PR #506, opened 2014-10-14
supporting higher order elements

norihiro-w:
I'd like to hear your opinion about how to implement the order-dependent edge nodes information. For example, Quad4 should return edges with 2 nodes and Quad8 return with 3 nodes. Since the information (topological rule) is currently stored in static member variable `_edge_nodes`, it is not possible to change its content and size in derived classes.  

My current approach implemented in this PR is passing the information as an additional template parameter to element classes. Let me know if you have better idea.

please ignore Line3 as it doesn't have the problem.


endJunction:
Summary of short discussion:
We prefer the second idea, though there are more changes of the codebase.

PR #542, opened 2014-12-03
MeshLib add properties

TomFischer:
This PR is based on the discussion in PR #531:
- It adds an implemention to manage properties within the Mesh.
- The properties are associated to specific mesh items like mesh nodes or mesh
  elements via an enum class (see enum class MeshItemType).
- The properties are not restricted to a particular data type, you can use several types like double, vector or matrix
  types at run time together (see tests).
- There is not a possibility to get the value data type.

### DIFF
--- a/MeshLib/Properties.h
+++ b/MeshLib/Properties.h
@@ -42,9 +42,10 @@ public:
 	/// Method to get a vector of property values.
 	template <typename T>
 	boost::optional<std::vector<T> const&>
-	getProperty(std::string const& name) const
+	getProperty(std::string const& name,
+		MeshItemType mesh_item_type = MeshItemType::Cell) const
 	{
-		PropertyKeyType property_key(name, MeshItemType::Cell);
+		PropertyKeyType property_key(name, mesh_item_type);
 		std::map<PropertyKeyType, boost::any>::const_iterator it(
 			_properties.find(property_key)
 		);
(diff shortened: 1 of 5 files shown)


MeshLib/Properties.h
```diff
@@ -0,0 +1,220 @@
...
+	/// mesh item and \f$n\f$ different property values.
+	/// The user has to ensure the correct usage of the vector later on.
+	/// @tparam T type of the property value
+	/// @param name the name of the property
+	/// @param mesh_item_type for instance node or element assigned properties
+	/// @return On success a reference to a PropertyVector packed into a
+	///   boost::optional else an empty boost::optional.
+	template <typename T>
+	boost::optional<PropertyVector<T> &>
+	createNewPropertyVector(std::string const& name, MeshItemType mesh_item_type)
```
TomFischer: I made class Properties a friend class of class PropertyVector and hid the constructors in protected scope of the class. So PropertyVector can only be constructed from Properties and not from the user.

norihiro-w:
:+1:

endJunction:
Very nice!

Two suggestions for future:
 - use make_pair, instead of pair ctor, which requires the types: `std::make_pair(a, b)` vs. `std::pair<TheAType, TheBType>(a, b)`.
 - Don't hessitate to make a type alias for often used structures, like for the `std::map<PropertyKey, boost::any>` and its `const_iterator` making code more compact (and abstract).

I like very much the `std::default_delete()` usage. No memory leaks and no compilation warnings.

PR #549, opened 2014-12-11
Node adjacency table

endJunction:
I have started a creation of a node adjacency table. The reason for this were mainly the endless memory reallocations in LIS. (When creating matrix there, the data is first written into a buffer, which has to be allocated). Here again the net gain is:
 - measured on the ExternalData/.../cube_1e4:  5.40s on master vs. 0.25s this branch.
 - measured on the ExternalData/.../cube_1e5:  820s on master vs.  2.50s this branch.

The (`createTable`) algorithm stores adjacent node ids to every node in a vector of vector structure. The nodes i and j are adjacent iff they are shared by an element (computed by `setNodeConnectedByElements()`.
Current implementation provides only the `getNodeDegree()` function, because I don't know the required access pattern for creation of CRS matrices etc.

I also add a `setMatrixSparsity()` function (empty implementation by default), which is currently used in the lis interface. (Unfortunatly a little too technical...)

Computation and application are in the `GroundwaterFlowProcess`.


MeshLib/NodeAdjacencyTable.h
```diff
@@ -0,0 +1,80 @@
...
+        return _data.size();
+    }
+
+    std::size_t getNodeDegree(std::size_t const node_id) const
+    {
+        return _data[node_id].size();
+    }
+
+    void createTable(std::vector<Node*> const& nodes)
+    {
```
wenqing: MeshComponentMap should be considered here or somewhere in the LIS solver initialization if a problem has more than one unknowns.

wenqing: or in setNodesConnectedByElements().

endJunction:
@wenqing Yes, what you say is correct. The current version is valid only for single variable problems. Together with the `MeshComponentMap` it shall become usable for multiple variables and/or meshes.
It's a starting point.

PR #550, opened 2015-01-05
Updates of the PETSc matrix vector interfaces: 1st splitted PR from 487 and 545.

wenqing:
To enhance the merging of #487 and #545, I split these two PRs into several PRs. This is the first one that mainly deals with the changes of the interfaces of PETSc matrix and vector.

A preliminary the performance test result of PETSc in ogs6 (using #545) for the solving  of a Poission type equation is given here (https://github.com/wenqing/Document/tree/master/Peformance_PETSC_FEM_ogs6).


endJunction:
Included in https://github.com/ufz/ogs/pull/561

PR #552, opened 2015-01-09
Fixed some compile errors due to a moc-bug in combination with Boost

bilke:
See https://bugreports.qt.io/browse/QTBUG-22829

This broke compilation on Mac OS, Clang 3.5, Boost 1.57.


TomFischer:
Compiled with g++ 4.9.1. :+1:

PR #560, opened 2015-01-16
MinGW support

bilke:
I finally figured out a way to setup MinGW (gcc) under Windows with all required dependencies (Boost, Vtk, Qt). Documentation will follow in the next days in the [developer guide](http://docs.opengeosys.org/docs/devguide).

The only required changes were to exclude the `MemWatch`-class and to another include path in the Boost finding logic in CMake. 

I also tested a new cross-platform IDE from JetBrains, [CLion](https://www.jetbrains.com/clion/) which requires gcc. It uses CMake as its project management system and also has a nice GDB frontend.


endJunction:
This implies also a jenkins mingw target?

PR #561, opened 2015-01-16
PETSc corrections

endJunction:
Collection of unrelated changes for PETSc code.

The MatSetType and Finalize do correct some bugs when using the matrix/vector.
The PETSc includes are marked as system includes and do not produce warnings while compilation.


bilke:
:+1:

norihiro-w:
:+1:

MathLib/LinAlg/PETSc/PETScLinearSolver.h
```diff
@@ -47,6 +49,9 @@ class PETScLinearSolver
 
         ~PETScLinearSolver()
         {
+            INFO("info: Time elapsed in PETSc ksp solver for equation: %g s.\n",
```
TomFischer: The macro INFO already outputs the string "info: " as default behaviour.

TomFischer:
Apart from tiny remark; :+1:

MathLib/LinAlg/PETSc/PETScLinearSolver.h
wenqing: @TomFischer Thanks.

PR #562, opened 2015-01-16
Node partitioned mesh reader

endJunction:
Based on the https://github.com/ufz/ogs/pull/561

The code can be tested by running
```sh
mpirun -np 3 bin/test_node_partitioned_mesh Tests/Data/NodePartitionedMesh/Binary/mesh_3d
diff Tests/Data/NodePartitionedMesh/Binary/mesh_3d_partition_0_expected.msh Tests/Data/NodePartitionedMesh/Binary/mesh_3d_partition_0.msh
```
for binary meshes. For ascii meshes substitute `Binary` with `ASCII`. The result is written into three files which are checked against expected ones.


FileIO/MPI_IO/NodePartitionedMeshReader.cpp
```diff
@@ -0,0 +1,521 @@
...
+
+    registerNodeDataMpiType();
+}
+
+NodePartitionedMeshReader::~NodePartitionedMeshReader()
+{
+    MPI_Type_free(&_mpi_node_type);
+}
+
+void NodePartitionedMeshReader::registerNodeDataMpiType()
```
wenqing: Change the name to registerMPI_NodeDataType (MPI must in capital).

FileIO/MPI_IO/NodePartitionedMeshReader.cpp
```diff
@@ -0,0 +1,521 @@
...
+            is_cfg >> std::ws;
+        }
+
+        MPI_Bcast(_mesh_info.data(), static_cast<int>(_mesh_info.size()),
+            MPI_LONG, 0, _mpi_comm);
+
+        //----------------------------------------------------------------------------------
+        // Read Nodes
+        if (!readCastNodesASCII(is_node, i, mesh_nodes, glb_node_ids))
+            return nullptr;
```
wenqing: Change ``return`` to ``break``. Then opened files can be closed.

FileIO/MPI_IO/NodePartitionedMeshReader.h
```diff
@@ -0,0 +1,252 @@
...
+                ASCII format or binary format.
+         \param file_name_base  Name of file to be read, and it must be base name without name extension.
+         \return                Pointer to Mesh object. If the creation of mesh object
+                                fails, return a null pointer.
+     */
+    MeshLib::NodePartitionedMesh* read(const std::string &file_name_base);
+
+private:
+    /// Pointer to MPI commumicator;
+    MPI_Comm _mpi_comm = MPI_COMM_WORLD;
```
norihiro-w: is it necessary to set initial value here? the variable is anyway overwritten in the constructor.

SimpleTests/MeshTests/MPI/NodePartitionedMeshTester.cpp
```diff
@@ -0,0 +1,120 @@
...
+             Distributed under a Modified BSD License.
+               See accompanying file LICENSE.txt or
+               http://www.opengeosys.org/project/license
+*/
+
+#include <iomanip>
+#include <fstream>
+#include <string>
+
+#ifdef USE_MPI
```
norihiro-w: i think all `#ifdef USE_MPI` in this file is unnecessary since this file is compiled only with MPI.

endJunction:
Improved and rebased. Correct copyright years.

FileIO/MPI_IO/NodePartitionedMeshReader.cpp
```diff
@@ -0,0 +1,521 @@
...
+            is_node >> node.index >> node.x >> node.y >> node.z >> std::ws;
+        }
+
+        if(part_id == 0)
+            setNodes(nodes, mesh_nodes, glb_node_ids);
+        else
+            MPI_Send(nodes.data(), static_cast<int>(_mesh_info.nodes),
+                _mpi_node_type, part_id, message_tag, _mpi_comm);
+    }
+    else if(part_id > 0 && _mpi_rank == part_id)
```
TomFischer: Condition ```_mpi_rank == part_id``` is enough since it holds ```_mpi_rank > 0``` true here.

FileIO/MPI_IO/NodePartitionedMeshReader.cpp
```diff
@@ -0,0 +1,521 @@
...
+            if(!process_ghost)
+                mesh_elems.resize(
+                    _mesh_info.regular_elements + _mesh_info.ghost_elements);
+            setElements(mesh_nodes, elem_data, mesh_elems, process_ghost);
+        }
+        else
+            MPI_Send(elem_data.data(), static_cast<int>(data_size), MPI_LONG,
+                part_id, message_tag, _mpi_comm);
+    }
+    else if(part_id > 0 && _mpi_rank == part_id)
```
TomFischer: See comment above.

FileIO/MPI_IO/NodePartitionedMeshReader.cpp
```diff
@@ -0,0 +1,521 @@
...
+    MeshLib::NodePartitionedMesh* np_mesh = nullptr;
+    std::vector<MeshLib::Node*> mesh_nodes;
+    std::vector<unsigned long> glb_node_ids;
+    std::vector<MeshLib::Element*> mesh_elems;
+
+    for(int i = 0; i < _mpi_comm_size; i++)
+    {
+        if(_mpi_rank == 0)
+        {
+            INFO("-->Parallel reading the partitioned mesh: ");
```
TomFischer: In contrast to the info statement in the following only the meta info is read.

TomFischer: As far as I understand the methods ```readCastNodesASCII()``` and ```readCastElemsASCII()``` the whole data is read from rank 0 and then appropriate portions of it are send to the other ranks. So the file is not really read in parallel, isn't it?

FileIO/MPI_IO/NodePartitionedMeshReader.cpp
```diff
@@ -0,0 +1,521 @@
...
+void NodePartitionedMeshReader::readElementASCII(std::ifstream &ins,
+    std::vector<unsigned long>& elem_data, const bool ghost) const
+{
+    // Set number of elements.
+    unsigned long const ne =
+        ghost ? _mesh_info.ghost_elements : _mesh_info.regular_elements;
+    unsigned long counter = ne;
+    for(unsigned long j = 0; j < ne; j++)
+    {
+        elem_data[j] = counter;
```
TomFischer: I do not understand this line.

FileIO/MPI_IO/NodePartitionedMeshReader.cpp
wenqing: @TomFischer  is right.

FileIO/MPI_IO/NodePartitionedMeshReader.cpp
wenqing: Yes, it is. The ASCII input is for test purpose. With that only rank 0 takes charge of reading data and cast them. I wanted to remove this functionality before because it is less usage but it is gained more coding and review efforts.  On the contrary binary input is purely parallelized for practical application.

FileIO/MPI_IO/NodePartitionedMeshReader.cpp
wenqing: Look at index ``j`` here, and the initial value of ``counter=ne``, you will see that:
*  from 0 to ``ne``,  ``elem_data[j] = counter``, records the start position of data of each element.
* element data in array  ``elem_data`` starts from position ``ne``, the initial value of ``counter``.

FileIO/MPI_IO/NodePartitionedMeshReader.cpp
```diff
@@ -0,0 +1,521 @@
...
+        _mesh_info.active_nodes);
+}
+
+void NodePartitionedMeshReader::readElementASCII(std::ifstream &ins,
+    std::vector<unsigned long>& elem_data, const bool ghost) const
+{
+    // Set number of elements.
+    unsigned long const ne =
+        ghost ? _mesh_info.ghost_elements : _mesh_info.regular_elements;
+    unsigned long counter = ne;
```
TomFischer: Why is counter intialized with number of elements?

wenqing: From entry ID ``ne``, the number of element, array ``elem_data`` saves the element data.  For any j in  ``[0: ne-1]``, array ``elem_data[j]` saves the starting positions of the data for each element (also see my reply below).

FileIO/MPI_IO/NodePartitionedMeshReader.cpp
TomFischer: What information is stored in ```elem_data[0, ..., ne-1]```?

TomFischer: @wenqing Thanks for the explanation! Now I understand. Before it was not clear to me that the entries [0,ne-1[ are used as offsets. It would be very good to add this to the documentation of the method setElements in the header file and maybe mention here that the format elem_data is described there.

FileIO/MPI_IO/NodePartitionedMeshReader.cpp
```diff
@@ -0,0 +1,521 @@
...
+{
+    // Number of elements, ether ghost or regular
+    unsigned long const ne =
+        ghost ? _mesh_info.ghost_elements : _mesh_info.regular_elements;
+    unsigned long const id_offset =
+        ghost ? _mesh_info.regular_elements : 0;
+
+    for(unsigned long i = 0; i < ne; i++)
+    {
+        unsigned long counter = elem_data[i];
```
TomFischer: I would prefer ```offset``` instead of ```counter```.

FileIO/MPI_IO/NodePartitionedMeshReader.h
```diff
@@ -0,0 +1,252 @@
...
+        std::size_t index;     ///< Global node index.
+        double x;
+        double y;
+        double z;
+    };
+
+    /// Define MPI data type for NodeData struct.
+    void registerNodeDataMpiType();
+
+    /// A collection of integers that congfigure the partitioned mesh data.
```
TomFischer: Spelling: congfigure -> configure

FileIO/MPI_IO/NodePartitionedMeshReader.cpp
```diff
@@ -0,0 +1,521 @@
...
+            elem_nodes[k] = mesh_nodes[ elem_data[counter++] ];
+
+        MeshLib::Element* elem = nullptr;
+
+        // The element types below are defined by the mesh_partition tool
+        // available at https://github.com/ufz/mesh_partition .
+        switch(e_type)
+        {
+        case 1:
+            elem = new MeshLib::Line(elem_nodes, mat_idx);
```
TomFischer: Why not ```mesh_elems[i+id_offset] = new MeshLib::Line(elem_nodes, mat_idx);```? The same for all the other elements. Then line 489 and 518 can be deleted. IMHO the readability would be the same.

TomFischer:
:+1: Apart from small remarks :fast_forward:

CMakeLists.txt
```diff
@@ -156,11 +156,17 @@ IF( OGS_BUILD_TESTS AND NOT IS_SUBPROJECT )
...
+
+	IF(OGS_USE_MPI)
+		ADD_SUBDIRECTORY( SimpleTests/MeshTests/MPI )
+	ELSE(OGS_USE_MPI)
+		ADD_SUBDIRECTORY( SimpleTests/MatrixTests EXCLUDE_FROM_ALL )
+		ADD_SUBDIRECTORY( SimpleTests/MeshTests EXCLUDE_FROM_ALL )
+		IF(NOT MSVC AND BLAS_FOUND AND LAPACK_FOUND)
+			ADD_SUBDIRECTORY( SimpleTests/SolverTests EXCLUDE_FROM_ALL )
+		ENDIF()
+	ENDIF(OGS_USE_MPI)
```
bilke: In general it is sufficient to omit the condition in the `ELSE()`- and `ENDIF()`-statements. This was required in old CMake versions.

FileIO/MPI_IO/NodePartitionedMeshReader.h
```diff
@@ -0,0 +1,252 @@
...
+                and return a pointer to it. Data files are either in
+                ASCII format or binary format.
+         \param file_name_base  Name of file to be read, and it must be base name without name extension.
+         \return                Pointer to Mesh object. If the creation of mesh object
+                                fails, return a null pointer.
+     */
+    MeshLib::NodePartitionedMesh* read(const std::string &file_name_base);
+
+private:
+    /// Pointer to MPI commumicator;
```
bilke: Typo: `communicator`

bilke:
:+1:

wenqing:
I will have an update concerning to the above comments.

endJunction:
@wenqing Now you can start preparation of the FEM part based on current master branch.

wenqing:
@endJunction There will be several small PRs soon.

PR #563, opened 2015-01-19
OGS version information is now handled by its git tag

bilke:
When the current commit is also a tag then the version is the tag name,
e.g. `6.0.1`. When the current commit is not a tag then the ogs version
is the last tag + number of commits since then + current commit short
hash, e.g. `6.0.1-140-g80ac475`.

This info is available in `BaseLib::BuildInfo::git_describe`. Use it to
uniquely set the OGS version.


TomFischer:
Looks good.

PR #566, opened 2015-01-20
New inheritance structure for Node

TomFischer:
The reason for this PR is that class ```MeshLib::Node``` is derived from ```GeoLib::GeoObject```. The consequence is that objects of class ```MeshLib::Node``` needs more memory than necessary.

In this PR class ```MeshLib::Node``` is directly derived from the new class MathPoint. Node objects are not anymore derived from GeoObject. Furthermore I fixed some small bugs.


MathLib/MathPoint.h
```diff
@@ -0,0 +1,44 @@
...
+ * lexicographical comparison of points taking an epsilon into account
+ * @param p0 first input MathPoint
+ * @param p1 first input MathPoint
+ * @param tol tolerance (if in the comparison operation the property fabs(p0[k] - p1[k]) < tol
+ *     holds for the k-th coordinate the points are assumed the be equal in this coordinate)
+ * @return true, if p0 is lexicographically smaller than p1
+ */
+bool lessEq(const MathLib::MathPoint& p0,
+            const MathLib::MathPoint& p1,
+            double tol = std::numeric_limits<double>::epsilon());
```
norihiro-w: need to include <limit>

MathLib/MathPoint.h
```diff
@@ -0,0 +1,44 @@
...
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef MATHPOINT_H_
+#define MATHPOINT_H_
+
+#include <cmath>
```
norihiro-w: where is this include used?

MathLib/MathPoint.h
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * lexicographic comparison of points
+ */
+bool operator<= (MathLib::MathPoint const & p0, MathLib::MathPoint const & p1);
+
+/**
+ * lexicographical comparison of points taking an epsilon into account
+ * @param p0 first input MathPoint
+ * @param p1 first input MathPoint
```
norihiro-w: first -> second

rinkk:
I have mixed feelings about this PR because my impression is that it confuses quite a few things.

First and foremost I think that the name ```MathPoint``` does not make a lot of sense. I also think it is weird to see algorithms in classes such as ```Polyline``` handle ```MathPoint```-objects instead of ```GeoPoints``` or just ```Points``` (e.g. if I need the distance of a point to a line -- which is a method in the ```Polyline``` class -- I have to use a ```MathLib::MathPoint```-object and a ```GeoLib::Polyline```-object ... I think this is a very strange naming policy)

I took a quick look at the code and I think that after the changes the ```Point```-class is pretty much useless. It's only benefit is that it also inherits from ```GeoObject``` in addition to the new ```MathPoint```. However, after ditching the old BC-classes, as far as I can see ```GeoObject``` is only used *once* in all the source code (in ```MeshNodeSearcher::getMeshNodeIDs()```.
In addition, checking the hierarchy for someone who isn't a core developer ist now even more complicated than before, e.g.
PointWithID -> Point -> MathPoint -> TemplatePoint
(you basically removed one layer from the MeshLib::Node hierarchy to add one layer to the GeoLib::Point hierarchy)

I would propose the following:
* rename ```MathPoint``` to ```Point```
* either rename ```Point``` to ```GeoPoint``` or remove the ```Point```-class and the ```GeoObject```-class completly

Also, can you quantify how substancial the mem-savings are? (e.g. for a mesh with 1 mio nodes?) ... Is it worth all the trouble?

endJunction:
@rinkk The naming could indeed be improved.
While it is non-optimal state (MathPoint and GeoPoint are almost identical), this is one of the efforts to untangle the class hierarchy; An improvement here is that the "MathPoint" does not require the dependency on GeoObject, which is now removed. As a nice side-effect we have smaller Node class, which result for a 10^6 elements quad mesh in memory saving around 3% (15MB/472MB).

Removing of GeoObject class leads to more changes, than the mesh node searcher only (especially in the DE); I would not touch it in this PR.

@TomFischer What is the difference between the (new) MathPoint and GeoPoint?

MathLib/MathPoint.h
TomFischer: Removed include.

TomFischer:
> What is the difference between the (new) MathPoint and GeoPoint?

GeoPoint does not exist anymore. GeoPoint was a class template and GeoLib::Point was an instance of the class template. In this PR I remove the template stuff and renamed GeoPoint to Point. GeoPoint was derived from GeoObject, the Point is derived from GeoObject.

> I also think it is weird to see algorithms in classes such as Polyline handle MathPoint-objects instead of GeoPoints or just Points (e.g. if I need the distance of a point to a line -- which is a method in the Polyline class -- I have to use a MathLib::MathPoint-object and a GeoLib::Polyline-object ... I think this is a very strange naming policy)

Most (propably all) places in GeoLib where MathPoint is used as data type are called from MeshLib. (MeshLib::Node are used for instance in Polyline::getDistanceAlonPolyline().) MathPoint has already all information needed to perform most of the algorithms in GeoLib.

> I took a quick look at the code and I think that after the changes the Point-class is pretty much useless. It's only benefit is that it also inherits from GeoObject in addition to the new MathPoint. 

The inheritance from GeoObject is the difference! It is used at several locations in OGS.

> However, after ditching the old BC-classes, as far as I can see GeoObject is only used once in all the source code (in MeshNodeSearcher::getMeshNodeIDs().

I searched for the string 'GeoObject' and found it in 37 files (h and cpp). The GUI uses the inheritance of GeoObject a lot :smile: If this feature should be remove in the future we should sit together and discuss the consequences.

**Concering the depths of the hierarchys**
- The new inheritance hierarchy is:
  PointWithID -> Point -> GeoObject and MathPoint = TemplatePoint<double,3>.
- This hierarchy is of the same depth as before:
  PointWithID -> Point = GeoPoint<double,3> -> GeoObject and TemplatePoint<double,3>.
- For the node the inheritance hierarchy is shorter. This was the aim of the PR :smile: 

A renaming is always possible, but this would touch even more files. I suggest to do it in another PR. The advantage of a "renaming only PR" is that the review will be easy.

rinkk:
My main problem with this PR *is* that the naming is confusing.
If we merge this now we have lots of ```MathPoints``` in the code and the follow-up "renaming-PR" would probably never happen, so we'd be stuck with it ....

TomFischer:
@rinkk Please propose a proper name schema. I will do the renaming asap.

rinkk:
Also, the hierarchy thing isn't quite true:
GeoPoint was just a typedef on Point and as far as I can see it was used pretty much nowhere at all.
PointWithID was derived from Point.

But that doesn't actually matter for this PR, I have no problem with shortening the hierarchy for nodes. (I am just wondering why you do it now when I would think there are more pressing issues ....)

wenqing:
Same confusion about the naming as the mentioned by @rinkk. Also the hierarchy. I guess MathPoint refers to the point definition in analytic geometry, e.g. (x,y,z). Then one may think the hierarchy could be Point (MathPoint)->PointWithID or -> any other point based objects. 

Since it works, I only suggest to document the depths of the hierarchy of such  point classes somewhere.

rinkk:
the alternative is to rename MathPoint to Point3D, this is probably the most straightforward solution.

rinkk:
other than the renaming stuff, I have no further comments

bilke:
:+1:

endJunction:
I'd also suggest to rename MathLib::MathPoint to MathLib::Point3 (or Point3d), because its dimension is the main difference to the TemplatePoint.
GeoLib::Point to GeoLib::GeoPoint seems also OK, because it indicates the derivation from GeoObject. (Or GeoLib::GeoPoint3[d] for consistency).

I would also agree to name both types simply Point3 in their respective namespaces: MathLib::Point3 and GeoLib::Point3.

@TomFischer If the rename is very large, create a new PR on top of this.

TomFischer:
Commit c32a9ed is the first step clarify the name confusion. I propose to use another PR for a possible renaming the original GeoLib::Point to something more consistent or meaningful.

PR #568, opened 2015-01-20
Fix warnings and small bugs

TomFischer:
as titled


GeoLib/PointVec.cpp
```diff
@@ -16,7 +16,6 @@
 #include "logog/include/logog.hpp"
```
rinkk: Please add <numeric> header to make this compile under Windows

GeoLib/MinimalBoundingSphere.cpp
```diff
@@ -154,7 +154,9 @@ MinimalBoundingSphere MinimalBoundingSphere::recurseCalculation(std::vector<GeoL
             if (i>start_idx)
             {
                 GeoLib::Point* tmp = sphere_points[start_idx+i];
-                std::copy(sphere_points.begin() + start_idx, sphere_points.begin() + (start_idx + i), sphere_points.begin() + (start_idx + 1));
+                std::copy(sphere_points.begin()+static_cast<std::vector<GeoLib::Point*>::iterator::difference_type>(start_idx),
+                    sphere_points.begin()+static_cast<std::vector<GeoLib::Point*>::iterator::difference_type>(start_idx + i),
+                    sphere_points.begin()+static_cast<std::vector<GeoLib::Point*>::iterator::difference_type>(start_idx + 1));
```
rinkk: Can you please explain what is the benefit of this change? It looks ... complicated ...

GeoLib/GEOObjects.cpp
```diff
@@ -90,29 +91,21 @@ bool GEOObjects::appendPoint(Point* point, std::string const &name, std::size_t&
 const std::vector<Point*>* GEOObjects::getPointVec(const std::string &name) const
 {
 	int idx = this->exists(name);
```
rinkk: we could change the ```exists()```-method to return ```std::numeric_limits::max()``` if the name does not exits, this would save some conversions here and other methods where this is checked.

rinkk:
For some reason I cannot see Dimas comments in the code, so I cannot answer directly. Concerning the comment about typedef: I am not a huge fan of typedefs because they are complicated to look up if you do not already know what they mean. Specifically, I don't think it makes sense to do a typedef here for something that is used only once.

Additional question: does the originally code in ```MinimalBoundingSphere``` result in a GCC-warning? Because it does not under Windows and simply adding a ```std::size_t``` is pretty much the default solution to this you find everywhere on the net?!

endJunction:
@rinkk A type alias (or typedef) inside a single function is a good solution to improve readability:
```c++
using DiffType = std::vector<GeoLib::Point*>::iterator::difference_type;
...
sphere_points.begin()+static_cast<DiffType>(start_idx + i),
...
```

endJunction:
The static_casts are preferable over the c-casts.  e.g. this [SO question](http://stackoverflow.com/questions/103512/in-c-why-use-static-castintx-instead-of-intx).

GeoLib/MinimalBoundingSphere.cpp
TomFischer: To avoid a warning about the types the std::size_t has to be static_casted to std::vector<GeoLib::Point*>::iterator::difference_type. I will use a type alias as @endJunction suggested in order to make it readable.

GeoLib/GEOObjects.cpp
TomFischer: Should we do this in a separate PR?

TomFischer:
> Additional question: does the originally code in MinimalBoundingSphere result in a GCC-warning?

Yes. Clang also gives a warning. By the way: Clang reports hundreds of warnings!

endJunction:
:fast_forward:

rinkk:
I am aware what a typedef does. I am just saying that I don't think it is useful if it is used at only one line of code.

Also, I still see the benefit of using ```difference_type``` at all as it seems to be only a typedef itself from what I gather from c++ documentation. The only practical result it seems to serve here is to make the code three times longer (i.e. one line to three lines) and as such making the line actually much less readable than it was before ....

TomFischer:
> The only practical result it seems to serve here is to make the code three times longer (i.e. one line to three lines) and as such making the line actually much less readable than it was before ....

This is necessary to fix a warning, since ```difference_type != std::size_t```. The goal of the PR is to reduce the number of warnings.

GeoLib/GEOObjects.cpp
rinkk: we can do this in a seperate PR. the problem is, that you are introducing quite a few casts now (basically everywhere where exists() is used) and I would assume that it is easier to track the changes now.
the actual difference would only be the return value and type of ```exists()``` and changing the test ```if (i>0)```  to ```if (i<std::numeric_limits<std:size_t>::max())```

TomFischer: Changed as suggested in  commit b3daf65.

PR #571, opened 2015-01-26
Cleanup

norihiro-w:
Some cleanup before merging new implementation of mesh element classes


endJunction:
I always agree to code removal.
Karsten?

TomFischer:
:+1:

PR #572, opened 2015-01-26
[MeshLib] Replacement of element classes with new implementation using TemplateElement

norihiro-w:
This PR is based on discussion in #506 and introduces new implementations of mesh element classes by combining TemplateElement class and element-specific topological/geometrical rules. 

Note: It includes two commits in #571


MeshLib/Elements/CellRule.cpp
```diff
@@ -0,0 +1,41 @@
...
+	const MathLib::Vector3 c (e->getCenterOfGravity());
+	const unsigned nFaces (e->getNFaces());
+	for (unsigned j=0; j<nFaces; ++j)
+	{
+		MeshLib::Face const*const face (dynamic_cast<const MeshLib::Face*>(e->getFace(j)));
+		const MeshLib::Node x (*(face->getNode(1)));
+		const MathLib::Vector3 cx (c, x);
+		const double s = MathLib::scalarProduct(face->getSurfaceNormal(), cx);
+		delete face;
+		if (s >= 0)
```
wenqing: Real comparison.

MeshLib/Elements/Cell.h
```diff
@@ -26,29 +25,13 @@ namespace MeshLib {
...
-
-	/// Returns the length, area or volume of a 1D, 2D or 3D element
-	double getContent() const { return _volume; }
-
-	/// Get dimension of the mesh element.
-	unsigned getDimension() const { return dimension; }
-
 	/// Get the volume of this 3d element.
-	virtual double getVolume() const { return _volume; }
+	virtual double getVolume() const { return getContent(); }
```
wenqing: CellSize?

MeshLib/Elements/Edge.h
```diff
@@ -0,0 +1,52 @@
...
+
+	/**
+		* Checks if the node order of an element is correct by testing surface normals.
+		* For 1D elements this always returns true.
+		*/
+	virtual bool testElementNodeOrder() const { return true; }
+
+protected:
+	/// Constructor for a generic mesh element without an array of mesh nodes.
+	Edge(unsigned value = 0, std::size_t id = std::numeric_limits<std::size_t>::max()) : Element(value, id) {}
```
wenqing: Needs doc. for arguments.

MeshLib/Elements/Element.h
```diff
@@ -207,19 +207,19 @@ class Element
...
 
-	/// Return a specific edge node.
-	virtual Node* getEdgeNode(unsigned edge_id, unsigned node_id) const = 0;
-
 	/// Sets the element ID.
 	virtual void setID(std::size_t id) { this->_id = id; }
 
 	Node** _nodes;
 	std::size_t _id;
+	double _content;
```
wenqing: ``_geo_size_quantity``. Or give a comment as ``Length for 1D, area for 2D, and volume for 3D, respectively.``

MeshLib/Elements/HexRule8.cpp
```diff
@@ -0,0 +1,119 @@
...
+#include "Quad.h"
+#include "Line.h"
+
+namespace MeshLib {
+
+const unsigned HexRule8::n_all_nodes;
+
+const unsigned HexRule8::n_base_nodes;
+
+const unsigned HexRule8::face_nodes[6][4] =
```
wenqing: Can it be put  constexpr modifer?

MeshLib/Elements/HexRule8.cpp
```diff
@@ -0,0 +1,119 @@
...
+	{4, 7}, // Edge 7
+	{0, 4}, // Edge 8
+	{1, 5}, // Edge 9
+	{2, 6}, // Edge 10
+	{3, 7}  // Edge 11
+};
+
+const Element* HexRule8::getFace(const Element* e, unsigned i)
+{
+	if (i < n_faces)
```
wenqing: this if-condition could be in ``assert`` for debugging only.

MeshLib/Elements/HexRule8.cpp
```diff
@@ -0,0 +1,119 @@
...
+			nodes[j] = const_cast<Node*>(e->getNode(face_nodes[i][j]));
+		return new Quad(nodes);
+	}
+	ERR("Error in MeshLib::Element::getFace() - Index %d does not exist.", i);
+	return nullptr;
+}
+
+double HexRule8::computeVolume(Node const* const* _nodes)
+{
+	return GeoLib::calcTetrahedronVolume(_nodes[4]->getCoords(), _nodes[7]->getCoords(), _nodes[5]->getCoords(), _nodes[0]->getCoords())
```
wenqing: some getCoords() is called more than twice. Is it better to put to a constant pointer in advance as ``const double *p7 = _nodes[7]->getCoords()`` .

MeshLib/Elements/HexRule8.h
```diff
@@ -0,0 +1,110 @@
...
+ *    0-----------1
+ *          0
+ *
+ * @endcode
+ */
+class HexRule8 : public CellRule
+{
+public:
+	/// Constant: The number of base nodes for this element
+	static const unsigned n_base_nodes = 8u;
```
wenqing: ``static constexpr``?

MeshLib/Elements/PrismRule6.cpp
```diff
@@ -0,0 +1,123 @@
...
+{
+	if (i<5)
+		return n_face_nodes[i];
+	ERR("Error in MeshLib::Element::getNFaceNodes() - Index %d does not exist.", i);
+	return 0;
+}
+
+const Element* PrismRule6::getFace(const Element* e, unsigned i)
+{
+	if (i < n_faces)
```
wenqing: assert()

MeshLib/Elements/PyramidRule5.cpp
```diff
@@ -0,0 +1,120 @@
...
+{
+	if (i<5)
+		return n_face_nodes[i];
+	ERR("Error in MeshLib::Element::getNFaceNodes() - Index %d does not exist.", i);
+	return 0;
+}
+
+const Element* PyramidRule5::getFace(const Element* e, unsigned i)
+{
+	if (i<e->getNFaces())
```
wenqing: assert()

MeshLib/Elements/LineRule2.h
```diff
@@ -0,0 +1,90 @@
...
+	 * @param eps tolerance for numerical algorithm used or computing the property
+	 * @return true if the point is not outside the element, false otherwise
+	 */
+	static bool isPntInElement(Node const* const* _nodes, MathLib::Point3d const& pnt, double eps);
+
+	/**
+	 * Tests if the element is geometrically valid.
+	 * @param check_zero_volume indicates if volume == 0 should be checked
+	 */
+	static ElementErrorCode validate(const Element* e);
```
wenqing: I think this sort of checking is only performed once. Therefore, it may be unnecessary to have this functionality as member, and is can be arranged in a member of mesh class.

wenqing:
I think the current implementation has a flexibility in handling different element orders.

MeshLib/Elements/CellRule.cpp
norihiro-w: @TomFischer @rinkk how do you think?

MeshLib/Elements/Cell.h
norihiro-w: Are you suggesting renaming getVolume() with getCellSize()? If so, why? What we return here is volume of a cell (which is 3D).

MeshLib/Elements/Element.h
norihiro-w: I'll give the comment now, because the name is already used in ogs6. you can discuss it in another PR.

MeshLib/Elements/Cell.h
TomFischer: IMHO getContent() is readable.

MeshLib/Elements/CellRule.cpp
TomFischer: I can not see a problem here. @wenqing What are your concerns?

MeshLib/Elements/CellRule.cpp
```diff
@@ -0,0 +1,41 @@
...
+const unsigned CellRule::dimension;
+
+bool CellRule::testElementNodeOrder(const Element* e)
+{
+	const MathLib::Vector3 c (e->getCenterOfGravity());
+	const unsigned nFaces (e->getNFaces());
+	for (unsigned j=0; j<nFaces; ++j)
+	{
+		MeshLib::Face const*const face (dynamic_cast<const MeshLib::Face*>(e->getFace(j)));
+		const MeshLib::Node x (*(face->getNode(1)));
```
TomFischer: I am curious: Is there a special reason to take node 1 of the face?

MeshLib/Elements/HexRule8.cpp
```diff
@@ -0,0 +1,119 @@
...
+	{5, 6}, // Edge 5
+	{6, 7}, // Edge 6
+	{4, 7}, // Edge 7
+	{0, 4}, // Edge 8
+	{1, 5}, // Edge 9
+	{2, 6}, // Edge 10
+	{3, 7}  // Edge 11
+};
+
+const Element* HexRule8::getFace(const Element* e, unsigned i)
```
TomFischer: I wonder if it is possible take a hexahedron here instead of a general element as an argument? In this case the code could be more specific, i.e. the number of face nodes is already known (=4). Or in line 57 one could use ```std::array<Node*, 4>``` for storage instead using ```new```. But I do not know if it is possible.

MeshLib/Elements/HexRule8.cpp
TomFischer: Maybe we can overload the ```GeoLib::calcTetrahedronVolume()``` for ```MeshLib::Node``` in order to make the code better readable?

TomFischer: See [PR 573](https://github.com/ufz/ogs/pull/573).

MeshLib/Elements/HexRule8.h
```diff
@@ -0,0 +1,110 @@
...
+
+	/// Get the number of nodes for face i.
+	static unsigned getNFaceNodes(unsigned /*i*/) { return 4; }
+
+	/// Returns the i-th face of the element.
+	static const Element* getFace(const Element* e, unsigned i);
+
+	/**
+	 * Checks if a point is inside the element.
+	 * @param pnt a 3D GeoLib::Point object
```
TomFischer: Docu: ```GeoLib::Point``` -> ```MathLib::Point3d```

MeshLib/Elements/CellRule.cpp
wenqing: 0.

MeshLib/Elements/CellRule.cpp
norihiro-w: i have no idea. please ask @rinkk

rinkk: This is the documentation I wrote:
```c++
	 * Checks if the node order of an element is correct by testing surface normals.
	 * For 3D elements true is returned if the normals of all faces points away from the centre of 
	 * the element.
	 * Note: This method might give wrong results if something else is wrong with the element 
	 * (non-planar faces, non-convex geometry, possibly zero volume) which causes the calculated 
	 * center of gravity to lie outside of the actual element
```
I *think* I chose node 1 because that way *all* nodes are used for the test at some point, while for node 0 at least one node in every element type would be used for checking twice and one wouldn't be checked at all. (based on the definition of the ```_face_nodes``` variable)

MeshLib/Elements/CellRule.cpp
rinkk: we just check if two vectors are pointing in *roughly* the same directing. ```s``` should be nowhere near ```0```, if it is the element is degenerated. the test is still correct here because we are testing just for node order so technically ```s=0``` is a valid value, even if other tests will fail because of that. (I am not 100% sure without looking up the theory again, but ```s=0``` almost certainly means this will be a zero volume element)

MeshLib/Elements/HexRule8.cpp
rinkk: I am confused ... what exactly is ```e``` here? Shouldn't ```getFace()``` be called on an element and ```e``` would be basically ```this```?

rinkk: @TomFischer this is a general method overloaded from the base class so the signature needs to be general, too.

MeshLib/Elements/HexRule8.cpp
rinkk: no, if you don't test this the programm will crash if ```ì < 8``` ...
in this implementation it will just return a ```nullptr``` for which you can test, so everything will be fine.

MeshLib/Elements/HexRule8.cpp
rinkk: the PR is okay, but I am confident that the compiler will do something clever here. getcoords is a const method that just returns a value, so it is very likely this will be replaced here during compiling.

MeshLib/Elements/LineRule2.h
rinkk: this may be called at any time from the GUI. 
also, no memory is used for this, as the result is not stored. please leave this as it is.

MeshLib/Elements/PrismRule6.cpp
rinkk: nope!

MeshLib/Elements/PyramidRule5.cpp
rinkk: no

rinkk:
For what it's worth: everything works fine under windows and with gui.
If I see this correctly, debugging will be more difficult using this implementation. (I don't like this fact but I suppose that it is not a priority.)

A general comment:
You should probably also adjust the schematics for the higher order elements, e.g. for Quad9:
```c++
 *              
 *        3-----6-----2
 *        |           |
 *        |           |
 *        7     9     5
 *        |           |
 *        |           |
 *        0-----4-----1
 *              
```

rinkk:
Another question:
I see that your higher order implementation-rules are currently very small.
Can you somehow use methods from another rule (e.g. the ```getVolume()``` method) or do you simply need none of that for higher order methods?

MeshLib/Elements/CellRule.cpp
norihiro-w: @rinkk shall I add your comment (about checking node 1) into the source so that we can avoid repating of the same question in the future?

MeshLib/Elements/HexRule8.cpp
norihiro-w: @rinkk no, this class is not derived from Element. It is collection of all element specific rules given as either static const variables or static const functions.

norihiro-w: @TomFischer I'll check if it's possible to specialize it.

norihiro-w:
@rinkk 
- you are right and agree with you. debugging is more difficult. I also don't like it. Any alternative idea for easier debugging is welcome
- Thanks, I'll modify the schematics for the higher order elements
- With this implementation, it is possible to override getVolume() for non-linear shape elements (e.g. quad with curved edgs). I'm not doing this at the moment, because so far we are using only linear shape. Someone like TN may want to use non-linear shape for representing coverns.

MeshLib/Elements/CellRule.cpp
rinkk: good idea, thank you :)

rinkk:
thanks for the info

MeshLib/Elements/CellRule.h
```diff
@@ -0,0 +1,36 @@
...
+class CellRule
+{
+public:
+	/// Constant: Dimension of this mesh element
+	static const unsigned dimension = 3u;
+
+	/**
+	 * Checks if the node order of an element is correct by testing surface normals.
+	 */
+	static bool testElementNodeOrder(const Element* e);
```
norihiro-w: the same function exists in Cell class

MeshLib/Elements/HexRule8.cpp
norihiro-w: @rinkk is constexprt supported in MSVC?

norihiro-w:
rebased, changed according to the comments

MeshLib/Elements/HexRule8.cpp
norihiro-w: @wenqing it seems current MSVC does not support constexprt for arrays

MeshLib/Elements/QuadRule8.h
```diff
@@ -0,0 +1,57 @@
...
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef QUADRULE8_H_
+#define QUADRULE8_H_
+
+#include "MeshLib/MeshEnums.h"
+#include "Element.h"
```
norihiro-w: not needed

MeshLib/Elements/QuadRule9.h
```diff
@@ -0,0 +1,49 @@
...
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef QUADRULE9_H_
+#define QUADRULE9_H_
+
+#include "MeshLib/MeshEnums.h"
+#include "Element.h"
```
norihiro-w: not needed

MeshLib/Elements/HexRule8.cpp
TomFischer: Nice! :-)

norihiro-w:
maybe it is better to define rule functions in header files so that it helps optimization?

MeshLib/Elements/Face.h
```diff
@@ -32,25 +30,10 @@ class Face : public Element
...
-
-	/// Returns the face i of the element.
-	const Element* getFace(unsigned i) const { return this->getEdge(i); }
-
-	/// Get the number of nodes for face i.
-	unsigned getNFaceNodes(unsigned i) const { (void)i; return 2; }
-
-	/// 2D elements have no faces.
-	unsigned getNFaces() const { return 0; }
+	virtual double getArea() const { return _content; }
```
endJunction: I think `getArea` is nowhere used, and there is the new `getContent` method in the base class.

MeshLib/Elements/Element.h
```diff
@@ -84,7 +84,7 @@ class Element
 	virtual unsigned getDimension() const = 0;
 
 	/// Returns the i-th edge of the element.
-	const Element* getEdge(unsigned i) const;
```
endJunction: The implementation of `getEdge` is the same as in `EdgeReturn.h`, `LinearEdgeReturn::getEdge`.
I don't yet see the big picture, but maybe like in the `TemplateElement` class this should return
`LinearEdgeReturn::EdgeReturn::getEdge(this, i)`?

MeshLib/Elements/EdgeReturn.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include "logog/include/logog.hpp"
+
+#include "MeshLib/Node.h"
+#include "Element.h"
+#include "Line.h"
+
+namespace MeshLib
+{
+
+const Element* LinearEdgeReturn::getEdge(const Element* e, unsigned i)
```
endJunction: Note: The linear and quadratic implementations return a Line or Line3, which are derived from Edge, so the return type could be `Edge*` too (consistently with the function name getEdge).

MeshLib/Elements/Face.h
norihiro-w: @rinkk is it true that getArea is nowhere used? 
@endJunction getContent() is not new.

MeshLib/Elements/Element.h
norihiro-w: i'll remove implementaiton of getEdge in Element.

endJunction:
An idea, before I forget it.
The `Edge`, `Face`, and `Cell` classes mainly provide the `dimension` (the two special functions aside, which could be implemented elsewhere). Is it possible to move the `dimension` into `TemplateElement` template parameter list?

norihiro-w:
moving dimension into TemplateElement is possible. do you mean deleting edge, face, and cell elements?

rinkk:
I like the face/cell classes because it is a very intuitive/logical concept in my opinion.
Even if removing those would make some things a teeny-tiny bit more efficient, adding yet another template param imo makes the meshlib as a whole a bit harder to understand and putting the methods in the face and cell classes (where they belong!!) elsewhere makes them unintuitive to use ....

MeshLib/Elements/Face.h
rinkk: yes, ```getArea``` is not used ... it is one of the methods we included when we first implemented the MeshLib but in reality we always end up using ```getContent()``` instead ...

rinkk:
i just a discussion with Dima and it seems i didn't understand the new structure not compeletely.
if we move the information and methods from face and cell classes into facerule and cellrule, we will end up with a similar structure as we have now and could remove the current face and cell classes.

endJunction:
Now (in this PR) we have following structure:
```
Element <- Edge <- TemplateElement<Edge | Face | Cell, Rules>
  ^ ^- Face <--------/  |
  |___ Cell <-----------|
```
Removing Edge, Face, and Cell classes yields:
```
Element <- TemplateElement<Rules>
```
where the (corresponding) Rules contain the dimension and the methods of the original Edge, Face, and Cell classes.

norihiro-w:
removed Edge, Face and Cell classes and moved their functions to EdgeRule, FaceRule and CellRule

norihiro-w:
Follwoing Dima's suggestion, removed T_BASE parameter from TemplateElement class

endJunction:
Good work!
Someone please look one more time for typos etc. The class hierarchy and overall structure looks good. (And we have more functionality with less code in total!)

:fast_forward:

bilke:
:+1:

PR #574, opened 2015-01-29
Mesh hole detection

rinkk:
implements algorithm to detect "holes" within meshes (e.g. missing triangles in a 2D mesh) by using a kind of surface marching algorithm.
works for 2D and 3D meshes, as long as they don't have "weird" geometries, e.g. more than two 2D elements sharing an edge.
the test is included in gui and in nori's command line tool for mesh testing.

here's an example if you don't know what i am referring to by "holes in a mesh" (from inside Feng's Nankou study):
mesh_hole2

currently i am only returning the number of holes found in the mesh. in theory i also know the boundaries of the holes at some point. i am not sure if it'd be useful to return those, too?!

### DIFF
--- a/MeshLib/MeshQuality/MeshValidation.cpp
+++ b/MeshLib/MeshQuality/MeshValidation.cpp
@@ -176,9 +176,9 @@ unsigned MeshValidation::detectHoles(MeshLib::Mesh const& mesh)
 	return (--current_surface_id); 
 }
 
-void MeshValidation::trackSurface(MeshLib::Element const*const element, std::vector<unsigned> &sfc_idx, unsigned const current_index)
+void MeshValidation::trackSurface(MeshLib::Element const* element, std::vector<unsigned> &sfc_idx, unsigned const current_index)
 {
-	std::stack<MeshLib::Element const*const> elem_stack;
+	std::stack<MeshLib::Element const*> elem_stack;
 	elem_stack.push(element);
 	while (!elem_stack.empty())
 	{
@@ -188,7 +188,7 @@ void MeshValidation::trackSurface(MeshLib::Element const*const element, std::vec
 		std::size_t const n_neighbors (elem->getNNeighbors());
 		for (std::size_t i=0; i<n_neighbors; ++i)
 		{
-			MeshLib::Element const*const neighbor (elem->getNeighbor(i));
+			MeshLib::Element const* neighbor (elem->getNeighbor(i));
 			if ( neighbor != nullptr && sfc_idx[neighbor->getID()] == std::numeric_limits<unsigned>::max())
 				elem_stack.push(neighbor);
 		}
--- a/MeshLib/MeshQuality/MeshValidation.cpp
+++ b/MeshLib/MeshQuality/MeshValidation.cpp
@@ -161,12 +161,7 @@ unsigned MeshValidation::detectHoles(MeshLib::Mesh const& mesh)
 
 	while (it != sfc_idx.cend())
 	{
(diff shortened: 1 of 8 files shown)


rinkk:
can anyone give me a hint why the travis builds fail? 
i don't get what gcc is trying to tell me...

endJunction:
For the compilation error.
In short: one can not create `std::vector<int const> a;`.
Long: From the Standard, 17.6.3.5 [allocator.requirements], table 27: T (which is the type for the allocator, like the `std::allocator<T>`) can be any non-const object type.

Just remove the second `const` qualifier.

MeshLib/MeshQuality/MeshValidation.cpp
```diff
@@ -147,4 +150,49 @@ MeshValidation::ElementErrorCodeOutput(const std::vector<ElementErrorCode> &erro
...
+	MeshLib::Mesh* boundary_mesh (MeshSurfaceExtraction::getMeshBoundary(mesh));
+	std::vector<MeshLib::Element*> const& elements (boundary_mesh->getElements());
+
+	std::vector<unsigned> sfc_idx (elements.size(), std::numeric_limits<unsigned>::max());
+	unsigned current_surface_id (0);
+	std::vector<unsigned>::const_iterator it = sfc_idx.cbegin();
+
+	while (it != sfc_idx.cend())
+	{
+		int const idx = static_cast<int>(std::distance(sfc_idx.cbegin(), it));
```
norihiro-w: i am not sure in which case you get `idx<0` and `idx>elements.size()`. It looks for me that `idx<0` is impossible because `it` can take a value only in a range of [begin, end). you may get `idx>elements.size()` if `it==sfc_idx.cend()`, but the condition should be checked before calling distance().

MeshLib/MeshQuality/MeshValidation.h
```diff
@@ -65,7 +66,26 @@ class MeshValidation
 	static std::array<std::string, static_cast<std::size_t>(ElementErrorFlag::MaxValue)>
 		ElementErrorCodeOutput(const std::vector<ElementErrorCode> &error_codes);
 
+	/** 
+	 * Tests if holes are located within the mesh. 
+	 * In this context, a hole is a boundary of an element with no neighbor that cannot be reached from
+	 * the actual boundary of the mesh. Examples include a missing triangle in a 2D mesh or a missing 
+	 * Tetrahedron in a 3D mesh.
+	 * Note, that this method does not work when complex 3D structures are build from 2D mesh elements,
```
norihiro-w: this note should go to comments of getMeshBoundary()

rinkk:
> as probably you know it, having holes is sometimes intention

Yes, but if you *know* there is a hole in your mesh and the method also tells you that it found one, everything is still fine. It's just information, just like with zero volume elements. If you *want* to have them in there, just ignore the message :)

MeshLib/MeshQuality/MeshValidation.h
rinkk: mmmh ... I think this note makes sense here, as *this* method will give completely wrong results in the given case.
I will add an additional note to ```getMeshBoundary()``` because the result might not be exactly what the user is expecting. (It is debatable wether the result is correct or not in this case, depending on the expectations you have when you call the method).

MeshLib/MeshQuality/MeshValidation.cpp
rinkk: okay, i removed the error code now.
I know there was a reason I put it there but I cannot reproduce any problems now and it *might* have been necessary at some point during implementation. anyway, it's gone now ...

norihiro-w:
:+1:

PR #575, opened 2015-01-29
Surface extraction dialog

rinkk:
added a dialog to GUI to account for extended functionality introduced in PR #488

### DIFF
--- a/Applications/DataExplorer/DataView/MshView.cpp
+++ b/Applications/DataExplorer/DataView/MshView.cpp
@@ -25,6 +25,7 @@
 #include "MeshLib/Node.h"
 #include "MeshLayerEditDialog.h"
 #include "MeshValueEditDialog.h"
+#include "SurfaceExtractionDialog.h"
 #include "MshItem.h"
 #include "MshModel.h"
 #include "OGSError.h"
@@ -120,7 +121,7 @@ void MshView::contextMenuEvent( QContextMenuEvent* event )
 		QAction* tetgenExportAction (nullptr);
 		if (mesh_dim==3)
 		{
-			surfaceMeshAction = menu.addAction("Extract surface");
+			surfaceMeshAction = menu.addAction("Extract surface...");
 			tetgenExportAction = menu.addAction("Export to TetGen...");
 		}
 		QAction* mesh2geoAction (nullptr);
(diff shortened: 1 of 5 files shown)


bilke:
:+1:

Applications/DataExplorer/DataView/MshView.cpp
```diff
@@ -187,8 +188,17 @@ void MshView::extractSurfaceMesh()
...
 
 	const MeshLib::Mesh* mesh = static_cast<MshModel*>(this->model())->getMesh(index);
-	const MathLib::Vector3 dir(0, 0, -1);
-	static_cast<MshModel*>(this->model())->addMesh( MeshLib::MeshSurfaceExtraction::getMeshSurface(*mesh, dir, 89) );
+	SurfaceExtractionDialog dlg;
+	if (dlg.exec() != QDialog::Accepted)
+		return;
+	
+	MathLib::Vector3 const& dir (dlg.getNormal());
+	int const tolerance (dlg.getTolerance());
```
norihiro-w: is integer correct data type for tolerance?

norihiro-w: sorry it seems this is your intention

Applications/DataExplorer/DataView/SurfaceExtractionDialog.cpp
```diff
@@ -0,0 +1,45 @@
...
+#include "SurfaceExtractionDialog.h"
+
+#include "MeshLib/MeshSurfaceExtraction.h"
+
+#include <QFileDialog>
+#include <QSettings>
+#include <QDoubleValidator>
+
+// ThirdParty/logog
+#include "logog/include/logog.hpp"
```
norihiro-w: logog not used

norihiro-w: it looks most headers except for QDoubleValidator.h are actually not used

Applications/DataExplorer/DataView/MshView.cpp
rinkk: The QSpinBox providing this value is configured to only return values in [0,90] and the extraction method want a double for calculation, so the int-value should be fine.

Applications/DataExplorer/DataView/SurfaceExtractionDialog.cpp
rinkk: you are right ... I removed all unnecesary includes.

norihiro-w:
+1

PR #576, opened 2015-01-30
Some travis script improvements

bilke:
- apt-get install outputs errors only
- install CMake 3.1.1 system-wide
- install VTK system-wide


TomFischer:
:+1:

PR #577, opened 2015-02-02
[TR] Only run MPI related tests under condition PETSC.

wenqing:
For travis: if PETSc is used, only MPI related unit tests are performed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,7 +55,8 @@ script:
   - cd build
   - cmake $CMAKE_ARGS ..
   - make
-  - ./bin/testrunner --gtest_filter=-MPITest* --gtest_shuffle --gtest_repeat=3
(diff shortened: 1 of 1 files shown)


norihiro-w:
+1

endJunction:
From discussion with Wenqing: There are some unit tests in the AssemblerLib not suitable for testing with PETSC, but other tests should run anyway. Wenqing shall exclude the AssemblerLib and some other tests here.
The basic tests (MathLib, BaseLib, etc.) must run independent of the current configuration.

norihiro-w:
@endJunction is it worth to let travis spend more time to check non-MPI test cases again? as we use OpenMPI based on gcc, the result should be same as gcc. Detailed check can be done with Jenkins.

endJunction:
This is good point, Nori. I didn't think about this.

PR #578, opened 2015-02-02
Added reading partitioned mesh to readMeshFromFile

wenqing:
As the title.

The first commit is PR #577.

### DIFF
--- a/FileIO/readMeshFromFile.cpp
+++ b/FileIO/readMeshFromFile.cpp
@@ -35,7 +35,7 @@
 #include "readMeshFromFile.h"
 // FileIO : for reading partitioned mesh.
 #ifdef USE_PETSC
-#include "MPI_IO/NodePartitionedMeshReader.h"
+#include "FileIO/MPI_IO/NodePartitionedMeshReader.h"
 #include "MeshLib/NodePartitionedMesh.h"
 #endif
 
@@ -55,9 +55,9 @@ MeshLib::Mesh* readMeshFromFile(const std::string &file_name)
 
 	if (BaseLib::hasFileExtension("vtu", file_name))
 		return VtuInterface::readVTUFile(file_name);
-#endif
 
 	ERR("readMeshFromFile(): Unknown mesh file format in file %s.", file_name.c_str());
 	return nullptr;
+#endif
 }
 } // end namespace FileIO
--- a/FileIO/readMeshFromFile.cpp
+++ b/FileIO/readMeshFromFile.cpp
@@ -15,6 +15,10 @@
  * @author Karsten Rink
  */
 
+#ifdef USE_PETSC
+#include <petscksp.h>
+#endif
+
 // ThirdParty/logog
 #include "logog/include/logog.hpp"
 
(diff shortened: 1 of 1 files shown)


MeshLib/Mesh.cpp
```diff
@@ -269,5 +269,15 @@ void Mesh::setNodesConnectedByElements()
...
+std::size_t Mesh:: getMaximumNConnectedNodesToNode() const
+{
+	std::size_t max_ncn = 0;
+	for(auto node : _nodes)
+	{
+		if(node->getConnectedNodes().size() > max_ncn)
+			max_ncn = node->getConnectedNodes().size();
+	}
+	return max_ncn + 1;
+}
```
rinkk: do you actually *need* this method here? I cannot find a call to this function anywhere in the code. but if you are going to need it only once or twice it makes more sense to implement it where you need it instead of enlarging the general mesh-interface. ( also, I think you could implement this more efficiently using ```std::max_element()``` )

MeshLib/MeshSubset.h
```diff
@@ -91,6 +91,11 @@ class MeshSubset
         return _msh.getElements().cend();
     }
 
+    const Mesh &getMesh() const
+    {
+        return _msh;
+    }
```
rinkk: is this method necessary? from what i gather you could access the mesh from the general mesh vector using the mesh id provided in this class? i am not sure, though.

MeshLib/Mesh.cpp
wenqing: Yes,  I need it for the pre-memory allocation for PETSc matrix. Its usage will be shown in  one coming PR.

MeshLib/MeshSubset.h
wenqing: Yes, its usage will be shown in another PR for parallel FEM with PETSc.

MeshLib/Mesh.cpp
wenqing: Removed and will be  added in another PR.

wenqing: And it will be moved to the derived class.

FileIO/readMeshFromFile.cpp
```diff
@@ -42,6 +55,7 @@ MeshLib::Mesh* readMeshFromFile(const std::string &file_name)
 
 	if (BaseLib::hasFileExtension("vtu", file_name))
 		return VtuInterface::readVTUFile(file_name);
+#endif
```
wenqing: Moved.

FileIO/readMeshFromFile.cpp
```diff
@@ -29,11 +33,20 @@
 #include "Legacy/MeshIO.h"
 #include "FileIO/VtkIO/VtuInterface.h"
 #include "readMeshFromFile.h"
+// FileIO : for reading partitioned mesh.
+#ifdef USE_PETSC
+#include "MPI_IO/NodePartitionedMeshReader.h"
```
wenqing: Added.

wenqing:
@endJunction  Removed one commit.

norihiro-w:
:+1:

PR #579, opened 2015-02-03
Cmdline tool for triangulating polylines

rinkk:
this is a request by @waltherm
a command line tool for specifying a polyline that should be triangulated into a surface.
tests if line is closed and coplaner, if not closed user can specify if line should be closed prior to triangulating.

### DIFF
--- a/GeoLib/Polyline.cpp
+++ b/GeoLib/Polyline.cpp
@@ -193,6 +193,26 @@ bool Polyline::isClosed() const
 		return false;
 }
 
+bool Polyline::isCoplanar() const
+{
+	std::size_t const n_points (_ply_pnt_ids.size());
+	if (n_points > 3)
+	{
+		GeoLib::Point const& p0 (*this->getPoint(0));
+		GeoLib::Point const& p1 (*this->getPoint(1));
+		GeoLib::Point const& p2 (*this->getPoint(2));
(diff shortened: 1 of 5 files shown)


GeoLib/Polyline.cpp
```diff
@@ -193,6 +193,26 @@ bool Polyline::isClosed() const
...
+		return true;
+
+	GeoLib::Point const& p0 (*this->getPoint(0));
+	GeoLib::Point const& p1 (*this->getPoint(1));
+	GeoLib::Point const& p2 (*this->getPoint(2));
+	for (std::size_t i=3; i<n_points; ++i)
+	{
+		if (!GeoLib::isCoplanar(p0, p1, p2, *this->getPoint(i)))
+		{
+			ERR ("Point %d is not coplanar to the first three points of the line.", i);
```
TomFischer: @norihiro-w Is @rinkk's  argument valid for you or do you still prefer the change?

norihiro-w: i prefer debug level here and error level at somewhere calling this function which clearly gives a context that copolanar is bad.

norihiro-w: anway this part is tiny and i'll not insisit my preferance if you want to keep it as it is.

PR #580, opened 2015-02-03
Parallelization of the FEM by using PETSc

wenqing:
As the title, this PR presents an implementation of parallelization of assembly and linear solver for the existing process modeler: groundwater flow.  There are changes mainly in ``AssemblerLib``, and ``ProcessLib``. This PR is categorized as follows:

1.  Changes in ``AssemblerLib``
  * [ASLib] Modified mesh component related classes for DDC. 	5745e8c
  * [ASLib] Added two new files. 	                                                f6b1a77
  * [ASLib] Moved two local variables to the place close to where they ar… … 9ae7b42
  * [ASLib] Updated some comments. 	26e0665
  * [MathLib] Replaced one indentation of 4 white spaces with a tab. 	1509518

2. Changes in ``ProcessLib``
  * [PCS] Parallelized the process calculation. e5864f6
  * [Mesh] Added a member to NodePartitionedMesh 3173e9d

3. Changes in main function.
  * [CLI] Added PETSc and MPI stuffs in main function of CLI. e136d4e
  * [CLI] Removed one blank line. bc042e0
  * [CLI] Added PETSc vector, matrix and solver specifications in Numeric… … 	0067366

4. Changes in test.
  * [CM] Disabled compiling SteadyDiffusion2DExample1 under USE_PETSC. 	38fb4db
  * [CM] Excluded non-MPI tests from the test list under MPI. 	4bd70c8
  * [TEST] Added a test for the parallel FEM and also its input data. 	abfee39


AssemblerLib/MeshComponentMap.h
```diff
@@ -107,8 +107,29 @@ class MeshComponentMap
...
+
+    /// Get the number of global unknowns.
+    std::size_t getNGlobalUnknowns() const
+    {
+        return _num_global_dof;
+    }
+
+    /// Get the ghost flags of the nodes at location \c l.
     template <ComponentOrder ORDER>
-    std::vector<std::size_t> getGlobalIndices(const std::vector<Location> &ls) const;
```
wenqing: @norihiro-w  This function is split into two functions:
```
 template <typename T_INT_TYPE>
 std::vector<T_INT_TYPE> getGlobalIndicesByLocation(const std::vector<Location> &ls) const;

/// The same function as getGlobalIndicesByLocation but getting indices by component
template <typename T_INT_TYPE>
std::vector<T_INT_TYPE> getGlobalIndicesByComponent(const std::vector<Location> &ls) const;
```
The definitions of these two functions are given in a new added file:
``` 
AssemblerLib/MeshComponentMap_imp.h.
```

Applications/CLI/NumericsConfig.h
```diff
@@ -37,6 +37,15 @@ namespace detail
...
+#elif defined(USE_PTESC)
+namespace detail
+{
+    using GlobalVectorType = MathLib::PETScVector;
+    using GlobalMatrixType = MathLib::PETScMatrix;
+
+    using LinearSolverType = MathLib::PETScLinearSolver;
+}
+
 #else    // USE_LIS
```
norihiro-w: comment wrong

Applications/CLI/NumericsConfig.h
```diff
@@ -52,13 +61,37 @@ namespace detail
...
+
+using GlobalVectorMatrixBuilderType =
+        AssemblerLib::PETScVectorMatrixBuilder<
+            MathLib::PETScMatrix, MathLib::PETScVector>;
+}
 
+///
+/// Global setup collects the previous configuration in single place.
+///
+using GlobalSetupType =
```
norihiro-w: GlobalSetupType can be shared with non-PETSc case, because  detail::LinearSolverType points to PETSc solver

Applications/CLI/ogs.cpp
```diff
@@ -58,6 +77,11 @@ int main(int argc, char *argv[])
 		"",
 		"PROJECT FILE");
 
+	BaseLib::RunTime run_timer;
+	run_timer.start();
+	BaseLib::CPUTime CPU_timer;
```
norihiro-w: is it really necessary to record CPU time?

Applications/CLI/ogs.cpp
```diff
@@ -96,6 +119,22 @@ int main(int argc, char *argv[])
 
 	output_file.close();
 
+	INFO("Elapsed running time in total computation %g s", run_timer.elapsed());
```
norihiro-w: suggest making it short:
"Elapsed running time" -> "Elapsed time"
"Elapsed CPU time" -> "CPU time"

Applications/CLI/ogs.cpp
```diff
@@ -96,6 +119,22 @@ int main(int argc, char *argv[])
...
+	INFO("Elapsed CPU time in total computation %g s", CPU_timer.elapsed());
+
+#ifdef USE_PETSC
+	INFO("Released the memory occupied by MPI based vector, matrix and linear solver.\n");
```
norihiro-w: why is this information relevant to users?

Applications/CLI/NumericsConfig.h
```diff
@@ -52,13 +61,37 @@ namespace detail
...
-
 //
 // Global executor and vector/matrix builder.
 //
 #include "AssemblerLib/SerialExecutor.h"
-#include "AssemblerLib/SerialVectorMatrixBuilder.h"
+#include "AssemblerLib/GlobalSetup.h"
+
+#ifdef USE_PETSC
+    #include "AssemblerLib/PETScVectorMatrixBuilder.h"
```
wenqing: Moved.

Applications/CLI/NumericsConfig.h
wenqing: Modified.

Applications/CLI/ogs.cpp
wenqing: May it be used together with run-time to estimate I/O time.

Applications/CLI/ogs.cpp
wenqing: Changed.

Applications/CLI/ogs.cpp
wenqing: Removed.

Applications/CLI/NumericsConfig.h
```diff
@@ -52,26 +64,35 @@ namespace detail
...
             GlobalVectorType>;
 }
+#endif
 
-
-#include "AssemblerLib/GlobalSetup.h"
-
+namespace detail
+{
+using GlobalExecutorType = AssemblerLib::SerialExecutor;
```
wenqing: @norihiro-w  How about change ``SerialExecutor`` to ``Executor`` in another PR?

MeshLib/NodePartitionedMesh.h
```diff
@@ -114,6 +118,20 @@ class NodePartitionedMesh : public Mesh
             return _n_nghost_elem;
         }
 
+        /// Get the maximum number of connected nodes to node.
+        std::size_t getMaximumNConnectedNodesToNode() const
+        {
+            std::vector<Node*>::const_iterator it_max_ncn
+                 = std::max_element(this->_nodes.cbegin(), this->_nodes.cend(),
```
wenqing: @rinkk  Thanks. Changed to const types  and it works fine.

wenqing:
To @norihiro-w 
The last two commits 5d83dcd and 1a95df6 optimized GroundwaterFlow::addToGlobal for USE_PETSC by
* replaced ``_element_ghost_node_flags`` with ``_element_non_ghost_local_ids`` in class ``LocalToGlobalIndexMap``, which avoids if-condition in a loop in local assembly.
* Added a overloaded general function ``add`` in ``PETScMatrix``, which makes the changes in  ``addToGlobal`` concise.

AssemblerLib/LocalToGlobalIndexMap.h
```diff
@@ -31,32 +35,31 @@ namespace AssemblerLib
 class LocalToGlobalIndexMap
 {
 public:
+#ifdef USE_PETSC
+    typedef MathLib::RowColumnIndices<PetscInt> RowColumnIndices;
```
endJunction: @wenqing How is the PetscInt different from maximum unsigned integer?
Actually this code should not be dependent on PETSc but on MeshLib::NodePartitionMesh or (base) MeshLib::Mesh.
This part handles indices of some meshes and parts of them and is not related to PETSc or any other type of parallelization.

wenqing: ``MeshLib::NodePartitionMesh`` is independent of PETSc, and it could be used to other node partitioned approach. However, members of RowColumnIndices are passed to PETSc matrix and vector, where the type of PetscInt (defined as int in current version) is needed. 

Of course, we can have a conversion safe assertion for unsigned long to int in the constructor of MeshComponentMap. But I think it is unnecessary, because in seldom case, the number of nodes can reach  the number of 2^32-1. Besides, we can use option  --with-64-bit-indices.

wenqing: I see you mean size_t. I can try a test to pass size_t type integer to PetscInt to see whether there is any problem.

wenqing: Tested. No complain from compiler. Change it soon.

wenqing: Tried. If this ``MathLib::RowColumnIndices<PetscInt> RowColumnIndices`` is moved, I have to replace PetscInt with std::size_t in the related arguments of all member functions of PETScMatrix and PETScVector. I would like to keep this for flexibility.  If it is definitely needed to remove this, I will do it in next PR, where the changes are only to remove ifdef and modify  PETScMatrix and PETScVector.

endJunction:
@wenqing Currently the ctest is failing:
```
Test project /home/naumov/w/ogs/buildP
    Start 1: NodePartitionedMeshTestASCII
1/3 Test #1: NodePartitionedMeshTestASCII .....   Passed    0.43 sec
    Start 2: NodePartitionedMeshTestBinary
2/3 Test #2: NodePartitionedMeshTestBinary ....   Passed    0.42 sec
    Start 3: ParalllelFEM_GroundWaterFlow
3/3 Test #3: ParalllelFEM_GroundWaterFlow .....***Failed    0.07 sec
```

Running manually
```bash
make && make data && bin/ogs Tests/Data/EllipticPETSc/quad_20x10_GroundWaterFlow.prj
```
yields
```
error: readMeshFromFile(): Unknown mesh file format in file Tests/Data/EllipticPETSc/quad_20x10_GroundWaterFlow.
error: Could not read mesh from 'Tests/Data/EllipticPETSc/quad_20x10_GroundWaterFlow' file. No mesh added.
error: A mesh is required to define process variables.
error: Expected process variable 'pressure' not found in provided variables list.
info: Initialize processes.
zsh: segmentation fault (core dumped)  bin/ogs Tests/Data/EllipticPETSc/quad_20x10_GroundWaterFlow.prj
```
Any ideas?

AssemblerLib/PETScVectorMatrixBuilder.h
```diff
@@ -0,0 +1,62 @@
...
+    typedef MatrixType_ MatrixType;
+
+public:
+    /*!
+        \param vec_size       The size of the vector, either global or local
+        \param is_global_size The flag of the type of vec_size, i.e. whether it is a global size
+                              or local size. The default is true.
+    */
+    static
+    VectorType* createVector(const PetscInt vec_size, const bool is_global_size = true)
```
endJunction: @wenqing I could not find a place where `is_global_size` is used. Can you tell please under which conditions one would pass `false` value here?

wenqing:
@endJunction [About the error in your test]. Just tested on another machine, it works except one error in ``make data``. The error you got looks like that the changes in the merged PR. #578 is missing, and the serial mesh reader is called. If you run the test manually, use command
```
make && make data && mpirun -np 3 bin/ogs Tests/Data/EllipticPETSc/quad_20x10_GroundWaterFlow.prj
```
or add more option at the end of the command as 
```
-- -gw_ksp_type bcgs -gw_pc_type bjacobi -gw_ksp_atol 1.e-10 
```
for comparison with its reference.

As mentioned at the beginning , there is an error in ``make data``. When I checkout the source code to my laptop and ran ``make data``, I got an error
```
 Object MD5=5763a0cca7adeb0fda1bbad6dc117c70 not found at:
```
I have checked that ``5763a0cca7ade...`` has already copied to /data/ogs/ogs6-data/MD5. Do you have any idea about this error?

AssemblerLib/PETScVectorMatrixBuilder.h
wenqing: In case the vector size is given locally for each sub-domain, then the global size is calculated in size the constructor . It is used when one want to partition vector manually, and it will be used late on. There are also tests about it in TestGlobalVectorInterface.cpp

endJunction:
@wenqing I synced the external data just now. (The sync target is build only every merge or so.)
The other thing was just forgotten `mpirun -np 3`. It's working. I would like to add some sanity checks for the wrong input...

wenqing:
@endJunction  make data works.

wenqing:
@endJunction  Just changed quad_20x10_GroundWaterFlow.prj by removing path tmp in the output file name. Please synchronize the external data.

wenqing:
I changed the label to ``please review``, and I am not sure a message of the change being sent out. Anyway, please review this rebased PR.

ProcessLib/GroundwaterFlowProcess.h
```diff
@@ -260,16 +306,33 @@ class GroundwaterFlowProcess : public Process
...
+#endif
         return true;
     }
 
     void post(std::string const& file_name)
     {
         DBUG("Postprocessing GroundwaterFlowProcess.");
+
+#ifdef USE_PETSC // Note: this is only a test
+        int rank;
```
wenqing: As mentioned before, this output is only for test purpose, and there will be another PR for a parallel VTK output.

wenqing:
@endJunction Removed the timer stuff and authors.

wenqing:
@endJunction  Rebased.

FileIO/MPI_IO/NodePartitionedMeshReader.cpp
```diff
@@ -71,21 +72,24 @@ MeshLib::NodePartitionedMesh* NodePartitionedMeshReader::read(
 
     MeshLib::NodePartitionedMesh* mesh = nullptr;
 
+    std::string confirmed_file_name_base = file_name_base;
+    boost::erase_last(confirmed_file_name_base, ".vtk");
```
endJunction: Use `boost::erase_last_copy()` here for simplicity.
Better to name the result as `filename_prefix` or something, because nothing is confirmed here.

MathLib/LinAlg/RowColumnIndices.h
```diff
@@ -15,6 +15,21 @@
 namespace MathLib
 {
 
+#ifdef USE_PETSC
```
endJunction: The `#ifdef` is easily avoidable when using `std::vector<IDX_TYPE>*` with default `nullptr` value.

MeshLib/NodePartitionedMesh.h
```diff
@@ -96,11 +101,11 @@ class NodePartitionedMesh : public Mesh
         bool isGhostNode(const std::size_t node_id) const
         {
             if(node_id < _n_active_base_nodes)
-                return true;
+                return false;
```
endJunction: This line and the following if/else's do change the logic completly. Is it a bug-fix or something?

ProcessLib/GroundwaterFlowProcess.h
```diff
@@ -329,6 +400,10 @@ class GroundwaterFlowProcess : public Process
     std::unique_ptr<typename GlobalSetup::VectorType> _rhs;
     std::unique_ptr<typename GlobalSetup::VectorType> _x;
 
+#ifdef USE_PETSC
+    std::unique_ptr<typename GlobalSetup::LinearSolver> _linearSolver;
```
endJunction: Why is it not possible to use the linar solver present in the `solve()` function above?

AssemblerLib/MeshComponentMap-impl.h
```diff
@@ -0,0 +1,85 @@
...
+#ifdef USE_PETSC
+#include "MeshLib/NodePartitionedMesh.h"
+#endif
+
+#include "MeshComponentMap.h"
+
+namespace AssemblerLib
+{
+
+using namespace detail;
```
endJunction: Including namespace in a header file is a bad idea.

MeshLib/NodePartitionedMesh.h
wenqing: It's a bug fixing.

ProcessLib/GroundwaterFlowProcess.h
wenqing: Yes, it is possible by adding a string type of argument to  ``GlobalSetup::LinearSolver(...) ``. However, in such way, PETSc ksp solver  has to be created  and deleted for every time when it is called, which, you know, would involves a lot of memory allocation/release operations for the computation of transient problems.

AssemblerLib/MeshComponentMap-impl.h
wenqing: This file was removed.

MathLib/LinAlg/RowColumnIndices.h
wenqing: Changed with a default value to the third argument of the constructor.

FileIO/MPI_IO/NodePartitionedMeshReader.cpp
wenqing: The name was changed. ``boost::erase_last_copy()`` needs an additional string variable as output. Since  boost::erase_last is only called once here, it is kept.

wenqing:
@endJunction std::size_t type variable cannot be passed to PETSc type argument directly. A new file ogsInt.h about a type declaration is created in BaseLib for a universal integer type to avoid some template definition.

TomFischer:
The style guide states:
> Type names start with a capital letter and have a capital letter for each new word, with no underscores

In the consequence `OgsInt` would be a style guide compliant name.

wenqing:
@TomFischer  The type name has been changed. Before this, I just thought that the term of ogs is special.

wenqing:
Rebased.

ProcessLib/GroundwaterFlowFEM.h
```diff
@@ -109,8 +114,17 @@ class LocalAssemblerData : public LocalAssemblerDataInterface<GlobalMatrix, Glob
     void addToGlobal(GlobalMatrix& A, GlobalVector& rhs,
             AssemblerLib::LocalToGlobalIndexMap::RowColumnIndices const& indices) const
     {
-        A.add(indices, *_localA);
-        rhs.add(indices.rows, *_localRhs);
+#ifdef USE_PETSC
```
endJunction: This part must go into the `addToGlobal` function. It will not be possible to maintain the PETSc support for the future process implementations.

MathLib/LinAlg/RowColumnIndices.h
```diff
@@ -14,19 +14,25 @@
...
+	                 const std::vector<std::size_t>& non_ghost_local_ids_
+	                 = dummy_non_ghost_local_ids)
+		: rows(rows_), columns(columns_), non_ghost_local_ids(non_ghost_local_ids_)
 	{ }
 
 	LineIndex const& rows;
 	LineIndex const& columns;
-};
 
+	std::vector<std::size_t> const& non_ghost_local_ids;
```
endJunction: Is it possible to avoid the non_ghost_local_ids here and move it entirely to AssemblerLib?
I'm asking, because this member is used outside of AssemblerLib only once in the GroundwaterFlowFEM, which should become independent of any mpi related code.

MathLib/LinAlg/RowColumnIndices.h
```diff
@@ -14,19 +14,25 @@
 
 namespace MathLib
 {
+// A dummy vector as the default value of the third argument of
+// the constructor.
+static std::vector<std::size_t> dummy_non_ghost_local_ids;
```
endJunction: This (global) variable is not necessary, as far I can understand it. Why not use default initialization of a member variable in the RowColumnIndices ctor?

wenqing: All reference member can only be initialized in the initialization list. Therefore, overload ctor with different argument is also not possible. Alternative, an ifdef as the previous commit or some other solution that I am thinking about.

MathLib/LinAlg/RowColumnIndices.h
wenqing: Thinking to have a new class to handle these DDC stuff in AssemblerLib

MathLib/LinAlg/RowColumnIndices.h
endJunction: Ahh, I see. Thanks for explanation.

ProcessLib/GroundwaterFlowProcess.h
```diff
@@ -208,6 +216,18 @@ class GroundwaterFlowProcess : public Process
...
+        DBUG("Allocate global matrix, vectors, and linear solver.");
+        MathLib::PETScMatrixOption mat_opt;
+        const MeshLib::NodePartitionedMesh &pmesh
+                    = static_cast<const MeshLib::NodePartitionedMesh&>(_mesh);
+        mat_opt.d_nz = pmesh.getMaximumNConnectedNodesToNode();
+        mat_opt.o_nz = mat_opt.d_nz;
+        _A.reset(_global_setup.createMatrix(_local_to_global_index_map->dofSizeGlobal(), mat_opt) );
+        _x.reset(_global_setup.createVector(_local_to_global_index_map->dofSizeGlobal()));
+        _rhs.reset(_global_setup.createVector(_local_to_global_index_map->dofSizeGlobal()));
+        _linearSolver.reset(new typename GlobalSetup::LinearSolver(*_A, "gw_"));
```
endJunction: Is the second argument to the LinearSolver ctor necessary?
If yes, please prepare a PR unifying existing LinearSolver ctor concepts.
Otherwise I strongly recommend to stick to the existing concept of LinearSolvers.

endJunction:
I have pushed several changes to [endJ/pFEM4 branch](https://github.com/endJunction/ogs/tree/pFEM4) including a removal of the proposed OgsInt type with changes to petsc vector/matrix classes. The included tests all are running just fine. The conversion is local to petsc classes.

Therefore I strongly recommend not to change the index types in the mesh component map and related classes.

Please let me know if the changes in the  [endJ/pFEM4 branch](https://github.com/endJunction/ogs/tree/pFEM4) do break any of other concepts/codes.

ProcessLib/GroundwaterFlowProcess.h
wenqing: It is necessary only for PETSc to allow commend line PETSc options, which enables more flexible settings for PETSc matrix and ksp solver, PCs, and external solvers.

wenqing:
@endJunction  I saw you removed OgsInt type by a type conversion inside ``add`` function. As @chleh  proposed a better idea to deal with ghost rows of matrix by using the negative row index, I would like to keep OgsInt. I am going to try the way of negative index now.

endJunction:
Handling the ghost nodes and the type of dof table's global index are two different issues.

wenqing:
@endJunction @chleh Recent changes are about allowing negative row/col indecies. With the changes, several ifdefs are gone and additional memory for ghost nodes are removed too.
PS: The commits include cherry-picked commits from PR #844  and PR #843.

MathLib/LinAlg/PETSc/PETScMatrix.h
```diff
@@ -159,6 +159,19 @@ class PETScMatrix
             MatSetValue(_A, i, j, value, ADD_VALUES);
         }
 
+        /// Add sub-matrix at positions given by \c indices.
```
wenqing: To be rebased from #844 .

MathLib/LinAlg/PETSc/PETScVector.cpp
```diff
@@ -42,6 +42,8 @@ PETScVector::PETScVector(const PetscInt vec_size, const bool is_global_size)
 
     VecGetLocalSize(_v, &_size_loc);
     VecGetSize(_v, &_size);
+
+    VecSetOption(_v, VEC_IGNORE_NEGATIVE_INDICES,PETSC_TRUE);
```
wenqing: to be rebased from #844 .

MeshLib/NodePartitionedMesh.h
```diff
@@ -97,15 +97,19 @@ class NodePartitionedMesh : public Mesh
             return _n_active_nodes;
         }
 
-        /// Check whether a node with ID of node_id is a ghost node
-        bool isGhostNode(const std::size_t node_id) const
+        /*! Check whether a node with ID of node_id is a ghost node.
```
wenqing: to be rebased from #843 .

ProcessLib/NumericsConfig.h
```diff
@@ -65,12 +77,20 @@ namespace detail
 #endif    // USE_LIS
 #endif    // OGS_USE_EIGEN
 
-
 //
 // Global vector/matrix builder.
 //
 
-#include "AssemblerLib/SerialVectorMatrixBuilder.h"
+#ifdef USE_PETSC
```
endJunction: The ifdef and the PETScVectorMatrixBuilder are not necessary. The SerialVectorMatrixBuilder will forward all arguments to the constructors of matrix and vector.
Do I overlook something obvious?

wenqing: createVector and createMatrix in PETScVectorMatrixBuilder have different argument compared to that of the serial version.

wenqing: Removed  PETScVectorMatrixBuilder.h

AssemblerLib/MeshComponentMap.cpp
```diff
@@ -20,9 +25,76 @@ namespace AssemblerLib
...
+                   = static_cast<const MeshLib::NodePartitionedMesh&>(mesh_subset.getMesh());
+
+            if (order == ComponentOrder::BY_LOCATION)
+            {
+                // mesh items are ordered first by node, cell, ....
+                for (std::size_t j=0; j<mesh_subset.getNNodes(); j++)
+                {
+                    const OgsInt global_id = static_cast<OgsInt>(components.size()
+                                               * mesh.getGlobalNodeID(j) + comp_id);
+                    const OgsInt signed_global_id = mesh.isGhostNode( mesh.getNode(j)->getID() )
```
wenqing: @endJunction Amended commit (04d8fe89) with your suggestion.

MathLib/LinAlg/Eigen/EigenLinearSolver.h
```diff
@@ -25,7 +25,17 @@ class EigenMatrix;
 class EigenLinearSolver final
 {
 public:
-    EigenLinearSolver(EigenMatrix &A, boost::property_tree::ptree const*const option = nullptr);
+    /**
```
wenqing: To be rebased from #847 .

wenqing:
rebased.

bilke:
Some comment lines are too long but beside that :+1:

wenqing:
@bilke long comment lines have been split.

endJunction:
Thanks everybody for your patience!

We have four open topics to resolve until the next version, two touching the parallelized part only:
 - Write tests, same as the currently existing square and cube examples.
 - Implement parallel (preferably pvtu) output.
And two where also the serial part is involved:
 - Generalize dirichlet boundary conditions
 - Generalize current sparsity pattern, s.t. it can be used with petsc too.

PR #584, opened 2015-02-04
[bugfix] use include_directories for a list of dir

norihiro-w:
CMake on my PC complained INCLUDE can take only one parameter.

### DIFF
--- a/Applications/Utils/OGSFileConverter/CMakeLists.txt
+++ b/Applications/Utils/OGSFileConverter/CMakeLists.txt
@@ -79,7 +79,7 @@ ADD_CATALYST_DEPENDENCY(OGSFileConverter)
 # Adds useful macros and variables
 # this is needed to correctly link the qt and vtk libraries through target_link_libraries
 INCLUDE( ${QT_USE_FILE} )
-INCLUDE( ${VTK_USE_FILE} )
+include_directories( ${VTK_USE_FILE} )
 
 # Set build configuration types
 # "RelWithDebInfo" and "MinSizeRelease" can be added here


wenqing:
:+1:

rinkk:
seems to work fine on windows.

bilke:
`${VTK_USE_FILE}` references a CMake-file usually found in your VTK installation and should therefore be included with `INCLUDE()`. It contains some `include_directories()`-calls but also other stuff. Maybe there is something wrong with your VTK installation?

endJunction:
I have just tried to reproduce this on envinf1, with the vtk/6.1.0-2_gcc_4.8.1_CentOS6, but there is no message from cmake.

norihiro-w:
thanks all for testing. i haven't used any preinstalled VTK and let CMake to download and compile it.

bilke:
Nori, can you give me more details on your setup (OS, compiler).

norihiro-w:
The CMake error 
```
CMake Error at Applications/Utils/OGSFileConverter/CMakeLists.txt:83 (INCLUDE):
  include called with wrong number of arguments.  include() only takes one
  file.
```
is reproducable on my environment: 
- Ubuntu 14.04
- CMake 3.0.2, `-DOGS_BUILD_CLI=ON -DOGS_BUILD_GUI=ON -DOGS_BUILD_UTILS=ON`
- No system VTK lib

I debugged CMake and found VTK_USE_FILE is actualyl empty, i.e. VTK not found because it's before compiling VTK. Should I add `if (VTK_FOUND)` there?

PR #585, opened 2015-02-05
Small util to reset properties in polygonal region.

TomFischer:
In this PR a small util is implemented that sets the material property id of an mesh element to a given new id iff all nodes of the element are within a polygonal region given by a polygon read from a gml file.


Applications/Utils/MeshEdit/ResetPropertiesInPolygonalRegion.cpp
```diff
@@ -0,0 +1,216 @@
...
+#include "MeshLib/Node.h"
+#include "MeshLib/Elements/Element.h"
+#include "MeshLib/MeshEditing/ElementValueModification.h"
+
+GeoLib::Polygon rotatePolygonToXY(GeoLib::Polygon const& polygon_in,
+	MathLib::Vector3 & plane_normal)
+{
+	// 1 copy all points
+	std::vector<GeoLib::Point*> *polygon_pnts(new std::vector<GeoLib::Point*>);
+	for (size_t k(0); k < polygon_in.getNumberOfPoints(); k++)
```
wenqing: Dima cares about using ``size_t`` with its namespace ``std``. Do we have a rule for this? I think if it is used in cpp file and the file does not include any C header, using size_t directly is OK and less verbose.

Applications/Utils/MeshEdit/ResetPropertiesInPolygonalRegion.cpp
```diff
@@ -0,0 +1,216 @@
...
+	{
+		FileIO::BoostXmlGmlInterface xml_io(geometries);
+		if (xml_io.readFile(geometry_fname.getValue())) {
+			INFO("Read geometry from file \"%s\".",
+				geometry_fname.getValue().c_str());
+		} else {
+			ERR("Problems to read geometry from file \"%s\".",
+				geometry_fname.getValue().c_str());
+			delete mesh;
+			return -1;
```
wenqing: Why return -1? 1 or EXIT_FAILURE?

Applications/Utils/MeshEdit/ResetPropertiesInPolygonalRegion.cpp
TomFischer: Changed ```size_t``` to ```std::size_t```.

Applications/Utils/MeshEdit/ResetPropertiesInPolygonalRegion.cpp
TomFischer: Returning more readable EXIT_FAILURE.

Applications/Utils/MeshEdit/ResetPropertiesInPolygonalRegion.cpp
```diff
@@ -0,0 +1,216 @@
...
+	std::size_t new_property(new_material_id_arg.getValue());
+
+	GeoLib::Polygon polygon(*((*plys)[polygon_id]));
+
+	resetProperty(*mesh, polygon, new_property);
+
+	FileIO::VtuInterface mesh_io(mesh);
+	mesh_io.writeToFile (mesh_out.getValue());
+
+	return 0;
```
wenqing: Then here EXIT_SUCCESS

TomFischer: You are right. Will change this.

Applications/Utils/MeshEdit/ResetPropertiesInPolygonalRegion.cpp
```diff
@@ -0,0 +1,216 @@
...
+		return EXIT_FAILURE;
+	}
+
+	std::size_t new_property(new_material_id_arg.getValue());
+
+	GeoLib::Polygon polygon(*((*plys)[polygon_id]));
+
+	resetProperty(*mesh, polygon, new_property);
+
+	FileIO::VtuInterface mesh_io(mesh);
```
norihiro-w: why not use `writeMeshToFile()`?

rinkk:
could you please add a short description in future PRs just so i know what i am looking at?
also, the utility itself does not give a description of what it does, you should add this.

just to make sure i see this correctly: you have a mesh and a polygon and the mesh elements inside the polygon region are assigned a new material id?

Applications/Utils/MeshEdit/ResetPropertiesInPolygonalRegion.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+	// *** read geometry
+	GeoLib::GEOObjects geometries;
+	{
+		FileIO::BoostXmlGmlInterface xml_io(geometries);
+		if (xml_io.readFile(geometry_fname.getValue())) {
+			INFO("Read geometry from file \"%s\".",
+				geometry_fname.getValue().c_str());
+		} else {
+			ERR("Problems to read geometry from file \"%s\".",
```
rinkk: i think this msg is not necessary, the reader will output a message that this is not a valid GML file.

Applications/Utils/MeshEdit/ResetPropertiesInPolygonalRegion.cpp
```diff
@@ -0,0 +1,215 @@
...
+	cmd.add(mesh_in);
+	TCLAP::ValueArg<std::string> mesh_out("o", "mesh-output-file",
+		"the name of the file the mesh will be written to", true,
+		"", "file name");
+	cmd.add(mesh_out);
+	TCLAP::ValueArg<std::string> geometry_fname("g", "geometry",
+		"the name of the file containing the input geometry", true,
+		"", "file name");
+	cmd.add(geometry_fname);
+	TCLAP::ValueArg<unsigned> polygon_id_arg("p", "polygon-id",
```
rinkk: i would not expect an id but the name of the polyline.
the problem with the id is that using the current implementation the id given in the gml file need *not* be the id the polyline ends up with when it is stored in geoobjects.

rinkk:
you rotate the polygon but not the mesh ... i suppose the logic is that this was necessary for the data you wrote this tool for ... but this should definitely be noted somewhere because it is not straightforward ... also, you flatten the polyline to z=0 which is again a special case .... i guess this only works with flat meshes where z=0 for all nodes? please note this somewhere ...

Applications/Utils/MeshEdit/ResetPropertiesInPolygonalRegion.cpp
```diff
@@ -0,0 +1,215 @@
...
+	// 3 set z coord to zero
+	std::for_each(polygon_pnts->begin(), polygon_pnts->end(),
+		[] (GeoLib::Point* p) { (*p)[2] = 0.0; }
+	);
+
+	// 4 create new polygon
+	GeoLib::Polyline rot_polyline(*polygon_pnts);
+	for (std::size_t k(0); k < polygon_in.getNumberOfPoints(); k++)
+		rot_polyline.addPoint (k);
+	rot_polyline.addPoint (0);
```
rinkk: is this correct here? if the polyline is already closed (which is a requirement) then the last point will already be in the list of points?

Applications/Utils/MeshEdit/ResetPropertiesInPolygonalRegion.cpp
```diff
@@ -0,0 +1,215 @@
...
+			rot_polygon.getPointsVec()
+		)
+	);
+	for (std::size_t k(0); k < rot_polygon_pnts.size(); k++)
+		delete rot_polygon_pnts[k];
+
+	return outside;
+}
+
+void resetProperty(MeshLib::Mesh &mesh, GeoLib::Polygon const& polygon,
```
rinkk: you should note somewhere that you only reset elements that are *completely* located within the polygon.

TomFischer:
@rinkk wrote:
> you rotate the polygon but not the mesh ... 

The mesh is rotated in function ```makeNodesOutsideOfPolygon()``` in the same way the polygon is rotated.

> also, you flatten the polyline to z=0 which is again a special case .... i guess this only works with flat meshes where z=0 for all nodes? please note this somewhere ...

The algorithm marks alls element that have at least one node in the cylindrical volume defined by the polygon. It is not restricted to "flat" meshes.

I hope I fixed all your issues.

Applications/Utils/MeshEdit/ResetPropertiesInPolygonalRegion.cpp
```diff
@@ -0,0 +1,217 @@
...
+#include "MeshLib/Mesh.h"
+#include "MeshLib/Node.h"
+#include "MeshLib/Elements/Element.h"
+#include "MeshLib/MeshEditing/ElementValueModification.h"
+
+GeoLib::Polygon rotatePolygonToXY(GeoLib::Polygon const& polygon_in,
+	MathLib::Vector3 & plane_normal)
+{
+	// 1 copy all points
+	std::vector<GeoLib::Point*> *polygon_pnts(new std::vector<GeoLib::Point*>);
```
norihiro-w: include vector

Applications/Utils/MeshEdit/ResetPropertiesInPolygonalRegion.cpp
```diff
@@ -0,0 +1,217 @@
...
+	for (std::size_t k(0); k < polygon_in.getNumberOfPoints(); k++)
+		polygon_pnts->push_back (new GeoLib::Point (*(polygon_in.getPoint(k))));
+
+	// 2 rotate points
+	double d_polygon (0.0);
+	GeoLib::getNewellPlane (*polygon_pnts, plane_normal, d_polygon);
+	MathLib::DenseMatrix<double> rot_mat(3,3);
```
norihiro-w: include DenseMatrix

Applications/Utils/MeshEdit/ResetPropertiesInPolygonalRegion.cpp
```diff
@@ -0,0 +1,217 @@
...
+	GeoLib::computeRotationMatrixToXY(plane_normal, rot_mat);
+	GeoLib::rotatePoints(rot_mat, *polygon_pnts);
+
+	// 3 set z coord to zero
+	std::for_each(polygon_pnts->begin(), polygon_pnts->end(),
```
norihiro-w: include algorithm

Applications/Utils/MeshEdit/ResetPropertiesInPolygonalRegion.cpp
```diff
@@ -0,0 +1,217 @@
...
+	// *** read geometry
+	GeoLib::GEOObjects geometries;
+	{
+		FileIO::BoostXmlGmlInterface xml_io(geometries);
+		if (xml_io.readFile(geometry_fname.getValue())) {
+			INFO("Read geometry from file \"%s\".",
+				geometry_fname.getValue().c_str());
+		} else {
+			delete mesh;
+			return EXIT_FAILURE;
```
norihiro-w: include cstdlib

Applications/Utils/MeshEdit/ResetPropertiesInPolygonalRegion.cpp
```diff
@@ -0,0 +1,217 @@
...
+// MathLib
+#include "MathLib/Vector3.h"
+
+// MeshLib
+#include "MeshLib/Mesh.h"
+#include "MeshLib/Node.h"
+#include "MeshLib/Elements/Element.h"
+#include "MeshLib/MeshEditing/ElementValueModification.h"
+
+GeoLib::Polygon rotatePolygonToXY(GeoLib::Polygon const& polygon_in,
```
norihiro-w: why not move this funciton to GeoLib so that the code can be reused by other people? do you have any concerns?

norihiro-w:
:+1:

Applications/Utils/MeshEdit/CMakeLists.txt
```diff
@@ -112,6 +112,17 @@ target_link_libraries (reviseMesh
 	${CATALYST_LIBRARIES}
 )
 ADD_CATALYST_DEPENDENCY(reviseMesh)
+set_target_properties(reviseMesh
+	PROPERTIES FOLDER Utilities)
 
-set_target_properties(reviseMesh PROPERTIES FOLDER Utilities)
+add_executable( ResetPropertiesInPolygonalRegion
+	ResetPropertiesInPolygonalRegion.cpp )
+target_link_libraries( ResetPropertiesInPolygonalRegion
```
bilke: Recently I removed extraneous whitespace in CMake-files for consistency. Please remove the whitespace before the first parameter.

bilke:
:+1: 
It would be nice to have a small tutorial page on http://docs.opengeosys.org as well!

TomFischer:
> It would be nice to have a small tutorial page on http://docs.opengeosys.org as well!

This site is already under construction :smiley:

PR #588, opened 2015-02-05
[MeL] Simple Element::operator<<() for debugging.

endJunction:


MeshLib/Elements/Element.cpp
```diff
@@ -186,4 +186,20 @@ bool Element::isBoundaryElement() const
         [](MeshLib::Element const*const e){ return e == nullptr; });
 }
 
+#ifndef NDEBUG
+std::ostream& operator<<(std::ostream& os, Element const& e)
+{
+	os << "Element " << e._id << " @ " << &e
```
norihiro-w: suggestion: adding an elemet type, mat ID to output

TomFischer:
:+1: Can be very useful.

rinkk:
can you give an example on how to use this? 
why do i need both the ostream param *and* the ostream return value?
why is the operator defined in ```Element``` class on not on the mesh? and why do i need an element as a param, wouldn't it be applied to ```this```?
(sorry if this seems annoying but somehow i don't get it)

endJunction:
```c++
std::vector<Element*> es;
for (auto e : es) std::cout << *e;
```
The operator<< is a binary operator (usually used in infix notation). To concatenate multiple applications of << the return ostream is required. The order of parameters is also important (we want write `std::cout << e` and not `e.operator<<(std::cout)`; therefore it is implemented as a non-member function.
One could move the definition into the Mesh.cpp, but it makes no sense. Btw  the operator is not defined in the Element class, it is a non-member function.

rinkk:
ah, thanks for the explanation. defining this so you can use the syntax in your example makes a lot of sense.
:+1:

MeshLib/Elements/Element.cpp
endJunction: I'm adding a MeshElemType (line, hex) , not the CellType (line2, hex8,...). The number of nodes is visible anyway...

endJunction: Also, I put the number in the enum, not the string for that type. The latter would require output operator(s) for MeshElemType (and CellType).

PR #592, opened 2015-02-09
Ignore XML comments in input files.

endJunction:
... only those, which are read by the read_xml() function.

Otherwise the comments are stored in extra xmlcomment boost::property_tree, which is not parsed by any of the config readers.
With the new function the xmlcomments are ignored.


norihiro-w:
:+1:

PR #594, opened 2015-02-09
fix CMake error due to empy VTK_USE_FILE

norihiro-w:
This is follow-up  of #584. Sorry I created a branch in the previous PR in the UFZ repo without awareness.


endJunction:
Checked on archlinux without paraview/vtk being installed.
:fast_forward:

bilke:
Sorry Nori, but can you try to simply remove the line `INCLUDE( ${VTK_USE_FILE} )`? I guess this will work because all this stuff should be already handled in `ExternalProjectCatalyst.cmake` and `ADD_CATALYST_DEPENDENCY(OGSFileConverter)`.

norihiro-w:
thanks! it worked

PR #596, opened 2015-02-10
[Fix] Several issues with GMSH geo file export

rinkk:
* moved gmsh-geo-export to GMSHInterface
* fixed issue where geo export would fail if only one geometry was loaded
* output of correct error messages in data explorer

### DIFF
--- a/FileIO/GMSHInterface.cpp
+++ b/FileIO/GMSHInterface.cpp
@@ -74,6 +74,44 @@ GMSHInterface::GMSHInterface(GeoLib::GEOObjects & geo_objs,
 	}
 }
 
+int GMSHInterface::writeGeoFile(GeoLib::GEOObjects &geo_objects, std::string const& file_name)
+{
+	std::vector<std::string> names;
+	geo_objects.getGeometryNames(names);
+
+	if (names.size() == 0)
+	{
+		ERR ("No geometry information available.");
+		return 1;
+	}
+
+	bool const multiple_geometries = (names.size() > 1) ? true : false;
+	std::string merge_name("MergedGeometry");
+	if (multiple_geometries)
+	{
+		if (geo_objects.mergeGeometries (names, merge_name) != 1)
+			return 2;
+		names.clear();
+		names.push_back(merge_name);
+	}
+	else
+		merge_name = names[0];
+
+	double param1(0.5); // mesh density scaling on normal points
+	double param2(0.05); // mesh density scaling on station points
+	size_t param3(2); // points per leaf
+	GMSHInterface gmsh_io(geo_objects, true, FileIO::GMSH::MeshDensityAlgorithm::AdaptiveMeshDensity, param1, param2, param3, names);
+	int const writer_return_val = gmsh_io.writeToFile(file_name);
+
+	if (multiple_geometries)
+	{
+		geo_objects.removeSurfaceVec(merge_name);
+		geo_objects.removePolylineVec(merge_name);
+		geo_objects.removePointVec(merge_name);
+	}
+	return (writer_return_val==1) ? 0 : 3;
+}
+
 bool GMSHInterface::isGMSHMeshFile(const std::string& fname)
 {
 	std::ifstream input(fname.c_str());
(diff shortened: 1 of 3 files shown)


FileIO/GMSHInterface.cpp
```diff
@@ -74,6 +74,44 @@ GMSHInterface::GMSHInterface(GeoLib::GEOObjects & geo_objs,
 	}
 }
 
+int GMSHInterface::writeGeoFile(GeoLib::GEOObjects &geo_objects, std::string const& file_name)
+{
+	std::vector<std::string> names;
+	geo_objects.getGeometryNames(names);
+
+	if (names.size() == 0)
```
endJunction: names.empty()

FileIO/GMSHInterface.cpp
```diff
@@ -74,6 +74,44 @@ GMSHInterface::GMSHInterface(GeoLib::GEOObjects & geo_objs,
...
+	{
+		ERR ("No geometry information available.");
+		return 1;
+	}
+
+	bool const multiple_geometries = (names.size() > 1) ? true : false;
```
endJunction: (names.size() > 1) is already a boolean, no need for ?:

FileIO/GMSHInterface.cpp
```diff
@@ -74,6 +74,44 @@ GMSHInterface::GMSHInterface(GeoLib::GEOObjects & geo_objs,
...
+			return 2;
+		names.clear();
+		names.push_back(merge_name);
+	}
+	else
+		merge_name = names[0];
+
+	double param1(0.5); // mesh density scaling on normal points
+	double param2(0.05); // mesh density scaling on station points
+	size_t param3(2); // points per leaf
```
endJunction: std:: or include c-compatibility header.

FileIO/GMSHInterface.cpp
rinkk: ich need that information later again after i've cleared names for the merge-vector

FileIO/GMSHInterface.h
```diff
@@ -85,6 +85,13 @@ class GMSHInterface : public Writer
 	 */
 	static MeshLib::Mesh* readGMSHMesh (std::string const& fname);
 
+	/**
+	 * Export script for writing geo files.
+	 * To do this, all geometries currently loaded are merged, the merged result is written to a
+	 * file and then the merged geometry is removed again.
+	 */
```
endJunction: I recommend to describe the different return values here.

FileIO/GMSHInterface.cpp
endJunction: I mean that this line
```bool const multiple_geometries = (names.size() > 1) ? true : false;```
is equivalent to
```bool const multiple_geometries = names.size() > 1;```

Comment (rinkk) [inline FileIO/GMSHInterface.cpp]: dumb me ....

**Comment by bilke** [comment]
:+1:

**Comment by waltherm** [comment]
I get compiler warnings:
```
[ 97%] Building CXX object Applications/DataExplorer/CMakeFiles/DataExplorer.dir/mainwindow.cpp.o
bin/Da	/home/waltherm/30_GIT/ogs6/ogs/Applications/DataExplorer/mainwindow.cpp: In member function ‘void MainWindow::callGMSH(std::vector<std::basic_string<char> >&, unsigned int, double, double, double, bool)’:
/home/waltherm/30_GIT/ogs6/ogs/Applications/DataExplorer/mainwindow.cpp:945:34: warning: ignoring return value of ‘int system(const char*)’, declared with attribute warn_unused_result [-Wunused-result]
      system(gmsh_command.c_str());
                                  ^
/home/waltherm/30_GIT/ogs6/ogs/Applications/DataExplorer/mainwindow.cpp:962:35: warning: ignoring return value of ‘int system(const char*)’, declared with attribute warn_unused_result [-Wunused-result]
     system(remove_command.c_str());
                                   ^
```

besides that, :+1:

rinkk:
neither MSVS nor gcc or clang give warnings.
do you use particularly picky compiler settings? i mean, it's true ... we don't use the return value .... but why should that be a problem?!

waltherm:
I used gcc 4.8.2.

endJunction:
It is a good practice to check the return value of a system call. It might have returned with an error.

rinkk:
i would like to point out that the system call is not part of this pr.

but even if it was i don't think it is straightforward to interpret the return value:
> If command is not a null pointer, the value returned depends on the system and library implementations, but it is generally expected to be the status code returned by the called command, if supported.

(from http://www.cplusplus.com/reference/cstdlib/system/)

endJunction:
@rinkk Sorry, didn't see that the system-call line is not touched here.

PR #597, opened 2015-02-10
Lower-case CMake function names

bilke:
Changed CMake code according to official CMake conventions:

- lower-case function names
- empty end-block statements

No need to review all of this because I used scripts to automatically convert this (see commit messages).


endJunction:
Sometimes its `endif ()` with space, sometimes without. This would be a good point to change to single style.
Same for other function calls.

PR #599, opened 2015-02-11
DOF table subset

endJunction:
This is preparation for the Neumann type boundary conditions.

The PR consists of three parts:
 - adding `MeshSubset::setIntersectionByNodes()`,
 - adding `MeshComponentMap::getSubset()` where the nodes subset is supplied by the set intersection by nodes function,
 - adding `LocalToGlobalIndexMap::deriveBoundaryConstrainedMap()` which is using the get subset function for mesh component map.

This feature in the `LocalToGlobalIndexMap` is used by the Neumann-type boundary condition implementation: The integration of the Neumann-type boundary conditions is essentially another process (like the groundwater flow process) contributing to the global matrix and vectors but it is executed only for boundary elements, for which a smaller DOF table is needed.

You can see ongoing development of the Neumann-type boundary conditions on [endJunction/NeumannBC](https://github.com/endJunction/ogs/tree/NeumannBC) branch.

WIP test for `LocalToGlobalIndexMap::deriveBoundaryConstrainedMap()` missing.

PR #603, opened 2015-02-13
Extract findGlobalIndices() method in LocalToGlobalIndexMap.

endJunction:
No logic change, just extracting a function. Merging this now would help with merging PETSc and NeumannBC stuff later.

This is used in the constructor and in another constructor added in #599.


bilke:
:+1:

wenqing:
:+1:

PR #605, opened 2015-02-18
Use editorconfig

bilke:
This PR adds a configuration file which configures your IDE to use consistent file formatting. Have a look at the config file to see the used settings. Although you need a IDE-specific plugin to make this happen. There are plugins for Visual Studio, Vim, Sublime Text and more.

See http://editorconfig.org for more info.


endJunction:
OK, though I didn't check if it's working or not (and probably shall not use it too).

PR #607, opened 2015-02-19
VTK-mapped properties

bilke:
This PR contains basically 4 parts which can be seen by the subsequent commits of the different authors:

- Some refactoring of the `Properties`-class to add better access methods
- Mapped data arrays in the `InSituLib` allowing for zero-copy usage of OGS mesh data structures (nodes, elements, `int` and `double` property vectors) in VTK
- VTK file input allowing to read the VTK data arrays into OGS mesh properties and the corresponding changes in the Data Explorer
- Some small fixes for the Data Explorer

What is missing at the moment is the correct handling of material ids. The idea is to have them both in a `PropertyVector` as well as in the mesh element itself due to data access problems (e.g. when the ids are stored in `PropertyVector`s only then you cannot access them from a mesh element). This will be added beginning next week.

Some notes / interesting points in the code for easier review:

- The first part can be seen [here](https://github.com/TomFischer/ogs/compare/ReorganizeProperties)
- VTK-mapped properties: [a28ef9](https://github.com/rinkk/ogs/commit/a28ef97269d45d51e88fbcc4d892e96ba130ec9f)
- Conversion from VTK array to mesh property: [9ccbf8](https://github.com/rinkk/ogs/commit/9ccbf848a2a37f6b78b8fdd7888afd5e05caf12c)


Applications/DataExplorer/VtkVis/VtkConsoleOutputWindow.h
```diff
@@ -0,0 +1,38 @@
...
+*
+* \copyright
+* Copyright (c) 2015, OpenGeoSys Community (http://www.opengeosys.org)
+*            Distributed under a Modified BSD License.
+*              See accompanying file LICENSE.txt or
+*              http://www.opengeosys.org/project/license
+*
+*/
+
+#ifndef VtkConsoleOutputWindow_h
```
norihiro-w: change macro name to upper case

InSituLib/VtkMappedMeshSource.cpp
```diff
@@ -78,29 +78,75 @@ int VtkMappedMeshSource::RequestData(vtkInformation *,
...
+		DBUG ("Mesh property \"%s\" with unknown data type.", *it->c_str());
+	}
+
+	output->GetPointData()->ShallowCopy(this->PointData.GetPointer());
+	output->GetCellData()->ShallowCopy(this->CellData.GetPointer());
+
+	return 1;
+}
+
+int VtkMappedMeshSource::addDoubleProperty(MeshLib::Properties const& properties,
```
norihiro-w: why not return boolean?

InSituLib/VtkMappedMeshSource.cpp
```diff
@@ -78,29 +78,75 @@ int VtkMappedMeshSource::RequestData(vtkInformation *,
...
+
+	if(propertyVector->getMeshItemType() == MeshLib::MeshItemType::Node)
+		this->PointData->AddArray(dataArray.GetPointer());
+	else if(propertyVector->getMeshItemType() == MeshLib::MeshItemType::Cell)
+		this->CellData->AddArray(dataArray.GetPointer());
+
+	return 1;
+}
+
+int VtkMappedMeshSource::addIntProperty(MeshLib::Properties const& properties,
```
norihiro-w: It looks for me that there is no difference in addDoubleProperty() and addIntProperty() except for data type. maybe you can define a template function and call it like `addProperty<double>()`, `addProperty<int>()`?

MeshLib/Mesh.h
```diff
@@ -124,6 +124,8 @@ class Mesh : BaseLib::Counter<Mesh>
 	MeshLib::Properties & getProperties() { return _properties; }
 	MeshLib::Properties const& getProperties() const { return _properties; }
 
+	void updateMaterialGroups();
```
norihiro-w: documentation missing

MeshLib/MeshGenerators/VtkMeshConverter.cpp
```diff
@@ -356,8 +358,63 @@ MeshLib::Mesh* VtkMeshConverter::convertUnstructuredGrid(vtkUnstructuredGrid* gr
...
+	convertScalarArrays(*grid, *mesh);
+
+	return mesh;
+}
+
+void VtkMeshConverter::convertScalarArrays(vtkUnstructuredGrid &grid, MeshLib::Mesh &mesh)
+{
+	vtkPointData* point_data = grid.GetPointData();
+	if (point_data == nullptr)
+		return;
```
norihiro-w: why skipping converting cell data?

MeshLib/Properties.h
```diff
@@ -153,21 +150,40 @@ class Properties
...
 			return boost::optional<PropertyVector<T> const&>();
 		}
 	}
 
-	void removeProperty(std::string const& name,
-		MeshItemType mesh_item_type)
+	/// Method to get a vector of property values.
+	template <typename T>
+	boost::optional<PropertyVector<T>&>
+	getProperty(std::string const& name)
```
norihiro-w: can this function be const?

InSituLib/VtkMappedMeshSource.cpp
bilke: Thanks, good idea!

MeshLib/Properties.h
endJunction: Yes. Actually there is a const version of `getProperty` in [line 142](https://github.com/ufz/ogs/pull/607/files#diff-91ee186ac317eab2fe2b6458c84f37bdR142). This non-const version is used for writing back the results etc.

MeshLib/PropertyVector.h
```diff
@@ -27,14 +27,44 @@ template <typename PROP_VAL_TYPE>
...
 protected:
-	PropertyVector()
-		: std::vector<PROP_VAL_TYPE>()
+	/// @brief The constructor taking meta information for the data.
+	/// @param property_name a string describing the property
+	/// @param mesh_item_type the values of the property are either assigned to
+	/// nodes or cells (see enumeration @MeshItemType) (default:
+	/// MeshItemType::Cell)
+	/// @param tuple_size the number of elements of a tuple (default: 1)
+	PropertyVector(std::string const& property_name,
```
norihiro-w: need explicit, as passing one argument to this constructor is allowed

MeshLib/PropertyVector.h
```diff
@@ -27,14 +27,44 @@ template <typename PROP_VAL_TYPE>
...
-		: std::vector<PROP_VAL_TYPE>(size)
+	/// @brief The constructor taking meta information for the data.
+	/// @param n_property_values number of property values (value can be a tuple
+	/// with several entries)
+	/// @param property_name a string describing the property
+	/// @param mesh_item_type the values of the property are either assigned to
+	/// nodes or cells (see enumeration @MeshItemType) (default:
+	/// MeshItemType::Cell)
+	/// @param tuple_size the number of elements of a tuple (default: 1)
+	explicit PropertyVector(std::size_t n_property_values,
```
norihiro-w: no need to have explicit, as it requires at least two parameters

MeshLib/Properties.h
norihiro-w: sorry i'm not getting why we need this non-const version. does this function change any class variables?

norihiro-w:
Concerning material IDs, why not store a pointer instead of an actual value in Element class? The pointer I mean gives access to actual value in a property vector. In this way, we don't need to have two separated memory and keep them syncronized. just idea...

MeshLib/Properties.h
TomFischer: The return value of a const getter has to be also const, which prohibits the modification of the values within the `PropertyVector`. For example, when updating the values.

norihiro-w: i see. thanks for the explanation

Applications/DataExplorer/VtkVis/VtkConsoleOutputWindow.cpp
```diff
@@ -0,0 +1,87 @@
+/**
+*
+* \copyright
+* Copyright (c) 2015, OpenGeoSys Community (http://www.opengeosys.org)
```
endJunction: 2012-2015

Applications/DataExplorer/VtkVis/VtkConsoleOutputWindow.cpp
```diff
@@ -0,0 +1,87 @@
...
+		int len = 0;
+		// Find the next new line in text
+		NewLinePos = strchr(someText, '\n');
+		// if no new line is found then just add the text
+		if(NewLinePos == 0)
+		{
+#ifdef WIN32
+			OutputDebugString(someText);
+#endif
+			cerr << someText;
```
endJunction: Is this `cerr` from the std namespace?

Applications/DataExplorer/VtkVis/VtkConsoleOutputWindow.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+	// Disable warnings
+	std::string someTextString(someText);
+	if((someTextString.find(
+		"This is very expensive for vtkMappedDataArray subclasses, since the scalar array must be generated for each call.") != std::string::npos) ||
+		(someTextString.find("Invalid framebuffer operation") != std::string::npos))
+		return;
+
+	// Create a buffer big enough to hold the entire text
+	char* buffer = new char[strlen(someText)+1];
```
endJunction: I'd recommend to rewrite the algorithm using std::string, instead of char* and libc functions.

Applications/DataExplorer/VtkVis/VtkVisPointSetItem.cpp
```diff
@@ -261,77 +268,52 @@ void VtkVisPointSetItem::SetActiveAttribute( const QString& name )
...
-				}
-				else
-				{
-					_activeArrayName = "";
-					_vtkProps->SetActiveAttribute("Solid Color");
-					_mapper->ScalarVisibilityOff();
-					return;
-				}
-			}
+			_algorithm->SetInputArrayToProcess(0, 0, 0, vtkDataObject::FIELD_ASSOCIATION_POINTS, arrayName);
```
endJunction: Instead of `arrayName` one can directly call `_activeArrayName.c_str()`.

InSituLib/VtkMappedMeshSource.h
```diff
@@ -63,6 +65,9 @@ class VtkMappedMeshSource : public vtkUnstructuredGridAlgorithm
 	VtkMappedMeshSource(const VtkMappedMeshSource &); // Not implemented.
 	void operator=(const VtkMappedMeshSource &);      // Not implemented.
 
+	template<typename T> bool addProperty(MeshLib::Properties const& properties,
```
endJunction: A short description of addProperty would be nice.

endJunction:
Is it feasible to resolve the remaining issues of this PR before end of the next week?

PR #608, opened 2015-02-19
More verbose error output when reading GML files

### LINKED ISSUE #590: add more warning info when tag missing in gml
I am not supplied with enough info about what tag is missing in a .gml:
```warning: BoostXmlGmlInterface::readPoints(): Attribute missing in <point> tag. Skipping point...```
Here, "ID" is missing. Would be good to add that output.
*BTW: why do I have to give an ID?*

(this works for polylines: ```warning: BoostXmlGmlInterface::readPolylines(): Attribute "id" missing in <polyline> tag. Skipping polyline..```)

### PR DESCRIPTION
**Pull request by bilke**
When attributes are missing the corresponding element is outputted in a JSON representation. Therefore I added `BaseLib::propertyTreeToString(boost::property_tree::ptree const& tree)`

Fixes #590.

### DIFF
--- a/BaseLib/StringTools.cpp
+++ b/BaseLib/StringTools.cpp
@@ -14,13 +14,13 @@
 
 #include "StringTools.h"
 
-#include "logog/include/logog.hpp"
-
-#include <boost/property_tree/json_parser.hpp>
-
 #include <algorithm>
 #include <cctype>
 #include <iomanip>
+
+#include "logog/include/logog.hpp"
+
+#include <boost/property_tree/json_parser.hpp>
 
 
 namespace BaseLib
--- a/BaseLib/StringTools.cpp
+++ b/BaseLib/StringTools.cpp
@@ -14,12 +14,13 @@
 
 #include "StringTools.h"
 
+#include "logog/include/logog.hpp"
+
+#include <boost/property_tree/json_parser.hpp>
+
 #include <algorithm>
 #include <cctype>
 #include <iomanip>
-
-// ThirdParty/logog
-#include "logog/include/logog.hpp"
 
 
 namespace BaseLib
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: FileIO/XmlIO/Boost/BoostXmlGmlInterface.cpp [hunk 1]
```diff
@@ -121,9 +121,10 @@ void BoostXmlGmlInterface::readPoints(boost::property_tree::ptree const & points
 		double   p_z  = static_cast<double>  (point.second.get("<xmlattr>.z",  std::numeric_limits<double>::max()));
 		std::string p_name = point.second.get("<xmlattr>.name", "");
 
-		if ( p_id == std::numeric_limits<unsigned>::max() || p_x == std::numeric_limits<double>::max() || 
-			 p_y  == std::numeric_limits<double>::max()   || p_z == std::numeric_limits<double>::max() ) 
-			WARN("BoostXmlGmlInterface::readPoints(): Attribute missing in <point> tag. Skipping point...")
+		if ( p_id == std::numeric_limits<unsigned>::max() || p_x == std::numeric_limits<double>::max() ||
+			 p_y  == std::numeric_limits<double>::max()   || p_z == std::numeric_limits<double>::max() )
+		WARN("BoostXmlGmlInterface::readPoints(): Skipping point, attribute missing in <point> tag:\n%s",
```
TomFischer: Indentation seems wrong in line 126.

TomFischer:
Travis fails?! I do not think the commit/PR is the reason. Is it possible to disable the tests temporarily that are causing Travis to fail?

Your PR: :+1:

BaseLib/StringTools.cpp
```diff
@@ -14,13 +14,14 @@
 
 #include "StringTools.h"
 
+#include "logog/include/logog.hpp"
```
norihiro-w: include these third party headers after C++ headers

BaseLib/StringTools.h
```diff
@@ -15,6 +15,8 @@
 #ifndef STRINGTOOLS_H
 #define STRINGTOOLS_H
 
+#include <boost/property_tree/ptree.hpp>
```
norihiro-w: same here

PR #609, opened 2015-02-23
Fix a bug in FEFLOW import

norihiro-w:
This PR includes
- fix a bug in importing a FEFLOW file reported by Sophie. Dimension of node coords given under "supermesh" section may not be equal to the dimension of the problem (e.g. only x,y coords are given while a problem is 3d).
- minor refactoring in FEFLOW2OGS.cpp


endJunction:
OK

Issue #606, opened 2015-02-18
Stalled travis testrunner for MPITest_Math.CheckInterface_PETScVector test.


endJunction:
```
Note: Google Test filter = MPITest*
Note: Randomizing tests' orders with a seed of 50138 .
[==========] Running 8 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 8 tests from MPITest_Math
[ RUN      ] MPITest_Math.CheckInterface_PETScVector
No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.
The build has been terminated
```

I have restarted travis builds 3 times today at different PRs. Could be an indication for a race condition in that test.

Obsolete after 18. of March. 2015

endJunction:
Three more restarts today (22. Feb.). Again USE_PETSc configuration, everything builds, tests target runs ok, tests_mpi stalled after few tests run. The last messages (all of the [OK] from gtest) were different.

wenqing:
Fixed (#610).

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/606).

PR #610, opened 2015-02-23
[CM] Fixed MPITest on travis.

wenqing:
Fixed the problem reported in #606.

### DIFF
--- a/Tests/CMakeLists.txt
+++ b/Tests/CMakeLists.txt
@@ -94,7 +94,7 @@ if(OGS_USE_PETSC)
 		DEPENDS testrunner
 	)
 	add_custom_target(tests_mpi
-		mpirun -np 3 $<TARGET_FILE:testrunner> ${TESTRUNNER_ADDITIONAL_ARGUMENTS}  --gtest_filter=MPITest*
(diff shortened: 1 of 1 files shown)


bilke:
Please explain why you are doing things when you alter functionality. In the parallel case (`mpirun -np 3 ...`) maybe it is necessary to remove the multiple run gtest parameter but I guess in the serial case (`mpirun -np 1 ...`) it is not necessary to be removed.

wenqing:
@bilke For the case of ``-np 3``, arguments ``--gtest_shuffle --gtest_repeat=3`` cause the problem,  because that each core run the same code but the tests are executed randomly at the same time, however MPI code needs execute the same piece of code at the same time.

For the case of ``np -1``, you are right that the arguments are OK. I will remove the last commit. However we do not need to perform such tests under travis because such tests are exactly the same tests by the serial version compiled by gcc.   The case of ``np -1`` is only used for run test manually. For example,  if one works the source code under USE_MPI, and he wants to test the whole code, he then can try with ``make tests`` and ``make tests_mpi``.

wenqing:
BTW,  the tests of ``np -1`` have already been removed for travis in ufz/ogs repository.

PR #611, opened 2015-02-24
Added build target archive-data on Linux / Mac.

bilke:
This target creates a gzipped tar and a zip-file of all files in `Tests/Data`. Symlinks are resolved and files are stored with their correct name.

Requirements:

- `zip`
- `realpath`


bilke:
Downloads can be found at http://docs.opengeosys.org/download (at the bottom of the page).

PR #615, opened 2015-02-26
Correct triangle Gauss integration

endJunction:
Currently in the `GroundwaterFlowProcess` the regular Gauss integration scheme was used. There is a special integration rule for triangle elements, which is different from the "regular" elements. This is used now.

Also move the choice of the integration scheme into the `LocalDataInitializer` from the processes.


norihiro-w:
:+1:

PR #616, opened 2015-02-27
[GeoLib] Output warning in case a polyline-name already exists

TomFischer:
as titled


endJunction:
:fast_forward:

rinkk:
2 things:
1) the warnings should specifically mention that the name of the polyline is removed
2) wouln't it be better if GEOObjects checked/handled this? this would require only one implementation of the check instead of the two you have now and it would also work for all other geometry-formats where names might pop up (e.g. shape files)

TomFischer:
> 2) wouln't it be better if GEOObjects checked/handled this? this would require only one implementation of the check instead of the two you have now and it would also work for all other geometry-formats where names might pop up (e.g. shape files)

The current design does not allow this. I would be happy if I could find the time to correct this. But there are more urgent things to do at the moment.

PR #617, opened 2015-02-27
Data explorer dialog calls

rinkk:
Changed a few dialogue calls such that dialogues get only the information they absolutely need.


endJunction:
OK

bilke:
:+1:

PR #619, opened 2015-02-27
Install dependencies for ogs cli

bilke:
This PR adds dependency packaging for the ogs cli. I would like to include this into the 6.0.1 release to have redistributable packages on the download page.


endJunction:
:fast_forward:

PR #620, opened 2015-03-09
[Fix] handling errors when reading raster files

rinkk:
* added error handling if file cannot be read
* added alternative tags supported by new arcgis version


endJunction:
OK

PR #621, opened 2015-03-10
[Fix] Extraction of mesh elements based on bounding box

rinkk:
we detected rounding error when string representations of double numbers are converted back to doubles (specifically in the bounding box specifications of the element extraction dialog)
this fix changes the dialog to only use text to double conversion for text fields that have been edited.

### DIFF
--- a/Applications/DataExplorer/DataView/MeshElementRemovalDialog.cpp
+++ b/Applications/DataExplorer/DataView/MeshElementRemovalDialog.cpp
@@ -44,7 +44,6 @@ MeshElementRemovalDialog::MeshElementRemovalDialog(const ProjectData &project, Q
 		OGSError::box("No meshes available.");
 		QMetaObject::invokeMethod(this, "close", Qt::QueuedConnection);
 	}
-
 }
 
 MeshElementRemovalDialog::~MeshElementRemovalDialog()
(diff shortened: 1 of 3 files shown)


Applications/DataExplorer/DataView/MeshElementRemovalDialog.cpp
```diff
@@ -132,19 +139,18 @@ void MeshElementRemovalDialog::on_boundingBoxCheckBox_toggled(bool is_checked)
...
+		MeshLib::Node const minAABB = aabb.getMinPoint();
+		MeshLib::Node const maxAABB = aabb.getMaxPoint();
 		this->xMinEdit->setText(QString::number(minAABB[0], 'f'));
 		this->xMaxEdit->setText(QString::number(maxAABB[0], 'f'));
 		this->yMinEdit->setText(QString::number(minAABB[1], 'f'));
 		this->yMaxEdit->setText(QString::number(maxAABB[1], 'f'));
 		this->zMinEdit->setText(QString::number(minAABB[2], 'f'));
 		this->zMaxEdit->setText(QString::number(maxAABB[2], 'f'));
-
+		std::fill(aabb_edits.begin(), aabb_edits.end(), false);
```
endJunction: Just for information: the line can be a little shorter with `aabb_edits.fill(false);`, but is ok the way it is.

Applications/DataExplorer/DataView/MeshElementRemovalDialog.cpp
```diff
@@ -78,13 +77,21 @@ void MeshElementRemovalDialog::accept()
...
-		                        this->zMaxEdit->text().toDouble(),0);
-		ex.searchByBoundingBox(min, max);
+		std::vector<MeshLib::Node*> const& nodes (_project.getMesh(this->meshNameComboBox->currentText().toStdString())->getNodes());
+		GeoLib::AABB<MeshLib::Node> aabb(nodes.begin(), nodes.end());
+		MeshLib::Node minAABB = aabb.getMinPoint();
+		MeshLib::Node maxAABB = aabb.getMaxPoint();
+		double const eps (std::numeric_limits<double>::epsilon());
+
+		// only extract bounding box parameters that have been edited (otherwise there will be rounding errors!)
+		minAABB[0] = (aabb_edits[0]) ? this->xMinEdit->text().toDouble() : (minAABB[0] - eps);
```
endJunction: Why is the eps subtracted from the original value, if there was no edit?

rinkk: when checking if a point p is contained within an ```AABB```, the method returns true if p is "larger" (>) than the min-point and "smaller" (<) than the max-point. i.e. if another point is located on the boundary, it is per definition *not* contained. to make sure such points are not removed, eps is subtracted or added, respectively.

bilke:
:+1:

Applications/DataExplorer/DataView/MeshElementRemovalDialog.cpp
endJunction: Can it happen, that  this function is invoked multiple times? If so, then the bounding box will grow every time by an eps.
I was wondering if this is the right location to make the bounding box larger.
The method [`AABB::containsPoint()`](http://doxygen.opengeosys.org/d6/d5d/classGeoLib_1_1AABB.html#ac4478094dbf00a1dd5de6f5d27f8d4db) checks for point `p` if `min <= p <= max` for all x, y, z, so it includes the boundary points already.

rinkk: it does?
then i mixed sth. up and will remove the eps.

Applications/DataExplorer/DataView/MeshElementRemovalDialog.cpp
```diff
@@ -78,13 +77,21 @@ void MeshElementRemovalDialog::accept()
...
 	{
-		const MeshLib::Node min(this->xMinEdit->text().toDouble(),
-		                        this->yMinEdit->text().toDouble(),
-		                        this->zMinEdit->text().toDouble(),0);
-		const MeshLib::Node max(this->xMaxEdit->text().toDouble(),
-		                        this->yMaxEdit->text().toDouble(),
-		                        this->zMaxEdit->text().toDouble(),0);
-		ex.searchByBoundingBox(min, max);
+		std::vector<MeshLib::Node*> const& nodes (_project.getMesh(this->meshNameComboBox->currentText().toStdString())->getNodes());
+		GeoLib::AABB<MeshLib::Node> aabb(nodes.begin(), nodes.end());
```
TomFischer: Maybe `aabb` can be `const`?

PR #622, opened 2015-03-11
Neumann type boundary conditions.

endJunction:
For discussion I propose an implementation of Neumann type boundary conditions used in the GroundwaterFlowProcess.

Benchmark description in http://docs.opengeosys.org/docs/benchmarks/elliptic/groundwater-flow-neumann

Known issues:
 - There are multiple passes needed to create the finally used objects (`NeumannBC`) in the process: Temporarly store parsed configuration in `NeumannBcConfig` objects, create the `NeumannBC` from the process using the config. There are multiple constructors and `init()`-like functions, which I find not so concise.
 - There is no concept for a function and its region of definition.


wenqing:
ctest falling.

ProcessLib/NeumannBCAssembler.h
```diff
@@ -0,0 +1,122 @@
...
+    {
+        A.add(indices, *_localA);
+        rhs.add(indices.rows, *_localRhs);
+    }
+
+private:
+    std::vector<ShapeMatrices> _shape_matrices;
+    double _neumann_bc_value;
+
+    std::unique_ptr<NodalMatrixType> _localA;
```
wenqing: Why there is a matrix? Neumann BC  is a known value, and it contributes to RHS through the associated weak form.

ProcessLib/NeumannBC.h
```diff
@@ -0,0 +1,155 @@
...
+    integrate(GlobalSetup const& global_setup)
+    {
+        global_setup.execute(*_global_assembler, _local_assemblers);
+    }
+
+
+private:
+    double const _scaling_value;
+
+    MathLib::ConstantFunction<double> const _function;
```
wenqing: I guess it is used as scaling factor. Documenation is needed.

ProcessLib/NeumannBC.h
```diff
@@ -0,0 +1,155 @@
...
+
+    std::vector<MeshLib::Element const*> _elements;
+    MeshLib::MeshSubset const* _mesh_subset_all_nodes = nullptr;
+    std::vector<MeshLib::MeshSubsets*> _all_mesh_subsets;
+
+    std::unique_ptr<AssemblerLib::LocalToGlobalIndexMap> _local_to_global_index_map;
+
+    unsigned const _integration_order;
+
+    using GlobalAssembler =
```
wenqing: It looks like that including MatrixType here is just for a unqiue interface. Am I right?

ProcessLib/NeumannBCAssembler.h
```diff
@@ -0,0 +1,122 @@
...
+            _localRhs->noalias() += sm.N * _neumann_bc_value
+                        * sm.detJ * wp.getWeight();
+        }
+    }
+
+    void
+    addToGlobal(GlobalMatrix& A, GlobalVector& rhs,
+            AssemblerLib::LocalToGlobalIndexMap::RowColumnIndices const& indices) const
+    {
+        A.add(indices, *_localA);
```
wenqing: Changing the global matrix is not needed.

norihiro-w:
:+1:

ProcessLib/BoundaryCondition.h
```diff
@@ -63,8 +57,12 @@ class UniformDirichletBoundaryCondition : public BoundaryCondition
 
         _value = config.get<double>("value", 0);
         DBUG("Using value %g", _value);
+
+        f = new MathLib::ConstantFunction<double>(_value);
```
norihiro-w: is this variable deleted somewhere?

norihiro-w:
Sorry for my late response.

In my opinion, it's not bad to have both `NeumannBcConfig` and `NeumannBC` if the former is used for storing general description of BC (geometry, distribution) and the latter is used for implementing the BC for FEM (boundary element search, assembly, integration). You may create `DirichletBcConfig` to make it consistent with Neumann BC.

endJunction:
I will update the Dirichlet BCs in another PR, such that both, the Neumann and Dirichlet BC interfaces/usage are equal.

endJunction:
After Nori's comments I did only some cleanups and documentation. No logic changes.

norihiro-w:
:+1:

ProcessLib/GroundwaterFlowProcess.h
```diff
@@ -179,9 +203,9 @@ class GroundwaterFlowProcess : public Process
     }
 
 private:
-    ProcessVariable const* _hydraulic_head = nullptr;
+    ProcessVariable* _hydraulic_head = nullptr;
 
-    double const _hydraulic_conductivity = 1e-6;
+    double const _hydraulic_conductivity = 1;
```
TomFischer: The variable is `const`. Is it intended that the value can not be changed later on?

ProcessLib/NeumannBc.h
```diff
@@ -0,0 +1,179 @@
...
+
+#include "NeumannBcConfig.h"
+#include "NeumannBcAssembler.h"
+#include "MeshLib/MeshSearcher.h"
+
+namespace ProcessLib
+{
+
+/// The NeumannBc class is a process which is integrating a single, Neumann type
+/// boundary conditions in to the global matrix and the right-hand-side.
```
TomFischer: Spelling: *condition* instead of *conditions*

ProcessLib/NeumannBc.h
```diff
@@ -0,0 +1,179 @@
...
+/// Neumann boundary conditions.
+template <typename GlobalSetup_>
+class NeumannBc
+{
+public:
+    /// Create a Neumann boundary condition process from given config,
+    /// DOF-table, and a mesh subset.
+    /// A local DOF-table, a subset of the given one, is constructed.
+    NeumannBc(
+        NeumannBcConfig* bc,
```
TomFischer: Why not const?

ProcessLib/NeumannBc.h
```diff
@@ -0,0 +1,179 @@
...
+        _local_assemblers.resize(_elements.size());
+        LocalAssemblerBuilder local_asm_builder(
+            initializer, *_local_to_global_index_map);
+
+        auto elementValueLookup = [this](MeshLib::Element const& e)
+        {
+            return _function();
+        };
+
+        DBUG("Calling local Neumann assembler builder for neumann boundary elements.");
```
TomFischer: neumann -> Neumann

ProcessLib/GroundwaterFlowProcess.h
endJunction: Yes, this is intended. When we start reading mesh properties, this will change.

Issue #625, opened 2015-03-16
Wrong PetSc package on Travis


bilke:
Travis builds are currently failing because in the PetSc configs the PetSc package cannot be installed anymore. Maybe the name or package repository has changed. See e.g. https://travis-ci.org/rinkk/ogs/jobs/54600770
@wenqing can you please have a look?

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/625).

PR #626, opened 2015-03-17
Use a new PPA for PETSc

norihiro-w:
as titled related to issue #625


wenqing:
:+1:

norihiro-w:
it would be better if we have our own ppa in future

PR #627, opened 2015-03-18
Petsc/mpi tests.

endJunction:
Three commits unrelated to each other:
Adding one test. Change configuration of petsc/mpi tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,7 +55,7 @@ script:
   - cd build
   - cmake $CMAKE_ARGS ..
   - make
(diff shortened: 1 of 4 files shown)


bilke:
:+1:

bilke:
Will there be MPI tests in the future (via ctest)?

wenqing:
@bilke MPI tests are still there but only for the code compiled with option ``USE_PETSC``.

TomFischer:
Added test is not tested.

PR #632, opened 2015-03-25
CSV interface for reading points

rinkk:
Reader for CSV files.
It is possible to read:
* 2D/3D Points, either specifying row-names (for CSV-files *with* header) or row-indices (for *headerless* files)
* arbitrary rows, either by specifying a row-name (for CSV-files *with* header) or a row-index (for *headerless* files)

Row-names and indeces don't have to be in correct order, i.e. the row for x-coordinates can be located *after* the row of y-coordinates. If something cannot be matched in one line, that line is simply skipped.

I also included a simple method for reading 3d points where it is assumed that x,y,z are located in rows 0,1,2. It was written first and I simply didn't remove it later. It should be faster than the other ones.


norihiro-w:
do you actually mean "column" instead of "row"?

FileIO/CsvInterface.cpp
```diff
@@ -0,0 +1,174 @@
...
+
+		if (fields.size() < 3)
+		{
+			ERR ("Line %d contains not enough rows of data. Skipping line...", line_count);
+			error_count++;
+			continue;
+		}
+		it = fields.begin();
+		std::array<double, 3> point;
+		point[0] = strtod(it->c_str(), 0);
```
TomFischer: `strtod` is a function of the stl (defined in header cstdlib, that is not included). For this reason I would write `std::strtod`.

FileIO/CsvInterface.cpp
```diff
@@ -0,0 +1,174 @@
...
+		std::list<std::string> const fields = BaseLib::splitString(line, delim);
+
+		if (fields.size() < 3)
+		{
+			ERR ("Line %d contains not enough rows of data. Skipping line...", line_count);
+			error_count++;
+			continue;
+		}
+		it = fields.begin();
+		std::array<double, 3> point;
```
TomFischer: Why create a `std::array<doubel,3>` here and not a `GeoLib::Point` that is created later on anyway?

FileIO/CsvInterface.cpp
```diff
@@ -0,0 +1,174 @@
...
+		point[1] = strtod((++it)->c_str(), 0);
+		point[2] = strtod((++it)->c_str(), 0);
+		points.push_back(new GeoLib::Point(point[0], point[1], point[2]));
+	}
+	return error_count;
+}
+
+int CsvInterface::readPointsFromCSV(std::string const& fname, char delim,
+                                    std::vector<GeoLib::Point*> &points,
+                                    std::string const& x_row_name,
```
TomFischer: In many cases I assume column is meant instead of row?

FileIO/CsvInterface.cpp
```diff
@@ -0,0 +1,174 @@
...
+			ERR ("Line %d contains not enough rows of data. Skipping line...", line_count);
+			error_count++;
+			continue;
+		}
+		it = fields.begin();
+		std::array<double, 3> point;
+		point[0] = strtod(it->c_str(), 0);
+		point[1] = strtod((++it)->c_str(), 0);
+		point[2] = strtod((++it)->c_str(), 0);
+		points.push_back(new GeoLib::Point(point[0], point[1], point[2]));
```
TomFischer: The `GeoLib::Point` can take the `std::array<double, 3>` as input directly.

FileIO/CsvInterface.h
```diff
@@ -0,0 +1,179 @@
...
+				ERR ("Line %d contains not enough rows of data. Skipping line...", line_count);
+				error_count++;
+				continue;
+			}
+			it = fields.begin();
+			std::advance(it, row_idx);
+
+			std::istringstream stream(*it);
+			T value;
+			stream >> value;
```
TomFischer: Maybe you can check the stream `stream` in order to detect errors?
```if (!(stream >> value)) {
  ERR("...");
  continue;
}```

**Comment by TomFischer** [comment]
It would be nice to have some tests. Especially test cases checking invalid inputs.

File: FileIO/CsvInterface.cpp (hunk 2 above)
Comment (rinkk) [inline FileIO/CsvInterface.cpp]: the x, y and z coordinates are read successively, so I'd either have to create a point with fake-coordinates and adjust these coordinates later on or I save the three coordinates in temp variables, which I am doing here.

File: FileIO/CsvInterface.cpp (hunk 3 above)
Comment (rinkk) [inline FileIO/CsvInterface.cpp]: in *all* cases .... i totally mixed that up ...

File: FileIO/CsvInterface.cpp (hunk 4 above)
Comment (rinkk) [inline FileIO/CsvInterface.cpp]: no, it cannot ... it can take a ```double*``` or three doubles.

Comment (TomFischer) [inline FileIO/CsvInterface.cpp]: I think it is comfortable to have such a constructor. I added it in PR https://github.com/ufz/ogs/pull/634.

File: FileIO/CsvInterface.h [hunk 6]
```diff
@@ -0,0 +1,183 @@
...
+	/** 
+	 * Reads 3D points from a CSV file. It is assumed that the file has a header
+	 * specifying a name for each of the columns. The first three columns will be
+	 * interpreted as x-, y- and z-coordinate, respectively.
+	 * \param fname    Name of the file to be read
+	 * \param delim    Deliminator, default is ','
+	 * \param points   A vector containing the 3D points read from the file
+	 * \return An error code (0 = ok, 0<i<max = number of skipped lines, -1 error reading file)
+	 */
+	static int readPointsFromCSV(std::string const& fname, char delim,
```
norihiro-w: I would remove "FromCSV" from the function name, because it is obvious from the class name

FileIO/CsvInterface.h
```diff
@@ -0,0 +1,183 @@
...
+	 * \return An error code (0 = ok, 0<i<max = number of skipped lines, -1 error reading file)
+	 */
+	template <typename T>
+	static int readColumnFromCSV(std::string const& fname, char delim,
+	                          std::vector<T> &data_array,
+	                          std::string const& column_name)
+	{
+		std::ifstream in(fname.c_str());
+		if (!in.is_open()) {
+			ERR ("CsvInterface::readPointsFromCSV(): Could not open file %s.", fname.c_str());
```
norihiro-w: the function name in the error message is not correct

FileIO/CsvInterface.cpp
```diff
@@ -0,0 +1,183 @@
...
+			ERR ("Error converting data to coordinates in line %d.", line_count);
+		}
+	}
+	return error_count;
+}
+
+std::size_t CsvInterface::findColumn(std::string const& line, char delim, std::string const& column_name)
+{
+	std::list<std::string> const fields = BaseLib::splitString(line, delim);
+	if (fields.size() < 3)
```
norihiro-w: why do we need at least 3 columns?

rinkk: good point ... i wrote this, when i needed it for reading points. you need 3 columns for points (x,y,z) but obviously not for the general case. i'll change it.

PR #633, opened 2015-03-30
Reorganize classes Properties and PropertyVector

TomFischer:
This PR improves the usability of the classes `Properties`and `PropertyVector` due to
- Renaming some methods in Properties.
- Adding some methods that are needed for the use.
- Using base class and derived classes instead of `boost::any` in `Properties`.
- Changed the implementation of class template `PropertyVector` in case of pointer types.
- Add more tests.

Many of the commits of this PR are broken off the PR https://github.com/ufz/ogs/pull/607 to make review easier. It will follow further PRs based on this PR. If possible I would appreciated I quick review :smile:

### DIFF
--- a/MeshLib/Properties.h
+++ b/MeshLib/Properties.h
@@ -69,7 +69,7 @@ public:
 		auto entry_info(
 			_properties.insert(
 				std::pair<std::string, boost::any>(
-					name, boost::any(PropertyVector<T>())
+					name, boost::any(PropertyVector<T>(name, mesh_item_type))
 				)
 			)
 		);
@@ -125,7 +125,8 @@ public:
 			_properties.insert(
 				std::pair<std::string, boost::any>(
 					name,
-					boost::any(PropertyVector<T>(n_prop_groups, item2group_mapping))
+					boost::any(PropertyVector<T>(n_prop_groups,
+						item2group_mapping, name, mesh_item_type))
 				)
 			)
 		);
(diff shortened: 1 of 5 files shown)


bilke:
:+1:

MeshLib/Properties.h
```diff
@@ -194,29 +208,53 @@ class Properties
...
-
-		bool operator<(PropertyKeyType const& other) const
-		{
-			int res(name.compare(other.name));
-			if (res == 0) {
-				return mesh_item_type < other.mesh_item_type;
-			}
-			return res < 0;
+		std::vector<std::string> names;
+		for (auto it(_properties.cbegin()); it != _properties.cend(); it++)
```
norihiro-w: ++it or use for (auto name : _properties)

MeshLib/PropertyVector.h
```diff
@@ -12,29 +12,84 @@
...
 {
 
+template <typename T>
+class PropertyVector;
+
+class PropertyVectorBase
+{
+public:
+	virtual PropertyVectorBase* clone(
+		std::vector<std::size_t> const& exclude_positions
```
norihiro-w: how about changing here to pass a pointer instead of a reference and setting a default value of nullptr?

norihiro-w:
Properties.h is getting longer and longer. it would be nice if you create Properties.hpp and move definitions of template functions in Properties.h there in the future.

MeshLib/PropertyVector.h
TomFischer: References are easier to check for validity in this case.

norihiro-w:
:+1:

PR #634, opened 2015-03-31
MathLib Point3d operator<

TomFischer:
- Implementation and tests for `Point3d::operator<`.
- Moved tests for `Point3d` from `GeoLib` to `MathLib`.
- Added constructor to GeoLib::Point class.


endJunction:
:fast_forward:

PR #637, opened 2015-04-13
Csv interface issues

TomFischer:
The locale at my environment was set to a comma instead of a period to separate the integer part from the non-integer part (for instance 1,2 instead of 1.2). Since some of the CsvInterface methods use locale dependend functions to read floating point numbers some of the CsvInterfaceTests failed. For this reason, I suggest to set the locale at start of the testrunner to the following: `setlocale(LC_NUMERIC,"C");`

Additional, there is one commit removing some warning in clang and one commit that remove some unnecessary whitespaces.

### DIFF
--- a/FileIO/CsvInterface.cpp
+++ b/FileIO/CsvInterface.cpp
@@ -65,7 +65,7 @@ int CsvInterface::readPoints(std::string const& fname, char delim,
                              std::string const& z_column_name)
 {
 	std::ifstream in(fname.c_str());
-	std::array<std::string, 3> const column_names = { x_column_name, y_column_name, z_column_name };
+	std::array<std::string, 3> const column_names = {{x_column_name, y_column_name, z_column_name}};
 
 	if (!in.is_open()) {
 		ERR ("CsvInterface::readPoints(): Could not open file %s.", fname.c_str());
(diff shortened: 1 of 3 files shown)


bilke:
:+1:

FileIO/CsvInterface.cpp
```diff
@@ -74,12 +74,13 @@ int CsvInterface::readPoints(std::string const& fname, char delim,
...
 	getline(in, line);
-	std::array<std::size_t, 3> const column_idx = 
-		{ CsvInterface::findColumn(line, delim, x_column_name),
-		  CsvInterface::findColumn(line, delim, y_column_name),
-		  (z_column_name.empty()) ? CsvInterface::findColumn(line, delim, y_column_name) : 
-		                            CsvInterface::findColumn(line, delim, z_column_name) };
-	
+	std::array<std::size_t, 3> const column_idx =
+		{{ CsvInterface::findColumn(line, delim, x_column_name),
+			CsvInterface::findColumn(line, delim, y_column_name),
```
rinkk: can you please align the three components?

FileIO/CsvInterface.cpp
```diff
@@ -114,18 +115,19 @@ int CsvInterface::readPoints(std::ifstream &in, char delim,
...
+	std::sort(order.begin(), order.end(),
 		[&column_idx](std::size_t idx1, std::size_t idx2) {return column_idx[idx1] < column_idx[idx2];});
-	std::array<std::size_t, 3> const column_advance = 
-		{ column_idx[order[0]], 
-		  column_idx[order[1]] - column_idx[order[0]], 
-		  column_idx[order[2]] - column_idx[order[1]] };
+	std::array<std::size_t, 3> const column_advance =
+		{{ column_idx[order[0]],
+			column_idx[order[1]] - column_idx[order[0]],
+			column_idx[order[2]] - column_idx[order[1]] }};
```
rinkk: can you please align the three components?

rinkk:
:+1: (apart from the two small formatting issues)

FileIO/CsvInterface.cpp
TomFischer: Components are aligned by tabs. Now, there is not a mixture using tabs and spaces anymore.

FileIO/CsvInterface.cpp
TomFischer: Same as above.

FileIO/CsvInterface.cpp
rinkk: i mean you should format the arrays as they were before with all array-components starting at the same row (without additional tabs for the 2nd and 3rd component)

TomFischer: A discussion with Lars hopefully enlighted me. There is a difference between indentation and alignment. Imho indentation only is better than indentation + alignment. But here the opinions are different. I will revert changes above.

TomFischer: Done.

FileIO/CsvInterface.cpp
TomFischer: Fixed.

PR #639, opened 2015-04-14
Using new interface of Mesh to access Properties and PropertyVector in class MeshRevision

TomFischer:
- Main changes in MeshLib/MeshEditing/MeshRevision.

### DIFF
--- a/MeshLib/MeshEditing/MeshRevision.cpp
+++ b/MeshLib/MeshEditing/MeshRevision.cpp
@@ -339,13 +339,13 @@ unsigned MeshRevision::subdivideQuad(MeshLib::Element const*const quad,
 	tri1_nodes[0] = nodes[quad->getNode(0)->getID()];
 	tri1_nodes[1] = nodes[quad->getNode(1)->getID()];
 	tri1_nodes[2] = nodes[quad->getNode(2)->getID()];
-	new_elements.push_back(new MeshLib::Tri(tri1_nodes, quad->getValue()));
+	new_elements.push_back(new MeshLib::Tri(tri1_nodes));
 
 	MeshLib::Node** tri2_nodes = new MeshLib::Node*[3];
 	tri2_nodes[0] = nodes[quad->getNode(0)->getID()];
 	tri2_nodes[1] = nodes[quad->getNode(2)->getID()];
 	tri2_nodes[2] = nodes[quad->getNode(3)->getID()];
-	new_elements.push_back(new MeshLib::Tri(tri2_nodes, quad->getValue()));
+	new_elements.push_back(new MeshLib::Tri(tri2_nodes));
 
 	return 2;
 }
@@ -361,7 +361,7 @@ unsigned MeshRevision::subdivideHex(MeshLib::Element const*const hex,
 	prism1_nodes[3] = nodes[hex->getNode(4)->getID()];
 	prism1_nodes[4] = nodes[hex->getNode(6)->getID()];
 	prism1_nodes[5] = nodes[hex->getNode(5)->getID()];
-	MeshLib::Prism* prism1 (new MeshLib::Prism(prism1_nodes, hex->getValue()));
+	MeshLib::Prism* prism1 (new MeshLib::Prism(prism1_nodes));
 	this->subdividePrism(prism1, nodes, new_elements);
 	delete prism1;
 
@@ -372,7 +372,7 @@ unsigned MeshRevision::subdivideHex(MeshLib::Element const*const hex,
 	prism2_nodes[3] = nodes[hex->getNode(0)->getID()];
 	prism2_nodes[4] = nodes[hex->getNode(2)->getID()];
 	prism2_nodes[5] = nodes[hex->getNode(3)->getID()];
-	MeshLib::Prism* prism2 (new MeshLib::Prism(prism2_nodes, hex->getValue()));
+	MeshLib::Prism* prism2 (new MeshLib::Prism(prism2_nodes));
 	this->subdividePrism(prism2, nodes, new_elements);
 	delete prism2;
 
@@ -388,14 +388,14 @@ unsigned MeshRevision::subdividePyramid(MeshLib::Element const*const pyramid,
 	tet1_nodes[1] = nodes[pyramid->getNode(1)->getID()];
 	tet1_nodes[2] = nodes[pyramid->getNode(2)->getID()];
 	tet1_nodes[3] = nodes[pyramid->getNode(4)->getID()];
-	new_elements.push_back(new MeshLib::Tet(tet1_nodes, pyramid->getValue()));
+	new_elements.push_back(new MeshLib::Tet(tet1_nodes));
 
 	MeshLib::Node** tet2_nodes = new MeshLib::Node*[4];
 	tet2_nodes[0] = nodes[pyramid->getNode(0)->getID()];
 	tet2_nodes[1] = nodes[pyramid->getNode(2)->getID()];
 	tet2_nodes[2] = nodes[pyramid->getNode(3)->getID()];
 	tet2_nodes[3] = nodes[pyramid->getNode(4)->getID()];
-	new_elements.push_back(new MeshLib::Tet(tet2_nodes, pyramid->getValue()));
+	new_elements.push_back(new MeshLib::Tet(tet2_nodes));
 
 	return 2;
 }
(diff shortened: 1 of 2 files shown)


bilke:
:+1:

MeshLib/MeshEditing/MeshRevision.cpp
```diff
@@ -67,39 +67,68 @@ MeshLib::Mesh* MeshRevision::simplifyMesh(const std::string &new_mesh_name, doub
...
-		else if (n_unique_nodes < (*elem)->getNBaseNodes() && n_unique_nodes>1)
-			reduceElement(*elem, n_unique_nodes, new_nodes, new_elements, min_elem_dim);
-		else
+		else if (n_unique_nodes < elem->getNBaseNodes() && n_unique_nodes>1) {
+			std::size_t const n_new_elements(reduceElement(
+				elem, n_unique_nodes, new_nodes, new_elements, min_elem_dim)
+			);
+			if (!material_vec)
+				continue;
+			for (std::size_t j(0); j<n_new_elements; ++j)
```
rinkk: i think it'd be better to remove the ```for```-loop and write
```new_material_vec.insert(new_material_vec.end(), n_new_elements, (*material_vec)[k]);```

File: MeshLib/MeshEditing/MeshRevision.cpp [hunk 2]
```diff
@@ -67,39 +67,68 @@ MeshLib::Mesh* MeshRevision::simplifyMesh(const std::string &new_mesh_name, doub
...
 					this->cleanUp(new_nodes, new_elements);
 					return nullptr;
 				}
+				if (!material_vec)
+					continue;
+				for (std::size_t j(0); j<n_new_elements; ++j)
+					new_material_vec->push_back((*material_vec)[k]);
+			} else {
+				new_elements.push_back(MeshLib::copyElement(elem, new_nodes));
+				// transmit material values
```
rinkk: "transmit" is the wrong word here ... better use "copy" or something ...

MeshLib/MeshEditing/MeshRevision.cpp
```diff
@@ -67,39 +67,68 @@ MeshLib::Mesh* MeshRevision::simplifyMesh(const std::string &new_mesh_name, doub
...
+		new_properties.createNewPropertyVector<int>(
+			"MaterialIDs", MeshItemType::Cell, 1
+		)
+	);
+
+	// original data
+	std::vector<MeshLib::Element*> const& elements(this->_mesh.getElements());
+	MeshLib::Properties const& properties(_mesh.getProperties());
+	boost::optional<MeshLib::PropertyVector<unsigned> const&> material_vec(
+		properties.getPropertyVector<unsigned>("MaterialIDs"));
```
rinkk: i think something is wrong here with the sequence of commands. you should first check if the old mesh had a "material id" property and *only* if it did should you create such a property for the new mesh.

additional comment: it might not be relevant now, but in theory you would have to copy *all* properties from the old mesh. i suppose we need some methods for the properties-object to be able to do this in the future ....

MeshLib/MeshEditing/MeshRevision.cpp
```diff
@@ -110,28 +139,52 @@ MeshLib::Mesh* MeshRevision::subdivideMesh(const std::string &new_mesh_name) con
...
-	for (auto elem = elements.begin(); elem != elements.end(); ++elem)
-	{
-		ElementErrorCode e((*elem)->validate());
-		if (e[ElementErrorFlag::NonCoplanar])
+	MeshLib::Properties new_properties;
+	boost::optional<PropertyVector<int> &> new_material_vec(
+		new_properties.createNewPropertyVector<int>(
+			"MaterialIDs", MeshItemType::Cell, 1
+		)
+	);
```
rinkk: same comment as above: first check if the vector actually exists before creating a new one.

MeshLib/MeshEditing/MeshRevision.cpp
```diff
@@ -110,28 +139,52 @@ MeshLib::Mesh* MeshRevision::subdivideMesh(const std::string &new_mesh_name) con
...
 			{
-				ERR("Error: Element %d has unknown element type.", std::distance(elements.begin(), elem));
+				ERR("Error: Element %d has unknown element type.", k);
 				this->cleanUp(new_nodes, new_elements);
 				return nullptr;
 			}
+			// transmit material values
+			if (!material_vec)
+				continue;
+			for (std::size_t j(0); j<n_new_elements; ++j)
```
rinkk: same comment as above: better use ```insert```

MeshLib/MeshEditing/MeshRevision.cpp
```diff
@@ -110,28 +139,52 @@ MeshLib::Mesh* MeshRevision::subdivideMesh(const std::string &new_mesh_name) con
...
+			std::size_t const n_new_elements(
+				subdivideElement(elem, new_nodes, new_elements));
+			if (n_new_elements == 0)
 			{
-				ERR("Error: Element %d has unknown element type.", std::distance(elements.begin(), elem));
+				ERR("Error: Element %d has unknown element type.", k);
 				this->cleanUp(new_nodes, new_elements);
 				return nullptr;
 			}
+			// transmit material values
```
rinkk: same comment as above: better use "copy"

MeshLib/MeshEditing/MeshRevision.cpp
```diff
@@ -67,39 +67,68 @@ MeshLib::Mesh* MeshRevision::simplifyMesh(const std::string &new_mesh_name, doub
...
 					return nullptr;
 				}
+				if (!material_vec)
+					continue;
+				for (std::size_t j(0); j<n_new_elements; ++j)
+					new_material_vec->push_back((*material_vec)[k]);
+			} else {
+				new_elements.push_back(MeshLib::copyElement(elem, new_nodes));
+				// transmit material values
+				if (!material_vec)
```
rinkk: better write:
```if (material_vec)
  new_mat_vec->push_back(...)```
(i.e. you save one line, ```continue``` is not needed here

File: MeshLib/MeshEditing/MeshRevision.cpp [hunk 8]
```diff
@@ -110,28 +139,52 @@ MeshLib::Mesh* MeshRevision::subdivideMesh(const std::string &new_mesh_name) con
...
 			}
+			// transmit material values
+			if (!material_vec)
+				continue;
+			for (std::size_t j(0); j<n_new_elements; ++j)
+				new_material_vec->push_back((*material_vec)[k]);
+		} else {
+			new_elements.push_back(MeshLib::copyElement(elem, new_nodes));
+			// transmit material values
+			if (!material_vec)
```
rinkk: same comment as above, remove ```continue```

MeshLib/MeshEditing/MeshRevision.cpp
```diff
@@ -253,24 +305,30 @@ void MeshRevision::reduceElement(MeshLib::Element const*const element,
 	/***************
 	 * TODO: modify neighbouring elements if one elements has been subdivided
 	 ***************/
-	if (element->getGeomType() == MeshElemType::TRIANGLE && min_elem_dim == 1)
+	if (element->getGeomType() == MeshElemType::TRIANGLE && min_elem_dim == 1) {
```
rinkk: everywhere else in this class the ```{``` is on a new line ... could you please do this here, too, otherwise it'll be easily overlooked.

MeshLib/MeshEditing/MeshRevision.cpp
```diff
@@ -67,39 +67,68 @@ MeshLib::Mesh* MeshRevision::simplifyMesh(const std::string &new_mesh_name, doub
...
-	const std::vector<MeshLib::Element*> &elements(this->_mesh.getElements());
-	for (auto elem = elements.begin(); elem != elements.end(); ++elem)
-	{
-		unsigned n_unique_nodes(this->getNUniqueNodes(*elem));
-		if (n_unique_nodes == (*elem)->getNBaseNodes() && (*elem)->getDimension() >= min_elem_dim)
+	MeshLib::Properties new_properties;
+	boost::optional<PropertyVector<int> &> new_material_vec(
+		new_properties.createNewPropertyVector<int>(
+			"MaterialIDs", MeshItemType::Cell, 1
+		)
```
rinkk: just a formatting thing:
if you write
```c++
boost::optional<PropertyVector<int> &> new_material_vec(
		new_properties.createNewPropertyVector<int>("MaterialIDs", MeshItemType::Cell, 1)
```
it is still less than 100 chars per line (which is our limit) but uses only 2 lines for the command instead of 4.
I don't know if it's just me, but i always get very confused if i see things that should/could be on one line subdivided into multiple lines ... and the more lines the more confusing ....
if you don't mind too much it would be really helpful if you could use the 100 char limit, there are multiple instances in this class alone where you would save a bunch of lines and (in my opinion) would make the source more easily comprehendable.

MeshLib/MeshEditing/MeshRevision.h
```diff
@@ -85,10 +85,18 @@ class MeshRevision {
...
+	std::size_t subdivideElement(MeshLib::Element const*const element,
+		std::vector<MeshLib::Node*> const& nodes,
+		std::vector<MeshLib::Element*> & elements) const;
 
 	// Revises an element by removing collapsed nodes, using the nodes vector from the result mesh.
-	void reduceElement(MeshLib::Element const*const element,
+	std::size_t reduceElement(MeshLib::Element const*const element,
 	                   unsigned n_unique_nodes,
 	    			   const std::vector<MeshLib::Node*> &nodes,
 	    			   std::vector<MeshLib::Element*> &elements,
```
rinkk: it's not your fault but can you align the parameters here please?

MeshLib/MeshEditing/MeshRevision.cpp
TomFischer: >i think something is wrong here with the sequence of commands. you should first check if the old mesh had a "material id" property and only if it did should you create such a property for the new mesh.

I'll change this.

> additional comment: it might not be relevant now, but in theory you would have to copy all properties from the old mesh. i suppose we need some methods for the properties-object to be able to do this in the future ....

Yes, that is true. This PR is one of the first steps. After that there is still much to do.

MeshLib/MeshEditing/MeshRevision.h
TomFischer: I indented the parameters of all methods in this class (where necessary).

endJunction:
@rinkk If the comments above were answered, please merge.

PR #640, opened 2015-04-14
Project mesh to plane

rinkk:
A small method that projects a mesh to an arbitrary plane defined by point of origin + normal vector.

### DIFF
--- a/MeshLib/MeshEditing/projectMeshOntoPlane.h
+++ b/MeshLib/MeshEditing/projectMeshOntoPlane.h
@@ -43,11 +43,10 @@ MeshLib::Mesh* projectMeshOntoPlane(MeshLib::Mesh const& mesh, MathLib::Point3d 
 	new_nodes.reserve(n_nodes);
 	for (std::size_t i=0; i<n_nodes; ++i)
 	{
-		MathLib::Vector3 const o (nodes[i]->getCoords());
-		MathLib::Vector3 const v (o, plane_origin);
-		double const dist (MathLib::scalarProduct(v, normal));
-		MathLib::Point3d result (o - (normal * -1 * dist));
-		new_nodes.push_back(new MeshLib::Node(result.getCoords()));
+		MeshLib::Node const& node(*nodes[i]);
+		MathLib::Vector3 const v(plane_origin, node);
+		new_nodes.push_back(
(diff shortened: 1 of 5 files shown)


MeshLib/MeshEditing/projectMeshOntoPlane.h
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * Projects all nodes of a mesh onto a plane specified by a point of origin and a normal vector.
+ * Overlapping elements, collapsed nodes, and other issues are not handled by the method.
+ * The normal vector need not be normalized.
+ */
+MeshLib::Mesh* projectMeshOntoPlane(MeshLib::Mesh const& mesh, MathLib::Point3d const& plane_origin, MathLib::Vector3 const& plane_normal)
+{
+	std::size_t const n_nodes (mesh.getNNodes());
+	std::vector<MeshLib::Node*> const& nodes (mesh.getNodes());
+	MathLib::Vector3 normal (plane_normal);
```
TomFischer: The copy can be omitted if an already normalized plane normal is given as argument.

MeshLib/MeshEditing/projectMeshOntoPlane.h
```diff
@@ -0,0 +1,58 @@
...
+	normal.normalize();
+	std::vector<MeshLib::Node*> new_nodes;
+	new_nodes.reserve(n_nodes);
+	for (std::size_t i=0; i<n_nodes; ++i)
+	{
+		MathLib::Vector3 const o (nodes[i]->getCoords());
+		MathLib::Vector3 const v (o, plane_origin);
+		double const dist (MathLib::scalarProduct(v, normal));
+		MathLib::Point3d result (o - (normal * -1 * dist));
```
TomFischer: This formula seems a l little bit strange to me, especially the two minus signs. I think the coordinates of the resulting point should be calculated as follow:
```
MathLib::Vector3 const node(nodes[i]->getCoords());
MathLib::Vector3 const v(plane_orgin, node); // switch the order of arguments compared to original alg.
MathLib::Point3d result(node - scalarProduct(v,plane_normal)*normal);
```

TomFischer: Did send you a PR.

PR #643, opened 2015-04-15
[Just for discussion] Classes for material property

wenqing:
Just a proposal for discussion.
In the implementation, material parameters are passed to PDEs by a container
```
template<typename T_PARAMETER_TYPE, typename T_MAT_MODEL, typename T_MATRIX> class TensorParameter  : public ParameterBase<T_PARAMETER_TYPE>
```
which allows  flexibility for different material model and computing variable dependent parameters via variadic members.  

The groundwater problem is used for test.


MaterialLib/Fluid/Viscosity/PressureDependentViscosity.h
```diff
@@ -0,0 +1,64 @@
...
+#include <string>
+#include <algorithm>    // std::max
+
+#include "ViscosityType.h"
+
+namespace MaterialLib
+{
+
+/// Pressure dependent viscosity model,
+class PressureDependentViscosity
```
endJunction: I like the idea in general. In ogs5 we have got several pressure dependent viscosity models numbered from 0 to 4 or something. It would be nice if we could avoid five classes having different ids.
Is there a generalization to this?

There are also models which are dependent on p only, on T only, on p and T, on p and T and t etc. So how to avoid a possible model name "PressureTemperatureTimeDependentViscosity7a"?

wenqing: I think the most of models are named already. My idea is to create a class for each mode, which may be function of p, T or other variables,  and  name the class after the name of the model.

endJunction:
No activity for more than 30 days.

PR #644, opened 2015-04-15
MeshIO using new Properties and PropertyVector.

TomFischer:
Reading material ids of elements from the old ogs mesh file format (*.msh) into the new data structures.

### DIFF
--- a/FileIO/Legacy/MeshIO.cpp
+++ b/FileIO/Legacy/MeshIO.cpp
@@ -163,9 +163,8 @@ MeshLib::Element* MeshIO::readElement(std::istream& in,
 	MeshElemType elem_type (MeshElemType::INVALID);
 
 	do {
-		in >> elem_type_str;
-		if (in.fail())
-			return NULL;
+		if (!(in >> elem_type_str))
+			return nullptr;
 		elem_type = String2MeshElemType(elem_type_str);
 	} while (elem_type == MeshElemType::INVALID);
 
@@ -176,7 +175,8 @@ MeshLib::Element* MeshIO::readElement(std::istream& in,
 	{
 	case MeshElemType::LINE: {
 		for (int i = 0; i < 2; ++i)
-			in >> idx[i];
+			if (!(in >> idx[i]))
+				return nullptr;
 		// edge_nodes array will be deleted from Line object
 		MeshLib::Node** edge_nodes = new MeshLib::Node*[2];
 		for (unsigned k(0); k < 2; ++k)
@@ -186,7 +186,8 @@ MeshLib::Element* MeshIO::readElement(std::istream& in,
 	}
 	case MeshElemType::TRIANGLE: {
 		for (int i = 0; i < 3; ++i)
-			in >> idx[i];
+			if (!(in >> idx[i]))
+				return nullptr;
 		MeshLib::Node** tri_nodes = new MeshLib::Node*[3];
 		for (unsigned k(0); k < 3; ++k)
 			tri_nodes[k] = nodes[idx[k]];
@@ -195,7 +196,8 @@ MeshLib::Element* MeshIO::readElement(std::istream& in,
 	}
 	case MeshElemType::QUAD: {
 		for (int i = 0; i < 4; ++i)
-			in >> idx[i];
+			if (!(in >> idx[i]))
+				return nullptr;
 		MeshLib::Node** quad_nodes = new MeshLib::Node*[4];
 		for (unsigned k(0); k < 4; ++k)
 			quad_nodes[k] = nodes[idx[k]];
@@ -204,7 +206,8 @@ MeshLib::Element* MeshIO::readElement(std::istream& in,
 	}
 	case MeshElemType::TETRAHEDRON: {
 		for (int i = 0; i < 4; ++i)
-			in >> idx[i];
+			if (!(in >> idx[i]))
+				return nullptr;
 		MeshLib::Node** tet_nodes = new MeshLib::Node*[4];
 		for (unsigned k(0); k < 4; ++k)
 			tet_nodes[k] = nodes[idx[k]];
@@ -213,7 +216,8 @@ MeshLib::Element* MeshIO::readElement(std::istream& in,
 	}
 	case MeshElemType::HEXAHEDRON: {
 		for (int i = 0; i < 8; ++i)
-			in >> idx[i];
+			if (!(in >> idx[i]))
+				return nullptr;
 		MeshLib::Node** hex_nodes = new MeshLib::Node*[8];
 		for (unsigned k(0); k < 8; ++k)
 			hex_nodes[k] = nodes[idx[k]];
@@ -222,7 +226,8 @@ MeshLib::Element* MeshIO::readElement(std::istream& in,
 	}
 	case MeshElemType::PYRAMID: {
 		for (int i = 0; i < 5; ++i)
-			in >> idx[i];
+			if (!(in >> idx[i]))
+				return nullptr;
 		MeshLib::Node** pyramid_nodes = new MeshLib::Node*[5];
 		for (unsigned k(0); k < 5; ++k)
 			pyramid_nodes[k] = nodes[idx[k]];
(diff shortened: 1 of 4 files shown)


FileIO/Legacy/MeshIO.cpp
```diff
@@ -119,6 +123,17 @@ MeshLib::Mesh* MeshIO::loadMeshFromFile(const std::string& file_name)
 		MeshLib::Mesh* mesh (new MeshLib::Mesh(BaseLib::extractBaseNameWithoutExtension(
 		                                               file_name), nodes, elements));
 
+		boost::optional<MeshLib::PropertyVector<int> &> opt_material_ids(
+			mesh->getProperties().createNewPropertyVector<int>(
+				"MaterialIDs", MeshLib::MeshItemType::Cell, 1)
+		);
+		if (!opt_material_ids) {
+			WARN("No material ids read.");
```
rinkk: is this possible? and if so, is it still a valid mesh file?

FileIO/Legacy/MeshIO.cpp
```diff
@@ -119,6 +123,17 @@ MeshLib::Mesh* MeshIO::loadMeshFromFile(const std::string& file_name)
...
+		} else {
+			MeshLib::PropertyVector<int> & material_ids(opt_material_ids.get());
+			material_ids.resize(materials.size());
+			std::copy(materials.cbegin(), materials.cend(), material_ids.begin());
```
rinkk: stupid question: since ```material_ids``` is empty at this point, wouldn't it be faster to append the mat-vector, e.g.
```material_ids.insert(material_ids.end(), materials.begin(), materials.end());``` ?

File: FileIO/Legacy/MeshIO.cpp [hunk 3]
```diff
@@ -133,18 +148,22 @@ MeshLib::Mesh* MeshIO::loadMeshFromFile(const std::string& file_name)
 	}
 }
 
-MeshLib::Element* MeshIO::readElement(const std::string& line,
-                                      const std::vector<MeshLib::Node*> &nodes)
+std::size_t MeshIO::readMaterialID(std::istream & in) const
+{
+	unsigned index, patch_index;
```
rinkk: why is the variable called ```patch_index``` and not ```material``` or so?

FileIO/Legacy/MeshIO.cpp
```diff
@@ -133,18 +148,22 @@ MeshLib::Mesh* MeshIO::loadMeshFromFile(const std::string& file_name)
...
 	MeshElemType elem_type (MeshElemType::INVALID);
-	unsigned index, patch_index;
-	ss >> index >> patch_index;
 
 	do {
-		ss >> elem_type_str;
-		if (ss.fail())
+		in >> elem_type_str;
+		if (in.fail())
 			return NULL;
```
rinkk: nullptr

FileIO/Legacy/MeshIO.cpp
```diff
@@ -133,18 +148,22 @@ MeshLib::Mesh* MeshIO::loadMeshFromFile(const std::string& file_name)
...
 	std::string elem_type_str("");
 	MeshElemType elem_type (MeshElemType::INVALID);
-	unsigned index, patch_index;
-	ss >> index >> patch_index;
 
 	do {
-		ss >> elem_type_str;
-		if (ss.fail())
+		in >> elem_type_str;
+		if (in.fail())
```
rinkk: why do you need to test here for a ```fail``` but not in the ```readMaterialID()```-method?

FileIO/Legacy/MeshIO.cpp
```diff
@@ -133,18 +148,22 @@ MeshLib::Mesh* MeshIO::loadMeshFromFile(const std::string& file_name)
...
 			return NULL;
 		elem_type = String2MeshElemType(elem_type_str);
 	} while (elem_type == MeshElemType::INVALID);
```
rinkk: sorry for the many questions but why do we need a ```do ... while``` here? (i see it is old code and presumably there is a reason for this but i cannot think of any ...)

FileIO/Legacy/MeshIO.cpp
```diff
@@ -156,66 +175,66 @@ MeshLib::Element* MeshIO::readElement(const std::string& line,
 	{
 	case MeshElemType::LINE: {
 		for (int i = 0; i < 2; ++i)
-			ss >> idx[i];
+			in >> idx[i];
```
rinkk: for this and all the other element types: what happens if you run out of values to read? (e.g. you need to read two node ids for a line but what happens if the stream ends after the first value?)

FileIO/Legacy/MeshIO.h
```diff
@@ -57,7 +57,8 @@ class MeshIO : public Writer
 
 private:
 	void writeElements(std::vector<MeshLib::Element*> const& ele_vec, std::ostream &out) const;
-	MeshLib::Element* readElement(const std::string& line, const std::vector<MeshLib::Node*> &nodes);
+	std::size_t readMaterialID(std::istream & in) const;
+	MeshLib::Element* readElement(std::istream& line, const std::vector<MeshLib::Node*> &nodes) const;
```
rinkk: documentation for both methods is missing ... the name is pretty obvious so i don't know if something is necessary or not ...

rinkk:
Two general things:
1. is it necessary to adjust the write method, too?
2. to have a robust read method, i think it would be best to do something like this:
  * read line
  * subdivide line into list of values
  * read 3rd value of line (element type id) and evaluate if number of read values is equal to numbers that need to be read (e.g. 7 values for a quad, etc.)
  * build elements based on these value.

this would require a lot of changes to the current read method and i really don't know if it is worth the effort for a legacy interface. i am just suggesting that the current implementation might crash in a number of cases. but it's also okay if we decide not to care about that ....

rinkk:
Another thing: we currently don't have tests for this. I am not sure if we need some (legacy interface and all that). if you feel like it you could write some ...

FileIO/Legacy/MeshIO.cpp
TomFischer: It should not be possible. But ...

FileIO/Legacy/MeshIO.cpp
TomFischer: Afaik GMSH uses patch_index as material id. But you are right: for readability the variable should be renamed.

TomFischer:
In my opinion this is a legacy file format. I do not want to put much more effort than using the new data structures for properties, i.e. I do not want write test nor I want to improve the robustness.

With the new commits I hope I addressed all your other comments.

FileIO/Legacy/MeshIO.cpp
```diff
@@ -103,7 +105,9 @@ MeshLib::Mesh* MeshIO::loadMeshFromFile(const std::string& file_name)
 				for (unsigned i = 0; i < nElements; ++i)
 				{
 					getline(in, line_string);
-					elements.push_back(readElement(line_string, nodes));
+					std::stringstream ss(line_string);
+					materials.push_back(readMaterialID(ss));
+					elements.push_back(readElement(ss, nodes));
```
norihiro-w: no need to stop reading if `readElement()` returns nullptr?

TomFischer: I did not change the already implemented behaviour. What do you think? Opinions of other developers are also wellcome. If we decide to change the behaviour should we do this in a separate PR?

rinkk: as i mentioned in my comments there are multiple ways to break this read-method. however, in my opinion we an leave it as it is for the most part as it is "only" a legacy interface and ogs5-read methods are similarly buggy (i.e. it is unlikely that someone comes up with a msh-file that will read in ogs5 but not in ogs6).

that said, in this particular case i would also *not* insert a nullptr but instead either abort reading the mesh or simply skip the element completely. in both cases we can guarantee consistent behaviour afterwards, whereas a nullptr within the element-vector will result in a crash later on...

bilke:
:+1:

FileIO/Legacy/MeshIO.cpp
TomFischer: Changed as requested.

PR #645, opened 2015-04-16
Using new mesh interface for Properties in class ElementValueModification 

TomFischer:
This is another PR that changes the handling of materials. `Properties` and `PropertyVector` are used in `ElementValueModification`.

### DIFF
--- a/MeshLib/MeshEditing/ElementValueModification.cpp
+++ b/MeshLib/MeshEditing/ElementValueModification.cpp
@@ -21,33 +21,7 @@
 #include "MeshLib/Mesh.h"
 #include "MeshLib/Elements/Element.h"
 
-
 namespace MeshLib {
-
-std::vector<unsigned> ElementValueModification::getMeshValues(const MeshLib::Mesh &mesh)
-{
-	const std::size_t nElements (mesh.getNElements());
-	std::vector<unsigned> value_mapping;
-	for (unsigned i=0; i<nElements; ++i)
-	{
-		bool exists(false);
-		unsigned value (mesh.getElement(i)->getValue());
-		const unsigned nValues (value_mapping.size());
-		for (unsigned j=0; j<nValues; ++j)
-		{
-			if (value == value_mapping[j])
-			{
-				exists = true;
-				break;
-			}
-		}
-		if (!exists)
-			value_mapping.push_back(value);
-	}
-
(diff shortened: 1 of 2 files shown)


bilke:
:+1:

MeshLib/MeshEditing/ElementValueModification.h
```diff
@@ -37,13 +41,40 @@ class ElementValueModification
...
+	static std::vector<T> getSortedPropertyValues(
+		MeshLib::PropertyVector<T> const& property_vector)
+	{
+		std::vector<T> value_mapping;
+		const std::size_t n_property_values(property_vector.size());
+		for (std::size_t i=0; i<n_property_values; ++i) {
+			bool exists(false);
+			T const& value (property_vector[i]);
+			std::size_t const size(value_mapping.size());
+			for (unsigned j=0; j<size; ++j) {
```
norihiro-w: though i'm not sure, it might be faster if you erase duplicated afterwards like

```
std::sort(value_mapping.begin(), value_mapping.end());
value_mapping.erase( std::unique(value_mapping.begin(), value_mapping.end()), value_mapping.end() );
```

MeshLib/MeshEditing/ElementValueModification.cpp
```diff
@@ -21,88 +21,105 @@
...
+
+	if (!optional_property_value_vec) {
+		return 0;
+	}
+
+	MeshLib::PropertyVector<int> & property_value_vector(
+		optional_property_value_vec.get()
+	);
+	std::vector<int> value_mapping(
+		ElementValueModification::getSortedPropertyValues(property_value_vector)
```
norihiro-w: you can remove ElementValueModification::

MeshLib/MeshEditing/ElementValueModification.h
TomFischer: While the number of property values is usually quite big, I expect a small number of different values here. The above implementation takes O(n) x O(m), where n is the total number of properties and m is the number of different property values.

If my understanding of the suggestion is correct 
(1) a copy of the values values is needed (O(n))
(2) sorting (O(n log n))
(3) erasing duplicate values (O(n))

If the number of different property values is in the order of O(log n) the implementations should have almost the same run time. So all depends on the concrete case.

PR #651, opened 2015-04-17
Prism node ordering

rinkk:
due to using InsituLib now (specifically, direct use of VTK methods for reading and writing meshes) we established in inconsistency for node ordering in prism-elements. this PR fixes that such that meshes are now again displayed correctly in OGS as well as in ParaView.

unfortunately, this also means that old ogs-meshes (created before including insitu-lib) need to have the prism nodes reordered. i adjusted the existing node-reordering tool to provide an easy way to do that (use parameter ```-m 2```).

note: the old meshes are not *per se* broken, it's just that the face-normals of prism elements point in the wrong direction. if this doesn't concern you, you need not re-order nodes.


bilke:
:+1:

PR #652, opened 2015-04-21
Using new data structures of Mesh to reset property values.

TomFischer:
As titled. Also added/updated [documentation](http://docs.opengeosys.org/docs/tools/meshing/set-properties-in-polygonal-region)


norihiro-w:
:+1:

PR #653, opened 2015-04-21
Command line tool for mapping geometry to the surface of a mesh

TomFischer:
as titled.

I am also open for suggestions to improve the [documentation](http://docs.opengeosys.org/docs/tools/meshing/map-geometric-object-to-the-surface-of-a-mesh) of the tool.

### DIFF
--- a/Applications/Utils/MeshEdit/CMakeLists.txt
+++ b/Applications/Utils/MeshEdit/CMakeLists.txt
@@ -26,6 +26,20 @@ if(QT4_FOUND)
 
 	set_target_properties(moveMeshNodes PROPERTIES FOLDER Utilities)
 
+	add_executable(MapGeometryToMeshSurface
+		MapGeometryToMeshSurface.cpp )
+	target_link_libraries(MapGeometryToMeshSurface
+		MeshLib
+		MeshGeoToolsLib
+		FileIO
+		InSituLib
+		${CATALYST_LIBRARIES}
+		${QT_LIBRARIES}
+	)
+	ADD_CATALYST_DEPENDENCY(MapGeometryToMeshSurface)
+
+set_target_properties(MapGeometryToMeshSurface
+	PROPERTIES FOLDER Utilities)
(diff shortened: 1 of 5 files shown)


Applications/Utils/MeshEdit/MapGeometryToMeshSurface.cpp
```diff
@@ -0,0 +1,104 @@
...
+		' ', "0.1");
+	TCLAP::ValueArg<std::string> mesh_in("m", "mesh-file",
+		"the name of the file containing the mesh", true,
+		"", "file name");
+	cmd.add(mesh_in);
+	TCLAP::ValueArg<std::string> input_geometry_fname("i", "input-geometry",
+		"the name of the file containing the input geometry", true,
+		"", "file name");
+	cmd.add(input_geometry_fname);
+	TCLAP::ValueArg<bool> advanced_mapping("", "advanced-mapping",
```
norihiro-w: can you give a more specific algorithm name here?

norihiro-w:
:+1: except for a minor comment

Applications/Utils/MeshEdit/CMakeLists.txt
```diff
@@ -26,6 +26,20 @@ if(QT4_FOUND)
...
 
+	add_executable(MapGeometryToMeshSurface
+		MapGeometryToMeshSurface.cpp )
+	target_link_libraries(MapGeometryToMeshSurface
+		MeshLib
+		MeshGeoToolsLib
+		FileIO
+		InSituLib
+		${CATALYST_LIBRARIES}
+		${QT_LIBRARIES}
```
endJunction: Are the QT libs necessary?

TomFischer: The catalyst libraries are necessary for the MeshLib. Only the part of Qt used for vtk is necessary. So I removed `${QT_LIBRARIES}`.

PR #595, opened 2015-02-09
[WIP] supporting lower dimension ele

norihiro-w:
in progress

PR #655, opened 2015-04-23
Add ElementCoordinatesMappingLocal class

norihiro-w:
This PR is  a part of #595 and adds mapping of lower dimensional elements (e.g. line in 2D, tri in 3D) between global and local coordinates. The mapping functionality will be used during calculation of shape functions and mapping of tensor material properties like permeability.

### DIFF
--- a/GeoLib/AnalyticalGeometry-impl.h
+++ b/GeoLib/AnalyticalGeometry-impl.h
@@ -78,29 +78,5 @@ void compute3DRotationMatrixToX(MathLib::Vector3  const& v,
     }
 }
 
-template <class T_MATRIX>
-void computeRotationMatrixToXY2(const std::vector<MathLib::Point3d*> &vec_pt,
-        T_MATRIX & rot_mat)
-{
-    assert(vec_pt.size()>2);
-    // x"_vec
-    MathLib::Vector3 xx((*vec_pt[0]), (*vec_pt[1]));
-    xx.normalize();
-    // a vector on the plane
-    MathLib::Vector3 yy((*vec_pt[1]), (*vec_pt[2]));
-    // z"_vec
-    MathLib::Vector3 zz(MathLib::crossProduct(xx, yy));
-    zz.normalize();
-    // y"_vec
-    yy = MathLib::crossProduct(zz, xx);
-    yy.normalize();
-
-    for (size_t i=0; i<3; ++i) {
-        rot_mat(0, i) = xx[i];
-        rot_mat(1, i) = yy[i];
-        rot_mat(2, i) = zz[i];
-    }
-}
-
 } // end namespace GeoLib
 
--- a/GeoLib/AnalyticalGeometry-impl.h
+++ b/GeoLib/AnalyticalGeometry-impl.h
@@ -1,0 +1,106 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2015, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+namespace GeoLib
+{
+
+template <class T_POINT>
+void getNewellPlane (const std::vector<T_POINT*>& pnts,
+                     MathLib::Vector3 &plane_normal,
+                     double& d)
+{
+    d = 0;
+    MathLib::Vector3 centroid;
+    std::size_t n_pnts(pnts.size());
+    for (size_t i(n_pnts - 1), j(0); j < n_pnts; i = j, j++) {
+        plane_normal[0] += ((*(pnts[i]))[1] - (*(pnts[j]))[1])
+                           * ((*(pnts[i]))[2] + (*(pnts[j]))[2]); // projection on yz
+        plane_normal[1] += ((*(pnts[i]))[2] - (*(pnts[j]))[2])
+                           * ((*(pnts[i]))[0] + (*(pnts[j]))[0]); // projection on xz
+        plane_normal[2] += ((*(pnts[i]))[0] - (*(pnts[j]))[0])
+                           * ((*(pnts[i]))[1] + (*(pnts[j]))[1]); // projection on xy
+
+        centroid += *(pnts[j]);
+    }
+
+    plane_normal.normalize();
+    d = MathLib::scalarProduct(centroid, plane_normal) / n_pnts;
+}
+
+template<class T_MATRIX>
+void compute2DRotationMatrixToX(MathLib::Vector3 const& v,
+        T_MATRIX & rot_mat)
+{
+    const double cos_theta = v[0];
+    const double sin_theta = v[1];
+    rot_mat(0,0) = rot_mat(1,1) = cos_theta;
+    rot_mat(0,1) = sin_theta;
+    rot_mat(1,0) = -sin_theta;
+}
+
+template <class T_MATRIX>
+void compute3DRotationMatrixToX(MathLib::Vector3  const& v,
+        T_MATRIX & rot_mat)
+{
+    // a vector on the plane
+    MathLib::Vector3 yy(0, 0, 0);
+    if (fabs(v[0]) > 0.0 && fabs(v[1]) + fabs(v[2]) < std::numeric_limits<double>::epsilon()) {
+        yy[2] = 1.0;
+    } else if (fabs(v[1]) > 0.0 && fabs(v[0]) + fabs(v[2]) < std::numeric_limits<double>::epsilon()) {
+        yy[0] = 1.0;
+    } else if (fabs(v[2]) > 0.0 && fabs(v[0]) + fabs(v[1]) < std::numeric_limits<double>::epsilon()) {
+        yy[1] = 1.0;
+    } else {
+        for (unsigned i = 0; i < 3; i++) {
+            if (fabs(v[i]) > 0.0) {
+                yy[i] = -v[i];
+                break;
+            }
+        }
+    }
+    // z"_vec
+    MathLib::Vector3 zz(MathLib::crossProduct(v, yy));
+    zz.normalize();
+    // y"_vec
+    yy = MathLib::crossProduct(zz, v);
+    yy.normalize();
+
+    for (unsigned i=0; i<3; ++i) {
+        rot_mat(0, i) = v[i];
+        rot_mat(1, i) = yy[i];
+        rot_mat(2, i) = zz[i];
+    }
+}
+
+template <class T_MATRIX>
+void computeRotationMatrixToXY2(const std::vector<MathLib::Point3d*> &vec_pt,
+        T_MATRIX & rot_mat)
+{
+    assert(vec_pt.size()>2);
+    // x"_vec
+    MathLib::Vector3 xx((*vec_pt[0]), (*vec_pt[1]));
+    xx.normalize();
+    // a vector on the plane
+    MathLib::Vector3 yy((*vec_pt[1]), (*vec_pt[2]));
+    // z"_vec
+    MathLib::Vector3 zz(MathLib::crossProduct(xx, yy));
+    zz.normalize();
+    // y"_vec
+    yy = MathLib::crossProduct(zz, xx);
+    yy.normalize();
+
+    for (size_t i=0; i<3; ++i) {
+        rot_mat(0, i) = xx[i];
+        rot_mat(1, i) = yy[i];
+        rot_mat(2, i) = zz[i];
+    }
+}
+
+} // end namespace GeoLib
+
--- a/GeoLib/AnalyticalGeometry-impl.h
+++ b/GeoLib/AnalyticalGeometry-impl.h
@@ -42,6 +42,8 @@ void compute2DRotationMatrixToX(MathLib::Vector3 const& v,
     rot_mat(0,0) = rot_mat(1,1) = cos_theta;
     rot_mat(0,1) = sin_theta;
     rot_mat(1,0) = -sin_theta;
+    rot_mat(0,2) = rot_mat(1,2) = rot_mat(2,0) = rot_mat(2,1) = 0.0;
+    rot_mat(2,2) = 1.0;
 }
 
 template <class T_MATRIX>
--- a/GeoLib/AnalyticalGeometry-impl.h
+++ b/GeoLib/AnalyticalGeometry-impl.h
@@ -78,5 +78,63 @@ void compute3DRotationMatrixToX(MathLib::Vector3  const& v,
     }
 }
 
+template <class T_MATRIX>
+void computeRotationMatrixToXY(MathLib::Vector3 const& n,
+		T_MATRIX & rot_mat)
+{
+	// check if normal points already in the right direction
+	if (sqrt(n[0]*n[0]+n[1]*n[1]) == 0) {
+		rot_mat(0,0) = 1.0;
+		rot_mat(0,1) = 0.0;
+		rot_mat(0,2) = 0.0;
+		rot_mat(1,0) = 0.0;
+		rot_mat(1,1) = 1.0;
+		rot_mat(1,2) = 0.0;
+		rot_mat(2,0) = 0.0;
+		rot_mat(2,1) = 0.0;
+		rot_mat(2,2) = 1.0;
+		return;
+	}
+
+	// sqrt (n_1^2 + n_3^2)
+	double const h0(sqrt(n[0]*n[0]+n[2]*n[2]));
+
+	// In case the x and z components of the normal are both zero the rotation
+	// to the x-z-plane is not required, i.e. only the rotation in the z-axis is
+	// required. The angle is either pi/2 or 3/2*pi. Thus the components of
+	// rot_mat are as follows.
+	if (h0 < std::numeric_limits<double>::epsilon()) {
+		rot_mat(0,0) = 1.0;
+		rot_mat(0,1) = 0.0;
+		rot_mat(0,2) = 0.0;
+		rot_mat(1,0) = 0.0;
+		rot_mat(1,1) = 0.0;
+		if (n[1] > 0)
+			rot_mat(1,2) = -1.0;
+		else
+			rot_mat(1,2) = 1.0;
+		rot_mat(2,0) = 0.0;
+		if (n[1] > 0)
+			rot_mat(2,1) = 1.0;
+		else
+			rot_mat(2,1) = -1.0;
+		rot_mat(2,2) = 0.0;
+		return;
+	}
+
+	double h1(1 / n.getLength());
+
+	// general case: calculate entries of rotation matrix
+	rot_mat(0, 0) = n[2] / h0;
+	rot_mat(0, 1) = 0;
+	rot_mat(0, 2) = - n[0] / h0;
+	rot_mat(1, 0) = - n[1]*n[0]/h0 * h1;
+	rot_mat(1, 1) = h0 * h1;
+	rot_mat(1, 2) = - n[1]*n[2]/h0 * h1;
+	rot_mat(2, 0) = n[0] * h1;
+	rot_mat(2, 1) = n[1] * h1;
+	rot_mat(2, 2) = n[2] * h1;
+}
+
 } // end namespace GeoLib
 
--- a/GeoLib/AnalyticalGeometry-impl.h
+++ b/GeoLib/AnalyticalGeometry-impl.h
@@ -50,15 +50,15 @@ void compute3DRotationMatrixToX(MathLib::Vector3  const& v,
 {
     // a vector on the plane
     MathLib::Vector3 yy(0, 0, 0);
-    if (fabs(v[0]) > 0.0 && fabs(v[1]) + fabs(v[2]) < std::numeric_limits<double>::epsilon()) {
+    if (std::abs(v[0]) > 0.0 && std::abs(v[1]) + std::abs(v[2]) < std::numeric_limits<double>::epsilon()) {
         yy[2] = 1.0;
-    } else if (fabs(v[1]) > 0.0 && fabs(v[0]) + fabs(v[2]) < std::numeric_limits<double>::epsilon()) {
+    } else if (std::abs(v[1]) > 0.0 && std::abs(v[0]) + std::abs(v[2]) < std::numeric_limits<double>::epsilon()) {
         yy[0] = 1.0;
-    } else if (fabs(v[2]) > 0.0 && fabs(v[0]) + fabs(v[1]) < std::numeric_limits<double>::epsilon()) {
+    } else if (std::abs(v[2]) > 0.0 && std::abs(v[0]) + std::abs(v[1]) < std::numeric_limits<double>::epsilon()) {
         yy[1] = 1.0;
     } else {
         for (unsigned i = 0; i < 3; i++) {
-            if (fabs(v[i]) > 0.0) {
+            if (std::abs(v[i]) > 0.0) {
                 yy[i] = -v[i];
                 break;
             }
--- a/GeoLib/AnalyticalGeometry.h
+++ b/GeoLib/AnalyticalGeometry.h
@@ -55,7 +55,8 @@ Orientation getOrientation (const GeoLib::Point* p0,
 /**
  * compute a supporting plane (represented by plane_normal and the value d) for the polygon
  * Let \f$n\f$ be the plane normal and \f$d\f$ a parameter. Then for all points \f$p \in R^3\f$ of the plane
- * it holds \f$ n \cdot p + d = 0\f$
+ * it holds \f$ n \cdot p + d = 0\f$. The Newell algorithm is described in
+ * \cite Ericson:2004:RCD:1121584 .
  * @param pnts points of a closed polyline describing a polygon
  * @param plane_normal the normal of the plane the polygon is located in
  * @param d parameter from the plane equation
--- a/GeoLib/AnalyticalGeometry.h
+++ b/GeoLib/AnalyticalGeometry.h
@@ -90,15 +90,6 @@ void computeRotationMatrixToXY(MathLib::Vector3 const& plane_normal,
                                MathLib::DenseMatrix<double> & rot_mat);
 
 /**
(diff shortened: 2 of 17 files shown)


norihiro-w:
Remark: one of the tests is not working right now due to an issue in computeRotationMatrixToXY.(). Tom will fix it.

GeoLib/AnalyticalGeometry.h
```diff
@@ -60,11 +60,28 @@ Orientation getOrientation (const GeoLib::Point* p0,
  * @param plane_normal the normal of the plane the polygon is located in
  * @param d parameter from the plane equation
  */
-void getNewellPlane (const std::vector<GeoLib::Point*>& pnts,
+template <class T_POINT>
+void getNewellPlane (const std::vector<T_POINT*>& pnts,
```
endJunction: @TomFischer Who is "Newell"?
Sarcasm aside, from the documentation it is not clear what is modified and how.

TomFischer: You are right, we should rename the method to something like `getSupportingPlane`. The algorithm uses Newell's method, described in the book "Real-Time Collision Detection" from Christer Ericson. This fact is also missing in the documentation.
@norihiro-w Should I update the branch or should we make a separat PR?

norihiro-w: you can make a separate PR

MeshLib/ElementCoordinatesMappingLocal.cpp
```diff
@@ -0,0 +1,101 @@
...
+        if (global_dim == 2)
+            GeoLib::compute2DRotationMatrixToX(xx, matR);
+        else
+            GeoLib::compute3DRotationMatrixToX(xx, matR);
+        matR.transposeInPlace();
+    } else if (global_dim == 3 && e.getDimension() == 2) {
+        std::vector<MathLib::Point3d*> pnts;
+        for (auto& p : vec_pt)
+            pnts.push_back(const_cast<MeshLib::Node*>(&p));
+#if 0
```
wenqing: What's this?

norihiro-w: these are testing another algorithm to compute a rotation matrix (actually the one used in OGS5). Once Tom fixes GeoLib::computeRotationMatrixToXY(), I'll remove them.

wenqing: I see.

norihiro-w:
This PR should be merged after #659 by Tom. I will rebase my commits to it

norihiro-w:
rebased to #659

MeshLib/CoordinateSystem.h
```diff
@@ -0,0 +1,100 @@
...
+{
+
+class Element;
+
+/**
+ * \brief Coordinate system type
+ */
+struct CoordinateSystemType
+{
+    enum type
```
endJunction: Better to use `enum class` directly.

norihiro-w: i initially tried to use it but changed it to this way  because the code is shorter and simpler as I don't need to make cast (from enum class to int) each time whenever bit operation is used.

GeoLib/AnalyticalGeometry-impl.h
```diff
@@ -0,0 +1,106 @@
...
+    rot_mat(1,0) = -sin_theta;
+}
+
+template <class T_MATRIX>
+void compute3DRotationMatrixToX(MathLib::Vector3  const& v,
+        T_MATRIX & rot_mat)
+{
+    // a vector on the plane
+    MathLib::Vector3 yy(0, 0, 0);
+    if (fabs(v[0]) > 0.0 && fabs(v[1]) + fabs(v[2]) < std::numeric_limits<double>::epsilon()) {
```
TomFischer: It is also possible to use ```std::abs``` instead of ```fabs``` here and in line 55 as well as line 57 and 61.

GeoLib/AnalyticalGeometry-impl.h
```diff
@@ -0,0 +1,106 @@
...
+    yy.normalize();
+
+    for (unsigned i=0; i<3; ++i) {
+        rot_mat(0, i) = v[i];
+        rot_mat(1, i) = yy[i];
+        rot_mat(2, i) = zz[i];
+    }
+}
+
+template <class T_MATRIX>
```
TomFischer: I am unsure if this version of `computeRotationMatrixToXY` is still necessary?

MeshLib/ElementCoordinatesMappingLocal.cpp
```diff
@@ -0,0 +1,97 @@
...
+namespace MeshLib
+{
+
+ElementCoordinatesMappingLocal::ElementCoordinatesMappingLocal(
+    const Element& e,
+    const CoordinateSystem &global_coords)
+: _coords(global_coords)
+{
+    assert(e.getDimension() <= global_coords.getDimension());
+    for(size_t i = 0; i < e.getNNodes(); i++)
```
TomFischer: `size_t`-> `std::size_t`

MeshLib/ElementCoordinatesMappingLocal.cpp
```diff
@@ -0,0 +1,97 @@
...
+    Eigen::VectorXd x_new = Eigen::VectorXd::Zero(3);
+    for(std::size_t i = 0; i < ele.getNNodes(); i++)
+    {
+        for (std::size_t j=0; j<global_dim; j++)
+            dx[j] = vec_pt[i].getCoords()[j];
+
+        x_new.head(global_dim) = matR2local * dx;
+        local_pt[i] = MeshLib::Node(x_new.data());
+    }
+};
```
TomFischer: Semicolon is not needed.

MeshLib/ElementCoordinatesMappingLocal.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+void ElementCoordinatesMappingLocal::getRotationMatrixToGlobal(
+    const Element &e,
+    const CoordinateSystem &global_coords,
+    const std::vector<MeshLib::Node> &vec_pt,
+    RotationMatrix &matR) const
+{
+    const std::size_t global_dim = global_coords.getDimension();
+
+    // compute R in x=R*x' where x is original coordinates and x' is local coordinates
```
TomFischer: where x **are** original coordinates and x' **are** local coordinates

MeshLib/ElementCoordinatesMappingLocal.cpp
```diff
@@ -0,0 +1,97 @@
...
+    } else if (global_dim == 3 && e.getDimension() == 2) {
+        std::vector<MathLib::Point3d*> pnts;
+        for (auto& p : vec_pt)
+            pnts.push_back(const_cast<MeshLib::Node*>(&p));
+
+        // get plane normal
+        MathLib::Vector3 plane_normal;
+        double d;
+        GeoLib::getNewellPlane (pnts, plane_normal, d);
+        //std::cout << "pn=" << plane_normal << std::endl;
```
TomFischer: Should the debug output be kept? If yes, please add a reason.

MeshLib/ElementCoordinatesMappingLocal.cpp
```diff
@@ -0,0 +1,97 @@
...
+        // compute a rotation matrix to XY
+        MathLib::DenseMatrix<double> matToXY(3,3,0.0);
+        GeoLib::computeRotationMatrixToXY(plane_normal, matToXY);
```
TomFischer: Do you use Eigen matrices because you want to use `transposeInPlace()`? Is this the only feature you want to use from Eigen matrix implementation that is not implemented in our `DenseMatrix` class?

One possibility to avoid the copy of matrices in line 90-92 would be to make `computeRotationMatrixToXY` a template also accepting Eigen matrices.

Or we could implement `DenseMatrix<T>::transposeInPlace` and avoid using Eigen matrix class here.

MeshLib/ElementCoordinatesMappingLocal.cpp
```diff
@@ -0,0 +1,97 @@
...
+    const std::size_t global_dim = global_coords.getDimension();
+    Eigen::VectorXd dx = Eigen::VectorXd::Zero(global_dim);
+    Eigen::VectorXd x_new = Eigen::VectorXd::Zero(3);
+    for(std::size_t i = 0; i < ele.getNNodes(); i++)
+    {
+        for (std::size_t j=0; j<global_dim; j++)
+            dx[j] = vec_pt[i].getCoords()[j];
+
+        x_new.head(global_dim) = matR2local * dx;
+        local_pt[i] = MeshLib::Node(x_new.data());
```
TomFischer: Is it possible to overload the operator* for the Eigen matrix using our `MeshLib::Node` as an argument? The it would be possible to write
```local_pt[i] = matR2local * vec_pt[i];```
instead of lines 48 - 52.

As an alternative we could use our `DenseMatrix` as an rotation matrix. Here we could implement such an operator. Maybe we can have short discussion if it is worth and if you need support.

File: MeshLib/ElementCoordinatesMappingLocal.h [hunk 12]
```diff
@@ -0,0 +1,79 @@
...
+
+#include "MeshLib/Elements/Element.h"
+#include "MeshLib/CoordinateSystem.h"
+
+namespace MeshLib
+{
+
+#ifdef OGS_USE_EIGEN
+typedef Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> RotationMatrix;
+#endif
```
TomFischer: I am unsure, but I think the whole class depends on the availability of Eigen.

MeshLib/ElementCoordinatesMappingLocal.h
```diff
@@ -0,0 +1,79 @@
...
+    ElementCoordinatesMappingLocal(const Element &e, const CoordinateSystem &global_coord_system);
+	
+    /// Destructor
+    virtual ~ElementCoordinatesMappingLocal() {}
+
+    /// return the global coordinate system
+    const CoordinateSystem getGlobalCoordinateSystem() const { return _coords; }
+
+    /// return mapped coordinates of the node
+    const MeshLib::Node* getMappedCoordinates(size_t node_id) const
```
TomFischer: `size_t` -> `std::size_t`

Tests/MeshLib/TestCoordinatesMappingLocal.cpp
```diff
@@ -0,0 +1,270 @@
...
+        nodes[1] = new MeshLib::Node(0.0, 0.0,  1.0);
+        return new MeshLib::Line(nodes);
+    }
+
+    static MeshLib::Line* createXY()
+    {
+        // 45degree inclined
+        MeshLib::Node** nodes = new MeshLib::Node*[e_nnodes];
+        nodes[0] = new MeshLib::Node(0.0, 0.0, 0.0);
+        nodes[1] = new MeshLib::Node(2./sqrt(2), 2./sqrt(2), 0.0);
```
TomFischer: 2./sqrt(2) == sqrt(2) :smile:

Tests/MeshLib/TestCoordinatesMappingLocal.cpp
```diff
@@ -0,0 +1,270 @@
...
+    // Element information
+    typedef MeshLib::Quad ElementType;
+    static const unsigned dim = 2; //ElementType::dimension;
+    static const unsigned e_nnodes = ElementType::n_all_nodes;
+
+    // 2.5D case: inclined
+    static MeshLib::Quad* createXYZ()
+    {
+        MeshLib::Node** nodes = new MeshLib::Node*[e_nnodes];
+        // rotate 40 degree around x axis
```
TomFischer: 40 or 45 degree? The values looks like cos(pi/4) and sin (pi/4) and pi/4 is corresponding to 45 degree.

Tests/MeshLib/TestCoordinatesMappingLocal.cpp
```diff
@@ -0,0 +1,270 @@
...
+        nodes[2] = new MeshLib::Node(0.0, -1.0, -1.0);
+        nodes[3] = new MeshLib::Node(0.0,  1.0, -1.0);
+        return new MeshLib::Quad(nodes);
+    }
+
+};
+
+void debugOutput(MeshLib::Element *ele, MeshLib::ElementCoordinatesMappingLocal &mapping)
+{
+    std::cout.precision(12);
```
TomFischer: Why not logog? Do you plan to put this code between the preprocessor directives `#ifndef NDEBUG` and `#endif`?

Tests/MeshLib/TestCoordinatesMappingLocal.cpp
```diff
@@ -0,0 +1,270 @@
...
+    //debugOutput(ele, mapping);
+
+    // results when using GeoLib::ComputeRotationMatrixToXY()
+    double exp_R[3*3] = {  1, 0,  0,
+                           0, 0, -1,
+                           0, 1,  0}; //row major
+//    // results when using GeoLib::ComputeRotationMatrixToXY2()
+//    double exp_R[3*3] = { -1,  0,  0,
+//                           0,  0, -1,
+//                           0, -1,  0}; //row major
```
TomFischer: Is it possible to remove the commented lines?

Tests/MeshLib/TestCoordinatesMappingLocal.cpp
```diff
@@ -0,0 +1,270 @@
...
+    auto ele = TestQuad4::createYZ();
+    MeshLib::ElementCoordinatesMappingLocal mapping(*ele, MeshLib::CoordinateSystem(*ele));
+    auto matR(mapping.getRotationMatrixToGlobal());
+    //debugOutput(ele, mapping);
+
+    // results when using GeoLib::ComputeRotationMatrixToXY()
+    double exp_R[3*3] = { 0, 0, 1,
+                          0, 1, 0,
+                         -1, 0, 0}; //row major
+//    // results when using GeoLib::ComputeRotationMatrixToXY2()
```
TomFischer: see above

Tests/MeshLib/TestCoordinatesMappingLocal.cpp
```diff
@@ -0,0 +1,270 @@
...
+    //debugOutput(ele, mapping);
+
+    // results when using GeoLib::ComputeRotationMatrixToXY()
+    double exp_R[3*3] = {  1, 0, 0,
+                           0, 0.70710678118654757, -0.70710678118654757,
+                           0, 0.70710678118654757,  0.70710678118654757}; //row major
+//    // results when using GeoLib::ComputeRotationMatrixToXY2()
+//    double exp_R[3*3] = { -1, 0, 0,
+//                           0, -0.70710678118654757, -0.70710678118654757,
+//                           0, -0.70710678118654757,  0.70710678118654757}; //row major
```
TomFischer: see above

TomFischer:
@norihiro-w Please let me know if I should help you fixing some of the issues.

MeshLib/ElementCoordinatesMappingLocal.cpp
norihiro-w: Eigen is used here because the rotation matrix will be later multiplied with shape matrices which I assume Eigen is used as data type

Tests/MeshLib/TestCoordinatesMappingLocal.cpp
norihiro-w: yes but I intended to write it so since it's clear that the line has a length of 2.

Tests/MeshLib/TestCoordinatesMappingLocal.cpp
norihiro-w: right, it's 45degree

Tests/MeshLib/TestCoordinatesMappingLocal.cpp
norihiro-w: I prefer cout because logog add prefix in output. I'll add #ifdef guard

MeshLib/ElementCoordinatesMappingLocal.h
norihiro-w: changed it to unsigned :white_check_mark:

MeshLib/ElementCoordinatesMappingLocal.cpp
norihiro-w: deleted :white_check_mark:

MeshLib/ElementCoordinatesMappingLocal.cpp
norihiro-w: changed it to unsigned

MeshLib/ElementCoordinatesMappingLocal.cpp
norihiro-w: technically yes as we have the code, however I would not do it for the third-party library. Instead i shortened the code using Eigen::Map

MeshLib/ElementCoordinatesMappingLocal.cpp
```diff
@@ -0,0 +1,86 @@
...
+void ElementCoordinatesMappingLocal::rotateToLocal(
+    const Element &ele,
+    const CoordinateSystem &global_coords,
+    const RotationMatrix &matR2local,
+    std::vector<MeshLib::Node*> &vec_pt) const
+{
+    // rotate the point coordinates
+    for(unsigned i = 0; i < ele.getNNodes(); i++)
+    {
+        Eigen::Vector3d  x_new = matR2local * Eigen::Map<Eigen::Vector3d>(const_cast<double*>(vec_pt[i]->getCoords()));
```
TomFischer: Is there an obstacle to implement something like
```
MeshLib::Node operator*(EigenMatrix const& mat, MeshLib::Node const& node)
```
or
```
MathLib::Point3d operator*(EigenMatrix mat, MathLib::Point3d const& )?
```
I would prefer the second version, which is more general.

This would be 
(1) more readable for the user and
(2) save some lines of code and
(3) omit some copies.

norihiro-w: updated. please see ee35769

norihiro-w:
It appears to me that DenseMatrix is used in GeoLib and MeshLib, whereas Eigen is used in NumLib. Should I move also ElementCoordinatesMappingLocal class to NumLib since it uses Eigen?

Tests/MeshLib/TestCoordinatesMappingLocal.cpp
```diff
@@ -0,0 +1,262 @@
...
+        nodes[1] = new MeshLib::Node(0.0, -1.0,  1.0);
+        nodes[2] = new MeshLib::Node(0.0, -1.0, -1.0);
+        nodes[3] = new MeshLib::Node(0.0,  1.0, -1.0);
+        return new MeshLib::Quad(nodes);
+    }
+
+};
+
+#ifndef NDEBUG
+void debugOutput(MeshLib::Element *ele, MeshLib::ElementCoordinatesMappingLocal &mapping)
```
TomFischer: The function is not used and this fact gives a warning when compiling in the debug mode (tested it with clang).

MathLib/Point3d.h
```diff
@@ -16,7 +16,12 @@
 
 #include <limits>
 
+#ifdef OGS_USE_EIGEN
+#include <Eigen/Eigen>
+#endif
```
endJunction: @TomFischer Is eigen needed now?

MeshLib/ElementCoordinatesMappingLocal.cpp
```diff
@@ -0,0 +1,80 @@
...
+{
+	for (auto p : _point_vec) delete p;
+}
+
+void ElementCoordinatesMappingLocal::rotateToLocal(
+    const RotationMatrix &matR2local,
+    std::vector<MeshLib::Node*> &vec_pt) const
+{
+    for(unsigned i = 0; i < vec_pt.size(); i++)
+        vec_pt[i]->setCoords((matR2local* (*vec_pt[i])).getCoords());
```
endJunction: More style than performance. If I'm not wrong this could be
```c++
for(MeshLib::Node* node : vec_pt)
    node->setCoords((matR2local* (*node)).getCoords());
```
Also there is a mismatch between vector of Node (pointers) and its name vec_pt.

MeshLib/ElementCoordinatesMappingLocal.cpp
```diff
@@ -0,0 +1,80 @@
...
+}
+
+void ElementCoordinatesMappingLocal::getRotationMatrixToGlobal(
+    const Element &e,
+    const CoordinateSystem &global_coords,
+    const std::vector<MeshLib::Node*> &vec_pt,
```
endJunction: Same naming mismatch here. Maybe `std::vector<MathLib::Point3d>` is sufficient here; No `Node` features seem to be used.

MeshLib/ElementCoordinatesMappingLocal.h
```diff
@@ -0,0 +1,76 @@
...
+
+    /// get a rotation matrix to the global coordinates
+    /// it computes R in x=R*x' where x is original coordinates and x' is local coordinates
+    void getRotationMatrixToGlobal(
+            const Element &e, const CoordinateSystem &coordinate_system,
+            const std::vector<MeshLib::Node*> &vec_pt, RotationMatrix &matR2original) const;
+
+private:
+    const CoordinateSystem _coords;
+    std::vector<MeshLib::Node*> _point_vec;
```
endJunction: Again, `point_vec` is a vector of nodes...

MathLib/Point3d.h
TomFischer: I think Eigen is not needed here. I forgot to remove these lines. @norihiro-w : can you remove the lines and merge the changes to the corresponding commit. By the way, I think the commits  c7deda4 and  ee35769 can also be squashed together. If you want do this go ahead.

MeshLib/ElementCoordinatesMappingLocal.cpp
TomFischer: The original "point" data are the `MeshLib::Node`s from an element. At least pointers to `Point3d` are needed for polymorphism.

MathLib/Point3d.h
norihiro-w: deleted and squashed

MeshLib/ElementCoordinatesMappingLocal.cpp
norihiro-w: renamed to vec_nodes

MeshLib/ElementCoordinatesMappingLocal.h
```diff
@@ -0,0 +1,76 @@
...
+
+#include "MeshLib/Elements/Element.h"
+#include "MeshLib/CoordinateSystem.h"
+
+namespace MeshLib
+{
+
+#ifdef OGS_USE_EIGEN
+typedef Eigen::Matrix<double, 3u, 3u, Eigen::RowMajor> RotationMatrix;
+#endif
```
endJunction: If eigen is not found or is disabled, the type RotationMatrix is not defined causing a compilation error. We need a default value.

@TomFischer Which parts of the interface are missing in DenseMatrix for this code to compile?
I found only `RotationMatrix::Identity()` to be missing...

norihiro-w: added DenseMatrix as a default case

norihiro-w:
tests are failing on Windows MSVC while it is fine on GCC and Clang. as i dont have windows environment at home, i will check this after i am back in the office.

bilke:
Jenkins test this please!

norihiro-w:
got the following errors with MSVC. is there anyone having idea what to do?

```
Error	28	error C2280: 'ProcessLib::GroundwaterFlow::LocalAssemblerData<NumLib::ShapeLine2,NumLib::IntegrationGaussRegular<0x01>,GlobalMatrix_,GlobalVector_>::LocalAssemblerData(void)' : attempting to reference a deleted function	C:\Users\watanabe\Documents\GitHub\ogs\AssemblerLib\LocalDataInitializer.h	59	1	ogs
Error	35	error C2542: 'NumLibFemIsoTest<Type>::D' : class object has no constructor for initialization	C:\Users\watanabe\Documents\GitHub\ogs\Tests\NumLib\TestFe.cpp	104	1	testrunner
Error	36	error C2542: 'NumLibFemIsoTest<Type>::expectedM' : class object has no constructor for initialization	C:\Users\watanabe\Documents\GitHub\ogs\Tests\NumLib\TestFe.cpp	104	1	testrunner
Error	37	error C2542: 'NumLibFemIsoTest<Type>::expectedK' : class object has no constructor for initialization	C:\Users\watanabe\Documents\GitHub\ogs\Tests\NumLib\TestFe.cpp	104	1	testrunner
Error	38	error C1903: unable to recover from previous error(s); stopping compilation	C:\Users\watanabe\Documents\GitHub\ogs\Tests\NumLib\TestFe.cpp	109	1	testrunner
```

endJunction:
I compiled the code with clang in RelWithDebInfo, which run through OK. But running tests ended in a SEGV for the:
```
[----------] 3 tests from NumLibFemIsoTest/5, where TypeParam = (anonymous namespace)::TestCase<FeTestData::TestFeLINE2, EigenFixedShapeMatrixPolicy>
[ RUN      ] NumLibFemIsoTest/5.CheckMassMatrix
```
test.
Running it in gdb does not always result in a SEGV: output and the backtrace [here](https://gist.github.com/endJunction/6473dbb722eda3f3c509)

Valgrind's memcheck does not show any invalid reads/writes until a SEGV. (see the above gist for details)

norihiro-w:
update: the test failing on Jenkins cannot be reproduced on my windows (win7, MSVC2013 with update 4, 64bit). could be there is something unitialized or machine dependent bug.

norihiro-w:
rebased to HEAD

norihiro-w:
@bilke Eigen version used in OGS6/Win-PRs seems to be bit old (3.1.4). Can you please replace it with the latest (3.2.4)?

bilke:
I updated eigen to 3.2.4.

Jenkins test this please!

bilke:
test this please

norihiro-w:
@bilke thanks!

norihiro-w:
the problem was solved!

PR #659, opened 2015-04-24
Fix bug in GeoLib::computeRotationMatrixToXY

TomFischer:
As titled. Also added a simple test.

### DIFF
--- a/Tests/GeoLib/TestComputeRotationMatrix.cpp
+++ b/Tests/GeoLib/TestComputeRotationMatrix.cpp
@@ -1,0 +1,84 @@
+/**
+ * @file TestComputeRotationMatrix.cpp
+ * @date 2015-04-23
+ *
+ * \copyright
+ * Copyright (c) 2012-2015, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ */
+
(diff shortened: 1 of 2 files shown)


Tests/GeoLib/TestComputeRotationMatrix.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include "gtest/gtest.h"
+#include "AnalyticalGeometry.h"
+
+TEST(GeoLib, ComputeRotationMatrixToXY)
+{
+	MathLib::Vector3 const n(0.0, -1.0, 0.0);
+	MathLib::DenseMatrix<double> rot_mat(3,3,0.0);
+
+	GeoLib::computeRotationMatrixToXY(n, rot_mat);
+	EXPECT_EQ(1.0, rot_mat(0,0));
```
norihiro-w: Can you take into account  a plane normal vector even for a XZ plane? The expected rotation matrix here is actually for the case of a normal vector (0, 1, 0). If a normal vector is (0, -1, 0), I would expect a matrix like

```
1 0 0
0 0 1
0 -1 0
```

TomFischer: Added test for this case and handle it in `computeRotationMatrixToXY`. Changed the commits, respectively.

norihiro-w: thanks, it worked!

norihiro-w:
:+1:

PR #660, opened 2015-04-24
Fix compilation error w/o explicit default initialization of a const object.

endJunction:
The compilation error:
```error: default initialization of an object of const type 'const std::vector<MeshLib::Element *>' without a user-provided default constructor```
is produced by clang-3.6.0, which is correct for the current standard, but will be corrected later as described in
http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_active.html#253

In gcc-5.1.0 this is only a warning.

See also [this SO thread](http://stackoverflow.com/questions/26077807/why-does-gcc-allow-a-const-object-without-a-user-declared-default-constructor-bu).


TomFischer:
:+1:

norihiro-w:
:+1:

Issue #662, opened 2015-04-27
[DE] crash when reading .gli


waltherm:
the following .gli setup will cause a crash without error / warning for the current DE.

```
#POINTS
0 0 0 0 $MD 10 $NAME P0
1 1000 0 0 $MD 10 $NAME P1
2 1000 0 100 $MD 10 
3 0 0 100 $MD 10 
4 400 0 0 $MD 10 
5 400 0 100 $MD 10 
6 300 0 0 $MD 10 
7 300 0 100 $MD 10 
#POLYLINE
 $NAME
  IN
 $POINTS
  0
  3
#POLYLINE
 $NAME
  OUT
 $POINTS
  1
  2
#POLYLINE
 $NAME
  S1
 $POINTS
  0
  6
  4
  1
  2
  5
  7
  3
  0
#POLYLINE
 $NAME
  first
 $POINTS
  7
  6
#POLYLINE
 $NAME
  second
 $POINTS
  5
  4
#SURFACE
 $NAME
  S1_S
 $POLYLINES
  S1
 $TYPE
  0
#STOP
#LIS 1 4
#LIS 1 5
```

TomFischer:
Fixed in PR #666.

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/662).

PR #666, opened 2015-04-28
GeoLib fix line segment intersection algorithm

TomFischer:
@waltherm discovered a bug while reading a geometry. See issue #662. This bug is fixed with this PR. The previous failing cases are included as new tests.


bilke:
This PR is evil and shall not be merged! :imp:

endJunction:
Using Schläfli notation the answer is {5,2}.

PR #668, opened 2015-05-04
Cleanup

endJunction:
Randomly placed changes. See commit messages for descriptions.

### DIFF
--- a/ProcessLib/NeumannBc.h
+++ b/ProcessLib/NeumannBc.h
@@ -111,7 +111,7 @@ public:
         LocalAssemblerBuilder local_asm_builder(
             initializer, *_local_to_global_index_map);
 
-        auto elementValueLookup = [this](MeshLib::Element const& e)
+        auto elementValueLookup = [this](MeshLib::Element const&)
         {
             return _function();
         };
--- a/ProcessLib/NeumannBc.h
+++ b/ProcessLib/NeumannBc.h
@@ -77,6 +77,11 @@ public:
 
     ~NeumannBc()
     {
+        for (auto p : _all_mesh_subsets)
+            delete p;
+
+        delete _mesh_subset_all_nodes;
+
         for (auto e : _elements)
             delete e;
 
--- a/ProcessLib/ProcessVariable.h
+++ b/ProcessLib/ProcessVariable.h
@@ -52,7 +52,7 @@ public:
     template <typename OutputIterator, typename GlobalSetup, typename ...Args>
     void createNeumannBcs(OutputIterator bcs,
         MeshGeoToolsLib::BoundaryElementsSearcher& searcher,
-        GlobalSetup const& global_setup,
+        GlobalSetup const&,
         Args&&... args)
     {
         for (NeumannBcConfig* config : _neumann_bc_configs)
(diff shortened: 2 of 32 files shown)


TomFischer:
:fast_forward:

MeshLib/MeshSubset.h
```diff
@@ -29,22 +29,31 @@ class MeshSubset
...
 
-    ~MeshSubset() {}
+    ~MeshSubset()
+    {
+        if (_delete_ptr)
+        {
+            delete _nodes;
+            delete _eles;
+        }
+    }
```
rinkk: I think the whole ```delete_ptr``` solution is very unintuitive. At the very least it should be well-documented in constructor/destructor why this variable is there and its effect is

ProcessLib/NeumannBc.h
```diff
@@ -111,7 +116,7 @@ class NeumannBc
         LocalAssemblerBuilder local_asm_builder(
             initializer, *_local_to_global_index_map);
 
-        auto elementValueLookup = [this](MeshLib::Element const& e)
+        auto elementValueLookup = [this](MeshLib::Element const&)
```
rinkk: why do you need the element if it isn't used?

ProcessLib/ProcessVariable.h
```diff
@@ -52,7 +52,7 @@ class ProcessVariable
     template <typename OutputIterator, typename GlobalSetup, typename ...Args>
     void createNeumannBcs(OutputIterator bcs,
         MeshGeoToolsLib::BoundaryElementsSearcher& searcher,
-        GlobalSetup const& global_setup,
+        GlobalSetup const&,
```
rinkk: why do you need the GlobalSetup if it isn't used?

scripts/cmake/CompilerSetup.cmake
```diff
@@ -33,7 +33,7 @@ if(COMPILER_IS_GCC)
 	endif()
 	if(NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
 		message(STATUS "Set GCC release flags")
-		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O3 -mtune=native -msse4.2 -DNDEBUG")
+		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O3 -mtune=native -DNDEBUG")
```
TomFischer: Should be `-march`.

ProcessLib/NeumannBc.h
endJunction: General interface.
When we start to use material ids the elementValueLookup will be dependent on the element itself. For the constant value for any of elements the argument is not used.

ProcessLib/ProcessVariable.h
endJunction: This is used to infer the `GlobalSetup` type using a mechanism called "argument dependent lookup".
I can not avoid this because I'd need to specify the unknown `Args` types in the process in the [function call](https://github.com/ufz/ogs/blob/master/ProcessLib/GroundwaterFlowProcess.h#L150-L156).

endJunction:
I removed the MeshSubset changes and pulled them into https://github.com/ufz/ogs/pull/671.
@TomFischer Please check the last six commits; these are new.

TomFischer:
:+1:

endJunction:
Just a rebase.

PR #669, opened 2015-05-06
[Fix] Added checks for VTU reader

rinkk:
* checks if a file of the given name actually exists
* checks if the reader was able to read an unstructured grid (by testing if number of nodes is > 0 which was the most low level check i could think of)


endJunction:
:fast_forward:

FileIO/VtkIO/VtuInterface.cpp
```diff
@@ -46,10 +46,18 @@ MeshLib::Mesh* VtuInterface::readVTUFile(std::string const &file_name)
 	vtkSmartPointer<vtkXMLUnstructuredGridReader> reader =
 		vtkSmartPointer<vtkXMLUnstructuredGridReader>::New();
 
+	std::ifstream in( file_name.c_str() );
+	if (!in.is_open())
+		return nullptr;
+	else
+		in.close();
+
```
TomFischer: Why not move line 46 and 47 behind the check?

PR #672, opened 2015-05-11
Fix MeshLayerMapper compilation

endJunction:
This fixes compilation error introduced by removing the empty constructor in the `MeshLayeredMapper`.

This happens with the mac's compiler, which is more strict and produces the following error:
```
<.../ogs/Applications/DataExplorer/DataView/MeshLayerEditDialog.cpp>:261:25: error: default initialization of an object of const type 'const MeshLayerMapper' requires a user-provided default constructor
                MeshLayerMapper const mapper;
                                      ^
<..../ogs/Applications/DataExplorer/DataView/MeshLayerEditDialog.cpp>:308:25: error: default initialization of an object of const type 'const MeshLayerMapper' requires a user-provided default constructor
                MeshLayerMapper const mapper;
                                      ^
```

... but the referred objects are not required at all: removing them.


TomFischer:
To your information: CLang did have also problems to compile the code.
:+1:

PR #673, opened 2015-05-18
Rework of id concept in point classes.

TomFischer:
This PR is a preparation for another PR that will fix a bug in `PointVec::makePointsUnique()` and is at the moment work in progress (branch [GeoLibReworkMergingGeometries](https://github.com/TomFischer/ogs/tree/GeoLibReworkMergingGeometries) in my repo).
For this reason, all geometrical points have to have an id. This PR introduces this behaviour using new implemented (base) class `MathLib::Point3dWithID`. `GeoLib::Point` as well as `MeshLib::Node` are derived from `MathLib::Point3dWithID`.


FileIO/FEFLOWInterface.cpp
```diff
@@ -372,11 +372,11 @@ void FEFLOWInterface::readPoints(QDomElement &nodesEle, const std::string &tag,
 		if (line_str.empty()) continue;
 		std::istringstream line_ss(line_str);
 		size_t pt_id = 0;
-		double pt_xyz[3] = { };
+		std::array<double,3> pt_xyz = { };
```
endJunction: The initialization is not needed here, isn't it?

MathLib/Point3dWithID.h
```diff
@@ -0,0 +1,73 @@
...
+
+#include <limits>
+
+#include "Point3d.h"
+
+namespace MathLib
+{
+/**
+ * Class Point3dWithID is derived from class Point3d in
+ * order to extend the class Point3d with an id.
```
endJunction: The descriptions needs an update.

endJunction:
:fast_forward:

GeoLib/Point.h
```diff
@@ -27,27 +27,31 @@ namespace GeoLib
...
+	{}
+
+	Point(double const* x) :
+		MathLib::Point3dWithID(std::array<double,3>({{x[0], x[1], x[2]}}), 0),
+		GeoLib::GeoObject()
 	{}
 
-	Point (double const* x) :
-		MathLib::Point3d(x), GeoLib::GeoObject()
+	Point(Point const& p, std::size_t id = std::numeric_limits<std::size_t>::max()) :
```
norihiro-w: i think everyone assume a copy constructor should copy all information without changing id. I suggest adding another constructor which explicitly takes Point and new id.

MathLib/Point3dWithID.h
```diff
@@ -0,0 +1,73 @@
...
+	Point3dWithID(std::array<double,3> & coords,
+		std::size_t id = std::numeric_limits<std::size_t>::max())
+		: Point3d(coords), _id(id)
+	{}
+
+	/// Constructs a point with the same coordinates as the given
+	/// Point3d pnt and the provided id.
+	/// @param pnt a MathLib::Point3d object containing the coordinates
+	/// @param id the id of the object [default: max of std::size_t]
+	Point3dWithID(MathLib::Point3d const& pnt,
```
norihiro-w: same here

norihiro-w:
:+1: besides the copy constructor issue

GeoLib/Point.h
TomFischer: You are right. I will remove the default value for the second argument. Then the compiler generates its own copy constructor.

MathLib/Point3dWithID.h
TomFischer: This is not the copy constructor of `Point3dWithID`. This constructor constructs the `Point3dWithID` object using th `Point3d` and the provided id. The compiler generates a copy constructor.

norihiro-w: have you tested it? though i am not sure, since Point3d is a base class of Point3dWithID, compilers might use this constructor as a copy constructor.

endJunction: I'd mark this one with `explicit` specifier to avoid possible usage as a copy ctor.

TomFischer: In the commit 437b9c8 I added tests for the above case. At least clang generates a separate copy constructor. Nevertheless  I added expcit keyword.

PR #657, opened 2015-04-23
Complete linear and quadratic FE types

norihiro-w:
as titled. this PR should be merged after #656

PR #674, opened 2015-05-20
fix compiling errors

norihiro-w:
as titled after PR #657


endJunction:
The changes are straight forward and should be merged asap.

For me it is not clear, why this particular code was not compiled on travis and all three jenkins *-PR builds? Any ideas?

norihiro-w:
@endJunction it's possible that the code is not compiled because FePRISM15 is not included in the test yet.

PR #675, opened 2015-05-21
[GeoLib] Speedup point name access.

TomFischer:
The suggested implementation uses a vector of strings to speed up the access the point name. Thus, more memory is used but the access, especially for point sets with more than 100.000 points is faster (_hours vs. seconds_).

### DIFF
--- a/GeoLib/PointVec.h
+++ b/GeoLib/PointVec.h
@@ -21,8 +21,6 @@
 #include <string>
 #include <vector>
 
-#include <boost/optional.hpp>
-
 #ifndef POINTVEC_H_
 #define POINTVEC_H_
 
--- a/GeoLib/PointVec.h
+++ b/GeoLib/PointVec.h
@@ -21,6 +21,8 @@
 #include <string>
 #include <vector>
 
+#include <boost/optional.hpp>
+
 #ifndef POINTVEC_H_
 #define POINTVEC_H_
 
@@ -100,6 +102,8 @@ public:
 
 	const GeoLib::AABB<GeoLib::Point>& getAABB () const;
 
+	boost::optional<std::string const&> getItemNameByID(std::size_t id) const;
+
 private:
 	/**
 	 * Removes points out of the given point set that have the (nearly) same coordinates, i.e.
@@ -137,6 +141,10 @@ private:
 	 * to their lexicographical order
 	 */
 	std::vector<std::size_t> _pnt_id_map;
+
+	/// The reverse map to the name to id map, for fast lookup of the name to a
+	/// given point id.
+	std::vector<std::string> _id_to_name_map;
 
 	AABB<GeoLib::Point> _aabb;
 };
--- a/GeoLib/PointVec.h
+++ b/GeoLib/PointVec.h
@@ -102,7 +102,7 @@ public:
 
 	const GeoLib::AABB<GeoLib::Point>& getAABB () const;
 
-	boost::optional<std::string const&> getItemNameByID(std::size_t id) const;
+	std::string const& getItemNameByID(std::size_t id) const;
 
 private:
 	/**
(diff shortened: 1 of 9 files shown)


GeoLib/PointVec.h
```diff
@@ -143,6 +145,10 @@ class PointVec : public TemplateVec<Point>
 	 */
 	std::vector<std::size_t> _pnt_id_map;
 
+	/// The reverse map to the name to id map, for fast lookup of the name to a
+	/// given point id.
+	std::vector<std::string> _id_to_name_map;
```
norihiro-w: for me it looks waste of memory because often we assign names to a few number of points. how about storing both string and point id as a lookup table like `std::vector<std::set<std::string, std::size_t>>` or `std::map<std::string, std::size_t>`?

TomFischer: The mapping "name -> item" is already there. It is implemented employing a `std::map<std::string, std::size_t>`. Sometimes the reverse lookup is needed (item or id -> name), for instance, in the case when all points should be written to a file. Then only the id of the point is given and a name is searched. Currently this name search is very slow. The changes in the PR try to speedup the name search.

You are right: this takes more memory. But the new implemented `std::vector<std::string>` allows a faster access (hours vs. seconds).

norihiro-w: sorry it was other way around. however i still assume `std::vector<std::set<std::size_t,std::string>>` or `std::map<std::size_t, std::string>` can also speed up (of course not the fastest, but could be enough fast), because comparison of integers is much faster than that of strings (i mean compared to `std::map<std::string, std::size_t>`).

TomFischer: Will measure time and mem for different versions.

TomFischer: A small util (bc30338) was used to measure the memory consumption of the [master](https://github.com/TomFischer/ogs/tree/Test) and a [map version](https://github.com/TomFischer/ogs/tree/GeoLibSpeedupName2PointAccessMapVersion) of this PR (that uses a `std::vector`). The results are surprising -  at least for me:
comparison
The first three test series are generated using executables created by clang (debug mode). The last three test series are generated using gcc executables (debug mode). In each test 100000 points are generated. The number of point names increases from the left to the right from 0 percent (no point has a name) to 100 percent (i.e. all points have names). Surprisingly, the implementation without a data structure to manage the id -> name mapping (current master) has the same memory usage than the the implementation using the `std::vector` (this PR). A second surprise is that the `std::map` version does not perform better than the `std::vector` version - even for low load. I have no explanation for this.

endJunction: I additionally analysed Tom's test program in valgrind's massif tool for heap memory profiling.
 - For 1e5 points, all with names, we are speaking about 1MiB heap memory usage, which is not that important IMO.
 - The vector version has same memory usage as the master, because the maximum memory is used in the makePointsUnique function, which frees some memory afterwards. This additional memory is used by the `std::vector<std::string>`.
 - The map version indeed uses less memory up to 1e3 names. Minimum is for 0 names around 300KiB.

Summary: I tend to use the vector version because of simpler implementation and faster access compared to the map. The memory usage issues here are within reasonable limits.

norihiro-w: @TomFischer @endJunction  i see. 1MB is not a big issue for us (ca 10MB even with 1e6 points). thanks for testing it!

endJunction:
Some graphics from the massif analysis for 1e5 points all with names.
Three versions:
 - The first two peaks are common code.
 - The vector version has a third peak of memory allocation equal to the previously used memory (that's why there is no difference in maximum memory usage).
 - The map version is less expensive for small amount of names (say 10%) but more expensive otherwise as shown in the third figure.
master

The results were produced with massif-visualizer using output of valgrind:
```sh
parallel 'valgrind --massif-out-file=master/massif.{} --tool=massif --time-unit=B --stacks=no --max-snapshots=1000 --detailed-freq=1 bin/TimeAndMemoryMeasurementForGeometries -p 100000 -n {} > master/out.{}' ::: `seq 0 1000 100000`
```

endJunction:
:fast_forward: after Tom accepted or declined my small cleanup PR https://github.com/TomFischer/ogs/pull/8.

PR #676, opened 2015-05-25
Mesh properties; small addons.

endJunction:
- Add `--print_properties` flag to `checkMesh` utility which is then printing the names of all available properties in the mesh file.
 - Mark `Properties::hasPropertyVector()` const.


TomFischer:
:+1:

norihiro-w:
:+1:

PR #681, opened 2015-05-28
CMake instructions for installing utilities

bilke:
Utilities will later on be available as binary downloads on ogs.org.


endJunction:
Looks ok. How can we check/try this?
:fast_forward:

TomFischer:
:+1:

bilke:
If utils are build then `make install` installs the binaries into `${CMAKE_INSTALL_PREFIX}/bin`.

bilke:
Utilities are now build in a separate Jenkins job (`OGS-6/*_Utils`) and binary downloads are available on http://docs.opengeosys.org/download under *Additional tools*.

TomFischer:
Great work, so the user can download the tools now.

norihiro-w:
:+1:

PR #683, opened 2015-05-29
[GeoLib] Restructure and bug fix

TomFischer:
This PR is based on PR https://github.com/ufz/ogs/pull/675 and must not be merged before PR https://github.com/ufz/ogs/pull/675.

- Small improvements in `GeoLib::PointVec::mergePoints()` 
  - Postfix to prefix increments,
  - Handles returned nullptr value,
  - Rework the loop structure a little bit.
- Using the same eps value for `makePntsUnique` and `uniqueInsert` fixes a inconsistency bug.

### DIFF
--- a/GeoLib/GEOObjects.cpp
+++ b/GeoLib/GEOObjects.cpp
@@ -404,19 +404,22 @@ bool GEOObjects::mergePoints(std::vector<std::string> const & geo_names,
 	std::vector<GeoLib::Point*>* merged_points (new std::vector<GeoLib::Point*>);
 	std::map<std::string, std::size_t>* merged_pnt_names(new std::map<std::string, std::size_t>);
 
-	for (std::size_t j(0); j < n_geo_names; j++) {
-		const std::vector<GeoLib::Point*>* pnts(this->getPointVec(geo_names[j]));
+	for (std::size_t j(0); j < n_geo_names; ++j) {
+		GeoLib::PointVec const*const pnt_vec(this->getPointVecObj(geo_names[j]));
+		if (pnt_vec == nullptr)
+			continue;
+		const std::vector<GeoLib::Point*>* pnts(pnt_vec->getVector());
 		if (pnts) {
 			std::size_t n_pnts(0);
 			// do not consider stations
 			if (!dynamic_cast<GeoLib::Station*>((*pnts)[0])) {
 				std::string tmp_name;
 				n_pnts = pnts->size();
-				for (std::size_t k(0); k < n_pnts; k++) {
-					merged_points->push_back(new GeoLib::Point(((*pnts)[k])->getCoords()));
-					if (this->getPointVecObj(geo_names[j])->getNameOfElementByID(k, tmp_name)) {
+				for (std::size_t k(0); k < n_pnts; ++k) {
+					merged_points->push_back(new GeoLib::Point(*(*pnts)[k]));
+					if (pnt_vec->getNameOfElementByID(k, tmp_name)) {
 						merged_pnt_names->insert(
-								std::pair<std::string, std::size_t>(tmp_name, pnt_offsets[j] + k));
+								std::make_pair(tmp_name, pnt_offsets[j] + k));
 					}
 				}
 			}
--- a/GeoLib/GEOObjects.cpp
+++ b/GeoLib/GEOObjects.cpp
@@ -409,25 +409,27 @@ bool GEOObjects::mergePoints(std::vector<std::string> const & geo_names,
 		if (pnt_vec == nullptr)
 			continue;
 		const std::vector<GeoLib::Point*>* pnts(pnt_vec->getVector());
-		if (pnts) {
-			std::size_t n_pnts(0);
-			// do not consider stations
-			if (!dynamic_cast<GeoLib::Station*>((*pnts)[0])) {
-				std::string tmp_name;
-				n_pnts = pnts->size();
-				for (std::size_t k(0); k < n_pnts; ++k) {
-					merged_points->push_back(new GeoLib::Point(*(*pnts)[k]));
-					if (pnt_vec->getNameOfElementByID(k, tmp_name)) {
-						merged_pnt_names->insert(
-								std::make_pair(tmp_name, pnt_offsets[j] + k));
-					}
-				}
+		if (pnts == nullptr) {
+			return false;
+		}
+
+		// do not consider stations
+		if (dynamic_cast<GeoLib::Station*>((*pnts)[0])) {
+			continue;
+		}
+
+		std::string tmp_name;
+		std::size_t const n_pnts(pnts->size());
+		for (std::size_t k(0); k < n_pnts; ++k) {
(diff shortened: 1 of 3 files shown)


endJunction:
Can you please explain the change of eps from the rel_eps to machine precision in the 	_[GL] Using the same eps for makePntsUnique and uniqueInsert()._ 1d2957d commit?
The rel_eps is usually larger than machine precision and was also used for the making points unique.

GeoLib/GEOObjects.cpp
```diff
@@ -404,27 +404,32 @@ bool GEOObjects::mergePoints(std::vector<std::string> const & geo_names,
...
+		if (pnts == nullptr) {
+			return false;
+		}
+
+		// do not consider stations
+		if (dynamic_cast<GeoLib::Station*>((*pnts)[0])) {
+			continue;
+		}
+
+		std::size_t n_pnts(pnts->size());
```
endJunction: const

GeoLib/PointVec.cpp
```diff
@@ -72,11 +82,12 @@ void PointVec::push_back (Point* pnt, std::string const*const name)
 	std::size_t id (uniqueInsert (pnt));
 	_pnt_id_map.push_back (id);
 	(*_name_id_map)[*name] = id;
+	_id_to_name_map.push_back(*name);
 }
 
 std::size_t PointVec::uniqueInsert (Point* pnt)
 {
-	const double eps (std::numeric_limits<double>::epsilon());
+	double const eps(_rel_eps);
```
endJunction: Just use the `_rel_eps` below. (single place)

endJunction:
:fast_forward:

GeoLib/GEOObjects.cpp
```diff
@@ -404,27 +404,32 @@ bool GEOObjects::mergePoints(std::vector<std::string> const & geo_names,
...
+		GeoLib::PointVec const*const pnt_vec(this->getPointVecObj(geo_names[j]));
+		if (pnt_vec == nullptr)
+			continue;
+		const std::vector<GeoLib::Point*>* pnts(pnt_vec->getVector());
+		if (pnts == nullptr) {
+			return false;
+		}
+
+		// do not consider stations
+		if (dynamic_cast<GeoLib::Station*>((*pnts)[0])) {
```
norihiro-w: need to check `!pnts->empty()`

norihiro-w:
:+1 besides one comment

GeoLib/GEOObjects.cpp
rinkk: I don't think so ... it is not possible to add a geometry without points (that is pretty much the only requirement).

GeoLib/PointVec.cpp
```diff
@@ -32,13 +32,14 @@ PointVec::PointVec (const std::string& name, std::vector<Point*>* points,
...
-	_aabb(points->begin(), points->end())
+	_aabb(points->begin(), points->end()),
+	_rel_eps(rel_eps)
 {
 	assert (_data_vec);
 	std::size_t const number_of_all_input_pnts (_data_vec->size());
 
-	rel_eps *= sqrt(MathLib::sqrDist (_aabb.getMinPoint(),_aabb.getMaxPoint()));
-	makePntsUnique (_data_vec, _pnt_id_map, rel_eps);
+	_rel_eps *= sqrt(MathLib::sqrDist (_aabb.getMinPoint(),_aabb.getMaxPoint()));
```
rinkk: can you pls explain what happens here? i suppose you want a larger eps if the RoI is large but why does *this* particular method work well?

GeoLib/GEOObjects.cpp
```diff
@@ -404,27 +404,32 @@ bool GEOObjects::mergePoints(std::vector<std::string> const & geo_names,
...
-						merged_pnt_names->insert(
-								std::pair<std::string, std::size_t>(tmp_name, pnt_offsets[j] + k));
-					}
-				}
-			}
-			if (n_geo_names - 1 > j) {
-				pnt_offsets[j + 1] = n_pnts + pnt_offsets[j];
+	for (std::size_t j(0); j < n_geo_names; ++j) {
+		GeoLib::PointVec const*const pnt_vec(this->getPointVecObj(geo_names[j]));
+		if (pnt_vec == nullptr)
```
rinkk: i don't see how pnt_vec could be a nullptr. you cannot create geometries without points and you cannot remove points without removing the complete geometry.

GeoLib/GEOObjects.cpp
```diff
@@ -404,27 +404,32 @@ bool GEOObjects::mergePoints(std::vector<std::string> const & geo_names,
...
+			continue;
+		const std::vector<GeoLib::Point*>* pnts(pnt_vec->getVector());
+		if (pnts == nullptr) {
```
rinkk: same here ... i don't think this *can* be a nullptr...

GeoLib/GEOObjects.cpp
```diff
@@ -404,27 +404,32 @@ bool GEOObjects::mergePoints(std::vector<std::string> const & geo_names,
...
-					merged_points->push_back(new GeoLib::Point(((*pnts)[k])->getCoords()));
-					if (this->getPointVecObj(geo_names[j])->getNameOfElementByID(k, tmp_name)) {
-						merged_pnt_names->insert(
-								std::pair<std::string, std::size_t>(tmp_name, pnt_offsets[j] + k));
-					}
-				}
-			}
-			if (n_geo_names - 1 > j) {
-				pnt_offsets[j + 1] = n_pnts + pnt_offsets[j];
+	for (std::size_t j(0); j < n_geo_names; ++j) {
```
rinkk: does it make sense to check if ```n_geo_names > 0```? (i.e. it does not make sense to "merge" a single geometry)

GeoLib/PointVec.cpp
TomFischer: No, at this place there was not the problem. The reason for using now an attribute instead of a local variable is that the same eps (`_rel_eps`) should be used in the method `uniqueInsert()`.

GeoLib/GEOObjects.cpp
TomFischer: It is not impossible to cast the const away and manipulate the internal data structures of `PointVec`. I do not think this check costs so much and can be valuable in case the user makes a mistake.

GeoLib/GEOObjects.cpp
TomFischer: The class `GEOObjects` offers a method that returns a pointer to the `PointVec` object. It is possible to delete the PointVec object from outside.

GeoLib/GEOObjects.cpp
TomFischer: The private method `mergePoints()` is called from `mergeGeometries()`. In `mergeGeometries()` the check is already performed. So at this point we can be sure to have at least two geometries.

rinkk:
:+1:

PR #686, opened 2015-06-02
Initialize with zeron in TemplatePoint default constructor

norihiro-w:
This PR is based on the comment by Tom (https://github.com/norihiro-w/ogs/commit/460d114a7241401719cf0b7c71162bd58cdea820#commitcomment-11468635), and changes the behaviour of a default constructor of MathLib::TemplatePoint to always initialize its coordinate with zero.


TomFischer:
:+1:

PR #687, opened 2015-06-02
Fixing warnings

rinkk:
... using MS Visual Studio.

Note: A few warnings are left in the ```CoordinateSystem``` class because I don't know how to handle them. I already dropped Nori an e-mail.

### DIFF
--- a/MeshLib/VtkOGSEnum.cpp
+++ b/MeshLib/VtkOGSEnum.cpp
@@ -92,7 +92,7 @@ int OGSToVtkCellType(CellType ogs)
 			return VTK_PYRAMID;
 		case CellType::PYRAMID13:
 			return VTK_QUADRATIC_PYRAMID;
-		case CellType::INVALID:
+		default:
 			return -1;
 	}
 }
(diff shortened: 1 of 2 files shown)


MeshLib/VtkOGSEnum.cpp
```diff
@@ -92,7 +92,7 @@ int OGSToVtkCellType(CellType ogs)
 			return VTK_PYRAMID;
 		case CellType::PYRAMID13:
 			return VTK_QUADRATIC_PYRAMID;
-		case CellType::INVALID:
+		default:
```
endJunction: This is reverting the commit 65e52ab9b which was intended to explicitly handle all possible enum types in the switch case.
If one would add an entry to the CellType enum a warning about not all handled cases would be shown.

Now adding a default case makes it possible to add another CellType entry without a warning.

rinkk: hmmm ... i see ... the problem is that Visual Studio gives a warning because there is no default case ...

norihiro-w: why not add macros to include a default case only when MSVC is used? we can still get warnings with gcc and clangs if one would add an entry to the CellType enum.

PR #691, opened 2015-06-04
Cleanup

endJunction:
Update submodules, small fixes.

### DIFF
--- a/GeoLib/Raster.cpp
+++ b/GeoLib/Raster.cpp
@@ -148,9 +148,7 @@ double Raster::interpolateValueAtPoint(MathLib::Point3d const& pnt) const
 	for (unsigned j=0; j<4; ++j)
 	{
 		// check if neighbour pixel is still on the raster, otherwise substitute a no data value
-		if ( (xIdx + x_nb[j]) < 0 ||
-		     (yIdx + y_nb[j]) < 0 ||
-		     (xIdx + x_nb[j]) > _ll_pnt[0]+(_n_cols*_cell_size) ||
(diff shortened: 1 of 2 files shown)


GeoLib/Raster.cpp
```diff
@@ -148,9 +148,7 @@ double Raster::interpolateValueAtPoint(MathLib::Point3d const& pnt) const
 	for (unsigned j=0; j<4; ++j)
 	{
 		// check if neighbour pixel is still on the raster, otherwise substitute a no data value
-		if ( (xIdx + x_nb[j]) < 0 ||
-		     (yIdx + y_nb[j]) < 0 ||
```
rinkk: why did you remove the check if the calculated index is < 0?
that was precisely why the programme had crashed before?!

rinkk:
I've looked it up and the calculation of the index is indeedd an int. however, in line 157 (in the original code) there is an implicit cast to std::size_t. I don't know why my compiler does not mention this. We could add a static cast there if other compilers give a warning.

GeoLib/Raster.cpp
endJunction: `xIdx` is a `std::size_t`, therefore the result of addition with an `int` is a `std::size_t`, which is never smaller than zero.

rinkk: why is ```std::size_t + int = std::size_t``` and not int?

endJunction: Please lookup the standard for integral conversions.
An int cannot be larger than std::size_t.

To end the discussion on this point here I opened an issue which needs a resolution: https://github.com/ufz/ogs/issues/693

bilke:
:+1: if the cast is added

PR #692, opened 2015-06-05
Vtk result output; vtkdiff

endJunction:
Write the GroundWaterflowProcess result to vtu file.
Add vtkdiff tester (also please review the [vtkdiff](https://github.com/ufz/vtkdiff) repository) replacing the numdiff.

The test input files now include the analytical solution. There is a [python code](https://github.com/endJunction/ogs/blob/python_code/solutions.py) for generating the solutions, which I would like to include too but I don't know where to put it. Ideas?

Known issues:
 - The code for writing the output shall be moved somewhere else and linked with the `Process::post()` function  via callback.
 - Some of the LARGE*vtkdiff tests are failing because the comparison method of vtkdiff is slightly different to numdiff.


endJunction:
I'm changing the neumann ctest a little; that's the reason why linux-pr fails. Compilation and gtests are ok.

bilke:
It compiles fine on Mac OS. Although I could not run `make ctest` because the file `796efdd29f55ed91c70d056a3faeb947` was not found on the server.

The python files can be placed in `scripts/test/...` maybe.

:+1:

endJunction:
Jenkins test this please

endJunction:
Jenkins test this please

endJunction:
OK. Following situation:
Code compiles, the gtests are green on all platforms. On linux we also run the ctests, which are now testing the Neumann setups. The tests run through and vtkdiff compares correct fields. Only the results are different from the expected. This has nothing to do with the vtkdiff or vtk output. So I remove the tests again and will fix this issue https://github.com/ufz/ogs/issues/697.

PR #698, opened 2015-06-08
Lower dim elems for neumann bcs

endJunction:
Following the vtk output PR.

Extend Neumann boundary conditions to elements not aligned with the natural coordinate system using a local coordinate system mapping.

Provide analytical solutions and add tests for the Neumann bc ctests. Also see the documentation at [opengeosys.org/docs](http://docs.opengeosys.org/docs/benchmarks/elliptic/groundwater-flow-neumann).

Fixes ufz/ogs#697.

PR #702, opened 2015-06-09
Supporting mixed dimensional elements in FE

norihiro-w:
@endJunction sorry i forgot to upload implementations supporting mixed dimensional elements in FE part. your #698 can be merged at the moment, as it should be no problem with neumann BCs because it doesn't require dshape. To really support mixed dim ele in groundwaterflow, we need something like this PR.

### DIFF
--- a/NumLib/Fem/CoordinatesMapping/NaturalCoordinatesMapping-impl.h
+++ b/NumLib/Fem/CoordinatesMapping/NaturalCoordinatesMapping-impl.h
@@ -16,6 +16,10 @@
 
 #include "logog/include/logog.hpp"
 
+#include "MeshLib/Node.h"
+#include "MeshLib/ElementCoordinatesMappingLocal.h"
+#include "MeshLib/CoordinateSystem.h"
+
 namespace NumLib
 {
 
@@ -28,6 +32,7 @@ template <class T_MESH_ELEMENT, class T_SHAPE_FUNC, class T_SHAPE_MATRICES>
 inline void computeMappingMatrices(
         const T_MESH_ELEMENT &/*ele*/,
         const double* natural_pt,
+        const MeshLib::ElementCoordinatesMappingLocal &/*ele_local_coord*/,
         T_SHAPE_MATRICES &shapemat,
         FieldType<ShapeMatrixType::N>)
 {
@@ -38,6 +43,7 @@ template <class T_MESH_ELEMENT, class T_SHAPE_FUNC, class T_SHAPE_MATRICES>
 inline void computeMappingMatrices(
         const T_MESH_ELEMENT &/*ele*/,
         const double* natural_pt,
+        const MeshLib::ElementCoordinatesMappingLocal &/*ele_local_coord*/,
         T_SHAPE_MATRICES &shapemat,
         FieldType<ShapeMatrixType::DNDR>)
 {
@@ -49,22 +55,23 @@ template <class T_MESH_ELEMENT, class T_SHAPE_FUNC, class T_SHAPE_MATRICES>
 inline void computeMappingMatrices(
         const T_MESH_ELEMENT &ele,
         const double* natural_pt,
+        const MeshLib::ElementCoordinatesMappingLocal &ele_local_coord,
         T_SHAPE_MATRICES &shapemat,
         FieldType<ShapeMatrixType::DNDR_J>)
 {
     computeMappingMatrices<T_MESH_ELEMENT, T_SHAPE_FUNC, T_SHAPE_MATRICES>
-        (ele, natural_pt, shapemat, FieldType<ShapeMatrixType::DNDR>());
+        (ele, natural_pt, ele_local_coord, shapemat, FieldType<ShapeMatrixType::DNDR>());
 
     const std::size_t dim = T_MESH_ELEMENT::dimension;
     const std::size_t nnodes = T_MESH_ELEMENT::n_all_nodes;
 
     //jacobian: J=[dx/dr dy/dr // dx/ds dy/ds]
     for (std::size_t k=0; k<nnodes; k++) {
-        double const* const xyz = ele.getNode(k)->getCoords();
+        const MeshLib::Node& mapped_pt = *ele_local_coord.getMappedCoordinates(k);
         // outer product of dNdr and xyz for a particular node
         for (std::size_t i_r=0; i_r<dim; i_r++) {
             for (std::size_t j_x=0; j_x<dim; j_x++) {
-                shapemat.J(i_r,j_x) += shapemat.dNdr(i_r,k) * xyz[j_x];
+                shapemat.J(i_r,j_x) += shapemat.dNdr(i_r,k) * mapped_pt[j_x];
             }
         }
     }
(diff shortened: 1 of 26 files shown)


bilke:
Formally: :+1:

NumLib/Fem/ShapeMatrixPolicy.h
```diff
@@ -34,7 +34,8 @@ struct EigenFixedShapeMatrixPolicy
         NumLib::ShapeMatrices<
             NodalVectorType,
             DimNodalMatrixType,
-            DimMatrixType>;
+            DimMatrixType,
+            DimNodalMatrixType>;
```
endJunction: Could one potentially reuse the second type in the list?

Tests/NumLib/TestCoordinatesMapping.cpp
```diff
@@ -278,4 +282,38 @@ TYPED_TEST(NumLibFemNaturalCoordinatesMappingTest, CheckZeroVolume)
...
+	typedef Eigen::Matrix<double, 1, 2, Eigen::RowMajor> DimNodalMatrix;
+	typedef Eigen::Matrix<double, 1, 1, Eigen::RowMajor> DimMatrix;
+	typedef Eigen::Matrix<double, 2, 2, Eigen::RowMajor> GlobalDimNodalMatrix;
+#endif
+	// Shape data type
+	typedef ShapeMatrices<NodalVector,DimNodalMatrix,DimMatrix,GlobalDimNodalMatrix> ShapeMatricesType;
+	typedef NaturalCoordinatesMapping<MeshLib::Line, ShapeLine2, ShapeMatricesType> MappingType;
+	double r[] = {0.5};
+	auto line = TestLine2::createY();
+	//std::cout << *line;
```
TomFischer: I guess this line and line 303 are debug outputs? Is it possible to remove them or do we need the output later on?

TomFischer:
@norihiro-w Could you put a reference in the source code to an article or a book that explains mixed dimensional elements a bit. For example, it is not clear to me what a 1.5 dimensional line is. Is it correct that mixed dimensional elements have different shape matrices?

NumLib/Fem/ShapeMatrixPolicy.h
norihiro-w: indeed this is a bug. I should pass a matrix type having a size of global dimension x number of nodes. see the latest where the bug is fixed.

norihiro-w:
@TomFischer i added a reference. see 4340340

TomFischer:
> @TomFischer i added a reference. see 4340340

Thank you!

norihiro-w:
Dear all, please review the code again. I had to introduce more change in the code to fix a bug commented by @endJunction

NumLib/Fem/CoordinatesMapping/NaturalCoordinatesMapping.h
```diff
@@ -23,6 +23,11 @@ namespace NumLib
 /**
  * Coordinates mapping tools for natural coordinates
  *
+ * This class also supports coordinates mapping of mixed dimensional elements,
+ * e.g. line elements in 2D space. Details of the mapping method can be found in
+ * Kolditz O (2001) Non-linear flow in fractured rock. International Journal of
+ * Numerical Methods for Heat & Fluid Flow 11:547–575.
+ *
```
endJunction: Please add this to doxygen references like [here](https://github.com/ufz/ogs/blob/master/GeoLib/AnalyticalGeometry.h#L59) and add the bibtex entry to [Documentation/bibliography.bib](https://github.com/ufz/ogs/blob/master/Documentation/bibliography.bib).

ProcessLib/GroundwaterFlowProcess.h
```diff
@@ -159,8 +140,39 @@ class GroundwaterFlowProcess : public Process
...
+        _x.reset(_global_setup.createVector(_local_to_global_index_map->dofSize()));
+        _rhs.reset(_global_setup.createVector(_local_to_global_index_map->dofSize()));
+        _linearSolver.reset(new typename GlobalSetup::LinearSolver(*_A));
 
+        if (_mesh.getDimension()==1)
+            createLocalAssemblers<1>();
+        else if (_mesh.getDimension()==2)
+            createLocalAssemblers<2>();
+        else if (_mesh.getDimension()==3)
+            createLocalAssemblers<3>();
```
endJunction: I'd add an extra else with assert:
```c++
else
    assert(false);
```

endJunction:
Regarding the global dimension template parameter:
I guess we could do some more decoupling, but first of all it works.

I run comparison of ufz/master with my lower-dim PR (based on this one) for dirichlet and neumann on 1e4 square ctest using callgrind:
Number of instructions executed:

          | dirichlet | neumann     |
----------|-----------|-------------|
master    |848,454,453|1,118,595,144|
lower_dim |941,440,774|1,200,626,996|
diff      |92,986,321 |82,031,852   |

The difference is not in lis solver (actually it uses exactly the same number of instructions).

Not that nice, but we need this generalization. I shall look into the detailed analysis later.

@TomFischer have one more look on this and :fast_forward:

NumLib/Fem/CoordinatesMapping/NaturalCoordinatesMapping-impl.h
```diff
@@ -28,6 +32,7 @@ template <class T_MESH_ELEMENT, class T_SHAPE_FUNC, class T_SHAPE_MATRICES>
 inline void computeMappingMatrices(
         const T_MESH_ELEMENT &/*ele*/,
         const double* natural_pt,
+        const MeshLib::ElementCoordinatesMappingLocal &/*ele_local_coord*/,
```
TomFischer: The `ElementCoordinatesMappingLocal` is constructed using an element, isn't it? And an element is given as the first argument of `computeMappingMatrices()`.

Maybe a stupid question: Does it make sense to construct the object of class `ElementCoordinatesMappingLocal` within `computeMappingMatrices()`?

NumLib/Fem/CoordinatesMapping/NaturalCoordinatesMapping-impl.h
```diff
@@ -49,22 +55,23 @@ template <class T_MESH_ELEMENT, class T_SHAPE_FUNC, class T_SHAPE_MATRICES>
...
+        (ele, natural_pt, ele_local_coord, shapemat, FieldType<ShapeMatrixType::DNDR>());
 
     const std::size_t dim = T_MESH_ELEMENT::dimension;
     const std::size_t nnodes = T_MESH_ELEMENT::n_all_nodes;
 
     //jacobian: J=[dx/dr dy/dr // dx/ds dy/ds]
     for (std::size_t k=0; k<nnodes; k++) {
-        double const* const xyz = ele.getNode(k)->getCoords();
+        const MeshLib::Node& mapped_pt = *ele_local_coord.getMappedCoordinates(k);
         // outer product of dNdr and xyz for a particular node
```
TomFischer: I think `xyz` int the comment should be `mapped_pt`?!

NumLib/Fem/CoordinatesMapping/NaturalCoordinatesMapping-impl.h
norihiro-w: I think it's better to create the object outside of `computeMappingMatrices()`  because some `computeMappingMatrices()` internally call other `computeMappingMatrices()`, e.g. `computeMappingMatrices()` for `FieldType<ShapeMatrixType::DNDX>` calls `computeMappingMatrices()` for `FieldType<ShapeMatrixType::DNDR>`. `ElementCoordinatesMappingLocal` object is required in both functions.

TomFischer: Thanks for explaination.

norihiro-w:
updated

Issue #697, opened 2015-06-08
Neumann tests failing.


endJunction:
Investigating.

endJunction:
Easy cure (for ogs6, not for me); I forgot to implement rotation of lower dimensional elements for the Neumann boundary conditions.
PR is following.

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/697).

Issue #705, opened 2015-06-09
Configuration of linear eqs. solvers through project file.


endJunction:
forward configuration of linear equation system solvers to the actual solvers instead of using the only hardcoded set.

chleh:
I think this issue is obsolete now.

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/705).

Issue #704, opened 2015-06-09
Improve error bounds for analytical solutions


endJunction:
provide a sharper margin for each test (probably based on the mesh resolution) instead of general margin for the vtkdiff tests.

chleh:
Maybe the primary reference solution should be one from a test run on some machine. Then the error bounds can be much smaller since we don't have to account for discretization errors.

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/704).

This issue / pull request was **open** before the GitLab migration.

PR #706, opened 2015-06-09
Neumann BC ctests

endJunction:
Based on the general approach to include lower dimensional elements presented in ufz/ogs#702 this adds the neumann boundary conditition tests.

 - Provide analytical solutions and add tests for the Neumann bc ctests. Also see the documentation at opengeosys.org/docs.
 - Fixes ufz/ogs#697.

Know issues:
 - I had to set the comparison error to 1e-2, which is pretty large for the analytical benchmarks. Possible resolution are
   1. to provide a sharper margin for each test (probably based on the mesh resolution) instead of general margin for the vtkdiff tests. (ufz/ogs#704)
   2. forward configuration of linear equation system solvers to the actual solvers instead of using the only hardcoded set. (ufz/ogs#705)


Applications/CLI/Tests.cmake
```diff
@@ -21,6 +21,8 @@ foreach(mesh_size 1e0 1e1 1e2 1e3)
 		EXECUTABLE ogs
 		EXECUTABLE_ARGS cube_${mesh_size}_neumann.prj
 		WRAPPER time
+		TESTER vtkdiff
+        DIFF_DATA cube_${mesh_size}_neumann.vtu D1_left_front_N1_right Result
```
TomFischer: Is the indentation correct?

TomFischer:
Apart from small comment and my comments in Noris PR :fast_forward:

bilke:
Needs to be rebased.

TomFischer:
Rebased.

Issue #710, opened 2015-06-10
Make AABB<Point3d> possible.


endJunction:
There is a bunch of incompatible constructors or arguments of those constructors for point like classes. It would be nice to unify how such points are created (at least those of dimension 3).

Example usage would be changing `MeshLib::CoordinateSystem::CoordinateSystem(Element const&)` ctor where a `GeoLib::AABB<MeshLib::Node>` class is instantiated to a `GeoLib::AABB<MeshLib::Point3d>`.
This would eliminate unnecessary constructions of very heavy (for this task) `MeshLib::Node`s. A callgrind graph showing some details:
![coordinate_system](https://cloud.githubusercontent.com/assets/329493/8089408/34ec2a7c-0f97-11e5-9d7a-9fc49815f1fe.png)

TomFischer:
If you want, I can take a look at this issue.

endJunction:
Benchmarked again and the creation and destruction of the Node's is gone because AABB accepts the result of `MeshLib::Element::getNode(i)` without conversion.
No numbers, because I forgot to note the exact benchmark setup.

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/710).

PR #711, opened 2015-06-11
Optimization of AABB

TomFischer:
This PR is a result of https://github.com/ufz/ogs/issues/710. @endJunction discovered, that creating AABB objects is relative expensive. So in this PR the size of management overhead of the class `AABB<Node>` is reduced  from 304 byte to 64 byte. I am curious on the new callgraph and run time comparisons.

### DIFF
--- a/Applications/DataExplorer/DataView/ElementTreeModel.cpp
+++ b/Applications/DataExplorer/DataView/ElementTreeModel.cpp
@@ -140,8 +140,8 @@ void ElementTreeModel::setMesh(MeshLib::Mesh const*const mesh)
 	_rootItem->appendChild(aabb_item);
 
 	const GeoLib::AABB<MeshLib::Node> aabb (MeshLib::MeshInformation::getBoundingBox(*mesh));
-	const MeshLib::Node& min = aabb.getMinPoint();
-	const MeshLib::Node& max = aabb.getMaxPoint();
+	auto const& min = aabb.getMinPoint();
+	auto const& max = aabb.getMaxPoint();
 
 	QList<QVariant> min_aabb;
 	min_aabb << "Min:" << QString::number(min[0], 'f') << QString::number(min[1], 'f') << QString::number(min[2], 'f');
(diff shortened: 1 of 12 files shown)


bilke:
:+1:

Applications/DataExplorer/DataView/ElementTreeModel.cpp
```diff
@@ -140,8 +140,8 @@ void ElementTreeModel::setMesh(MeshLib::Mesh const*const mesh)
 	_rootItem->appendChild(aabb_item);
 
 	const GeoLib::AABB<MeshLib::Node> aabb (MeshLib::MeshInformation::getBoundingBox(*mesh));
-	const MeshLib::Node& min = aabb.getMinPoint();
-	const MeshLib::Node& max = aabb.getMaxPoint();
+	auto min = aabb.getMinPoint();
```
norihiro-w: no need to write `auto&`? maybe there is no difference but `aabb.getMinPoint()` returns a reference.

TomFischer: Changed and rebased.

norihiro-w:
:+1:

endJunction:
One could move the default initialization of min and max points in aabb to variable declaration such that not to repeat it in init lists of the constructors.
(The mobile version of the site does not allow in-code comments.)
:fast_forward: in any case.

endJunction:
Two general questions.
1. Why are the ids of input points used for selection of a subset? Couldn't this be done in the calling function? This would allow to switch to mathlib::point3d, or even to templatePoint.
2. Would the following test fail?
```c++
Point a = ...;
std::vector pnts({{a}});
Auto box = AABB(pnts.begin(), pnts.end());
assert(box.containsPoint(a));
```

TomFischer:
**Answering your first bunch of questions:**
> Why are the ids of input points used for selection of a subset?

Only one constructor is offering this possibilty.

> Couldn't this be done in the calling function?

This would be possible. As far as I see this constructor is used seldom. Then it is possible to remove the corresponding constructor.

> This would allow to switch to mathlib::point3d, or even to templatePoint.

This question is unclear to me. What exactly would allow to switch to `MathLib::Point3d` and what does "switch" mean. Can you explain it?

**Answer to 2.**
I do not see a reason why the test should not run succesfully.

endJunction:
Regarding the first: I think that selection of points and AABB algorithms should be separate. Therefore the suggestions. Would it be possible then to remove the template and work only on point3d derived objects?
Second: those who can read have definitely an advantageous position :-) . Never mind.

PR #714, opened 2015-06-12
GeoLib OctTree

TomFischer:
This PR contains a rework of class `OctTree`, which will be used in another PR fixing some issues in `GeoLib::GEOObject` that is already in preparation (see branch [GeoLibReworkMergingGeometries](https://github.com/TomFischer/ogs/tree/GeoLibReworkMergingGeometries)). Also in this PR a tests for `OctTree` are added.

Maybe the `OctTree` can also be useful for other things like mesh node search?

### DIFF
--- a/GeoLib/OctTree-impl.h
+++ b/GeoLib/OctTree-impl.h
@@ -1,0 +1,267 @@
+/**
+ * \date   2015-06-12
+ * \brief  Implementation of the OctTree class.
+ *
+ * \copyright
+ * Copyright (c) 2012-2015, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+
+namespace GeoLib {
+
+template <typename POINT, std::size_t MAX_POINTS>
+template <typename T>
+OctTree<POINT, MAX_POINTS> OctTree<POINT, MAX_POINTS>::createOctTree(T ll, T ur,
+	double eps)
+{
+	// compute an axis aligned cube around the points ll and ur
+	const double dx(ur[0] - ll[0]);
+	const double dy(ur[1] - ll[1]);
+	const double dz(ur[2] - ll[2]);
+	if (dx >= dy && dx >= dz) {
+		ll[1] -= (dx-dy)/2.0;
+		ur[1] += (dx-dy)/2.0;
+		ll[2] -= (dx-dz)/2.0;
+		ur[2] += (dx-dz)/2.0;
+	} else {
+		if (dy >= dx && dy >= dz) {
+			ll[0] -= (dy-dx)/2.0;
+			ur[0] += (dy-dx)/2.0;
+			ll[2] -= (dy-dz)/2.0;
+			ur[2] += (dy-dz)/2.0;
+		} else {
+			ll[0] -= (dz-dx)/2.0;
+			ur[0] += (dz-dx)/2.0;
+			ll[1] -= (dz-dy)/2.0;
+			ur[1] += (dz-dy)/2.0;
+		}
+	}
+	return OctTree<POINT, MAX_POINTS>(ll, ur, eps);
+}
+
+template <typename POINT, std::size_t MAX_POINTS>
+OctTree<POINT, MAX_POINTS>::~OctTree()
+{
+	for (auto c : _children)
+		delete c;
+}
+
+template <typename POINT, std::size_t MAX_POINTS>
+bool OctTree<POINT, MAX_POINTS>::addPoint(POINT * pnt, POINT *& ret_pnt)
+{
+	// first do a range query using a epsilon box around the point pnt
+	std::vector<POINT*> query_pnts;
+	MathLib::Point3d min(
+		std::array<double,3>{{(*pnt)[0]-_eps, (*pnt)[1]-_eps, (*pnt)[2]-_eps}});
+	MathLib::Point3d max(
+		std::array<double,3>{{(*pnt)[0]+_eps, (*pnt)[1]+_eps, (*pnt)[2]+_eps}});
+	getPointsInRange(min, max, query_pnts);
+	if (! query_pnts.empty()) {
+		// check Euclidean norm
+		for (auto p : query_pnts) {
+			if (MathLib::sqrDist(*p, *pnt) < _eps*_eps) {
+				ret_pnt = p;
+				return false;
+			}
+		}
+	}
+
+	// the point pnt is not yet in the OctTree
+	if (isOutside(pnt)) {
+		ret_pnt = nullptr;
+		return false;
+	}
+
+	// at this place it holds true that the point is within [_ll, _ur]
+	if (!_is_leaf) {
+		for (auto c : _children) {
+			if (c->addPoint_(pnt, ret_pnt)) {
+				return true;
+			} else {
+				if (ret_pnt != nullptr)
+					return false;
+			}
+		}
+	}
+
+	ret_pnt = pnt;
+
+	if (_pnts.size () < MAX_POINTS) {
+		_pnts.push_back(pnt);
+	} else { // i.e. _pnts.size () == MAX_POINTS
+		splitNode(pnt);
+		_pnts.clear();
+	}
+	return true;
+}
+
+template <typename POINT, std::size_t MAX_POINTS>
+template <typename T>
+void OctTree<POINT, MAX_POINTS>::getPointsInRange(T const& min, T const& max,
+	std::vector<POINT*> &pnts) const
+{
+	if (_ur[0] < min[0] || _ur[1] < min[1] || _ur[2] < min[2])
+		return;
+
+	if (max[0] < _ll[0] || max[1] < _ll[1] || max[2] < _ll[2])
+		return;
+
+	if (_is_leaf) {
+		for (auto p : _pnts) {
+			if (min[0] <= (*p)[0] && (*p)[0] < max[0]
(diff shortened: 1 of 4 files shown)


GeoLib/OctTree-impl.h
```diff
@@ -0,0 +1,260 @@
...
+namespace GeoLib {
+
+template <typename POINT>
+OctTree<POINT> OctTree<POINT>::createOctTree(POINT ll, POINT ur,
+	double eps, std::size_t max_items_per_node)
+{
+	OctTree<POINT>::_max_items_per_node = max_items_per_node;
+	OctTree<POINT>::_eps = eps;
+	// compute an axis aligned cube around the points ll and ur
+	const double dx(ur[0] - ll[0]);
```
endJunction: Without repeating the algorithm an AABB could be used here.

GeoLib/OctTree-impl.h
```diff
@@ -0,0 +1,260 @@
...
+
+template <typename POINT>
+OctTree<POINT>::~OctTree()
+{
+	for (std::size_t k(0); k < 8; ++k)
+		delete _childs[k];
+}
+
+template <typename POINT>
+bool OctTree<POINT>::addPoint(POINT * pnt, POINT *& ret_pnt)
```
endJunction: The return type could be a boost::optional<point> if not too difficult to change.
The input could be a reference to a point.

GeoLib/OctTree-impl.h
```diff
@@ -0,0 +1,260 @@
...
+		return false;
+	}
+	if ((*pnt)[0] > _ur[0] || (*pnt)[1] > _ur[1] || (*pnt)[2] > _ur[2]) {
+		ret_pnt = nullptr;
+		return false;
+	}
+
+	// at this place it holds true that the point is within [_ll, _ur]
+	if (!_is_leaf) {
+		for (std::size_t k(0); k < 8; ++k) {
```
endJunction:
```c++
for (auto child : _children)
    child = ...
```

GeoLib/OctTree-impl.h
```diff
@@ -0,0 +1,260 @@
...
+		std::array<double,3>{{(*pnt)[0]+_eps, (*pnt)[1]+_eps, (*pnt)[2]+_eps}});
+	getPointsInRange(min, max, query_pnts);
+	if (! query_pnts.empty()) {
+		// point can not be added since it is already there
+		ret_pnt = query_pnts[0];
+		return false;
+	}
+
+	// the point pnt is not yet in the OctTree
+	if ((*pnt)[0] < _ll[0] || (*pnt)[1] < _ll[1] || (*pnt)[2] < _ll[2]) {
```
endJunction: Componentwise comparison is used more than once. I'd outsource it.

GeoLib/OctTree-impl.h
```diff
@@ -0,0 +1,260 @@
...
+	}
+}
+
+template <typename POINT>
+OctTree<POINT>::OctTree(MathLib::Point3d const& ll, MathLib::Point3d const& ur)
+	: _ll(ll), _ur(ur), _is_leaf(true)
+{
+	// init childs
+	for (std::size_t k(0); k<8; ++k)
+		_childs[k] = nullptr;
```
endJunction:
```c++
Children.fill(nullptr);
```

GeoLib/OctTree-impl.h
TomFischer: I do not see how the AABB can help at this place. Here a special AABB is computed - a cube. The input points `ll` and `ur` needs not to describe a cube.

GeoLib/OctTree-impl.h
```diff
@@ -0,0 +1,260 @@
...
+{
+	if ((*pnt)[0] < _ll[0] || (*pnt)[1] < _ll[1] || (*pnt)[2] < _ll[2])
+		return false;
+	if ((*pnt)[0] > _ur[0] || (*pnt)[1] > _ur[1] || (*pnt)[2] > _ur[2])
+		return false;
+
+	if (_pnts.size() < OctTree<POINT>::_max_items_per_node) {
+		_pnts.push_back(pnt);
+	} else { // i.e. _pnts.size () == OctTree<POINT>::_max_items_per_node
+		splitNode(pnt);
```
endJunction: What will happen here?:
```c++
std::vector<point> ps(100, point(0,0,0));
Octree o(ps);
```

Tests/GeoLib/TestOctTree.cpp
```diff
@@ -0,0 +1,325 @@
...
+TEST_F(GeoLibOctTree, TestWithEquidistantPoints3d)
+{
+	generateEquidistantPoints3d();
+	double const eps(10*std::numeric_limits<double>::epsilon());
+	GeoLib::OctTree<GeoLib::Point> oct_tree(
+		GeoLib::OctTree<GeoLib::Point>::createOctTree(*ps_ptr.front(),
+		*ps_ptr.back(), eps, 2));
+
+#ifndef NDEBUG
+	MathLib::Point3d const& ll(oct_tree.getLowerLeftCornerPoint());
```
endJunction: I'd remove the `#ifndef NDEBUG` and let all checks run.

GeoLib/OctTree-impl.h
endJunction: I missed  the cube part. Thanks.

Tests/GeoLib/TestOctTree.cpp
TomFischer: The reason I activated this code only in debug mode is that some additional methods in the OctTree are only activated in debug mode. These OctTree methods are useful only to investigate the structure of the OctTree object at hand and are not needed in the "usual" use. So a removal of `#ifndef NDEBUG` here needs a removal of `#ifndef NDEBUG` in OctTree. I am unsure if this is a good idea.

GeoLib/OctTree-impl.h
TomFischer: With the current interface this would not compile. 
  1. The OctTree will  only manage pointers to points.
  2. It is only possible to add pointer to points one after another.
  3. Inserting equal points (equal except for eps radius) is not possible, the second occurence will be rejected via a returned false (see addPoint())

TomFischer:
Summary before weekend: 
- The question for the signature of `addPoint()` is still open.
- Other points are either answered or already changed.

GeoLib/OctTree-impl.h
endJunction: Thanks for clarification.I was originally asking about the third point, because it seemed to me that adding the same point more then 8 times  would be a special and not covered case. Now I understand better.

GeoLib/OctTree.h
```diff
@@ -15,116 +15,80 @@
...
-			_pnts.push_back (pnt);
-		else
-			return false;
-
-		if (_pnts.size () > OctTree<POINT>::_max_points_per_node)
-			splitNode ();
-		return true;
-	}
+private:
+	static double _eps;
```
norihiro-w: why static? should we share the same epsilon for all Octree objects?

TomFischer:
@endJunction wrote:
> The return type could be a boost::optional if not too difficult to change. 

The `addPoint()` method is returning a bit more information than would fit in a `boost::optional`. In case the point could not be inserted (return value false) the user also get the reason via the `ret_pnt`: If `ret_pnt` is a `nullptr` the point is outside of the `OctTree` cubic domain. If the `ret_pnt` is not equal to a `nullptr` the returned value is the pointer to a point within the eps-radius that was already inserted into the `OctTree` before.

@endJunction wrote:
> The input could be a reference to a point.

If the `OctTree` stores references or objects instead of pointers in my opinion this would result in some performance issues. I expect that `OctTree` nodes are often split. As a consequence a recreation of `POINT` objects for the childs and destroying the `POINT` objects for the split node are necessary very often.

GeoLib/OctTree.h
TomFischer: Good point. On the one hand I want only one epsilon (stored once) for one `OctTree`. On the other hand different `OctTree`s can and should have different epsilon. I am thinking of a good solution. Do you (or somebody else) have any suggestions how to solve this issue?

TomFischer: Changed from class variable to class member variable. I also changed the handling of the number of maximal points per `OctTree` node from a static variable to a template parameter. So it is possible to have for each `OctTree` it is own value.

endJunction:
Thanks Tom for clarifying the open questions.
I would though insist on use proper English "children" for the plural of a child.
:fast_forward:

norihiro-w:
:+1:

TomFischer:
Rebased and added changelog entry. There are not any further source code changes.

GeoLib/OctTree.h
```diff
@@ -15,116 +15,84 @@
...
+	/// the OctTree.
+	/// @param ll lower left front point, used for computation of cubic domain
+	/// @param ur upper right back point, used for computation of cubic domain
+	/// @param eps the euclidean distance as a threshold to make objects unique
+	/// [default std::numeric_limits<double>::epsilon()]
+	/// @param max_items_per_node The max number of items per OctTree node.
+	/// Adding a new item to an already "filled" OctTree node results in a
+	/// split of the OctTree node. The smaller this number is the more leafes
+	/// the OctTree will have, i.e. it needs more memory and more time to walk
+	/// through the OctTree, but the search inside a leafe is fast. In
```
bilke: Typo: `leaf`

GeoLib/OctTree.h
```diff
@@ -15,116 +15,84 @@
...
+	/// contrast a big value results into a smaller number of OctTree leafes,
```
bilke: Typo: `leafs`

TomFischer: Corrected.

PR #716, opened 2015-06-16
PropertyVector usage improvements.

TomFischer:
For better usablility and readability of the `PropertyVector` this PR contains:
- Added `PropertyVector::getNumberOfTuples()` which returns the number of tuples.
- Changed the behaviour of `PropertyVector::size()`. The method now returns the number of tuples times the number of components per tuple.


norihiro-w:
:+1:

PR #717, opened 2015-06-16
Cleanup: Add point ids.

TomFischer:
In PR [673](https://github.com/ufz/ogs/pull/673) an id was introduced to the `GeoLib::Point`. At some places I missed to add this id when points are created. This PR is adding the ids to the points.


bilke:
:+1:

endJunction:
:fast_forward:

PR #720, opened 2015-06-19
GeoLib OctTree - some small improvements

TomFischer:
- Handle case where the given `eps` is exactly zero -> the `eps` will be set internally to `std::numeric_limits<double>::epsilon()`
- Due to technical reasons OctTree manages now the spatial domain [_ll[0], _ur[0]**)** x [_ll[1], _ur[1]**)** x [_ll[2], _ur[2]**)**. The given upper bound at the construction time will be raised a little bit.

### DIFF
--- a/GeoLib/OctTree-impl.h
+++ b/GeoLib/OctTree-impl.h
@@ -72,7 +72,7 @@ bool OctTree<POINT, MAX_POINTS>::addPoint(POINT * pnt, POINT *& ret_pnt)
 	if (! query_pnts.empty()) {
 		// check Euclidean norm
 		for (auto p : query_pnts) {
-			if (MathLib::sqrDist(*p, *pnt) < _eps*_eps) {
+			if (MathLib::sqrDist(*p, *pnt) <= _eps*_eps) {
 				ret_pnt = p;
 				return false;
 			}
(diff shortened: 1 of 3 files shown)


GeoLib/OctTree-impl.h
```diff
@@ -40,7 +40,16 @@ OctTree<POINT, MAX_POINTS> OctTree<POINT, MAX_POINTS>::createOctTree(T ll, T ur,
 			ur[1] += (dz-dy)/2.0;
 		}
 	}
-	return OctTree<POINT, MAX_POINTS>(ll, ur, eps);
+	if (eps == 0.0)
+		eps = std::numeric_limits<double>::epsilon();
+	for (std::size_t k(0); k<3; ++k) {
+		if (ur[k] - ll[k] > 0.0) {
+			ur[k] += (ur[k] - ll[k]) * 1e-6;
+		} else {
```
norihiro-w: do you mean there is a negative case `ur[k] - ll[k] < 0.0`? or only `ur[k] - ll[k] == 0.0`

norihiro-w:
:+1:

GeoLib/OctTree-impl.h
TomFischer: `ur[k] - ll[k] < 0.0` must not happen.

norihiro-w: thanks

PR #721, opened 2015-06-22
GeoLib: Rework PointVec

TomFischer:
The main change in this PR is the substitution of the method `GeoLib::PointVec::makePntsUnique()` with an `OctTree` based algorithm. This substitution fixes some mistakes that are nearly unavoidable using the old implementation. Furthermore it is much faster to use `GeoLib::PointVec::push_back()` methods. The PR is based on PR https://github.com/ufz/ogs/pull/720 and shall be merged after https://github.com/ufz/ogs/pull/720.

### DIFF
--- a/GeoLib/PointVec.cpp
+++ b/GeoLib/PointVec.cpp
@@ -33,23 +33,44 @@ PointVec::PointVec (const std::string& name, std::vector<Point*>* points,
 	TemplateVec<Point> (name, points, name_id_map),
 	_type(type),
 	_aabb(points->begin(), points->end()),
-	_rel_eps(rel_eps)
+	_rel_eps(rel_eps*sqrt(MathLib::sqrDist(_aabb.getMinPoint(), _aabb.getMaxPoint()))),
+	_oct_tree(OctTree<GeoLib::Point, 16>::createOctTree(
+		_aabb.getMinPoint(), _aabb.getMaxPoint(), _rel_eps))
 {
 	assert (_data_vec);
-	std::size_t const number_of_all_input_pnts (_data_vec->size());
+	std::size_t const number_of_all_input_pnts(_data_vec->size());
 
-	_rel_eps *= sqrt(MathLib::sqrDist (_aabb.getMinPoint(),_aabb.getMaxPoint()));
-	makePntsUnique (_data_vec, _pnt_id_map, _rel_eps);
+	// add all points in the oct tree in order to make them unique
+	_pnt_id_map.resize(number_of_all_input_pnts);
+	std::iota(_pnt_id_map.begin(), _pnt_id_map.end(), 0);
+	GeoLib::Point * ret_pnt(nullptr);
+	std::size_t cnt(0);
+	for (std::size_t k(0); k<_data_vec->size(); ++k) {
+		GeoLib::Point *const pnt((*_data_vec)[k]);
+		if (! _oct_tree->addPoint(pnt, ret_pnt)) {
+			_pnt_id_map[pnt->getID()] = ret_pnt->getID();
+			delete (*_data_vec)[k];
+			(*_data_vec)[k] = nullptr;
+		} else {
+			_pnt_id_map[k] = cnt;
+			cnt++;
+		}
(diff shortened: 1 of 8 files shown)


GeoLib/PointVec.cpp
```diff
@@ -33,23 +33,55 @@ PointVec::PointVec (const std::string& name, std::vector<Point*>* points,
...
+	std::iota(_pnt_id_map.begin(), _pnt_id_map.end(), 0);
+	GeoLib::Point * ret_pnt(nullptr);
+	std::size_t cnt(0);
+	for (std::size_t k(0); k<_data_vec->size(); ++k) {
+		GeoLib::Point *const pnt((*_data_vec)[k]);
+		if (! _oct_tree->addPoint(pnt, ret_pnt)) {
+			DBUG("Point %d (%f,%f,%f) *not* inserted since it is equal to "
+				"point %d (%f,%f,%f).", pnt->getID(), (*pnt)[0], (*pnt)[1], (*pnt)[2],
+				ret_pnt->getID(), (*ret_pnt)[0], (*ret_pnt)[1], (*ret_pnt)[2]);
+			DBUG("Removing point %d", pnt->getID());
```
endJunction: The two DBUG messages are very noisy when running tests. Please remove or comment.
Same in line 149ff.

GeoLib/PointVec.cpp
```diff
@@ -85,102 +117,40 @@ void PointVec::push_back (Point* pnt, std::string const*const name)
...
+		for (std::size_t k(0); k<_data_vec->size(); ++k) {
+			GeoLib::Point *const p((*_data_vec)[k]);
+			_oct_tree->addPoint(p, ret_pnt);
 		}
+		// add the new point
+		ret_pnt = nullptr;
+		_oct_tree->addPoint(pnt, ret_pnt);
+		pnt->setID(_data_vec->size());
+		_data_vec->push_back(pnt);
+		return _data_vec->size()-1;
```
endJunction: Not sure if critical, but the return value could cause underflow for _data_vec->size() == 0.

TomFischer: This can not happen since in line 143 something is pushed into the vector `_data_vec`.

GeoLib/PointVec.cpp
TomFischer: Ok, will do this tomorrow.

endJunction:
Looks good! :fast_forward:

GeoLib/PointVec.cpp
```diff
@@ -33,23 +33,51 @@ PointVec::PointVec (const std::string& name, std::vector<Point*>* points,
...
 	assert (_data_vec);
-	std::size_t const number_of_all_input_pnts (_data_vec->size());
+	std::size_t const number_of_all_input_pnts(_data_vec->size());
 
-	_rel_eps *= sqrt(MathLib::sqrDist (_aabb.getMinPoint(),_aabb.getMaxPoint()));
-	makePntsUnique (_data_vec, _pnt_id_map, _rel_eps);
+	// correct the ids if necessary
+	for (std::size_t k(0); k<_data_vec->size(); ++k) {
+		if ((*_data_vec)[k]->getID() == std::numeric_limits<std::size_t>::max()) {
+			(*_data_vec)[k]->setID(k);
```
norihiro-w: it looks you assume a point ID should be an index in `_data_vec`. then it might be safer if you check `if ((*_data_vec)[k]->getID() == k)`? otherwise, there could be points having the same ID

norihiro-w:
:+1: besides one minor comment

GeoLib/PointVec.cpp
TomFischer: @norihiro-w You are right with your assumption that the point index is or at least should be a position in the `_data_vec`. In most cases the index of the point in position k in `_data_vec` should be k, too. I wonder if this is a requirement in the gli or gml files?

In my point of view the index of a point should be a unique identifier within a geometrical set.

The code in line 45 should correct the ids if the user did not specify them.

PR #723, opened 2015-06-25
add a test checking Eigen behavior

norihiro-w:
This PR adds a test checking Eigen behavior for matrix multiplication with `auto` keyword in order to track any future change in Eigen.  

A problem I found with version 3.2.0 and 3.2.5 is that the following two expressions with `auto` keywords result in different solution. 

```
auto result = c * (m * r);
```

```
auto v = m*r;
auto result = c *v;
```

A result of the first expression is wrong as far as I checked. If I explicitly set a type of `result` instead of `auto`, both provide the same correct result.

Note: `NaturalCoordinatesMapping::computeShapeMatrices()` should be modified as it uses the first expression with `auto` keyword.

### DIFF
--- a/Tests/NumLib/TestEigen.cpp
+++ b/Tests/NumLib/TestEigen.cpp
@@ -1,0 +1,60 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2015, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#include <gtest/gtest.h>
+
+#ifdef OGS_USE_EIGEN
+#include <Eigen/Eigen>
+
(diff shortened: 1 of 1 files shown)


norihiro-w:
Jenkins failed on Mac and Win, but worked on Linux...

endJunction:
Checked on archlinux with gcc-5.1.0 and clang-3.6.1 with eigen-3.2.4 and eigen-3.2-develop-r6545.
The early version of eigen produces incorrect results independent of compiler.
The developer version is producing correct results independent of compiler.

clang address sanitizer finds an error
```
SUMMARY: AddressSanitizer: heap-use-after-free
/usr/include/eigen3/Eigen/src/Core/products/GeneralBlockPanelKernel.h:1187 
Eigen::internal::gemm_pack_lhs<double, long, 4, 2, 1, false, false>::operator()(...)
```

Probably related to the bug http://eigen.tuxfamily.org/bz/show_bug.cgi?id=825

norihiro-w:
close with #726

PR #726, opened 2015-06-25
fix Eigen issue with auto keyword in NaturalCoordinatesMapping

norihiro-w:
This PR fixes the issue reported in #723 together with a test

### DIFF
--- a/NumLib/Fem/CoordinatesMapping/NaturalCoordinatesMapping-impl.h
+++ b/NumLib/Fem/CoordinatesMapping/NaturalCoordinatesMapping-impl.h
@@ -120,7 +120,8 @@ inline void computeMappingMatrices(
             shapemat.dNdx.topLeftCorner(ele_dim, nnodes) = shapemat.invJ * shapemat.dNdr;
         } else {
             auto const& matR = ele_local_coord.getRotationMatrixToGlobal(); // 3 x 3
-            auto dshape_global = matR.topLeftCorner(3u, ele_dim) * (shapemat.invJ * shapemat.dNdr); //3 x nnodes
+            auto invJ_dNdr = shapemat.invJ * shapemat.dNdr;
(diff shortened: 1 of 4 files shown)


NumLib/Fem/CoordinatesMapping/NaturalCoordinatesMapping-impl.h
```diff
@@ -120,7 +120,8 @@ inline void computeMappingMatrices(
             shapemat.dNdx.topLeftCorner(ele_dim, nnodes) = shapemat.invJ * shapemat.dNdr;
         } else {
             auto const& matR = ele_local_coord.getRotationMatrixToGlobal(); // 3 x 3
-            auto dshape_global = matR.topLeftCorner(3u, ele_dim) * (shapemat.invJ * shapemat.dNdr); //3 x nnodes
+            auto invJ_dNdr = shapemat.invJ * shapemat.dNdr;
```
TomFischer: Can `invJ_dNdr` be `const`? If yes, I am unsure if the `auto`-type is already const?

endJunction: The result of eigen's operations are not always a resulting matrix or vector but an expression for that operation holding the operands.
I don't know what optimization will be inhibited by extra constness.

Anyway, as soon as next eigen release fixes this (and this is already done in dev branch) this is going to the normal expression x=a*b*c;

endJunction:
:fast_forward:

NumLib/Fem/CoordinatesMapping/NaturalCoordinatesMapping-impl.h
TomFischer: Thanks for explanation. So if I understand this correctly this is only a bug fix for the eigen bug? If so feel free to merge.

PR #730, opened 2015-06-29
Cleanup includes fwd decls

endJunction:
- One large commit, unfortunately, forcing full include paths in FileIO, which also cleanups them and use forward declarations where possible.
 - Remove unused zLibDataCompressor.
 - Some include related changes here and there.

I tried to use same scheme for the includes:
 - class declarations (if in the cpp file)
 - system (stl and alike includes)
 - our third party lib includes
 - alphabetically BaseLib, GeoLib etc.
 - other local includes.

~~The compilation times did not change unfortunately.~~
__Update:__
With the second batch of changes there compile time is reduced by ~4% relative to the ufz/master.


FileIO/FEFLOWInterface.cpp
```diff
@@ -9,24 +9,25 @@
...
+#include <logog/include/logog.hpp>
+
+#include "BaseLib/FileTools.h"
+#include "BaseLib/StringTools.h"
+
+#include "GeoLib/GEOObjects.h"
 #include "GeoLib/Point.h"
+#include "GeoLib/Polygon.h"
+
+#include "MeshLib/Elements/Hex.h"
```
wenqing: It might be better to put these mesh related includes, which occur multiply in different files,  in one header file.

TomFischer:
Looks good. :fast_forward:

FileIO/FEFLOWInterface.cpp
endJunction: Created an Elements.h include for this case.

norihiro-w:
:+1:

PR #731, opened 2015-07-02
Use all available shape functions

endJunction:
This makes all currently implemented shape functions available for the local assemblers.

The mapping between mesh elements and the shape functions is one-to-one.
In general the higher-order shape functions can be instantiated on the lower-order mesh elements too. Those interested in this feel free to provide another template argument to the `LocalDataInitializer` for the order of shape functions and extend the builder map.


norihiro-w:
:+1:

endJunction:
Does anybody have an idea, why there are multiple definitions in the resulting object files?
When reverting the last commit, there are no compile errors.

norihiro-w:
i have sent you a PR to fix the error.

endJunction:
Just a rebase.

PR #732, opened 2015-07-02
[MaL] Inline sqrDist(double*,...).

endJunction:
This reduces number of instructions executed in the `GeoLib::getNearestPoint()` from 1087e6 to 707e6 or around 34% (measured in SearchNearestPointsInDenseGrid test), which is not so bad for a single inline :)

The reason for this is that the `sqrDist` computation is around 20 instructions, but its call costs in order of 100 instructions. Use the assembly, Luke!


norihiro-w:
:+1:

PR #733, opened 2015-07-03
[T/GL] Fix PolygonTest.containsSegment.

TomFischer:
In the previous version the variables a and b were references initialized
to the point 2 and 6 respectively. Later on, the point 2 and 6 were modified.
This modification was not intended by the test and leads to wrong testing.


endJunction:
:+1: The division by zero disappeared!

Addendum to the description: This is a first bug in the series of undefined behaviour bugs as ufz/ogs#734 and ufz/ogs#735.

norihiro-w:
:+1:

PR #737, opened 2015-07-03
introduce GeoDomain class

norihiro-w:
I would like to propose adding `GeoDomain` class as a member of `GeoObject`. With the class, we can provide consistent interface for any geometric definition in some functions such as `MeshNodeSearcher::getMeshNodeIDs()`.

### DIFF
--- a/GeoLib/GEOObjects.cpp
+++ b/GeoLib/GEOObjects.cpp
@@ -522,6 +522,10 @@ const GeoLib::GeoObject* GEOObjects::getGeoObject(const std::string &geo_name,
 							sfc_vec->getElementByName(geo_obj_name)));
 		break;
 	}
+	case GeoLib::GEOTYPE::GEODOMAIN: {
+		geo_obj = const_cast<GeoLib::GeoDomain*>(&_geo_domain);
+		break;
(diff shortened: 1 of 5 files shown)


endJunction:
Formally OK.
@norihiro-w  Actually I didn't get the explanation. Could you explain it in different words (or example), please?

norihiro-w:
let me try... Inconsistency comes from that geo type GEODOMAIN doesn't have its corresponding class while other geo types have them. With current code, for example we need to write it like

```
if (geo_type == GEOTYPE::GEODOMAIN) {
    vec_node_ids = ...; // no geoObj can be geo domain
} else {
    vec_node_ids = MeshNodeSearcher::getMeshNodeIDs(geoObj);
}
```

which can be shortened by having GeoDomain class

```
vec_node_ids = MeshNodeSearcher::getMeshNodeIDs(geoObj);
```

bilke:
:+1:

endJunction:
@norihiro-w I'd merge it after a rebase.

endJunction:
Merged, but have to manually close it, strange, isn't it?

PR #739, opened 2015-07-05
[NUM] add another version of  reset() in ShapeMatrices 

norihiro-w:
as titled. new reset() takes golobal dim


endJunction:
:fast_forward:

norihiro-w:
First of all, this change is aimed at fixing inconsistency caused by #702 where the new constructor taking `global_dim` has been added. 

At least for Eigen, the check is done within `resize()` of its matrix class. 

I don't think `resize()` is not used in current AssemblerLib. Originally the feature was implemented for the case that one prefers recalculating shape matrices for each element instead of storing shape matrices for every elements.

TomFischer:
Thanks for explanation.

TomFischer:
@norihiro-w 
> At least for Eigen, the check is done within resize() of its matrix class.

In our code the check `local_dim == dim` has to be performed only once, i.e. we could omit 4 Eigen checks (3 check total) in case `local_dim` equals `dim`. Contrary to this if the dimension really changes we have one additional check. What frequency of occurence of `local_dim != dim` do you expect?

TomFischer:
I discussed this with @endJunction . Now I understand and I think the solution is ok.

PR #740, opened 2015-07-06
Cleanup

TomFischer:
PR covers cleanup of
- whitespaces,
- documentation,
- const correctness,
- using `std::size_t` instead of `size_t` and
- using `std::abs` instead of `std::fabs`.


norihiro-w:
:+1:

PR #741, opened 2015-07-06
[A/U] GMSH2OGS: Add nullptr check.

TomFischer:
@waltherm discovered a bug in the GMSH2OGS tool. Using a switch it is possible that line elements are not converted to the OGS mesh. In case the GMSH mesh does not contain line elements and the switch is activated the previous code crashed. The changes in this PR/commit will fix this.


norihiro-w:
:+1:

PR #743, opened 2015-07-06
[NL] Fix possible alignment issues with eigen.

endJunction:
See http://eigen.tuxfamily.org/dox/group__TopicStructHavingEigenMembers.html for details.


norihiro-w:
looks good.

PR #746, opened 2015-07-06
Fix memory leaks in tests

endJunction:
Following ufz/ogs#745.

This removes all remaining memory leaks detected by clang's address-sanitizer in testrunner (caused by ogs).
There are few remaining but their traces lead to Qt.

One real memory leak (which would occur when using MeshSurfaceExtraction) is detected.
Other memory leaks are due to careless usage in the tests.


TomFischer:
Good work! :beer: :+1:

norihiro-w:
:+1:

Issue #734, opened 2015-07-03
Division by zero in CreatePolygonTreesTest.P0AndP1


endJunction:
Found with undefined behaviour sanitzer:

```
 bin/testrunner --gtest_filter=CreatePolygonTreesTest.P0AndP1           
Note: Google Test filter = CreatePolygonTreesTest.P0AndP1
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from CreatePolygonTreesTest
[ RUN      ] CreatePolygonTreesTest.P0AndP1
/home/naumov/w/ogs/GeoLib/../MathLib/LinAlg/Solvers/TriangularSolve-impl.h:44:19: runtime error: division by zero
[       OK ] CreatePolygonTreesTest.P0AndP1 (2 ms)
[----------] 1 test from CreatePolygonTreesTest (3 ms total)
```

Backtrace under gtest code:
```
(gdb) bt
#0  MathLib::backwardSolve<double, double [2]> (mat=..., b=...) at /home/naumov/w/ogs/GeoLib/../MathLib/LinAlg/Solvers/TriangularSolve-impl.h:44
#1  0x0000000002801ae4 in MathLib::GaussAlgorithm<MathLib::DenseMatrix<double, unsigned long>, double*>::solve<double [2]> (this=0x7fffffffc738, b=..., decompose=true)
    at /home/naumov/w/ogs/GeoLib/../MathLib/LinAlg/Solvers/GaussAlgorithm-impl.h:76
#2  0x0000000002808d27 in GeoLib::lineSegmentIntersect (a=..., b=..., c=..., d=..., s=...) at /home/naumov/w/ogs/GeoLib/AnalyticalGeometry.cpp:150
#3  0x000000000268c856 in GeoLib::Polygon::getAllIntersectionPoints (this=0x5ccf970, a=..., b=...) at /home/naumov/w/ogs/GeoLib/Polygon.cpp:123
#4  0x000000000268ce5e in GeoLib::Polygon::containsSegment (this=0x5ccf970, a=..., b=...) at /home/naumov/w/ogs/GeoLib/Polygon.cpp:133
#5  0x000000000269260f in GeoLib::Polygon::isPolylineInPolygon (this=0x5ccf970, ply=...) at /home/naumov/w/ogs/GeoLib/Polygon.cpp:189
[...]

PR #747, opened 2015-07-07
[GL] Fix bug in lineSegmentIntersect.

TomFischer:
The special case where the line segments (a,b) and (c,d) are parallel but are not located at the same straight line was not handled properly. This led to a singular matrix and to a not a number solution. This commit fixes the handling of the special case.

This PR is related to issue https://github.com/ufz/ogs/issues/734.

fixes ufz/ogs#734


endJunction:
Can you add a test for this, because the current tests do not directly cover this issue.

norihiro-w:
:+1:

PR #748, opened 2015-07-07
[A] Initialize lis.

endJunction:
The lis initialization was forgotten in the ogs.cpp and resulted in failing simulations for openmp lis configuration.


norihiro-w:
:+1:

Issue #728, opened 2015-06-25
Memory leak in DenseMatrix::operator*(FP_TYPE*).


endJunction:
When creating a `MathLib::Vector3` directly from the result of a `DenseMatrix` and double pointer as in Tests/GeoLib/TestComputeRotationMatrix.cpp:67 the result values allocated in
MathLib/LinAlg/Dense/DenseMatrix-impl.h:123 are not deallocated.

backtrace from clang's address sanitizer.
```
Direct leak of 24 byte(s) in 1 object(s) allocated from
    #1 0xa8127a in MathLib::DenseMatrix<double, unsigned long>::operator*(double const* const&) const /home/naumov/w/ogs/MathLib/LinAlg/Dense/DenseMatrix-impl.h:123
    #2 0xa7f084 in GeoLib_ComputeRotationMatrixToXYpositive_Test::TestBody() /home/naumov/w/ogs/Tests/GeoLib/TestComputeRotationMatrix.cpp:67
```

For reference see the following version of ufz/master [DenseMatrix-impl.h](https://github.com/ufz/ogs/blob/0e3affe5f291c3f209a4a7359165899c6b654e16/MathLib/LinAlg/Dense/DenseMatrix-impl.h#L123)

endJunction:
The issue is resolved by removing Vector3(double*) ctor. Now such code (with hidden memory leak) is not compiling:
```
Tests/GeoLib/TestComputeRotationMatrix.cpp:79:25: error: no matching constructor for initialization of 'const MathLib::Vector3' (aka 'const TemplateVector3<double>')
        MathLib::Vector3 const r1(rot_mat*x1.getCoords());
...
```

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/728).

PR #749, opened 2015-07-08
Using other Point constructors

TomFischer:
The reason for this PR is to make OGS more secure and to avoid the creation of unnecessary temporary objects.
Main changes:
- Generalized one constructor of class `GeoLib::Point` (commit f3f0b13).
- Avoid using potential insecure Point constructor (see https://github.com/TomFischer/ogs/blob/master/GeoLib/Point.h#L46-49). This constructor is potential insecure because there is not a check if the length of the given `x` is sufficient.
- Creating `Point3d` objects instead of `double [3]` fields at some places.

**Update**
Forgot to mention: Removing the insecure constructor is in preparation.


endJunction:
:+1:

Info: This series of changes by Tom aims to prohibit memory  leaks caused by some constructors taking a pointer to data and copying this data like the one described in ufz/ogs#728

norihiro-w:
:+1:

PR #750, opened 2015-07-08
Improve point type constructors

TomFischer:
The reason for this PR is the same as for PR #749, i.e. to make OGS more secure and to avoid the creation of unnecessary temporary objects. The PR is based on PR #749 and should be merged after 749 is merged.

The main point is the introduction of the keyword `explicit` in `TemplateVector` constructors (commit 14e2702) and removing the potential insecure constructor (commit 93741af).


norihiro-w:
:+1:

bilke:
:+1:

PR #751, opened 2015-07-08
Improve TemplatePoint::setCoords()

TomFischer:
Small PR based on PR #750. The data type given to `TemplatePoint::setCoords()` can be checked at compile time. Only commit fb67b52 and 167bbd8 needs a review.


TomFischer:
@norihiro-w remarked that it is possible to remove `setCoords()`. I changed the PR according to the remark.

norihiro-w:
:+1:

PR #752, opened 2015-07-08
Generalized interfaces to avoid temporaries.

TomFischer:
PR is based on PR #750.
- It generalizes the interfaces in classes `Surface`, `Triangle` and `Grid`such that some of the methods accept `MathLib::Point3d` instead of `GeoLib::Point`. This avoids creating temporary objects.
- Because of the changes of 1. it is possible to remove the potential insecure constructor in class `GeoLib::Point` (commit eaf7f10).
- At the moment there is no need to have a templated version of `sqrDist()`. For this reason the template version is substituted with a version that takes to `MathLib::Point3d` objects. `MathLib::Point3d` is the base class of most of the in OGS used point types.

### DIFF
--- a/GeoLib/Grid.h
+++ b/GeoLib/Grid.h
@@ -131,7 +131,8 @@ public:
 	 * @return a pointer to the point with the smallest distance within the grid cells that are
 	 * outlined above or nullptr
 	 */
-	POINT* getNearestPoint(POINT const& pnt) const
+	template <typename P>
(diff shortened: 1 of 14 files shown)


norihiro-w:
:+1:

endJunction:
@TomFischer Thank you!

I'll check if the ufz/ogs#/728 is resolved now.

PR #754, opened 2015-07-08
Optimize compile time

endJunction:
just testing this on all platforms.
Not for merging.


endJunction:
Result is that measuring of compile time is not straight forward. On jenkins the time for linux pr compilation varies from 10min to over 1h. Because optimizing for compile time increases code complexity sometimes, I try to include only changes which are relevant for compile time.
Thank you for your attention ;)

PR #755, opened 2015-07-09
Add Eigen sparse solvers

norihiro-w:
This PR adds 
- Global matrix, vector, linear solver based on Eigen
- EigenLis linear solver which uses Eigen for matrix and vector objects but Lis for linear solve

For those linear solvers, imposing of prescribed values is done by modification of matrix entries (exact), not by the penalty method.


MathLib/LinAlg/ApplyKnownSolution.h
```diff
@@ -13,6 +13,10 @@
 
 #include "MathLib/LinAlg/Dense/DenseTools.h"
 
+#ifdef OGS_USE_EIGEN
+#include "MathLib/LinAlg/Eigen/EigenTools.h"
+#endif // USE_LIS
```
endJunction: Change comment to USE_EIGEN, or drop it and the following one too.

MathLib/LinAlg/Eigen/EigenLinearSolver.cpp
```diff
@@ -0,0 +1,145 @@
...
+public:
+    EigenDirectSolver(EigenMatrix::RawMatrixType &A) : _A(A)
+    {
+        // fill A and b;
+        // Compute the ordering permutation vector from the structural pattern of A
+        _solver.analyzePattern(A);
+        // Compute the numerical factorization
+        _solver.factorize(A);
+        if(_solver.info()!=Eigen::Success) {
+            INFO("\t failed");
```
endJunction: ERR, or is the failure handled somehow?
I even think assert is better here.

MathLib/LinAlg/Eigen/EigenLinearSolver.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+    void solve(EigenVector::RawVectorType &b, EigenVector::RawVectorType &x, EigenOption &) override
+    {
+        //Use the factors to solve the linear system
+        INFO("-> solve");
+        x = _solver.solve(b);
+    }
+
+private:
+    T_SOLVER _solver;
```
endJunction: Is EigenIterativeSolver owning the _solver or just using it?
Precaution to avoid copies of _solver.

MathLib/LinAlg/Eigen/EigenLinearSolver.h
```diff
@@ -0,0 +1,80 @@
...
+ *
+ */
+
+#ifndef EIGENLINEARSOLVER_H_
+#define EIGENLINEARSOLVER_H_
+
+#include <vector>
+
+#include <boost/any.hpp>
+#include <boost/property_tree/ptree.hpp>
```
endJunction: ptree_fwd.hpp should be sufficient here. (Compile time minimization).

MathLib/LinAlg/Eigen/EigenLinearSolver.h
```diff
@@ -0,0 +1,80 @@
...
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef EIGENLINEARSOLVER_H_
+#define EIGENLINEARSOLVER_H_
+
+#include <vector>
+
+#include <boost/any.hpp>
```
endJunction: I don't see where boost::any is used, and the std::vector too. If unused -> remove.

MathLib/LinAlg/Eigen/EigenMatrix.h
```diff
@@ -0,0 +1,170 @@
...
+    /// y = mat * x
+    virtual void multiply(const EigenVector &x, EigenVector &y) const
+    {
+        y.getRawVector() = _mat * x.getRawVector();
+    }
+
+    /// return if this matrix is already assembled or not
+    bool isAssembled() const { return true; }
+
+    /// printout this matrix for debugging
```
endJunction: If used only in debug build please add NDEBUG around both writes. (And then probably the iostream include also goes into NDEBUG condition.)

MathLib/LinAlg/Eigen/EigenTools.cpp
```diff
@@ -0,0 +1,58 @@
...
+namespace MathLib
+{
+
+void applyKnownSolution(EigenMatrix &A_, EigenVector &b_, const std::vector<std::size_t> &vec_knownX_id,
+		const std::vector<double> &vec_knownX_x, double /*penalty_scaling*/)
+{
+    typedef EigenMatrix::RawMatrixType SpMat;
+    auto &A = A_.getRawMatrix();
+    auto &b = b_.getRawVector();
+    const size_t n_rows = A.rows();
```
endJunction: auto const or std::size_t

There are some more of size_t's below in the for-loops
A very safe way is to write
```c++
for (auto i = decltype(n_rows){0}; i < n_rows; ++i)
```
because there will be no conversion by definition. (but it's a little verbose)

MathLib/LinAlg/Eigen/EigenVector.h
```diff
@@ -0,0 +1,114 @@
...
+    template<class T_SUBVEC>
+    void add(const std::vector<std::size_t> &pos, const T_SUBVEC &sub_vec)
+    {
+        for (std::size_t i=0; i<pos.size(); ++i) {
+            this->add(pos[i], sub_vec[i]);
+        }
+    }
+
+    /// printout this equation for debugging
+    void write (const std::string &filename) const { std::ofstream os(filename); os << _vec; }
```
endJunction: NDEBUG here and for fstream include, if you decide to make it in debug mode available only.

MathLib/LinAlg/EigenLis/EigenLisLinearSolver.h
```diff
@@ -0,0 +1,76 @@
...
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef EIGENLISLINEARSOLVER_H_
+#define EIGENLISLINEARSOLVER_H_
+
+#include <vector>
+
+#include <boost/property_tree/ptree.hpp>
```
endJunction: ptree_fwd here, and ptree (and probably optional) in cpp

MathLib/LinAlg/Eigen/EigenMatrix.h
```diff
@@ -0,0 +1,170 @@
...
+    /// get a maximum value in diagonal entries
+    virtual double getMaxDiagCoeff() {return 1;}
+
+    /// y = mat * x
+    virtual void multiply(const EigenVector &x, EigenVector &y) const
+    {
+        y.getRawVector() = _mat * x.getRawVector();
+    }
+
+    /// return if this matrix is already assembled or not
```
wenqing: Seems this function does nothing because it always returns true.

MathLib/LinAlg/Eigen/EigenLinearSolver.cpp
norihiro-w: _solver is initialized in the constructor and used in solve(). am not sure what you mean "avoid of copies"

MathLib/LinAlg/Eigen/EigenMatrix.h
norihiro-w: this function is a part of the matrix interface and needed for petsc and lis matrices.

MathLib/LinAlg/Eigen/EigenMatrix.h
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    /// get value
+    double operator() (std::size_t row, std::size_t col) const
+    {
+        return _mat.coeff(row, col);
+    }
+
+    /// get a maximum value in diagonal entries
+    virtual double getMaxDiagCoeff() {return 1;}
```
TomFischer: I wonder if the implementation of `getMaxDiagCoeff()` is correct?

MathLib/LinAlg/Eigen/EigenMatrix.h
```diff
@@ -0,0 +1,173 @@
...
+        std::ofstream of(filename);
+        if (of)
+            write(of);
+    }
+
+    /// printout this matrix for debugging
+    void write(std::ostream &os) const
+    {
+        for (int k=0; k<_mat.outerSize(); ++k)
+          for (Eigen::SparseMatrix<double>::InnerIterator it(_mat,k); it; ++it)
```
TomFischer: The condition of this for loop looks a little bit wired to me. Can you explain why this works?

MathLib/LinAlg/Eigen/EigenMatrix.h
```diff
@@ -0,0 +1,173 @@
...
+    /// printout this matrix for debugging
+    void write(const std::string &filename) const
+    {
+        std::ofstream of(filename);
+        if (of)
+            write(of);
+    }
+
+    /// printout this matrix for debugging
+    void write(std::ostream &os) const
```
TomFischer: Maybe it is also possible to offer binary output in standard sparse matrix formats like compressed row storage (CRS). This would give the user the possiblity the use other software, for instance Matlab, to inspect the matrix entries.

MathLib/LinAlg/Eigen/EigenMatrix.h
```diff
@@ -0,0 +1,173 @@
...
+    /// Add sub-matrix at positions given by \c indices.
+    template<class T_DENSE_MATRIX>
+    void add(RowColumnIndices<std::size_t> const& indices,
+            const T_DENSE_MATRIX &sub_matrix,
+            double fkt = 1.0)
+    {
+        this->add(indices.rows, indices.columns, sub_matrix, fkt);
+    }
+
+    ///
```
TomFischer: Documentation is missing. Do you assume that `row_pos.size() == sub_matrix.getNRows()` and `col_pos.size() == sub_matrix.getNCols()`? Maybe it is also worth to mention that the entries of `row_pos` and `col_pos` should be valid entries with in the sparse matrix. What happens if not?

MathLib/LinAlg/Eigen/EigenOption.h
```diff
@@ -0,0 +1,82 @@
...
+struct EigenOption
+{
+    /// Solver type
+    enum class SolverType : int
+    {
+        INVALID,
+        CG,
+        BiCGSTAB,
+        SparseLU
+    };
```
TomFischer: Is it right that we have to update this enum if new solvers added to eigen? Is it possible to import such a list from eigen directly?

MathLib/LinAlg/Eigen/EigenOption.h
```diff
@@ -0,0 +1,82 @@
...
+
+    /// Preconditioner type
+    enum class PreconType : int
+    {
+        NONE,
+        DIAGONAL
+    };
```
TomFischer: Some as above: Is it right that we have to update this enum if new preconditioners added to eigen? Is it possible to import such a list from eigen directly?

MathLib/LinAlg/Eigen/EigenOption.h
```diff
@@ -0,0 +1,82 @@
...
+
+    /// Linear solver type
+    SolverType solver_type;
+    /// Preconditioner type
+    PreconType precon_type;
+    /// Maximum iteration count
+    long max_iterations;
```
TomFischer: Using `long` for really difficult problems? :smile:

MathLib/LinAlg/Eigen/EigenTools.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+#include <logog/include/logog.hpp>
+
+#include "EigenMatrix.h"
+#include "EigenVector.h"
+
+namespace MathLib
+{
+
+void applyKnownSolution(EigenMatrix &A_, EigenVector &b_, const std::vector<std::size_t> &vec_knownX_id,
```
TomFischer: Is this the method used for applying Dirichlet type conditions?

MathLib/LinAlg/Eigen/EigenMatrix.h
norihiro-w: `InnerIterator` has the following `operator bool()` function so that you can write it as `if (it)`

```
    inline operator bool() const { return (m_id < m_end); }
```
see http://www.artima.com/cppsource/safebool.html for  `operator bool()`

MathLib/LinAlg/Eigen/EigenMatrix.h
norihiro-w: :white_check_mark:

MathLib/LinAlg/Eigen/EigenOption.h
norihiro-w: no, eigen doesn't provide such a list

MathLib/LinAlg/Eigen/EigenOption.h
norihiro-w: see above

MathLib/LinAlg/Eigen/EigenTools.cpp
norihiro-w: yes.

MathLib/LinAlg/Eigen/EigenMatrix.h
norihiro-w: it would be better if there is such a function, but currently I don't find no need to have the function.  someone wants to use it can implement it later.

MathLib/LinAlg/Eigen/EigenOption.h
endJunction: There are few more solvers available http://eigen.tuxfamily.org/dox/group__TopicSparseSystems.html
Any reasons not to include all of them in the SolverType?

MathLib/LinAlg/Eigen/EigenMatrix.h
```diff
@@ -0,0 +1,190 @@
...
+public:
+    typedef Eigen::SparseMatrix<double, Eigen::RowMajor> RawMatrixType;
+
+    /**
+     * constructor
+     * @param n_rows the number of rows (that is equal to the number of columns)
+     */
+    explicit EigenMatrix(std::size_t n) :_mat(n, n) {}
+
+    virtual ~EigenMatrix() {}
```
endJunction: Will we ever derive from the EigenMatrix class?
If not, the virtual specifier can be removed. Additionally I recommend to use the `final` specifier in class definition and `override` specifier for all derived functions. This can prevent some types of mistakes.

MathLib/LinAlg/Eigen/EigenOption.h
norihiro-w: Technically there is no reason for it. I just don't need others at the moment (BCGSTAB and SparseLU solve most of problems and are enough for testing). You can extend it later if you need them. There are a few more solvers available in Eigen/unsupported such as GMRES.

endJunction:
when compiling on my platform with clang I get a compilation error:
```
Building CXX object MathLib/CMakeFiles/MathLib.dir/LinAlg/Eigen/EigenLinearSolver.cpp.o
/home/naumov/w/ogs/MathLib/LinAlg/Eigen/EigenLinearSolver.cpp:29:59: error: 'IEigenSolver' is a protected member of 'MathLib::EigenLinearSolver'
class EigenDirectLinearSolver : public EigenLinearSolver::IEigenSolver
                                                          ^
/home/naumov/w/ogs/MathLib/LinAlg/Eigen/EigenLinearSolver.h:64:11: note: declared protected here
    class IEigenSolver
          ^
```
(eigen is 3.2.5, clang-3.6.1)

But gcc-5.1.0 is happily compiling...

endJunction:
Also when compiling LisMatrix.cpp.o fails with
```
Building CXX object MathLib/CMakeFiles/MathLib.dir/LinAlg/Lis/LisMatrix.cpp.o
/home/naumov/w/ogs/MathLib/LinAlg/Lis/LisMatrix.cpp:59:16: error: use of undeclared identifier 'lis_matrix_unset'
        ierr = lis_matrix_unset(_AA);
               ^
```

OGS_USE_LIS is set.

norihiro-w:
The first compiling error is fixed. The second one should be gone if you use Lis version 1.5.57.

endJunction:
:fast_forward: 
@norihiro-w  Thank you. Works fine.
@bilke Would you please have a quick glance on this and merge if OK?

bilke:
I will merge after #765 is merged to get green windows builds on Jenkins.

endJunction:
This was working on Linux before gcc package was broken. Later it was also checked on Windows, so both platforms are checked.
I shall merge it tonight, even if the gcc  is not repaired. Any objections?

endJunction:
Jenkins, test this please

endJunction:
@norihiro-w Could you please add the patch for GroundWaterFlowProcess to correct instantiation of eigen's linear solver. Otherwise the jenkins is stuck when running `ctest -E LARGE`.

norihiro-w:
added

bilke:
:+1:

PR #756, opened 2015-07-09
[TCLAP] add ignoreUnmatched

norihiro-w:
`ignoreUnmatched()` was copied from TCLAP codes in vtkdiff. The official TCLAP doesn't have such function but am not sure where it comes from...

Anyway, we need the function because current TCLAP throws an error if one provides any arguments which are not registered in TCLAP. For example, this is required for PETSc which is often configured via command line arguments.

### DIFF
--- a/ThirdParty/tclap/CmdLine.h
+++ b/ThirdParty/tclap/CmdLine.h
@@ -183,6 +183,11 @@ private:
 		 */
 		bool _helpAndVersion;
 
+		/**
+		 * Whether or not to ignore unmatched args.
+		 */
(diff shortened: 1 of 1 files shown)


endJunction:
:+1: 

Discussion: I see there will be more changes coming to TCLAP, which is unfortunately not evolving anymore.
Suggestion to create a ufz/TCLAP from the same source as used in ufz/vtkdiff and include it as a submodule/subtree.

bilke:
@endJunction :+1:

norihiro-w:
:+1:

PR #758, opened 2015-07-09
Remove some try catch blocks

endJunction:
This are the leftovers of my DisableExceptions branch.

__Update:__ There are no logic changes anymore compared to the original PR. Only replace `throw` with ERR() + `std::abort`.

I checked on a small code sample that removing `throw`s indeed decreases binary size (by 8 bytes :-). Unfortunately this is not the case with logog messages, which are quite expensive (much more expensive as the throw solution).

bin/ogs sizes in byte (clang, O3, march=native, stripped.) :

  | with logog | without logog | saved by removing logog 
---- | ------------ | ----------------- | ----------------------
base ufz/master | 1839296 | 1739496 | 99800
this PR | 1841984 | 1739472 | 102512
saved by this PR | -2688 | 24 | 

Removing logog means replacing DBUG, INFO, ERR macros with a std::printf calls, not removing the messages entirely.

### DIFF
--- a/GeoLib/AABB.h
+++ b/GeoLib/AABB.h
@@ -17,6 +17,7 @@
 
 #include <limits>
 #include <cstddef>
+#include <cstdlib>
 #include <iterator>
 #include <cassert>
 #include <vector>
@@ -72,9 +73,11 @@ public:
 	template <typename InputIterator>
 	AABB(InputIterator first, InputIterator last)
 	{
-		if (!(std::distance(first,last) > 0))
+		if (std::distance(first,last) <= 0)
+		{
 			ERR("AABB::AABB(InputIterator first, InputIterator last): first > last");
-		assert(std::distance(first,last) > 0);
+			std::abort();
+		}
 		init(*first);
 		InputIterator it(first);
 		while (it != last) {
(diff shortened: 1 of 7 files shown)


norihiro-w:
:+1:

GeoLib/AABB.h
```diff
@@ -70,9 +72,9 @@ class AABB
 	template <typename InputIterator>
 	AABB(InputIterator first, InputIterator last)
 	{
-		if (! (std::distance(first,last) > 0)) {
-			throw std::invalid_argument("AABB::AABB(InputIterator first, InputIterator last): first == last");
-		}
+		if (!(std::distance(first,last) > 0))
+			ERR("AABB::AABB(InputIterator first, InputIterator last): first > last");
+		assert(std::distance(first,last) > 0);
```
TomFischer: Assuming that `std::distance(first, last) > 0`, does `assert()` something useful in release mode? On the other hand if the programm is not aborted at this place, it will either crash in the init function or will run forever in the while loop starting in line 80. Contrary, the previous solution was throwing an exception in release and debug mode, i.e. the same behaviour for debug and release mode.

endJunction:
@TomFischer I restored the previous behaviour. Thank you for pointing this out.

endJunction:
The logog discussion is not related to this PR, but I thought this is a good example of how much the logog messages costs. The PR stands as it is.

PR #759, opened 2015-07-11
fix circular dependency between Node and Mesh

norihiro-w:
I encountered a circular dependency problem between Node and Mesh class, when I tried to include Node.h in Mesh.h. The problem is basically due to declaration of friend functions in Node class. In this PR, I suggest to remove those declarations and make some functions public, as I don't see a reason to make them non-public. Let me know if there is any reason for it and you have better solutions.


TomFischer:
Thank you for fixing includes. I want to remark two things:

1. `GeoLib::Point` is derived from `MathLib::Point3dWithID` and has also a protected method `setID()`. I think this could be a problem later on.
2. In `GeoLib` there are some rare cases where it is necessary to (re-) set the id outside of the constructor, for instance when reading some strange input data. But he necessity to (re-)set the ids should not occure in general. For this reason I want to keep the method `setID()` protected in `GeoLib` and suggest to remove the `setID()` from its base class `MathLib::Point3d` and to reconsider if it is really necessary to make the method in the class `MeshLib::Node` public.

MeshLib/Node.h
```diff
@@ -37,14 +31,6 @@ class Element;
  */
 class Node : public MathLib::Point3dWithID
 {
-	/* friend functions: */
-	friend bool MeshLayerMapper::layerMapping(MeshLib::Mesh &mesh, const GeoLib::Raster &raster, double noDataReplacementValue);
-	friend MeshLib::Mesh* MeshSurfaceExtraction::getMeshSurface(const MeshLib::Mesh &mesh, const MathLib::Vector3 &dir, double angle, bool keep3dMeshIds);
```
rinkk: Does surface extraction and layer mapping still work when you remove the friend declarations?

norihiro-w: yes they work because I changed non-public functions to public. let me know if you have any concern about it.

norihiro-w:
@TomFischer thank you for the comments. I'll revert the changes about setID() and declare Mesh as a friend class of Node because setID() is used only from Mesh class.

MeshLib/Node.h
```diff
@@ -79,13 +70,16 @@ class Node : public MathLib::Point3dWithID
...
 	}
 
-protected:
-	/**
-	 * Add an element the node is part of.
-	 * This method is called by Mesh::addElement(Element*), see friend definition.
-	 */
+	/// Update coordinates of a node.
+	/// This method automatically also updates the areas/volumes of all connected elements.
+	virtual void updateCoordinates(double x, double y, double z);
```
TomFischer: Before methods `updateCoordinates()` and `addElement()` where protected. Now they are in the public interface. Was this change intended?

MeshLib/Node.h
```diff
@@ -79,13 +70,16 @@ class Node : public MathLib::Point3dWithID
...
+
+	/// Add an element the node is part of.
 	void addElement(Element* elem) { _elements.push_back(elem); }
 
+	/// clear stored elements connecting to this node
+	void clearElements() { _elements.clear(); }
```
TomFischer: Since in my opinion the method should be used with special caution, I wonder if it is necessary to make this method public? This opens the possibility to delete to the elements from all places within OGS.

MeshLib/Node.h
norihiro-w: yes, is there any problem?

MeshLib/Node.h
norihiro-w: I changed it to protected.

MeshLib/Node.h
norihiro-w: changed addElement() to protected because it's called from Mesh. updateCoordinates() is still kept as public.

bilke:
Linux jobs are failing because of a broken gcc module on envinf1. I already notified the cluster guys..

endJunction:
Jenkins, test this please.

PR #760, opened 2015-07-11
fix includes in MeshLib and MeshGeoToolsLib

norihiro-w:
This PR ~~is based on #759 and~~ fixes some include issues in MeshLib and MeshGeoToolsLib, i.e. adding absolute path, forward declarations, correcting order.


endJunction:
@norihiro-w If you would make a PR with include's corrections only, it could be merged without delay. Then the 'setId' can be easier reviewed...

norihiro-w:
updated only with include's correction

TomFischer:
:+1:

Issue #679, opened 2015-05-27
DataExplorer crashes when closing the application.


TomFischer:
The destructor of `GEOObject` is never called.

endJunction:
This is new info compared to https://github.com/ufz/ogs/issues/486, where it crashed only on windows. 

rinkk:
> The DataExplorer currently crashes in Windows in GeoModels::removeGeometry() when closing the programme.
> Apparently there is no problem using clang/gcc.

I think these issues address the same bug. Closing the old one.

rinkk:
does the DE crash on linux, too, now?
is there a way to fix this (because on windows we couldn't locate the actual problem) ?

TomFischer:
Yes, the DataExplorer crashes in `GeoModel::removeGeometry` when closing the programme under linux, too. I guess some signals are emitted to call `GeoMode::removeGeometry`. Maybe later on some destructors try to remove the geometry a second time? But this is only a speculation.

endJunction:
@TomFischer Is this still the case?

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/679).

PR #761, opened 2015-07-13
Added test for instantiating ProjectData

bilke:
For the moment the crash described in #679 seems to be gone...


endJunction:
:fast_forward:

PR #762, opened 2015-07-13
change std::vector<MeshLib::Element const*> to std::vector<MeshLib::Element*>

norihiro-w:
as titled. I suggest not using `std::vector<MeshLib::Element const*>` if there is no easy way to cast it to `std::vector<MeshLib::Element*>`.


endJunction:
:+1:

PR #765, opened 2015-07-14
Removed using namespace GeoLib.

bilke:
Should fix windows builds on Jenkins.

### DIFF
--- a/FileIO/Legacy/OGSIOVer4.cpp
+++ b/FileIO/Legacy/OGSIOVer4.cpp
@@ -36,8 +36,6 @@
 
 #include "FileIO/TINInterface.h"
 
-using namespace GeoLib;
-
 namespace FileIO
 {
 namespace Legacy {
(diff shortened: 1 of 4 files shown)


TomFischer:
:+1:

What was the problem with the code before?

bilke:
I do not know what was the reason (maybe an updated CMake) but since yesterday the Jenkins Windows build includes more Windows default directories where some more non-namespaced classes (i.e. `Polygon` in `wingdi.h`) get included (see current Jenkins Windows builds). This fixes it.

TomFischer:
Thanks for explanation. So the lesson is to structure our sources into namespaces **and** avoid the keyword `using` even in cpp files?!

bilke:
I don't really know.. Maybe I can also check how these windows headers get included, because a developer should not care about things which are not related.. but also because the `GeoLib::` namespace prefix was sometimes used and sometimes not I made those changes.

bilke:
retest this please

endJunction:
@TomFischer Actually those who put something into global namespace break other's code.
Using `using namespace X` is perfectly correct (and in my opinion desirable, because of less code) in cpp files; not in header files.

When green, :fast_forward:

endJunction:
Jenkins, test this please.

endJunction:
It's green :), thanks Lars!

@norihiro-w Would you please rebase all your PRs on new ufz/master; the PRs will be tested on Windows too then. (A merge of the ufz/master into the PR is also OK, but less clean.)

norihiro-w:
@endJunction rebased

PR #767, opened 2015-07-16
Cross compiling for Windows with MXE

bilke:
This allows for cross-compilation on a Linux machine for Windows with [MXE](http://mxe.cc) which uses the MinGW-compiler (gcc).

I have tested this in a [Docker](https://www.docker.com/whatisdocker)-environment. See [this Dockerfile](https://github.com/ufz/dockerfiles/blob/master/mingw/ogs-gui/Dockerfile) for setup instructions. The corresponding docker image is not published yet. See also [this Jenkins job](https://svn.ufz.de:8443/job/OGS-6/job/Docker/job/gcc/) which utilizes this Docker infrastructure.

CMake requirement is bumped to 3.1 because of a modernized usage of the Qt4 CMake module. Also this CMake version allows for improved handling of library linking dependencies which I will address in another PR.


endJunction:
:+1:

wenqing:
:+1:

PR #768, opened 2015-07-16
Jenkins Docker workflow script for gcc cli and gui

bilke:
Thanks to the [Workflow-plugin](https://github.com/jenkinsci/workflow-plugin/blob/master/TUTORIAL.md) Jenkins jobs can now be configured via scripts similar to Travis. This has the advantage that the Jenkins config is now part of the code itself and it allows for a compact representation of the Jenkins logic.

The script compiles the cli and gui config with gcc (via a [Docker](https://www.docker.com/whatisdocker)-environment, configuration can be found [here](https://github.com/ufz/dockerfiles/tree/master/gcc)) and runs the testrunner.

See the [Jenkins job](https://svn.ufz.de:8443/job/OGS-6/job/Docker/job/gcc-scm/) and also how the [output of every step](https://svn.ufz.de:8443/job/OGS-6/job/Docker/job/gcc-scm/5/flowGraphTable/) can be examined (click on the console icons).

Already includes (commented out) instructions for cross-compiling for Windows (see also #767).

**Note:** This will not replace the current Jenkins jobs in the near future because at the moment not the full Jenkins functionality can be used in a Workflow script.

### DIFF
--- a/scripts/jenkins/gcc.groovy
+++ b/scripts/jenkins/gcc.groovy
@@ -1,0 +1,72 @@
+node('docker')
+{
+	// Checks out into subdirectory ogs
+	stage 'Checkout'
+	checkout([$class: 'GitSCM',
+		branches: [[name: '*/master']],
+		doGenerateSubmoduleConfigurations: false,
+		extensions:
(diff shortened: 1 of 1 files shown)


wenqing:
:+1:

PR #770, opened 2015-07-17
Math lib geo lib includes cleanup

endJunction:
No algorithm changes.
Moved Point3d operators into Point3d header from MathTools, and same for TemplatePoint.
Moved Color class into DataExlorer tree, where it is used.
Use consistent pi-related constants from boost.
Use std namespace.

And the usual include's cleanup: full path, remove superfluous, include missing, fwd decls.


norihiro-w:
:+1:

endJunction:
Jenkins, test this please!

bilke:
:+1:

PR #772, opened 2015-07-18
Remove operator<=() for Point3d

endJunction:
The operator is unused. There is a replacement named LessEq(), which is also taking an arbitrary epsilon into account and is used.

The operator is not consistent with the operator<() for Point3d, because of the epsilon;
The operator<() has no epsilon;

> Given a point a and a small positive variation epsilon of any of coordinates.
> Then for any epsilon > 0 "operator<(a, a+epsilon) is true". This is equivalent to
> "operator<=(a+epsilon, a) is false".
> But there exist a small epsilon>0 s.t. the "operator<=(a+epsilon, a) is true",
> which is equivalent to "operator<(a, a+epsilon) is false",
> which is contradicting the first statement.


bilke:
:+1:

norihiro-w:
:+1:

PR #775, opened 2015-07-20
Use new travis infrastructure

bilke:
Travis migrates its infrastructure to a Docker-based one, all jobs then run in a Linux container. They propose to [migrate](http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade) Travis scripts to the new infrastructure because they start almost instantly and the average build time is almost halved. Another advantage that installed dependencies such as CMake and VTK get cached between builds.

Installation of packages is now limited to a [whitelisted set](https://github.com/travis-ci/apt-package-whitelist) and I have not found a way to install PetSc. So PetSc builds are currently disabled.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,68 +1,64 @@
+sudo: false
 language: cpp
 
 compiler:
   - gcc
   - clang
 
+addons:
+  apt:
+    sources:
+      - boost-latest
+      - ubuntu-toolchain-r-test
+      # TODO: feelpp/petsc
+    packages:
+      - g++-4.8
+      - libeigen3-dev
+      - libboost1.55-dev
+      - libboost-date-time1.55-dev
+      - libboost-filesystem1.55-dev
+      - libboost-program-options1.55-dev
+      - libboost-system1.55-dev
+      - qt4-dev-tools
+      - libxt-dev
+      # TODO
+      # - libpetsc3.4.2-dev
+      # Not allowed yet:
+      # - numdiff
+      # - libshp-dev
+      # - libgeotiff-dev
+
+
+
 cache:
-  - apt
+  directories:
+    - $HOME/cmake-3.1.1-Linux-x86_64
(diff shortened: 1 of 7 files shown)


wenqing:
:+1:

.travis.yml
```diff
@@ -1,68 +1,64 @@
...
+      - libboost-date-time1.55-dev
+      - libboost-filesystem1.55-dev
+      - libboost-program-options1.55-dev
+      - libboost-system1.55-dev
+      - qt4-dev-tools
+      - libxt-dev
+      # TODO
+      # - libpetsc3.4.2-dev
+      # Not allowed yet:
+      # - numdiff
```
endJunction: numdiff, I think, is not necessary. VtkDiff will do for a while.

scripts/cmake/ThirdPartyLibVersions.cmake
```diff
@@ -1,7 +1,7 @@
 set(OGS_BOOST_URL "http://sourceforge.net/projects/boost/files/boost/1.58.0/boost_1_58_0.tar.bz2/download")
 set(OGS_BOOST_MD5 "b8839650e61e9c1c0a89f371dd475546")
 
-set(OGS_EIGEN_URL "http://bitbucket.org/eigen/eigen/get/3.2.5.tar.gz")
+set(OGS_EIGEN_URL "http://opengeosys.s3.amazonaws.com/ogs6-lib-sources/eigen-3.2.5.tar.gz")
```
endJunction: Just curious, why not bitbucket's version?

bilke: CMake internally uses curl to download which per default does not follow redirections. The BitBucket http-url gets redirected to http**s** and the CMake internal curl does not support https.

I can share S3 credentials if you want.

PR #778, opened 2015-07-22
Heterogeneous groundwater flow.

endJunction:
Add parameter options used as the hydraulic conductivity coefficient in the groundwater flow process. The parameters are of two kinds: constant value and a value read from the input mesh file.

I will readily answer any questions.

**Update:** I rewrote the PR for most parts. Now the config files are parsed in the ProjectData and in creat.*Parameter functions located in ProcessLib close to the definition of the corresponding parameter classes.

I like in the new version the placement of the config parsers.
What I'm not sure about is the usage in the fem part: I simply don't know how anybody is going to implement own parameters and worst of it I don't know the requirements for the parameters. For now it is pretty general with some overhead for the abstractions both in compile and run time.

Actually this is now only heterogeneous, not anisotropic any longer.


NumLib/Fem/ShapeMatrixPolicy.h
```diff
@@ -31,6 +31,7 @@ struct EigenFixedShapeMatrixPolicy
     using DimMatrixType = _MatrixType<ShapeFunction::DIM, ShapeFunction::DIM>;
     using GlobalDimNodalMatrixType = _MatrixType<GlobalDim, ShapeFunction::NPOINTS>;
     using GlobalDimMatrixType = _MatrixType<GlobalDim, GlobalDim>;
+    using DimVectorType = _MatrixType<ShapeFunction::DIM, ShapeFunction::DIM>;
```
norihiro-w: isn't this a matrix type?

endJunction: It should be a vector of length ShapeFunction::DIM. Will change. Thanks!

ProcessLib/GroundwaterFlowFEM.h
```diff
@@ -97,7 +117,8 @@ class LocalAssemblerData : public LocalAssemblerDataInterface<GlobalMatrix, Glob
         for (std::size_t ip(0); ip < n_integration_points; ip++) {
             auto const& sm = _shape_matrices[ip];
             auto const& wp = integration_method.getWeightedPoint(ip);
-            _localA->noalias() += sm.dNdx.transpose() * _hydraulic_conductivity *
+            _localA->noalias() += sm.dNdx.transpose() *
+                _hydraulic_conductivity().asDiagonal() *
```
wenqing: I suggestion to replace ``_hydraulic_conductivity().asDiagonal()`` with a template parameter, which then can be a scalar or tensor variable for isotropic, anisotropic or transversely isotropic/anisotropic. Under such consideration, , the computation time is saved for the case of scalar parameter.

endJunction:
Wenqing, although the computation cost (CPU time) for a scalar value versus a diagonal matrix is smaller, it has some drawbacks:
 - Using a template for this parameter is good method to reduce the runtime costs, but at the costs of compile time; for P processes and S shape functions there are P*S template instantiations of the `assemble()` function. With T different types for the parameter there will be P*S*T instantiations, which will increase the compile time further (from current ~30s for ogs.cpp.o). And the binary will be also larger, which reduces the probability for instruction cache hits.
 - Using a dynamic switch case also might be more expensive than to multiplication with a matrix in all cases.

I would recommend a benchmark comparison before any specializations.

__Question:__ Do we have any groundwater flow simulations (real projects, not test examples) where a symmetric positive defenite matrix is used?

wenqing:
@endJunction  I see. The multiplication operation times  of A*B (A, B are matrix) is roughly larger than ``(N*N)^(N*N)`` (N is the dimension of the matrix), while that of A*b (b is a constant) is ``N^N``. Since N<=3 in this material parameter, the computation efficiency could not lose much by treating isotropic material tensor as a diagonal matrix other than using a constant instead.

For the  groundwater flow simulations of real project, Agnes has a little complex model. Unfortunately, the model has a monthly recharge data to be considered. To modelling the model, the recharge data were integrated on the top surface of the domain and assigned to the nodes on the top surface as point wise source terms. So far I think ogs6 does not have a functionality such that directly assign the point wise source terms to the RHS for a modelling. In addition, Benny also has a nice model, which I think Karsten should know it.  Since these models are all linear,  their  stiffness matricies are  symmetric and positive.

Applications/ApplicationsLib/ProjectData.cpp
```diff
@@ -190,6 +193,55 @@ void ProjectData::parseProcessVariables(
...
+		{
+			INFO("ConstantParameter: %s.", name->c_str());
+			_parameters.push_back(createConstParameter(parameter_config));
+			_parameters.back()->name = *name;
+		}
+		else if (*type == "MeshProperty")
+		{
+			INFO("MeshPropertyParameter: %s", name->c_str());
+			_parameters.push_back(
+			    createMeshPropertyParameter(parameter_config, *_mesh_vec[0]));
```
TomFischer: Is it correct that properties are only assignable to the first mesh in `_mesh_vec`?

endJunction: Yes. Although there is a support for arbitrary many meshes in the simulation part, we still read only single meshes from project configuration.

ProcessLib/GroundwaterFlowFEM.h
```diff
@@ -80,7 +81,10 @@ class LocalAssemblerData : public LocalAssemblerDataInterface<GlobalMatrix, Glob
                     _shape_matrices[ip]);
         }
 
-        _hydraulic_conductivity = hydraulic_conductivity;
+        _hydraulic_conductivity = [&hydraulic_conductivity, &e]()
```
norihiro-w: why not simply store an evaluated value here instead of a function? the conductivity is a function of only an element.

ProcessLib/Parameter.h
```diff
@@ -0,0 +1,131 @@
...
+namespace MeshLib
+{
+template <typename T>
+class PropertyVector;
+}
+
+namespace ProcessLib
+{
+/// Base class for all parameters, not an interface class. This avoids using of
+/// void* when storing parameters and convinient destruction.
```
bilke: Typo: conv**e**nient

ProcessLib/GroundwaterFlowFEM.h
endJunction: With time (or other values for non-linear equations) dependency this will become a function anyway. So this is more conceptual thing here, about which I'm still not very happy.

If for a particular process a constant is needed per element, saving a constant here will save few bytes because the std::function<...> is relatively heavy object.

ProcessLib/Parameter.h
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	std::size_t tupleSize() const override
+	{
+		return 1;
+	}
+
+	std::vector<ReturnType> getTuple(MeshLib::Element const&) const override
+	{
+		return std::vector<ReturnType>(_value);
```
norihiro-w: `{_value}`?

ProcessLib/Parameter.h
```diff
@@ -0,0 +1,131 @@
...
+/// The ReturnType can represent an underlying type of an aggregate type like
+/// tuple or matrix (\see tupleSize()).
+/// The total number of stored tuples is provided.
+template <typename ReturnType, typename... Args>
+struct Parameter : public ParameterBase
+{
+	virtual ~Parameter() = default;
+
+	virtual ReturnType operator()(Args&&... args) const = 0;
+	virtual std::size_t numberOfTuples() const = 0;
```
norihiro-w: is this function used somewhere? why do we need it?

bilke:
:+1:

norihiro-w:
:+1:

ProcessLib/GroundwaterFlowFEM.h
```diff
@@ -111,7 +116,7 @@ class LocalAssemblerData : public LocalAssemblerDataInterface<GlobalMatrix, Glob
 
 private:
     std::vector<ShapeMatrices> _shape_matrices;
-    double _hydraulic_conductivity;
+    std::function<double(void)> _hydraulic_conductivity;
```
wenqing: Here we need return matrix in future development for anisotropic conductivity. Everything else looks fine.

endJunction: Definetly. There are already some functions managing the tuple size information; See the [[PL] Remove unused parameter tuple functions.](https://github.com/endJunction/ogs/commit/b313ed5c923465311b5d8f3054674560d009032b).

wenqing: I see that you already prepared a branch for the case.

wenqing: :+1:

PR #779, opened 2015-07-24
add MeshNodesOnPoint

norihiro-w:
This PR suggests adding `MeshNodesOnPoint` class which is used to search nodes near the given point for Dirichlet BC, assuming the number of points given in a geometry file is not quite large. We used `GeoLib::Grid<MeshLib::Node>` until now but it is not supporting 
 * a case when more than one nodes exist at the given location
 * a case when no node is located at the location (maybe search radius is not considered?)

it is also possible to extend GeoLib::Grid, however I don't think we need such a sophisticated algorithm to search nodes for boundary conditions because the number of geometric points we give in simulations are not so large. how do you think?

### DIFF
--- a/MeshGeoToolsLib/MeshNodeSearcher.cpp
+++ b/MeshGeoToolsLib/MeshNodeSearcher.cpp
@@ -24,6 +24,7 @@
 
 // MeshGeoToolsLib
 #include "MeshGeoToolsLib/HeuristicSearchLength.h"
+#include "MeshGeoToolsLib/MeshNodesOnPoint.h"
 #include "MeshGeoToolsLib/MeshNodesAlongPolyline.h"
 #include "MeshGeoToolsLib/MeshNodesAlongSurface.h"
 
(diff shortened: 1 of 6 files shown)


TomFischer:
In my opinion the assumption that the number of points given in a geometry file is not quite large is not always fulfilled. In some projects there are a **huge** number of geometry points.

endJunction:
Please specify what is "not quite large" and "huge", otherwise reasoning on the algorithms is difficult.

TomFischer:
Huge means a number in the order of mesh nodes.

norihiro-w:
i meant up to 1e5. 

Anyway I would not argue here which algorithm is to be used. I argue that current `GeoLib::Grid` doesn't satisfy the requirements for mesh node searching and therefore suggest using the simpler alternative at the moment since I don't have time to understand and extend `GeoLib::Grid`. However i also don't mind to keep using `GeoLib::Grid` if someone extends it.

TomFischer:
@norihiro-w Can you give some examples where `GeoLib::Grid<MeshLib::Node>` does not return the expected result? Then I will fix this in `GeoLib::Grid<MeshLib::Node>`.

The algorithm suggested in this PR is **very** expensive for large meshes even for a small number of search points since it depends on the number of  mesh nodes. If I remember correctly the introduction of a Grid based search in OGS-5 the search time for a large example was reduced from 45 min to below 1 second.

norihiro-w:
thank you for your kind offer fixing it. please take a look at a test modified in this PR (TestMeshNodeSearch?) where the requirement is checked.

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

TomFischer:
I'll start changing the search (the class `Grid`) according to the requirements. I am unsure if I can finish this work before my vacation, that will start on thursday.

norihiro-w:
close this PR. Tom has already extended Grid class (https://github.com/TomFischer/ogs/tree/MeshNodeSearch) and will make a PR after his vacation.

PR #780, opened 2015-07-24
Cleanup MathLib/CMake

norihiro-w:
as titled


MathLib/CMakeLists.txt
```diff
@@ -59,14 +59,6 @@ if(METIS_FOUND)
 	include_directories(${METIS_INCLUDE_DIR})
 endif()
 
-if(OGS_USE_PETSC)
```
norihiro-w: the same code appears at line 41

PR #782, opened 2015-07-27
Small fix for error in GeoLib::gaussPointInTriangle.

TomFischer:
PR fixes an error in `GeoLib::gaussPointInTriangle(MathLib::Point3d const& q, ...)`. In cases the point `q` and the orthogonal projection of `q` into the plane defined by the triangle are identical **and** the tolerance (`eps_pnt_out_of_plane`) is zero, the algorithm returned a wrong result.


endJunction:
:fast_forward:

PR #783, opened 2015-07-27
Add another function for construction of a surface mesh approximating a function.

TomFischer:
This PR is a preparation for another PR that will improve the search for points/nodes along surfaces.


norihiro-w:
:+1:

PR #784, opened 2015-07-27
Restrict the interface of class MeshLib::Node.

TomFischer:
- Moved class `MeshLayerMapper` into namespace `MeshLib`.
- Moved method `Node::updateCoordinates()` form public to protected, which reduces the interface of class `Node`.


endJunction:
:fast_forward:

norihiro-w:
why is it necessary to move `Node::updateCoordinates()` to protected?

TomFischer:
> why is it necessary to move `Node::updateCoordinates()` to protected?

As you remarked, it is not necessary. But in my opinion the interfaces of classes should be as small as possible. For this reason I suggest to make the method `Node::updateCoordinates()` protected and allow only a small number of friend classes to update the coordinates.

norihiro-w:
:+1: thanks for the explanation

PR #786, opened 2015-07-28
[GeoLib] Another forward declaration.

TomFischer:
- Added forward declaration of class `Triangle` in class `Surface`.
- Fixed include issue with class `Color`.


norihiro-w:
:+1:

PR #787, opened 2015-07-29
Add SurfaceGrid for the accelaration of searching

TomFischer:
In this PR class `SurfaceGrid` is used to accelerate the method `Surface::isPntInSfc(Point3d const& p)`. For this reason, the triangles of the `Surface` are sorted in an axis aligned structured grid - the `SurfaceGrid`. The cell of the axis aligned structured grid that contains the point p can be computed in O(1). Then only the triangles within the computed grid cell has to be tested.

### DIFF
--- a/GeoLib/Surface.cpp
+++ b/GeoLib/Surface.cpp
@@ -17,12 +17,12 @@
 // ThirdParty/logog
 #include "logog/include/logog.hpp"
 
+#include "Surface.h"
+
 // GeoLib
 #include "AABB.h"
 #include "Polygon.h"
-#include "Surface.h"
-
-// MathLib
+#include "SurfaceGrid.h"
 #include "AnalyticalGeometry.h"
 
 #include "Triangle.h"
@@ -32,7 +32,8 @@
 namespace GeoLib
 {
 Surface::Surface (const std::vector<Point*> &pnt_vec) :
-	GeoObject(), _sfc_pnts(pnt_vec), _bounding_volume(nullptr)
+	GeoObject(), _sfc_pnts(pnt_vec), _bounding_volume(nullptr),
+	_surface_grid(nullptr)
 {}
 
 Surface::~Surface ()
@@ -40,9 +41,10 @@ Surface::~Surface ()
 	for (std::size_t k(0); k < _sfc_triangles.size(); k++)
 		delete _sfc_triangles[k];
 	delete _bounding_volume;
+	delete _surface_grid;
 }
 
-void Surface::addTriangle (std::size_t pnt_a, std::size_t pnt_b, std::size_t pnt_c)
+void Surface::addTriangle(std::size_t pnt_a, std::size_t pnt_b, std::size_t pnt_c)
 {
 	assert (pnt_a < _sfc_pnts.size() && pnt_b < _sfc_pnts.size() && pnt_c < _sfc_pnts.size());
 
@@ -50,17 +52,31 @@ void Surface::addTriangle (std::size_t pnt_a, std::size_t pnt_b, std::size_t pnt
 	if (pnt_a == pnt_b || pnt_a == pnt_c || pnt_b == pnt_c)
 		return;
 
-	_sfc_triangles.push_back (new Triangle(_sfc_pnts, pnt_a, pnt_b, pnt_c));
+	_sfc_triangles.push_back(new Triangle(_sfc_pnts, pnt_a, pnt_b, pnt_c));
 	if (!_bounding_volume) {
 		std::vector<size_t> ids(3);
 		ids[0] = pnt_a;
 		ids[1] = pnt_b;
 		ids[2] = pnt_c;
 		_bounding_volume = new AABB<Point>(_sfc_pnts, ids);
+		if (_surface_grid == nullptr) {
+			_surface_grid = new SurfaceGrid(this);
+		}
 	} else {
-		_bounding_volume->update (*_sfc_pnts[pnt_a]);
-		_bounding_volume->update (*_sfc_pnts[pnt_b]);
-		_bounding_volume->update (*_sfc_pnts[pnt_c]);
+		bool bbx_updated(_bounding_volume->update(*_sfc_pnts[pnt_a]));
+		bbx_updated = bbx_updated || _bounding_volume->update(*_sfc_pnts[pnt_b]);
+		bbx_updated = bbx_updated || _bounding_volume->update(*_sfc_pnts[pnt_c]);
+		if (bbx_updated) {
+			delete _surface_grid;
+			_surface_grid = new SurfaceGrid(this);
+		} else {
+			if (! _surface_grid->sortTriangleInGridCells(_sfc_triangles.back())) {
+				ERR("Fatal: Could not insert triangle into surface grid. "
+					"To keep things consistent, the triangle is removed from "
+					"the surface!");
+				_sfc_triangles.pop_back();
+			}
+		}
 	}
 }
 
@@ -125,7 +141,8 @@ bool Surface::isPntInBoundingVolume(MathLib::Point3d const& pnt) const
 
 bool Surface::isPntInSfc(MathLib::Point3d const& pnt) const
 {
-	return (findTriangle(pnt)!=nullptr);
+	return _surface_grid->isPointInSurface(
+		pnt, std::numeric_limits<double>::epsilon());
 }
 
 const Triangle* Surface::findTriangle (MathLib::Point3d const& pnt) const
--- a/GeoLib/SurfaceGrid.cpp
+++ b/GeoLib/SurfaceGrid.cpp
@@ -1,0 +1,207 @@
+/*
+ * \date 2012-09-22
+ * \brief Definition of the SurfaceGrid class.
+ *
+ * \copyright
+ * Copyright (c) 2012-2015, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ */
+
+#include <algorithm>
+#include <bitset>
+
+#include <logog/include/logog.hpp>
+
+#include "SurfaceGrid.h"
+
+#include "Surface.h"
+#include "Triangle.h"
+
+#include "MathLib/Point3d.h"
+
+namespace GeoLib {
+
+SurfaceGrid::SurfaceGrid(Surface const*const sfc) :
+	AABB(sfc->getAABB()), _n_steps({{1,1,1}})
+{
+	// enlarge the bounding, such that the points with maximal coordinates
+	// fits into the grid
+	for (std::size_t k(0); k<3; ++k) {
+		_max_pnt[k] += std::abs(_max_pnt[k]) * 1e-6;
+		if (std::abs(_max_pnt[k]) < std::numeric_limits<double>::epsilon()) {
+			_max_pnt[k] = (_max_pnt[k] - _min_pnt[k]) * (1.0 + 1e-6);
+		}
+	}
+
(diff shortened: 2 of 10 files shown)


GeoLib/Surface.cpp
```diff
@@ -50,17 +52,30 @@ void Surface::addTriangle (std::size_t pnt_a, std::size_t pnt_b, std::size_t pnt
...
+	_sfc_triangles.push_back(new Triangle(_sfc_pnts, pnt_a, pnt_b, pnt_c));
 	if (!_bounding_volume) {
 		std::vector<size_t> ids(3);
 		ids[0] = pnt_a;
 		ids[1] = pnt_b;
 		ids[2] = pnt_c;
 		_bounding_volume = new AABB<Point>(_sfc_pnts, ids);
+		if (_surface_grid == nullptr) {
+			_surface_grid = new SurfaceGrid(this);
+	}
```
norihiro-w: indentation

GeoLib/Surface.cpp
```diff
@@ -50,17 +52,30 @@ void Surface::addTriangle (std::size_t pnt_a, std::size_t pnt_b, std::size_t pnt
...
 	} else {
-		_bounding_volume->update (*_sfc_pnts[pnt_a]);
-		_bounding_volume->update (*_sfc_pnts[pnt_b]);
-		_bounding_volume->update (*_sfc_pnts[pnt_c]);
+		if (_bounding_volume->update(*_sfc_pnts[pnt_a])
+			|| _bounding_volume->update(*_sfc_pnts[pnt_b])
```
norihiro-w: is it unnecessary to call _bounding_volume->update() for all pnt_a, pnt_b, and pnt_c? if it is necessary, `|` should be used instead of `||`.

GeoLib/Surface.cpp
```diff
@@ -50,17 +52,30 @@ void Surface::addTriangle (std::size_t pnt_a, std::size_t pnt_b, std::size_t pnt
...
+			|| _bounding_volume->update(*_sfc_pnts[pnt_c])) {
+			delete _surface_grid;
+			_surface_grid = new SurfaceGrid(this);
```
norihiro-w: is it right that surface grid can be frequently recreated for each time addTriangle() is called? If creating SurfaceGrid object is expensive, it might be better to create it after all triangles are added.

GeoLib/Surface.cpp
```diff
@@ -50,17 +52,30 @@ void Surface::addTriangle (std::size_t pnt_a, std::size_t pnt_b, std::size_t pnt
...
+		} else {
+			if (! _surface_grid->sortTriangleInGridCells(_sfc_triangles.back())) {
+				ERR("Fatal: Could not insert triangle into surface grid. "
```
norihiro-w: what could be a reason for this error? how can a user avoid this?

GeoLib/SurfaceGrid.h
```diff
@@ -0,0 +1,47 @@
...
+
+namespace GeoLib {
+
+// forward declarations
+class Triangle;
+class Surface;
+
+class SurfaceGrid : public AABB<GeoLib::Point> {
+public:
+	friend GeoLib::Surface;
```
norihiro-w: move a friend declaration to private. conceptually public don't need to know it.

GeoLib/SurfaceGrid.h
```diff
@@ -0,0 +1,47 @@
...
+	SurfaceGrid(GeoLib::Surface const*const sfc);
```
norihiro-w: explicit

GeoLib/SurfaceGrid.h
```diff
@@ -0,0 +1,47 @@
...
+	bool isPointInSurface(MathLib::Point3d const & pnt, double eps = 0) const;
```
norihiro-w: why not use double epsilon as a default value?

GeoLib/Surface.cpp
endJunction: To the second part: `|` is bit-wise-or. I think a logic-or, which is `||`, is correct here.

GeoLib/SurfaceGrid.h
```diff
@@ -0,0 +1,47 @@
...
+
+#include "AABB.h"
+
+namespace GeoLib {
+
+// forward declarations
+class Triangle;
+class Surface;
+
+class SurfaceGrid : public AABB<GeoLib::Point> {
```
norihiro-w: you might need to include Point.h

GeoLib/SurfaceGrid.cpp
```diff
@@ -0,0 +1,207 @@
...
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ */
+
+#include <algorithm>
+#include <bitset>
+
+#include <logog/include/logog.hpp>
+
+#include "SurfaceGrid.h"
```
norihiro-w: move this to line 10

GeoLib/SurfaceGrid.cpp
```diff
@@ -0,0 +1,207 @@
...
+namespace GeoLib {
+
+SurfaceGrid::SurfaceGrid(Surface const*const sfc) :
+	AABB(sfc->getAABB()), _n_steps({{1,1,1}})
+{
+	// enlarge the bounding, such that the points with maximal coordinates
+	// fits into the grid
+	for (std::size_t k(0); k<3; ++k) {
+		_max_pnt[k] += std::abs(_max_pnt[k]) * 1e-6;
+		if (std::abs(_max_pnt[k]) < std::numeric_limits<double>::epsilon()) {
```
norihiro-w: would you please explain what you are trying to do here? it looks for me if a maximum coordinate is nearly zero, you change its value to `- (min. coord)*(1+1e-6)`. if the minimum coordinate is positive sign, the new value becomes negative...

GeoLib/SurfaceGrid.cpp
```diff
@@ -0,0 +1,207 @@
...
+	//                where n_cells = _n_steps[0] * _n_steps[1] * _n_steps[2]
+	// *** with _n_steps[0] = ceil(pow(n_tris*delta[0]*delta[0]/(n_tris_per_cell*delta[1]*delta[2]), 1/3.)));
+	//          _n_steps[1] = _n_steps[0] * delta[1]/delta[0],
+	//          _n_steps[2] = _n_steps[0] * delta[2]/delta[0]
+	auto ceil = [](double v){
+		return static_cast<std::size_t>(std::ceil(v));
+	};
+	switch (dim.count()) {
+	case 3: // 3d case
+		_n_steps[0] = ceil(std::cbrt(
```
norihiro-w: std::ceil?

GeoLib/SurfaceGrid.cpp
```diff
@@ -0,0 +1,207 @@
...
+		_max_pnt[k] += std::abs(_max_pnt[k]) * 1e-6;
+		if (std::abs(_max_pnt[k]) < std::numeric_limits<double>::epsilon()) {
+			_max_pnt[k] = (_max_pnt[k] - _min_pnt[k]) * (1.0 + 1e-6);
+		}
+	}
+
+	std::array<double, 3> delta{{ _max_pnt[0] - _min_pnt[0],
+		_max_pnt[1] - _min_pnt[1], _max_pnt[2] - _min_pnt[2] }};
+
+	if (delta[0] < std::numeric_limits<double>::epsilon()) {
```
wenqing: Since std::numeric_limits is called many times in the entire source code, I suggest to set several constants in a namespace for that, e.g.
```
namespace Constant
{
    const double double_epsion = std::numeric_limits<double>::epsilon();
    const double double_max = std::numeric_limits<double>::max();
   ...
}
```
This could save a little compilation time. If interest, I will may prepare a PR concerning to this point.

GeoLib/SurfaceGrid.h
endJunction: There is no difference in which section a friend is declared.

GeoLib/SurfaceGrid.cpp
endJunction: @wenqing If you can show a compilation speedup of  at least 1%, then and only then we can consider to introduce the `Constant` namespace.

I doubt there will be any difference, because almost all of the numeric_limits class' members are static const (constexpr) defined to preprocessor constants, no deeply nested templates etc...

GeoLib/Surface.cpp
norihiro-w: using `|`, one can let all the conditions evaluated, which is not the case for `||`.

GeoLib/SurfaceGrid.h
norihiro-w: sure there is no difference in compilation. it matters when someone reading interface of this class. friend is a tool for encapsulation and its declaration should not be put in public section where a user of this class will be interested.

GeoLib/SurfaceGrid.cpp
wenqing: I see.

GeoLib/Surface.cpp
TomFischer: > is it right that surface grid can be frequently recreated for each time addTriangle() is called?

Yes, it is true.

> If creating SurfaceGrid object is expensive, it might be better to create it after all triangles are added.

In this case we need a method like `finalize()` which signals that nothing will be added to the geometry. In the past I had this idea several times. But the implementation would have extensive consequences to the source code of OGS. We need to find and update all places where geometry is added. If we want to do this, I suggest a separate PR.

GeoLib/Surface.cpp
TomFischer: At the moment I can not figure out a case where this piece of code will be executed. Should it be deleted?

GeoLib/SurfaceGrid.h
TomFischer: No special reason for this default value. I'll changed it to your suggestion.

GeoLib/SurfaceGrid.cpp
TomFischer: In the case both values, `_min_pnt[k]` and `_max_pnt[k]`, are small positive values the property `_min_pnt[k] <= _max_pnt[k]` holds true and the new value should never be negative.

GeoLib/SurfaceGrid.cpp
TomFischer: `ceil` is the lambda above (line 74) which uses `std::ceil` and additionally does a static_cast.

GeoLib/Surface.cpp
TomFischer: The update should be executed for all three points. I change this.

GeoLib/SurfaceGrid.h
TomFischer: I move the friend declaration to the private section.

TomFischer:
Rebased and changes according to the comments.

GeoLib/Surface.cpp
norihiro-w: I would keep it as a safe check. we can improve the message when the reason is identified.

norihiro-w:
:+1:

GeoLib/SurfaceGrid.cpp
endJunction: Nori is right, better to rename it to `sc_ceil` like in the ufz/ogs#796.

endJunction: Actually this `switch-case` is very similar to that in ufz/ogs#796 in `Grid.h`, `initNumberOfSteps()`. Is a single meaningful algorithm covering both cases possible?

TomFischer: This can be done later in a separate PR. `sc_ceil` will be corrected in some minutes.

TomFischer:
Rebased again and some changes incorporating the comments.

PR #789, opened 2015-07-30
Grid extension to return mulitiple points located in given search radius

norihiro-w:
maybe we can start reviewing the code while Tom is in vacation


endJunction:
I reviewed the `Grid(Iterator, Iterator, size_t)` ctor and the `getNearestPoint()`. Both versions are equivalent.

For the `initNumberOfSteps()` function I _guess_ they are doing some similar calculations.

The review could be much easier, if the changes of algorithms were separate from superficial changes.

PR #791, opened 2015-07-30
Raster intensity to scalar array

rinkk:
When raster files are converted into meshes, intensity can now be interpreted as a scalar array.

### DIFF
--- a/MeshLib/MeshGenerators/VtkMeshConverter.cpp
+++ b/MeshLib/MeshGenerators/VtkMeshConverter.cpp
@@ -204,7 +204,7 @@ MeshLib::Mesh* VtkMeshConverter::constructMesh(const double* pixVal,
 
 			if (set_node)
 			{
-				double zValue = (intensity_type == UseIntensityAs::ELEVATION) ? pixVal[index] : origin[2];
+				double zValue = (intensity_type == UseIntensityAs::ELEVATION) ? pixVal[index] : 0;
 				MeshLib::Node* node (new MeshLib::Node(x_offset + (scalingFactor * j), y_offset + (scalingFactor * i), zValue));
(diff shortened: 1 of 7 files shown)


bilke:
:+1:

endJunction:
Jenkins, test this please.

MeshLib/MeshGenerators/VtkMeshConverter.cpp
```diff
@@ -250,7 +260,14 @@ MeshLib::Mesh* VtkMeshConverter::constructMesh(const double* pixVal,
 	if (elements.empty())
 		return nullptr;
 
-	return new MeshLib::Mesh("RasterDataMesh", nodes, elements); // the name is only a temp-name, the name given in the dialog is set later
+	if (value_vec->empty())
+		properties.removePropertyVector("Colour");
+
+	boost::optional< MeshLib::PropertyVector<unsigned>& > materials =
+		properties.createNewPropertyVector<unsigned>("MaterialIDs", MeshLib::MeshItemType::Cell, 1);
+	materials->insert(materials->begin(), elements.size(), 0);
```
endJunction: What is the size of `*materials` vector after the creation?
 - If it is 0, then the insert is OK, though `std::fill_n(std::back_inserter(*materials), elements.size(), 0)` is better, because the intention is to fill the vector, not to insert zeros in front of the existing values.
 - If it is not 0, then the insert is probably wrong here, because the existing elements would be preserved and the resulting  `*materials` vector would grow by `elements.size()`.

rinkk: it returns an empty vector. i changed to insert location to ```materials->end()``` anyway to keep it consistent with other cases where this is done.

PR #801, opened 2015-08-12
MATERIAL is not in UseIntensityAs anymore

norihiro-w:
This PR fixes compiling errors due to missing MATERIAL enum type in MeshLib::UseIntensityAs, though Jenkins is strangely not complaining it. 

In this PR, I replaced MATERIAL with DATAVECTOR, but I have no idea which one should be used. Let me know if this is wrong.


endJunction:
:+1:

rinkk:
It's correct the way you did it. I renamed this a while back because "material" seemed too specific a name for the colour array.
:+1:

PR #802, opened 2015-08-12
bugfix in LinearInterpolationOnSurface

norihiro-w:
fix a bug that an interpolation point was not rotated to XY plane

### DIFF
--- a/GeoLib/AnalyticalGeometry-impl.h
+++ b/GeoLib/AnalyticalGeometry-impl.h
@@ -15,22 +15,25 @@ void getNewellPlane (InputIterator pnts_begin, InputIterator pnts_end,
                      MathLib::Vector3 &plane_normal,
                      double& d)
 {
-    d = 0;
     MathLib::Vector3 centroid;
-    std::size_t n_pnts(std::distance(pnts_begin, pnts_end));
     for (auto i=std::prev(pnts_end), j=pnts_begin; j!=pnts_end; i = j, ++j) {
-        plane_normal[0] += ((*(*i))[1] - (*(*j))[1])
-                           * ((*(*i))[2] + (*(*j))[2]); // projection on yz
-        plane_normal[1] += ((*(*i))[2] - (*(*j))[2])
-                           * ((*(*i))[0] + (*(*j))[0]); // projection on xz
-        plane_normal[2] += ((*(*i))[0] - (*(*j))[0])
-                           * ((*(*i))[1] + (*(*j))[1]); // projection on xy
+        auto &pt_i = *(*i);
+        auto &pt_j = *(*j);
+        plane_normal[0] += (pt_i[1] - pt_j[1])
+                           * (pt_i[2] + pt_j[2]); // projection on yz
+        plane_normal[1] += (pt_i[2] - pt_j[2])
+                           * (pt_i[0] + pt_j[0]); // projection on xz
+        plane_normal[2] += (pt_i[0] - pt_j[0])
+                           * (pt_i[1] + pt_j[1]); // projection on xy
 
-        centroid += *(*j);
+        centroid += pt_j;
     }
 
     plane_normal.normalize();
-    d = MathLib::scalarProduct(centroid, plane_normal) / n_pnts;
+    auto n_pnts(std::distance(pnts_begin, pnts_end));
+    assert(n_pnts > 2);
+    if (n_pnts > 0)
+        d = MathLib::scalarProduct(centroid, plane_normal) / n_pnts;
 }
 
 template <class T_POINT>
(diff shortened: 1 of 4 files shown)


norihiro-w:
@TomFischer would you please review ce84489 where I extended `rotatePointsToXY()` taking two ranges of point vectors; first one used for computing rotation matrix and second one for points to be rotated? This is needed when an interpolation point is located at one of triangle vertices.

GeoLib/AnalyticalGeometry.h
```diff
@@ -123,6 +148,19 @@ void rotatePoints(MathLib::DenseMatrix<double> const& rot_mat, std::vector<GeoLi
 void rotatePointsToXY(std::vector<GeoLib::Point*> &pnts);
 
 /**
+ * rotate points to X-Y plane
+ * @param p_pnts_begin Iterator pointing to the initial element in a vector of points used for computing a rotation matrix
+ * @param p_pnts_end Iterator pointing to the final element in a vector of points used for computing a rotation matrix
```
TomFischer: Do you mean an iterator to the element following the last element of the vector? Same for line 155.

TomFischer:
IMHO can be merged if the very small documentation issue is solved. :+1:

GeoLib/AnalyticalGeometry.h
norihiro-w: fixed

GeoLib/AnalyticalGeometry-impl.h
```diff
@@ -10,23 +10,23 @@
...
 {
 
-template <class T_POINT>
-void getNewellPlane (const std::vector<T_POINT*>& pnts,
+template <typename InputIterator>
+void getNewellPlane (InputIterator pnts_begin, InputIterator pnts_end,
                      MathLib::Vector3 &plane_normal,
                      double& d)
 {
     d = 0;
```
endJunction: setting d = 0 is not required. See last line of the function.

GeoLib/AnalyticalGeometry-impl.h
```diff
@@ -10,23 +10,23 @@
...
-    for (std::size_t i(n_pnts - 1), j(0); j < n_pnts; i = j, j++) {
-        plane_normal[0] += ((*(pnts[i]))[1] - (*(pnts[j]))[1])
-                           * ((*(pnts[i]))[2] + (*(pnts[j]))[2]); // projection on yz
-        plane_normal[1] += ((*(pnts[i]))[2] - (*(pnts[j]))[2])
-                           * ((*(pnts[i]))[0] + (*(pnts[j]))[0]); // projection on xz
-        plane_normal[2] += ((*(pnts[i]))[0] - (*(pnts[j]))[0])
-                           * ((*(pnts[i]))[1] + (*(pnts[j]))[1]); // projection on xy
-
-        centroid += *(pnts[j]);
+    std::size_t n_pnts(std::distance(pnts_begin, pnts_end));
```
endJunction: `n_pnts` are not used here but at the end of the function; please move down.
Also it is possible that std::distance returns a negative value causing underflow, or zero causing division by zero.
Suggest to use signed type and check for negative and zero values.

GeoLib/AnalyticalGeometry.h
```diff
@@ -110,6 +125,16 @@ void computeRotationMatrixToXZ(MathLib::Vector3 const& plane_normal,
 /**
  * rotate points according to the rotation matrix
  * @param rot_mat 3x3 dimensional rotation matrix
+ * @param pnts_begin Iterator pointing to the initial element in a vector of points to be rotated
+ * @param pnts_end Iterator pointing to the element following the last element in a vector of points to be rotated
+ */
+void rotatePoints(
+        MathLib::DenseMatrix<double> const& rot_mat,
+        std::vector<GeoLib::Point*>::iterator pnts_begin, std::vector<GeoLib::Point*>::iterator pnts_end);
```
endJunction: Somewhat inconsistent interface compared to `getNewellPlane`: Here the iterator type is fixed, while in `getNewellPlane` it is a template type.

GeoLib/AnalyticalGeometry-impl.h
```diff
@@ -10,23 +10,23 @@
...
+    for (auto i=std::prev(pnts_end), j=pnts_begin; j!=pnts_end; i = j, ++j) {
+        plane_normal[0] += ((*(*i))[1] - (*(*j))[1])
```
wenqing: It might be better to introduce two intermediate variables as
```
const * double x = (*(*i));
const * double y = (*(*j));
```

norihiro-w:
updated

endJunction:
:fast_forward:
@TomFischer ?

PR #803, opened 2015-08-12
[Mesh] rework on ElementStatus. no changes allowed after construction

norihiro-w:
`MeshLib::ElementStatus` class was reworked in order to cache searched active nodes and elements  for the given condition. To enable this, all set functions in the class were removed and one can set active elements only via constructors.

### DIFF
--- a/MeshLib/ElementStatus.h
+++ b/MeshLib/ElementStatus.h
@@ -20,53 +20,52 @@
 namespace MeshLib {
 	class Mesh;
 	class Element;
+	class Node;
 
-/**
- * Manages the active/inactive flags for mesh elements and their nodes
- */
 class ElementStatus
 {
 
 public:
-	/// Constructor
+	/// Constructor assuming all nodes and elements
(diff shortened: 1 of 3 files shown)


MeshLib/ElementStatus.h
```diff
@@ -20,47 +20,42 @@
...
+
 	/// Returns the total number of active nodes
 	std::size_t getNActiveNodes() const;
 
 	/// Returns the total number of active elements
 	std::size_t getNActiveElements() const;
 
-	/// Activates/Deactives all mesh elements
-	void setAll(bool status);
+	~ElementStatus() {}
```
endJunction: This can be removed.

endJunction:
:fast_forward:

bilke:
:+1: but maybe @rinkk may have a look too.

rinkk:
I implemented this based on what you needed back then. Feel free to change whatever you need to... :+1:

PR #806, opened 2015-08-12
Repleacement of some Eigen functions with own implementations in shape stuff

norihiro-w:
as titled. `setZeo()`, `inverse()`, `determinant()` in Eigen::Matrix can be slow (at least with dynamic matrix).

### DIFF
--- a/MathLib/LinAlg/MatrixTools.h
+++ b/MathLib/LinAlg/MatrixTools.h
@@ -1,0 +1,124 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2015, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef MATHLIB_MATRIXTOOLS_H_
+#define MATHLIB_MATRIXTOOLS_H_
+
+#include <type_traits>
+
+#ifdef OGS_USE_EIGEN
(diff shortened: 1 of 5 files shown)


bilke:
Although 40 lines is not much I would not complicate the code because of small performance improvement in a particular use case. In case the performance improvement is not that small and it is crucial for you give some more motivation / explanation. If not then I suggest to replace these Eigen function if the need arises in the future.

endJunction:
Tested on cube mesh with 1e3 elements, clang-3.6.2, -O3 -DNDEBUG.

Indeed, the determinant and inverse computations for dynamic eigen matrices are significantly more expensive than the by-hand coded solutions; 64.9M instructions vs. 16.7M instructions for the `computeMappingMatrices<..., DNDX>(...)` call.
For fixed size matrices the difference is negligible 15.1M instructions vs. 14.7M instructions.

I would suggest to create a free template functions in `MathLib` with specializations for fixed size and dynamic size matrices for the determinant and the inverse calculations. (Or at least in the same file...)

_skip this:_ Another possibility would be to call Eigen's `PartialPivLU` only once before computing determinant and the inverse. Currently the `PartialPivLU` is called twice, once from determinant, once from the inverse calculations. This would save around 14M instruction executions. (But I see now, it's not worth it because the inverse is still much more expensive...)

norihiro-w:
@endJunction  thank you for the test. I'll add template functions in MathLib

norihiro-w:
@endJunction I added template functions in MathLib.

TomFischer:
:fast_forward:

PR #807, opened 2015-08-14
add STL_NO_DEBUG cmake option

norihiro-w:
In current code, debugging of STL is always enabled if a build type is debug. This is usually useful but can make debugging difficult for large problems because some STL functions are very slow in debug mode (e.g.  std::sort() in Mesh::setElementNeighbors()). With STL_NO_DEBUG=ON, one can debug the OGS code without using the debug version of STL.


endJunction:
:+1:

PR #809, opened 2015-08-14
Full path includes cleanup

endJunction:
Now only the DataExplorer and SimpleTests take advantage of cmake's include_directories.


bilke:
:+1:

norihiro-w:
:+1:

Issue #799, opened 2015-08-11
Unified Indentation: Tabs or Spaces


chleh:
I've started with OGS6 this week, and I find it very annoying that the indentation convention changes from one source file to another. It is not a big issue to change the associated setting in my editor. But as with every small but repeated issue: It makes me mad.
It would be great if you could stick to a single indentation style throughout the whole project.
For me the specific style does not matter so much: Either tabs (and spaces for continued expressions?) or (4?) spaces...
You maybe have a strong opinion about which style is the right one, but I think having one style in all files outweighs having your style in 50% of files by far.

bilke:
:+1:

I am more the Tab-guy but because of IDEs and text editors can handle spaces equally well for me it does not matter which one we choose...

endJunction:
Using clang-format (as recommended by google style guide creator on CppCon14) would be a good idea:
```
            \ "Language" : "Cpp",
            \ "BasedOnStyle" : "Google",
            \ "AccessModifierOffset" : -4,
            \ "AllowShortLoopsOnASingleLine" : "false",
            \ "AllowShortIfStatementsOnASingleLine" : "false",
            \ "AllowShortFunctionsOnASingleLine" : "false",
            \ "BreakBeforeBraces" : "Allman",
            \ "ExperimentalAutoDetectBinPacking" : "true",
            \ "Standard" : "C++11"
```
... with "UseTab" being set no "Never" or "ForIndentation".
In a good editor there should be a possibility to automatically detect whether there are tabs or spaces; I use such feature together with above clang-format style and therefor I don't care about the spaces or tabs.

FYI: There are currently around 17489 lines starting with four+ spaces and  42672 starting with tabs, which is roughly 1:2 ratio.

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/799).

PR #813, opened 2015-08-19
Add clang-format.

endJunction:
In consequence of ufz/ogs#799 a *suggested* .clang-format.
See http://clang.llvm.org/docs/ClangFormatStyleOptions.html for more options.

Fixes  ufz/ogs#799

See [available IDE / editor integrations](http://clang.llvm.org/docs/ClangFormat.html) or [this Sublime-plugin](https://packagecontrol.io/packages/Clang%20Format).

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,13 @@
+---
+Language:        Cpp
+BasedOnStyle:    Google
+
+AccessModifierOffset: -4
+AllowShortLoopsOnASingleLine: false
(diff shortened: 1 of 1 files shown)


bilke:
When a line is too long it indents on the next line with 4 spaces instead of a tab.

Switching to `UseTab: Always` fixes that but then it also aligns with tab...

bilke:
:+1:

PR #814, opened 2015-08-19
Replace GEOObjects info/err msg with DBUG.

endJunction:
This reduces noise on the console.

Someone more knowledgable about the GEOObjects might have a look if the messages are needed at all.


TomFischer:
:+1:

norihiro-w:
:+1:

PR #819, opened 2015-08-28
Added time-stepping loop.

chleh:
This PR covers three things:
* the process initialization and solving loops has been moved from ogs.cpp to ProjectData.
* there is a new element `<time_stepping>` in the project file.
* ProjectData now also has a time loop. Currently it is enabled for all processes regardless of whether they are time-dependent or not.
* I have added a timestepper `DummyTimeStepping` which does exactly one timestep. It is intended for time-independent processes.

The `<time_stepping>` element currently looks as follows:
```
    <time_stepping>
        <type>FixedTimeStepping</type>
        <t_initial>   0.0 </t_initial>
        <t_end>    2000.0 </t_end>
        <dt>          0.5 </dt>
    </time_stepping>
```

Since it is mandatory, many project files will have to be changed.
Also the `solve()` method in `Process` now has a parameter, so I will have to change the GroundwaterFlowProcess. I'd like to do that in a subsequent PR.


TomFischer:
In contrast to Jenkins for me the code compiles and links fine. clang version 3.6.2 gives the following two warnings:
`GroundwaterFlowProcess.h:197:10: warning: 'initialize' overrides a member function but is not marked 'override'`
and `GroundwaterFlowProcess.h:253:10: warning: 'post' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]`.

NumLib/TimeStepping/Algorithms/DummyTimeStepping.h
```diff
@@ -0,0 +1,70 @@
...
+namespace NumLib
+{
+
+/**
+ * \brief Fixed time stepping algorithm
+ *
+ * This algorithm is meant to perform exactly one timestep.
+ * To be used e.g. for time-invariant processes in a time stepping loop
+ */
+class DummyTimeStepping : public ITimeStepAlgorithm
```
endJunction: I would rename it to `SingleStepTimeAlgorithm` or something similar, repeating the behaviour described in the documentation.
A dummy variable/code/class has in contrary no behaviour at all (in my understanding).

Applications/ApplicationsLib/ProjectData.cpp
```diff
@@ -272,3 +278,89 @@ void ProjectData::parseOutput(ConfigTree const& output_config,
 
 	_output_file_prefix = path + *file;
 }
+
+void ProjectData::parseTimeStepping(ConfigTree const& timestepping_config)
```
wenqing: change return type to ``bool``.

Applications/ApplicationsLib/ProjectData.cpp
```diff
@@ -272,3 +278,89 @@ void ProjectData::parseOutput(ConfigTree const& output_config,
...
+{
+	using namespace ProcessLib;
+
+	DBUG("Reading timestepping configuration.");
+
+	auto const type = timestepping_config.get_optional<std::string>("type");
+	if (!type)
+	{
+		ERR("Could not find required timestepper type.");
+		std::abort();
```
wenqing: return false.

Applications/ApplicationsLib/ProjectData.cpp
```diff
@@ -69,6 +72,9 @@ ProjectData::ProjectData(ConfigTree const& project_config,
 
 	// output
 	parseOutput(project_config.get_child("output"), path);
+
+	// timestepping
+	parseTimeStepping(project_config.get_child("time_stepping"));
```
wenqing: Here, with a boolean return type, the program can determine whether the time stepping is needed. That means that the input of time stepping data can not be mandatory.  If there is not any time stepping data in the input file, the problem is solved as steady state one. Under such consideration, the change of the existing input files is not needed.

endJunction: I think it is better to explicitly state the time-stepping scheme in the config files; this can reduce number of surprised users.

endJunction:
Jenkins, test this please!

Applications/ApplicationsLib/ProjectData.cpp
```diff
@@ -272,3 +277,89 @@ void ProjectData::parseOutput(ConfigTree const& output_config,
...
+}
+
+void ProjectData::solveProcesses()
+{
+	INFO("Solve processes.");
+
+	std::string const out_pref = getOutputFilePrefix();
+	std::string const output_file_name(getOutputFilePrefix() + ".vtu");
+
+    while (_time_stepper->next()) // skips zeroth timestep, but OK since end of first timestep is after first delta t
```
endJunction: Just an idea:

if limiting number of time steps, then this could become a for loop.
```c++
std::size_t const maximum_timesteps = 42; // zero for unlimited.
for (std::size_t timestep = 1; timestep < maxium_timesteps; ++i)
{
    ...
}
```
but maybe there is already a mechanism for this limit inside the time_stepper objects...

NumLib/TimeStepping/Algorithms/FixedTimeStepping.h
```diff
@@ -58,6 +60,13 @@ class FixedTimeStepping : public ITimeStepAlgorithm
      */
     FixedTimeStepping(double t_initial, double t_end, const std::vector<double> &vec_all_dt);
 
+    /**
+     * @brief Create timestepper from the given configuration
+     *
+     * Currently thisd function only covers uniform timestep size.
```
endJunction: typo: thisd.

NumLib/TimeStepping/Algorithms/FixedTimeStepping.h
```diff
@@ -58,6 +60,13 @@ class FixedTimeStepping : public ITimeStepAlgorithm
...
+     */
+    static FixedTimeStepping* newInstance(ConfigTree const& config);
```
endJunction: returning a unique_ptr could be better indicating ownership transfer.

Applications/ApplicationsLib/ProjectData.cpp
chleh: In general (e.g. adaptive timestepping) there can only be an artificail limit for timesteps like "abort after 1 million steps". Is it desirable to have such a limit?

NumLib/TimeStepping/Algorithms/FixedTimeStepping.h
chleh: In GroundwaterFlowProcess there is a line `_A.reset(_global_setup.createMatrix(_local_to_global_index_map->dofSize()));`, i.e. `createMatrix()` returns a plain pointer.
So at least this case is not handled always with `unique_ptr`.
I think, it is not so much of an issue here, because the name `newInstance` already indicates ownership transfer.

Applications/ApplicationsLib/ProjectData.h
```diff
@@ -109,21 +111,15 @@ using ConfigTree = boost::property_tree::ptree;
...
-	{
-		return _processes.end();
-	}
+	/// Initializes all processes.
+	///
+	/// \pre buildProcesses() must habe been called
+	void initializeProcesses();
+	void solveProcesses();
 
+	/// Solves all processes
```
bilke: I guess this comment should be moved over to the function declaration.

bilke:
:+1:

chleh:
Jenkins fails with the message
```
...
Scanning dependencies of target data
[  6%] Generating /gpfs1/data/ogs/jenkins/workspace/OGS-6/Linux-PRs/build/Tests/Data/Elliptic/cube_1x1x1_GroundWaterFlow/cube_1e0.prj
-- Fetching "http://www.opengeosys.org/images/dev/MD5/ce002c594365b012dd338bd27467bd31"
-- [download 49% complete]
-- [download 100% complete]
-- Fetching "http://opengeosys.s3.amazonaws.com/ogs6-data/MD5/ce002c594365b012dd338bd27467bd31"
CMake Error at /gpfs0/global/local/cmake/3.1.1-1/share/cmake-3.1/Modules/ExternalData.cmake:774 (message):
  Object MD5=ce002c594365b012dd338bd27467bd31 not found at:

    http://www.opengeosys.org/images/dev/MD5/ce002c594365b012dd338bd27467bd31 (wrong hash MD5=4eeda8a614a8fb35916db2d7b6233018)
    http://opengeosys.s3.amazonaws.com/ogs6-data/MD5/ce002c594365b012dd338bd27467bd31 ("HTTP response code said error")
Call Stack (most recent call first):
  /gpfs0/global/local/cmake/3.1.1-1/share/cmake-3.1/Modules/ExternalData.cmake:796 (_ExternalData_download_object)
...
```

I uploaded a file with the md5-hash `ce002c594365b012dd338bd27467bd31` yesterday evening to eve. Do Jenkins' external data files need to be updated? I thought, that happens automatically every night.

endJunction:
I have seen that some of the ogs.cpp code was moved to ProjectData class. This is not good because the ProjectData class is also used by DataExplorer, for example, and is just a container for simulation. The simulation itself should be done in the ogs.cpp.

PR #820, opened 2015-08-28
Small issues

chleh:
This PR contains only changes to some small issues, most notably:
* XML Parser now trims whitespace
* changed ICs and BCs in ProcessVariable to unique_ptr


TomFischer:
Very nice! :+1:

bilke:
:+1:

PR #823, opened 2015-08-31
Small cleanup

endJunction:
- Avoid some implicit type conversions.
 - Reuse existing type definitons.


TomFischer:
:fast_forward:

PR #824, opened 2015-09-01
Extrapolation methods

chleh:
This PR introduces classes for linear least squares extrapolation from integration points to mesh nodes.

TODO:
- [ ] add documentation
- [ ] add tests


wenqing:
:+1: 
You can add the formulas about the least square extrapolation to the documentation.  PS: For quadrilateral and tetrahedral elements, the interpolation method can be applied as  a fast extrapolation method.

chleh:
@wenqing: I know that for certain elements one can use the interpolation method for extrapolation,
but I wanted a method that works for any element out of the box.

PR #825, opened 2015-09-01
Added function for shape matrix interpolation.

chleh:
The provided function implements the usual shape matrix interpolation of nodal values to integration points. It can be used, e.g. in any local assembly where matrix coefficients depend on the value of the unknowns.

### DIFF
--- a/NumLib/Function/Interpolation.h
+++ b/NumLib/Function/Interpolation.h
@@ -1,0 +1,43 @@
+#ifndef NUMLIB_INTERPOLATION_H
+#define NUMLIB_INTERPOLATION_H
+
+namespace NumLib
+{
+
+/**
+ * Interpolates variables given at element nodes according to the given shape matrix.
+ *
+ * This function simply does the usual finite-element interpolation, i.e. multiplication
+ * of nodal values with the shape function.
+ *
+ * @param nodal_values   vector of nodal values
+ * @param shape_matrix_N shape matrix of the point to which will be interpolated
+ * @param num_nodal_dof  number of nodal variables that will be interpolated
+ * @param interpolated_values array of addresses to which the interpolated values will be written
+ *
+ * The size of interpolated_values must be equal to num_nodal_dof.
+ */
+template<typename NodalValues, typename ShapeMatrix>
+void shapeFunctionInterpolate(
+        const NodalValues& nodal_values,
+        const ShapeMatrix& shape_matrix_N,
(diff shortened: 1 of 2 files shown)


endJunction:
Yes! Array is perfect.

TomFischer:
Do you plan to implement a test?

chleh:
I added a simple test withthe last commit. I hope, it suffices, since the functionality of the interpolation method is really simple.

NumLib/Function/Interpolation.h
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * This function simply does the usual finite-element interpolation, i.e. multiplication
+ * of nodal values with the shape function.
+ *
+ * @param nodal_values   vector of nodal values, ordered by component
+ * @param shape_matrix_N shape matrix of the point to which will be interpolated
+ * @param interpolated_values array of addresses to which the interpolated values will be written
+ */
+template<typename NodalValues, typename ShapeMatrix, std::size_t NodalDOF>
+void shapeFunctionInterpolate(
```
wenqing: The name of ``elementInterpolation`` or ``interpolateNodeValues`` could be more meaningful.

wenqing: If the third party matrix like Eigen is used, we can skip this function.

bilke:
:+1:

NumLib/Function/Interpolation.h
chleh: The purpose of this function is to avoid writing the loop in each process. And I think, even with Eigen one would need a loop and, e.g. `block` expressions.

Tests/NumLib/TestFunctionInterpolation.cpp
```diff
@@ -0,0 +1,100 @@
...
+    // set up mesh element
+    auto pt_a = MeshLib::Node({ 0.0, 0.0, 0.0 }, 0);
+    auto pt_b = MeshLib::Node({ 1.0, 0.0, 0.0 }, 0);
+
+    auto const element = MeshLib::Line(std::array<MeshLib::Node*, 2>{
+                                           &pt_a, &pt_b
+                                       }, 0u, 0u);
+
+    // set up shape function
+    FemType finite_element(*static_cast<const typename ShapeFunction::MeshElement*>(&element));
```
endJunction: I guess the windows compiler does not like the `typename` here...

bilke:
:+1:

PR #826, opened 2015-09-02
Explicit instantiation of GroundwateFlowProcess<GlobalSetup>.

endJunction:
This decouples the ogs.cpp compilation unit from the GroundwaterFlowProcess.

@chleh Is this compatible with your implementation of the TES process?


chleh:
As far as I understand, this PR effects that not all processes will be recompiled each time only a single process is changed.
In my implementation I have "deactivated" some template parameters, especially for shape matrices. Thus, my code will not be compiled once for every possible shape matrix type.

chleh:
:+1:

bilke:
:+1:

endJunction:
Jenkins, test this please!

endJunction:
OK, Jenkins is reporting green again.

chleh:
:+1:

PR #827, opened 2015-09-02
Initial conditions

endJunction:
This adds an initial condition handling to set the start vector for the linear solver.
Aside the uniform value a mesh property based initial condition is added; the latter allows one to reuse previous solutions.

There are three parts:
 - clang-format
 - Adding/moving InitialConditions code [middle four commits](https://github.com/endJunction/ogs/compare/a6ac6a1...496bb2c)
 - Using the initial conditions in GroundwaterFlowProcess [last three commits]
(https://github.com/endJunction/ogs/compare/496bb2c...InitialConditions)

*(The commits below are ordered differently)*

### DIFF
--- a/ProcessLib/GroundwaterFlowProcess.h
+++ b/ProcessLib/GroundwaterFlowProcess.h
@@ -233,7 +233,8 @@ public:
         std::size_t const n = _mesh.getNNodes();
         for (std::size_t i = 0; i < n; ++i)
         {
-            MeshLib::Location l(_mesh.getID(), MeshLib::MeshItemType::Node, i);
+            MeshLib::Location const l(_mesh.getID(),
+                                      MeshLib::MeshItemType::Node, i);
             std::size_t const global_index =
                 _local_to_global_index_map->getGlobalIndex(
                     l, 0);  // 0 is the component id.
--- a/ProcessLib/GroundwaterFlowProcess.h
+++ b/ProcessLib/GroundwaterFlowProcess.h
@@ -216,6 +216,8 @@ public:
         _x.reset(_global_setup.createVector(_local_to_global_index_map->dofSize()));
         _rhs.reset(_global_setup.createVector(_local_to_global_index_map->dofSize()));
 
+        setInitialConditions(*_hydraulic_head);
+
         if (_mesh.getDimension()==1)
             createLocalAssemblers<1>();
         else if (_mesh.getDimension()==2)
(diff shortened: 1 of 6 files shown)


ProcessLib/GroundwaterFlowProcess.h
```diff
@@ -226,6 +228,20 @@ class GroundwaterFlowProcess : public Process
             assert(false);
     }
 
+    void setInitialConditions(ProcessVariable const& variable)
+    {
+        std::size_t const n = _mesh.getNNodes();
+        for (std::size_t i = 0; i < n; ++i)
+        {
+            MeshLib::Location l(_mesh.getID(), MeshLib::MeshItemType::Node, i);
```
TomFischer: Is there a reason why `l` is not `const`?

ProcessLib/ProcessVariable.cpp
```diff
@@ -18,81 +18,81 @@
...
+
+			// Find corresponding GeoObject
+			std::string const geometrical_set_name =
+			    bc_config.get<std::string>("geometrical_set");
+			std::string const geometry_name =
+			    bc_config.get<std::string>("geometry");
+
+			GeoLib::GeoObject const* const geometry =
+			    geometries.getGeoObject(geometrical_set_name, geometry_name);
+			DBUG(
```
TomFischer: I like the previous formatting better :smiley:

TomFischer:
Apart from the "const" question: :+1:

ProcessLib/InitialCondition.h
```diff
@@ -10,38 +10,75 @@
...
+	virtual double getValue(MeshLib::Node const&) const override
+	{
+		return _value;
+	}
+
+private:
+	double _value;
+};
+
+using ConfigTree = boost::property_tree::ptree;
```
chleh: Can that interfere with other definitions of ConfigTree? At least, there should be no different definitions, but maybe the compiler could complain?

ProcessLib/InitialCondition.h
```diff
@@ -10,38 +10,75 @@
...
+/// Construct a UniformInitialCondition from configuration.
+std::unique_ptr<InitialCondition> createUniformInitialCondition(
```
chleh: Do you want all builders to be free functions? Maybe static member functions (`newInstance()`) of the respective class would be preferable. However, maybe code can be better decoupled when using free functions.

chleh:
:+1: but please answer the questions.

ProcessLib/InitialCondition.h
endJunction: I do prefer free functions because of extensibility, reuse, and class interface minimization.

bilke:
:+1:

PR #829, opened 2015-09-08
Quicksort: fix error.

TomFischer:
Quicksort sorts only the part [beg, end) of the array now.

### DIFF
--- a/BaseLib/quicksort.h
+++ b/BaseLib/quicksort.h
@@ -77,7 +77,7 @@ void quicksort(std::vector<T1*>& array, std::size_t beg, std::size_t end, std::v
 }
 
 template <typename T1, typename T2 = std::size_t>
-void quicksort(T1* & array, std::size_t beg, std::size_t end, T2* & perm)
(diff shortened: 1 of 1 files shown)


BaseLib/quicksort.h
```diff
@@ -41,10 +41,10 @@ void quicksort(std::vector<T1>& array, std::size_t beg, std::size_t end, std::ve
 		});
 
 	// Unzip sorted data.
-	for (std::size_t i = 0; i < data.size(); i++)
+	for (std::size_t i = beg; i < end; i++)
```
chleh: Maybe it's better to have the same loop as in the function below, i.e.
```
for (std::size_t i = 0; i < data.size(); i++)
{
	array[beg+i] = data[i].first;
	perm[beg+i] = data[i].second;
}
```

chleh:
Aren't the two functions pretty identiacel, such that the `vector` case can be handled by passing `array.data()` to the pointer case?

BaseLib/quicksort.h
TomFischer: You are right. Changed it.

PR #830, opened 2015-09-08
Added a funtion to remove a substring from a string.

wenqing:
as Titled.

### DIFF
--- a/BaseLib/StringTools.cpp
+++ b/BaseLib/StringTools.cpp
@@ -143,4 +143,20 @@ void correctScientificNotation(std::string filename, size_t precision)
 	remove(filename.c_str());
 	rename(tmpFilename.c_str(), filename.c_str());
 }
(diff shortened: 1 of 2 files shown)


endJunction:
How is this used?

There are several other functions for dropping file extensions or matching substrings in FileTools.h. Boost's filesystem library has also several more functions for path modifications.

Also a very similar functionality is already provided by boost string algorithms library, e.g. the [erase_last](http://www.boost.org/doc/libs/1_59_0/doc/html/boost/algorithm/erase_last.html).

If I understood correctly the function drops the end of an input string iff the end is the same as another substring; I would name it like `dropSubstringAtEnd` or something along this lines.

wenqing:
@endJunction erase_last seems work, but I have to check whether it can handle the case of an empty substring. Thanks.

PR #831, opened 2015-09-11
[CLI] Added PETSc and MPI stuffs in main function of CLI.

endJunction:
This is a split-off from ufz/ogs#580.

When we add some more init/finalize stuff using preprocessor here it will be a preprocessor-jungle.
Any ideas how to avoid these ifdef's? Feel free to merge otherwise.


bilke:
:+1:

PR #833, opened 2015-09-15
Node partitioned mesh

endJunction:
This is a split-off from ufz/ogs#580

Both new functions are indeed used in the above mentioned PR.

Unfortunatly jenkins went down some time ago and I can only hope for correct compilation on other platforms but linux.
@wenqing Please confirm this code is working for you, then someone merge it.

### DIFF
--- a/MeshLib/NodePartitionedMesh.h
+++ b/MeshLib/NodePartitionedMesh.h
@@ -130,6 +130,7 @@ class NodePartitionedMesh : public Mesh
                     return (node_a->getConnectedNodes().size() <
                             node_b->getConnectedNodes().size());
                 });
+            // Return the number of connected nodes +1 for the node itself.
             return (*it_max_ncn)->getConnectedNodes().size() + 1;
         }
 
--- a/MeshLib/NodePartitionedMesh.h
+++ b/MeshLib/NodePartitionedMesh.h
@@ -16,16 +16,15 @@
 #ifndef NODE_PARTITIONED_MESH_H_
 #define NODE_PARTITIONED_MESH_H_
 
-#include <vector>
+#include <algorithm>
 #include <string>
+#include <vector>
 
 #include "Mesh.h"
+#include "Node.h"
 
 namespace MeshLib
 {
-class Node;
-class Element;
-
 /// A subdomain mesh.
 class NodePartitionedMesh : public Mesh
 {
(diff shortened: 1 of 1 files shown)


MeshLib/NodePartitionedMesh.h
```diff
@@ -115,6 +120,19 @@ class NodePartitionedMesh : public Mesh
...
+        std::size_t getMaximumNConnectedNodesToNode() const
+        {
+            std::vector<Node *>::const_iterator it_max_ncn = std::max_element(
+                _nodes.cbegin(), _nodes.cend(),
+                [](Node const *const node_a, Node const *const node_b)
+                {
+                    return (node_a->getConnectedNodes().size() <
+                            node_b->getConnectedNodes().size());
+                });
+            return (*it_max_ncn)->getConnectedNodes().size() + 1;
```
TomFischer: Why isn't the `size()` as return value enough, i.e. why is it necessary to add one here?

wenqing: The number of connected nodes should include the node itself. Besides, this number is used to preallocate memory for PETSc matrix, the number plus one guarantees an enough space for the sparsity of the matrix arising from FEM.

endJunction: Thanks Wenqing; I added a comment for this.

wenqing:
All tests run except a compilation error caused by
```
#ifdef USE_PETSC
	{
	}
...
#endif
```
which should be in #580, but has been committed to master via #831. Anyway, there is not any test for ogs#PETSc in the master, and everything looks good. Therefore, this PR can be merged.

MeshLib/NodePartitionedMesh.h
```diff
@@ -115,6 +120,20 @@ class NodePartitionedMesh : public Mesh
...
+        std::size_t getMaximumNConnectedNodesToNode() const
+        {
+            std::vector<Node *>::const_iterator it_max_ncn = std::max_element(
+                _nodes.cbegin(), _nodes.cend(),
+                [](Node const *const node_a, Node const *const node_b)
+                {
+                    return (node_a->getConnectedNodes().size() <
+                            node_b->getConnectedNodes().size());
+                });
+            // Return the number of connected nodes +1 for the node itself.
```
wenqing: // Return the number of the connected nodes and the node itself.

bilke:
:+1:

PR #834, opened 2015-09-16
Erase vtk filename extension from the base name.

endJunction:
*Forgot to include this one yesterday into the NodePartitionedMesh PR.*

This makes a file name with ".vtk" extension a valid input for the partitioned mesh reader.

### DIFF
--- a/FileIO/MPI_IO/NodePartitionedMeshReader.cpp
+++ b/FileIO/MPI_IO/NodePartitionedMeshReader.cpp
@@ -73,8 +73,8 @@ MeshLib::NodePartitionedMesh* NodePartitionedMeshReader::read(
     MeshLib::NodePartitionedMesh* mesh = nullptr;
 
     /// Prefix of the filename of the partitioned mesh files.
-    std::string filename_prefix = file_name_base;
-    boost::erase_last(filename_prefix, ".vtk");
+    std::string const filename_prefix =
+        boost::erase_last_copy(file_name_base, ".vtk");
 
     // Always try binary file first
     std::string const fname_new = filename_prefix + "_partitioned_msh_cfg" +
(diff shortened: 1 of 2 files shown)


FileIO/MPI_IO/NodePartitionedMeshReader.cpp
```diff
@@ -71,21 +72,25 @@ MeshLib::NodePartitionedMesh* NodePartitionedMeshReader::read(
 
     MeshLib::NodePartitionedMesh* mesh = nullptr;
 
+    /// Prefix of the filename of the partitioned mesh files.
+    std::string const filename_prefix =
+        boost::erase_last_copy(filename_prefix, ".vtk");
```
TomFischer: The object `filename_prefix` is created here and at the same time used as an argument to the method. Is this working?

wenqing: Not work, line 77 should be changed to
```
      boost::erase_last_copy(file_name_base, ".vtk");
```

wenqing: @endJunction Tested with the above correction. You can amend the commit with that.

endJunction:
Thanks for pointing out the mistake.
I have moved up the file extension handling, such that all of the extensions are handled in the same place, the `readMeshFromFile`.

@wenqing is it OK?

endJunction:
BTW. This code is not being compiled; Why?

wenqing:
:+1: .  All tests run. You can merge it.

The compilation error is caused by
```
#ifdef USE_PETSC
	{
	}
```
in ogs,cpp, which should by in PR #580 , but it has been merged into the master vis  PR #831 (I have pointed out  this issue before in a comment). Anyway, there is not any test for ogs (#PETSC) executable file.

PR #835, opened 2015-09-17
Add a member to PETSc matrix for adding submatrix

wenqing:
Add a member to PETSc matrix for adding submatrix with ghost entry checking.
With the change and the next two PRs, the ghost entry checking in PR #580 (https://github.com/ufz/ogs/pull/580/files#r39689719) under ifdef disappears.

Next two PRs:
1. In order to make consistency with the type of the submatrix argument in member ``add``,   the const modifier of the same argument of member ``add`` of other matrix classes will be removed. 
2. The const modifier of the same argument of member ``add`` of Vector classes will be removed too.

### DIFF
--- a/MathLib/LinAlg/PETSc/PETScMatrix.h
+++ b/MathLib/LinAlg/PETSc/PETScMatrix.h
@@ -164,10 +164,10 @@ class PETScMatrix
         void add(RowColumnIndices<PetscInt> const& indices,
                  T_DENSE_MATRIX &sub_matrix)
         {
-            for(std::size_t i=0; i<indices.non_ghost_local_ids.size(); i++)
+            for(std::size_t i=0; i<indices.ghost_local_ids.size(); i++)
             {
-                const std::size_t non_ghost_id = indices.non_ghost_local_ids[i];
-                sub_matrix->row(non_ghost_id).setZero();
+                const std::size_t ghost_ids = indices.ghost_local_ids[i];
+                sub_matrix->row(ghost_ids).setZero();
             }
             this->add(indices.rows, indices.columns, sub_matrix);
         }
--- a/MathLib/LinAlg/PETSc/PETScMatrix.h
+++ b/MathLib/LinAlg/PETSc/PETScMatrix.h
@@ -159,6 +159,19 @@ class PETScMatrix
             MatSetValue(_A, i, j, value, ADD_VALUES);
         }
 
+        /// Add sub-matrix at positions given by \c indices.
+        template<class T_DENSE_MATRIX>
+        void add(RowColumnIndices<PetscInt> const& indices,
+                 T_DENSE_MATRIX &sub_matrix)
+        {
+            for(std::size_t i=0; i<indices.non_ghost_local_ids.size(); i++)
+            {
+                const std::size_t non_ghost_id = indices.non_ghost_local_ids[i];
+                sub_matrix->row(non_ghost_id).setZero();
+            }
+            this->add(indices.rows, indices.columns, sub_matrix);
+        }
+
         /*!
           \brief         Add a submatrix to this.
           \param row_pos The row indices of the entries of the submatrix.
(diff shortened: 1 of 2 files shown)


MathLib/LinAlg/RowColumnIndices.h
```diff
@@ -14,19 +14,25 @@
...
+// the constructor.
+static std::vector<std::size_t> dummy_non_ghost_local_ids;
 
 template <typename IDX_TYPE>
 struct RowColumnIndices
 {
 	typedef typename std::vector<IDX_TYPE> LineIndex;
-	RowColumnIndices(LineIndex const& rows_, LineIndex const& columns_)
-		: rows(rows_), columns(columns_)
+	RowColumnIndices(LineIndex const& rows_, LineIndex const& columns_,
```
wenqing: So far, I do not have a more better idea about initialization the reference member. In order make the argument of member of add of class matrix concise, add  ``non_ghost_local_ids`` in this truct is adequate.

MathLib/LinAlg/PETSc/PETScMatrix.h
```diff
@@ -159,6 +159,19 @@ class PETScMatrix
...
 
+        /// Add sub-matrix at positions given by \c indices.
+        template<class T_DENSE_MATRIX>
+        void add(RowColumnIndices<PetscInt> const& indices,
+                 T_DENSE_MATRIX &sub_matrix)
+        {
+            for(std::size_t i=0; i<indices.non_ghost_local_ids.size(); i++)
+            {
+                const std::size_t non_ghost_id = indices.non_ghost_local_ids[i];
+                sub_matrix->row(non_ghost_id).setZero();
```
TomFischer: Why are the non ghost node values set to zero?

wenqing: The name is wrong, and it is correct. These ghost ID related  names  were forgotten to be changed after changing the return value of a function about ghost node status. Thanks for pointed it out.

MathLib/LinAlg/RowColumnIndices.h
chleh: At least you could make `dummy_ghost_local_ids` static const, or you make a pointer out of `ghost_local_ids`. Sure, that would be somehow inconsistent as the other members are references, but it removes the need for a static dummy constant.

chleh:
Concerning the announcement of your next two PRs:
I am not very happy that te `add()` methods are about to change the matrix that is added to the global matrix. For example: When I call the `addToGlobal()` method of a local assembler I do not expect thet my locally assembled matrix might be changed while being added.
Sure, most of the time the locally assembled matrix will not be used anymore after being added to the global matrix. But if, e.g., I want to implement a Crank-Nicolson scheme, I also need the matrix from the previous timestep.
I am not entirely sure whether this can be done entirely on the level of global matrices. Probably it can, then my concern is meaningless. But at least we should think about it. (Maybe you already have done so)

chleh:
I have seen that in the file `PETScMatrix.h` there is a line
`MatSetValues(_A, nrows, &row_pos[0], ncols, &col_pos[0], &sub_mat(0,0), ADD_VALUES);`
This line takes the data pointer of `sub_mat` verbatim. Does this have any implications whether local matrices must be row major or column major?

endJunction:
1. I also agree that surprising the user by modification of the local matrix is not a good idea.
2. It seems that we didn't think about this before and all of the ogs6 code is based on the assumption of row major storage order for matrices. This is to be corrected. (Opening issue for this...)

wenqing:
The modification of nulling the ghost node related rows  is only needed for PetscMatrix, and it is the last modification before the local matrix being added when DDC is applied. const modifier  can be kept for other matrix classes. So far this is the best solution within the current frame work.

chleh:
From http://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Mat/MatSetValues.html
> Negative indices may be passed in idxm and idxn, these rows and columns are simply ignored. This allows easily inserting element stiffness matrices with homogeneous Dirchlet boundary conditions that you don't want represented in the matrix. 

Wouldn't it also be a solution to have negative indices for the ghost entries?

wenqing:
@chleh  Thanks. That could be a good solution. I am going to try it.

PR #836, opened 2015-09-17
Corrected an error in PR #834

wenqing:
Corrected a value to a function argument.


wenqing:
I think that I can merge

PR #837, opened 2015-09-17
Fix PETSc compilation

endJunction:
Fixing two ~~three~~ issues here:
 - The findPetsc.cmake could not find my petsc-3.6.1 on archlinux, so the first commit is fixing this for me. Does it work for you?
 - The code did not compile, because releaseEquationMemory() in the GWFProcess does not exist. There is a vivid discussion how to pack this into OO structure (like dtor), but the code is not yet available. Just adding a dummy here (based on https://github.com/ufz/ogs/pull/580)
 - ~~The tests for GWF are missing (in OGS_USE_PETSC configuration); so simply add them. Of course  the mpi ctest will not succeed, but it takes some code out of the https://github.com/ufz/ogs/pull/580 PR.~~

@wenqing Could you please upload the 94783db hash to our ogs6-data place?

I removed the tests from this PR, because those are misplaced in the SimpleTests directory.


ProcessLib/GroundwaterFlowProcess.h
```diff
@@ -316,6 +316,18 @@ class GroundwaterFlowProcess : public Process
         delete _mesh_subset_all_nodes;
     }
 
+    /// Explicitly release memory of global vector, matrix and linear solvers
+    /// for MPI based parallel computing
+    void releaseEquationMemory()
+    {
+    // #ifdef USE_PETSC
+    //     delete _A.release();
```
chleh: If `releaseEquationMemory()` is called before the process is destroyed, then maybe these delete statements should stay, such that memory will really be released. Otherwise: why don't you remove these lines?

endJunction: Yes, better it is!
Sorry, I sometimes forget, that there are other processes.

scripts/cmake/findPETSC/FindPETSc.cmake
```diff
@@ -109,7 +109,13 @@ find_package_multipass (PETSc petsc_config_current
 
 # Determine whether the PETSc layout is old-style (through 2.3.3) or
 # new-style (>= 3.0.0)
-if (EXISTS "${PETSC_DIR}/${PETSC_ARCH}/include/petscconf.h")   # > 2.3.3
+
```
wenqing: Did you update FindPETSc from https://github.com/jedbrown/cmake-modules/blob/master/FindPETSc.cmake?

bilke:
Jenkins test this please.

wenqing:
@endJunction  Permission denied when I copied the file to /data/...

ProcessLib/Process.h
```diff
@@ -32,6 +32,14 @@ class Process
     virtual void initialize() = 0;
     virtual bool solve(const double delta_t) = 0;
 
+    /// Explicitly release memory of global vector, matrix and linear solvers
```
wenqing: Because I already added this function in PR #580, I suggest not to change this file but  to remove the following lines in ogs.cpp
```
	{
	}
```

PR #839, opened 2015-09-18
Move linear solver creation back process construction

endJunction:
Revert a bug fix and fix it in the same way at a different place.
This is better because it is Eigen specific.

This is *important* change unifying the LinearSolver concept.

### DIFF
--- a/MathLib/LinAlg/Eigen/EigenLinearSolver.cpp
+++ b/MathLib/LinAlg/Eigen/EigenLinearSolver.cpp
@@ -32,16 +32,17 @@ public:
     explicit EigenDirectLinearSolver(EigenMatrix::RawMatrixType &A) : _A(A)
     {
         INFO("-> initialize with the coefficient matrix");
-        _solver.compute(A);
-        if(_solver.info()!=Eigen::Success) {
-            ERR("Failed during Eigen linear solver initialization");
-            return;
-        }
     }
 
     void solve(EigenVector::RawVectorType &b, EigenVector::RawVectorType &x, EigenOption &/*opt*/) override
     {
         INFO("-> solve");
+        _solver.compute(_A);
+        if(_solver.info()!=Eigen::Success) {
+            ERR("Failed during Eigen linear solver initialization");
+            return;
+        }
+
         x = _solver.solve(b);
         if(_solver.info()!=Eigen::Success) {
             ERR("Failed during Eigen linear solve");
@@ -62,11 +63,6 @@ public:
     explicit EigenIterativeLinearSolver(EigenMatrix::RawMatrixType &A) : _A(A)
     {
         INFO("-> initialize with the coefficient matrix");
-        _solver.compute(A);
-        if(_solver.info()!=Eigen::Success) {
-            ERR("Failed during Eigen linear solver initialization");
-            return;
-        }
     }
 
     void solve(EigenVector::RawVectorType &b, EigenVector::RawVectorType &x, EigenOption &opt) override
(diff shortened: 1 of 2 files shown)


MathLib/LinAlg/Eigen/EigenLinearSolver.cpp
```diff
@@ -32,16 +32,19 @@ class EigenDirectLinearSolver final : public T_BASE
...
-        if(_solver.info()!=Eigen::Success) {
-            ERR("Failed during Eigen linear solver initialization");
-            return;
-        }
     }
 
     void solve(EigenVector::RawVectorType &b, EigenVector::RawVectorType &x, EigenOption &/*opt*/) override
     {
         INFO("-> solve");
+        if (!_A.isCompressed())
```
chleh: this check is already present in the constructor.

MathLib/LinAlg/Eigen/EigenLinearSolver.cpp
```diff
@@ -32,16 +32,19 @@ class EigenDirectLinearSolver final : public T_BASE
...
+            _A.makeCompressed();
+        _solver.compute(_A);
```
chleh: Maybe someone does not always want to always compute a new decomposition of A for each solve step?

ProcessLib/GroundwaterFlowProcess.h
```diff
@@ -215,6 +215,8 @@ class GroundwaterFlowProcess : public Process
         _A.reset(_global_setup.createMatrix(_local_to_global_index_map->dofSize()));
         _x.reset(_global_setup.createVector(_local_to_global_index_map->dofSize()));
         _rhs.reset(_global_setup.createVector(_local_to_global_index_map->dofSize()));
+        _linearSolver.reset(new typename GlobalSetup::LinearSolver(*_A));
```
chleh: This will work for elliptic problems, but for transient problems the `reset()` has to be done in any iteration.

MathLib/LinAlg/Eigen/EigenLinearSolver.cpp
```diff
@@ -32,16 +32,19 @@ class EigenDirectLinearSolver final : public T_BASE
     explicit EigenDirectLinearSolver(EigenMatrix::RawMatrixType &A) : _A(A)
```
chleh: Why not get rid of the argument `A` entirely?

chleh:
I guess, you (@endJunction, @norihiro-w) know the reason for the commit "fix a bug in linear solver creation. it should be created after assemblying A." (0237275a0e11860556ee7540af9f01baa3bc7df3) that this PR refers to.
So far I did not know it, but by chance I probably learned about it. From the Eigen docs:
> Derived& compute (const EigenBase<MatrixDerived> &A)
> Initializes the iterative solver with the matrix A for further solving Ax=b problems.

> Currently, this function mostly initializes/computes the preconditioner. In the future we might, for instance, implement column reordering for faster matrix vector products.

> Warning
>    this class stores a reference to the matrix A as well as some precomputed values that depend on it. Therefore, if A is changed this class becomes invalid. Call compute() to update it with the new matrix A, or modify a copy of A.

(http://eigen.tuxfamily.org/dox-devel/classEigen_1_1IterativeSolverBase.html#a7dfa55c55e82d697bde227696a630914)

That is, the former solution is very likely to be applied wrong by users writing own processes. The new solution introduced by this PR is much better in this regard: Compute and solve step are tied together tightly, so the user cannot accidentally invalidate the solver in between.

Maybe it should be pointed out more strongly in the PR title that this is a serious issue and not only some second rank change.

MathLib/LinAlg/Eigen/EigenLinearSolver.cpp
endJunction: This is possible, but then we need to extend the LinearSolver concept.

MathLib/LinAlg/Eigen/EigenLinearSolver.cpp
endJunction: Must suffice the LinearSolver concept

endJunction:
The unanswered questions were discussed with Christoph. Please post new questions after this comment ;)

PR #846, opened 2015-09-22
0d element

endJunction:
This adds a point element to the mesh lib, suitable shape function, and makes it available for the local assemblers.

The reason behind this element is that the boundary of a line mesh consists of point elements and to implement Neumann boundary conditions, for example, integration over such elements must be performed.

The only work-in-progress item is to update the ctest suite.
- [ ] Update ctests.

### DIFF
--- a/NumLib/Fem/Integration/IntegrationPoint.h
+++ b/NumLib/Fem/Integration/IntegrationPoint.h
@@ -1,0 +1,88 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2015, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef NUMLIB_INTEGRATIONPOINT_H_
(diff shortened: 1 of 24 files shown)


NumLib/Fem/Integration/IntegrationPoint.h
```diff
@@ -0,0 +1,88 @@
+/**
+ * \copyright
```
wenqing: Point wise Neumann BC dose not need integration, the value of the BC can be input directly.

wenqing:
Looks like this 0 element is only for the consistency of interface for BC. Why not to use Node directly? Find the nodes in the line element by the search function, and assign the BC values directly from the input data. Besides, the Neumann type BC for node (potin-wise source term) can be directly  put to the input data and added to the RHS vector without any additional processing.

TomFischer:
> Looks like this 0 element is only for the consistency of interface for BC.

This is a good point. We need consistent interfaces and clear structures. The node should not implement the interface of an element.

endJunction:
Zero sized arrays (occuring at some point of creation of 0x0 matrices) are not allowed by c++. This PR needs a general rework.

PR #847, opened 2015-09-24
Make solver interface unique

wenqing:
Make solver interface unique by adding a string argument to the constructor of linear solver classes, which could be used as a prefix for command line options of linear solvers if they are available.
PS: removed the calling ``releaseEquationMemory()`` from main to avoid compilation error.

### DIFF
--- a/Applications/CLI/ogs.cpp
+++ b/Applications/CLI/ogs.cpp
@@ -153,10 +153,6 @@ int main(int argc, char *argv[])
 	solveProcesses(project);
 
 #ifdef USE_PETSC
-	for (auto p_it = project.processesBegin(); p_it != project.processesEnd(); ++p_it)
-	{
-		(*p_it)->releaseEquationMemory();
-	}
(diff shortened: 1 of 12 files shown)


Applications/CLI/ogs.cpp
```diff
@@ -153,10 +153,6 @@ int main(int argc, char *argv[])
 	solveProcesses(project);
 
 #ifdef USE_PETSC
-	for (auto p_it = project.processesBegin(); p_it != project.processesEnd(); ++p_it)
```
wenqing: Removed because the associated implementation is not merged yet.

wenqing:
Changed the type of the new argument  from reference to value.

endJunction:
:fast_forward:

This change will be reversed as soon as we start handling the linear solver configurations (parsing project's config and passing the {Lis,Eigen,Petsc}Options to corresponding linear solver.

PR #849, opened 2015-09-24
Global index type

endJunction:
The important changes start at *Add IndexType to global matrix and vector types.* commit.
The preceding part is only a cleanup w/o any logic changes.

A concept of a global index type is introduced.
The type of global indices stored in the DOF table must correspond to the index type used for addressing global matrix/vector entries. And the currently used `std::size_t` type requires additional type casts which are not always possible w/o copies.
The decision is to use the index types given by global matrices/vectors from corresponding linear equation system solver libraries.

### DIFF
--- a/MathLib/LinAlg/Eigen/EigenMatrix.h
+++ b/MathLib/LinAlg/Eigen/EigenMatrix.h
@@ -62,14 +62,14 @@ public:
     void setZero()
     {
         auto const N = _mat.nonZeros();
-        for (std::remove_const<decltype(N)>::type i=0; i<N; i++)
+        for (auto i = decltype(N){0}; i<N; i++)
             _mat.valuePtr()[i] = 0;
         // don't use _mat.setZero(). it makes a matrix uncompressed
     }
 
     /// set a value to the given entry. If the entry doesn't exist, this class
     /// dynamically allocates it.
-    int setValue(std::size_t row, std::size_t col, double val)
+    int setValue(IndexType row, IndexType col, double val)
     {
         _mat.coeffRef(row, col) = val;
         return 0;
@@ -77,7 +77,7 @@ public:
 
     /// add a value to the given entry. If the entry doesn't exist, the value is
     /// inserted.
-    int add(std::size_t row, std::size_t col, double val)
+    int add(IndexType row, IndexType col, double val)
     {
         assert(row < getNRows() && col < getNCols());
         _mat.coeffRef(row, col) += val;
@@ -87,7 +87,7 @@ public:
     /// Add sub-matrix at positions \c row_pos and same column positions as the
     /// given row positions. If the entry doesn't exist, the value is inserted.
     template<class T_DENSE_MATRIX>
-    void add(std::vector<std::size_t> const& row_pos,
+    void add(std::vector<IndexType> const& row_pos,
             const T_DENSE_MATRIX &sub_matrix,
             double fkt = 1.0)
     {
@@ -97,7 +97,7 @@ public:
     /// Add sub-matrix at positions given by \c indices. If the entry doesn't exist,
     /// this class inserts the value.
     template<class T_DENSE_MATRIX>
-    void add(RowColumnIndices<std::size_t> const& indices,
+    void add(RowColumnIndices<IndexType> const& indices,
             const T_DENSE_MATRIX &sub_matrix,
             double fkt = 1.0)
     {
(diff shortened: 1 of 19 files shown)


MathLib/LinAlg/Eigen/EigenMatrix.h
```diff
@@ -76,9 +77,9 @@ class EigenMatrix final
 
     /// add a value to the given entry. If the entry doesn't exist, the value is
     /// inserted.
-    int add(std::size_t row, std::size_t col, double val)
+    int add(IndexType row, IndexType col, double val)
     {
-        assert(row < getNRows() && col < getNCols());
+        assert(row < _mat.rows() && col < _mat.cols());
```
chleh: Maybe now you should also assert that row and col are not neagative.

ProcessLib/NumericsConfig.h
```diff
@@ -103,5 +103,18 @@ using GlobalSetupType =
...
+static_assert(std::is_integral<detail::GlobalVectorType::IndexType>::value,
+              "The index type for global vectors is not an integral type.");
+static_assert(std::is_same<detail::GlobalMatrixType::IndexType,
+                           detail::GlobalVectorType::IndexType>::value,
+              "The global matrix and vector index types do not match.");
+// Both types are integral types and equal, define a single GlobalIndexType.
+
+/// A type used for indexing of global vectors and matrices. It is equal to the
+/// GlobalMatrixType::IndexType and the GlobalVectorType::IndexType.
+using GlobalIndexType = detail::GlobalMatrixType::IndexType;
```
wenqing: :+1:

wenqing:
:fast_forward:

MathLib/LinAlg/Eigen/EigenMatrix.h
endJunction: I checked EigenMatrix.h code and this assertions are already present there. So I removed our assertions altogether.
See [Eigen/SparseMatrix.h:184](https://bitbucket.org/eigen/eigen/src/25c4b01d1bbce878477809eb86da86d2d2906d53/Eigen/src/SparseCore/SparseMatrix.h?at=default&fileviewer=file-view-default#SparseMatrix.h-184) for example.

endJunction:
Just a rebase.

wenqing:
Is it ready to be merged?

endJunction:
From my side there are no further changes planned. We will wait for a review.

chleh:
:+1:

endJunction:
Beautified three `for` loops in LisMatrix.

MathLib/LinAlg/Eigen/EigenMatrix.h
```diff
@@ -170,18 +168,17 @@ class EigenMatrix final
...
+                      const T_DENSE_MATRIX& sub_matrix, double fkt)
 {
-    const std::size_t n_rows = row_pos.size();
-    const std::size_t n_cols = col_pos.size();
-    for (std::size_t i = 0; i < n_rows; i++) {
-        const std::size_t row = row_pos[i];
-        for (std::size_t j = 0; j < n_cols; j++) {
-            const std::size_t col = col_pos[j];
+    const IndexType n_rows = row_pos.size();
+    const IndexType n_cols = col_pos.size();
```
TomFischer: `row_pos` and `col_pos` are of type `std::vector`. The `size()` method returns `size_type` which has not to be equal to `IndexType`. So maybe these changes are not necessary?

MathLib/LinAlg/Lis/LisVector.h
```diff
@@ -22,97 +22,95 @@
...
+
+	/// return a start index of the active data range
+	std::size_t getRangeBegin() const { return 0; }
+	/// return an end index of the active data range
+	std::size_t getRangeEnd() const { return this->size(); }
+	/// set all values in this vector
+	LisVector& operator=(double v);
+
+	/// access entry
+	double operator[](std::size_t rowId) const { return get(rowId); }
```
TomFischer: I wonder why the parameter is not of type `IndexType`? The same question arises for the `set`, `get` and `add` methods.

TomFischer:
In the "Eigen files" `IndexType` is used almost everywhere. In the "Lis files" `LisInt` is used. For consistency reasons I would prefer `IndexType` also in the  "Lis files".

MathLib/LinAlg/Lis/LisMatrix.h
```diff
@@ -62,7 +64,8 @@ class LisMatrix
      * @param col_idx  array of column indexes
      * @param data     the non-zero entry values
      */
-    LisMatrix(std::size_t n_rows, int nonzero, int* row_ptr, int* col_idx, double* data);
+    LisMatrix(std::size_t n_rows, int nonzero, IndexType* row_ptr, int* col_idx,
```
TomFischer: I would expect `IndexType* col_idx`, too.

MathLib/LinAlg/Lis/LisVector.h
```diff
@@ -22,97 +22,95 @@
...
+
+	/// copy constructor
+	LisVector(LisVector const& src);
+
+	/**
+	 *
+	 */
+	virtual ~LisVector();
+
+	/// return a vector length
```
TomFischer: The docu sounds a little bit strange to me. Why not: "return the length of the vector".
The same holds true for the docu of `getRangeBegin()` and `getRangeEnd()`.

Why do this methods return `std::size_t`? I wonder if `IndexType` would be more appropriate?

endJunction: Didn't touch the docu here but I agree that the English prose is not our strength.

Regarding the std::size_t return type: This same as for all other (global) vectors. I checked the documentation of eigen and petsc, the do return the IndexType for sizes too. Suggest to update it in next step. But there will be some point at which our (for example meshlib) types do not agree with the linare solver library types and a conversion will happen.

MathLib/LinAlg/Lis/LisMatrix.h
endJunction: Fixed.

PR #851, opened 2015-09-25
Added OgsInitialize and OgsFinalize.

wenqing:
As titled, two functions were added to hold Initialize/Finalize functions from different packages. In addition, explicitly releasing MPI related memory is moved to OgsFinalize.

### DIFF
--- a/Applications/ApplicationsLib/OgsInitFinalize.h
+++ b/Applications/ApplicationsLib/OgsInitFinalize.h
@@ -28,6 +28,8 @@
 
 #include "Applications/ApplicationsLib/ProjectData.h"
 
+namespace detail
+{
 /// Initialize MPI, PETSc, LIS or any other database from third party
 /// packages.
 void OgsInitialize(int argc, char *argv[])
@@ -82,4 +84,5 @@ void OgsFinalize(ProjectData &project)
 	(void) project;
 }
 
+} // end of detail
 #endif  // OGS_INIT_FINALIZE_H
--- a/Applications/ApplicationsLib/OgsInitFinalize.h
+++ b/Applications/ApplicationsLib/OgsInitFinalize.h
@@ -59,7 +59,7 @@ void OgsFinalize(ProjectData &project)
 {
 #ifdef USE_PETSC
 	// Since the global matrix, vector and linear equation in ProjectData
-	// are defined as smarter point type (unique_ptr or might be shared_ptr)
+	// are defined as smart pointer types (unique_ptr or might be shared_ptr)
 	// variables, their memory occupations are automatically released at
 	// the end of ProjectData terminated, i.e. the end of the main program.
 	// However if  the global matrix, vector and linear equation are created
--- a/Applications/ApplicationsLib/OgsInitFinalize.h
+++ b/Applications/ApplicationsLib/OgsInitFinalize.h
@@ -28,9 +28,8 @@
 
 #include "Applications/ApplicationsLib/ProjectData.h"
 
-namespace BaseLib
-{
-
+/// Initialize MPI, PETSc, LIS or any other database from third party
+/// packages.
 void OgsInitialize(int argc, char *argv[])
 {
 #ifdef USE_MPI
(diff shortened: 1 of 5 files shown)


endJunction:
This is a first step towards eliminating the PETSc specific code in processes.

@wenqing Please add a note why the process' destructor is not sufficient in the case of MPI parallelization.

Following suggestions:
 - Move this into ApplicationsLib, because it has very limited usage outside of any applications.
 - Replace finalize call with a destructor of an object.
 - The functions are related to the linear solvers libraries; call the class something like LinearSolverSystem.

wenqing:
@endJunction Moved it to ApplicationsLib, and added more documentation. 
As for replacing finalize call with a destructor of an object, turn the function to an object works. However, an instance of the object has to be created and never used, it is  just for the destructor to be called automatically. Therefore, I think have one single function of OgsInitFinalize is straightforward, and it also has a consistency meaning of the functions called inside this function.

ProcessLib/GroundwaterFlowProcess.h
```diff
@@ -317,6 +317,24 @@ class GroundwaterFlowProcess : public Process
...
+    /// Explicitly release memory of global vector, matrix and linear solvers
+    /// for MPI based parallel computing
+    /// Since the global matrix, vector and linear equation in ProjectData
+    /// are defined as smarter point type (unique_ptr or might be shared_ptr)
+    /// variables, their memory occupations are automatically released at
+    /// the end of ProjectData terminated, i.e. the end of the main program.
+    /// However if  the global matrix, vector and linear equation are created
+    /// under MPI environment, their memory occupations must be released before
+    /// calling of MPI_Finalize, PetscFinalize. That is why the following fucntion
+    /// is needed.
```
bilke: Documentation is already in base class.

Applications/ApplicationsLib/OgsInitFinalize.h
```diff
@@ -0,0 +1,88 @@
...
+}
+
+/// Release MPI related memory in project, and finalize MPI, PETSc,
+/// LIS or any other database from third party
+/// packages.
+void OgsFinalize(ProjectData &project)
+{
+#ifdef USE_PETSC
+	// Since the global matrix, vector and linear equation in ProjectData
+	// are defined as smarter point type (unique_ptr or might be shared_ptr)
```
bilke: Typo: "smart pointer types"

ProcessLib/GroundwaterFlowProcess.h
wenqing: Removed.

Applications/ApplicationsLib/OgsInitFinalize.h
wenqing: Corrected.

Applications/CLI/ogs.cpp
```diff
@@ -85,14 +78,9 @@ int main(int argc, char *argv[])
...
-#ifdef USE_MPI
-	MPI_Init(&argc, &argv);
-#endif
-
-#ifdef USE_PETSC
-	char help[] = "ogs6 with PETSc \n";
-	PetscInitialize(&argc, &argv, nullptr, help);
-#endif
+	// Initialize MPI, PETSc, LIS or any other database from third party
+	// packages.
```
bilke: I would remove this comment and also the one on `OgsFinalize()` because the name already implies what the methods do and the same comment is already given in the method documentation.

bilke:
Beside my last comment: :+1:

Applications/CLI/ogs.cpp
wenqing: Removed.

Applications/ApplicationsLib/OgsInitFinalize.h
```diff
@@ -0,0 +1,88 @@
...
+	}
+	PetscFinalize();
+#endif
+
+#ifdef USE_MPI
+	MPI_Finalize();
+#endif
+
+#ifdef USE_LIS
+	lis_finalize();
```
chleh: LIS is initialized after Petsc and MPi and is finalized after them. Thus, it seems that LIS is completely independent from the other two.
Is it necessary to have the LIS stuff in this Petsc-related class?

Applications/ApplicationsLib/OgsInitFinalize.h
```diff
@@ -0,0 +1,88 @@
...
+void OgsFinalize(ProjectData &project)
+{
+#ifdef USE_PETSC
+	// Since the global matrix, vector and linear equation in ProjectData
+	// are defined as smart pointer types (unique_ptr or might be shared_ptr)
+	// variables, their memory occupations are automatically released at
+	// the end of ProjectData terminated, i.e. the end of the main program.
+	// However if  the global matrix, vector and linear equation are created
+	// under MPI environment, their memory occupations must be released before
+	// calling of MPI_Finalize, PetscFinalize. That is why the following loop is needed.
```
chleh: This problem can be circumvented more elegantly by making `_A`, `_rhs` etc. `shared_ptr`s and using the aliasing constructor (case 8 on http://en.cppreference.com/w/cpp/memory/shared_ptr/shared_ptr).

wenqing: Thanks. Will try this.

PR #856, opened 2015-10-05
Linear solver library setup

endJunction:
Move initialization of linear solver libraries into own class `LinearSolverLibrarySetup` (which is chosen by preprocessor for now and will be chosen by cmake soon)  with purpose of correct initialization and *shutdown* of the external libraries.
This resolves the issue of core dumps in the pFEM4 PR (https://github.com/ufz/ogs/pull/580) because of improper order of matrix/vectors/linear-solver destruction. No explicit `releaseEquationsMemory()` workaround is longer necessary.

Logog is set up same way too.

 - First fix some issues on logog first,
 - then move logog initialization to LogogSetup,
 - finally implement LinearSolverLibrarySetup for PETSc, LIS, and a default (which is empty).

This closes https://github.com/ufz/ogs/pull/851.

### DIFF
--- a/Applications/ApplicationsLib/LinearSolverLibrarySetup.h
+++ b/Applications/ApplicationsLib/LinearSolverLibrarySetup.h
@@ -1,0 +1,64 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2015, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef APPLICATIONSLIB_LINEARSOLVERLIBRARYSETUP_H_
+#define APPLICATIONSLIB_LINEARSOLVERLIBRARYSETUP_H_
+
+namespace ApplicationsLib
+{
+
+/// The LinearSolverLibrarySetup takes care of proper initialization and
+/// shutting down of an external linear solver library. The concrete
+/// implementation is chosen by the build system.
+/// An object of this class must be created at the begining of the scope where
+/// it is used. When the scope closes (or the object is destroyed explicitly)
+/// library shutting down functions are automatically called.
+/// The default implementation is empty providing polymorphic behaviour when
+/// using this class.
+
+#if defined(USE_PETSC)
+#include <petsc.h>
+#include <mpi.h>
+struct LinearSolverLibrarySetup final
+{
+	LinearSolverLibrarySetup(int argc, char* argv[])
+	{
+		MPI_Init(&argc, &argv);
+		char help[] = "ogs6 with PETSc \n";
+		PetscInitialize(&argc, &argv, nullptr, help);
+	}
+
+	~LinearSolverLibrarySetup()
+	{
+		PetscFinalize();
+		MPI_Finalize();
+	}
+};
+#elif defined(USE_LIS)
+#include <lis.h>
+struct LinearSolverLibrarySetup final
+{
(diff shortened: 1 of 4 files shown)


endJunction:
Thanks to Christoph (@chleh) for reminding us about allocation/deallocation order on the stack!

bilke:
:+1:

Applications/ApplicationsLib/LinearSolverLibrarySetup.h
```diff
@@ -0,0 +1,64 @@
...
+/// using this class.
+
+#if defined(USE_PETSC)
+#include <petsc.h>
+#include <mpi.h>
+struct LinearSolverLibrarySetup final
+{
+	LinearSolverLibrarySetup(int argc, char* argv[])
+	{
+		MPI_Init(&argc, &argv);
```
wenqing: Removed this call, let it done by PetscInitialize.

Applications/ApplicationsLib/LinearSolverLibrarySetup.h
```diff
@@ -0,0 +1,64 @@
...
+		char help[] = "ogs6 with PETSc \n";
+		PetscInitialize(&argc, &argv, nullptr, help);
+	}
+
+	~LinearSolverLibrarySetup()
+	{
+		PetscFinalize();
+		MPI_Finalize();
```
wenqing: This can be also removed. Or remove this, keep MPI_init.

Applications/ApplicationsLib/LinearSolverLibrarySetup.h
```diff
@@ -0,0 +1,64 @@
...
+/// An object of this class must be created at the begining of the scope where
+/// it is used. When the scope closes (or the object is destroyed explicitly)
+/// library shutting down functions are automatically called.
+/// The default implementation is empty providing polymorphic behaviour when
+/// using this class.
+
+#if defined(USE_PETSC)
+#include <petsc.h>
+#include <mpi.h>
+struct LinearSolverLibrarySetup final
```
wenqing: PETSc is not only for linear solver, and also MPI can be also initialized here as an individual option. Therefore, the term of  LinearSolverLibrarySetup is not adequate.

Applications/ApplicationsLib/LinearSolverLibrarySetup.h
endJunction: What is the difference between explicitly calling MPI_Init versus let it be called by PetscInitialize?

From petsc [documentation](http://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Sys/PetscInitialize.html): "PetscInitialize() calls MPI_Init() if that has yet to be called"
And [later](http://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Sys/PetscFinalize.html#PetscFinalize): "MPI_Finalize() is called only if the user had not called MPI_Init() before calling PetscInitialize()."
Se the constructor and destructor are consistent regarding mpi initialization/shutdown.

Applications/ApplicationsLib/LinearSolverLibrarySetup.h
endJunction: I agree.

endJunction:
I'm having some issues with LIS at the moment. Resolving them and the will reset the review flag.

endJunction:
OK. Everything is green now.

Applications/ApplicationsLib/LinearSolverLibrarySetup.h
wenqing: It solved by discussion.

PR #858, opened 2015-10-07
Added CMake option OGS_CPU_ARCHITECTURE which defaults to native

bilke:
See http://docs.opengeosys.org/docs/devguide/advanced/configuration-options

This will enable redistributable binaries build with Jenkins.


TomFischer:
:+1:

PR #862, opened 2015-10-13
Small bug fixes

TomFischer:
- `GEOObjects`: Omit crash in adding an empty vector of points.
-  `createSurfaceMesh`: Fix a logical bug. Mesh size is as expected now.
- `NetCdfConfigureDialog`: Fix a mem leak and const correctness.


endJunction:
Looks good!

bilke:
:+1:

PR #866, opened 2015-10-15
Separate GEOObjects and GEOModels.

endJunction:
Replace `GEOObjects <- GEOModels` polymorphism by callbacks. The callback implementation closely follows the "command pattern".

I have replaced the logic in some places with equivalent. In one case there is an extra behavior in `GEOModels`, which was not implemented in `GEOObjects`. The place I will mark below.

Because there is no automatic testing for this code (but for `GEOObjects`) I would like to ask you to actually compile this branch and try out the DataExplorer in your favorite setup.


Applications/DataExplorer/DataView/GEOModels.cpp
```diff
@@ -74,114 +81,87 @@ void GEOModels::updateGeometry(const std::string &geo_name)
...
-			sfc->push_back(surfaces[i]);
-		this->addSurfaceVec(sfc, name);
+	    std::vector<GeoLib::Surface*>* sfc = new std::vector<GeoLib::Surface*>;
+	    for (std::size_t i = 0; i < surfaces.size(); i++)
+	        sfc->push_back(surfaces[i]);
+	    _geo_objects->addSurfaceVec(sfc, name);
 	}
-
-	return ret;
+	*/
```
endJunction: Not sure what to do about this special handling. It is special because the extra `addSurfaceVec` calls were executed only when the code was compiled with `OGS_BUILD_GUI` flag set.

Applications/DataExplorer/DataView/GEOModels.cpp
```diff
@@ -74,114 +81,87 @@ void GEOModels::updateGeometry(const std::string &geo_name)
 void GEOModels::removeGeometry(std::string geo_name, GeoLib::GEOTYPE type)
```
TomFischer: I wonder if the first argument can be given as `const&`?

Applications/DataExplorer/DataView/GEOModels.cpp
```diff
@@ -74,114 +81,87 @@ void GEOModels::updateGeometry(const std::string &geo_name)
...
 {
-	GEOObjects::addPointVec(points, name, name_pnt_id_map, eps);
-	_geoModel->addPointList(QString::fromStdString(name), *this->getPointVecObj(name));
+	_geoModel->addPointList(QString::fromStdString(name),
+	                        *_geo_objects->getPointVecObj(name));
 	emit geoDataAdded(_geoModel, name, GeoLib::GEOTYPE::POINT);
 }
 
-bool GEOModels::removePointVec( const std::string &name )
+void GEOModels::removePointVec(std::string const& name)
```
TomFischer: Is there any reason why the return value changed from `bool` to `void`?

bilke:
Cool stuff! I did not understand how the callbacks-stuff works but it works..

endJunction:
The callbacks work so:
 - there is a [bunch of "default" callbacks](https://github.com/endJunction/ogs/blob/GeoModels/GeoLib/GEOObjects.h#L67) in GEOObjects doing nothing.
 - `GEOObjects::addPointVec()` is calling the corresponding callback after adding the points.
That's all for the GEOObjects part.

Now instead of the default callbacks, the `GEOModels` class is providing [special callbacks](https://github.com/endJunction/ogs/blob/GeoModels/Applications/DataExplorer/DataView/GEOModels.h#L111) which are derived from the `GEOObjects::Callbacks`. The polymorphism kicks in and the qt stuff is updated.

Should there be another GUI, it would only derive special callbacks from `GEOObjects::Callbacks` and get the update information.

That's all!

Applications/DataExplorer/DataView/GEOModels.cpp
endJunction: - The result of any of the callbacks is not evaluated in GEOObjects.
 - There is no need to mimic the GEOObjects interface anymore.
 - The gui gives no feedback anyway; GeoTreeModel::removeGeoList has return type `void`.

endJunction:
Just a rebase and one more commit.

Applications/DataExplorer/DataView/GEOModels.cpp
TomFischer: Thanks for explanation.

bilke:
Thanks, I just missed the [instantiation of the derived callbacks](https://github.com/endJunction/ogs/blob/GeoModels/Applications/DataExplorer/DataView/GEOModels.cpp#L44).

bilke:
:+1:

Applications/DataExplorer/DataView/GEOModels.cpp
TomFischer: @rrink, @bilke and I would move the code to the class `GEOObjects`.

Applications/DataExplorer/DataView/GEOModels.h
```diff
@@ -42,6 +42,8 @@ class GEOModels : public QObject, public GeoLib::GEOObjects
 	GEOModels(QObject* parent = 0);
 	~GEOModels();
 
+	void setGEOObjects(GeoLib::GEOObjects* geo_objects);
```
TomFischer: Why not initialize the member `_geo_objects` using the constructor and omit the method?

Applications/DataExplorer/DataView/SHPImportDialog.cpp
```diff
@@ -24,11 +24,20 @@
...
-SHPImportDialog::SHPImportDialog(std::string filename, GEOModels* geoModels, QDialog* parent)
-	: QDialog(parent), _buttonBox(NULL), _layout(NULL), _shpContentLabel(NULL), _nameLabel(NULL),
-	  _listName(new QLineEdit()), _choice1(NULL), _choice2(NULL), _filename(filename),
-	  _fileType(0),
-	  _shpInterface(new FileIO::SHPInterface(geoModels))
+SHPImportDialog::SHPImportDialog(std::string filename,
+                                 GeoLib::GEOObjects* geo_objects,
+                                 QDialog* parent)
+    : QDialog(parent),
+      _buttonBox(NULL),
```
TomFischer: It is possible to substitute `NULL` with `nullptr` in the complete initializer list.

Applications/DataExplorer/DataView/SHPImportDialog.h
```diff
@@ -40,7 +42,8 @@ class SHPImportDialog : public QDialog
 
 public:
 	/// Constructor
-	SHPImportDialog(std::string filename, GEOModels* geoModels, QDialog* parent = 0);
+	SHPImportDialog(std::string filename, GeoLib::GEOObjects* geo_objects,
```
TomFischer: `filename` can be a `const &`.

endJunction: Now it's passed by value and moved into _filename.

endJunction:
I did not simplified any of the GEOObjects algorithms, though it is very tempting. Let's have a look on them after introducing unique_ptr to TemplateVec.

PR #867, opened 2015-10-17
Use unique_ptr in TemplateVec for data.

endJunction:
Follow up of https://github.com/ufz/ogs/pull/866 *Separate GEOObjects and GEOModels.*
This improves (and actually fixes few memory leaks) memory management.
Especially use of unique_ptr indicates ownership of the data, whereas raw memory pointer are just pointers and are not allowed to be released or reset.

Main change is only in `TemplateVec` class; all other changes just required changes to keep the code correct.
The bulk of changes follow the same pattern:
 - replace `std::vector<Point*>* points` with `std::unique_ptr<std::vector<Point*>>`
 - move in the call of e.g. `geo_objects.addPointVec(std::move(points), ...)`
 - the points pointer is no longer usable (ownership transferred to `geo_objects`) and if needed is fetched by `auto points = geo_objects.getPointVec(name)`.

### DIFF
--- a/Applications/DataExplorer/mainwindow.cpp
+++ b/Applications/DataExplorer/mainwindow.cpp
@@ -545,11 +545,14 @@ void MainWindow::loadFile(ImportFileType::type t, const QString &fileName)
 	{
 		if (fi.suffix().toLower() == "txt") // GMS borehole files
 		{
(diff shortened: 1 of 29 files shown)


TomFischer:
:+1:

endJunction:
Just a rebase.
@bilke Would you please have a look on the macs?

PR #868, opened 2015-10-17
Fix SimpleTests.

endJunction:
The targets in SimpleTests are not checked automatically, so this was forgotten.

The PR fixes 
 - [x] the include paths
 - [x] MatMult, MatTestRemoveRowsCols, MatVecMultPthreads and all SolverTests/ compilation failures with
`make MatMult` fails with `MathLib/LinAlg/Sparse/CRSMatrix.h:26:10: fatal error: 'sparse.h' file not found`
 - [x] MeshSearchTest (there are multiple compiler errors regarding GeoLib::Point, sqrDist, etc.)


TomFischer:
The added commits should fix the second and third point of the list mentioned in the PR description.

endJunction:
Thanks Tom!
:fast_forward:

PR #869, opened 2015-10-19
Parallel VTU output

wenqing:
With this PR, the parallel output of the parallel FEM computation is enabled. When the output is required after solving the the equations, the results of the domain are written into individual binary or ASCII vtu files simultaneously by the corresponding compute cores, and then a pvtu file is created for merging the vtu files in Paraview. In such a way, no post-process is required.

PS:
@bilke  Could you please tell me how to enable the vtk module that contains  vtkXMLPUnstructuredGridWriter.h in Cmake? I use the vtk module on eve to compile the code.

@endJunction  property_name in GroundwaterFlowProcess::post() was changed to  "Head [m]". Due to this change, the reference results of ctest have to be changed. I can update the reference results either in this PR or in the next one.


ProcessLib/GroundwaterFlowProcess.h
```diff
@@ -313,52 +312,51 @@ class GroundwaterFlowProcess : public Process
...
-            std::ofstream os(output_file_name);
-            os << "SCALARS HEAD double 1"<<endl;
-            os << "LOOKUP_TABLE default"<< endl;
-            for (std::size_t i = 0; i < u.size(); ++i)
-                 os << u[i] << "\n";
-            os.close();
-        }
-#else
-        std::string const property_name = "Result";
+        std::string const property_name = "Head [m]";
```
wenqing: All ctests run successfully with the changes in this function. However their reference results, in which the string of  'Result' is replaced with "Head [m]",  need to be updated.

MathLib/LinAlg/Dense/DenseVector.h
```diff
@@ -75,6 +75,14 @@ class DenseVector : public std::valarray<T>
 		}
 	}
 
+	/// get entry values to an array
+	void getValues(T u[])
+	{
+		for (std::size_t i=0; i<this.size(); ++i) {
```
TomFischer: This line is not compiling.

bilke:
To link the parallel VTK IO library add `vtkIOParallelXML` to line 11 in `ExternalProjectCatalyst.cmake`:

    set(PARAVIEW_MODULES vtkIOXML vtkIOParallelXML)

MathLib/LinAlg/Dense/DenseVector.h
wenqing: I see. clang does not compile this line, I am going to check it.

wenqing:
@bilke  Thanks. Added but there came a compilation error
```
*** No rule to make target `vtkIOParallelXML'.  Stop.
```
Is there any setting in cmake for vtkIOParallelXML?

bilke:
The module `vtk/6.1.0-2_gcc_4.8.1` does not provide the parallel xml lib. I have created a new module:

```bash
module load /opt/modulefiles/vtk/6.3.0_openmpi_1.8.2
```

In `ExternalProjectCatalyst.cmake` please add it like this:

```cmake
set(PARAVIEW_MODULES vtkIOXML)
if(OGS_USE_MPI)
	set(PARAVIEW_MODULES ${PARAVIEW_MODULES} vtkIOParallelXML)
endif()
```

Please also not that the automatic VTK compilation when VTK was not found is not yet updated, see ufz/catalyst-io#1.

MathLib/LinAlg/Dense/DenseVector.h
wenqing: Fixed.

wenqing:
@bilke Added module vtkIOParallelXML. If I load  module vtk/6.1.0-2_gcc_4.8.1, I do not need to modify ExternalProjectCatalyst.cmake to compile the source code under USE_PETSC. However, if  module vtk/6.1.0-2_gcc_4.8 is not loaded and vtkIOParallelXML is added, the compiler gives an error of
```
*** No rule to make target `vtkIOParallelXML'.  Stop.
```
I guess that the update of ufz/catalyst-io#1 will have vtkIOParallelXML sources.

FileIO/readMeshFromFile.cpp
```diff
@@ -44,8 +44,16 @@ MeshLib::Mesh* readMeshFromFile(const std::string &file_name)
...
 #ifdef USE_PETSC
 	NodePartitionedMeshReader read_pmesh(PETSC_COMM_WORLD);
+
+	std::string file_name_ext;
+	if (BaseLib::hasFileExtension("msh", file_name))
+	   file_name_ext = ".msh";
+	if (BaseLib::hasFileExtension("vtk", file_name))
+	   file_name_ext = ".vtk";
+	if (BaseLib::hasFileExtension("vtu", file_name))
+	   file_name_ext = ".vtu";
```
endJunction: What shall happen in other cases, if the mesh extension is neither msh, nor vtk, nor vtu?

ProcessLib/NumericsConfig.h
```diff
@@ -129,4 +129,7 @@ static_assert(std::is_same<detail::GlobalMatrixType::IndexType,
 /// GlobalMatrixType::IndexType and the GlobalVectorType::IndexType.
 using GlobalIndexType = detail::GlobalMatrixType::IndexType;
 
+/// A type for scalar number
+using ScalarType = detail::GlobalMatrixType::ScalarType;
```
endJunction: Why can not we assume that the scalar type used in matrices is always convertable to a `double`?

ProcessLib/GroundwaterFlowProcess.h
endJunction: One can not assume that the head is computed in meter. Could be feet or dimensionless.
Please move this to another PR.

FileIO/VtkIO/VtuInterface-impl.h
```diff
@@ -0,0 +1,79 @@
...
+{
+	if(!_mesh)
+	{
+		ERR("VtuInterface::write(): No mesh specified.");
+		return false;
+	}
+
+	// See http://www.paraview.org/Bug/view.php?id=13382
+	if(_data_mode == vtkXMLWriter::Appended)
+		WARN("Appended data mode is currently not supported, written file is not valid!");
```
endJunction: Either make it an error and abort here, or change the _data_mode to something which will produce valid results; the latter needs to be indicated in the warning message.

FileIO/VtkIO/VtuInterface-impl.h
```diff
@@ -0,0 +1,79 @@
...
+	vtkSmartPointer<UnstructuredGridWriter> vtuWriter =
+		vtkSmartPointer<UnstructuredGridWriter>::New();
+
+	vtuWriter->SetInputConnection(vtkSource->GetOutputPort());
+
+	if(_use_compressor && num_partitions == 0)
+		vtuWriter->SetCompressorTypeToZLib();
+
+	vtuWriter->SetDataMode(_data_mode);
+	if (_data_mode == vtkXMLWriter::Appended)
```
endJunction: Probably I overlook something; what about the the binary (not appended) mode? Is there a default decision somewhere?

FileIO/VtkIO/VtuInterface.cpp
```diff
@@ -61,43 +70,34 @@ MeshLib::Mesh* VtuInterface::readVTUFile(std::string const &file_name)
...
 bool VtuInterface::writeToFile(std::string const &file_name)
 {
-	if(!_mesh)
+#ifdef USE_PETSC
+	// Also for other approach with DDC.
+	// In such case, a MPI_Comm argument is need to this member,
+	// and PETSC_COMM_WORLD shoud be replaced with the argument.  
+	int mpi_rank;
+	MPI_Comm_rank(PETSC_COMM_WORLD, &mpi_rank);
+	const std::string file_name_base = boost::erase_last_copy(file_name, ".vtu");
```
endJunction: What if the `file_name` variable does not end with vtu? Or is it guarantied somewhere else that the ending is `".vtu"`?

FileIO/VtkIO/VtuInterface-impl.h
bilke: The VTK bug [is fixed](https://gitlab.kitware.com/vtk/vtk/commit/f2fd417af9fdb4d92a87f3bd5e2cb888a82b0883) but not yet in a release, maybe in VTK 6.4...

FileIO/VtkIO/VtuInterface-impl.h
bilke: In binary mode nothing more needs to be set.

ProcessLib/GroundwaterFlowProcess.h
```diff
@@ -313,52 +312,51 @@ class GroundwaterFlowProcess : public Process
...
         }
         assert(result && result->size() == _x->size());
 
-        // Copy result
-        for (std::size_t i = 0; i < _x->size(); ++i)
-            (*result)[i] = (*_x)[i];
+        // If in the case of DDC under parallel computation
+        if ( (_x->getRangeEnd() - _x->getRangeEnd()) != _x->size())
+        {
+            std::vector<ScalarType>  u(_x->size());
```
endJunction: Not sure here: is every rank allocating a `std::vector` of size of the global vector?

FileIO/VtkIO/VtuInterface-impl.h
endJunction: OK. Thanks!

FileIO/VtkIO/VtuInterface-impl.h
endJunction: OK. Just to be sure: in ASCII mode below there is a deep copy. For binary mode the copy is not needed?

FileIO/readMeshFromFile.cpp
wenqing: I just assume that the default mesh file is either vtk or vtu, and other mesh file could be OGS5 mesh. I also think about to subtract the substring after the last ``.`` in file_name.

ProcessLib/NumericsConfig.h
wenqing: we may need complex type in future.

FileIO/VtkIO/VtuInterface.cpp
wenqing: The extension is set in void solveProcesses(ProjectData &project)  in ogs.cpp.

ProcessLib/GroundwaterFlowProcess.h
wenqing: We should use SI unit. OK, this will be in another PR.

ProcessLib/GroundwaterFlowProcess.h
wenqing: Yes. It is for every rank. For huge mesh, there will be another approach by dumping raw data of rank wise vector, and extract solution by a post-process as what I did in OGS5. Therefore, there will be two methods for parallel output  of solutions:
1. multiple vtu plus pvtu in this PR.
2. parallel raw local vector dumping into a single file (future work).

wenqing: The memory usage in method 1 can be also reduced by using openmp to share the global vector within a rank. This would be  also a future effort.

FileIO/readMeshFromFile.cpp
wenqing: Changed.

ProcessLib/NumericsConfig.h
endJunction: So, if I understand correctly, introduction of the `ScalarType` type is not required for VTK output, I suggest to make a separate PR.

wenqing: It is for the newly added member of matrix class,  void getValues(PetscScalar u[]) of  void getValues(double u[]). It is used to avoid converting float type of every entries of the solution vector.

ProcessLib/GroundwaterFlowProcess.h
```diff
@@ -313,52 +312,51 @@ class GroundwaterFlowProcess : public Process
...
                     MeshLib::MeshItemType::Node);
             result->resize(_x->size());
         }
         assert(result && result->size() == _x->size());
 
-        // Copy result
-        for (std::size_t i = 0; i < _x->size(); ++i)
-            (*result)[i] = (*_x)[i];
+        // If in the case of DDC under parallel computation
+        if ( (_x->getRangeEnd() - _x->getRangeEnd()) != _x->size())
```
TomFischer: Did you mean `_x->getRangeEnd() - _x->getRangeBegin()`?

wenqing: Tom, thanks. Yes it is.

wenqing: Corrected now.

ProcessLib/GroundwaterFlowProcess.h
```diff
@@ -313,52 +312,51 @@ class GroundwaterFlowProcess : public Process
...
         }
         assert(result && result->size() == _x->size());
 
-        // Copy result
-        for (std::size_t i = 0; i < _x->size(); ++i)
-            (*result)[i] = (*_x)[i];
+        // If in the case of DDC under parallel computation
+        if ( (_x->getRangeEnd() - _x->getRangeBegin()) != _x->size())
+        {
+            std::vector<ScalarType>  u(_x->size());
```
endJunction: Please find a way around this copy. The only reason for the copy is the convenient access over the global_index. There could be a more efficient way.

ProcessLib/NumericsConfig.h
endJunction: But there is a copy anyway!

Regarding some special types: vtk (as of version 6.3) supports only 32 and 64 bit floating point number types. No complex, no long double. Please correct me if I'm wrong.

I'm not very knowledgeable about existing problems/benchmarks where complex type is used. Would you please point to such examples (in ogs5 for example)?

FileIO/VtkIO/VtuInterface-impl.h
bilke: Correct.

ProcessLib/NumericsConfig.h
wenqing: 1. For getValues PETScVector, all copies are carried out through PETSc routines.
2. This is not only for VTK,
3. The interfaces  of PETSc vector and matrix enables complex variable, which can be used for solving some problems with complex unknowns. In future, I think, we can come cross such problems.

Under the current development,  PetscScalar is defined as double without an additional preprocess definition. Therefore ScalarType is just removed.

bilke:
Jenkins test this please.

ProcessLib/GroundwaterFlowProcess.h
wenqing: Going to try IS routines for less memory and less communication.

wenqing: It is complex to get a global solution array from distributed Petsc vector data meanwhile to avoid allocate memory for the global array. The reason to have a global solution array is that the solution has to be updated onto the ghost nodes as well. The current approach dose not simply copy vector, it concatenates the local arrays to a global array by a very fast gather function.

For the mesh partitioned by nodes, a straightforward approach without global solution array is to collect ghost entry data among neighbor partitions. The drawbacks of this approach are:
1. Additional memory is also needed for  neighboring ghost nodes.
2. More computation load.

Although the current global array approach involves a little bit more memory than the approach addressed above, its computation speedup is much better.

To improve the memory usage of the current approach in the shared memory node, the application of openmp can be investigated. Also, a software package is available specially for the global array under distrusted machine (http://hpc.pnl.gov/globalarrays/documentation.shtml). However it also involves additional computer resources and memory usage. I would like to add these to the future study list. 

So far, I would like to keep the current implementation.

-------------------------------------------
PS: IS routines are only for index mapping.

scripts/cmake/ExternalProjectCatalyst.cmake
```diff
@@ -9,6 +9,9 @@ endif()
 
 # CLI modules
 set(PARAVIEW_MODULES vtkIOXML)
+#if(OGS_USE_MPI OR OGS_USE_PETSC)
+#	set(PARAVIEW_MODULES ${PARAVIEW_MODULES} vtkIOParallelXML)
+#endif()
```
endJunction: Is there a reason why this is commented?

endJunction:
The result files from the ctest run shall be removed from the external data repository, otherwise running ctest (at least for the cube1e3 example) overwrites the files.
Don't hessitate to ask if I'm unclear here...

endJunction:
I have run the cube_1e3.prj from EllipticPETSc directory with
`mpirun -np 3 bin/ogs paths_to/cube_1e3.prj`. Simulation log output seems ok, everything converged etc.
Now comparing the solution to an analytical solution (f(x,y,z) = -2x+1) the maximum error is ~0.002 which seems a little to high for such a trivial example. For the serial case the maximum error is O(1e-15).

x

wenqing:
@endJunction 
* The result files: They were copied accidentally by using *vtu filter, and they are removed now from the repository. 
* Accuracy: decrease the tolerance, e.g. by command
```
mpirun -np 3 ogs cube_1e3.prj -- -gw_ksp_type bcgs -gw_pc_type bjacobi -gw_ksp_atol 1.e-10
```

scripts/cmake/ExternalProjectCatalyst.cmake
wenqing: It is not used now but future. Leave it here just as a reminder.

wenqing: Added a short comment about this.

wenqing:
@endJunction or add one more command option ``-gw_ksp_rtol 1.e-10`` of the relative tolerance if the accuracy does not satisfy still.

endJunction:
Tried with
```sh
mpirun -np 3 ~/w/ogs/buildp/bin/ogs cube_1e3.prj -- \
   -gw_ksp_type bcgs -gw_pc_type bjacobi \
   -gw_ksp_atol 1.e-100 -gw_ksp_rtol 1.e-100
```
resulting in maxium difference of 0.0244768. (previous difference was 0.0244943).
The solver indeed spent 109 iterations. (previously around 20).
Might the problem be elsewhere?

wenqing:
The tolerance of 1.e-100 is too small, which also increases error.
Try

mpirun -np 3 ~/w/ogs/buildp/bin/ogs cube_1e3.prj -- \
    -gw_ksp_type bcgs -gw_pc_type bjacobi \
    -gw_ksp_rtol 1.e-10

scripts/cmake/ExternalProjectCatalyst.cmake
```diff
@@ -9,6 +9,11 @@ endif()
 
 # CLI modules
 set(PARAVIEW_MODULES vtkIOXML)
+### The following if-condition is not used now but future,
+### and it is commented out and left here just as a reminder. 
+#if(OGS_USE_MPI OR OGS_USE_PETSC)
+#	set(PARAVIEW_MODULES ${PARAVIEW_MODULES} vtkIOParallelXML)
+#endif()
```
endJunction: With this block commented out the compilation fails with undefined reference to vtk...:
```
../lib/libFileIO.a(VtuInterface.cpp.o): In function `vtkSmartPointer<vtkXMLPUnstructuredGridWriter>::New()':
/usr/include/vtk/vtkSmartPointer.h:117: undefined reference to `vtkXMLPUnstructuredGridWriter::New()'
clang-3.7: error: linker command failed with exit code 1 (use -v to see invocation)
Tests/CMakeFiles/testrunner.dir/build.make:2357: recipe for target 'bin/testrunner' failed
make[2]: *** [bin/testrunner] Error 1
```

endJunction:
Also the expected result files you provide do not contain the correct solution of this equation which would be a projection of the analytical solution to the particular mesh. Please have a look on the vtu files provided in Elliptic/cube.../*vtu, which do contain the analytical solutions. The particular one is called "Linear_1_to_minus1".

wenqing:
Since the mesh is partitioned, I would like use the verified computation 
results as reference. Besides, I did not find the file called 
"Linear_1_to_minus1".

endJunction:
It is not a file but a data array in the cube_...1e3.vtu file. The names of the data arrays inside the vtu files are used for the comparison of the results e.g. in vtkdiff.

endJunction:
The following test is failing on my machine. Anybody else can confirm it too?
```
[ RUN      ] InSituMesh.MappedMeshSourceRoundtrip
ERROR: In /build/vtk/src/VTK-6.3.0/IO/XML/vtkXMLReader.cxx, line 234
vtkXMLUnstructuredGridReader (0x17347a0): Error opening file /home/naumov/w/ogs/build_p/Tests//MappedMeshSourceRoundtrip.vtu

ERROR: In /build/vtk/src/VTK-6.3.0/Common/ExecutionModel/vtkExecutive.cxx, line 784
vtkCompositeDataPipeline (0x1758610): Algorithm vtkXMLUnstructuredGridReader(0x17347a0) returned failure for request: vtkInformation (0x1739790)
  Debug: Off
  Modified Time: 2175
  Reference Count: 1
  Registered Events: (none)
  Request: REQUEST_INFORMATION
  ALGORITHM_AFTER_FORWARD: 1
  FORWARD_DIRECTION: 0



/home/naumov/w/ogs/Tests/InSituLib/TestVtkMappedMeshSource.cpp:226: Failure
Value of: output->GetNumberOfPoints()
  Actual: 216
Expected: vtkMesh->GetNumberOfPoints()
Which is: 0
[  FAILED  ] InSituMesh.MappedMeshSourceRoundtrip (38 ms)
```

wenqing:
@endJunction Disabled InSituMesh.MappedMeshSourceRoundtrip in MPI environment because that VtuInterface::writeToFile under MPI writes pvtu and vtu. 

The test can be run in MPI environment by replacing writeToFile with writeVTU<vtkXMLUnstructuredGridWriter>, which just the same as that is done in the test under serial computing. Since it is a test for serial computing, it is disabled in MPI. 

Under MPI, make tests_mpi is normally performed.

wenqing:
How did you write the binary array of the analytics solution into the 
result file? I did not find it in the source code.

endJunction:
Just wrote a small for loop printing the solution, then imported the table into paraview and associated it with the mesh, which was then written. Or maybe it was directly in paraview using python.
Is there any problem reusing the existing cube_1e3.vtu with the "Linear_1_to_minus1" data array?

wenqing:
Deleted a question by accident. @endJunction has answered the question.   
---------

How did you write the binary array of the analytics solution into the result file? I did not find it in the source code.

-------

Just wrote a small for loop printing the solution, then imported the table into paraview and associated it with the mesh, which was then written. Or maybe it was directly in paraview using python.
Is there any problem reusing the existing cube_1e3.vtu with the "Linear_1_to_minus1" data array?

endJunction:
Opening `quad_20x10_GroundWaterFlow_result_pcs_0_ts_1_expected.pvtu` in paraview results in an error:
"vtkXMLPUnstructuredGridReader (0x48833c0): File for piece 0 cannot be read."

wenqing:
|File quad_20x10_GroundWaterFlow_result_pcs_0_ts_1_expected.pvtu is 
copied from ||quad_20x10_GroundWaterFlow_result_pcs_0_ts_1.pvtu, and it 
is used as the reference of its origin for file comparison. If you want 
to use it to open vtu files that has string of ``expected`` in file 
names, you have to modify it manually by adding string of ``expected`` 
to file names in the pieces source element as
```
<?xml version="1.0"?>
<VTKFile type="PUnstructuredGrid" version="0.1" 
byte_order="LittleEndian" compressor="vtkZLibDataCompressor">
   <PUnstructuredGrid GhostLevel="0">
     <PPointData>
       <PDataArray type="Float64" Name="Result"/>
     </PPointData>
     <PPoints>
       <PDataArray type="Float64" Name="Points" NumberOfComponents="3"/>
     </PPoints>
     <Piece Source="cube_1e3_pcs_0_ts_1_expected_0.vtu"/>
     <Piece Source="cube_1e3_pcs_0_ts_1_expected_1.vtu"/>
     <Piece Source="cube_1e3_pcs_0_ts_1_expected_2.vtu"/>
   </PUnstructuredGrid>
</VTKFile>
```
Meanwhile changing file name |||cube_1e3_pcs_0_ts_1_?_expected.vtu to 
|||||cube_1e3_pcs_0_ts_1_expected_?.vtu.||

The file names of pvtu and its related vtu follows a rule, which looks 
like the following example:
||foo.pvtu
||foo_0.vtu
||foo_1.vtu

where foo.pvtu contains the piece sources as:
```|
||<Piece Source="foo_0.vtu"/>
<Piece Source="foo_1.vtu"/>
|```

Therefore, I prefer to only perform file comparison for pvtu files.


|On 10/28/2015 04:11 PM, Dmitry Yu. Naumov wrote:
>
> Opening |quad_20x10_GroundWaterFlow_result_pcs_0_ts_1_expected.pvtu| 
> in paraview results in an error:
> "vtkXMLPUnstructuredGridReader (0x48833c0): File for piece 0 cannot be 
> read."
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/ufz/ogs/pull/869#issuecomment-151876190>.
>

**Comment by wenqing** [comment]
@bilke 
I compiled the ogs6 on my laptop, on which vtk-lib6-dev is installed. It seems that the system installed VTK is not used in cmake configuration on my laptop.  After cmake configuration,  the source code is compiled and the compilation was broken with an error message
```
 *** No rule to make target `vtkIOParallelXML'.  Stop.
```
I saw that a part of VTK source code is downloaded to a subdirectory of the building directory as: ``External/catalyst/src/Catalyst/VTK/IO/XML``, where exists the file of  ``vtkXMLPStructuredGridReader.cxx``. The complication error occurs at the end of these ``VTK/IO/XML`` source files being compiled. In ``CMakeLists.txt`` of that XML subdirectory, only  ``vtkIOXML`` is added to the vtk library by  command vtk_module_library(vtkIOXML ${Module_SRCS}). I guess this could be the reason to breaking the compilation.  Would you please take a look at cmake files about this issue?

**Comment by bilke** [comment]
Standard vtk installations do not provide `vtkIOParallelXML` because this is part of vtks mpi-group. See my attached PR for a working solution.

**Comment by wenqing** [comment]
@bilke  Thanks.

**Comment by wenqing** [comment]
@bilke 
Just based on master that is merged with #882 ,  and the new cmake/VTK stuffs work very well on EVE. However the travis compilation dose not start due to the cmake server problem. As said in the following error message on travis, certificate subject of cmake has problem during downloading cmake.

------------------------   
$ bash scripts/travis/cmake.sh

--2015-11-10 15:37:51--  http://www.cmake.org/files/v3.1/cmake-3.1.1-Linux-x86_64.tar.gz

Resolving www.cmake.org (www.cmake.org)... 66.194.253.19

Connecting to www.cmake.org (www.cmake.org)|66.194.253.19|:80... connected.

HTTP request sent, awaiting response... 301 Moved Permanently

Location: http://cmake.org/files/v3.1/cmake-3.1.1-Linux-x86_64.tar.gz [following]

--2015-11-10 15:37:52--  http://cmake.org/files/v3.1/cmake-3.1.1-Linux-x86_64.tar.gz

Resolving cmake.org (cmake.org)... 66.194.253.19

Connecting to cmake.org (cmake.org)|66.194.253.19|:80... connected.

HTTP request sent, awaiting response... 301 Moved Permanently

Location: https://cmake.org/files/v3.1/cmake-3.1.1-Linux-x86_64.tar.gz [following]

--2015-11-10 15:37:52--  https://cmake.org/files/v3.1/cmake-3.1.1-Linux-x86_64.tar.gz

Connecting to cmake.org (cmake.org)|66.194.253.19|:443... connected.

ERROR: no certificate subject alternative name matches

	requested host name `cmake.org'.

To connect to cmake.org insecurely, use `--no-check-certificate'.

The command "bash scripts/travis/cmake.sh" failed and exited with 5 during .

**Comment by bilke** [comment]
Merge in #885 which fixes the certificate error.

**Comment by wenqing** [comment]
@endJunction Replaced.

**Comment by endJunction** [comment]
:fast_forward:

File: FileIO/VtkIO/VtuInterface.h [hunk 13]
```diff
@@ -38,13 +38,20 @@ class VtuInterface
 	~VtuInterface();
 
 	/// Read an unstructured grid from a VTU file
-	/// \returns The converted mesh or a nullptr if reading failed
+	/// \return The converted mesh or a nullptr if reading failed
 	static MeshLib::Mesh* readVTUFile(std::string const &file_name);
 
 	/// Writes the given mesh to file.
-	/// \returns True on success, false on error
+	/// Note: a MPI_Comm type argument is needed.
```
bilke: Can you clarify this note. It is a bit misleading because there is no argument other than the filename to this method.

FileIO/readMeshFromFile.cpp
```diff
@@ -44,8 +44,12 @@ MeshLib::Mesh* readMeshFromFile(const std::string &file_name)
 {
 #ifdef USE_PETSC
 	NodePartitionedMeshReader read_pmesh(PETSC_COMM_WORLD);
+
+	std::size_t pos = file_name.find_last_of(".");
+	std::string file_name_ext = file_name.substr(pos, file_name.size()-1);
```
bilke: You can also use helper methods from BaseLib which extract the file extension and base names.

bilke:
:+1:

FileIO/readMeshFromFile.cpp
wenqing: Found two functions in FileTools.h and StringTools.h, extractBaseName and splitString, respectively.
extractBaseName removes the path, while splitString cannot handle file name with a delimiter appearing in multiple places, e.g. test.old.mesh.vtu.

FileIO/VtkIO/VtuInterface.h
wenqing: Just removed. Since there is MPI communicator used inside this function, this is just a reminder for future development if other communicator is employed. Actually there is already a comment in the definition of this function as a reminder.

FileIO/readMeshFromFile.cpp
TomFischer: @wenqing Does your code implement a different behaviour than `BaseLib::dropFileExtension()`?

wenqing: Did not check that. Just replaced the current implementation with the function. Thanks.

PR #870, opened 2015-10-21
[Cmake] Make git secondary working directory work

chleh:
Starting from git 2.5 one can create secondary working directories using `git worktree add ...`.
This PR extends our CMake configuration s.t. it works on secondary working directories.


TomFischer:
What is a secondary working directory and how is it used?

chleh:
Sorry, gave the wrong command above.
A secondary working directory allows to check out two different branches to two different directories. Those two worktrees will then operate on the same repository.

See also: https://git-scm.com/docs/git-worktree

I thought this feature might be useful to have in OGS6 development.

TomFischer:
Thank you for the explanation.

wenqing:
:+1:

PR #821, opened 2015-08-28
Global Index Calculation for Multiple nodal d.o.f.

chleh:
VectorMatrixAssembler now has a pointer to the variables the global problem is solved for.
The indices of a variable in the global matrix are now computed based on the setting BY_COMPONENT/BY_LOCATION.
However, I changed VectorMatrixAssembler such that in the local matrices are always ordered by component.

PR #871, opened 2015-10-21
Multiple nodal d.o.f.

chleh:
This PR is a reincarnation of #821.

It allows for using multiple degrees of freedom per node.

Changes are:
* VectorMatrixAssembler computes local nodal values for each degree of freedom and passes them on to the local assembler.
* Thus, local assemblers have to accept parameters: `assemble(x, xold)`
* MeshComponentMap stores number of components and global matrix order (by location vs. by component)
* some methods now have component id and dof_table as additional input parameters.
* dof table now saves global indices in a matrix indexed by mesh element and component id

TODOs (will not be handled in this PR):
* VectorMatrixAssembler passes all local nodal values in a single vector. In the future that might be changed to passing each component separately.
* using `LocalToGlobalIndexMap::deriveBoundaryConstrainedMap()` on a derived dof table is allowed only if certain conditions hold (see doc in source file). Yet, that might be a rather rare case to use.

### DIFF
--- a/AssemblerLib/LocalToGlobalIndexMap.h
+++ b/AssemblerLib/LocalToGlobalIndexMap.h
@@ -16,6 +16,8 @@
 
 #include <vector>
 
+#include <Eigen/Dense>
+
 #include "AssemblerLib/MeshComponentMap.h"
 #include "MathLib/LinAlg/RowColumnIndices.h"
 #include "MeshLib/MeshSubsets.h"
@@ -43,19 +45,25 @@ public:
     /// each mesh element of the given mesh_subsets.
     explicit LocalToGlobalIndexMap(
         std::vector<MeshLib::MeshSubsets*> const& mesh_subsets,
-        AssemblerLib::ComponentOrder const order =
-            AssemblerLib::ComponentOrder::BY_COMPONENT);
+        AssemblerLib::ComponentOrder const order);
 
     /// Derive a LocalToGlobalIndexMap constrained to a set of mesh subsets and
     /// elements. A new mesh component map will be constructed using the passed
     /// mesh_subsets.
     ///
-    /// \note The elements are not necessary those used in the mesh_subsets.
+    /// \param mesh_subsets the subsets to which this map will be constrained.
+    ///        It must hold that: mesh_subsets.size() == _mesh_subsets.size()
+    ///        If a component shall not be present in the derived map, put a nullptr
+    ///        to the respective position in mesh_subsets.
+    ///
+    /// \note The elements are not necessarily those used in the mesh_subsets.
+    ///
+    /// \note It is possible to use this method on the returned map only
+    ///       if \c mesh_subsets contains no nullptr!
+    ///
     LocalToGlobalIndexMap* deriveBoundaryConstrainedMap(
         std::vector<MeshLib::MeshSubsets*> const& mesh_subsets,
-        std::vector<MeshLib::Element*> const& elements,
-        AssemblerLib::ComponentOrder const order =
-            AssemblerLib::ComponentOrder::BY_COMPONENT) const;
+        std::vector<MeshLib::Element*> const& elements) const;
 
     /// Returns total number of degrees of freedom.
     std::size_t dofSize() const;
@@ -68,10 +76,11 @@ public:
 
     std::size_t size() const;
 
-    RowColumnIndices operator[](std::size_t const mesh_item_id) const;
+    std::size_t getNumComponents() const { return _mesh_subsets.size(); }
 
-    LineIndex rowIndices(std::size_t const mesh_item_id) const;
-    LineIndex columnIndices(std::size_t const mesh_item_id) const;
+    RowColumnIndices operator()(std::size_t const mesh_item_id, const unsigned component_id) const;
+
+    std::size_t getNumElementDOF(std::size_t const mesh_item_id) const;
 
     GlobalIndexType getGlobalIndex(MeshLib::Location const& l,
                                std::size_t const c) const
@@ -86,27 +95,30 @@ private:
     /// \attention The passed mesh_component_map is in undefined state after
     /// this construtor.
     explicit LocalToGlobalIndexMap(
-        std::vector<MeshLib::MeshSubsets*> const& mesh_subsets,
+        std::vector<MeshLib::MeshSubsets*>&& mesh_subsets,
+        std::vector<std::size_t> const& original_indices,
         std::vector<MeshLib::Element*> const& elements,
-        AssemblerLib::MeshComponentMap&& mesh_component_map,
-        AssemblerLib::ComponentOrder const order);
+        AssemblerLib::MeshComponentMap&& mesh_component_map);
 
     template <typename ElementIterator>
     void
     findGlobalIndices(ElementIterator first, ElementIterator last,
-        std::size_t const mesh_id, AssemblerLib::ComponentOrder const order);
+        std::size_t const mesh_id,
+        const unsigned component_id, const unsigned comp_id_write);
 
 private:
-    std::vector<MeshLib::MeshSubsets*> const& _mesh_subsets;
+    std::vector<MeshLib::MeshSubsets*> const _mesh_subsets;
     AssemblerLib::MeshComponentMap _mesh_component_map;
 
-    /// Vector contains for each element a vector of global row/or entry indices
-    /// in the global stiffness matrix or vector
-    std::vector<LineIndex> _rows;
+    using Table = Eigen::Matrix<LineIndex, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>;
+
+    /// Table contains for each element (first index) and each component (second index)
+    /// a vector (\c LineIndex) of indices in the global stiffness matrix or vector
+    Table _rows;
 
     /// Vector alias to that contains for each element a vector of global column
     /// indices in the global stiffness matrix
-    std::vector<LineIndex> const& _columns = _rows;
+    Table const& _columns = _rows;
 
 #ifndef NDEBUG
     /// Prints first rows of the table, every line, and the mesh component map.
@@ -114,43 +126,6 @@ private:
 #endif  // NDEBUG
 
 };
-
-
-template <typename ElementIterator>
-void
-LocalToGlobalIndexMap::findGlobalIndices(ElementIterator first, ElementIterator last,
-    std::size_t const mesh_id, AssemblerLib::ComponentOrder const order)
-{
-    // For each element find the global indices for node/element
-    // components.
-    for (ElementIterator e = first; e != last; ++e)
-    {
-        std::vector<MeshLib::Location> vec_items;
-        std::size_t const nnodes = (*e)->getNNodes();
-        vec_items.reserve(nnodes);
-
-        for (unsigned n = 0; n < nnodes; n++)
-        {
-            vec_items.emplace_back(
-                mesh_id,
-                MeshLib::MeshItemType::Node,
-                (*e)->getNode(n)->getID());
-        }
-
-        // Save a line of indices for the current element.
-        switch (order)
-        {
-            case AssemblerLib::ComponentOrder::BY_LOCATION:
-                _rows.push_back(
-                    _mesh_component_map.getGlobalIndicesByLocation(vec_items));
-                break;
-            case AssemblerLib::ComponentOrder::BY_COMPONENT:
-                _rows.push_back(
-                    _mesh_component_map.getGlobalIndicesByComponent(vec_items));
-                break;
-        }
-    }
-}
 
 }   // namespace AssemblerLib
 
--- a/AssemblerLib/LocalToGlobalIndexMap.h
+++ b/AssemblerLib/LocalToGlobalIndexMap.h
@@ -116,8 +116,7 @@ private:
     /// a vector (\c LineIndex) of indices in the global stiffness matrix or vector
     Table _rows;
 
-    /// Vector alias to that contains for each element a vector of global column
-    /// indices in the global stiffness matrix
+    /// \see _rows
     Table const& _columns = _rows;
 
 #ifndef NDEBUG
--- a/AssemblerLib/MeshComponentMap.cpp
+++ b/AssemblerLib/MeshComponentMap.cpp
@@ -31,6 +31,8 @@ GlobalIndexType const MeshComponentMap::nop =
 #ifdef USE_PETSC
 MeshComponentMap::MeshComponentMap(
     const std::vector<MeshLib::MeshSubsets*> &components, ComponentOrder order)
+    : _num_components(components.size())
+
 {
     // construct dict (and here we number global_index by component type)
     std::size_t global_index_offset = 0;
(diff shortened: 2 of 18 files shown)


AssemblerLib/MeshComponentMap.cpp
```diff
@@ -259,4 +263,34 @@ std::vector<GlobalIndexType> MeshComponentMap::getGlobalIndicesByComponent(
     return global_indices;
 }
 
+std::vector<GlobalIndexType>
+MeshComponentMap::getIndicesForComponent(
+        const std::vector<GlobalIndexType>& cnt,
+        const unsigned component_id) const
+{
+    std::vector<GlobalIndexType> new_cnt;
+    auto const N = cnt.size() / _num_components;
```
wenqing: In some case, DOF per node can vary. For example, using linear finite element for one PDE (e.g. that of fluid flow) and high order finite element for another PDE (e.g. that of solid deformation) in one geometry element is needed for solving of some coupled equations.

ProcessLib/ProcessVariable.h
```diff
@@ -60,8 +60,9 @@ class ProcessVariable
 	MeshLib::Mesh const& getMesh() const;
 
 	void initializeDirichletBCs(MeshGeoToolsLib::MeshNodeSearcher& searcher,
-	                            std::vector<std::size_t>& global_ids,
-	                            std::vector<double>& values);
+	        const AssemblerLib::LocalToGlobalIndexMap& local2global,
+	        const unsigned nodal_dof_idx,
+	        std::vector<std::size_t>& global_ids, std::vector<double>& values);
```
wenqing: When you merge or rebase, you need to change ``td::size_t`` here to ``GlobalIndexType``.

AssemblerLib/VectorMatrixAssembler.h
```diff
@@ -49,15 +56,46 @@ class VectorMatrixAssembler
...
+        auto const& x = *_x;
+        auto const& x_pts = *_x_prev_ts;
+
+        const unsigned num_comp = _data_pos.getNumComponents();
+
+        if (_x)         localX.reserve(element_dof);
+        if (_x_prev_ts) localX_pts.reserve(element_dof);
+        remapped_rows_cols.reserve(element_dof);
+
+        // The local matrix will always be ordered by component,
```
chleh: Yes, vector might be more appropriate, here. But also the local matrix has to be ordered by component due to the `addToGlobal()` call below.

AssemblerLib/MeshComponentMap.cpp
chleh: That's a valid point. So far I've only considered scalar components which all operate on basically the same mesh, i.e. all have same order finite elements.
I will think about that.

chleh:
Test has been updated: Less indentation, more independent cases.

Tests/AssemblerLib/LocalToGlobalIndexMapMultiComponent.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+	std::size_t operator()(std::size_t const node,
+	                       std::size_t const component) const
+	{
+		return node * num_components + component;
+	}
+};
+
+
+template <enum AL::ComponentOrder ComponentOrder>
```
endJunction: My fault here:
```
F:\jenkins\workspace\OGS-6\Win-PRs\ogs\Tests\AssemblerLib\LocalToGlobalIndexMapMultiComponent.cpp(166): error C3431: 'ComponentOrder' : a scoped enumeration cannot be redeclared as an unscoped enumeration [F:\jenkins\workspace\OGS-6\Win-PRs\build\Tests\testrunner.vcxproj]
```

@chleh could you remove the `enum` in templat argument list please.

AssemblerLib/LocalToGlobalIndexMap.h
```diff
@@ -86,27 +95,31 @@ class LocalToGlobalIndexMap
...
     /// Vector contains for each element a vector of global row/or entry indices
     /// in the global stiffness matrix or vector
-    std::vector<LineIndex> _rows;
+    /// first index: element, second index: component
+    Table _rows;
 
     /// Vector alias to that contains for each element a vector of global column
     /// indices in the global stiffness matrix
-    std::vector<LineIndex> const& _columns = _rows;
+    Table const& _columns = _rows;
```
endJunction: The comments (the previously bad spelling aside) needs an update.

endJunction:
When building with PETSc following errors emerge:
```
/home/naumov/w/ogs/AssemblerLib/VectorMatrixAssembler.h:79:49: error: type 'const MathLib::PETScVector' does not provide a subscript operator
            if (_x)         localX.emplace_back((*_x)[i]);
                                                ^~~~~ ~
/home/naumov/w/ogs/AssemblerLib/SerialExecutor.h:40:13: note: in instantiation of function template specialization 'AssemblerLib::VectorMatrixAssembler<MathLib::PETScMatrix,
      MathLib::PETScVector>::operator()<ProcessLib::GroundwaterFlow::LocalAssemblerDataInterface<MathLib::PETScMatrix, MathLib::PETScVector> >' requested here
            f(i, c[i], std::forward<Args_>(args)...);
            ^
/home/naumov/w/ogs/AssemblerLib/GlobalSetup.h:46:26: note: in instantiation of function template specialization
      'AssemblerLib::SerialExecutor::execute<AssemblerLib::VectorMatrixAssembler<MathLib::PETScMatrix, MathLib::PETScVector>,
      std::__debug::vector<ProcessLib::GroundwaterFlow::LocalAssemblerDataInterface<MathLib::PETScMatrix, MathLib::PETScVector> *,
      std::allocator<ProcessLib::GroundwaterFlow::LocalAssemblerDataInterface<MathLib::PETScMatrix, MathLib::PETScVector> *> >>' requested here
        return Executor::execute(std::forward<Args>(args)...);
                         ^
/home/naumov/w/ogs/ProcessLib/GroundwaterFlowProcess.h:278:23: note: in instantiation of function template specialization
      'AssemblerLib::GlobalSetup<AssemblerLib::VectorMatrixBuilder<MathLib::PETScMatrix, MathLib::PETScVector>, AssemblerLib::SerialExecutor,
      MathLib::PETScLinearSolver>::execute<AssemblerLib::VectorMatrixAssembler<MathLib::PETScMatrix, MathLib::PETScVector> &,
      std::__debug::vector<ProcessLib::GroundwaterFlow::LocalAssemblerDataInterface<MathLib::PETScMatrix, MathLib::PETScVector> *,
      std::allocator<ProcessLib::GroundwaterFlow::LocalAssemblerDataInterface<MathLib::PETScMatrix, MathLib::PETScVector> *> > &>' requested here
        _global_setup.execute(*_global_assembler, _local_assemblers);
                      ^
In file included from /home/naumov/w/ogs/ProcessLib/GroundwaterFlowProcess.cpp:10:
In file included from /home/naumov/w/ogs/ProcessLib/GroundwaterFlowProcess-fwd.h:13:
In file included from /home/naumov/w/ogs/ProcessLib/GroundwaterFlowProcess.h:28:
/home/naumov/w/ogs/AssemblerLib/VectorMatrixAssembler.h:80:53: error: type 'const MathLib::PETScVector' does not provide a subscript operator
            if (_x_prev_ts) localX_pts.emplace_back((*_x_prev_ts)[i]);
                                                    ^~~~~~~~~~~~~ ~
2 errors generated.
```

AssemblerLib/MeshComponentMap.h
```diff
@@ -46,7 +46,18 @@ class MeshComponentMap
         ComponentOrder order);
 
     /// Creates a subset of the current mesh component map.
-    /// The order of components is the same as of the current map.
+    /// The components are those from \c components which are not nullptrs.
+    /// The order (BY_LOCATION/BY_NODE) of components is the same as of the current map.
```
endJunction: `BY_NODE` is confusing because such order does not exist.
Would simply keep `BY_LOCATION`.

ProcessLib/UniformDirichletBoundaryCondition.h
```diff
@@ -48,13 +49,27 @@ class UniformDirichletBoundaryCondition
...
+            std::vector<double>& values)
     {
         // Find nodes' ids on the given mesh on which this boundary condition
         // is defined.
         std::vector<std::size_t> ids = searcher.getMeshNodeIDs(*_geometry);
 
+        // convert mesh node ids to global index for the given component
+        for (auto& id : ids)
+        {
+            MeshLib::Location l(searcher.getMesh().getID(),
```
endJunction: I would replace the `searcher.getMesh().getID()` by `searcher.getMeshId()` just forwarding that id and not exposing the mesh of the searcher. If I didn't miss something, this is the only place where node searcher's `getMesh()` is used.

AssemblerLib/VectorMatrixAssembler.h
```diff
@@ -49,15 +57,41 @@ class VectorMatrixAssembler
...
+
+        std::vector<double> localX;
+        std::vector<double> localX_pts;
+
+        if (_x)         localX.reserve(indices.size());
+        if (_x_prev_ts) localX_pts.reserve(indices.size());
+
+        for (auto i : indices)
+        {
+            if (_x)         localX.emplace_back(_x->get(i));
```
endJunction: Note for later:
The get() call for negative indices in case of PETSc vectors will not return the expected value in all cases.
See [PETSc::VecGetValues](http://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Vec/VecGetValues.html) documentation.

AssemblerLib/VectorMatrixAssembler.h
```diff
@@ -38,6 +38,14 @@ class VectorMatrixAssembler
 
     ~VectorMatrixAssembler() {}
 
+    void setX(GLOBAL_VECTOR_ const * x, GLOBAL_VECTOR_ const * x_prev_ts)
+    {
+        assert((!x == !x_prev_ts) && "either no or both inputs have to be set");
+        assert((!x) || x->size() == x_prev_ts->size());
+        _x = x;
+        _x_prev_ts = x_prev_ts;
+    }
```
endJunction: Passing the solution `x` to assembler is required for the non-linear assembly; OK.
How is the `x_prev_ts` used?

chleh: I use `x_prev_ts` for the assembly of the global equation system. E.g. for a parabolic problem the linear equation system to solve is `M x' + K x = b`, where `x' = (x - y) / dt`
So (currently) in need both `x` and `y := x_prev_ts`.
Maybe later on this can be moved to a different place.

endJunction:
:fast_forward:

AssemblerLib/VectorMatrixAssembler.h
wenqing: Get values entry by entry is slow in PETSc. There is a member in class PETScVector called  
```
get(const std::vector<PetscInt> &e_idxs,  T_SUBVEC &sub_vec)
```
You can try to perform it like
```
_x->get(indices, localX);
```
For this purpose, you need to add        
```
 template<class T_SUBVEC> void get(const std::vector<[Int Type of the owner package]> &e_idxs,
                                          T_SUBVEC &sub_vec)
```
to all vector classes in MathLib/LinAlg. 

Besides, do not forget to set  indices be positive.

AssemblerLib/VectorMatrixAssembler.h
```diff
@@ -49,15 +57,41 @@ class VectorMatrixAssembler
...
 
-        LocalToGlobalIndexMap::RowColumnIndices const& indices = _data_pos[id];
+        std::vector<GlobalIndexType> indices;
+
+        // Local matrices and vectors will always be ordered by component,
+        // no matter what the order of the global matrix is.
+        for (unsigned c=0; c<_data_pos.getNumComponents(); ++c)
+        {
+            auto const& idcs = _data_pos(id, c).rows;
+            indices.reserve(indices.size() + idcs.size());
```
wenqing: set fabs indices

AssemblerLib/LocalToGlobalIndexMap.h
```diff
@@ -86,27 +95,30 @@ class LocalToGlobalIndexMap
...
 
 private:
-    std::vector<MeshLib::MeshSubsets*> const& _mesh_subsets;
+    std::vector<MeshLib::MeshSubsets*> const _mesh_subsets;
     AssemblerLib::MeshComponentMap _mesh_component_map;
 
-    /// Vector contains for each element a vector of global row/or entry indices
-    /// in the global stiffness matrix or vector
-    std::vector<LineIndex> _rows;
+    using Table = Eigen::Matrix<LineIndex, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>;
```
wenqing: It is possible to place this alias some where outside of this file, e.g. a a file like ``MathLib/LinAlg/MatrixTools.h`` for comment access? Under such consideration, we can replace the eigen with other third party algebraic package easily.

chleh: I think, currently there are no benefits from putting this definition to some other place: It is an implementation detail of the dof table and only serves as a special lookup table having two indices.
But maybe I should define it in terms of either `Eigen::Matrix` or `DenseMatrix` depending on the setting `OGS_USE_EIGEN`.

AssemblerLib/VectorMatrixAssembler.h
chleh: > Get values entry by entry is slow in PETSc.

Is there really a big performance impact from getting values index-by-index? If I use `get(const std::vector<PetscInt> &e_idxs,  T_SUBVEC &sub_vec)`, the PETSc function probably is faster (though, not with zero runtime), but results have to be cached.

Additionally, extending all vector classes would clutter this PR. Sure, this is a rather weak argument. But, as @endJunction and you pointed out, there is an issue with negative indices and PETSc. This is definitely out of scope of this PR. I think ghost nodes should be handled transparently in the class `PetscVector`.

I suggest to leave this issue open for this PR and to work on a solution for it afterwards.

Do you agree?

AssemblerLib/VectorMatrixAssembler.h
chleh: Excuse me, I don't understand what you mean. Could you explain again, please?

wenqing: Just related the negative ID for PETSc, which you mentioned to put it in the next PR in your reply to the next comment.

AssemblerLib/VectorMatrixAssembler.h
wenqing: In PETScVector, get(PestcInt idx) calls PETSc routine
```
VecGetValues(_v, 1, &idx, &x);
```
e.g. it get one entry through an array with the size of one.
To fetch values from petsc vector, communication operation is involved among node. To get a bunch of values through an array in one time costs much less time than to get it one by one.

chleh: OK, but I think extending (and unifying) matrix classes' interfaces should also be done in a separate PR.

wenqing: That's good.

wenqing:
:+1:

Tests/AssemblerLib/LocalToGlobalIndexMapMultiComponent.cpp
```diff
@@ -0,0 +1,277 @@
...
+	}
+
+	template <AL::ComponentOrder order>
+	void test(const unsigned num_components, const unsigned selected_component,
+	          std::function<std::size_t(std::size_t, std::size_t)> const
+	              compute_global_index);
+
+	std::unique_ptr<const MeshLib::Mesh> mesh;
+	std::unique_ptr<const MeL::MeshSubset> mesh_items_all_nodes;
+    std::vector<MeL::MeshSubsets*> components_boundary;
```
TomFischer: Can you check the indentation?

TomFischer:
:+1:

PR #875, opened 2015-10-22
Implementation of plane plane intersection

TomFischer:
The PR conatins
- An implementation of a class `Plane` and
- The algorithm to compute the intersection of two planes as well as
- Tests for the intersection algorithm.

### DIFF
--- a/GeoLib/AnalyticalGeometry.cpp
+++ b/GeoLib/AnalyticalGeometry.cpp
@@ -48,6 +48,11 @@ Orientation getOrientation(const GeoLib::Point* p0, const GeoLib::Point* p1,
 	return getOrientation((*p0)[0], (*p0)[1], (*p1)[0], (*p1)[1], (*p2)[0], (*p2)[1]);
 }
 
+bool isParallel(MathLib::Vector3 const& v, MathLib::Vector3 const& w, double eps)
+{
+	return MathLib::crossProduct(v,w).getSqrLength() <= eps*eps;
+}
+
 bool parallel(MathLib::Vector3 v, MathLib::Vector3 w)
 {
 	const double eps(std::numeric_limits<double>::epsilon());
--- a/GeoLib/Plane.h
+++ b/GeoLib/Plane.h
@@ -93,7 +93,7 @@ computePlanePlaneIntersection(GeoLib::Plane const& p0, GeoLib::Plane const& p1,
 	// form (n1,p) - d1 = 0 and (n2,p) - d2 = 0
 
 	// create right hand side of the system of linear equations
-	std::array<double, 2> rhs{{p0.getDistance(), p1.getDistance()}};
+	std::array<double, 2> rhs{{-p0.getDistance(), -p1.getDistance()}};
 
 	// copy vector components into the matrix of the system of linear equations
 	MathLib::DenseMatrix<double, std::size_t> mat(2,3);
@@ -103,6 +103,10 @@ computePlanePlaneIntersection(GeoLib::Plane const& p0, GeoLib::Plane const& p1,
 	mat(1,0) = n2[0];
 	mat(1,1) = n2[1];
 	mat(1,2) = n2[2];
+
+	std::cerr << "########" << std::endl;
+	std::cerr << "n1: " << p0.getNormal() << ", d1: " << p0.getDistance() << std::endl;
+	std::cerr << "n2: " << p1.getNormal() << ", d2: " << p1.getDistance() << std::endl;
 
 	// full pivot search
 	std::size_t i(0), j(0);
@@ -132,6 +136,8 @@ computePlanePlaneIntersection(GeoLib::Plane const& p0, GeoLib::Plane const& p1,
 		std::swap(n2[0], n2[j]);
 	}
 
+	std::cerr << "pivoted: " << mat << std::endl;
+
 	// eliminate the entry (1,0) and apply changes to the other entries of the
 	// column and the right hand side
 	long double const l(mat(1,0) / mat(0,0));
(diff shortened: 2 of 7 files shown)


GeoLib/AnalyticalGeometry.cpp
```diff
@@ -48,6 +48,11 @@ Orientation getOrientation(const GeoLib::Point* p0, const GeoLib::Point* p1,
 	return getOrientation((*p0)[0], (*p0)[1], (*p1)[0], (*p1)[1], (*p2)[0], (*p2)[1]);
 }
 
+bool isParallel(MathLib::Vector3 const& v, MathLib::Vector3 const& w, double eps)
+{
+	return MathLib::crossProduct(v,w).getSqrLength() <= eps;
```
chleh: Shouldn't the square of the length rather be compared to eps^2?

GeoLib/Plane.h
```diff
@@ -0,0 +1,118 @@
...
+#include "MathLib/Vector3.h"
+
+namespace GeoLib
+{
+
+class Plane final
+{
+public:
+	Plane(MathLib::Vector3 const& unit_normal, double d)
+		: _unit_normal(unit_normal), _oriented_distance(d)
```
chleh: Maybe it's better to check if unit_normal really has unit length, and if not, to scale.
`getDistance()` relies on that feature.

GeoLib/Plane.h
```diff
@@ -0,0 +1,118 @@
...
+	double d2(p1.getDistance());
+	// determine a point p that fulfills both plane equations in Hessian normal
+	// form (n1,p) - d1 = 0 and (n2,p) - d2 = 0
+	MathLib::Point3d p;
+	if (std::abs(direction[2]) >= eps) {
+		// direction vector is not in parallel with xy plane and it is save to
+		// set p[2] zero
+		p[2] = 0.0;
+		// solve the 2x3 system of linear equations manually
+		double const l10(n2[0]/n1[0]);
```
chleh: Check if `n1[0] == 0.0`.
Or is it not necessary?

GeoLib/Plane.h
```diff
@@ -0,0 +1,118 @@
...
+private:
+	MathLib::Vector3 _unit_normal;
+	double _oriented_distance;
+};
+
+/// Function computes the intersection of two planes
+/// @param p0 first plane
+/// @param p1 second plane
+/// @return a line described by the pair of a vector (direction) and a point of
+/// the line
```
chleh: Add prerequisite: It must be checked beforehand that the two planes actually intersect.
Or return pair(0, 0) in the case of parallel planes?

GeoLib/Plane.h
```diff
@@ -0,0 +1,118 @@
...
+		// direction vector is not in parallel with xy plane and it is save to
+		// set p[2] zero
+		p[2] = 0.0;
+		// solve the 2x3 system of linear equations manually
+		double const l10(n2[0]/n1[0]);
+		double const s(n2[1] - n1[1] * l10);
+		p[1] = 1.0/s * (-d2 + d1*l10);
+		p[0] = (-d1 - n1[1] * p[1])/n1[0];
+	} else if (std::abs(direction[1]) >= eps){
+		// Direction vector is in parallel with xy plane and the y component is
```
chleh: xz plane?

GeoLib/Plane.h
```diff
@@ -0,0 +1,118 @@
...
+		// not zero. Consequently, it is save to set p[1] to zero.
+		p[1] = 0.0;
+		// solve the 2x3 system of linear equations manually
+		double const l10(n2[0]/n1[0]);
+		double const s(n2[2] - n1[2] * l10);
+		p[2] = 1.0/s * (-d2 + d1*l10);
+		p[0] = (-d1 - n1[2]*p[2])/n1[0];
+	} else {
+		// direction vector is parallel to e_x = (1,0,0)
+		// => either (n1 = e_y and n2 = e_z) or (n1 = e_z and n2 = e_y)
```
chleh: I think there are many other cases, where the cross product of two vectors is `e_x`.
Basically for any two n1 != n2, which both are in the y-z-plane, their cross product will be parallel to e_x.

Tests/GeoLib/TestPlanePlaneIntersection.cpp
```diff
@@ -0,0 +1,235 @@
...
+	ac::randomTupleGenerator<double, 3> tuple_generator;
+	ac::cons_generator<MathLib::Point3d, ac::randomTupleGenerator<double, 3>>
+	    points_gen{tuple_generator};
+
+	ac::gtest_reporter gtest_reporter;
+};
+
+TEST_F(GeoLibComputePlanePlaneIntersection, TestPlanePlaneIntersection)
+{
+	auto checkPlanePlaneIntersection =
```
chleh: Could you add a comment like: "Takes a list of four points (p0, ... p3) , constructs two planes (p0, p1, p2), (p1, p2, p3), computes their intersection".
Sure, one can reaad that information also directly from the code, but a comment would save one minute of thinking.

GeoLib/Plane.h
```diff
@@ -0,0 +1,118 @@
...
+};
+
+/// Function computes the intersection of two planes
+/// @param p0 first plane
+/// @param p1 second plane
+/// @return a line described by the pair of a vector (direction) and a point of
+/// the line
+std::pair<MathLib::Vector3, MathLib::Point3d>
+computePlanePlaneIntersection(GeoLib::Plane const& p0, GeoLib::Plane const& p1,
+	double eps = std::numeric_limits<double>::epsilon())
```
endJunction: eps is missing in the doxygen parameter description.

TomFischer:
Will try another approach for the original problem.

PR #877, opened 2015-10-24
Move boost::ptree type instantiations in single TU

endJunction:
TU = translation unit.

This moves an expensive read_xml() instantiation in the first step from ogs.cpp into an own TU BaseLib/ConfigTree.cpp and reduces the compilation time for ogs.cpp from 6.80s (100%) to 5.84s (85%).
The compile time for the whole project increases slightly from 210.7s to 212.6s.

In the second step an explicit template instantiation of the boost::property_tree::ptree is added to ConfigTree.cpp avoiding multiple instantiations. This reduces the compile time for ogs.cpp to 5.46s (80%) and the whole project compilation time to 204.2s. (~3% faster)

The advantage of this all is that when actively working on processes the recompilation of the simulator is faster.

### DIFF
--- a/BaseLib/ConfigTree.h
+++ b/BaseLib/ConfigTree.h
@@ -1,0 +1,23 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2015, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef BASELIB_CONFIGTREE_H_
+#define BASELIB_CONFIGTREE_H_
+
+#include <boost/property_tree/ptree_fwd.hpp>
+#include <boost/filesystem.hpp>
+
+namespace BaseLib
+{
+
+boost::property_tree::ptree read_xml_config(
+    boost::filesystem::path const& path);
+}
+
+#endif  // BASELIB_CONFIGTREE_H_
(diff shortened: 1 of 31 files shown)


bilke:
:+1:

BaseLib/ConfigTree.h
```diff
@@ -0,0 +1,31 @@
...
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef BASELIB_CONFIGTREE_H_
+#define BASELIB_CONFIGTREE_H_
+
+#include <string>
+#include <boost/property_tree/ptree.hpp>
```
chleh: There is also the header file `boost/property_tree/ptree_fwd.hpp`, which only has some forward declarations (?).
Maybe it's better to use that one here.

chleh: :+1:
Besides the question: Does it make sense to split off a forward declaration part from `BaseLib/ConfigTree.h`?

endJunction: 1. Unfortunately it is not possible to include ptree_fwd in `ConfigTree.h` because the explicit template declaration requires the full template...
 2. In my opinion no, because `ConfigTree.h` already constains only declarations.

BaseLib/ConfigTree.h
```diff
@@ -0,0 +1,31 @@
...
+
+namespace BaseLib
+{
+using ConfigTree = boost::property_tree::ptree;
+
+boost::property_tree::ptree read_xml_config(
+    boost::filesystem::path const& path);
+
+/// Returns the JSON-representation of the given boost::property_tree.
+std::string propertyTreeToString(boost::property_tree::ptree const& tree);
```
TomFischer: Is the JSON format used anywhere? Maybe the function name can give a hint that JSON format is written?

PR #879, opened 2015-10-28
Switch behaviour for CMAKE_BUILD_TYPE in CMake gui

bilke:
You can toggle through possible values for CMAKE_BUILD_TYPE with ENTER
in ccmake and by popup-menu in cmake-gui.


endJunction:
Nice!
:fast_forward:

wenqing:
:+1:

PR #880, opened 2015-10-28
CMake script which marks variables as advanced

bilke:
The marked variables will not show up by default in ccmake or cmake-gui
Enable them by pressing `t` or the *Advanced* checkbox.


chleh:
looks good :+1:

PR #884, opened 2015-11-10
Use logogCustomCout

wenqing:
Use logogCustomCout to allow the logog output from rank 0 only under MPI environment.

### DIFF
--- a/scripts/jenkins/gcc.groovy
+++ b/scripts/jenkins/gcc.groovy
@@ -14,7 +14,7 @@ node('docker')
 
 	// Multiple configurations are build in parallel
 	parallel linux: {
-		docker.image('ogs6/gcc-ogs-cli').inside
+		docker.image('ogs6/gcc-ogs-cli:latest').inside
 		{
 			build 'build', '', 'package'
 
@@ -26,13 +26,13 @@ node('docker')
 	},
 
 	linux_gui: {
-		docker.image('ogs6/gcc-ogs-gui').inside {
+		docker.image('ogs6/gcc-ogs-gui:latest').inside {
 			build 'build_gui', '-DOGS_BUILD_GUI=ON -DOGS_BUILD_TESTS=OFF -DOGS_BUILD_CLI=OFF', 'package'
 		}
 	}
 
 //	windows: {
-//		docker.image('ogs6/mingw-ogs-gui').inside
+//		docker.image('ogs6/mingw-ogs-gui:latest').inside
 //		{
 //			build 'build_win', '-DCMAKE_TOOLCHAIN_FILE=$CMAKE_TOOLCHAIN_FILE', 'package'
 //
@@ -44,7 +44,7 @@ node('docker')
 //	},
 //
 //	windows_gui: {
-//		docker.image('ogs6/mingw-ogs-gui').inside
+//		docker.image('ogs6/mingw-ogs-gui:latest').inside
 //		{
 //			build 'build_win_gui',
 //						'-DCMAKE_TOOLCHAIN_FILE=$CMAKE_TOOLCHAIN_FILE -DOGS_BUILD_GUI=ON -DOGS_BUILD_TESTS=OFF -DOGS_BUILD_CLI=OFF',


Applications/ApplicationsLib/LogogSetup.h
```diff
@@ -25,8 +26,9 @@ class LogogSetup final
 	LogogSetup()
 	{
 		LOGOG_INITIALIZE();
-		fmt = new BaseLib::LogogSimpleFormatter;
-		logog_cout = new logog::Cout;
+		fmt = new BaseLib::TemplateLogogFormatterSuppressedGCC<TOPIC_LEVEL_FLAG
+		                      | TOPIC_FILE_NAME_FLAG | TOPIC_LINE_NUMBER_FLAG>;
```
chleh: This way filename and line number will be printed, won't they? Do we want that?

Applications/CLI/ogs.cpp
```diff
@@ -92,9 +92,9 @@ int main(int argc, char *argv[])
 	cmd.add(project_arg);
 	cmd.parse(argc, argv);
 
-	ApplicationsLib::LogogSetup logog_setup;
 	ApplicationsLib::LinearSolverLibrarySetup linear_solver_library_setup(
 	    argc, argv);
+	ApplicationsLib::LogogSetup logog_setup;
```
chleh: Unnecessary edit.

BaseLib/LogogCustomCout.h
```diff
@@ -25,21 +27,25 @@ namespace BaseLib
...
-#ifdef USE_MPI
+#if defined(USE_MPI) || defined(USE_PETSC)
 	/**
 	 * Constructor when MPI is involved
 	 *
 	 * @param all_rank_output_level  Minimum level to output messages from all MPI processes
 	 * @param mpi_comm               MPI communicator
 	 */
+#if defined(USE_PETSC)
+	LogogCustomCout(LOGOG_LEVEL_TYPE all_rank_output_level = LOGOG_LEVEL_INFO, MPI_Comm mpi_comm = PETSC_COMM_WORLD)
```
chleh: So the maximum output level will be "Info"? What about debug output in debug builds?

BaseLib/LogogCustomCout.h
```diff
@@ -48,12 +54,15 @@ class LogogCustomCout : public logog::Target
 
 	virtual int Output( const LOGOG_STRING &data )
 	{
+#if defined(USE_MPI) || defined(USE_PETSC)
+		if (_is_rank0)
```
chleh: If you include "_is_rank0" in all cases, not only for MPI and PETSC, and set it to true in the other cases, then the ifdefs can be omitted.

Applications/CLI/ogs.cpp
TomFischer: @chleh I don't understand your point. Can you explain, please?

chleh: Forget about my comment. I did not understand, why that line was moved. But obviously it is because now the formatter might depend on MPI/PETSC which are initialized in the linear solver lib setup.

wenqing: Yes, MPI/PETSc initialization is required before  logog_setup, which needs rank and rank size of MPI.

Applications/ApplicationsLib/LogogSetup.h
wenqing: They will not be printed under the flags.

BaseLib/LogogCustomCout.h
wenqing: Debugging output is also available. I just committed some changes in the suppression flags. With the changed format, logog makes the same output as the in the master in serial computations, while in parallel computation it makes an output looking like
```
rank 0| debug: Parse process variables:
rank 0| debug: Constructing process variable pressure
rank 0| debug: Using value 0
rank 0| debug: GEOObjects::getGeoObject(): Could not find POINT "left"
```

BaseLib/LogogCustomCout.h
wenqing: In the serial computation, it is faster without this if-condition.

chleh:
:+1:

endJunction:
The `LogogSimpleFormatter` is derived from `logog::FormatterMSVC`, which, I guess, adds special behaviour for windows executables. How is this handled from now on?


Multiple ifdefs scattered through the code do not make it more readable. I can imagine a logog formatter selected in case of multi-threaded run with special formatting for SuppressedGCC at run time is a more lasting implementation.

wenqing:
@endJunction The root of  ``LogogFormatterSuppressedGCC`` can be replaced with ``logog::FormatterMSVC`` directly. However, we conduct parallel computing mostly on Linux/UNIX.
The content of this PR has been rewritten.

wenqing:
Multiple ifdefs is avoided.

Applications/ApplicationsLib/LogogSetup.h
```diff
@@ -12,11 +12,28 @@
...
+#include "BaseLib/LogogFormatterSuppressedMPI.h"
+	using LogogCout = BaseLib::LogogCustomCoutMPI;
+	using LogogFormat = BaseLib::LogogFormatterSuppressedMPI
+	                         <  TOPIC_GROUP_FLAG | TOPIC_FILE_NAME_FLAG
+	                          | TOPIC_LINE_NUMBER_FLAG>;
+#elif defined(USE_PETSC)
+#include "BaseLib/LogogCustomCoutPETSc.h"
+#include "BaseLib/LogogFormatterSuppressedPETSc.h"
+	using LogogCout = BaseLib::LogogCustomCoutPETSc;
+	using LogogFormat = BaseLib::LogogFormatterSuppressedPETSc
```
wenqing: With such template arguments, this format is equivalent to use  BaseLib::LogogSimpleFormatter.

PR #885, opened 2015-11-11
Replace wget by curl

bilke:
Old wget on Travis (Ubuntu 12.04) has some bugs checking certificates, see e.g.:

https://github.com/haskell/cabal/issues/1796


wenqing:
:+1:

PR #886, opened 2015-11-11
Surface and polyline output for class BoostXmlGmlInterface.

TomFischer:
The functionality is needed for tools running on computer that do not have Qt installed. Furthermore more properties are tested after writing and (re-)reading the data.

### DIFF
--- a/FileIO/XmlIO/Boost/BoostXmlGmlInterface.cpp
+++ b/FileIO/XmlIO/Boost/BoostXmlGmlInterface.cpp
@@ -21,7 +21,6 @@
 
 #include <boost/version.hpp>
 #include <boost/foreach.hpp>
-#include <boost/tokenizer.hpp>
 #include <boost/property_tree/xml_parser.hpp>
 
 #include <logog/include/logog.hpp>
@@ -32,6 +31,7 @@
 #include "GeoLib/PointVec.h"
 #include "GeoLib/Polyline.h"
 #include "GeoLib/Surface.h"
+#include "GeoLib/Triangle.h"
 
 namespace FileIO
 {
@@ -325,6 +325,8 @@ bool BoostXmlGmlInterface::write()
 			pnt_tag.put("<xmlattr>.name", point_name);
 	}
 
+	addSurfacesToPropertyTree(geometry_set);
+
 #if BOOST_VERSION <= 105500
 	boost::property_tree::xml_writer_settings<char> settings('\t', 1);
 #else
@@ -332,6 +334,49 @@ bool BoostXmlGmlInterface::write()
 #endif  // BOOST_VERSION
 	write_xml(_out, pt, settings);
 	return true;
+}
+
+
+
+void BoostXmlGmlInterface::addSurfacesToPropertyTree(
+	BaseLib::ConfigTree & geometry_set)
+{
+	GeoLib::SurfaceVec const*const sfc_vec(_geo_objects.getSurfaceVecObj(_exportName));
+	if (!sfc_vec) {
+		INFO("BoostXmlGmlInterface::addSurfacesToPropertyTree(): "
+			"No surfaces within the geometry \"%s\".", _exportName.c_str());
+		return;
+	}
+
+	std::vector<GeoLib::Surface*> const*const surfaces(sfc_vec->getVector());
+	if (! surfaces) {
+		INFO("BoostXmlGmlInterface::addSurfacesToPropertyTree(): "
+			"No surfaces within the geometry \"%s\".", _exportName.c_str());
+		return;
+	}
+
+	if (surfaces->empty()) {
+		INFO("BoostXmlGmlInterface::addSurfacesToPropertyTree(): "
+			"No surfaces within the geometry \"%s\".", _exportName.c_str());
+		return;
+	}
+
+	BaseLib::ConfigTree & surfaces_tag = geometry_set.add("surfaces", "");
+	for (std::size_t i=0; i<surfaces->size(); ++i) {
+		GeoLib::Surface const*const surface((*surfaces)[i]);
+		std::string sfc_name("");
+		sfc_vec->getNameOfElement(surface, sfc_name);
+		BaseLib::ConfigTree &surface_tag = surfaces_tag.add("surface", "");
+		surface_tag.put("<xmlattr>.id", i);
+		if (!sfc_name.empty())
+			surface_tag.put("<xmlattr>.name", sfc_name);
+		for (std::size_t j=0; j<surface->getNTriangles(); ++j) {
+			BaseLib::ConfigTree &element_tag = surface_tag.add("element", "");
+			element_tag.put("<xmlattr>.p1", (*(*surface)[j])[0]);
+			element_tag.put("<xmlattr>.p2", (*(*surface)[j])[1]);
+			element_tag.put("<xmlattr>.p3", (*(*surface)[j])[2]);
+		}
+	}
 }
 
 } // end namespace FileIO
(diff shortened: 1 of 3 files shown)


FileIO/XmlIO/Boost/BoostXmlGmlInterface.cpp
```diff
@@ -334,4 +338,73 @@ bool BoostXmlGmlInterface::write()
 	return true;
 }
 
+void BoostXmlGmlInterface::addSurfacesToPropertyTree(
+	boost::property_tree::ptree & geometry_set)
```
chleh: please use `BaseLib::ConfigTree`

FileIO/XmlIO/Boost/BoostXmlGmlInterface.h
```diff
@@ -68,6 +68,9 @@ class BoostXmlGmlInterface : public XMLInterface
 	                  const std::vector<std::size_t>& pnt_id_map,
 	                  std::map<std::string, std::size_t>*& sfc_names);
 
+	void addSurfacesToPropertyTree(boost::property_tree::ptree & pt);
+	void addPolylinesToPropertyTree(boost::property_tree::ptree & geometry_set);
```
chleh: Documentation missing (although maybe self-explanatory)

Tests/FileIO/TestXmlGmlReader.cpp
```diff
@@ -19,120 +19,242 @@
...
+
+	void checkPointProperties()
+	{
+		auto const& pointvec(*geo_objects.getPointVecObj(geo_name));
+		auto const& read_points(*pointvec.getVector());
+		ASSERT_EQ(test_pnts.size(), read_points.size());
+		for (std::size_t k(0); k<test_pnts.size(); ++k) {
+			GeoLib::Point const& read_pnt = *read_points[k];
+			// compare the coordinates
+			ASSERT_EQ(test_pnts[k][0], read_pnt[0]);
```
chleh: Maybe rather use `EXPECT_EQUAL` s.t. test does not stop on first error.

Tests/FileIO/TestXmlGmlReader.cpp
```diff
@@ -19,120 +19,242 @@
...
+		checkSurface(*read_sfcs, 0,
+			{{{0,3,1}}, {{1,3,4}}, {{1,4,2}}, {{2,4,5}}}, "");
+		checkSurface(*read_sfcs, 1, {{{3,6,8}}, {{3,8,5}}}, "SecondSurface");
+	}
+};
+
+TEST_F(FileIOXmlGml, QtXmlGmlWriterReaderTest)
+{
+	// Writer test
+	std::string test_data_file(BaseLib::BuildInfo::source_path + "/Tests/FileIO/xmlgmltestdata.gml");
```
chleh: Will the test write to the source tree? I think it would be better to write to some temporary directory.

Tests/FileIO/TestXmlGmlReader.cpp
```diff
@@ -19,120 +19,242 @@
...
+
 	// Reader test
 	result = xml.readFile(QString::fromStdString(test_data_file));
+	ASSERT_EQ(1, result);
+
+	checkPointProperties();
+	checkPolylineProperties();
+	checkSurfaceProperties();
+
+	boost::filesystem::remove(test_data_file);
```
chleh: Will the temporary files be removed if a test fails?

FileIO/XmlIO/Boost/BoostXmlGmlInterface.cpp
```diff
@@ -334,4 +338,73 @@ bool BoostXmlGmlInterface::write()
...
+	}
+
+	std::vector<GeoLib::Surface*> const*const surfaces(sfc_vec->getVector());
+	if (! surfaces) {
+		INFO("BoostXmlGmlInterface::addSurfacesToPropertyTree(): "
+			"No surfaces within the geometry \"%s\".", _exportName.c_str());
+		return;
+	}
+
+	using boost::property_tree::ptree;
```
chleh: BaseLib::ConfigTree

Tests/FileIO/TestXmlGmlReader.cpp
TomFischer: Changed as suggested.

Tests/FileIO/TestXmlGmlReader.cpp
TomFischer: Changed to "BaseLib::BuildInfo::test_tmp_path".

Tests/FileIO/TestXmlGmlReader.cpp
TomFischer: No, but since I changed all tests from "ASSERT_EQUAL" to "EXPECT_EQUAL" I hope this will not be an issue.

FileIO/XmlIO/Boost/BoostXmlGmlInterface.h
TomFischer: Since the method names are self-explanatory I omitted the documentation. Do you think documentation is necessary?

chleh: OK, it's fine.

chleh:
:+1:
Nice, I didn't know that we have a `tests_tmp_path`.

FileIO/XmlIO/Boost/BoostXmlGmlInterface.cpp
```diff
@@ -334,4 +337,71 @@ bool BoostXmlGmlInterface::write()
...
+{
+	GeoLib::SurfaceVec const*const sfc_vec(_geo_objects.getSurfaceVecObj(_exportName));
+	if (!sfc_vec) {
+		INFO("BoostXmlGmlInterface::addSurfacesToPropertyTree(): "
+			"No surfaces within the geometry \"%s\".", _exportName.c_str());
+		return;
+	}
+
+	std::vector<GeoLib::Surface*> const*const surfaces(sfc_vec->getVector());
+	if (! surfaces) {
```
endJunction: Could the `surfaces` be empty?

Tests/FileIO/TestXmlGmlReader.cpp
```diff
@@ -19,120 +19,242 @@
...
+	test_data_file += ".md5";
+	boost::filesystem::remove(test_data_file);
+}
+
+TEST_F(FileIOXmlGml, BoostXmlGmlWriterReaderTest)
+{
+	// Writer test
+	std::string test_data_file(BaseLib::BuildInfo::tests_tmp_path + "xmlgmltestdata.gml");
```
endJunction: Could be improved with boost's [`filesystem::unique_path()`](http://www.boost.org/doc/libs/master/libs/filesystem/doc/reference.html#unique_path) to avoid race-conditions on the filesystem or leftovers from last (e.g. segfaulted) run.

It, unfortunately, does not solve the problem of a dangling file like posix' `mkstemp`, which would remove the temporary after the program exits, or stream is closed; _but_ it is not portable.... :(

endJunction: Another _idea_ to delete the temporaries almost always but on segfaults is to create the filestream in the constructor of the test and remove in the destructor...

endJunction:
:fast_forward: otherwise...

FileIO/XmlIO/Boost/BoostXmlGmlInterface.cpp
TomFischer: This should not happen. If there are no surfaces in the vector `surfaces` then this should be a `nullptr`. However, I will insert an additional check to prevent errors in the future when the behaviour of `GEOObjects::getSurfaceVec()` or/and `SurfaceVec::getVector()` will be eventually changed, i.e., returns an empty vector instead of `nullptr`.

PR #888, opened 2015-11-12
[MaL] faster applyKnownSolution

chleh:
The provided implementation of `applyKnownSolution()` transposes the matrix and thereby avoids expensive searching for the right column.
This drastically increases perfomance for big matrices.


wenqing:
:+1:

PR #889, opened 2015-11-12
Rewrite of LisOption

chleh:
I largely rewrote the class `LisOption`. Now any Lis option can be given in the input file. The new implementation `LisOption` just stores options. `LisLinearSolver` then passes them through to Lis.

I simplified setting of options: There are no longer any conditions that decide what will happen. I also removed unused methods, members and enums.

### DIFF
--- a/MathLib/LinAlg/Lis/LisLinearSolver.cpp
+++ b/MathLib/LinAlg/Lis/LisLinearSolver.cpp
@@ -12,6 +12,10 @@
  *
  */
 
+#ifdef _OPENMP
+#include <omp.h>
+#endif
+
 #include "LisLinearSolver.h"
 
 #include "logog/include/logog.hpp"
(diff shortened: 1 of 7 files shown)


MathLib/LinAlg/Lis/LisLinearSolver.cpp
```diff
@@ -70,62 +42,64 @@ void LisLinearSolver::solve(LisVector &b, LisVector &x)
 
     INFO("------------------------------------------------------------------");
     INFO("*** LIS solver computation");
-#ifdef _OPENMP
```
wenqing: Looks great. Only one question hereby: where to print number of threads used by LIS if this ifdef is removed?

chleh: I see: This can be an interesting fact to know. I will re-enable it.

chleh: Re-Introduced OpenMP output. See below.

wenqing: :+1:

wenqing:
:fast_forward:

endJunction:
Merged.

PR #890, opened 2015-11-13
srqDist2d() and writeGeometryToFile()

TomFischer:
- At the moment the 2-D (xy-plane) (squared) distance of of two nodes / points is accomplished by copying the nodes or points and setting the *z*-coordinate explicitely to zero. Then the usual 3-D `sqrDist()` is used. To omit the copies and setting zero of the *z*-coordinate this PR suggest a function `sqrDist2d()`.
- The second commit is a function `writeGeometryToFile()` that takes the file name extension for the decision which format the data will be written. This will be used in several simulation model preprocessing tools.

### DIFF
--- a/FileIO/writeGeometryToFile.cpp
+++ b/FileIO/writeGeometryToFile.cpp
@@ -1,0 +1,37 @@
+/**
+ *
+ * \copyright
+ * Copyright (c) 2012-2015, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#include "writeGeometryToFile.h"
+
+#include "BaseLib/FileTools.h"
+
+#include "FileIO/XmlIO/Boost/BoostXmlGmlInterface.h"
+#include "FileIO/Legacy/OGSIOVer4.h"
+
+#include "GeoLib/GEOObjects.h"
+
+namespace FileIO
+{
+void writeGeometryToFile(std::string const& geo_name,
+	GeoLib::GEOObjects& geo_objs, std::string const& fname)
+{
+	std::string const extension(BaseLib::getFileExtension(fname));
+	if (extension == "gml" || extension == "GML") {
+		FileIO::BoostXmlGmlInterface xml(geo_objs);
+		xml.setNameForExport(geo_name);
+		xml.writeToFile(fname);
+	} else if (extension == "gli" || extension == "GLI") {
+		FileIO::Legacy::writeGLIFileV4(fname, geo_name, geo_objs);
+	} else {
+		ERR("Writing of geometry failed, since it was not possible to determine"
+			" the required format from file extension.");
+	}
+}
+}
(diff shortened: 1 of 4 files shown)


FileIO/writeGeometryToFile.cpp
```diff
@@ -0,0 +1,33 @@
...
+#include "FileIO/Legacy/OGSIOVer4.h"
+
+#include "GeoLib/GEOObjects.h"
+
+namespace FileIO
+{
+void writeGeometryToFile(std::string const& geo_name,
+	GeoLib::GEOObjects& geo_objs, std::string const& fname)
+{
+	if (BaseLib::getFileExtension(fname).compare("gml") == 0) {
```
chleh: I would prefer `string1 == "gml"`. It's more explicit. Or is there any reason against it?

FileIO/writeGeometryToFile.cpp
```diff
@@ -0,0 +1,33 @@
...
+		FileIO::BoostXmlGmlInterface xml(geo_objs);
+		xml.setNameForExport(geo_name);
+		xml.writeToFile(fname);
+	} else {
```
chleh: Maybe also compare to "gli" and print an error message if none of the options match?

MathLib/Point3d.h
```diff
@@ -50,6 +50,14 @@ double sqrDist(MathLib::Point3d const& p0, MathLib::Point3d const& p1)
 	return MathLib::scalarProduct<double,3>(v,v);
 }
 
+/// Computes the squared distance between the orthogonal projection of the two
+/// points \c p0 and \c p1 onto the \f$xy\f$-plane.
+inline
+double sqrDist2d(MathLib::Point3d const& p0, MathLib::Point3d const& p1)
+{
+	return std::pow(p0[0]-p1[0], 2) + std::pow(p0[1]-p1[1], 2);
```
chleh: I think, `std::pow` is very expensive. Maybe better square by hand.

chleh:
You said that "At the moment the 2-D (xy-plane) (squared) distance of of two nodes / points is accomplished by copying the nodes or points and setting the z-coordinate explicitely to zero."
Is there currently any code in master that calculates 2D distances? Because you have not updated any of those occurences.

FileIO/writeGeometryToFile.cpp
TomFischer: You are right. It is more readable.

FileIO/writeGeometryToFile.cpp
TomFischer: Changed behaviour.

MathLib/Point3d.h
TomFischer: The are overloads of `std::pow` for integer exponents that are nearly as fast as the multiplication. I changed the code according your suggestion.

TomFischer:
@chleh At least in several tool the "copy, set to zero and calc distance" is used. The only reason I included `srqDist2D()` is that I want to use it in at least two branches and hoping of a fast review of a smaller part of code. :smile:

chleh:
:+1:

MathLib/Point3d.h
endJunction: I also thought that pow(x,2) would result in non-optimal code. I proved wrong myself: the assembly code for the current multiplication algorithm and that one using std::pow(x,2) is identical, which is pretty cool. Seems there is an optimization for integral exponents.

clang-3.7, -O3.

PR #891, opened 2015-11-16
Integration of boundary conditions of first-type

TomFischer:
- Integration of boundary conditions of first-type into the system of linear equations using the `CRSMatrix` as system matrix
- Converter to use this integration together with the LIS data structures

### DIFF
--- a/MathLib/LinAlg/Lis/LisTools.cpp
+++ b/MathLib/LinAlg/Lis/LisTools.cpp
@@ -14,34 +14,89 @@
 
 #include "LisTools.h"
 
+#include <cassert>
+
 #include "logog/include/logog.hpp"
 
 #include "LisMatrix.h"
 #include "LisVector.h"
 
+#include "BaseLib/quicksort.h"
+#include "MathLib/LinAlg/Sparse/CRSMatrix.h"
+#include "MathLib/LinAlg/Sparse/CRSTools.h"
+
 namespace MathLib
 {
 
-void applyKnownSolution(LisMatrix &A, LisVector &b, const std::vector<std::size_t> &vec_knownX_id,
-		const std::vector<double> &vec_knownX_x, double penalty_scaling)
+namespace detail
 {
-    //Use penalty parameter
-    const double max_diag_coeff = A.getMaxDiagCoeff();
-    const double penalty = max_diag_coeff * penalty_scaling;
-    INFO("-> max. absolute value of diagonal entries = %e", max_diag_coeff);
-    INFO("-> penalty scaling = %e", penalty_scaling);
-    const std::size_t n_bc = vec_knownX_id.size();
-    for (std::size_t i_bc=0; i_bc<n_bc; i_bc++) {
-        const std::size_t rowId = vec_knownX_id[i_bc];
-        const double x = vec_knownX_x[i_bc];
-        //A(k, k) = penalty
-        A.setValue(rowId, rowId, penalty);
-        //b(k) = x*penalty
-        b.set(rowId, x*penalty);
-    }
+/// Converts the internal format of a lis matrix into the compressed row storage
+/// format. Since the column entries of a row in the internal lis format aren't
+/// sorted lis2crs not only transfers the indices and entries, it also
+/// sorts the columns and values, accordingly.
+MathLib::CRSMatrix<double, typename LisMatrix::IndexType>* lis2crs(LisMatrix &a)
+{
+	using IndexType = LisMatrix::IndexType;
+
+	LIS_MATRIX &A = a.getRawMatrix();
+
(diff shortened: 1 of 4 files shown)


MathLib/LinAlg/Lis/LisTools.cpp
```diff
@@ -14,34 +14,86 @@
...
-    const std::size_t n_bc = vec_knownX_id.size();
-    for (std::size_t i_bc=0; i_bc<n_bc; i_bc++) {
-        const std::size_t rowId = vec_knownX_id[i_bc];
-        const double x = vec_knownX_x[i_bc];
-        //A(k, k) = penalty
-        A.setValue(rowId, rowId, penalty);
-        //b(k) = x*penalty
-        b.set(rowId, x*penalty);
-    }
+MathLib::CRSMatrix<double, typename LisMatrix::IndexType>* lis2crs(LisMatrix &a)
```
endJunction: Some comments for the main steps of the algorithm would be helpful. (Yes, it's just a deep copy but maybe there is some special treatment like sorting or indices counting from 1...)

TomFischer: Added documentation.

MathLib/LinAlg/Lis/LisTools.cpp
```diff
@@ -14,34 +14,89 @@
...
-        //A(k, k) = penalty
-        A.setValue(rowId, rowId, penalty);
-        //b(k) = x*penalty
-        b.set(rowId, x*penalty);
-    }
+	using IndexType = LisMatrix::IndexType;
+
+	LIS_MATRIX &A = a.getRawMatrix();
+
+	IndexType const n(A->n);
```
chleh: Could you add a hint that this is the number of rows, please.

MathLib/LinAlg/Sparse/CRSTools-impl.h
```diff
@@ -0,0 +1,69 @@
...
+	typename VEC_T::IndexType const*const jA(mat->getColIdxArray()); // col idx
+	double * entries(const_cast<double*>(mat->getEntryArray()));
+
+	// set row entries, except the diagonal entry, to zero
+	for (std::size_t r(0); r<rows.size(); ++r) {
+		auto const row = rows[r];
+		for (unsigned j(iA[row]); j<iA[row+1]; ++j) {
+			if (jA[j] == row) {
+				// rhs_{row} = a_{row,row} * val
+				rhs.set(row, entries[j]*vals[r]);
```
chleh: What if a matrix has a zero on the diagonal?
In this case I would set the diagonal to one for the moment and leave a better solution for later.

wenqing: In addition, for this serial solver, it it better to also set off-diagonal entries of the corresponding column be zero. Under such consideration,  symmetrical matrix is still symmetry after calling this function.

TomFischer: @wenqing What do you mean by "this serial solver".  The function at hand does integrate boundary conditions of first-type. The algorithm works as follows:
1. Compute the transposed matrix and set the non-zero column entries (except the diagonal entry) for the given rows to zero and integrate the information into the rhs.
2. Compute the transposed matrix of the transposed matrix (= original matrix) and again, remove the non-zero column entries.
The symmetry of the matrix is preserved. Did I made a mistake?

wenqing: It's clear to me now. Thanks for explanation.

MathLib/LinAlg/Lis/LisTools.cpp
TomFischer: Renamed variable and added docu.

MathLib/LinAlg/Sparse/CRSTools-impl.h
TomFischer: @chleh I am unsure what it means when the diagonal entries equals zero. I wonder which conditions have to be fulfilled if this happens? Maybe this is the consequence of an inconsistent first-type boundary condition?! If there are no objections I would discuss this separately from this PR.

chleh: In general matrices can have zeroes on the diagonal. Maybe that's rather unlikely for FEM. But can you guarantee that there are never zeroes even for indefinite matrices, or when there are multiple primary variables or when you are solving the equations arising from a transient or nonlinear problem?
Does anybody else see those issues?

chleh:
:+1:
Please make an issue from my last comment (unless it is obviously irrelevant)

endJunction:
Simply check for A(row,row) == 0 and set to 1 in that case.

PR #893, opened 2015-11-16
Lis performance

chleh:
Changes of this PR are made for improving Lis performance.
Currently there is an issue with slow performance of Lis matrix assembly: Repeated invocation of `LisMatrix::add()` triggers many unnecessary checks in Lis. This PR makes using some workaround easier.
* There is the class `EigenLisLinearSolver` in master. This PR adds a CMake configuration option `OGS_USE_EigenLis` which makes ogs use this linear solver.
* EigenMatrix and LisMatrix now do not set values of 0.0 anymore. In the TES process I have many zeroes in local matrices. This filter speeds up the linear solvers quite a bit
* `applyKnownSolution()` for `EigenMatrix` has been updated to not degrade matrix condition by inserting 1.0 on the diagonal.
* A simple Test has been added to check if the internal storage of `Eigen::SparseMatrix` is a CSR matrix. Sure, this test is not perfect, though better than nothing, and it complements the test cases `TestLinearSolver.cpp` to simplify detection of possible errors that might arise if the Eigen project changes their internal matrix storage format.

All in all the changes from this PR saved very roughly 30% of runtime in my case.

### DIFF
--- a/MathLib/LinAlg/Lis/LisMatrix.cpp
+++ b/MathLib/LinAlg/Lis/LisMatrix.cpp
@@ -87,6 +87,7 @@ void LisMatrix::setZero()
 
 int LisMatrix::setValue(IndexType rowId, IndexType colId, double v)
 {
+    if (v == 0.0) return 0;
     lis_matrix_set_value(LIS_INS_VALUE, rowId, colId, v, _AA);
     if (rowId==colId)
         lis_vector_set_value(LIS_INS_VALUE, rowId, v, _diag);
@@ -96,6 +97,7 @@ int LisMatrix::setValue(IndexType rowId, IndexType colId, double v)
 
 int LisMatrix::add(IndexType rowId, IndexType colId, double v)
 {
+    if (v == 0.0) return 0;
     lis_matrix_set_value(LIS_ADD_VALUE, rowId, colId, v, _AA);
     if (rowId==colId)
         lis_vector_set_value(LIS_ADD_VALUE, rowId, v, _diag);
--- a/Tests/MathLib/TestEigenCSR.cpp
+++ b/Tests/MathLib/TestEigenCSR.cpp
@@ -84,10 +84,11 @@ TEST(MathLibEigen, Eigen2CSR)
     // adapt entries of CSR matrix
     for (auto& v : values) v *= 4.0;
 
-    ASSERT_EQ(nnz, mat.nonZeros());
-    ASSERT_EQ(nnz, ja.size());
-
     mat.makeCompressed();
+
+    ASSERT_EQ(nnz, mat.nonZeros());
+    ASSERT_EQ(nnz, values.size());
+    ASSERT_EQ(nnz, ja.size());
 
     int* ptr = mat.outerIndexPtr();
     int* col = mat.innerIndexPtr();
--- a/Tests/MathLib/TestEigenCSR.cpp
+++ b/Tests/MathLib/TestEigenCSR.cpp
@@ -70,15 +70,12 @@ TEST(MathLibEigen, Eigen2CSR)
     }
 
     // change matrix
-    for (int repeat=0; repeat < 2; ++repeat) {
-        for (int row=0; row<nrows; ++row) {
-            for (int col = -1; col<=1; ++col) {
-                int cidx = row + col;
-                if (cidx < 0 || cidx >= ncols) continue;
+    for (int row=0; row<nrows; ++row) {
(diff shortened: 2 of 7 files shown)


Tests/MathLib/TestEigenCSR.cpp
```diff
@@ -0,0 +1,105 @@
...
+            if (cidx < 0 || cidx >= ncols) continue;
+
+            const double val = (col == 0) ? 2.0 : -1.0;
+            values.push_back(val);
+            mat.coeffRef(row, cidx) = val;
+            ja.push_back(cidx);
+        }
+    }
+
+    // change matrix
```
TomFischer: What is the purpose of doing some math with the matrix entries?

Tests/MathLib/TestEigenCSR.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+                // doubles matrix entry
+                mat.coeffRef(row, cidx) += mat.coeff(row, cidx);
+            }
+        }
+    }
+    // adapt entries of CSR matrix
+    for (auto& v : values) v *= 4.0;
+
+    ASSERT_EQ(nnz, mat.nonZeros());
```
TomFischer: Maybe you can also check `ASSERT_EQ(values.size(), mat.nonZeros())`?!

Tests/MathLib/TestEigenCSR.cpp
```diff
@@ -0,0 +1,105 @@
...
+    ASSERT_EQ(nnz, ja.size());
+
+    mat.makeCompressed();
```
TomFischer: Maybe the previous check from line 87 and 88 should be moved after `math.makeCompressed()`?

MathLib/LinAlg/Lis/LisMatrix.cpp
```diff
@@ -87,6 +87,7 @@ void LisMatrix::setZero()
 
 int LisMatrix::setValue(IndexType rowId, IndexType colId, double v)
 {
+    if (v == 0.0) return 0;
```
TomFischer: How does this influence the generation of the sparsitty pattern of the LisMatrix? I assume there will be no entry in the sparsitty pattern. Will/Can this be a problem when the matrix is reused in the next assembly iteration?
Same question for the handling in `LisMatrix::add()`.

TomFischer:
> All in all the changes from this PR saved very roughly 30% of runtime in my case.

@chleh This is great!

MathLib/LinAlg/Lis/LisMatrix.cpp
chleh: It does not influence the generation of the sparsity pattern, only its application:
In general the sparsity pattern now will be slightly too large, so more space than needed will be reserved. However, I think we won't run out of space because of that.
`if (v==0.0)` will also not influence subsequent iterations, because currently the method `setZero()` of LisMatrix rebuilds the whole matrix and the sparsity pattern has to be applied in every step.

Tests/MathLib/TestEigenCSR.cpp
chleh: Probably there is not much point in doiing so. If Eigen did some fancy stuff (like switching the internal storage format) upon changing matrix entries, this might be triggered here. But probably Eigen does not do so, so this loop could also be removed.

Tests/MathLib/TestEigenCSR.cpp
chleh: Check added.

Tests/MathLib/TestEigenCSR.cpp
chleh: Check added.

MathLib/LinAlg/Lis/LisMatrix.cpp
TomFischer: Thanks for explanation!

Tests/MathLib/TestEigenCSR.cpp
TomFischer: Personally, I would remove the loops (line 73 - 83) and lines 84, 85. In case you want to keep them I would remove at least the outermost ("repeat") loop and modify the matrix entries in the same manner than the values in the vector `values` (line 85) are modified, i.e., `mat.coeffRef(row, cidx) *= 4.0;`

wenqing:
:+1:

TomFischer:
:+1:

Issue #881, opened 2015-10-30
Eigen assertions


chleh:
I just noticed that there is the option `EIGEN_NO_DEBUG` in our cmake files.
Unfortunately, as I think, it is enabled by default:
`option(EIGEN_NO_DEBUG "Disables Eigen's assertions" ON)`
That means that Eigen assertions are disabled, even in debug mode. Is there a good reason for this default behaviour?
Some time ago I had some headache because Eigen did not check bounds on dynamically sized matrices, and it was rather hard for me to find the places where my code was wrong. Somehow I also didn't find out, how to enable bounds checking. Now I know: Disabling `EIGEN_NO_DEBUG`.

I suggest to change the default to
`option(EIGEN_NO_DEBUG "Disables Eigen's assertions" OFF)`.

endJunction:
There is also the "EIGEN_NO_STATIC_ASSERT - if defined, compile-time static assertions are replaced by runtime assertions; this saves compilation time. Not defined by default." (from http://eigen.tuxfamily.org/dox/TopicPreprocessorDirectives.html).
This could be interesting too.

Feel free to make a PR.

Looking up the log for this, there seem to be no particular reason why the tests were disabled...

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/881).

PR #894, opened 2015-11-17
Eigen assertions enabled by default

chleh:
See issue #881.
I think setting Eigen assertions to "ON" by default is a better choice than disabling them.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,7 +73,7 @@ option(OGS_USE_MPI "Use MPI" OFF)
 
 # Eigen
 option(OGS_USE_EIGEN "Use EIGEN for local matrix and vector" ON)
-option(EIGEN_NO_DEBUG "Disables Eigen's assertions" ON)
+option(EIGEN_NO_DEBUG "Disables Eigen's assertions" OFF)
 
 # Logging
 option(OGS_DISABLE_LOGGING "Disables all logog messages." OFF)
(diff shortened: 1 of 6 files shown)


chleh:
Jenkins already found out about the sane new default :smile:

chleh:
Sorry, I don't see the issue in the test immediately. Maybe someone of you sees easily why the test fails. Can you please help me fixing it?

bilke:
There was an assertion triggered:

```
testrunner: /gpfs1/data/ogs/jenkins/workspace/OGS-6/Linux-PRs/build/External/eigen/src/Eigen/Eigen/src/Core/DenseCoeffsBase.h:337:
Eigen::DenseCoeffsBase<Derived, 1>::Scalar& Eigen::DenseCoeffsBase<Derived, 1>::operator()(Eigen::DenseCoeffsBase<Derived, 1>::Index, Eigen::DenseCoeffsBase<Derived, 1>::Index)
[with Derived = Eigen::Matrix<double, -1, -1, 1>; Eigen::DenseCoeffsBase<Derived, 1>::Scalar = double; Eigen::DenseCoeffsBase<Derived, 1>::Index = long int]:
Assertion `row >= 0 && row < rows() && col >= 0 && col < cols()' failed.
```

chleh:
Thanks, but I also saw, that the assertion was triggered. Maybe, I should have said that.
I thought, maybe someone knows how to fix the constructor of `NumLibFemIsoTest` properly.
in particular: What to change s.t. the product `globalD = R^T D R` is meaningful AND the tests pass?

chleh:
The last commit fixes an assertion violation with Neumann BCs. I think there is a problem with the constructors and `resize()` methods of the shape matrices: They both come in versions with and without a `global_dim` parameter. This difference can be easy overlooked. So I suggest to drop the methods without `global_dim` parameter. Though this must be done in a subsequent PR.

endJunction:
Replaced by https://github.com/ufz/ogs/pull/904 without the GroundwaterFlow changes.

PR #895, opened 2015-11-19
Added CMake options to specify third-party library searching

### LINKED ISSUE #892: Possibility to disable usage of a system library in cmake
For testing (but also for some other reasons) it is necessary to use _not_ a library (boost, vtk, eigen) provided by the system, but to rely on library which would be downloaded (and compiled) by cmake when there is not applicable version of that library provided by the system.

For example: The default vtk-6.3 compilation on my system (Archlinux) does not include parallel xml io. For testing of the petsc code a vtk library with parallel io is required. Would be nice to be able to disable usage of vtk via cmake configuration.

bilke:
Boost, VTK, Eigen
Closes #892.


endJunction:
Thanks Lars!
This almost works ;)

1. Didn't read the docu, just specified `-DOGS_LIB_VTK=Download` (instead of `Local`) and it was downloaded. Not important.
2. vtkdiff didn't build (although the vtk target was made):
```
...
[100%] Linking CXX shared library ../../lib/libvtkIOXML-6.3.so
[100%] Built target vtkIOXML
[  2%] No install step for 'vtk'
[  2%] Completed 'vtk'
[  2%] Built target vtk
...
Scanning dependencies of target vtkdiff
[ 12%] Building CXX object ThirdParty/vtkdiff/CMakeFiles/vtkdiff.dir/vtkdiff.cpp.o
/home/naumov/w/ogs/ThirdParty/vtkdiff/vtkdiff.cpp:20:10: fatal error: 'vtkDataArray.h' file not found
#include <vtkDataArray.h>
```
and continuing with `make -k`
```
/home/naumov/w/ogs/FileIO/VtkIO/VtuInterface.h:19:10: fatal error: 'vtkXMLWriter.h' file not found
#include <vtkXMLWriter.h>
```
_etc._

Ideas?

bilke:
- to 1.: local building is the else-case so the value is not explicitly checked
- to 2.: should be fixed

endJunction:
Works now for non-petsc builds. When using petsc the only error remains is:
```
[ 16%] Building CXX object FileIO/CMakeFiles/FileIO.dir/VtkIO/VtuInterface.cpp.o
/home/naumov/w/ogs/FileIO/VtkIO/VtuInterface.cpp:20:10: fatal error: 'vtkXMLPUnstructuredGridWriter.h' file not found
#include <vtkXMLPUnstructuredGridWriter.h>
```
Do you have a fix for this too?

endJunction:
Parallel builds are happy!

PR #896, opened 2015-11-23
Lin solver config

endJunction:
Generalization of linear solver settings from project files.

LIS and PETSc do support setting of the linear solver options via single string equivalent to command line settings.
Eigen does not, and we have a parser for ConfigTree description of options.

The way it is presented here is the most general way to set linear solver's options without any assumptions on which parameters are allowed.

If there are strong objections regarding this way of setting options, suggestions for other ways are welcome.

The options added to all ctest project files are:
```xml
<linear_solver>
    <lis>-i cg -p jacobi -tol 1e-16 -maxiter 10000</lis>
    <eigen>
        <solver_type>CG</solver_type>
        <precon_type>jacobi</precon_type>
        <max_iteration_step>10000</max_iteration_step>
        <error_tolerance>1e-16</error_tolerance>
    </eigen>
    <petsc>-gw_ksp_type cg -gw_pc_type bjacobi -gw_ksp_rtol 1e-16 -gw_ksp_max_it 10000</petsc>
</linear_solver>
```


chleh:
For later: Since now the linear solver options are given per process, we could get rid of the "gw_" prefix in the Petsc options.

MathLib/LinAlg/Lis/LisLinearSolver.h
```diff
@@ -65,7 +65,7 @@ class LisLinearSolver final
 
 private:
     LisMatrix& _A;
-    LisOption _option;
+    LisOption _lis_options;
```
chleh: Minor inconsistency: singular vs. plural forms. I suggest sticking with singular since the class is also called `LisOption`. See also LisEigenLinearSolver.

MathLib/LinAlg/Lis/LisMatrix.h
```diff
@@ -45,6 +45,22 @@ struct SetMatrixSparsity<LisMatrix, SPARSITY_PATTERN>;
 class LisMatrix
 {
 public:
+    /// Matrix type
+    enum class MatrixType : int
```
chleh: I it used somewhere? Since now you've moved it here, you could also have dropped it entirely.

MathLib/LinAlg/Lis/LisOption.h
```diff
@@ -17,88 +17,43 @@
...
 struct LisOption
 {
-    using Key = std::string;
-    using Value = std::string;
+	LisOption(BaseLib::ConfigTree const* const options)
+	{
+		if (options)
+		{
+			_option_string += " " + options->get<std::string>("lis", "");
+			boost::algorithm::trim_right(_option_string);
```
chleh: I still think the following is better:
```
auto const s = options->get_optional<std::string>("lis");
if (s) _option_string += *s;
```

MathLib/LinAlg/Lis/LisOption.h
```diff
@@ -17,88 +17,43 @@
...
 #include "BaseLib/ConfigTree.h"
 
 namespace MathLib
 {
-
-/**
- * \brief Option for Lis solver
+/** Option for Lis solver.
+ *
+ * Options are stored as a strings obtained from lis-tag of the passed options.
```
chleh: as a string
from the

MathLib/LinAlg/Lis/LisOption.h
```diff
@@ -17,88 +17,43 @@
...
+ * The string will be passed as is to the lis.  In particular it will not be
```
chleh: to Lis

Tests/MathLib/TestLinearSolver.cpp
```diff
@@ -203,11 +203,13 @@ TEST(MathLib, CheckInterface_GaussAlgorithm)
 {
     boost::property_tree::ptree t_root;
     boost::property_tree::ptree t_solver;
-    t_root.put_child("LinearSolver", t_solver);
+    t_root.put_child("ogs", t_solver);
```
chleh: Why did you change it from "LinearSolver" to "ogs"?

MathLib/LinAlg/Lis/LisLinearSolver.h
TomFischer: Although the class manages several option**s**. :smile:

MathLib/LinAlg/Lis/LisMatrix.h
endJunction: Yes, in ctor of LisMatrix as a default value for lis matrix type.
I can also imagine to have
```xml
<lis>
   <options> -i cg -tol 3</options>
   <matrix_type>CRS</matrix_type>
</lis>
```

MathLib/LinAlg/Lis/LisOption.h
endJunction: This would need an extra space and a trim anyway:
```c++
auto const s = options->get_optional<std::string>("lis");
if (s) _option_string += " " + *s;
trim_right(_option_string);
```

Tests/MathLib/TestLinearSolver.cpp
endJunction: Yes. When we have an internal ogs linear equations solver, this could become the tag.
For the test only "eigen", "lis", or "petsc" ptrees are passed; we are already in the "linear_solver" ptree.

MathLib/LinAlg/Lis/LisOption.h
chleh: Why would we need trim_right? Currently the boost xml reader already trims all whitespace.
But nevermind.

MathLib/LinAlg/Lis/LisMatrix.h
chleh: But for the time being the matrix type is not passed on to Lis, so it wouldn't hurt to remove that code.
But I won't insist on that.

TomFischer:
:+1: for the implementation. On my computer the test AssemblerLibSerialLinearSolver.Steady2DdiffusionQuadElem still fails. Ctests seem to be ok.

wenqing:
I prefer not to set
```
    <petsc> ... </petsc>
```
as a mandatory element of  ``<linear_solver>`` in the input file.  To run a parallel job via queue system, sometimes it is convenience to set PETSc options in job script than to modify input data for the options.

endJunction:
Neither the `<linear_solver>` nor its subtrees are mandatory, though the project file options have precedence over the command line options.
One can still use the command line parameters to set petsc or lis options

wenqing:
:+1:

PR #897, opened 2015-11-24
Fixes for AlgorithmProperties and get- / setValue()-removal

bilke:
**TODOs:**
- [x] FeflowInterface.cpp -> use property groups (@TomFischer ), not necessary
- [x] TetgenInterface     -> not checked yet (@rinkk)
- [x] LayeredVolume.cpp   -> see TODO (@rinkk)
- [x] Threshold filter uses first array per default, not selected array from parent, created issue #901
- [x] Utils (@bilke)


.idea/codeStyleSettings.xml
```diff
@@ -63,6 +63,7 @@
           <indentOptions>
```
endJunction: Merge the .idea/ separately...

TomFischer:
@bilke In my opinion the setting of property values to mesh elements in the Feflow interface is already implemented well. Is there a special reason/requirement to use the "group property version" of `PropertyVector`?

rinkk:
A list where material ids are still used as parameters in the constructor. Also, 3 warnings.
```
Error	C:\Projects\ogs\MeshLib\MeshGenerators\VtkMeshConverter.cpp	323
Error	C:\Projects\ogs\MeshLib\MeshGenerators\VtkMeshConverter.cpp	358
Error	C:\Projects\ogs\MeshLib\MeshGenerators\VtkMeshConverter.cpp	403
Error	C:\Projects\ogs\MeshLib\MeshGenerators\VtkMeshConverter.cpp	344
Error	C:\Projects\ogs\FileIO\TetGenInterface.cpp	489
Error	C:\Projects\ogs\Tests\NumLib\TestFunctionInterpolation.cpp	63
Error	C:\Projects\ogs\MeshLib\MeshGenerators\MeshLayerMapper.cpp	209
Error	C:\Projects\ogs\MeshLib\MeshGenerators\MeshLayerMapper.cpp	204
Error	C:\Projects\ogs\MeshLib\MeshGenerators\MeshLayerMapper.cpp	98
Error	C:\Projects\ogs\MeshLib\MeshGenerators\MeshLayerMapper.cpp	195
Error	C:\Projects\ogs\MeshLib\MeshGenerators\MeshLayerMapper.cpp	101
Error	C:\Projects\ogs\MeshLib\MeshGenerators\LayeredVolume.cpp	93
Error	C:\Projects\ogs\MeshLib\MeshGenerators\LayeredVolume.cpp	127
Error	C:\Projects\ogs\MeshLib\MeshGenerators\LayeredVolume.cpp	132
Error	C:\Projects\ogs\MeshLib\MeshGenerators\LayeredVolume.cpp	101
Error	C:\Projects\ogs\MeshLib\MeshGenerators\LayeredVolume.cpp	166
Error	C:\Projects\ogs\FileIO\GMSInterface.cpp	293
Error	C:\Projects\ogs\FileIO\GMSInterface.cpp	303
Error	C:\Projects\ogs\FileIO\GMSInterface.cpp	283
Error	C:\Projects\ogs\MeshLib\MeshGenerators\ConvertRasterToMesh.cpp	136
Error	C:\Projects\ogs\MeshLib\MeshGenerators\ConvertRasterToMesh.cpp	137
Error	C:\Projects\ogs\MeshLib\MeshGenerators\ConvertRasterToMesh.cpp	145
Error	C:\Projects\ogs\MeshGeoToolsLib\BoundaryElementsAtPoint.cpp	32
Error	C:\Projects\ogs\MeshGeoToolsLib\AppendLinesAlongPolyline.cpp	65

Warning	28	warning C4018: '<' : signed/unsigned mismatch	C:\Projects\ogs\SimpleTests\MatrixTests\MatMult.cpp	147
Warning	15	warning C4258: 'i' : definition from the for loop is ignored; the definition from the enclosing scope is used	C:\Projects\ogs\FileIO\FEFLOWInterface.cpp	219
Warning	29	warning C4373: 'MathLib::CRSMatrixOpenMP<double,unsigned int>::amux': virtual function overrides 'MathLib::CRSMatrix<FP_TYPE,IDX_TYPE>::amux', previous versions of the compiler did not override when parameters only differed by const/volatile qualifiers	C:\Projects\ogs\MathLib\LinAlg\Sparse\CRSMatrixOpenMP.h	43
```

TomFischer:
```Error   C:\Projects\ogs\MeshGeoToolsLib\BoundaryElementsAtPoint.cpp 32```
The second parameter of the constructor seems to be already the mesh element id and not the material id.

**Comment by bilke** [comment]
:+1:

**Comment by bilke** [comment]
Jenkins test this please.

File: Applications/DataExplorer/DataView/MeshElementRemovalDialog.cpp [hunk 2]
```diff
@@ -170,13 +170,13 @@ void MeshElementRemovalDialog::on_materialIDCheckBox_toggled(bool is_checked)
...
-		const std::vector<MeshLib::Element*> elements (_project.getMesh(this->meshNameComboBox->currentText().toStdString())->getElements());
-		auto it = std::max_element(elements.begin(), elements.end(),
-			[](MeshLib::Element const*const x, MeshLib::Element const*const y) { return x->getValue() < y->getValue(); }
-		);
-		unsigned max_material ((*it)->getValue());
+		auto mesh = _project.getMesh(this->meshNameComboBox->currentText().toStdString());
+		const std::vector<MeshLib::Element*> elements = mesh->getElements();
 
-		for (unsigned i=0; i<=max_material; ++i)
+		auto materialIds = mesh->getProperties().getPropertyVector<int>("MaterialIDs");
```
endJunction: `materialIds` could be empty, but it is not checked below...

rinkk: The problem is larger than that. MaterialIDs mustn't be selectable if no material array exists. Everything afterwards depends on that and I'd have to look closer to see where more detailed checks need to be performed. I'll fix the class on Monday morning if that's alright with everyone.

FileIO/FEFLOWInterface.cpp
```diff
@@ -156,8 +156,69 @@ MeshLib::Mesh* FEFLOWInterface::readFEFLOWFile(const std::string &filename)
...
+	if (_geoObjects && points)
+		_geoObjects->addPointVec(
+			std::unique_ptr<std::vector<GeoLib::Point*>>(points), project_name);
+	if (_geoObjects && lines)
+		_geoObjects->addPolylineVec(
+			std::unique_ptr<std::vector<GeoLib::Polyline*>>(lines),
+			project_name);
+
 	INFO("setting material IDs");
-	setMaterialID(fem_class, fem_dim, lines, vec_elementsets, vec_elements);
```
endJunction: The `setMaterialID()` function was actually more readable than inplace code. Any strong reasons to move it here? (The commit message didn't provide any information on 'why'.)

FileIO/GMSInterface.cpp
```diff
@@ -240,6 +240,7 @@ MeshLib::Mesh* GMSInterface::readGMS3DMMesh(const std::string &filename)
 	INFO("GMSInterface::readGMS3DMMesh(): Read GMS-3DM mesh.");
 	std::vector<MeshLib::Node*> nodes;
 	std::vector<MeshLib::Element*> elements;
+	std::vector<int> mat_ids;
```
endJunction: `mat_id` below is declared as `unsigned` but should be consistent with `mat_ids` vector type.

FileIO/GMSInterface.cpp
```diff
@@ -317,7 +321,17 @@ MeshLib::Mesh* GMSInterface::readGMS3DMMesh(const std::string &filename)
...
+	if (mat_ids.size() == elements.size()) {
+		boost::optional<MeshLib::PropertyVector<int> &> opt_pv
+			= properties.createNewPropertyVector<int>(
+				"MaterialIDs", MeshLib::MeshItemType::Cell);
+		opt_pv->resize(mat_ids.size());
+		std::copy(mat_ids.cbegin(), mat_ids.cend(), opt_pv->begin());
+	}
+	MeshLib::Mesh* mesh(new MeshLib::Mesh(mesh_name, nodes, elements, properties));
+
+	return mesh;
```
endJunction: Simply `return new MeshLib::Mesh(mesh_name...);`

FileIO/GMSInterface.cpp
```diff
@@ -317,7 +321,17 @@ MeshLib::Mesh* GMSInterface::readGMS3DMMesh(const std::string &filename)
 	INFO("GMSInterface::readGMS3DMMesh(): \tfinished.");
 
 	const std::string mesh_name = BaseLib::extractBaseNameWithoutExtension(filename);
-	return new MeshLib::Mesh(mesh_name, nodes, elements);
+	MeshLib::Properties properties;
+	if (mat_ids.size() == elements.size()) {
+		boost::optional<MeshLib::PropertyVector<int> &> opt_pv
```
endJunction: A message that material ids are going to be ignored (in the other case) could be helpful.

FileIO/TetGenInterface.cpp
```diff
@@ -230,8 +231,18 @@ MeshLib::Mesh* TetGenInterface::readTetGenMesh (std::string const& nodes_fname,
 		return nullptr;
 	}
 
+	MeshLib::Properties properties;
+	// remove material vector again if all mat-ids are "0"
+	if (*(std::max_element(materials.cbegin(), materials.cend())) > 0)
```
endJunction: Better to use `any_of()` algorithm here both in readability and speed (it will stop looking after first non-zero element is found).

FileIO/TetGenInterface.cpp
```diff
@@ -618,7 +634,7 @@ void TetGenInterface::write3dElements(std::ofstream &out,
 	const std::streamoff before_elems_pos (out.tellp());
 	const unsigned n_spaces (static_cast<unsigned>(std::floor(log(nElements*8))) + 1);
 	out << std::string(n_spaces, ' ') << "\n";
-
+	auto materialIds = mesh.getProperties().getPropertyVector<int>("MaterialIDs");
```
endJunction: `materialIds` could be empty, but it is not checked below.

FileIO/TetGenInterface.cpp
```diff
@@ -655,16 +671,17 @@ void TetGenInterface::write3dElements(std::ofstream &out,
 	out.seekp(after_elems_pos);
 }
 
-void TetGenInterface::writeElementToFacets(std::ofstream &out, const MeshLib::Element &element, unsigned &element_count) const
+void TetGenInterface::writeElementToFacets(std::ofstream &out, const MeshLib::Element &element, unsigned &element_count, const MeshLib::Mesh &mesh) const
 {
+	auto materialIds = mesh.getProperties().getPropertyVector<int>("MaterialIDs");
```
endJunction: Could be empty...

MeshGeoToolsLib/AppendLinesAlongPolyline.cpp
```diff
@@ -32,9 +32,18 @@ MeshLib::Mesh* appendLinesAlongPolylines(const MeshLib::Mesh &mesh, const GeoLib
...
 	std::vector<MeshLib::Element*> vec_new_eles = MeshLib::copyElementVector(mesh.getElements(), vec_new_nodes);
-	unsigned max_matID = 0;
-	for (auto e : vec_new_eles)
-		max_matID = std::max(max_matID, e->getValue());
+
+	std::vector<int> new_mat_ids;
+	{
+		auto mat_ids = mesh.getProperties().getPropertyVector<int>("MaterialIDs");
+		if (mat_ids) {
+			new_mat_ids.resize((*mat_ids).size());
```
endJunction: reserve + push_back is faster, because no initialization happens as with resize.

MeshGeoToolsLib/AppendLinesAlongPolyline.cpp
```diff
@@ -57,13 +66,23 @@ MeshLib::Mesh* appendLinesAlongPolylines(const MeshLib::Mesh &mesh, const GeoLib
...
+				new MeshLib::Line(element_nodes, vec_new_eles.size()));
+			new_mat_ids.push_back(max_matID+k+1);
 		}
 	}
 
 	// generate a mesh
-	const std::string new_mesh_name = mesh.getName() + "_with_lines";
-	return new MeshLib::Mesh(new_mesh_name, vec_new_nodes, vec_new_eles);
+	const std::string name = mesh.getName() + "_with_lines";
+	MeshLib::Mesh* new_mesh(new MeshLib::Mesh(name, vec_new_nodes, vec_new_eles));
```
endJunction: Using `unique_ptr` would be better because of the ownership indication.

MeshLib/ElementStatus.cpp
```diff
@@ -34,10 +34,11 @@ ElementStatus::ElementStatus(Mesh const* const mesh,
                              std::vector<unsigned> const& vec_inactive_matIDs)
     : ElementStatus(mesh, !vec_inactive_matIDs.empty())
 {
+	auto materialIds = mesh->getProperties().getPropertyVector<int>("MaterialIDs");
```
endJunction: `materialIds` could be empty...

MeshLib/ElementStatus.cpp
```diff
@@ -34,10 +34,11 @@ ElementStatus::ElementStatus(Mesh const* const mesh,
...
 	const std::size_t nElems (_mesh->getNElements());
 	for (auto material_id : vec_inactive_matIDs) {
 		for (auto e : _mesh->getElements())
-			if (e->getValue() == material_id)
+			if ((*materialIds)[e->getID()] == material_id)
```
endJunction: `material_id` is unsigned but `materialIds` contains signed values.

MeshLib/Elements/Element.cpp
```diff
@@ -190,7 +190,7 @@ bool Element::isBoundaryElement() const
 #ifndef NDEBUG
 std::ostream& operator<<(std::ostream& os, Element const& e)
 {
-	os << "Element #" << e._id << " value " << e._value << " @ " << &e
+	os << "Element #" << e._id << " value " << " @ " << &e
```
endJunction: forgot to remove " value" string

MeshLib/Elements/Element.h
```diff
@@ -89,7 +89,7 @@ class Element
 	virtual const Element* getFace(unsigned i) const = 0;
 
 	/// Returns the ID of the element.
-	virtual std::size_t getID() const { return this->_id; }
```
endJunction: Not sure on removal of the virtual function specifier here.
I think it is better to add a final specifier here such that `getID()` cannot be overridden by any derived class. Without any specifiers one can implement `getID()` in a derived class with funny consequences.

MeshLib/MeshGenerators/ConvertRasterToMesh.cpp
```diff
@@ -142,14 +147,24 @@ MeshLib::Mesh* ConvertRasterToMesh::constructMesh(const double* pix_vals, const
...
 		}
 	}
 	delete [] node_idx_map;
 
-	return new MeshLib::Mesh("RasterDataMesh", nodes, elements); // the name is only a temp-name, the name given in the dialog is set later
+	// the name is only a temp-name, the name given in the dialog is set later
+	MeshLib::Mesh* mesh(new MeshLib::Mesh("RasterDataMesh", nodes, elements));
+	auto opt_pv = mesh->getProperties().createNewPropertyVector<int>(
+		"MaterialIDs", MeshLib::MeshItemType::Cell);
+	if (opt_pv) {
```
endJunction: Would abort if the material ids property vector could not be created, otherwise the returned mesh will sometimes have material ids and sometimes not.

MeshLib/MeshGenerators/LayeredMeshGenerator.cpp
```diff
@@ -17,6 +17,8 @@
 #include <vector>
 #include <fstream>
 
+#include "logog/include/logog.hpp"
```
endJunction: system include <>

MeshLib/MeshGenerators/LayeredMeshGenerator.cpp
```diff
@@ -56,7 +61,20 @@ MeshLib::Mesh* LayeredMeshGenerator::getMesh(std::string const& mesh_name) const
     if (_nodes.empty() || _elements.empty())
         return nullptr;
 
-    MeshLib::Mesh* result (new MeshLib::Mesh(mesh_name, _nodes, _elements));
+	MeshLib::Properties properties;
+	if (!_materials.empty())
```
endJunction: `_materials` being empty is covered by the next test of `_materials`' sizes being equal to `_elements`' sizes.

MeshLib/MeshGenerators/LayeredMeshGenerator.cpp
```diff
@@ -56,7 +61,20 @@ MeshLib::Mesh* LayeredMeshGenerator::getMesh(std::string const& mesh_name) const
...
+		{
+			boost::optional<MeshLib::PropertyVector<int> &> materials = 
+				properties.createNewPropertyVector<int>("MaterialIDs", MeshLib::MeshItemType::Cell);
+			materials->resize(_materials.size());
+			std::copy(_materials.cbegin(), _materials.cend(), materials->begin());
+		}
+		else
+			ERR ("Skipping MaterialID information, number of entries does not match element number");
+	}
+    MeshLib::Mesh* result (new MeshLib::Mesh(mesh_name, _nodes, _elements, properties));
```
endJunction: Better to use `unique_ptr` to indicate `result`'s owneship.

MeshLib/MeshGenerators/LayeredMeshGenerator.cpp
```diff
@@ -56,7 +61,20 @@ MeshLib::Mesh* LayeredMeshGenerator::getMesh(std::string const& mesh_name) const
...
+	{
+		if (_materials.size() == _elements.size())
+		{
+			boost::optional<MeshLib::PropertyVector<int> &> materials = 
+				properties.createNewPropertyVector<int>("MaterialIDs", MeshLib::MeshItemType::Cell);
+			materials->resize(_materials.size());
+			std::copy(_materials.cbegin(), _materials.cend(), materials->begin());
+		}
+		else
+			ERR ("Skipping MaterialID information, number of entries does not match element number");
```
endJunction: Would suggest to abort on any errors, but if not an error, just emit a warning...

MeshLib/MeshGenerators/LayeredVolume.cpp
```diff
@@ -58,6 +59,7 @@ bool LayeredVolume::createRasterLayers(const MeshLib::Mesh &mesh,
 	this->_minimum_thickness = minimum_thickness;
 	_nodes = MeshLib::copyNodeVector(bottom->getNodes());
 	_elements = MeshLib::copyElementVector(bottom->getElements(), _nodes);
+	_materials.insert(_materials.begin(), _elements.size(), 0);
```
endJunction: The `_materials` size after this line will grow by number of stored `_elements`; is it correct?
Especially the elements in materials vector will be shifted.

MeshLib/MeshGenerators/LayeredVolume.cpp
```diff
@@ -159,14 +166,17 @@ void LayeredVolume::removeCongruentElements(std::size_t nLayers, std::size_t nEl
 			{
 				delete _elements[upper_offset+j];
 				_elements[upper_offset+j] = nullptr;
+				_materials[upper_offset+j] = -1;
```
endJunction: The special value `-1` (and also the above inserted `0` entries) are not documented.
There is also some `noDataReplacementValue`, which might be appropriate instead of the `-1`, but I'm not sure here.

MeshLib/MeshGenerators/MeshLayerMapper.cpp
```diff
@@ -60,6 +61,10 @@ MeshLib::Mesh* MeshLayerMapper::createStaticLayers(MeshLib::Mesh const& mesh, st
 	std::vector<MeshLib::Node*> new_nodes(nNodes + (nLayers * nNodes));
 	std::vector<MeshLib::Element*> new_elems;
 	new_elems.reserve(nElems * nLayers);
+	MeshLib::Properties properties;
+	boost::optional<PropertyVector<int> &> materials = 
+		properties.createNewPropertyVector<int>("MaterialIDs", MeshLib::MeshItemType::Cell);
+	materials->reserve(nElems * nLayers);
```
endJunction: `materials` could be empty.

MeshLib/MeshGenerators/VtkMeshConverter.cpp
```diff
@@ -265,8 +265,8 @@ MeshLib::Mesh* VtkMeshConverter::constructMesh(const double* pixVal,
 	if (value_vec->empty())
 		properties.removePropertyVector("Colour");
 
-	boost::optional< MeshLib::PropertyVector<unsigned>& > materials =
-		properties.createNewPropertyVector<unsigned>("MaterialIDs", MeshLib::MeshItemType::Cell, 1);
+	boost::optional< MeshLib::PropertyVector<int>& > materials =
+		properties.createNewPropertyVector<int>("MaterialIDs", MeshLib::MeshItemType::Cell, 1);
 	materials->insert(materials->end(), elements.size(), 0);
```
endJunction: `materials` could be empty...

MeshLib/MeshInformation.cpp
```diff
@@ -19,15 +19,16 @@
...
-        [](MeshLib::Element const*const a, MeshLib::Element const*const b)
-            {
-                return a->getValue() < b->getValue();
-        });
-	return std::make_pair<unsigned, unsigned>((*minmax.first)->getValue(), (*minmax.second)->getValue());
+	boost::optional<MeshLib::PropertyVector<int> const&> materialIds
+		= mesh.getProperties().getPropertyVector<int>("MaterialIDs");
+	if (materialIds) {
+		auto mat_bounds = std::minmax_element(materialIds->cbegin(), materialIds->cend());
+		return std::make_pair(*(mat_bounds.first), *(mat_bounds.second));
```
endJunction: If `*materialIds` are empty the `mat_bounds` members can not be dereferenced, which usuall ends in SEGV.

MeshLib/MeshInformation.cpp
```diff
@@ -19,15 +19,16 @@
...
+	}
+	else
+		return std::make_pair(-1,-1);
```
endJunction: (-1, -1) pair is not distinguishable from valid bounds of materials = {-1, -1, ..., -1};

Tests/MeshLib/TestElementStatus.cpp
```diff
@@ -27,18 +27,24 @@ TEST(MeshLib, ElementStatus)
...
 	const unsigned elements_per_side (20);
 	auto const mesh = std::unique_ptr<MeshLib::Mesh>{
-        MeshLib::MeshGenerator::generateRegularQuadMesh(width, elements_per_side)};
+		MeshLib::MeshGenerator::generateRegularQuadMesh(width, elements_per_side)};
+
+	boost::optional<MeshLib::PropertyVector<int> &> material_id_properties(
+		mesh->getProperties().createNewPropertyVector<int>("MaterialIDs",
+		                                                   MeshLib::MeshItemType::Cell)
+	);
+	(*material_id_properties).resize(mesh->getNElements());
```
endJunction: `ASSERT(material_id_properties)` before access.

MeshLib/MeshGenerators/LayeredMeshGenerator.cpp
rinkk: is a return value, leaving as is

MeshLib/MeshGenerators/LayeredVolume.cpp
rinkk: don't understand what you mean. the existing elements of the 2d mesh are carried over as "0th" layer of the 3d mesh. material properties reflect this by assigned "0" as material id for said elements.

MeshLib/MeshGenerators/LayeredVolume.cpp
rinkk: if the upper and lower facet of the element collapse, element + mat entry are marked for deletion. added a comment. this has no relevance for dealing with no data values.

MeshLib/MeshGenerators/MeshLayerMapper.cpp
rinkk: no ... materials can only be empty if elements are empty, in which case the mesh cannot be read in the first place

MeshLib/MeshInformation.cpp
rinkk: we had this test in here and Tom said to remove it again because empty arrays/correct array size should only be tested in interfaces

MeshLib/MeshInformation.cpp
rinkk: yes, so?

rinkk: i misunderstood your comment ... changing undefined bounds to numeric_limits::max

MeshLib/MeshGenerators/ConvertRasterToMesh.cpp
rinkk: rewrote this part. the check is unneccessary as the mesh is created right there.

MeshLib/MeshGenerators/LayeredVolume.cpp
endJunction: I think that `insert` is the wrong function here.
Given a vector `a = {0,1,2,3}`; After `a.insert(a.begin(), 3, -1)` a is equal to {-1,-1,-1,0,1,2,3}. There are other functions to _fill_ a vector or to _resize_ a vector.

rinkk: I have an empty vector and, say, 100 elements. What is the difference between inserting 100 x "0" and filling it with 100 x "0" ?

MeshLib/MeshGenerators/MeshLayerMapper.cpp
endJunction: I do not mean the product of `nElems` and `nLayers` being 0; The `materials` variable of type `boost::optional<...>` can be empty. This must be checked before using pointer access for reserve call.

rinkk: The ```materials``` variable is definitely empty at that point because it's created just one line above. It's *supposed* to be empty ... I am just creating the mesh...

MeshLib/MeshGenerators/LayeredVolume.cpp
endJunction: Semantic difference. For someone who does not know that the vector is empty there is a big difference.
Please document and add an assertion that the `_materials` are empty; or use `resize`.

MeshLib/MeshGenerators/MeshLayerMapper.cpp
endJunction: If the `materials` after the `createNewPropertyVector` call has an uninitialized optional object, than `*materials` contains no valid value and a nullptr in best case. Calling `reserve` on an arbitrary pointer usually leads to a segv.
There must be a check
```c++
if (!materials)
{
    ERR("Could not create bla bla");
    std::abort();
}
```
then one is free to call `materials->reserve(...)`.

FileIO/TetGenInterface.cpp
```diff
@@ -230,8 +228,18 @@ MeshLib::Mesh* TetGenInterface::readTetGenMesh (std::string const& nodes_fname,
 		return nullptr;
 	}
 
+	MeshLib::Properties properties;
+	// remove material vector again if all mat-ids are "0"
```
endJunction: The comment is different from the code; in the code material vector is removed if any of material ids is 0.

FileIO/TetGenInterface.cpp
```diff
@@ -230,8 +228,18 @@ MeshLib::Mesh* TetGenInterface::readTetGenMesh (std::string const& nodes_fname,
...
+	if (std::any_of(materials.cbegin(), materials.cend(), [](int m){ return m != 0; }))
+	{
+		boost::optional<MeshLib::PropertyVector<int> &> mat_props = 
+			properties.createNewPropertyVector<int>("MaterialIDs", MeshLib::MeshItemType::Cell);
+		mat_props->resize(elements.size());
```
endJunction: `mat_props` could contain an uninitialized optional object but is accessed without a check before.

FileIO/TetGenInterface.cpp
rinkk: ???? the material vector is created if any of the materials is not equal 0 ... otherwise (if all mat-ids *are* 0), the mat vec is removed again (or rather not copied into properties in the first place)

MeshLib/MeshGenerators/MeshLayerMapper.cpp
rinkk: I just talked to Lars and we both dont get it .... I create a properties object and immediatly create the mat-vector. how can the vector be uninitalised? judging from the implementation of ```createPropertyVector``` this is only possible if there is already a vector of the same name, which cannot happen here because the property object has been created just for the purpose of creating the mat vector.

FileIO/TetGenInterface.cpp
rinkk: same as in the other comment ... properties has just been created specifically for returning the mat vector....

MeshLib/MeshGenerators/MeshLayerMapper.cpp
endJunction: There is a simple pattern: if the returned value can have two states, initialized and uninitialized, it is good programming practice to check for uninitialized state.
I do not judge the implementation of `createPropertyVector` here. The changed code must be readable and unambiguous; I'm not going to find out under which circumstances the `materials` object is initialized or not. There are two possible states; the code must take care of both possibilities.

There could be a programming error in `createPropertyVector` or in any function called there in. There could be a system error. If there is a goal to write safe code then we must take care of both possible states. Otherwise we are quickly at the point where ogs5 is now.

Applications/DataExplorer/DataView/MeshElementRemovalDialog.cpp
TomFischer: Fixed.

MeshLib/MeshGenerators/VtkMeshConverter.cpp
TomFischer: Yes, `materials` should be empty. In my opninion the values are inserted into the `std::vector` correctly. I added a check if `createNewPropertyVector` creates a valid object.

MeshLib/MeshGenerators/LayeredVolume.cpp
TomFischer: Test if the vector is empty and resize the vector afterwards.

bilke:
:beers:

PR #899, opened 2015-12-01
Rework reset properties in polygonal region

TomFischer:
- Tool can reset different property data types (int, boolean and character).
- Updated [documentation](http://docs.opengeosys.org/docs/tools/meshing/set-properties-in-polygonal-region)


endJunction:
Looks OK.
From clang warnings:
 - Add static qualifier to `markNodesOutSideOfPolygon` function declaration (missing prototype warning)
 - ResetPropertiesInPolygonalRegion.cpp:112:32: warning: implicit conversion loses integer precision: 'std::size_t' (aka 'unsigned long') to 'unsigned int' [-Wshorten-64-to-32]
```
if (! outside[elem->getNode(k)->getID()]) {
              ~~~~          ^
```

bilke:
:+1:

PR #902, opened 2015-12-02
Updated CLion project files

bilke:
Especially `SMART_TABS` is important because this is for "indent with tabs, align with spaces".


wenqing:
:+1:

PR #907, opened 2015-12-09
Fixes for AlgorithmProperties and get- / setValue()-removal; FileIO

endJunction:
Extracted from https://github.com/ufz/ogs/pull/897

TODO:
 - [x] Separate `void setMaterialID(...)`

### DIFF
--- a/FileIO/FEFLOWInterface.cpp
+++ b/FileIO/FEFLOWInterface.cpp
@@ -156,8 +156,29 @@ MeshLib::Mesh* FEFLOWInterface::readFEFLOWFile(const std::string &filename)
 	}
 	in.close();
 
-	INFO("setting material IDs");
-	setMaterialID(fem_class, fem_dim, lines, vec_elementsets, vec_elements);
+	std::string project_name(
+		BaseLib::extractBaseNameWithoutExtension(filename));
+	if (_geoObjects && points)
+		_geoObjects->addPointVec(
+			std::unique_ptr<std::vector<GeoLib::Point*>>(points), project_name);
+	if (_geoObjects && lines)
+		_geoObjects->addPolylineVec(
+			std::unique_ptr<std::vector<GeoLib::Polyline*>>(lines),
+			project_name);
+
+	INFO("Create mesh");
+	auto mesh(std::unique_ptr<MeshLib::Mesh>(
+		new MeshLib::Mesh(project_name, vec_nodes, vec_elements)));
+	INFO("Set values for material property.");
+	auto opt_material_ids(mesh->getProperties().createNewPropertyVector<int>(
+			"MaterialIDs", MeshLib::MeshItemType::Cell, 1)
+	);
+	if (!opt_material_ids) {
+		WARN("Could not create PropertyVector for MaterialIDs in Mesh.");
+	} else {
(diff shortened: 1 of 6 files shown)


FileIO/FEFLOWInterface.cpp
```diff
@@ -156,8 +156,69 @@ MeshLib::Mesh* FEFLOWInterface::readFEFLOWFile(const std::string &filename)
...
+		_geoObjects->addPointVec(
+			std::unique_ptr<std::vector<GeoLib::Point*>>(points), project_name);
+	if (_geoObjects && lines)
+		_geoObjects->addPolylineVec(
+			std::unique_ptr<std::vector<GeoLib::Polyline*>>(lines),
+			project_name);
+
 	INFO("setting material IDs");
-	setMaterialID(fem_class, fem_dim, lines, vec_elementsets, vec_elements);
+	auto mesh = new MeshLib::Mesh(project_name, vec_nodes, vec_elements);
```
TomFischer: Could be a `unique_ptr`?!

FileIO/GMSInterface.cpp
```diff
@@ -314,10 +319,23 @@ MeshLib::Mesh* GMSInterface::readGMS3DMMesh(const std::string &filename)
...
 
 	const std::string mesh_name = BaseLib::extractBaseNameWithoutExtension(filename);
-	return new MeshLib::Mesh(mesh_name, nodes, elements);
+	MeshLib::Properties properties;
+	if (mat_ids.size() == elements.size())
+	{
+		boost::optional<MeshLib::PropertyVector<int> &> opt_pv
+			= properties.createNewPropertyVector<int>(
+			"MaterialIDs", MeshLib::MeshItemType::Cell);
+		assert(opt_pv != boost::none);
```
TomFischer: `assert` is only active in debug mode. Should we check this in release mode, too?

rinkk: I don't think using assert is a good idea. If the programme is closed when I am testing complex workflow this is going to be really annoying. Even if it's only in Debug-mode.

rinkk: extended comment to the last one: I still think assert is not a good idea here PLUS tom is right and a test + error message should be given in release mode. the programme should definitely not close but mat-ids should just be ignored.

endJunction: @rinkk What is the desired behavior if the `mat_ids` were read correctly (eg `mat_ids.size == elemenst.size`) but can not be copied to `opt_pv` because its creation failed?

rinkk:
> @rinkk What is the desired behavior if the mat_ids were read correctly (eg mat_ids.size == elemenst.size) but can not be copied to opt_pv because its creation failed?
The best solution would be to return ```nullptr```, alternatively we could return the mesh without mat ids.

endJunction:
Returning `nullptr`. :+1: 
:fast_forward:

TomFischer:
The changes concerning `setMaterialIDs()` are not included yet.

endJunction:
@TomFischer Took your changes for `setMaterialIds()` in feflow interface. I think this one is good to be merged, then please check if something was forgotten from the "main" `algprops` branch.

bilke:
:+1:

PR #910, opened 2015-12-10
Test for Neumann BC under PETSc and unified known solution applying interface.

wenqing:
1. Unified known solution applying interface for Dirichlet BC.
2. Corrected index type in all that is used in the equation assembly.
3. Added a Neumann type test for FEM#PETSc.


MathLib/LinAlg/Eigen/EigenTools.h
```diff
@@ -11,12 +11,14 @@
...
 
 #include <vector>
+#include <Eigen/Sparse> //for RawMatrixType::Index
 
 namespace MathLib
 {
 class EigenMatrix;
 class EigenVector;
 
+using RawMatrixType = Eigen::SparseMatrix<double, Eigen::RowMajor>;
```
chleh: Isn't that already defined inside the `EigenMatrix` class? I would prefer to reuse that definition.

MathLib/LinAlg/Dense/DenseTools.h
```diff
@@ -16,12 +16,11 @@
...
 
 #include <vector>
-#include "DenseMatrix.h"
-#include "DenseVector.h"
+#include "GlobalDenseMatrix.h"
 
 namespace MathLib
 {
-
+using DenseMatrixIndexType = GlobalDenseMatrix<double>::IndexType;
```
chleh: I would prefer to drop that typedef.

MathLib/LinAlg/Dense/DenseTools.cpp
```diff
@@ -17,8 +17,8 @@
 namespace MathLib
 {
 
-void applyKnownSolution(DenseMatrix<double> &eqsA, DenseVector<double> &eqsRHS, std::size_t k,
-		double val)
+void applyKnownSolution(DenseMatrix<double> &eqsA, DenseVector<double> &eqsRHS,
+		DenseMatrixIndexType k, double val)
```
chleh: Strictly speaking, this function deals with type `DenseMatrix`, so the index type should be `DenseMatrix<double>::IdxType` and not a typedef from `GlobalDenseMatrix`.

MathLib/LinAlg/Lis/LisTools.h
```diff
@@ -34,8 +35,8 @@ class LisVector;
  * @param rows a vector of known solution entry IDs
  * @param vals a vector of known solutions
  */
-void applyKnownSolution(LisMatrix &eqsA, LisVector &eqsRHS,
-	const std::vector<std::size_t> &rows,
+void applyKnownSolution(LisMatrix &eqsA, LisVector &eqsRHS, LisVector &/*eqsX*/,
+	const std::vector< LIS_INT> &rows, //LIS_INT has an alias of LisMatrix::IndexType
```
chleh: Completely consistent with LisTools/EigenTools would be to use the mentioned alias and not LIS_INT.

ProcessLib/GroundwaterFlowProcess.h
```diff
@@ -289,33 +289,12 @@ class GroundwaterFlowProcess : public Process
         // Call global assembler for each local assembly item.
         _global_setup.execute(*_global_assembler, _local_assemblers);
 
-#ifdef USE_PETSC
```
chleh: Great improvement that now we don't need to make that distinction anymore!

ProcessLib/UniformDirichletBoundaryCondition.h
```diff
@@ -67,16 +69,13 @@ class UniformDirichletBoundaryCondition
                                 MeshLib::MeshItemType::Node,
                                 id);
             // TODO: that might be slow, but only done once
-            id = dof_table.getGlobalIndex(l, component_id);
+            const auto g_idx = dof_table.getGlobalIndex(l, component_id);
+            if (g_idx >= 0)
+            {
+                global_ids.emplace_back(g_idx);
```
chleh: Maybe it's worth to call `reserve()` on global_ids before the loop.

ProcessLib/UniformDirichletBoundaryCondition.h
```diff
@@ -67,16 +69,13 @@ class UniformDirichletBoundaryCondition
                                 MeshLib::MeshItemType::Node,
                                 id);
             // TODO: that might be slow, but only done once
-            id = dof_table.getGlobalIndex(l, component_id);
+            const auto g_idx = dof_table.getGlobalIndex(l, component_id);
+            if (g_idx >= 0)
```
chleh: Maybe it there should be a comment that this is Petsc-specific behaviour?

Tests/AssemblerLib/SteadyDiffusion2DExample1.h
```diff
@@ -140,7 +140,7 @@ struct SteadyDiffusion2DExample1
 	static const std::size_t mesh_stride = mesh_subdivs + 1;
 	static const std::size_t dim_eqs = mesh_stride * mesh_stride;
 	MeshLib::Mesh* msh;
-	std::vector<std::size_t> vec_DirichletBC_id;
+	std::vector<int> vec_DirichletBC_id;
```
chleh: Should it be some `IndexType`? But in my opinion, it's a test. As long as it works, it's fine. If will fail due to the index type, we can easily fix it later on.

wenqing:
@endJunction  Compilation errors under LIS have been removed.

Tests/MathLib/TestLinearSolver.cpp
```diff
@@ -207,10 +207,12 @@ TEST(MathLib, CheckInterface_GaussAlgorithm)
     boost::property_tree::ptree t_solver;
     t_root.put_child("ogs", t_solver);
 
+    using Example = Example1<std::size_t>; 
+
     typedef MathLib::GaussAlgorithm<MathLib::GlobalDenseMatrix<double>, MathLib::DenseVector<double> > LinearSolverType;
-    MathLib::GlobalDenseMatrix<double> A(Example1::dim_eqs, Example1::dim_eqs);
+    MathLib::GlobalDenseMatrix<double> A(Example::dim_eqs, Example::dim_eqs);
```
chleh: Why the change `Example1` → `Example`? I don't see it immediately.

chleh:
I only saw small issues that are very easy to address.

MathLib/LinAlg/Eigen/EigenTools.h
wenqing: Just want to take the advantage of the forward class declaration, and to avoid include the header of the matrix classes that have the definition of IndexType.

MathLib/LinAlg/Dense/DenseTools.h
wenqing: Same reason as before.

MathLib/LinAlg/Dense/DenseTools.cpp
wenqing: Will be changed.

MathLib/LinAlg/Lis/LisTools.h
wenqing: Same reason as above: just to avoid include LisMatrix.h

ProcessLib/UniformDirichletBoundaryCondition.h
wenqing: The thing is that we have to count how many non-ghost entries in advance in order to call reserve().

ProcessLib/UniformDirichletBoundaryCondition.h
wenqing: Will add a comment.

ProcessLib/UniformDirichletBoundaryCondition.h
chleh: Possibly it would be slightly more performant to reserve space for all entries of `ids`. Sure, that would waste some space, but memory consumption won't blow up by that.
But I have not measured the performace impact and this function is only called once per simulation (or in future maby once per timestep).
So in my opinion: It would be nice if you kept the `global_ids.reserve(global_ids.size() + ids.size());`, at least if you do not strongly object against the waste of space.

wenqing: Take your idea.

ProcessLib/UniformDirichletBoundaryCondition.h
wenqing: Added.

Tests/AssemblerLib/SteadyDiffusion2DExample1.h
wenqing: Changed by turning class SteadyDiffusion2DExample1.h into a template class.

Tests/MathLib/TestLinearSolver.cpp
wenqing: Above line 210, there is a newly introduced alias, Example, to ``Example1<std::size_t>`` as
```
using Example = Example1<std::size_t>; 
```

wenqing:
@chleh  Your comments have been addressed (See commit d487ad0).

Tests/MathLib/TestLinearSolver.cpp
chleh: Thanks.

ProcessLib/UniformDirichletBoundaryCondition.h
```diff
@@ -61,22 +63,27 @@ class UniformDirichletBoundaryCondition
...
+            const auto g_idx = dof_table.getGlobalIndex(l, component_id);
+            // For the DDC approach (e.g. with PETSc option), the negative
+            // index of g_idx means that the entry by that index is a ghost one,
+            // which should be dropped. Especially for PETSc routines MatZeroRows
+            // and MatZeroRowsColumns, which are called to apply the Dirichlet BC,
+            // the negative index is not accepted like other matrix or vector
+            // PETSc routines. Therefore, the following if-condition is applied.  
+            if (g_idx >= 0)
+            {
+                global_ids.push_back(g_idx);
```
chleh: Just a short note: You could have kept it as emplace_back.

chleh:
:+1:

ProcessLib/UniformDirichletBoundaryCondition.h
wenqing: Reverted to  emplace_back by amending (see the amended commit 68f93c1).

MathLib/LinAlg/Dense/DenseTools.cpp
wenqing: Changed.

SimpleTests/MeshTests/MPI/CMakeLists.txt
```diff
@@ -43,10 +44,16 @@ ExternalData_Add_Test(
...
-set(FilePath "DATA{${ExternalData_SOURCE_ROOT}/EllipticPETSc/,REGEX:.*}")
 set(MPITestParameters -np 3 "${PROJECT_BINARY_DIR}/bin/ogs" "${FilePath}/cube_1e3.prj" --  -gw_ksp_type bcgs -gw_pc_type bjacobi -gw_ksp_atol 1.e-10)
 ExternalData_Add_Test(
     data
-    NAME ParallelFEM_GroundWaterFlow3D
+    NAME ParallelFEM_GroundWaterFlow3D_DirichletBC
+    COMMAND  "mpirun" ${MPITestParameters} ${SolverType}
+)
+
+set(MPITestParameters -np 3 "${PROJECT_BINARY_DIR}/bin/ogs" "${FilePath}/cube_1e3_neumann.prj" --  -gw_ksp_type bcgs -gw_pc_type bjacobi -gw_ksp_atol 1.e-10)
```
endJunction: For another PR I suggest to put the linear solver options into the project file; this way the simulation will run independent of the way it was started---by `ctest` or manually from prompt.

PR #911, opened 2015-12-11
Rework axis aligned bounding box

TomFischer:
Until now, it was always necessary to enlarge the axis aligned bounding box (AABB) to assure that all points of the point set that defined the AABB are located inside this AABB. AABB should implement a from the right half-open interval. Unfortunately the implementation was not consistent.

The PR changes this behaviour. The AABB will enlarged a very small amount (unit in the last place) by default. The method `containsPoint()` is implemented such that "min <= p < max".

The doxygen documentation for the class template is not generated. If somebody has a hint ...

ToDo
 - [x] fix documentation

### DIFF
--- a/GeoLib/AABB.h
+++ b/GeoLib/AABB.h
@@ -108,9 +108,9 @@ public:
 	template <typename T>
 	bool containsPoint(T const & pnt) const
 	{
-		if (pnt[0] < _min_pnt[0] || _max_pnt[0] < pnt[0]) return false;
-		if (pnt[1] < _min_pnt[1] || _max_pnt[1] < pnt[1]) return false;
-		if (pnt[2] < _min_pnt[2] || _max_pnt[2] < pnt[2]) return false;
+		if (pnt[0] < _min_pnt[0] || _max_pnt[0] <= pnt[0]) return false;
+		if (pnt[1] < _min_pnt[1] || _max_pnt[1] <= pnt[1]) return false;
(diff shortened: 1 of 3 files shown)


GeoLib/AABB.h
```diff
@@ -160,9 +181,33 @@ class AABB
 	{
 		init(*pnt);
 	}
+
+	/// Updates the bounding box using point \f$p\f$ without resizing it.
+	std::pair<bool,std::bitset<3>> update_(PNT_TYPE const & p)
```
endJunction: Would you please elaborate more on what update without resizing means?
Maybe return value explanation would already help.

GeoLib/AABB.h
```diff
@@ -150,6 +160,17 @@ class AABB
 		std::numeric_limits<double>::lowest(),
 		std::numeric_limits<double>::lowest()}}};
 private:
+	/// Enlarge the bounding box the smallest possible amount.
+	void enlarge(std::bitset<3> to_update = 7)
```
endJunction: Please add a note that only right boundary is moved.

endJunction:
What is the difference btw. `update` and `update_`? Maybe there is a better name?

TomFischer:
Renamed `update_()` to `updateWithoutEnlarge()`.

endJunction:
:fast_forward:

TomFischer:
All changes included in PR #929.

PR #912, opened 2015-12-12
ConfigTree with some checks added

chleh:
This PR introduces a wrapper class for the Boost Property Tree. The purpose of the wrapper is:
* to tell the user about settings that have not been read. The messages include some hints about where something went wrong.
* to enforce (to some extent) that each setting is parsed exactly once
* to make (by rather long method names) settings easily greppable from source code. I think this way it will be more easy later on to provide a comprehensive documentation of all possible OGS-6 input parameters.
* to enforce a certain style of XML tag names [a-z0-9_].
* to avoid or reduce the amount of input error handling code in OGS-6 routines.
* to enforce that a particular parameter is always read with the same type.

Those points are basically implemented by keeping track of how many times a routine asks for a certain parameter. The destructor then will check that everything has been read.

Currently the changes in this PR are somewhat scattered. i will clean it up later on. I believe that everything we need can be done with the new class and that a transition is easy.

The main part (interface) of this PR is in `BaseLib/ConfigTreeNew.h`. It is used pretty much in the same way as the Boost property tree (cf. `Tests/BaseLib/TestConfigTree.cpp`). Please have a look if any feature is missing or if you disagree with my implementation.

A typical waring message will look like this:
`ConfigTree: At path <OpenGeoSysProject/processes/process/linear_solver>: Key <eigen> has been read 1 time(s) less than it was present in the configuration tree.`
(of cause this particular message will not occur.)

In this PR the options of `ProcessVariable`s are parsed with the new `ConfigTree`. Once this PR is accepted I will make the transition to the new ConfigTree in all of OGS-6.

There is one design question to be discussed: Currently the new ConfigTree cannot be passed as `const &` but only as reference to a modifiable object. That can be changed if I make one member variable `mutable`. Should I do that?

### DIFF
--- a/BaseLib/ConfigTreeNew-impl.h
+++ b/BaseLib/ConfigTreeNew-impl.h
@@ -1,0 +1,159 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2015, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#include "ConfigTreeNew.h"
+
+#include <logog/include/logog.hpp>
+
+namespace BaseLib
+{
+
+//! Wraps a pair of iterators for use as a range in range-based for-loops.
+template<typename Iterator>
+class Range
+{
+public:
+    explicit Range(Iterator begin, Iterator end)
+        : _begin(begin), _end(end)
+    {}
+
+    Iterator begin() const { return _begin; }
+    Iterator end()   const { return _end; }
+private:
+    Iterator _begin;
+    Iterator _end;
+};
+
+template<typename T>
+T
+ConfigTreeNew::
+getConfParam(std::string const& param)
+{
+    auto p = getConfParamOptional<T>(param);
+    if (p) return *p;
+
+    error("Key <" + param + "> has not been found");
+    return T();
+}
+
+template<typename T>
+T
+ConfigTreeNew::
+getConfParam(std::string const& param, T const& default_value)
+{
+    auto p = getConfParamOptional<T>(param);
+    if (p) return *p;
+    return default_value;
+}
+
+template<typename T>
+boost::optional<T>
+ConfigTreeNew::
+getConfParamOptional(std::string const& param)
+{
+    checkUnique(param);
+    auto p = _tree->get_child_optional(param);
+
+    bool peek_only = p == boost::none;
+    markVisited<T>(param, peek_only);
+
+    if (p) {
+        auto v = p->get_value_optional<T>();
+        if (v) {
+            return v;
+        } else {
+            error("Value for key <" + param + "> `" + shortString(p->data())
+                  + "' not convertible to the desired type.");
+        }
+    }
+
+    return boost::none;
+}
+
+template<typename T>
+Range<ConfigTreeNew::ValueIterator<T> >
+ConfigTreeNew::
+getConfParamList(std::string const& param)
+{
+    checkUnique(param);
+    markVisited<T>(param, true);
+
+    auto p = _tree->equal_range(param);
+    return Range<ValueIterator<T> >(
+                ValueIterator<T>(p.first,  param, *this),
+                ValueIterator<T>(p.second, param, *this));
+}
+
+template<typename T>
+T
+ConfigTreeNew::
+peekConfParam(std::string const& param)
+{
+    checkKeyname(param);
+
+    auto p =_tree->get_child_optional(param);
+
+    if (!p) {
+        error("Key <" + param + "> has not been found");
+    } else {
+        try {
+            return p->get_value<T>();
+        } catch (boost::property_tree::ptree_bad_data) {
+            error("Value for key <" + param + "> `" + shortString(p->data())
+                  + "' not convertible to the desired type.");
+        }
+    }
+
+    return T();
+}
+
+template<typename T>
+void
+ConfigTreeNew::
+checkConfParam(std::string const& param, T const& value)
+{
+    if (getConfParam<T>(param) != value) {
+        error("The value of key <" + param + "> is not the expected one.");
+    }
+}
+
+template<typename Ch>
+void
+ConfigTreeNew::
+checkConfParam(std::string const& param, Ch const* value)
+{
+    if (getConfParam<std::string>(param) != value) {
+        error("The value of key <" + param + "> is not the expected one.");
+    }
+}
+
+
+template<typename T>
+ConfigTreeNew::CountType&
+ConfigTreeNew::
+markVisited(std::string const& key, bool peek_only)
+{
+    auto const type = std::type_index(typeid(T));
+
+    auto p = _visited_params.emplace(key, CountType{peek_only ? 0 : 1, type});
+
+    if (!p.second) { // no insertion happened
+        auto& v = p.first->second;
+        if (v.type == type) {
+            if (!peek_only) ++v.count;
+        } else {
+            error("There already was an attempt to obtain key <" + key
+                  + "> with type \"" + v.type.name() + "\" (now: \"" + type.name() + "\").");
+        }
+    }
+
+    return p.first->second;
+}
+
+}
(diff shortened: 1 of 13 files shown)


endJunction:
Looks OK. I think that to truly find out if the design is good an exemplary change to any of the current config parsers would be helpful. Maybe you could choose one particularly ugly example :)

BaseLib/ConfigTreeNew.h
```diff
@@ -0,0 +1,325 @@
...
+/*!
+ * Wrapper around a Boost Property Tree with some basic error reporting features.
+ *
+ * Features. This class:
+ *  * makes sure that every configuration setting in a Property Tree is read
+ *    exactly once. If some settings is not read (e.g. due to a typo), a warning message
+ *    is generated. The message contains a hint where it occured.
+ *  * enforces a naming scheme of settings: letters a-z, numbers 0-9, underscore
+ *  * provides some functionality to read lists of values using range-based for loops.
+ *  * has rater lng method names that are easily greppable from the source code. So a list
```
TomFischer: lng -> long

BaseLib/ConfigTreeNew.h
```diff
@@ -0,0 +1,325 @@
...
+
+    using PTree = boost::property_tree::ptree;
+
+    //! Creates a new instance wrapping the given Boost Property Tree.
+    explicit ConfigTreeNew(PTree const& tree);
+
+    /*! Get parameter \c param of type \c T from the configuration tree.
+     *
+     * \return the value looked for or a default constructed value \c T() in the case of
+     *         an error. For the behaviur in case of an error, see also the documentation
```
TomFischer: behaviur -> behaviour

BaseLib/ConfigTreeNew.h
```diff
@@ -0,0 +1,325 @@
...
+     *
+     * \pre \c param must not have been read before from this ConfigTree.
+     */
+    template<typename T> T
+    getConfParam(std::string const& param, T const& default_value);
+
+    /*! Get parameter \c param of type \c T from the configuration tree if present
+     *
+     * This method has a similar behaviour as getConfParam(std::string const&) except
+     * no errors are raised. Rather it can be told from thr return value if the
```
TomFischer: thr -> the

TomFischer:
Good job, go on!

chleh:
Telling from this thread: http://stackoverflow.com/questions/10783391/is-it-possible-to-move-a-boostoptional
My new class needs boost >= 1.56, because of move-construction and move-assignment for `boost::optional`. Is it possible that we require at least boost 1.56? Otherwise I will either have to come up with a different way of describing optional values (unique pointers?), or I will have to drop this PR.

TomFischer:
I haven't any objections to require a higher boost version.

endJunction:
Jenkins test this please.

It was just the DenseMatrix test again...

bilke:
Cool stuff! :+1:

BaseLib/ConfigTreeNew.cpp
```diff
@@ -0,0 +1,225 @@
...
+getConfSubtreeOptional(std::string const& root)
+{
+    checkUnique(root);
+    auto subtree = _tree->get_child_optional(root);
+
+    if (subtree) {
+        markVisited(root);
+        boost::optional<ConfigTreeNew> t;
+        t.emplace(*subtree, *this, root);
+        return std::move(t);
```
endJunction: BaseLib/ConfigTreeNew.cpp:99:16: warning: moving a local object in a return statement prevents copy elision
clang's suggestion:  note: remove std::move call here

@chleh could you explain why making `t` an rvalue is needed?

BaseLib/ConfigTreeNew.h
```diff
@@ -0,0 +1,392 @@
...
+    //! The first argument denotes the path in the tree at which an event (warning/error)
+    //! occured, the second argument is the associated message
+    using Callback = std::function<void(const std::string& path,
+                                        const std::string& message)>;
+
+    /*! Creates a new instance wrapping the given Boost Property Tree.
+     *
+     * \param tree the Boost Property Tree to be wrapped
+     * \param onerror callback function to be called on error.
+     * \param onwarning callback function to be called on warning.
```
endJunction: the onerror and onwarning are noew error_cb and warning_cb.
clang complains here:
```
/home/naumov/w/ogs/BaseLib/ConfigTreeNew.h:173:15: warning: parameter 'onerror' not found in the function declaration [-Wdocumentation]
     * \param onerror callback function to be called on error.
              ^~~~~~~
/home/naumov/w/ogs/BaseLib/ConfigTreeNew.h:174:15: warning: parameter 'onwarning' not found in the function declaration [-Wdocumentation]
     * \param onwarning callback function to be called on warning.
              ^~~~~~~~~
```

BaseLib/ConfigTreeNew.cpp
```diff
@@ -0,0 +1,225 @@
...
+{
+	_onwarning(_path, message);
+}
+
+
+void ConfigTreeNew::onerror(const std::string& path, const std::string& message)
+{
+    ERR("ConfigTree: At path <%s>: %s", path.c_str(), message.c_str());
+    std::abort();
+}
```
endJunction: Just for info:
the function could be declared with noreturn attribute, but the gcc/clang syntax for attributes is different from msvc's. Not sure what a good solution is.

BaseLib/ConfigTreeNew.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+ConfigTreeNew::
+ConfigTreeNew(PTree const& tree,
+              Callback const& error_cb,
+              Callback const& warning_cb)
+    : _tree(&tree), _onerror(error_cb), _onwarning(warning_cb)
+{
+    if (!_onerror) {
+        ERR("ConfigTree: No valid error handler provided.");
+        std::abort();
```
wenqing: Is it good to call abort inside the constructor?

BaseLib/ConfigTreeNew.cpp
chleh: No, I can't, will remove it.

BaseLib/ConfigTreeNew.cpp
chleh: If the check fails, the program __will__ segfault later on, so I think there is no additional problem with `std::abort()`.
Also, I think calling `std::abort()` in the constructor is as fatal as calling it outside: The program will terminate without cleanup anyway.
Maybe we could discuss, if it is a good idea to use abort() anywhere, because it does not clean up. And maybe in the future it would be nice to clean up, e.g., finish writing to some file or remove temporary files. But that discussion can take place another day.

BaseLib/ConfigTreeNew.h
chleh: I will change the names.

BaseLib/ConfigTreeNew.cpp
chleh: I think in this special case there would not be much benifit from it: onerror() is wrapped inside a std::function. Does std::function have special support for noreturn? So far I would prefer to not do the extra work.

BaseLib/ConfigTreeNew-impl.h
```diff
@@ -0,0 +1,159 @@
...
+    if (p) return *p;
+    return default_value;
+}
+
+template<typename T>
+boost::optional<T>
+ConfigTreeNew::
+getConfParamOptional(std::string const& param)
+{
+    checkUnique(param);
```
endJunction: Why are the other getters (`getConfParam` and its optional version) without this `checkUnique`?

chleh: The other methods call this one, so additional checks are not necessary.

endJunction:
I'd recommend to have consistent tabs/whitespaces at least in a single file. (Or use clang-format)

chleh:
Since this PR has already been merged, I will fix the indentation in a subsequent PR.

BaseLib/ConfigTreeNew.cpp
```diff
@@ -0,0 +1,224 @@
...
+boost::optional<ConfigTreeNew>
+ConfigTreeNew::
+getConfSubtreeOptional(std::string const& root)
+{
+    checkUnique(root);
+    auto subtree = _tree->get_child_optional(root);
+
+    if (subtree) {
+        markVisited(root);
+        return boost::optional<ConfigTreeNew>(std::move(
```
endJunction: Same as for the other move: copy-elision with `std::move` is not possible. I think the return statement could be just
```c++
return ConfigTreeNew(*subtree, *this, root);
```

PR #928, opened 2015-12-15
Element value

endJunction:
The remains of the original "algprops" PR https://github.com/ufz/ogs/pull/897.

The `value` is finally removed from the `Element` class.
A changelog entry would be good.

Closes https://github.com/ufz/ogs/pull/897


TomFischer:
I do not see where the constructor declarations in TemplateElement.h are removed.

endJunction:
@TomFischer Updated `TemplateElement.h`; see Lars' commit:  https://github.com/endJunction/ogs/commit/33abca59185d3b0d7591f5b9b92613c20fcc4634

bilke:
:+1:

endJunction:
@bilke Some tests are delayed (if Jenkins' queue is full for example) but the checks are already marked as done/green. Is it possible to improve this and leave the pending/yellow status until all *-PR builds are executed?

bilke:
No unfortunately that is not possible.

PR #929, opened 2015-12-16
Transformed class template AABB -> class AABB

TomFischer:
Removing the template stuff makes 
- AABB more flexible and
- easier to use.

This PR is based on the PR #911 and must not be merged before PR #911.

**Update 2016-01-05**
All changes from PR #911 are included now in the PR.

### DIFF
--- a/GeoLib/AABB.h
+++ b/GeoLib/AABB.h
@@ -108,9 +108,9 @@ public:
 	template <typename T>
 	bool containsPoint(T const & pnt) const
 	{
-		if (pnt[0] < _min_pnt[0] || _max_pnt[0] < pnt[0]) return false;
-		if (pnt[1] < _min_pnt[1] || _max_pnt[1] < pnt[1]) return false;
-		if (pnt[2] < _min_pnt[2] || _max_pnt[2] < pnt[2]) return false;
+		if (pnt[0] < _min_pnt[0] || _max_pnt[0] <= pnt[0]) return false;
+		if (pnt[1] < _min_pnt[1] || _max_pnt[1] <= pnt[1]) return false;
+		if (pnt[2] < _min_pnt[2] || _max_pnt[2] <= pnt[2]) return false;
 		return true;
 	}
 
--- a/GeoLib/AABB.h
+++ b/GeoLib/AABB.h
@@ -29,7 +29,6 @@
 
 #include "MathLib/Point3d.h"
 #include "MathLib/MathTools.h"
-#include "Point.h"
 
 namespace GeoLib
 {
@@ -50,7 +49,7 @@ namespace GeoLib
  * @tparam PNT_TYPE a point type supporting accessing the coordinates via
  * operator[]
  */
-template <typename PNT_TYPE = GeoLib::Point>
+template <typename PNT_TYPE>
 class AABB
 {
 public:
--- a/GeoLib/AABB.h
+++ b/GeoLib/AABB.h
@@ -46,16 +46,16 @@ namespace GeoLib
  * available floating point number such that all input points are contained in
  * the bounding box.
  *
- * @tparam PNT_TYPE a point type supporting accessing the coordinates via
- * operator[]
  */
-template <typename PNT_TYPE>
 class AABB
 {
 public:
 	/**
 	 * construction of object, initialization the axis aligned bounding box
+	 * @tparam PNT_TYPE a point type supporting accessing the coordinates via
+	 * operator[]
 	 * */
+	template <typename PNT_TYPE>
 	AABB(std::vector<PNT_TYPE*> const& pnts, std::vector<std::size_t> const& ids)
 	{
 		assert(! ids.empty());
@@ -69,8 +69,8 @@ public:
 	/**
 	 * copy constructor.
 	 * @param src an axis aligned bounding box
-	 */
-	AABB(AABB<PNT_TYPE> const& src) :
+	 * */
+	AABB(AABB const& src) :
 		_min_pnt(src._min_pnt), _max_pnt(src._max_pnt)
 	{}
 
@@ -102,6 +102,7 @@ public:
 
 	/// Checks if the bounding box has to be updated.
 	/// @return true if AABB is updated.
+	template <typename PNT_TYPE>
 	bool update(PNT_TYPE const & p)
 	{
 		// First component of the pair signals if the minimum point is changed
@@ -164,7 +165,7 @@ public:
 	 * @return true if the other AABB is contained in the AABB
 	 * represented by this object
 	 */
-	bool containsAABB(AABB<PNT_TYPE> const& other_aabb) const
+	bool containsAABB(AABB const& other_aabb) const
 	{
 		return containsPoint(other_aabb.getMinPoint()) && containsPoint(other_aabb.getMaxPoint());
 	}
@@ -192,13 +193,16 @@ private:
 		}
 	}
 
+	template <typename PNT_TYPE>
 	void init(PNT_TYPE const & pnt)
 	{
 		_min_pnt[0] = _max_pnt[0] = pnt[0];
 		_min_pnt[1] = _max_pnt[1] = pnt[1];
 		_min_pnt[2] = _max_pnt[2] = pnt[2];
 	}
-	void init(PNT_TYPE const * pnt)
+
+	template <typename PNT_TYPE>
+	void init(PNT_TYPE * const & pnt)
 	{
 		init(*pnt);
 	}
@@ -208,6 +212,7 @@ private:
 	/// box. Using this method the bounding box of the initial point set is
 	/// enlarged only once.
 	/// @param p point that will possibly change the bounding box points
+	template <typename PNT_TYPE>
 	void  updateWithoutEnlarge(PNT_TYPE const & p)
 	{
 		for (std::size_t k(0); k<3; k++) {
@@ -220,11 +225,13 @@ private:
 		}
 	}
 
-	void updateWithoutEnlarge(PNT_TYPE const * pnt)
+	template <typename PNT_TYPE>
+	void updateWithoutEnlarge(PNT_TYPE * const & pnt)
 	{
 		updateWithoutEnlarge(*pnt);
 	}
 
+	template <typename PNT_TYPE>
 	void update(PNT_TYPE const * pnt)
 	{
 		update(*pnt);
--- a/GeoLib/AABB.h
+++ b/GeoLib/AABB.h
@@ -34,12 +34,22 @@
 namespace GeoLib
 {
 /**
- *
- * \ingroup GeoLib
- *
  * \brief Class AABB is an axis aligned bounding box around a given
  * set of geometric points of (template) type PNT_TYPE.
- * */
+ *
+ * Let \f$P = \{p_k \in \mathbb{R}^3, \ k=1, \dotsc, n\}\f$ a set of 3d points.
+ * The bounding volume is described by its lower, left, front point \f$\ell\f$
+ * and the upper, right, back point \f$u\f$, i.e. the coordinates of \f$\ell\f$
+ * and \f$u\f$ are computed as follows \f$\ell_i = \min \limits_{p \in P}
+ * p_i\f$, \f$u_i = \max \limits_{p \in P} p_i\f$, respectively. The bounding
+ * box consists of half-open intervals \f$[\ell_x, u_x) \times [\ell_y, u_y)
+ * \times [\ell_z, u_z).\f$ The bounding box is enlarged up to the next
+ * available floating point number such that all input points are contained in
+ * the bounding box.
+ *
+ * @tparam PNT_TYPE a point type supporting accessing the coordinates via
+ * operator[]
+ */
 template <typename PNT_TYPE = GeoLib::Point>
 class AABB
 {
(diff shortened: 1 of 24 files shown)


GeoLib/AABB.h
```diff
@@ -81,25 +94,40 @@ class AABB
...
+		enlarge();
 	}
 
-	bool update(PNT_TYPE const & pnt)
+	template <typename PNT_TYPE>
+	bool update(PNT_TYPE const & p)
 	{
-		bool updated(false);
+		// First component of the pair signals if the minimum point is changed
+		// Second component signals not only if the max point is changed.
```
chleh: "signals not only if"?

GeoLib/AABB.h
```diff
@@ -81,25 +94,40 @@ class AABB
...
+		// Furthermore it is signaled what coordinate (0,1,2) is changed.
+		std::pair<bool,std::bitset<3>> updated(0,0);
 		for (std::size_t k(0); k<3; k++) {
-			if (pnt[k] < _min_pnt[k]) {
-				_min_pnt[k] = pnt[k];
-				updated = true;
+			// if the minimum point is updated pair.first==true
+			if (p[k] < _min_pnt[k]) {
+				_min_pnt[k] = p[k];
+				updated.first = true;
```
chleh: did you forget `[k]`?

GeoLib/AABB.h
```diff
@@ -81,25 +94,40 @@ class AABB
...
-			update(*it);
+			updateWithoutEnlarge(*it);
 			it++;
 		}
+		enlarge();
 	}
 
-	bool update(PNT_TYPE const & pnt)
+	template <typename PNT_TYPE>
+	bool update(PNT_TYPE const & p)
```
chleh: Can you add a comment what the return value says?

GeoLib/AABB.h
```diff
@@ -150,19 +178,60 @@ class AABB
...
 	void init(PNT_TYPE const & pnt)
 	{
 		_min_pnt[0] = _max_pnt[0] = pnt[0];
 		_min_pnt[1] = _max_pnt[1] = pnt[1];
 		_min_pnt[2] = _max_pnt[2] = pnt[2];
 	}
-	void init(PNT_TYPE const * pnt)
+
+	template <typename PNT_TYPE>
+	void init(PNT_TYPE * const & pnt)
```
chleh: Is the pointer-version needed?

GeoLib/AABB.h
```diff
@@ -150,19 +178,60 @@ class AABB
...
+	void updateWithoutEnlarge(PNT_TYPE * const & pnt)
+	{
+		updateWithoutEnlarge(*pnt);
+	}
+
+	template <typename PNT_TYPE>
 	void update(PNT_TYPE const * pnt)
 	{
-		update (*pnt);
+		update(*pnt);
```
chleh: Is the pointer version needed?

GeoLib/AABB.h
```diff
@@ -150,19 +178,60 @@ class AABB
...
+				_min_pnt[k] = p[k];
+			}
+			if (p[k] >= _max_pnt[k]) {
+				_max_pnt[k] = p[k];
+			}
+		}
+	}
+
+	template <typename PNT_TYPE>
+	void updateWithoutEnlarge(PNT_TYPE * const & pnt)
```
chleh: Is the pointer version needed?

MeshLib/CoordinateSystem.cpp
```diff
@@ -31,4 +30,23 @@ CoordinateSystem::CoordinateSystem(const Element &ele)
...
+unsigned char CoordinateSystem::getCoordinateSystem(const GeoLib::AABB &bbox) const
+{
+    unsigned char coords = 0;
+
+    const MathLib::Vector3 pt_diff(bbox.getMinPoint(), bbox.getMaxPoint());
+
+    // The axis aligned bounding box is a from the right half-open interval.
+    // Therefore, the difference between the particular coordinates of the
+    // points is modified by the unit in the last place towards zero.
+    if (std::nexttoward(std::abs(pt_diff[0]), 0.0) > .0)
```
chleh: Are there any implications from those tests, e.g., if something associated with a 3D model (e.g. a surface) happens to lie in a coordinate plane (e.g. x-y plane)? (I don't know, where this method is used, so maybe it's never will be a problem.)

Tests/GeoLib/TestAABB.cpp
```diff
@@ -220,3 +238,42 @@ TEST(GeoLib, AABBAllPointsWithNegativeCoordinatesII)
...
+	pnts.push_back(p);
+
+	ASSERT_EQ(1u, pnts.size());
+
+	// construct from points of the vector a axis aligned bounding box
+	GeoLib::AABB aabb(pnts.begin(), pnts.end());
+
+	double const to_lowest(std::numeric_limits<double>::lowest());
+	double const to_max(std::numeric_limits<double>::max());
+	// modify coordinates of p and check again
```
chleh: A very short comment for a very monstrous block of three nested for loops...

Tests/GeoLib/TestAABB.cpp
```diff
@@ -177,6 +190,11 @@ TEST(GeoLibAABB, RandomNumberOfPointsRandomBox)
 	 ASSERT_LE(minus_half_box_size_x, min_pnt[0]) << "coordinate 0 of min_pnt is smaller than " << minus_half_box_size_x;
 	 ASSERT_LE(minus_half_box_size_y, min_pnt[1]) << "coordinate 1 of min_pnt is smaller than " << minus_half_box_size_y;
 	 ASSERT_LE(minus_half_box_size_z, min_pnt[2]) << "coordinate 2 of min_pnt is smaller than " << minus_half_box_size_z;
+
+	// since the interval is half-open we have to move the upper bound also
+	 half_box_size_x = std::nexttoward(half_box_size_x, std::numeric_limits<double>::max());
+	 half_box_size_y = std::nexttoward(half_box_size_y, std::numeric_limits<double>::max());
+	 half_box_size_z = std::nexttoward(half_box_size_z, std::numeric_limits<double>::max());
```
chleh: That's why I don't like your solution that fiddles around with 'infinitesimal floats': It requires this ugly fiddlings also in other places. ^^

GeoLib/AABB.h
TomFischer: See next line: "Furthermore ...."

chleh:
Minor things. Please add the comments as requested and maybe remove the pointer methods. Afterwards: :+1:

GeoLib/AABB.h
TomFischer: Yes, for instance the class `MeshLib::Mesh` stores pointers to nodes in a vector (`std::vector<MeshLib::Node*>`). The pointer versions are used computing the AABB for meshes.

GeoLib/AABB.h
TomFischer: See answer above.

GeoLib/AABB.h
TomFischer: See answer above.

GeoLib/AABB.h
chleh: I see... You probably need the pointer version in `AABB(InputIterator first, InputIterator last)` to have a uniform interface to the outside world.

MeshLib/CoordinateSystem.cpp
endJunction: Yes, this can happen, but at this point we have not enough information to handle this.

Tests/GeoLib/TestAABB.cpp
```diff
@@ -220,3 +238,47 @@ TEST(GeoLib, AABBAllPointsWithNegativeCoordinatesII)
 	ASSERT_NEAR(-1.0, max_pnt[2], std::numeric_limits<double>::epsilon());
 }
 
+TEST(GeoLib, AABBSinglePoint)
+{
```
chleh: This test creates an AABB containing a single point.
This point is moved in every space direction to the next possible double value.
It is checked that the AABB contains the point iff it has not been moved.

GeoLib/AABB.h
```diff
@@ -108,9 +138,9 @@ class AABB
 	template <typename T>
 	bool containsPoint(T const & pnt) const
 	{
-		if (pnt[0] < _min_pnt[0] || _max_pnt[0] < pnt[0]) return false;
-		if (pnt[1] < _min_pnt[1] || _max_pnt[1] < pnt[1]) return false;
-		if (pnt[2] < _min_pnt[2] || _max_pnt[2] < pnt[2]) return false;
+		if (pnt[0] < _min_pnt[0] || _max_pnt[0] <= pnt[0]) return false;
+		if (pnt[1] < _min_pnt[1] || _max_pnt[1] <= pnt[1]) return false;
+		if (pnt[2] < _min_pnt[2] || _max_pnt[2] <= pnt[2]) return false;
```
rinkk: Why does ```<=``` return false? I don't think that is practical.  
Example: I want to check if a Point x is within the bounding box of a mesh. I get AABB and if x is located on the bottom of the mesh, the method returns true. If x is located on the top of the mesh, the method returns false.  
I would definitely prefer the original version.

GeoLib/AABB.h
```diff
@@ -150,19 +180,60 @@ class AABB
...
 	{
 		init(*pnt);
 	}
+
+	/// Private method that is used internally to update the min and max point
+	/// of the bounding box using point \f$p\f$ without enlarging the bounding
+	/// box.
+	/// @param p point that will possibly change the bounding box points
+	template <typename PNT_TYPE>
+	void  updateWithoutEnlarge(PNT_TYPE const & p)
```
rinkk: I think the name ```updateWithoutEnlarge()``` is unfortunate. It suggests that there is an ```updateWithEnlarge()``` that doesn't exist. Also, even though I read the comments I don't understand why you need this method. (Maybe lets talk about this instead of writing long comments)

GeoLib/AABB.h
TomFischer: The maximal point is internally updated such that point located at the right boundaries (back, right or top surface) will fit into the aabb.

The problem with the previous version was that when the bounding box was created from a single point (min point == max point) the point was not in the bounding box itself.

MeshLib/CoordinateSystem.h
```diff
@@ -47,8 +47,7 @@ class CoordinateSystem
     explicit CoordinateSystem(const Element &ele);
 
     /// Decides a coordinate system from a bounding box
-    template <class T>
-    explicit CoordinateSystem(const GeoLib::AABB<T> &bbox) : _type(getCoordinateSystem(bbox)) {}
+    explicit CoordinateSystem(const GeoLib::AABB &bbox) : _type(getCoordinateSystem(bbox)) {}
```
rinkk: I don't really understand what this whole class does and the documentation is not great. Not the point of this PR though. Is it possible that class has really nothing to do with coordinate systems but with extends and dimensionality of data sets?

GeoLib/AABB.h
rinkk: Ah, alright then. :+1:

MeshLib/CoordinateSystem.h
endJunction: Yes, the `CoordinateSystem` describes the dimensionality of a point set but also if a 2D set lies in a plane parallel to the xz plane (or any other plane), or 1D set is parallel to any of the x, y, z axes.

GeoLib/AABB.h
```diff
@@ -15,49 +15,62 @@
...
+ *
+ * Let \f$P = \{p_k \in \mathbb{R}^3, \ k=1, \dotsc, n\}\f$ a set of 3d points.
+ * The bounding volume is described by its lower, left, front point \f$\ell\f$
+ * and the upper, right, back point \f$u\f$, i.e. the coordinates of \f$\ell\f$
+ * and \f$u\f$ are computed as follows \f$\ell_i = \min \limits_{p \in P}
+ * p_i\f$, \f$u_i = \max \limits_{p \in P} p_i\f$, respectively. The bounding
+ * box consists of half-open intervals \f$[\ell_x, u_x) \times [\ell_y, u_y)
+ * \times [\ell_z, u_z).\f$ In order to avoid that points of the defining point
+ * set \f$P\f$ do not lie within the bounding box it is enlarged up to the next
+ * available floating point number.
```
rinkk: The last sentence is imo hard to understand if you don't know already know how it works. Wouldn't it be better and more correct to remove the sentence and instead write the intervalls above as [l_x, u_x+\eps) ?

MeshLib/CoordinateSystem.h
rinkk: Okay. Maybe I am a bit pedantic here but I still think the name ```CoordinateSystem``` is wrong (i.e. suggesting something the class does not do). We can discuss this later, though, it's not related to this PR.

GeoLib/AABB.h
TomFischer: `updateWithoutEnlarge()` is a private method. I think for the user interface the `update()` method is readable and understandable. Internally there is another update mechanism reflected in the different name.

GeoLib/AABB.h
TomFischer: Updated last sentence of the docu.

GeoLib/AABB.h
TomFischer: > Also, even though I read the comments I don't understand why you need this method. (Maybe lets talk about this instead of writing long comments)

Updated the documentation. I think it is now more clear why the method exists.

endJunction:
Now all questions answered, this can be merged, isn't it?
:fast_forward:

PR #931, opened 2015-12-16
Lis tools refactorization

TomFischer:
Moved the transformation "crs matrix" to "lis matrix" to a separate function.


bilke:
:+1:

wenqing:
:+1:

PR #933, opened 2015-12-16
[GL]  Fix bug in GEOObjects.

TomFischer:
Use removeSurfaceVec() instead of removePointVec().


endJunction:
Happened to me while splitting GeoModels from GeoObjects. Not catched because there is no test.
:fast_forward:

Issue #735, opened 2015-07-03
Division by zero in MeshValidation.DetectHolesHex


endJunction:
Found with undefined behaviour sanitzer:

```
bin/testrunner --gtest_filter=MeshValidation.DetectHolesHex      
Note: Google Test filter = MeshValidation.DetectHolesHex
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from MeshValidation
[ RUN      ] MeshValidation.DetectHolesHex
Extracting mesh surface...
/home/naumov/w/ogs/MathLib/Vector3.h:109:19: runtime error: division by zero
Extracting mesh surface...
Extracting mesh surface...
Extracting mesh surface...
[       OK ] MeshValidation.DetectHolesHex (416 ms)
[----------] 1 test from MeshValidation (416 ms total)
```

Backtrace under gtest code:
```
#0  MathLib::TemplateVector3<double>::normalize (this=0x7fffffffc6c0) at /home/naumov/w/ogs/MathLib/Vector3.h:109
#1  0x0000000002b64d3e in MathLib::TemplateVector3<double>::getNormalizedVector (this=0x7fffffffd160) at /home/naumov/w/ogs/MathLib/Vector3.h:118
#2  0x0000000002b5bcd5 in MeshLib::MeshSurfaceExtraction::get2DSurfaceElements (all_elements=..., sfc_elements=..., dir=..., angle=90, mesh_dimension=3)
    at /home/naumov/w/ogs/MeshLib/MeshSurfaceExtraction.cpp:165
#3  0x0000000002b5751c in MeshLib::MeshSurfaceExtraction::getMeshSurface (mesh=..., dir=..., angle=90, keepOriginalNodeIds=false) at /home/naumov/w/ogs/MeshLib/MeshSurfaceExtraction.cpp:80
#4  0x0000000002b5fc22 in MeshLib::MeshSurfaceExtraction::getMeshBoundary (mesh=...) at /home/naumov/w/ogs/MeshLib/MeshSurfaceExtraction.cpp:131
#5  0x0000000002b2c1e6 in MeshLib::MeshValidation::detectHoles (mesh=...) at /home/naumov/w/ogs/MeshLib/MeshQuality/MeshValidation.cpp:155
#6  0x000000000181da27 in MeshValidation_DetectHolesHex_Test::TestBody (this=0x5ce0d50) at /home/naumov/w/ogs/Tests/MeshLib/TestMeshValidation.cpp:82
[...]

PR #935, opened 2015-12-16
CMake options for enabling Clang sanitizer

bilke:
- `OGS_ADDRESS_SANITIZER` for [AddressSanitizer](http://clang.llvm.org/docs/AddressSanitizer.html)
- `OGS_UNDEFINED_BEHAVIOR_SANITIZER` for [UndefinedBehaviorSanitizer](http://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html)

Detects #735.

I will setup Jenkins jobs after merge.


scripts/cmake/ClangSanitizer.cmake
```diff
@@ -0,0 +1,19 @@
...
+option(OGS_ADDRESS_SANITIZER OFF "Use Clangs AddressSanitizer")
+option(OGS_UNDEFINED_BEHAVIOR_SANITIZER OFF "Use Clangs UndefinedBehaviorSanitizer")
+
+if(OGS_ADDRESS_SANITIZER)
+	set(SANITIZE_FLAG_VALUE "address")
+	set(ADDITIONAL_FLAGS "-fno-omit-frame-pointer")
+endif()
+
+if(OGS_UNDEFINED_BEHAVIOR_SANITIZER)
+	set(SANITIZE_FLAG_VALUE "${SANITIZE_FLAG_VALUE},undefined")
```
endJunction: I'm not sure if both sanitizers can be use at the same time. Did not find anything on the [clang docu](http://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html) but usually they specify if sanitizers can be combined (like the [LeakSanitizer](http://clang.llvm.org/docs/LeakSanitizer.html)).

endJunction: Please add `-fsanitize=integer` too for full strength.

bilke: [Docs](http://clang.llvm.org/docs/UsersManual.html#id33) state that:

> It is not possible to combine more than one of the -fsanitize=address, -fsanitize=thread, and -fsanitize=memory checkers in the same program.

So combining `undefined` with `address` should be ok.

bilke: Should `integer` be an independent CMake option or should it be added with `undefined`? And would it then be `-fsanitize=undefined,integer`?

endJunction: Shall be added to `undefined`. Looks like they can be just comma-separated.

Issue #944, opened 2016-01-07
GeoMapper bug


TomFischer:
The method `advancedMapOnMesh()` in class `GeoMapper` does not map polylines correctly when the mesh contains vertical surface elements. As one can see in the figure, the course of the polyline is sometimes below the mesh surface. Furthermore the algorithm is very slow for big meshes and big geometries.

![mappedrivers-meshsurfacesmallunstrutclipping-rivernetwork](https://cloud.githubusercontent.com/assets/800065/12170904/7e528752-b545-11e5-8e96-1e3d698880c6.png)

Imho following points should be worked on:
- [x] Rewrite/correct the algorithm such that the new algorithm is more readable and parts are reusable.
- [x] Test refactored functionality.
- [x] Improve the speed.
- [x] Implement tests.

endJunction:
@TomFischer Is this still WIP?

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/944).

PR #945, opened 2016-01-07
[GL] Make tetgen predicates in OGS available.

TomFischer:
The predicates will be used for a robust line segment intersection algorithm. This algorithm will be used in the reimplementation of the surface mapping algorithm, see issue #944.

### DIFF
--- a/GeoLib/CMakeLists.txt
+++ b/GeoLib/CMakeLists.txt
@@ -2,7 +2,9 @@
 GET_SOURCE_FILES(SOURCES_GeoLib)
 
 # Create the library
-add_library(GeoLib STATIC ${SOURCES_GeoLib})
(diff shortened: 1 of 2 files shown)


bilke:
Until now tetgen is a required submodule for the GUI only. See [SubmoduleSetup.cmake](https://github.com/ufz/ogs/blob/master/scripts/cmake/SubmoduleSetup.cmake#L7-L17), maybe you have to move `ThirdParty/tetgen` to line 12.

TomFischer:
@bilke Thank you for your hint.

endJunction:
Is there a namespace for the predicates, or will they land in global namespace? Or will there be some wrappers in GeoLib?

endJunction:
From discussion with Tom: There predicates will be placed in ExactPredicates namespace. There will be interfaces in GeoLib.

PR #950, opened 2016-01-12
Local boost dependencies

bilke:
- Fixes compilation error due to missing library dependencies on local boost
- When a CMake option `OGS_LIB_XXX=System` is set then stop CMake when the required lib is not found on the system, see #895


TomFischer:
looks good.

Issue #949, opened 2016-01-11
Extract common Process parts/algorithms


endJunction:
Currently the processes (the GroundwaterFlowProcess) is controlling the linear solver creation and matrices. In order to implement non-linear solver loop the linear solver must be independent of the actual process because the non-linear solver loop will control it.

The first step is the extraction of GlobalSetup template parameter from GroundwateFlowProcess up to Process
 - [x] Extract linear solvers; https://github.com/ufz/ogs/pull/951.

To move up the boundary conditions the Dirichlet boundary condition class must be rewritten in same manner as the NeumannBcs.
 - [x] Rewrite DirichletBCs; https://github.com/ufz/ogs/pull/963
 - [x] Move up common BCs initialization and assembly parts. https://github.com/ufz/ogs/pull/982
 - [x] Move up `_mesh_subset_all_nodes` https://github.com/ufz/ogs/pull/982

Output needs some discussion about which mesh properties to write and their corresponding process variables. This is also touching multicomponent issues.
 - [x] Move up process output code. https://github.com/ufz/ogs/pull/998
 - [x] Move `assembleConcreteProcess` to Process.

Renames/refactorings.
 - [x] Construct GWFP outside. https://github.com/ufz/ogs/pull/1021
 - [x] Use unique_ptr for mesh subset(s) https://github.com/ufz/ogs/pull/1061
 - [ ] Rename Process to MonolithicProcess
 - [x] Generalize MonolithicProcess' multicomponent stuff.
 - [x] Replace possible `this->` usages with `Process<GS>::` if any
 - [x] Fix return values of `assemble` and `solve`.

chleh:
Finding the right ProcessVariable and -Parameter in the GWFlow constructor now takes 50+ lines of code. There should be a method in `Process<>` that does the work, so the corresponding section in GWFlow boils down to two l.o.c.

chleh:
TODO: move up '_all_mesh_subsets'

chleh:
@endJunction Can this be closed now?

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/949).

PR #951, opened 2016-01-12
Move up linear solver

endJunction:
This is one step (of many following) to separate process specific functionality (like for the GroundwaterFlowProcess) from common parts of all monolithic processes. This is required for introduction of non-linear solver loop which will control the linear solver solve calls.

In this PR there is no functionality change nor their is any (but unavoidable) generalization, for example to handle multiple DOF per node. It's just moving some of the GroundwaterFlowProcess variables up into the Process class. Following the variables there are some functions.

Major changes in functions are:
 - Process::solve() is split into two parts: GroundwaterFlowProcess::assemble() followed common linear solver solve call.
 - Process::initialize() is split into two parts: Common process initialization and specific GroundwaterFlowProcess::init() part.

There is one thing I like to point out: because the GroundwaterFlow class is derived from a template Process<GlobalSetup> `this` pointer is used to access Process<GlobalSetup>'s protected variables; I'll reduce their numbers in following PRs when more interfaces will be introduced.

_Update:_ I'v just seen that diff for the review look like a mess, so maybe commit-wise review is easier...

### DIFF
--- a/Applications/ApplicationsLib/ProjectData.cpp
+++ b/Applications/ApplicationsLib/ProjectData.cpp
@@ -84,9 +84,6 @@ ProjectData::~ProjectData()
 {
 	delete _geoObjects;
 
-	for(ProcessLib::Process<GlobalSetupType>* p : _processes)
-		delete p;
-
 	for (MeshLib::Mesh* m : _mesh_vec)
 		delete m;
 }
--- a/Applications/ApplicationsLib/ProjectData.cpp
+++ b/Applications/ApplicationsLib/ProjectData.cpp
@@ -84,7 +84,7 @@ ProjectData::~ProjectData()
 {
 	delete _geoObjects;
 
-	for(ProcessLib::Process* p : _processes)
+	for(ProcessLib::Process<GlobalSetupType>* p : _processes)
 		delete p;
 
 	for (MeshLib::Mesh* m : _mesh_vec)
--- a/ProcessLib/GroundwaterFlowProcess.h
+++ b/ProcessLib/GroundwaterFlowProcess.h
@@ -219,7 +219,7 @@ public:
     {
         DBUG("Initialize GroundwaterFlowProcess.");
 
-        setInitialConditions(*_hydraulic_head);
+        setInitialConditions(*_hydraulic_head, 0);
 
         if (this->_mesh.getDimension()==1)
             createLocalAssemblers<1>();
@@ -231,15 +231,17 @@ public:
             assert(false);
     }
 
-    void setInitialConditions(ProcessVariable const& variable)
+    void setInitialConditions(ProcessVariable const& variable,
+                              int const component_id)
     {
         std::size_t const n = this->_mesh.getNNodes();
         for (std::size_t i = 0; i < n; ++i)
         {
             MeshLib::Location const l(this->_mesh.getID(),
                                       MeshLib::MeshItemType::Node, i);
-            auto const global_index = // 0 is the component id.
-              std::abs(this->_local_to_global_index_map->getGlobalIndex(l, 0) );
+            auto const global_index =  // 0 is the component id.
+                std::abs(this->_local_to_global_index_map->getGlobalIndex(
+                    l, component_id));
             this->_x->set(
                 global_index,
                 variable.getInitialConditionValue(*this->_mesh.getNode(i)));
--- a/ProcessLib/GroundwaterFlowProcess.h
+++ b/ProcessLib/GroundwaterFlowProcess.h
@@ -193,9 +193,7 @@ public:
                     *_mesh_subset_all_nodes);
         }
 
-        for (auto bc : _neumann_bcs)
-            bc->initialize(this->_global_setup, *(this->_A), *(this->_rhs),
-                           this->_mesh.getDimension());
+        Process<GlobalSetup>::initializeNeumannBcs(_neumann_bcs);
     }
 
     void initializeMeshSubsets(MeshLib::Mesh const& mesh) override
--- a/ProcessLib/GroundwaterFlowProcess.h
+++ b/ProcessLib/GroundwaterFlowProcess.h
@@ -215,23 +215,9 @@ public:
         return "gw_";
     }
 
-    void initialize() override
+    void init() override
     {
         DBUG("Initialize GroundwaterFlowProcess.");
-
-        DBUG("Construct dof mappings.");
-        initializeMeshSubsets(this->_mesh);
-
-        this->_local_to_global_index_map.reset(
-            new AssemblerLib::LocalToGlobalIndexMap(this->_all_mesh_subsets, AssemblerLib::ComponentOrder::BY_COMPONENT));
-
-        DBUG("Compute sparsity pattern");
-        Process<GlobalSetup>::computeSparsityPattern(
-            *this->_local_to_global_index_map);
-
-        // create global vectors and linear solver
-        Process<GlobalSetup>::createLinearSolver(*this->_local_to_global_index_map,
-                                                 getLinearSolverName());
 
         setInitialConditions(*_hydraulic_head);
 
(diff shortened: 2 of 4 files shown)


Applications/ApplicationsLib/ProjectData.cpp
```diff
@@ -84,7 +84,7 @@ ProjectData::~ProjectData()
 {
 	delete _geoObjects;
 
-	for(ProcessLib::Process* p : _processes)
+	for(ProcessLib::Process<GlobalSetupType>* p : _processes)
```
chleh: now it might be good to have a typedef for it. here also auto would suffice, but typedef would also work below.

ProcessLib/GroundwaterFlowProcess.h
```diff
@@ -146,7 +127,7 @@ class GroundwaterFlowProcess : public Process
     {
         DBUG("Create local assemblers.");
         // Populate the vector of local assemblers.
-        _local_assemblers.resize(_mesh.getNElements());
+        _local_assemblers.resize(this->_mesh.getNElements());
```
chleh: Sure, later on most of the `this->` will vanish, but if there will be some of those constructs left, it might be advisable to rather use `Process<>::` in order to not encourage users to use `this->` all over the place.

ProcessLib/GroundwaterFlowProcess.h
```diff
@@ -131,14 +118,8 @@ class GroundwaterFlowProcess : public Process
...
         // Linear solver options
-        {
-            auto const par = config.get_child_optional("linear_solver");
-
-            if (par)
-            {
-                _linear_solver_options.reset(new BaseLib::ConfigTree(*par));
-            }
-        }
+        Process<GlobalSetup>::setLinearSolveOptions(
```
chleh: there is a typo in the method name: setLinearSolverOptions

ProcessLib/GroundwaterFlowProcess.h
```diff
@@ -200,102 +177,65 @@ class GroundwaterFlowProcess : public Process
...
         }
 
-        for (auto bc : _neumann_bcs)
-            bc->initialize(_global_setup, *_A, *_rhs, _mesh.getDimension());
-
+        Process<GlobalSetup>::initializeNeumannBcs(_neumann_bcs);
     }
 
-    void initialize() override
+    void initializeMeshSubsets(MeshLib::Mesh const& mesh) override
```
chleh: Only a note: For now it is inconsistent to pass the mesh here, but to use the member in other places.

chleh:
General note: I just noticed that there is a problem with commit-by-commit code review:
https://help.github.com/articles/why-are-my-commits-in-the-wrong-order/

chleh:
Good work so far. I've had quite some comments, but none of them addresses a fundamental issue.
So I think, there is no need to make changes to this PR (except for the `computeSparsityPattern()` issue), but it would be nice if you could make a TODO list which will either be worked off in a separate PR or at the end, after the whole refactoring has been completed.
In summary: Answer `computeSparsityPattern()`, make TODO list, then :+1:

chleh:
Maybe for such a big change it would be good, the next time, to initially add some todo-comments (what will be added, moved, removed) to the source code. This way people not directly involved might more easily get an idea of where the changes head to.

ProcessLib/GroundwaterFlowProcess.h
wenqing: @chleh  :+1:

ProcessLib/Process.h
```diff
@@ -12,35 +12,187 @@
...
+
+	std::unique_ptr<GlobalAssembler> _global_assembler;
+
+	std::unique_ptr<AssemblerLib::LocalToGlobalIndexMap>
+	    _local_to_global_index_map;
+
+	std::unique_ptr<BaseLib::ConfigTree> _linear_solver_options;
+	std::unique_ptr<typename GlobalSetup::LinearSolver> _linear_solver;
+
+	std::unique_ptr<typename GlobalSetup::MatrixType> _A;
```
wenqing: We also need a possibility that ``_A, _x, _rhs`` could be shared by other processes.

wenqing:
@endJunction :+1:  This is the right way.

Applications/ApplicationsLib/ProjectData.cpp
endJunction: Don't get it; to be discussed.

ProcessLib/GroundwaterFlowProcess.h
endJunction: Reduces one usage of `this->` :)
Never mind; will rework with `_all_mesh_subsets`.

ProcessLib/Process.h
endJunction: While this is not implemented yet, the current changes don't prohibit it. We shall see how it is possible when there are multiple weakly coupled processes, which shall not happen before March or so.

endJunction:
Addressed all open questions so far.
Updated todo list https://github.com/ufz/ogs/issues/949

__Do not merge__ yet. I'll squash the 'md' (= merge down) commits to the origins before it goes to master. This way you can easier verify the changes.

ProcessLib/GroundwaterFlowProcess.h
```diff
@@ -200,102 +179,65 @@ class GroundwaterFlowProcess : public Process
...
-              std::abs( _local_to_global_index_map->getGlobalIndex(l, 0) );
-            _x->set(global_index,
-                   variable.getInitialConditionValue(*_mesh.getNode(i)));
-        }
-    }
-
-    bool solve(const double /*delta_t*/) override
+    bool assemble(const double /*delta_t*/) override
     {
         DBUG("Solve GroundwaterFlowProcess.");
```
TomFischer: Mismatch of the name of the method (`assemble`) and debug output (Solve ...) .

ProcessLib/Process.h
```diff
@@ -12,35 +12,180 @@
...
-    virtual void postTimestep(std::string const& file_name, const unsigned timestep) = 0;
+	bool solve(const double delta_t)
+	{
+		_A->setZero();
+		MathLib::setMatrixSparsity(*_A, _sparsity_pattern);
+
+		bool const result = assemble(delta_t);
+
+		_linear_solver->solve(*_rhs, *_x);
+		return result;
```
TomFischer: The assemble method of the groundwater process returns always true. The `solve` method return the result from `assemble()`, i.e., always true. This does not depend on the solve process of the system of linear equations. Is this intended?

ProcessLib/Process.h
```diff
@@ -12,35 +12,180 @@
...
+			auto const global_index = std::abs(
+			    _local_to_global_index_map->getGlobalIndex(l, component_id));
+			_x->set(global_index,
+			        variable.getInitialConditionValue(*_mesh.getNode(i)));
+		}
+	}
+
+private:
+	/// Creates global matrix, rhs and solution vectors, and the linear solver.
+	void createLinearSolver(std::string const solver_name)
```
TomFischer: Can the argument be a reference or is the copy of the string necessary?

ProcessLib/Process.h
```diff
@@ -12,35 +12,180 @@
 
 #include <string>
 
+#include <boost/optional.hpp>
```
TomFischer: Is this include still needed?

ProcessLib/Process.h
endJunction: This is in horrible state now, I know. The reason for the return value is to report to nonlinear solver or timestepper that current damping or delta-t are too large. Will fix later. Todo added.

endJunction:
@TomFischer if OK, I'll rebase and merge it.

TomFischer:
Is is ok. Is there any label we could set for the case that the review is finished and the author of the PR can/shall be merge down changes? Maybe this is in general a good approach?!

PR #953, opened 2016-01-14
Conversion tool + warning

rinkk:
Added a warning for vtu to msh conversion that all scalar data will be lost (except mat ids).
Geometry + mesh conversion work in both directions, bc's are still deactivated.


endJunction:
If there is a corresponding command line tool, then it should be added there too.
:fast_forward: otherwise.

PR #955, opened 2016-01-14
ConfigTree -- Follow-Up PR

chleh:
This PR completes the transition to the new ConfigTree for project files.

Changes:
* make a member `mutable` and all methods `const` in order to be able to pass temporary instances directly to function calls
* change parsing of the project file completely to the new ConfigTree
* addresses two comments by @endJunction from #912

TODO (subsequent PR):
* remove the old ConfigTree code
* rename the new ConfigTree

PR #956, opened 2016-01-14
ConfigTree -- Default to strict checking

chleh:
This PR is a follow-up of #955, please only review the last two commits.

Changes:
* The default behaviour of ConfigTreeNew is changed s.t. it crashes also if a warning occurs
* A cmdline flag `--config-warnings-nonfatal` is added that allows to override that behaviour, i.e. warning messages are only printed and do not cause program abortion.

This way users will definitely notice immediately when it is likely that there are errors in their project file.
Additionally this way it is ensured that benchmark input files only contain the minimal set of parameters needed and no additional settings that do not even get parsed.

I guess, there might be a discussion if we should employ such a policy.


chleh:
It seems that there already is some warning raised in the benchmarks...
Mea culpa: I think, I added the failing section to the project file. Will fix it tomorrow.

chleh:
I also uploaded the input file to envinf1.

TomFischer:
Looks good.

bilke:
:+1:

TomFischer:
Jenkins, test this please.

ProcessLib/Process.h
```diff
@@ -150,6 +150,7 @@ class Process
 		_rhs.reset(_global_setup.createVector(num_unknowns));
 		_linear_solver.reset(new typename GlobalSetup::LinearSolver(
 		    *_A, solver_name, _linear_solver_options.get()));
+		_linear_solver_options.reset(nullptr); // trigger error checks early
```
endJunction: This needs a rename. Already discussed with Christoph.

chleh: renamed.

chleh:
My last two commits have been pushed after your reviews. @bilke, @TomFischer, @endJunction, maybe you want to check them.

chleh:
Btw: The changes in the benchmark file are just the removal of a duplicate configuration:
```
<time_stepping>
        <type>SingleStep</type>
    </time_stepping>
```

endJunction:
:fast_forward:

PR #963, opened 2016-01-18
Extract Dirichlet type boundary conditions from GWF process.

endJunction:
This is a preparation PR before moving all of the boundary conditions to Process class (from the GroundwaterFlowProcess).

Also see https://github.com/ufz/ogs/issues/949

### DIFF
--- a/ProcessLib/GroundwaterFlowProcess.h
+++ b/ProcessLib/GroundwaterFlowProcess.h
@@ -34,6 +34,7 @@
 #include "GroundwaterFlowFEM.h"
 #include "NeumannBcAssembler.h"
 #include "NeumannBc.h"
+#include "DirichletBc.h"
 #include "Parameter.h"
 #include "Process.h"
 
@@ -159,9 +160,10 @@ public:
                 _hydraulic_head->getMesh());
 
         _hydraulic_head->initializeDirichletBCs(
-                hydraulic_head_mesh_node_searcher,
-                *this->_local_to_global_index_map, 0,
-                _dirichlet_bc.global_ids, _dirichlet_bc.values);
+            std::back_inserter(_dirichlet_bcs),
+            hydraulic_head_mesh_node_searcher,
+            *this->_local_to_global_index_map,
+            0);
 
         //
         // Neumann boundary conditions.
@@ -232,9 +234,9 @@ public:
         for (auto bc : _neumann_bcs)
             bc->integrate(this->_global_setup);
 
-        MathLib::applyKnownSolution(*this->_A, *this->_rhs, *this->_x,
-                                    _dirichlet_bc.global_ids,
-                                    _dirichlet_bc.values);
+        for (auto const& bc : _dirichlet_bcs)
+            MathLib::applyKnownSolution(*this->_A, *this->_rhs, *this->_x,
+                                        bc.global_ids, bc.values);
 
         return true;
     }
@@ -313,13 +315,7 @@ private:
 
     std::vector<LocalAssembler*> _local_assemblers;
 
-    /// Global ids in the global matrix/vector where the dirichlet bc is
-    /// imposed and their corresponding values.
-    struct DirichletBC {
-        std::vector<GlobalIndexType> global_ids;
-        std::vector<double> values;
-    } _dirichlet_bc;
-
+    std::vector<DirichletBc<GlobalIndexType>> _dirichlet_bcs;
     std::vector<NeumannBc<GlobalSetup>*> _neumann_bcs;
 };
 
--- a/ProcessLib/ProcessVariable.h
+++ b/ProcessLib/ProcessVariable.h
@@ -67,8 +67,7 @@ public:
 		for (auto& bc_config : _dirichlet_bc_configs)
 		{
 			DirichletBc<GlobalIndexType> bc;
-			bc_config->initialize(searcher, dof_table, nodal_dof_idx,
-			                      bc.global_ids, bc.values);
+			bc_config->initialize(searcher, dof_table, nodal_dof_idx, bc);
 			output_bcs++ = bc;
 		}
 	}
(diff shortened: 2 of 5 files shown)


ProcessLib/GroundwaterFlowProcess.h
```diff
@@ -232,9 +234,9 @@ class GroundwaterFlowProcess : public Process<GlobalSetup>
         for (auto bc : _neumann_bcs)
             bc->integrate(this->_global_setup);
 
-        MathLib::applyKnownSolution(*this->_A, *this->_rhs, *this->_x,
-                                    _dirichlet_bc.global_ids,
-                                    _dirichlet_bc.values);
+        for (auto bc : _dirichlet_bcs)
```
chleh: Why not `auto const&`?

endJunction: forgot it... changing now.

ProcessLib/ProcessVariable.h
```diff
@@ -56,10 +57,21 @@ class ProcessVariable
 	/// Returns a mesh on which the process variable is defined.
 	MeshLib::Mesh const& getMesh() const;
 
-	void initializeDirichletBCs(MeshGeoToolsLib::MeshNodeSearcher& searcher,
-	        const AssemblerLib::LocalToGlobalIndexMap& dof_table,
-	        const unsigned nodal_dof_idx,
-	        std::vector<GlobalIndexType>& global_ids, std::vector<double>& values);
+	template <typename OutputIterator, typename... Args>
```
chleh: As far as I can see, `Args...` are not used, so I'd prefer to remove those parameters.

endJunction: Impossible. Who wrote it in first place?

ProcessLib/ProcessVariable.h
```diff
@@ -56,10 +57,21 @@ class ProcessVariable
...
-	        const unsigned nodal_dof_idx,
-	        std::vector<GlobalIndexType>& global_ids, std::vector<double>& values);
+	template <typename OutputIterator>
+	void initializeDirichletBCs(
+	    OutputIterator bcs,
+	    MeshGeoToolsLib::MeshNodeSearcher& searcher,
+	    const AssemblerLib::LocalToGlobalIndexMap& dof_table,
+	    const unsigned nodal_dof_idx)
+	{
+		for (auto& bc : _dirichlet_bcs)
```
chleh: I think, nomenclature is rather confusing: Everything is named `bc`, but has a different meaning for the output iterator than for `_dirichlet_bcs`. I think that should change in the near future.

ProcessLib/ProcessVariable.h
```diff
@@ -56,10 +57,21 @@ class ProcessVariable
...
+		{
+			DirichletBc<GlobalIndexType> new_bc;
+			bc->initialize(searcher, dof_table, nodal_dof_idx,
+			               new_bc.global_ids, new_bc.values);
```
chleh: Could you make the `initialize()` method accept `DirichletBc<>` objects?

ProcessLib/ProcessVariable.h
```diff
@@ -56,10 +57,21 @@ class ProcessVariable
...
+			bcs = new_bc;  // output iterator
```
chleh: The disadvantages of the output iterator are:
* It is rather implicit (the assignment has to be commented for those not used to output iterators)
* Are there any performance issues because of resizing?

chleh:
Three small things. Apart from them :+1:

ProcessLib/ProcessVariable.h
endJunction: - Ideas for better "output iterator" pattern?
 - Yes, but negligible for one-time initialization.

chleh: If you would pass the whole vector, you would explicitly `push_back`. Sure, that would expose more than the minimal set of features. So nevermind. No better idea.

ProcessLib/ProcessVariable.h
```diff
@@ -56,10 +57,20 @@ class ProcessVariable
...
+	    OutputIterator output_bcs,
+	    MeshGeoToolsLib::MeshNodeSearcher& searcher,
+	    const AssemblerLib::LocalToGlobalIndexMap& dof_table,
+	    const unsigned nodal_dof_idx)
+	{
+		for (auto& bc_config : _dirichlet_bc_configs)
+		{
+			DirichletBc<GlobalIndexType> bc;
+			bc_config->initialize(searcher, dof_table, nodal_dof_idx, bc);
+			output_bcs = bc;  // output iterator
```
chleh: Different question: Do you have to increment the output iterator? I'm not used to it, so I don't know.

bilke:
:+1:

ProcessLib/ProcessVariable.h
chleh: Just for the record: Returning a vector and moving/concatenating it on the caller side would be an option.

ProcessLib/ProcessVariable.h
chleh: Found something: http://stackoverflow.com/a/20108111
There must be an increment in order to fulfill the iterator concept.

wenqing:
:+1:

ProcessLib/ProcessVariable.h
endJunction: Thanks Christoph. There is a mistake here. But it works!
How? I think because operator= call on std::back_inserter makes a push_back on the underlying container, this works. But it would not (as pointed out in the SO answer above) for pointers, which are also OutputIterator's.

Will fix it, rebase and merge.
@chleh OK?

chleh: :+1:

endJunction:
All tests passed but on PETSc which is not related to this PR.

PR #970, opened 2016-01-19
Automated Jenkins job creation

bilke:
This PR adds functionality to automatically create Jenkins jobs for a given git repository and then automatically build all branches when pushed (i.e. pull request testing before submitting the pull request). The only thing to be done is to mark a repo to be tested (see [my repo](https://svn.ufz.de:8443/job/OGS-6/job/Docker/job/bilke/) and the [ufz repo](https://svn.ufz.de:8443/job/OGS-6/job/Docker/job/bufz/)).

For the moment these tests will be run in addition to other tests as I am using the [Jenkins Pipeline](https://github.com/jenkinsci/workflow-plugin)-functionality which still misses some features. Also the tests are a little bit slow because they are run on an old machine.

Starting point is the `Jenkinsfile` which loads the `gcc.groovy` where the actual job config is given:

- builds inside a docker container
- runs testrunner and ctest
- publishes test results
- scans console output for error and warnings

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,4 @@
+node('master') {
+	checkout scm
+	load 'scripts/jenkins/gcc.groovy'
+}
--- a/Tests/CMakeLists.txt
+++ b/Tests/CMakeLists.txt
@@ -69,16 +69,21 @@ endif()
 
 ADD_VTK_DEPENDENCY(testrunner)
 
-# Add make-target test which runs the testrunner
-# This should override CTest's predefined test-target but it does not
+# Add make-target tests which runs the testrunner
 if(DEFINED ENV{CI})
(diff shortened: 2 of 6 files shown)


chleh:
I like the idea, but there is one thing:
> automatically build all branches when pushed 

Can that behaviour be changed s.t. only a branch with a specific name or only a specific tag (like `jenkins-test-this`) is built upon push? Because I don't want Jenkins to build all of my branches (especially since some of them are in a nonfunctional state).

bilke:
Yes we can set regexes for branch inclusion and exclusion (based on branch name). I propose to use a postfix, e.g. `my-feature-test` or `my-feature-jenkins`.

endJunction:
Postfix `jenkins` +1

endJunction:
It seems not break the current builds and looks clean. :fast_forward:

PR #974, opened 2016-01-20
MeshAnalysisDialog fix

rinkk:
fixed two issues with analysis dialog

### DIFF
--- a/MeshLib/MeshQuality/MeshValidation.cpp
+++ b/MeshLib/MeshQuality/MeshValidation.cpp
@@ -126,6 +126,9 @@ MeshValidation::ElementErrorCodeOutput(const std::vector<ElementErrorCode> &erro
 
 unsigned MeshValidation::detectHoles(MeshLib::Mesh const& mesh)
 {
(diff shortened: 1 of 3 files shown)


MeshLib/MeshQuality/MeshValidation.cpp
```diff
@@ -126,6 +126,9 @@ MeshValidation::ElementErrorCodeOutput(const std::vector<ElementErrorCode> &erro
 
 unsigned MeshValidation::detectHoles(MeshLib::Mesh const& mesh)
 {
+	if (mesh.getDimension() == 1)
```
endJunction: Does this fix https://github.com/ufz/ogs/issues/735?
If so, please add a comment about fixing this.

TomFischer:
:fast_forward:

MeshLib/MeshQuality/MeshValidation.cpp
rinkk: No it doesn't but I'll take a look...

PR #972, opened 2016-01-20
Test for GeoMapper

TomFischer:
PR contains a test for the class ` GeoMapper`. Test cases are generated using some new implemented autocheck generators (`IntervalGenerator` and `IntervalTupleGenerator`).

PR #977, opened 2016-01-21
GeoMapper refactorization

TomFischer:
The main points of the PR are:
- Using C++11 features like `std::unique_ptr`and range based loops
- Introducing readable type checks for classes `Station`and `Borehole`
- Fix a mem leak (and hopefully not introduce new ones)
- Split of `mapData()` of class `GeoMapper` into several methods:
  - `mapStationData()`
  - `mapPointDataToDEM()`
  - `mapPointDataToMeshSurface()`
- Cleanup (remove old code)
- Improve the speed (the test in PR #972 takes only half of the time)

### DIFF
--- a/MeshGeoToolsLib/GeoMapper.cpp
+++ b/MeshGeoToolsLib/GeoMapper.cpp
@@ -279,10 +279,11 @@ void GeoMapper::advancedMapOnMesh(
 	std::vector<int> closest_geo_point(nMeshNodes);
 	// distance between geo points and mesh nodes in (x,y)-plane
 	std::vector<double> dist(nMeshNodes);
+	auto zero_coords = GeoLib::Point{};  // All coordinates zero.
 	for (std::size_t i=0; i<nMeshNodes; ++i)
 	{
-		auto const zero_coords = GeoLib::Point((*mesh->getNode(i))[0],
-			(*mesh->getNode(i))[1], 0.0, mesh->getNode(i)->getID());
+		zero_coords[0] = (*mesh->getNode(i))[0];
+		zero_coords[1] = (*mesh->getNode(i))[1];
 		GeoLib::Point* pnt = grid.getNearestPoint(zero_coords);
 		dist[i] = MathLib::sqrDist(*pnt, zero_coords);
 		closest_geo_point[i] = (dist[i]<=max_segment_length) ? pnt->getID() : -1;
--- a/MeshGeoToolsLib/GeoMapper.cpp
+++ b/MeshGeoToolsLib/GeoMapper.cpp
@@ -83,12 +83,17 @@ void GeoMapper::mapOnMesh(const MeshLib::Mesh* mesh)
 		return;
 	}
 
+	// the variable _surface_mesh is reused below, so first the existing
+	// _surface_mesh has to be cleaned up
+	if (_surface_mesh)
+		delete _surface_mesh;
+
 	if (mesh->getDimension()<3)
-		this->_surface_mesh = new MeshLib::Mesh(*mesh);
+		_surface_mesh = new MeshLib::Mesh(*mesh);
 	else
 	{
 		const MathLib::Vector3 dir(0,0,-1);
-		this->_surface_mesh = MeshLib::MeshSurfaceExtraction::getMeshSurface(*mesh, dir, 90);
+		_surface_mesh = MeshLib::MeshSurfaceExtraction::getMeshSurface(*mesh, dir, 90);
 	}
 
 	// init grid
--- a/MeshGeoToolsLib/GeoMapper.cpp
+++ b/MeshGeoToolsLib/GeoMapper.cpp
@@ -50,12 +50,22 @@ GeoMapper::~GeoMapper()
 
 void GeoMapper::mapOnDEM(const std::string &file_name)
 {
-	this->_raster = FileIO::AsciiRasterInterface::getRasterFromASCFile(file_name);
+	_raster = FileIO::AsciiRasterInterface::getRasterFromASCFile(file_name);
 	if (! _raster) {
 		ERR("GeoMapper::mapOnDEM(): failed to load %s", file_name.c_str());
 		return;
 	}
-	this->mapData();
+
+	std::vector<GeoLib::Point*> const* pnts(_geo_objects.getPointVec(_geo_name));
+	if (! pnts) {
+		ERR("Geometry \"%s\" does not exist.", _geo_name.c_str());
+		return;
+	}
+	if (GeoLib::isStation((*pnts)[0])) {
+		mapStationData(*pnts);
+	} else {
+		mapPointDataToDEM(*pnts);
+	}
 }
 
 void GeoMapper::mapOnMesh(const std::string &file_name)
@@ -72,7 +82,6 @@ void GeoMapper::mapOnMesh(const MeshLib::Mesh* mesh)
 		ERR("Geometry \"%s\" does not exist.", _geo_name.c_str());
 		return;
 	}
-	bool const is_station(GeoLib::isStation((*pnts)[0]));
 
 	if (mesh->getDimension()<3)
 		this->_surface_mesh = new MeshLib::Mesh(*mesh);
@@ -90,10 +99,10 @@ void GeoMapper::mapOnMesh(const MeshLib::Mesh* mesh)
 	std::vector<MeshLib::Node*> const& flat_nodes (flat_mesh->getNodes());
 	_grid = new GeoLib::Grid<MeshLib::Node>(flat_nodes.cbegin(), flat_nodes.cend());
 
-	if (is_station) {
-		mapStationData();
+	if (GeoLib::isStation((*pnts)[0])) {
+		mapStationData(*pnts);
 	} else {
-		mapData();
+		mapPointDataToMeshSurface();
 	}
 
 	delete _grid;
@@ -111,7 +120,7 @@ void GeoMapper::mapToConstantValue(double value)
 	std::for_each(points->begin(), points->end(), [value](GeoLib::Point* pnt){ (*pnt)[2] = value; });
 }
 
-void GeoMapper::mapStationData()
+void GeoMapper::mapStationData(std::vector<GeoLib::Point*> const& points)
 {
 	double min_val(0), max_val(0);
 	if (_surface_mesh)
@@ -122,11 +131,8 @@ void GeoMapper::mapStationData()
 		max_val = bounding_box.getMaxPoint()[2];
 	}
 
-	std::vector<GeoLib::Point*> const* points (_geo_objects.getPointVec(_geo_name));
-	std::size_t nPoints (points->size());
-	for (unsigned j=0; j<nPoints; ++j)
+	for (auto * pnt : points)
 	{
-		GeoLib::Point* pnt ((*points)[j]);
 		double offset =
 		    (_grid)
 		        ? (getMeshElevation((*pnt)[0], (*pnt)[1], min_val, max_val) -
@@ -146,15 +152,32 @@ void GeoMapper::mapStationData()
 	}
 }
 
-void GeoMapper::mapData()
+void GeoMapper::mapPointDataToDEM(std::vector<GeoLib::Point*> const& points)
 {
-	const std::vector<GeoLib::Point*> *points (this->_geo_objects.getPointVec(this->_geo_name));
-
-	for (auto pnt : pnts)
+	for (auto * pnt : points)
 	{
-		GeoLib::Point* pnt ((*points)[j]);
-		(*pnt)[2] = (_grid) ? getMeshElevation((*pnt)[0],(*pnt)[1], min_val, max_val)
-			                : getDemElevation(*pnt);
+		GeoLib::Point &p(*pnt);
+		p[2] = getDemElevation(p);
+	}
+}
+
+void GeoMapper::mapPointDataToMeshSurface()
+{
+	GeoLib::AABB const aabb(
+		_surface_mesh->getNodes().cbegin(), _surface_mesh->getNodes().cend());
+	double const min_val(aabb.getMinPoint()[2]);
+	double const max_val(aabb.getMaxPoint()[2]);
+
+	for (auto * pnt : pnts) {
+		// check if pnt is inside of the bounding box of the _surface_mesh
+		// projected onto the y-x plane
+		GeoLib::Point &p(*pnt);
+		if (p[0] < aabb.getMinPoint()[0] || aabb.getMaxPoint()[0] < p[0])
+			continue;
+		if (p[1] < aabb.getMinPoint()[1] || aabb.getMaxPoint()[1] < p[1])
+			continue;
+
+		p[2] = getMeshElevation(p[0], p[1], min_val, max_val);
 	}
 }
 
--- a/MeshGeoToolsLib/GeoMapper.cpp
+++ b/MeshGeoToolsLib/GeoMapper.cpp
@@ -99,9 +99,13 @@ void GeoMapper::mapOnMesh(const MeshLib::Mesh* mesh)
 	// init grid
 	MathLib::Point3d origin(std::array<double,3>{{0,0,0}});
 	MathLib::Vector3 normal(0,0,-1);
-	MeshLib::Mesh const*const flat_mesh =
-		MeshLib::projectMeshOntoPlane(*_surface_mesh, origin, normal);
-	std::vector<MeshLib::Node*> const& flat_nodes (flat_mesh->getNodes());
+	std::vector<MeshLib::Node*> flat_nodes;
+	// copy nodes and project the copied nodes to the x-y-plane, i.e. set
+	// z-coordinate to zero
+	for (auto n_ptr : _surface_mesh->getNodes()) {
+		flat_nodes.push_back(new MeshLib::Node(*n_ptr));
+		(*flat_nodes.back())[2] = 0.0;
+	}
 	_grid = new GeoLib::Grid<MeshLib::Node>(flat_nodes.cbegin(), flat_nodes.cend());
 
 	if (GeoLib::isStation((*pnts)[0])) {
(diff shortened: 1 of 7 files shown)


MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -67,25 +78,43 @@ void GeoMapper::mapOnMesh(const std::string &file_name)
 
 void GeoMapper::mapOnMesh(const MeshLib::Mesh* mesh)
 {
+	std::vector<GeoLib::Point*> const* pnts(_geo_objects.getPointVec(_geo_name));
+	if (! pnts) {
+		ERR("Geometry \"%s\" does not exist.", _geo_name.c_str());
+		return;
+	}
+	bool const is_station(GeoLib::isStation((*pnts)[0]));
```
endJunction: Move this down close to the place where it is used. ~109

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -67,25 +78,43 @@ void GeoMapper::mapOnMesh(const std::string &file_name)
...
+
+	if (_surface_mesh)
```
endJunction: This needs a short comment why deleting already existing _surface_mesh.

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -115,34 +138,55 @@ void GeoMapper::mapData()
...
+			GeoLib::Point* layer_pnt = layers[k];
+			(*layer_pnt)[2] = (*layer_pnt)[2] + offset;
 		}
 	}
-	else
+}
+
+void GeoMapper::mapPointDataToDEM()
+{
+	std::vector<GeoLib::Point*> const* pnts(_geo_objects.getPointVec(_geo_name));
```
endJunction: Could be a 'std::vector<> const&'.

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -115,34 +138,55 @@ void GeoMapper::mapData()
...
+	}
+}
+
+void GeoMapper::mapPointDataToMeshSurface(std::vector<GeoLib::Point*> const& pnts)
+{
+	double min_val(0), max_val(0);
+	GeoLib::AABB aabb(
+		_surface_mesh->getNodes().cbegin(), _surface_mesh->getNodes().cend());
+	min_val = aabb.getMinPoint()[2];
+	max_val = aabb.getMaxPoint()[2];
```
endJunction: The `aabb`, `min_val`, and `max_val` all could be `const`.

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -163,20 +207,20 @@ double GeoMapper::getMeshElevation(
 	    _surface_mesh->getNode(pnt->getID())->getElements());
 	GeoLib::Point* intersection(nullptr);
 
-	for (std::size_t i = 0; i < elements.size(); ++i)
+	for (auto & element : elements)
```
endJunction: Not sure about accesses below, but could be `auto const&` probably.

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -185,27 +229,19 @@ double GeoMapper::getMeshElevation(
...
-}
-
-std::unique_ptr<std::vector<GeoLib::Polyline*>> copyPolylinesVector(
-    std::vector<GeoLib::Polyline*> const& polylines,
-    std::vector<GeoLib::Point*> const& points)
-{
-	std::size_t nLines = polylines.size();
-	auto new_lines = std::unique_ptr<std::vector<GeoLib::Polyline*>>(
-	    new std::vector<GeoLib::Polyline*>(nLines));
+	std::size_t nLines = polylines->size();
```
endJunction: const

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -185,27 +229,19 @@ double GeoMapper::getMeshElevation(
...
+	auto new_lines = new std::vector<GeoLib::Polyline*>(nLines);
```
endJunction: Why not keeping it a unique_ptr?

MeshGeoToolsLib/GeoMapper.cpp
TomFischer: Using `auto const&` is possible. Changed.

MeshGeoToolsLib/GeoMapper.cpp
TomFischer: Rebase mistake. Changed back.

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -50,12 +50,22 @@ GeoMapper::~GeoMapper()
...
 		return;
 	}
-	this->mapData();
+
+	std::vector<GeoLib::Point*> const* pnts(_geo_objects.getPointVec(_geo_name));
+	if (! pnts) {
+		ERR("Geometry \"%s\" does not exist.", _geo_name.c_str());
+		return;
+	}
+	if (GeoLib::isStation((*pnts)[0])) {
```
chleh: Can `pnts` be empty?

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -67,25 +77,44 @@ void GeoMapper::mapOnMesh(const std::string &file_name)
...
 	// init grid
 	MathLib::Point3d origin(std::array<double,3>{{0,0,0}});
 	MathLib::Vector3 normal(0,0,-1);
-	MeshLib::Mesh const*const flat_mesh =
-		MeshLib::projectMeshOntoPlane(*_surface_mesh, origin, normal);
-	std::vector<MeshLib::Node*> const& flat_nodes (flat_mesh->getNodes());
+	std::vector<MeshLib::Node*> flat_nodes;
+	for (auto n_ptr : _surface_mesh->getNodes()) {
+		flat_nodes.push_back(new MeshLib::Node(*n_ptr));
+		(*flat_nodes.back())[2] = 0.0;
```
chleh: Looks awkward. Could you add a comment like
`// setting z-component to zero`

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -67,25 +77,44 @@ void GeoMapper::mapOnMesh(const std::string &file_name)
...
+	}
 	_grid = new GeoLib::Grid<MeshLib::Node>(flat_nodes.cbegin(), flat_nodes.cend());
-	this->mapData();
+
+	if (GeoLib::isStation((*pnts)[0])) {
```
chleh: Again: What if empty?

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -115,34 +138,54 @@ void GeoMapper::mapData()
 		min_val = bounding_box.getMinPoint()[2];
 		max_val = bounding_box.getMaxPoint()[2];
 	}
-	std::size_t nPoints (points->size());
 
-	if (!is_borehole)
+	std::vector<GeoLib::Point*> const* points (_geo_objects.getPointVec(_geo_name));
```
chleh: you already get this vector at the caller site, so it could be passed as an argument. (not so important)

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -115,34 +138,54 @@ void GeoMapper::mapData()
...
+	std::size_t nPoints (points->size());
+	for (unsigned j=0; j<nPoints; ++j)
```
chleh: `for (auto const* pnt : points)` would save two lines. ^^
Similar thing applies to the inner loop. But merely a question of taste.

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -115,34 +138,54 @@ void GeoMapper::mapData()
...
 		}
 	}
-	else
+}
+
+void GeoMapper::mapPointDataToDEM()
+{
+	std::vector<GeoLib::Point*> const& pnts(*_geo_objects.getPointVec(_geo_name));
+
+	for (auto pnt : pnts)
```
chleh: Caution, implicit stuff! The loop only works because `pnts` contains pointers. It would not work if `pnts` contained objects. This subtlety can be easily overlooked.
I think `for (auto* pnt : pnts)` should work here and would be better, because it stresses the fact of coping with pointers.

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -115,34 +138,54 @@ void GeoMapper::mapData()
...
+}
+
+void GeoMapper::mapPointDataToMeshSurface(std::vector<GeoLib::Point*> const& pnts)
+{
+	GeoLib::AABB const aabb(
+		_surface_mesh->getNodes().cbegin(), _surface_mesh->getNodes().cend());
+	double const min_val(aabb.getMinPoint()[2]);
+	double const max_val(aabb.getMaxPoint()[2]);
+
+	for (auto pnt : pnts) {
```
chleh: Same `auto*` issue here.

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -115,34 +138,54 @@ void GeoMapper::mapData()
...
+		GeoLib::Point &p(*pnt);
+		if (p[0] < aabb.getMinPoint()[0] || aabb.getMaxPoint()[0] < p[0])
```
chleh: Could you add a comment like
`// checking if pnt is inside the bounding box of the _surface_mesh projected onto the x-y plane`
I guess that that's what is checked

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -185,27 +228,20 @@ double GeoMapper::getMeshElevation(
...
-                          std::vector<GeoLib::Point*> const& points)
-{
-	auto it (std::find(points.begin(), points.end(), pnt));
-	return static_cast<unsigned>(std::distance(points.begin(), it));
-}
-
 std::unique_ptr<std::vector<GeoLib::Polyline*>> copyPolylinesVector(
-    std::vector<GeoLib::Polyline*> const& polylines,
-    std::vector<GeoLib::Point*> const& points)
+	const std::vector<GeoLib::Polyline*>* polylines,
```
chleh: Actually I liked the old signature better. But there probably is a reason for switching to pointers.

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -185,27 +228,20 @@ double GeoMapper::getMeshElevation(
...
+	std::vector<GeoLib::Point*>* points)
```
chleh: Why not a `const` pointer?

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -228,32 +264,30 @@ void GeoMapper::advancedMapOnMesh(
...
 		[](GeoLib::Point* p) { return new GeoLib::Point((*p)[0],(*p)[1],0.0); });
 
-	auto new_lines = copyPolylinesVector(
-	    *_geo_objects.getPolylineVec(this->_geo_name), *new_points);
+	auto new_lines = copyPolylinesVector(_geo_objects.getPolylineVec(_geo_name),
+	                                     new_points.get());
 
 	GeoLib::Grid<GeoLib::Point> grid(new_points->begin(), new_points->end());
-	double max_segment_length (this->getMaxSegmentLength(*new_lines));
+	double max_segment_length(getMaxSegmentLength(*(new_lines.get())));
```
chleh: Why not simply `*new_lines`?

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -340,7 +374,7 @@ GeoLib::Point* GeoMapper::calcIntersection(MathLib::Point3d const*const p1, Math
 	const double y1 = (*p1)[1], y2 = (*p2)[1], y3 = (*q1)[1], y4 = (*q2)[1];
 
 	const double det = (x1 - x2) * (y3 - y4) - (y1 - y2) * (x3 - x4);
-	if (fabs(det) < std::numeric_limits<double>::epsilon()) return NULL;
+	if (fabs(det) < std::numeric_limits<double>::epsilon()) return nullptr;
```
chleh: `std::abs` might be better. (in the sense of being consistent in only using math functions from `std` namespace)

chleh:
Only minor things. Please add the comments as suggested and if you agree with me, also change the `auto` related things. The rest is not so important, so do the changes ony if you want to.
Besides that :+1:

MeshGeoToolsLib/GeoMapper.cpp
TomFischer: The geometry management object `_geo_objects` (class `GEOObjects`) doesn't allow to insert an empty vector of points. Later, access to the vector of points is only possible via const methods
```vector<Point*> const* GEOObjects::getPointVec(...)```.

Of course, the user could cast away the const and remove points from the vector. This would have fatal consequences at some other locations in the GeoLib, too.

So the answer is no, but the user can use `GEOObjects` such that `pnts` could be empty.

File: MeshGeoToolsLib/GeoMapper.cpp (hunk 10 above)
Comment (TomFischer) [inline MeshGeoToolsLib/GeoMapper.cpp]: The `pnts` vector should not be empty.

File: MeshGeoToolsLib/GeoMapper.cpp (hunk 8 above)
Comment (chleh) [inline MeshGeoToolsLib/GeoMapper.cpp]: OK, if people want to shoot themselves into their feet, they should do so.

**Comment by chleh** [comment]
:+1:

File: MeshGeoToolsLib/GeoMapper.cpp (hunk 10 above)
Comment (endJunction) [inline MeshGeoToolsLib/GeoMapper.cpp]: @chleh It's checked in line 81 above.

Comment (chleh) [inline MeshGeoToolsLib/GeoMapper.cpp]: @endJunction No, line 81 checks that `pnts` is not a `nullptr`. It does not check if `*pnts` is not empty. But I think @TomFischer's explanation above is OK.

File: MeshGeoToolsLib/GeoMapper.cpp [hunk 20]
```diff
@@ -67,25 +77,46 @@ void GeoMapper::mapOnMesh(const std::string &file_name)
...
+		_surface_mesh = MeshLib::MeshSurfaceExtraction::getMeshSurface(*mesh, dir, 90);
 	}
 
 	// init grid
 	MathLib::Point3d origin(std::array<double,3>{{0,0,0}});
 	MathLib::Vector3 normal(0,0,-1);
-	MeshLib::Mesh const*const flat_mesh =
-		MeshLib::projectMeshOntoPlane(*_surface_mesh, origin, normal);
-	std::vector<MeshLib::Node*> const& flat_nodes (flat_mesh->getNodes());
+	std::vector<MeshLib::Node*> flat_nodes;
```
endJunction: This could be simple vector of Nodes, because you make a copy + change it later.
```diff
diff --git a/MeshGeoToolsLib/GeoMapper.cpp b/MeshGeoToolsLib/GeoMapper.cpp
index 39cd19d..d46284d 100644
--- a/MeshGeoToolsLib/GeoMapper.cpp
+++ b/MeshGeoToolsLib/GeoMapper.cpp
@@ -99,12 +99,13 @@ void GeoMapper::mapOnMesh(const MeshLib::Mesh* mesh)
        // init grid
        MathLib::Point3d origin(std::array<double,3>{{0,0,0}});
        MathLib::Vector3 normal(0,0,-1);
-       std::vector<MeshLib::Node*> flat_nodes;
+       std::vector<MeshLib::Node> flat_nodes;
        // copy nodes and project the copied nodes to the x-y-plane, i.e. set
        // z-coordinate to zero
+       flat_nodes.reserve(_surface_mesh->getNNodes());
        for (auto n_ptr : _surface_mesh->getNodes()) {
-               flat_nodes.push_back(new MeshLib::Node(*n_ptr));
-               (*flat_nodes.back())[2] = 0.0;
+               flat_nodes.emplace_back(*n_ptr);
+               flat_nodes.back()[2] = 0.0;
        }
        _grid = new GeoLib::Grid<MeshLib::Node>(flat_nodes.cbegin(), flat_nodes.cend());
 
@@ -115,8 +116,6 @@ void GeoMapper::mapOnMesh(const MeshLib::Mesh* mesh)
        }
 
        delete _grid;
-       for (auto n_ptr : flat_nodes)
-               delete n_ptr;
 }
 
 void GeoMapper::mapToConstantValue(double value)
```

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -115,34 +140,54 @@ void GeoMapper::mapData()
...
+		double offset =
+		    (_grid)
+		        ? (getMeshElevation((*pnt)[0], (*pnt)[1], min_val, max_val) -
+		           (*pnt)[2])
+		        : getDemElevation(*pnt);
+
+		if (!GeoLib::isBorehole(pnt))
+			continue;
+		GeoLib::StationBorehole* borehole = static_cast<GeoLib::StationBorehole*>(pnt);
+		const std::vector<GeoLib::Point*> layers = borehole->getProfile();
```
endJunction: Could be `... = pnt->getProfile();` and then the `borehole` pointer is not needed.

MeshGeoToolsLib/GeoMapper.cpp
TomFischer: See commit 2f86094

MeshGeoToolsLib/GeoMapper.cpp
TomFischer: Fixed in commit: 8e9ff7f7de2606e8a59fe50de26d70bdf6278408

MeshGeoToolsLib/GeoMapper.cpp
endJunction: Right :zap:

Issue #971, opened 2016-01-20
Dirichlet BC brocken under PETSc after PR #963 being merged


wenqing:
It will be fixed in the next week.

wenqing:
The test could not be run on EVE either.

endJunction:
here is my backtrace for the Tests/Data/EllipticPETSc/quad_20x10_GroundWaterFlow.prj on one of the threads:

In essence it is:
```
BaseLib::ConfigTreeNew::onerror (path="processes/process", message="Key <numerics> has been read 1 time(s) less than it was present in the configuration tree.")
    at /home/naumov/w/ogs/BaseLib/ConfigTreeNew.cpp:152
ConfigTreeNew.cpp:145
ConfigTreeNew.cpp:239
ConfigTreeNew.cpp:57
ProjectData.h:113
CLI/ogs.cpp:125
```

Full
```
(gdb) bt
#0  0x00007fffee52c5f8 in raise () from /usr/lib/libc.so.6
#1  0x00007fffee52da7a in abort () from /usr/lib/libc.so.6
#2  0x00000000005cb530 in BaseLib::ConfigTreeNew::onerror (path="processes/process", message="Key <numerics> has been read 1 time(s) less than it was present in the configuration tree.")
    at /home/naumov/w/ogs/BaseLib/ConfigTreeNew.cpp:152
#3  0x00000000005ca5e4 in std::function<void (std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)>::operator()(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) const (this=<optimized out>, __args=..., __args=...) at /usr/bin/../lib64/gcc/x86_64-unknown-linux-gnu/5.3.0/../../../../include/c++/5.3.0/functional:2271
#4  BaseLib::ConfigTreeNew::warning (this=0xb04990, message=...) at /home/naumov/w/ogs/BaseLib/ConfigTreeNew.cpp:145
#5  BaseLib::ConfigTreeNew::checkAndInvalidate (this=0xb04990) at /home/naumov/w/ogs/BaseLib/ConfigTreeNew.cpp:239
[...]

PR #981, opened 2016-01-21
make ConfigTreeNew more robust against nullptr accesses

chleh:
When examining #971 it was found that there was—additionally to the tag issue—a nullptr dereference. This PR introduces a methods that can savely be run on nullptrs, avoiding such errors.

Note: the call in `Process.h` now uses argument dependent lookup.


endJunction:
:fast_forward:

bilke:
:+1:

PR #982, opened 2016-01-21
Move up process bcs

endJunction:
Second chunk of moves from GroundwaterFlowProcess up into the Process class.
As the result the GWFP now has only the local assemblers and the hydraulic conductivity members.

Initial and both types of boundary conditions were directly moved. To generalize the initialization the Process class has a new member _process_variables pointing to the used set of process variables (defined in ProjectData).

Initialization of the mesh subsets (components) was finally moved up.

Next steps are:
 - move up post() for output
 - extend the Process class to handle multiple components. (For now at the corresponding places there is a 0 passed around.)

### DIFF
--- a/ProcessLib/Process.h
+++ b/ProcessLib/Process.h
@@ -49,7 +49,7 @@ public:
 	}
 
 	/// Process specific initialization called by initialize().
-	virtual void init() = 0;
+	virtual void createLocalAssemblers() = 0;
 	virtual bool assemble(const double delta_t) = 0;
 
 	virtual std::string getLinearSolverName() const = 0;
@@ -86,7 +86,7 @@ public:
 		_global_assembler.reset(
 		    new GlobalAssembler(*_A, *_rhs, *_local_to_global_index_map));
 
-		init();  // Execute proces specific initialization.
+		createLocalAssemblers();
 
 		for (auto const& pv : _process_variables)
 		{
--- a/ProcessLib/Process.h
+++ b/ProcessLib/Process.h
@@ -10,7 +10,11 @@
 #ifndef PROCESS_LIB_PROCESS_H_
 #define PROCESS_LIB_PROCESS_H_
 
+#include <memory>
 #include <string>
+
+#include <logog/include/logog.hpp>
+
 
 #include "AssemblerLib/ComputeSparsityPattern.h"
 #include "AssemblerLib/LocalToGlobalIndexMap.h"
@@ -27,7 +31,11 @@
 #include "MathLib/LinAlg/PETSc/PETScMatrixOption.h"
 #endif
 
+#include "DirichletBc.h"
+#include "NeumannBc.h"
+#include "NeumannBcAssembler.h"
 #include "ProcessVariable.h"
+#include "UniformDirichletBoundaryCondition.h"
 
 namespace MeshLib
 {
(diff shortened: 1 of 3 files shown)


ProcessLib/ProcessVariable.h
```diff
@@ -81,8 +81,9 @@ class ProcessVariable
 		for (auto& config : _neumann_bc_configs)
 		{
 			config->initialize(searcher);
-			bcs++ = new NeumannBc<GlobalSetup>(*config,
-			                                   std::forward<Args>(args)...);
+			bcs++ = std::unique_ptr<NeumannBc<GlobalSetup>>{
```
chleh: For the future: The parameter `GlobalSetup const&` is not used. So maybe it can be removed in the future?

ProcessLib/Process.h
```diff
@@ -111,6 +135,19 @@ class Process
 			std::move(config)));
 	}
 
+private:
+	/// Creates mesh subsets, i.e. components, for given mesh.
+	void initializeMeshSubsets()
+	{
+		// Create single component dof in every of the mesh's nodes.
+		_mesh_subset_all_nodes =
+		    new MeshLib::MeshSubset(_mesh, &_mesh.getNodes());
```
chleh: I can't see a corresponding `delete`.

chleh:
One small thing, apart from that :+1:

ProcessLib/ProcessVariable.h
endJunction: This one is explicitly for template argument lookup. Without GlobalSetup function argument I'd need to specify all template arguments when calling, which is cumbersome.

chleh: I see. So, we should tackle this with 30k lines of template metaprogramming. ^^

ProcessLib/Process.h
endJunction: Agree with transformation. Note taken for later.
Making it unique_ptr requires changes in the dof table...

wenqing:
:+1:

ProcessLib/Process.h
```diff
@@ -162,7 +232,10 @@ class Process
 	}
 
 protected:
+	unsigned const _integration_order = 2;
```
TomFischer: Does this mean the intergration order is fixed to 2 for all proecesses?

endJunction: By default yes, but it is a protected var, so can be set to anything.

TomFischer: The variable is declared using const qualifier. How is it possible to changed the value later on?

ProcessLib/Process.h
```diff
@@ -184,6 +257,11 @@ class Process
 	std::unique_ptr<typename GlobalSetup::VectorType> _x;
 
 	AssemblerLib::SparsityPattern _sparsity_pattern;
+
+	std::vector<DirichletBc<GlobalIndexType>> _dirichlet_bcs;
+	std::vector<std::unique_ptr<NeumannBc<GlobalSetup>>> _neumann_bcs;
+
+	std::vector<ProcessVariable*> _process_variables;
```
TomFischer: Why not `std::vector<std::unique_ptr<ProcessVariable>> _process_variables;`?

TomFischer:
:+1:

ProcessLib/Process.h
endJunction: As it is, yes, in the initialization list.
Otherwise needs to be changed to non-constant...

ProcessLib/Process.h
endJunction: The vector of process variables is not owning. Therefore there is also no dtor here but in ProjectData. I'm adding a comment here.

PR #985, opened 2016-01-22
Fixes for Windows cross-compilation

bilke:
In preparation for building on a new Jenkins Docker slave (will be available after some WKDV-work).

### DIFF
--- a/FileIO/CMakeLists.txt
+++ b/FileIO/CMakeLists.txt
@@ -61,6 +61,12 @@ target_link_libraries(FileIO INTERFACE
 )
 if(QT4_FOUND)
 	target_link_libraries(FileIO PUBLIC Qt4::QtXml Qt4::QtXmlPatterns)
+	if(WIN32 AND CMAKE_CROSSCOMPILING)
+		find_package(OpenSSL)
+		if(OPENSSL_FOUND)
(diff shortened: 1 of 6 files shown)


FileIO/CMakeLists.txt
```diff
@@ -61,6 +61,12 @@ target_link_libraries(FileIO INTERFACE
 )
 if(QT4_FOUND)
 	target_link_libraries(FileIO PUBLIC Qt4::QtXml Qt4::QtXmlPatterns)
+	if(WIN32 AND CMAKE_CROSSCOMPILING)
+		find_package(OpenSSL)
+		if(OPENSSL_FOUND)
```
chleh: Just curious: Why do we need OpenSSL?

chleh:
Looks good to me, but I don't know much about it. :+1:

FileIO/CMakeLists.txt
bilke: We don't need SSL but the prebuilt Qt package from http://pkg.mxe.cc is built against it which means that `QtNetwork` depends on it (and `QtXmlPatterns` (containing the XML validator) depends on `QtNetwork`).

TomFischer:
Looks good to me, too.

Issue #968, opened 2016-01-19
Issue with scalar data for AddTopLayer


rinkk:
The AddTopLayer algorithms expands the MaterialIDs array but not any other scalar arrays that might be stored in the mesh.
If a mesh contains scalar arrays in addition to MaterialIDs the result of the algorithm is an invalid mesh that can not be loaded into OGS.

TomFischer:
One possible solution is to copy and to update only the MaterialIDs property.

TomFischer:
Fixed by PR #986.

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/968).

PR #986, opened 2016-01-22
AddLayer properties

TomFischer:
This PR fixes the bug from issue #968. The property vectors "MaterialIDs" is updated correctly. All other property vectors of the original mesh are removed from the new mesh. Furthermore a mem leaks is removed. The tool can output the new mesh in vtu or old mesh format.

### DIFF
--- a/Applications/Utils/MeshEdit/AddTopLayer.cpp
+++ b/Applications/Utils/MeshEdit/AddTopLayer.cpp
@@ -18,14 +18,11 @@
 
 #include "BaseLib/FileTools.h"
 
-// FileIO
 #include "FileIO/readMeshFromFile.h"
-#include "FileIO/VtkIO/VtuInterface.h"
+#include "FileIO/writeMeshToFile.h"
 
-// MeshLib
 #include "MeshLib/Mesh.h"
 #include "MeshLib/MeshEditing/AddLayerToMesh.h"
-
 
 int main (int argc, char* argv[])
 {
@@ -69,8 +66,7 @@ int main (int argc, char* argv[])
 	}
 
 	INFO("Writing mesh \"%s\" ... ", mesh_out_arg.getValue().c_str());
-	FileIO::VtuInterface mesh_io(result.get(), vtkXMLWriter::Binary);
-	mesh_io.writeToFile(mesh_out_arg.getValue());
+	FileIO::writeMeshToFile(*result, mesh_out_arg.getValue());
 	INFO("done.");
 
 	return EXIT_SUCCESS;
--- a/Applications/Utils/MeshEdit/AddTopLayer.cpp
+++ b/Applications/Utils/MeshEdit/AddTopLayer.cpp
@@ -53,17 +53,25 @@ int main (int argc, char* argv[])
 	cmd.parse(argc, argv);
 
 	INFO("Reading mesh \"%s\" ... ", mesh_arg.getValue().c_str());
-	MeshLib::Mesh * subsfc_mesh(FileIO::readMeshFromFile(mesh_arg.getValue()));
+	auto subsfc_mesh = std::unique_ptr<MeshLib::Mesh>(
+	    FileIO::readMeshFromFile(mesh_arg.getValue()));
+	if (!subsfc_mesh) {
+		ERR("Error reading mesh \"%s\".", mesh_arg.getValue().c_str());
+		return EXIT_FAILURE;
+	}
 	INFO("done.");
 
-	std::unique_ptr<MeshLib::Mesh> result (MeshLib::addTopLayerToMesh(
-		*subsfc_mesh, layer_thickness_arg.getValue(), mesh_out_arg.getValue()
-	));
+	std::unique_ptr<MeshLib::Mesh> result(MeshLib::addTopLayerToMesh(
+	    *subsfc_mesh, layer_thickness_arg.getValue(), mesh_out_arg.getValue()));
+	if (!result) {
+		ERR("Failure while adding top layer.")
+		return EXIT_FAILURE;
+	}
 
 	INFO("Writing mesh \"%s\" ... ", mesh_out_arg.getValue().c_str());
 	FileIO::VtuInterface mesh_io(result.get(), vtkXMLWriter::Binary);
 	mesh_io.writeToFile(mesh_out_arg.getValue());
 	INFO("done.");
 
-	return 0;
+	return EXIT_SUCCESS;
 }
--- a/Applications/Utils/MeshEdit/AddTopLayer.cpp
+++ b/Applications/Utils/MeshEdit/AddTopLayer.cpp
@@ -14,11 +14,8 @@
 // TCLAP
 #include "tclap/CmdLine.h"
 
-// ThirdParty/logog
-#include "logog/include/logog.hpp"
+#include "Applications/ApplicationsLib/LogogSetup.h"
 
-// BaseLib
-#include "BaseLib/LogogSimpleFormatter.h"
 #include "BaseLib/FileTools.h"
 
 // FileIO
@@ -32,10 +29,7 @@
 
 int main (int argc, char* argv[])
 {
-	LOGOG_INITIALIZE();
-	logog::Cout* logog_cout (new logog::Cout);
-	BaseLib::LogogSimpleFormatter *custom_format (new BaseLib::LogogSimpleFormatter);
-	logog_cout->SetFormatter(*custom_format);
+	ApplicationsLib::LogogSetup logog_setup;
 
 	TCLAP::CmdLine cmd(
 		"Adds a top layer to an existing mesh",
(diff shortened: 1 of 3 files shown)


Applications/Utils/MeshEdit/AddTopLayer.cpp
```diff
@@ -17,25 +17,20 @@
 // ThirdParty/logog
 #include "logog/include/logog.hpp"
 
-// BaseLib
+#include "Applications/ApplicationsLib/LogogSetup.h"
```
endJunction: Please have a look if logog.hpp and LogogSimpleFormatter.h are still required.

endJunction:
:fast_forward: when logog includes are clarified.

MeshLib/MeshEditing/AddLayerToMesh.cpp
```diff
@@ -114,7 +115,15 @@ MeshLib::Mesh* addLayerToMesh(MeshLib::Mesh const& mesh, double thickness,
...
+	std::vector<std::string> prop_names(subsfc_props.getPropertyVectorNames());
+	for (auto const& name : prop_names) {
+		if (name != "MaterialIDs") {
+			subsfc_props.removePropertyVector(name);
+		}
+	}
+	// update the MaterialIDs property
 	boost::optional<MeshLib::PropertyVector<int> &> opt_materials(
 		subsfc_props.getPropertyVector<int>("MaterialIDs")
 	);
```
rinkk: If you only leave MatIDs in there, wouldn't it be faster/better to create a new Properties object instead of first copying everything and then deleting most of it?

Applications/Utils/MeshEdit/AddTopLayer.cpp
```diff
@@ -59,17 +50,23 @@ int main (int argc, char* argv[])
 	cmd.parse(argc, argv);
 
 	INFO("Reading mesh \"%s\" ... ", mesh_arg.getValue().c_str());
-	MeshLib::Mesh * subsfc_mesh(FileIO::readMeshFromFile(mesh_arg.getValue()));
+	auto subsfc_mesh = std::unique_ptr<MeshLib::Mesh>(
+	    FileIO::readMeshFromFile(mesh_arg.getValue()));
+	if (! subsfc_mesh.get()) {
```
chleh: Only a note: `!subsfc_mesh` would have been sufficient.

Applications/Utils/MeshEdit/AddTopLayer.cpp
```diff
@@ -59,17 +50,23 @@ int main (int argc, char* argv[])
...
+	    *subsfc_mesh, layer_thickness_arg.getValue(), mesh_out_arg.getValue()));
+	if (! result.get()) {
+		ERR("Failure while adding top layer.")
+		return EXIT_FAILURE;
+	}
 
 	INFO("Writing mesh \"%s\" ... ", mesh_out_arg.getValue().c_str());
-	FileIO::VtuInterface mesh_io(result.get(), vtkXMLWriter::Binary);
-	mesh_io.writeToFile(mesh_out_arg.getValue());
+	FileIO::writeMeshToFile(*(result.get()), mesh_out_arg.getValue());
```
chleh: Also only a note: `*result` does the job.

MeshLib/MeshEditing/AddLayerToMesh.cpp
```diff
@@ -114,23 +114,28 @@ MeshLib::Mesh* addLayerToMesh(MeshLib::Mesh const& mesh, double thickness,
...
 		);
 
-	MeshLib::Properties subsfc_props (mesh.getProperties());
-	boost::optional<MeshLib::PropertyVector<int> &> opt_materials(
-		subsfc_props.getPropertyVector<int>("MaterialIDs")
+	boost::optional<MeshLib::PropertyVector<int> const&> opt_materials(
+		mesh.getProperties().getPropertyVector<int>("MaterialIDs")
 	);
-	if (!opt_materials) {
-		ERR("Can not set material properties for new layer");
```
chleh: Now there is no error message anymore when `if (!opt_materials)` holds. Or is that normal behaviour?

MeshLib/MeshEditing/AddLayerToMesh.cpp
```diff
@@ -114,23 +114,28 @@ MeshLib::Mesh* addLayerToMesh(MeshLib::Mesh const& mesh, double thickness,
...
+	auto new_mesh = new MeshLib::Mesh(name, subsfc_nodes, subsfc_elements);
+	if (opt_materials) {
+		boost::optional<PropertyVector<int> &> new_opt_materials(
+		new_mesh->getProperties().createNewPropertyVector<int>("MaterialIDs",
+			MeshLib::MeshItemType::Cell, 1));
+		if (!new_opt_materials) {
+			ERR("Can not set material properties for new layer");
+		} else {
+			unsigned new_layer_id(*(std::max_element(
+				(*opt_materials).cbegin(), (*opt_materials).cend()))+1);
```
chleh: Only a note: `opt_materials->cbegin()` would look better.

chleh:
Code looks good. Only one question regarding an error message. :+1:

Applications/Utils/MeshEdit/AddTopLayer.cpp
TomFischer: ` FileIO::writeMeshToFile` expects a reference to a MeshLib::Mesh. Writing `result.get()` the internal pointer of `std::unique_ptr` is returned and with `*` the returned pointer is dereferenced. Maybe I overlooked something?

MeshLib/MeshEditing/AddLayerToMesh.cpp
TomFischer: Fixed in commit 6a49fde.

Applications/Utils/MeshEdit/AddTopLayer.cpp
chleh: No, you didn't overlook anything, only `*result` would be shorter, cf. http://en.cppreference.com/w/cpp/memory/unique_ptr/operator*, thats why I only "noted" (only wanting to annoy people with little details ^^)

Applications/Utils/MeshEdit/AddTopLayer.cpp
```diff
@@ -59,17 +50,23 @@ int main (int argc, char* argv[])
 	cmd.parse(argc, argv);
 
 	INFO("Reading mesh \"%s\" ... ", mesh_arg.getValue().c_str());
-	MeshLib::Mesh * subsfc_mesh(FileIO::readMeshFromFile(mesh_arg.getValue()));
+	auto subsfc_mesh = std::unique_ptr<MeshLib::Mesh>(
+	    FileIO::readMeshFromFile(mesh_arg.getValue()));
```
rinkk: this might be a stupid question but wouldn't it be better to use ```std::unique_ptr<MeshLib::Mesh> subsfc_mesh (...)``` to avoid the new allocation?

chleh: I guess you are referring to the `=`. I think, there is no allocation in this `=` construct. AFAIK if you write something like `Type var = Type(...)` the C++ standard says that this is the same as `Type var(...)`. So this line should automatically behave as you suggest.

chleh: And additionally `unique_ptr` never allocates by itself. If you do
```
std::unique_ptr<T>(new T());
```
then it's the `new` that does the allocation, `unique_ptr` only destroys things. ^^

Applications/Utils/MeshEdit/AddTopLayer.cpp
```diff
@@ -59,17 +50,23 @@ int main (int argc, char* argv[])
...
+	if (!subsfc_mesh) {
+		return EXIT_FAILURE;
```
endJunction: Error message on why it exits would be nice.

Applications/Utils/MeshEdit/AddTopLayer.cpp
```diff
@@ -59,17 +50,23 @@ int main (int argc, char* argv[])
...
+	}
 	INFO("done.");
 
-	std::unique_ptr<MeshLib::Mesh> result (MeshLib::addTopLayerToMesh(
-		*subsfc_mesh, layer_thickness_arg.getValue(), mesh_out_arg.getValue()
-	));
+	std::unique_ptr<MeshLib::Mesh> result(MeshLib::addTopLayerToMesh(
+	    *subsfc_mesh, layer_thickness_arg.getValue(), mesh_out_arg.getValue()));
+	if (! result.get()) {
```
endJunction: `!result`, same as above's `!subsfc_mesh`...

Applications/Utils/MeshEdit/AddTopLayer.cpp
TomFischer: Added error message. The low level reading functions output already error messages.

Issue #987, opened 2016-01-24
Assertion failure in Element test


chleh:
```
[----------] 2 tests from NumLibFunctionInterpolationTest
[ RUN      ] NumLibFunctionInterpolationTest.TwoVariablesTwoNodes
[       OK ] NumLibFunctionInterpolationTest.TwoVariablesTwoNodes (0 ms)
[ RUN      ] NumLibFunctionInterpolationTest.Linear1DElement
testrunner: /usr/include/eigen3/Eigen/src/Core/DenseCoeffsBase.h:378: Eigen::DenseCoeffsBase<Derived, 1>::Scalar& Eigen::DenseCoeffsBase<Derived, 1>::operator[](Eigen::DenseCoeffsBase<Derived, 1>::Index) [with Derived = Eigen::Matrix<double, -1, 1>; Eigen::DenseCoeffsBase<Derived, 1>::Scalar = double; Eigen::DenseCoeffsBase<Derived, 1>::Index = long int]: Assertion `index >= 0 && index < size()' failed.
Abgebrochen (Speicherabzug geschrieben)
```

I compiled it using dynamically allocated Eigen matrices.

endJunction:
@bilke Can we add a CMake configuration option for choosing eigen static/dynamic matrices?
It would require changing between lines 110 and 111 in [ShapeMatrixPolicy.h](https://github.com/ufz/ogs/blob/master/NumLib/Fem/ShapeMatrixPolicy.h#L111).

If possible, then to add a debug build with dynamic matrices to CI.

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/987).

PR #988, opened 2016-01-25
Added CMake option OGS_EIGEN_DYNAMIC for using dynamically allocated eigen matrices

bilke:
See #987.

- [x] Update Jenkins jobs to use fixed-size matrices

### DIFF
--- a/NumLib/Fem/ShapeMatrixPolicy.h
+++ b/NumLib/Fem/ShapeMatrixPolicy.h
@@ -106,10 +106,11 @@ struct EigenDynamicShapeMatrixPolicy
 };
 
 /// Default choice of the ShapeMatrixPolicy.
+#if OGS_EIGEN_DYNAMIC_SHAPE_MATRICES
 template <typename ShapeFunction, unsigned GlobalDim>
-#if EIGEN_DYNAMIC_MATRICES
 using ShapeMatrixPolicyType = EigenDynamicShapeMatrixPolicy<ShapeFunction, GlobalDim>;
 #else
+template <typename ShapeFunction, unsigned GlobalDim>
 using ShapeMatrixPolicyType = EigenFixedShapeMatrixPolicy<ShapeFunction, GlobalDim>;
 #endif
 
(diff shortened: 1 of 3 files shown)


bilke:
Fails as assumed in debug with dynamic eigen matrices.

In general to speed things up in Travis I suggest to test with only one compiler which reduces parallel builds to 3. Testing of the other compiler can then be done on Jenkins.

bilke:
Tried to exclude specific build combinations without success. Excluding seems not to work together with global environment variables...

chleh:
Maybe it's only a matter of taste, but I'd prefer a very verbose name, like `OGS_EIGEN_DYNAMIC_SHAPE_MATRICES`.

NumLib/Fem/ShapeMatrixPolicy.h
```diff
@@ -107,8 +107,11 @@ struct EigenDynamicShapeMatrixPolicy
 
 /// Default choice of the ShapeMatrixPolicy.
 template <typename ShapeFunction, unsigned GlobalDim>
+#if EIGEN_DYNAMIC_MATRICES
```
chleh: Could you pull the `template<...>` line inside the `#ifdef', please. I think, that's more clear.

chleh:
I suggest that pull requests be built with dynamic shape matrices only, since that compiles faster than fixed-size matrices and thereby reduces load on the CI systems. Then, when PRs get merged, there could be builds with fixed-size matrices, too, in order to catch all errors.
What do you think?

chleh:
@bilke when it works in general, I'm fine with it :fast_forward:

bilke:
Any volunteers to fix the crashing eigen test before merging this?

chleh:
Maybe the test really should be fixed before merging this PR. It seems to be the only one that crashes. I can have a look at it by end of this week.

chleh:
@bilke: By your last commit, will there now be any test of fixed-size eigen matrices? I think they have to be tested, too, because: (1) They might use entirely different template functions inside our OGS code, (2) there will be compile-time checks that matrix dimensions in producty match only for them.
In general I agree that we should default to dynamic matrices in order to save much time, but there should be a check of fixed-size matrices, e.g., after a PR has been merged.

bilke:
Yes I will update Jenkins jobs to test fixed-size matrices.

PR #989, opened 2016-01-25
ConfigTreeNew: Added support for XML attributes

chleh:
Said support is necessary in order to be able to process our geometry files with ConfigTreeNew.

This PR actually does many things:
* adds support for XML attributes in `ConfigTreeNew`
* along with that adds `getValue()` methods to access data from XML nodes,
* adds `_filename` member to `ConfigTreeNew` that will be printed in error messages,
* contains some cleanup of `ConfigTreeNew`, and extends its documentation
* adds some functionality to conveniently read an XML file into a `ConfigTreeNew`.

Furthermore, in order to demonstate the handiness of the introduced changes:
* parsing of GML files now happens through `ConfigTreeNew`.
* Therefore the old `ConfigTree` typdefs and functions are not needed anymore and have been removed

The tests for `ConfigTreeNew` have been extended to cover most of the added functionality. However, not everything is tested yet. I want to complete the tests later on, not in this PR.

Sorry for the long PR, but the only thing that can be separated from it easily are the changes in `BoostXmlGmlInterface`. If you think, that it is too much work for you to review it, I'd suggest to make an Q&A session with you about it rather than having to disassemble all of my commits. Unfortunately reviewing commit-wise will not be an option, I'm afraid.

### Suggested order of review (to make it easier for you)
First packet (reading tree from file):
1. Applications/CLI/ogs.cpp
2. BaseLib/ConfigTreeUtil.h
3. BaseLib/ConfigTreeUtil.cpp
Second packet (ConfigTreeNew):
4. BaseLib/ConfigTreeNew.h
5. BaseLib/ConfigTreeNew.cpp
6. BaseLib/ConfigTreeNew-impl.h
Third packet (Tests):
7. Tests/BaseLib/TestConfigTree.cpp
8. Tests/AssemblerLib/TestSerialLinearSolver.cpp (few changes)
9. Tests/MathLib/TestLinearSolver.cpp (few changes)
Fourth packet (GML parsing):
10. FileIO/XmlIO/Boost/BoostXmlGmlInterface.h
11. FileIO/XmlIO/Boost/BoostXmlGmlInterface.cpp
12. BaseLib/ConfigTree.h (only deletions)
13. BaseLib/ConfigTree.cpp (only deletions)

OK, maybe there should have been four PRs. ^^

### DIFF
--- a/BaseLib/ConfigTreeNew-impl.h
+++ b/BaseLib/ConfigTreeNew-impl.h
@@ -144,10 +144,11 @@ ConfigTreeNew::
 getConfAttribute(std::string const& attr) const
 {
     checkUniqueAttr(attr);
-    markVisited<T>(attr, true);
+    auto& ct = markVisited<T>(attr, true, true);
 
     if (auto attrs = _tree->get_child_optional("<xmlattr>")) {
         if (auto a = attrs->get_child_optional(attr)) {
+            ++ct.count; // count only if attribute has been found
             if (auto v = a->get_value_optional<T>()) {
                 return *v;
             } else {
@@ -156,7 +157,7 @@ getConfAttribute(std::string const& attr) const
                       + "' not convertible to the desired type.");
             }
         } else {
-            error("No XML attribute named \"" + attr + "\"");
+            error("Did not find XML attribute with name \"" + attr + "\"");
         }
     } else {
         error("This parameter has no XML attributes");
--- a/BaseLib/ConfigTreeNew-impl.h
+++ b/BaseLib/ConfigTreeNew-impl.h
@@ -9,8 +9,6 @@
 
 #include "ConfigTreeNew.h"
 
-#include <logog/include/logog.hpp>
-
 namespace BaseLib
 {
 
@@ -123,7 +121,6 @@ T
 ConfigTreeNew::
 getValue() const
 {
-    // TODO test this
     if (_have_read_data) {
         error("The data of this subtree has already been read.");
     }
--- a/BaseLib/ConfigTreeNew-impl.h
+++ b/BaseLib/ConfigTreeNew-impl.h
@@ -71,7 +71,7 @@ ConfigTreeNew::
 getConfParamList(std::string const& param) const
 {
     checkUnique(param);
-    markVisited<T>(param, true);
+    markVisited<T>(param, false, true);
 
     auto p = _tree->equal_range(param);
     return Range<ValueIterator<T> >(
@@ -144,7 +144,7 @@ ConfigTreeNew::
 getAttribute(std::string const& attr) const
 {
     checkUniqueAttr(attr);
-    markVisited(attr, true);
+    markVisited<T>(attr, true);
 
     if (auto attrs = _tree->get_child_optional("<xmlattr>")) {
         if (auto a = attrs->get_child_optional(attr)) {
@@ -166,11 +166,13 @@ getAttribute(std::string const& attr) const
 template<typename T>
 ConfigTreeNew::CountType&
 ConfigTreeNew::
-markVisited(std::string const& key, bool const peek_only) const
+markVisited(std::string const& key, bool const is_attr,
+            bool const peek_only) const
 {
     auto const type = std::type_index(typeid(T));
 
-    auto p = _visited_params.emplace(key, CountType{peek_only ? 0 : 1, type});
+    auto p = _visited_params.emplace(std::make_pair(is_attr, key),
+                                     CountType{peek_only ? 0 : 1, type});
 
     if (!p.second) { // no insertion happened
         auto& v = p.first->second;
--- a/BaseLib/ConfigTreeNew-impl.h
+++ b/BaseLib/ConfigTreeNew-impl.h
@@ -71,7 +71,7 @@ ConfigTreeNew::
 getConfParamList(std::string const& param) const
 {
     checkUnique(param);
-    markVisited<T>(param, false, true);
+    markVisited<T>(param, Attr::TAG, true);
 
     auto p = _tree->equal_range(param);
     return Range<ValueIterator<T> >(
@@ -155,7 +155,7 @@ ConfigTreeNew::
 getConfAttributeOptional(std::string const& attr) const
 {
     checkUniqueAttr(attr);
-    auto& ct = markVisited<T>(attr, true, true);
+    auto& ct = markVisited<T>(attr, Attr::ATTR, true);
 
     if (auto attrs = _tree->get_child_optional("<xmlattr>")) {
         if (auto a = attrs->get_child_optional(attr)) {
@@ -176,7 +176,7 @@ getConfAttributeOptional(std::string const& attr) const
 template<typename T>
 ConfigTreeNew::CountType&
 ConfigTreeNew::
-markVisited(std::string const& key, bool const is_attr,
+markVisited(std::string const& key, Attr const is_attr,
             bool const peek_only) const
 {
     auto const type = std::type_index(typeid(T));
--- a/BaseLib/ConfigTreeNew-impl.h
+++ b/BaseLib/ConfigTreeNew-impl.h
@@ -141,7 +141,7 @@ getValue() const
 template<typename T>
 T
 ConfigTreeNew::
-getAttribute(std::string const& attr) const
+getConfAttribute(std::string const& attr) const
 {
     checkUniqueAttr(attr);
     markVisited<T>(attr, true);
--- a/BaseLib/ConfigTreeNew-impl.h
+++ b/BaseLib/ConfigTreeNew-impl.h
@@ -143,6 +143,17 @@ T
 ConfigTreeNew::
 getConfAttribute(std::string const& attr) const
 {
+    if (auto a = getConfAttributeOptional<T>(attr))
+        return *a;
+
+    error("Did not find XML attribute with name \"" + attr + "\".");
+}
+
+template<typename T>
+boost::optional<T>
+ConfigTreeNew::
+getConfAttributeOptional(std::string const& attr) const
+{
     checkUniqueAttr(attr);
     auto& ct = markVisited<T>(attr, true, true);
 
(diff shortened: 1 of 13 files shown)


chleh:
One point to address in the review: I added methods `getConfParam()` etc. to `ConfigTreeNew` that have no template parameters, but there already are methods `getConfParam<>()` with template parameters. While the former actually return a subtree that has no children (hence also the name with `Param`), the latter return the value contained in the subtree converted to the given template type. That might not be the most obvious distinction between names. So upon request I would change the method names somehow.

TomFischer:
The first review packet looks good to me. Will continue review tomorrow.

chleh:
@TomFischer: Thanks for going through all of that.

BaseLib/ConfigTreeNew.h
```diff
@@ -181,13 +225,15 @@ class ConfigTreeNew final
     //! Type of the function objects used as callbacks.
     //! The first argument denotes the path in the tree at which an event (warning/error)
     //! occured, the second argument is the associated message
-    using Callback = std::function<void(const std::string& path,
+    using Callback = std::function<void(const std::string& filename,
```
TomFischer: Docu above needs update.

BaseLib/ConfigTreeNew.h
```diff
@@ -399,17 +578,26 @@ class ConfigTreeNew final
     //! A path printed in error/warning messages.
     std::string _path;
 
-    //! A map key -> (count, type) keeping track which parameters have been read how often
-    //! and which datatype they have.
+    //! The path of the file from which this tree has been read.
+    std::string _filename;
+
+    //! A pair (is attribute, tag/attribute name)
+    using KeyType = std::pair<bool, std::string>;
```
TomFischer: Maybe you can clearify: Is the first component of the pair true if the second component is a *attribute* name or is it true if the second component is a *tag* name?

FileIO/XmlIO/Boost/BoostXmlGmlInterface.cpp
```diff
@@ -58,222 +81,148 @@ bool BoostXmlGmlInterface::readFile(const std::string &fname)
...
-			WARN("BoostXmlGmlInterface::readPoints(): Skipping point, attribute missing in <point> tag:\n%s",
-				BaseLib::propertyTreeToString(point.second).c_str())
-		else
+		auto const p_id = pt.getConfAttribute<std::size_t>("id");
+		auto const p_x  = pt.getConfAttribute<double>("x");
+		auto const p_y  = pt.getConfAttribute<double>("y");
+		auto const p_z  = pt.getConfAttribute<double>("z");
+
+		auto const p_size = points.size();
+		insert_if_key_unique_else_error(_idx_map, p_id, p_size,
```
TomFischer: Nice check. I am curious how is the run time behaviour if we have a huge number of points. We will see ...

FileIO/XmlIO/Boost/BoostXmlGmlInterface.cpp
```diff
@@ -328,17 +277,13 @@ bool BoostXmlGmlInterface::write()
 	addPolylinesToPropertyTree(geometry_set);
 	addSurfacesToPropertyTree(geometry_set);
 
-#if BOOST_VERSION <= 105500
```
TomFischer: :cake:

TomFischer:
Very good for all part: :fast_forward:

FileIO/XmlIO/Boost/BoostXmlGmlInterface.cpp
chleh: I think the runtime behaviour of that call should be equal to the original solution, because there also an `_idx_map.insert()` wwas performed. If there is a huge number of points, then probably the overhead by `ConfigTreeNew` is much more severe. But huge numbers of points should certainly not be stored in plain XML files.

FileIO/XmlIO/Boost/BoostXmlGmlInterface.cpp
```diff
@@ -58,222 +81,148 @@ bool BoostXmlGmlInterface::readFile(const std::string &fname)
...
+    std::vector<std::size_t> const& pnt_id_map,
+    std::map<std::string, std::size_t>& ply_names)
 {
-	BOOST_FOREACH( BaseLib::ConfigTree::value_type const & polyline, polylinesRoot )
+	for (auto const pl : polylinesRoot.getConfSubtreeList("polyline"))
 	{
-		if (polyline.first.compare("polyline") != 0)
-			continue;
+		auto const id = pl.getConfAttribute<std::size_t>("id");
+		(void) id; // id not used
```
chleh: TODO for myself: The id is not used but it must be there in the gml file. That's why `ignore...()` cannot be used.

FileIO/XmlIO/Boost/BoostXmlGmlInterface.cpp
chleh: Thank you, but that's only because now OGS6 requires boost >= 1.56 ^^

FileIO/XmlIO/Boost/BoostXmlGmlInterface.cpp
TomFischer: We have to discuss if the id is necessary anymore. @rinkk your opinion?

BaseLib/ConfigTreeUtil.h
```diff
@@ -0,0 +1,86 @@
...
+     * \param be_ruthless if true, then warnings will raise errors, .i.e. lead to program abortion,
+     *                    else warnings will only warn
+     * \param ptree the underlying ptree of the created ConfigTree
+     */
+    explicit ConfigTreeTopLevel(std::string const& filepath,
+                                bool const be_ruthless,
+                                ConfigTreeNew::PTree&& ptree);
+
+    //! Access the contained ConfigTree.
+    ConfigTreeNew const& operator*() const;
```
endJunction: Is the non-const dereference operator disable with purpose or by accident?
Even if not required now, I'd complete the interface.

BaseLib/ConfigTreeUtil.h
```diff
@@ -0,0 +1,86 @@
...
+};
+
+/*! Create a ConfigTree from an XML file.
+ *
+ * \param filepath     see ConfigTreeTopLevel::ConfigTreeTopLevel()
+ * \param be_ruthless  see ConfigTreeTopLevel::ConfigTreeTopLevel()
+ * \param toplevel_tag name of the outermost tag in the XML file. The returned ConfigTree is rooted
+ *                     one level below that tag.
+ *
+ * The parameter \c toplevel_tag is provided for compatibility with our exeisting configuration
```
endJunction: typo: existing

BaseLib/ConfigTreeUtil.h
```diff
@@ -0,0 +1,86 @@
...
+ * ConfigTree. Via that parameter the naming rules do not apply to the toplevel tag.
+ *
+ * Unfortunately the XML parser shipped with <tt>boost::property_tree</tt> does not fully support
+ * the XML standard. Additionally there might be encoding issues. From their docs:
+ *
+ * > Please note that RapidXML does not understand the encoding specification. If you pass it a
+ * > character buffer, it assumes the data is already correctly encoded; if you pass it a filename,
+ * > it will read the file using the character conversion of the locale you give it (or the global
+ * > locale if you give it none). This means that, in order to parse a UTF-8-encoded XML file into
+ * > a wptree, you have to supply an alternate locale, either directly or by replacing the global one.
```
endJunction: Is "wptree" a typo or a type?

FileIO/XmlIO/Boost/BoostXmlGmlInterface.cpp
```diff
@@ -328,17 +279,13 @@ bool BoostXmlGmlInterface::write()
...
-#else
 	boost::property_tree::xml_writer_settings<std::string> settings('\t', 1);
-#endif  // BOOST_VERSION
 	write_xml(_out, pt, settings);
 	return true;
 }
 
 void BoostXmlGmlInterface::addSurfacesToPropertyTree(
-	BaseLib::ConfigTree & geometry_set)
+	boost::property_tree::ptree & geometry_set)
```
endJunction: The ptree is used here because the caller of `addSurfaces..` does not understand `ConfigTreeNew`, isn't it? Or maybe I plain wrong...

FileIO/XmlIO/Boost/BoostXmlGmlInterface.cpp
```diff
@@ -33,6 +25,38 @@
 #include "GeoLib/Surface.h"
 #include "GeoLib/Triangle.h"
 
+
+namespace
+{
+
+//! Method for handling conversion to string uniformly
```
endJunction: I'd add "... to string uniformly across all types and std::string; see std::string overload below."

BaseLib/ConfigTreeNew-impl.h
```diff
@@ -133,15 +116,70 @@ checkConfParam(std::string const& param, Ch const* value) const
...
+
+    return boost::none;
+}
 
 template<typename T>
 ConfigTreeNew::CountType&
 ConfigTreeNew::
-markVisited(std::string const& key, bool peek_only) const
+markVisited(std::string const& key, bool const is_attr,
+            bool const peek_only) const
```
endJunction: Maybe it is better to introduce an enum for `is_attr` parameter with the values like `ATTR` and `PARAM`. This would result in better code where `markVisited` is used, _e.g._:
```c++
auto& ct = markVisited<T>(attr, ATTR, true); //line 155 above
markVisited<T>(param, PARAM, true); // line 72 above
```
The implementation would not change when choosing ATTR=1 and PARAM=0. And it should result in identical assembler code.

BaseLib/ConfigTreeUtil.h
chleh: Initially it was by purpose, but when writing `operator->()` I forgot about it. ^^
Actually I'd rather remove the non-const `operator->()` and add a comment. The idea of all of this is to prevent that the user calls `checkAndInvalidate()` on a passed around instance of the `ConfigTreeNew`, instead the respective method of this wrapper should be used.

BaseLib/ConfigTreeUtil.h
chleh: A typo type ^^

FileIO/XmlIO/Boost/BoostXmlGmlInterface.cpp
chleh: Do not understand and never will understand since `ConfigTreeNew` is in a sense read-only by design.

Issue #983, opened 2016-01-22
DE: AddLayer (bottom)


TomFischer:
The node ordering of the elements added as bottom layer seems to be wrong.

![addlayerbottom](https://cloud.githubusercontent.com/assets/800065/12506878/51c3f780-c0f2-11e5-9204-784266b9873a.png)

rinkk:
This was solved with PR #990

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/983).

PR #990, opened 2016-01-25
Element flipping

rinkk:
adds a function to flip elements (i.e. reverse direction of 1d elems, change sign of normals for 2d elements). this solves #983.

### DIFF
--- a/MeshLib/MeshEditing/AddLayerToMesh.cpp
+++ b/MeshLib/MeshEditing/AddLayerToMesh.cpp
@@ -82,11 +82,10 @@ MeshLib::Mesh* addLayerToMesh(MeshLib::Mesh const& mesh, double thickness,
 	std::unique_ptr<MeshLib::Mesh> sfc_mesh (nullptr);
 	
 	if (mesh.getDimension() == 3)
-		sfc_mesh = std::unique_ptr<MeshLib::Mesh>(
-			MeshLib::MeshSurfaceExtraction::getMeshSurface(mesh, dir, angle, true));
+		sfc_mesh.reset(MeshLib::MeshSurfaceExtraction::getMeshSurface(mesh, dir, angle, true));
 	else
 		sfc_mesh = (on_top) ? std::unique_ptr<MeshLib::Mesh>(new MeshLib::Mesh(mesh)) :
-		                      std::unique_ptr<MeshLib::Mesh>(MeshLib::flipMeshElements(mesh));
+		                      std::unique_ptr<MeshLib::Mesh>(MeshLib::createFlippedMesh(mesh));
 	INFO("done.");
 
 	// *** add new surface nodes
--- a/MeshLib/MeshEditing/AddLayerToMesh.cpp
+++ b/MeshLib/MeshEditing/AddLayerToMesh.cpp
@@ -24,6 +24,7 @@
 #include "MeshLib/Elements/Elements.h"
 #include "MeshLib/MeshSurfaceExtraction.h"
 #include "MeshLib/MeshEditing/DuplicateMeshComponents.h"
+#include "MeshLib/MeshEditing/FlipElements.h"
 
 namespace MeshLib
 {
@@ -78,10 +79,14 @@ MeshLib::Mesh* addLayerToMesh(MeshLib::Mesh const& mesh, double thickness,
 	int const flag = (on_top) ? -1 : 1;
 	const MathLib::Vector3 dir(0, 0, flag);
 	double const angle(90);
-	std::unique_ptr<MeshLib::Mesh> sfc_mesh = (mesh.getDimension() == 3) ?
-		std::unique_ptr<MeshLib::Mesh>(
-			MeshLib::MeshSurfaceExtraction::getMeshSurface(mesh, dir, angle, true)) :
-		std::unique_ptr<MeshLib::Mesh>(new MeshLib::Mesh(mesh));
+	std::unique_ptr<MeshLib::Mesh> sfc_mesh (nullptr);
+	
+	if (mesh.getDimension() == 3)
+		sfc_mesh = std::unique_ptr<MeshLib::Mesh>(
+			MeshLib::MeshSurfaceExtraction::getMeshSurface(mesh, dir, angle, true));
+	else
+		sfc_mesh = (on_top) ? std::unique_ptr<MeshLib::Mesh>(new MeshLib::Mesh(mesh)) :
+		                      std::unique_ptr<MeshLib::Mesh>(MeshLib::flipMeshElements(mesh));
 	INFO("done.");
 
 	// *** add new surface nodes
--- a/MeshLib/MeshEditing/FlipElements.cpp
+++ b/MeshLib/MeshEditing/FlipElements.cpp
@@ -19,35 +19,30 @@
 namespace MeshLib
 {
 
-MeshLib::Element* flipElement(MeshLib::Element const& elem, std::vector<MeshLib::Node*> const& nodes)
+std::unique_ptr<MeshLib::Element> createFlippedElement(MeshLib::Element const& elem, std::vector<MeshLib::Node*> const& nodes)
 {
 	if (elem.getDimension()>2)
 		return nullptr;
 
-	std::size_t const n_nodes (elem.getNNodes());
+	unsigned const n_nodes (elem.getNNodes());
 	MeshLib::Node** elem_nodes = new MeshLib::Node*[n_nodes];
 	for (unsigned i=0; i<n_nodes; ++i)
 		elem_nodes[i] = nodes[elem.getNode(i)->getID()];
-	MeshLib::Node* tmp (elem_nodes[0]);
-	elem_nodes[0] = elem_nodes[1];
-	elem_nodes[1] = tmp;
-
+	std::swap(elem_nodes[0], elem_nodes[1]);
+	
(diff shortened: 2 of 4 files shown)


MeshLib/MeshEditing/FlipElements.h
```diff
@@ -0,0 +1,36 @@
...
+namespace MeshLib
+{
+class Mesh;
+class Node;
+
+/**
+ * Reverses the node order of a 1D / 2D element such that the direction 
+ * of a line changes and normals of 2D elements changes its sign.
+ */
+MeshLib::Element* flipElement(MeshLib::Element const& elem, std::vector<MeshLib::Node*> const& nodes);
```
TomFischer: Why do you need the second argument? If my understanding from the documentation is correct, the sequence of nodes of the element is changed only, i.e. no new nodes should be necessary.

MeshLib/MeshEditing/FlipElements.cpp
```diff
@@ -0,0 +1,69 @@
...
+	if (elem.getDimension()>2)
+		return nullptr;
+
+	std::size_t const n_nodes (elem.getNNodes());
+	MeshLib::Node** elem_nodes = new MeshLib::Node*[n_nodes];
+	for (unsigned i=0; i<n_nodes; ++i)
+		elem_nodes[i] = nodes[elem.getNode(i)->getID()];
+	MeshLib::Node* tmp (elem_nodes[0]);
+	elem_nodes[0] = elem_nodes[1];
+	elem_nodes[1] = tmp;
```
TomFischer: Why not use `std::swap()`?

MeshLib/MeshEditing/FlipElements.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+	if (elem.getGeomType() == MeshElemType::LINE)
+		return new MeshLib::Line(elem_nodes, elem.getID());
+	else if (elem.getGeomType() == MeshElemType::TRIANGLE)
+		return new MeshLib::Tri(elem_nodes, elem.getID());
+	else if (elem.getGeomType() == MeshElemType::QUAD)
+	{
+		tmp = elem_nodes[2];
+		elem_nodes[2] = elem_nodes[3];
+		elem_nodes[3] = tmp;
```
TomFischer: Maybe you could `std::swap()`?

MeshLib/MeshEditing/FlipElements.cpp
```diff
@@ -0,0 +1,69 @@
...
+		return new MeshLib::Quad(elem_nodes, elem.getID());
+	}
+
+	return nullptr;
+}
+
+MeshLib::Mesh* flipMeshElements(MeshLib::Mesh const& mesh)
```
TomFischer: I think either the input mesh or the output mesh are inconsistent concerning the node ordering in elements. I assume the elements of the input mesh has wrong node ordering. Why not fix this in the mesh itself?

MeshLib/MeshEditing/AddLayerToMesh.cpp
```diff
@@ -77,10 +78,14 @@ MeshLib::Mesh* addLayerToMesh(MeshLib::Mesh const& mesh, double thickness,
...
 	double const angle(90);
-	std::unique_ptr<MeshLib::Mesh> sfc_mesh = (mesh.getDimension() == 3) ?
-		std::unique_ptr<MeshLib::Mesh>(
-			MeshLib::MeshSurfaceExtraction::getMeshSurface(mesh, dir, angle, true)) :
-		std::unique_ptr<MeshLib::Mesh>(new MeshLib::Mesh(mesh));
+	std::unique_ptr<MeshLib::Mesh> sfc_mesh (nullptr);
+	
+	if (mesh.getDimension() == 3)
+		sfc_mesh = std::unique_ptr<MeshLib::Mesh>(
+			MeshLib::MeshSurfaceExtraction::getMeshSurface(mesh, dir, angle, true));
```
chleh: Only a note: You could also simply do `sfc_mesh.reset(...)`.

MeshLib/MeshEditing/FlipElements.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+	std::size_t const n_nodes (elem.getNNodes());
+	MeshLib::Node** elem_nodes = new MeshLib::Node*[n_nodes];
+	for (unsigned i=0; i<n_nodes; ++i)
+		elem_nodes[i] = nodes[elem.getNode(i)->getID()];
+	MeshLib::Node* tmp (elem_nodes[0]);
+	elem_nodes[0] = elem_nodes[1];
+	elem_nodes[1] = tmp;
+
+	if (elem.getGeomType() == MeshElemType::LINE)
```
chleh: If you used switch, there would be warnings generated if at some point in time somebody extends the enum.

MeshLib/MeshEditing/FlipElements.cpp
```diff
@@ -0,0 +1,69 @@
...
+namespace MeshLib
+{
+
+MeshLib::Element* flipElement(MeshLib::Element const& elem, std::vector<MeshLib::Node*> const& nodes)
+{
+	if (elem.getDimension()>2)
+		return nullptr;
+
+	std::size_t const n_nodes (elem.getNNodes());
+	MeshLib::Node** elem_nodes = new MeshLib::Node*[n_nodes];
```
chleh: Only a note: In general I would prefer some `unique_ptr`. You can call `release()` on it if ownership is passed on. Sure, this piece of code is rather simple, but in other occasions it might help to prevent memory leaks.

MeshLib/MeshEditing/FlipElements.cpp
```diff
@@ -0,0 +1,69 @@
...
+	for (unsigned i=0; i<n_nodes; ++i)
```
chleh: Probably `n_nodes` (being `size_t`) will never break the limits of `unsigned`.

MeshLib/MeshEditing/FlipElements.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include "MeshLib/Node.h"
+#include "MeshLib/Elements/Line.h"
+#include "MeshLib/Elements/Tri.h"
+#include "MeshLib/Elements/Quad.h"
+#include "DuplicateMeshComponents.h"
+
+namespace MeshLib
+{
+
+MeshLib::Element* flipElement(MeshLib::Element const& elem, std::vector<MeshLib::Node*> const& nodes)
```
chleh: I'd also like it more to return `unique_ptr` if memory must be managed.

MeshLib/MeshEditing/FlipElements.cpp
```diff
@@ -0,0 +1,69 @@
...
+	MeshLib::Node** elem_nodes = new MeshLib::Node*[n_nodes];
+	for (unsigned i=0; i<n_nodes; ++i)
+		elem_nodes[i] = nodes[elem.getNode(i)->getID()];
+	MeshLib::Node* tmp (elem_nodes[0]);
+	elem_nodes[0] = elem_nodes[1];
+	elem_nodes[1] = tmp;
+
+	if (elem.getGeomType() == MeshElemType::LINE)
+		return new MeshLib::Line(elem_nodes, elem.getID());
+	else if (elem.getGeomType() == MeshElemType::TRIANGLE)
```
chleh: Why don't you flip triangles?

chleh:
Only very minor things from my side :+1:

MeshLib/MeshEditing/FlipElements.cpp
rinkk: discussed, both meshes are correct.

MeshLib/MeshEditing/FlipElements.h
rinkk: discussed, nodes are necessary.

MeshLib/MeshEditing/FlipElements.cpp
rinkk: not really, there are plenty of element types we are not considering here (i.e. all 3d elements)

chleh:
`std::make_unique` is a C++14 feature.
Unfortunately we have to resort to manually creating `unique_ptr`s for now.

rinkk:
thx @chleh

Tests/MeshLib/TestFlipElements.cpp
```diff
@@ -0,0 +1,77 @@
...
+	ASSERT_EQ(mesh->getNElements(), result->getNElements());
+	for (std::size_t i=0; i<result->getNElements(); ++i)
+	{
+		ASSERT_TRUE(mesh->getElement(i)->getNode(0)->getID() == 
+		            result->getElement(i)->getNode(1)->getID());
+		ASSERT_TRUE(mesh->getElement(i)->getNode(1)->getID() == 
+		            result->getElement(i)->getNode(0)->getID());
+		ASSERT_TRUE(mesh->getElement(i)->getNode(2)->getID() == 
+		            result->getElement(i)->getNode(3)->getID());
+		ASSERT_TRUE(mesh->getElement(i)->getNode(3)->getID() ==
```
endJunction: Using `ASSERT_EQ` would generate better error messages showing both values...

endJunction:
Another good test would be to flip the element/mesh twice, which should be identity operation.

rinkk:
added double flipping test

Issue #991, opened 2016-01-26
GMSH crash


TomFischer:
Erik discovered, that generating a gmsh mesh using the gui crashes.

TomFischer:
Fixed with PR #994.

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/991).

PR #994, opened 2016-01-26
[FileIO] GMSH

TomFischer:
Insert points using a valid/correct id used for the visualization of the Quadtree. Before the ids were invalid which leads to crash the DataExplorer (see issue #991).

### DIFF
--- a/FileIO/GmshIO/GMSHAdaptiveMeshDensity.cpp
+++ b/FileIO/GmshIO/GMSHAdaptiveMeshDensity.cpp
@@ -123,7 +123,9 @@ void GMSHAdaptiveMeshDensity::getSteinerPoints (std::vector<GeoLib::Point*> & pn
 			const double delta ((ur[0] - ll[0]) / (2 * n_pnts_per_quad_dim));
 			for (std::size_t i(0); i<n_pnts_per_quad_dim; i++) {
 				for (std::size_t j(0); j<n_pnts_per_quad_dim; j++) {
-					pnts.push_back(new GeoLib::Point (ll[0] + (2*i+1) * delta, ll[1] + (2*j+1) * delta, 0.0));
+					pnts.push_back(new GeoLib::Point(
+					    ll[0] + (2 * i + 1) * delta,
+					    ll[1] + (2 * j + 1) * delta, 0.0, pnts.size()));
 				}
 			}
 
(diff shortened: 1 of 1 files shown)


chleh:
Whatever this code does, your changes look fine :+1:

FileIO/GmshIO/GMSHAdaptiveMeshDensity.cpp
```diff
@@ -123,7 +123,9 @@ void GMSHAdaptiveMeshDensity::getSteinerPoints (std::vector<GeoLib::Point*> & pn
 			const double delta ((ur[0] - ll[0]) / (2 * n_pnts_per_quad_dim));
 			for (std::size_t i(0); i<n_pnts_per_quad_dim; i++) {
 				for (std::size_t j(0); j<n_pnts_per_quad_dim; j++) {
-					pnts.push_back(new GeoLib::Point (ll[0] + (2*i+1) * delta, ll[1] + (2*j+1) * delta, 0.0));
+					pnts.push_back(new GeoLib::Point(
+					    ll[0] + (2 * i + 1) * delta,
+					    ll[1] + (2 * j + 1) * delta, 0.0, pnts.size()));
```
rinkk: shouldn't this be ```pnts.size()-1``` ?

FileIO/GmshIO/GMSHAdaptiveMeshDensity.cpp
```diff
@@ -141,13 +143,13 @@ void GMSHAdaptiveMeshDensity::getQuadTreeGeometry(std::vector<GeoLib::Point*> &p
...
-		(*it)->getSquarePoints(*ll, *ur);
-		std::size_t pnt_offset (pnts.size());
-		pnts.push_back(ll);
-		pnts.push_back(new GeoLib::Point((*ur)[0], (*ll)[1], 0.0));
-		pnts.push_back(ur);
-		pnts.push_back(new GeoLib::Point((*ll)[0], (*ur)[1], 0.0));
+		GeoLib::Point ll, ur;
+		(*it)->getSquarePoints(ll, ur);
+		std::size_t const pnt_offset (pnts.size());
+		pnts.push_back(new GeoLib::Point(ll, pnt_offset));
```
rinkk: not sure but if you don't correct the idx in my other comment you might end up with two points having idx ```pnt_offset``` (i might be wrong, though, i don't know how these two methods interact / in which sequence they are called.

FileIO/GmshIO/GMSHAdaptiveMeshDensity.cpp
chleh: I guess it works, because `pnts.size()` is evaluated before `push_back()`, so it still returns the old size. Anyway, it might be clearer to introduce an auxiliary variable, like `new_point_id`, that is set before the `push_back`.

TomFischer: @chleh is right. I would introduce an  auxiliary variable if the `id` would be used at several places.

FileIO/GmshIO/GMSHAdaptiveMeshDensity.cpp
TomFischer: I think the code is correct so far. See the answer to your first comment.

PR #998, opened 2016-01-27
[PL] Move up process' output. Split post and out.

endJunction:
No real code changes here, just moving of the output.

_Next steps:_ I'll introduce number of components in the ProcessVariable and extend the output to all of the process variables available in the Process.


TomFischer:
:+1:

PR #1000, opened 2016-01-28
Data submodule instead of CMake External Data

bilke:
This PR ditches the [CMake External Data](https://docs.opengeosys.org/docs/devguide/testing/test-data#How%20it%20works) stuff due to complicated handling and instead uses a git submodule in `Tests/Data` to provide test files. It will get checked out automatically when `OGS_BUILD_TESTS=ON`.

I also added an option (`-o [dir]`) to the cli to specify an output directory, see af0f0dc8. Should the output directory be stored somewhere for later use? At the moment it is simply passed to `solveProcesses()`.

The [data submodule](https://github.com/ufz/ogs-data) is a normal git repository (I also tried git-lfs but is slower when downloading / uploading and it is one dependency more because it has to be installed in addition to git). We just fill it with data and see if we hit some performance problems in the future...

Also unit tests can makes use of input files now. You can use `data_path` pointing `[source-dir/Tests/Data]`. For writing output files you can use `data_binary_path` pointing to `[build-dir]/Tests/Data` (see BuildInfo.h).

- [x] Update docs
- [ ] Fix MPI ctests


bilke:
:tropical_drink:  Whohoo! I landed the 1000th pull request!! :tropical_drink:

wenqing:
:clap:

BaseLib/FileTools.cpp
```diff
@@ -133,4 +133,18 @@ std::string extractPath(std::string const& pathname)
...
+const char pathSeparator =
+#ifdef _WIN32
+                            '\\';
+#else
+                            '/';
+#endif
+
+std::string appendPathSeparator(std::string const& path)
+{
+	if(findLastPathSeparator(path) == path.length() - 1)
```
chleh: If `path` is empty, the check will true, as far as i can see. That's good, but not obvious.
In general I'd prefer a method like `join(path1, path2)` to this one, because that would make sure that there are (back)slashes inserted appropriately whenever you use it.
If you don't agree, please at least document the behaviour for empty strings. In particular since that is the reason why OGS does not write to `/` when the new `-o` option is omitted.

Tests/Data/Elliptic/cube_1x1x1_GroundWaterFlow/cube_4e4_neumann.prj.md5
```diff
@@ -1 +0,0 @@
-38284d5144b7718309286a39725d9193
```
chleh: Can't read that without brute-forcing for a hash collision. ^^

chleh:
:joy:

chleh:
Btw: There was that discussion about whether the CMake scripts should get the whole history of the test data repo by default. Do they, or do they restrict the history to, e.g., the most recent commit?

bilke:
Unfortunately a shallow clone on submodules is not an option because when you clone e.g. only the last commit of a repo but the submodule points to another commit of that repo then it breaks...

bilke:
@chleh Unfortunately it is not possible to use shallow clones for submodules: if you just clone the most recent commit from a repos default branch and your submodule points to another commit in that repo things break...

chleh:
@bilke Have you considered this: http://stackoverflow.com/a/17692710
Maybe that solution is only applicable if done manually by the user. I don't know, I ain't a submodule guru.

TomFischer:
:fast_forward:

bilke:
@chleh Yes I have seen this but it does not help.

bilke:
@wenqing `ParallelFEM_GroundWaterFlow3D_NeumannBC` fails on file comparison (see almost the end of the log):

https://svn.ufz.de:8443/job/OGS-6/job/Linux-PRs/1147/console

Is the result still correct? If yes then I update the reference file.

endJunction:
Looks good and will definitely simplify updating of the tests. Hopefully it will not stress the github traffic too much.

:fast_forward: after the petsc neumann test is fixed by Wenqing.

wenqing:
@bilke  I've copied the results to directory of  ``/data/envinf/results_cube_1e3_neumann_pcs``. You can copy the data from the directory and remove it late on.

bilke:
Thanks! Please note that the correct name for reference files is `[original_file_name]_expected.vtu`.

PR #1002, opened 2016-01-28
Rework MeshSurfaceExtraction::getSurfaceMesh

TomFischer:
Copy subsurface node ids to a MeshLib::PropertyVector of the surface mesh.

### DIFF
--- a/Applications/DataExplorer/DataView/DirectConditionGenerator.cpp
+++ b/Applications/DataExplorer/DataView/DirectConditionGenerator.cpp
@@ -13,6 +13,7 @@
  */
 
 #include <fstream>
+#include <memory>
 
 // ThirdParty/logog
 #include "logog/include/logog.hpp"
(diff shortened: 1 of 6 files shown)


Applications/DataExplorer/DataView/DirectConditionGenerator.cpp
```diff
@@ -61,33 +62,50 @@ const std::vector< std::pair<std::size_t,double> >& DirectConditionGenerator::di
...
+	std::vector<double> node_area_vec =
+		MeshLib::MeshSurfaceExtraction::getSurfaceAreaForNodes(*surface_mesh);
+	const std::vector<MeshLib::Node*> &surface_nodes(surface_mesh->getNodes());
+	const std::size_t nNodes(surface_mesh->getNNodes());
+	const double no_data(raster->getNoDataValue());
+
+	boost::optional<MeshLib::PropertyVector<int> const&> opt_node_id_pv(
+	    surface_mesh->getProperties().getPropertyVector<int>(prop_name));
+	if (!opt_node_id_pv) {
+		ERR("Need subsurface node id, but such a property is not available.")
```
chleh: Maybe it's better to print the concrete property name "OriginalSubsurfaceNodeIDs" in the error message.

Applications/DataExplorer/DataView/DirectConditionGenerator.cpp
```diff
@@ -61,33 +62,50 @@ const std::vector< std::pair<std::size_t,double> >& DirectConditionGenerator::di
...
+		return _direct_values;
```
chleh: Just wondering: Whenever an error occurs, `_direct_values` will be returned immediately, which looks like a valid return value. Is that return value meaningful in the case of an error? Is there any indicator in the DE for the user that something went wrong? Or is such an indicator unnecessary?

MeshLib/MeshSurfaceExtraction.h
```diff
@@ -45,12 +45,20 @@ class MeshSurfaceExtraction
...
-	 * \param angle               The angle of the allowed deviation from the given direction (0 <= angle <= 90 degrees)
-	 * \param keepOriginalNodeIds If true, ids of mesh nodes are set to ids in original mesh, otherwise node ids are reset (as usual when creating a mesh)
-	 * \return                    A 2D mesh representing the surface in direction dir
+	 * \param dir                 The direction in which face normals have to
+	 * point to be considered surface elements
+	 * \param angle               The angle of the allowed deviation from the
+	 * given direction (0 <= angle <= 90 degrees)
+	 * \param prop_name If not empty, ids of nodes of the original subsurface
+	 * mesh are stored in a \c MeshLib::PropertyVector with the name stored in
+	 * \c prop_name
```
chleh: And if empty original ids are dropped? If so:
* Please document that.
* Then I'd prefer a more verbose argument name, like `subsfc_nodeid_backup_prop`.
If not, then what? ^^

MeshLib/MeshEditing/AddLayerToMesh.cpp
```diff
@@ -31,6 +32,7 @@ namespace MeshLib
 
 MeshLib::Element* extrudeElement(std::vector<MeshLib::Node*> const& subsfc_nodes,
 	MeshLib::Element const& sfc_elem,
+	MeshLib::PropertyVector<int> const& sfc_to_subsfc_id_map,
```
chleh: The subsequent argument is a `std::map<std::size_t, std::size_t>`, why not also `size_t` here? Probably that doesn't matter so much, but maybe there should be only one index type.

chleh: Can you document this method, even though it is in an anonymous namespace? The purpose can be guessed from the name, but not necessarily the purpose of the arguments.

MeshLib/MeshEditing/AddLayerToMesh.cpp
```diff
@@ -80,12 +83,27 @@ MeshLib::Mesh* addLayerToMesh(MeshLib::Mesh const& mesh, double thickness,
...
-		sfc_mesh.reset(MeshLib::MeshSurfaceExtraction::getMeshSurface(mesh, dir, angle, true));
-	else
+		sfc_mesh.reset(MeshLib::MeshSurfaceExtraction::getMeshSurface(
+		    mesh, dir, angle, prop_name));
+	else {
 		sfc_mesh = (on_top) ? std::unique_ptr<MeshLib::Mesh>(new MeshLib::Mesh(mesh)) :
 		                      std::unique_ptr<MeshLib::Mesh>(MeshLib::createFlippedMesh(mesh));
+		// add property storing node ids
+		boost::optional<MeshLib::PropertyVector<int>&> pv(
+		    sfc_mesh->getProperties().createNewPropertyVector<int>(
```
chleh: Also the `size_t` question here.

MeshLib/MeshEditing/AddLayerToMesh.cpp
```diff
@@ -80,12 +83,27 @@ MeshLib::Mesh* addLayerToMesh(MeshLib::Mesh const& mesh, double thickness,
...
+		        prop_name, MeshLib::MeshItemType::Node, 1));
+		if (pv) {
+			pv->resize(sfc_mesh->getNNodes());
+			std::iota(pv->begin(), pv->end(), 0);
+		} else {
+			ERR("Could not create and initialize property.");
```
chleh: Also "OriginalSubsurfaceNodeIDs" into the error message.

MeshLib/MeshEditing/AddLayerToMesh.cpp
```diff
@@ -99,24 +117,33 @@ MeshLib::Mesh* addLayerToMesh(MeshLib::Mesh const& mesh, double thickness,
 	std::vector<MeshLib::Node*> const& sfc_nodes(sfc_mesh->getNodes());
 	std::size_t const n_sfc_nodes(sfc_nodes.size());
 
+	// fetch subsurface node ids PropertyVector
+	boost::optional<MeshLib::PropertyVector<int> const&> opt_node_id_pv(
+	    sfc_mesh->getProperties().getPropertyVector<int>(prop_name));
+	if (!opt_node_id_pv) {
+		ERR("Need subsurface node ids, but such a property is not available.")
```
chleh: Also "OriginalSubsurfaceNodeIDs" into the error message.

chleh:
Some questions but afterwards :+1:

Applications/DataExplorer/DataView/DirectConditionGenerator.cpp
TomFischer: This behaviour was implemented already. I hope I did not changed this. To be honest: I don't know how this is handled in the DataExplorer. I touched the code only because it uses `getMeshSurface()`.

Applications/DataExplorer/DataView/DirectConditionGenerator.cpp
TomFischer: Fixed in commit f1ba0866fa2608fd8ce89a760f8f5559b78f16fd.

MeshLib/MeshSurfaceExtraction.h
TomFischer: See commit 9712bd762ceb81778303c6585484586ffe15c37e.

MeshLib/MeshEditing/AddLayerToMesh.cpp
TomFischer: See commit 9712bd762ceb81778303c6585484586ffe15c37e for type change (`int` -> `size_t`) and 9712bd7 for documentation.

MeshLib/MeshSurfaceExtraction.cpp
```diff
@@ -107,17 +109,23 @@ MeshLib::Mesh* MeshSurfaceExtraction::getMeshSurface(const MeshLib::Mesh &mesh,
...
-			(*node)->setID(id_map[(*node)->getID()]);
-
+	// transmit the original node ids of the subsurface mesh as a property
+	if (!subsfc_node_id_backup_prop_name.empty()) {
+		boost::optional<MeshLib::PropertyVector<std::size_t>&> orig_node_ids(
+		    result->getProperties().createNewPropertyVector<std::size_t>(
+		        subsfc_node_id_backup_prop_name , MeshLib::MeshItemType::Node, 1));
+		if (orig_node_ids) {
+			orig_node_ids->resize(id_map.size());
+			std::copy(id_map.cbegin(), id_map.cend(), orig_node_ids->begin());
```
endJunction: reserve and back_inserter would safe initialization...

endJunction:
:fast

Issue #965, opened 2016-01-19
Improvement in ogs#PETSc


wenqing:
- Create vector with ghost entries.
- Use local size to create matrix.

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/965).

PR #1006, opened 2016-01-29
Improvement in the usage of PETSc matrix and vector

wenqing:
- Create vector with ghost entries

     To save memory when the solution is fetched.
- Use local size to create vector and matrix

    To improve the computational and communication load balance in the global assembly and the linear solver.

(c.f. issue #965)

### DIFF
--- a/AssemblerLib/LocalToGlobalIndexMap.h
+++ b/AssemblerLib/LocalToGlobalIndexMap.h
@@ -103,7 +103,7 @@ public:
     }
 
     /// Get ghost indices, forwarded from MeshComponentMap.
-    std::vector<GlobalIndexType> getGhostIndices() const
+    std::vector<GlobalIndexType> const& getGhostIndices() const
     {
         return _mesh_component_map.getGhostIndices();
     }
--- a/AssemblerLib/LocalToGlobalIndexMap.h
+++ b/AssemblerLib/LocalToGlobalIndexMap.h
@@ -68,7 +68,9 @@ public:
     /// Returns total number of degrees of freedom.
     std::size_t dofSize() const;
 
-    /// Returns total number of local degrees of freedom for DDC.
+    /// Returns total number of local degrees of freedom
+    /// of the present rank, which does not count the unknowns
+    /// associated with ghost nodes (for DDC with node-wise mesh partitioning).
     std::size_t dofSizeLocal() const
     {
         return _mesh_component_map.getNLocalUnknowns();
--- a/AssemblerLib/LocalToGlobalIndexMap.h
+++ b/AssemblerLib/LocalToGlobalIndexMap.h
@@ -68,6 +68,12 @@ public:
     /// Returns total number of degrees of freedom.
     std::size_t dofSize() const;
 
+    /// Returns total number of local degrees of freedom for DDC.
+    std::size_t dofSizeLocal() const
+    {
+        return _mesh_component_map.getNLocalUnknowns();
+    }
+
     /// Returns total number of global degrees of freedom for DDC.
     std::size_t dofSizeGlobal() const
     {
@@ -92,6 +98,12 @@ public:
     std::vector<GlobalIndexType> getGlobalIndices(const MeshLib::Location &l) const
     {
         return _mesh_component_map.getGlobalIndices(l);
+    }
+
+    /// Get ghost indices, forwarded from MeshComponentMap.
+    std::vector<GlobalIndexType> getGhostIndices() const
+    {
+        return _mesh_component_map.getGhostIndices();
     }
 
 private:
--- a/AssemblerLib/MeshComponentMap.cpp
+++ b/AssemblerLib/MeshComponentMap.cpp
@@ -40,6 +40,7 @@ MeshComponentMap::MeshComponentMap(
     std::size_t comp_id = 0;
 
     _num_global_dof = 0;
+    _num_local_dof = 0;
     for (auto c = components.cbegin(); c != components.cend(); ++c)
     {
         assert (*c != nullptr);
@@ -57,8 +58,18 @@ MeshComponentMap::MeshComponentMap(
                 {
                     const GlobalIndexType global_id = static_cast<GlobalIndexType>(components.size()
                                                * mesh.getGlobalNodeID(j) + comp_id);
-                    const GlobalIndexType signed_global_id = mesh.isGhostNode( mesh.getNode(j)->getID() )
-                                                    ? -global_id : global_id;
+                    const bool is_ghost = mesh.isGhostNode( mesh.getNode(j)->getID() );
+                    GlobalIndexType signed_global_id
+                                          = is_ghost ? -global_id : global_id;
+                    if (is_ghost)
+                    {
+                        _ghosts_indices.push_back(-signed_global_id);
+                        if (signed_global_id == 0)
+                             signed_global_id = -9999;
+                    }
+                    else
+                        _num_local_dof++;
+                       
                     _dict.insert(Line(Location(mesh_id, MeshLib::MeshItemType::Node, j),
                                 comp_id, signed_global_id) );
                 }
@@ -77,8 +88,18 @@ MeshComponentMap::MeshComponentMap(
                 {
                     const GlobalIndexType global_id = static_cast<GlobalIndexType>(global_index_offset
                                                           + mesh.getGlobalNodeID(j) );
-                    const GlobalIndexType signed_global_id = mesh.isGhostNode( mesh.getNode(j)->getID() )
-                                                    ? -global_id : global_id;
+                    const bool is_ghost = mesh.isGhostNode( mesh.getNode(j)->getID() );
+                    GlobalIndexType signed_global_id
+                                        = is_ghost ? -global_id : global_id;
+                    if (is_ghost)
+                    {
+                        _ghosts_indices.push_back(-signed_global_id);
+                        if (signed_global_id == 0)
+                             signed_global_id = -9999;
+                    }
+                    else
+                        _num_local_dof++;
+
                     _dict.insert(Line(Location(mesh_id, MeshLib::MeshItemType::Node, j),
                                 comp_id, signed_global_id) );
                 }
--- a/AssemblerLib/MeshComponentMap.cpp
+++ b/AssemblerLib/MeshComponentMap.cpp
@@ -50,7 +50,6 @@ MeshComponentMap::MeshComponentMap(
     }
 
     // construct dict (and here we number global_index by component type)
-    std::size_t global_index_offset = 0;
     std::size_t cell_index = 0;
     std::size_t comp_id = 0;
     _num_global_dof = 0;
@@ -79,7 +78,8 @@ MeshComponentMap::MeshComponentMap(
                  }
                  else
                  {
-                     global_id = static_cast<GlobalIndexType>(global_index_offset
+                     // _num_global_dof is used as the global index offset
+                     global_id = static_cast<GlobalIndexType>(_num_global_dof
                                              + mesh.getGlobalNodeID(j) );
                  }
                  const bool is_ghost = mesh.isGhostNode( mesh.getNode(j)->getID() );
@@ -105,12 +105,6 @@ MeshComponentMap::MeshComponentMap(
              for (std::size_t j=0; j<mesh_subset.getNElements(); j++)
                  _dict.insert(Line(Location(mesh_id, MeshLib::MeshItemType::Cell, j),
                               comp_id, cell_index++));
-
-             if (order == ComponentOrder::BY_COMPONENT)
-             {
-                  // Include base nodes.
-                  global_index_offset += mesh.getNGlobalNodes();
-             }
 
              _num_global_dof += mesh.getNGlobalNodes();
         }
--- a/AssemblerLib/MeshComponentMap.cpp
+++ b/AssemblerLib/MeshComponentMap.cpp
@@ -38,13 +38,16 @@ MeshComponentMap::MeshComponentMap(
     GlobalIndexType num_unknowns = 0;
     for (auto const c : components)
     {
-        assert (c != nullptr);
-        for (unsigned mesh_subset_index = 0; mesh_subset_index < c->size(); mesh_subset_index++)
+        assert(c != nullptr);
+        for (unsigned mesh_subset_index = 0; mesh_subset_index < c->size();
+             mesh_subset_index++)
         {
-            MeshLib::MeshSubset const& mesh_subset = c->getMeshSubset(mesh_subset_index);
+            MeshLib::MeshSubset const& mesh_subset =
+                c->getMeshSubset(mesh_subset_index);
             // PETSc always works with MeshLib::NodePartitionedMesh.
-            const MeshLib::NodePartitionedMesh &mesh
-                   = static_cast<const MeshLib::NodePartitionedMesh&>(mesh_subset.getMesh());
+            const MeshLib::NodePartitionedMesh& mesh =
+                static_cast<const MeshLib::NodePartitionedMesh&>(
+                    mesh_subset.getMesh());
             num_unknowns += mesh.getNGlobalNodes();
         }
     }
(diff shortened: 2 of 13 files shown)


ProcessLib/Process.h
```diff
@@ -258,24 +269,8 @@ class Process
...
-			auto const global_index = std::abs(  // 0 is the component id.
-			    _local_to_global_index_map->getGlobalIndex(l, 0));
-			(*result)[i] = u[global_index];
-		}
-#else
 		// Copy result
-		for (std::size_t i = 0; i < _x->size(); ++i)
-			(*result)[i] = (*_x)[i];
-#endif
+		_x->getValues(&(*result)[0]);
```
endJunction: In the nearest feature the `_x` vector will contain multiple components, which must be redistributed to different mesh properties. Will the proposed approach work?

Essentially it will be very similar to the previous version of the petsc code:
```c++
std::size_t const n = _mesh.getNNodes();
for (std::size_t i = 0; i < n; ++i)
{
	MeshLib::Location const l(_mesh.getID(),
	                          MeshLib::MeshItemType::Node, i);
	for (int component_id = 0; component_id < DisplacementDim;
	     ++component_id)
	{
		auto const global_index =
		    _local_to_global_index_map->getGlobalIndex(
		        l, component_id);
		(*result)[i * DisplacementDim + component_id] =
		    (*_x)[global_index]; /// <<< Copy _x to result in different order
	}
}
```

AssemblerLib/MeshComponentMap.cpp
```diff
@@ -34,12 +34,26 @@ MeshComponentMap::MeshComponentMap(
     : _num_components(components.size())
 
 {
+    // get number of unknows
+    GlobalIndexType num_unknows = 0;
```
endJunction: Typo: unknowns

AssemblerLib/MeshComponentMap.cpp
```diff
@@ -34,12 +34,26 @@ MeshComponentMap::MeshComponentMap(
...
+    for (auto c = components.cbegin(); c != components.cend(); ++c)
```
endJunction: using for-range-loop simplifies the code

AssemblerLib/LocalToGlobalIndexMap.h
```diff
@@ -68,6 +68,12 @@ class LocalToGlobalIndexMap
...
     std::size_t dofSize() const;
 
+    /// Returns total number of local degrees of freedom for DDC.
+    std::size_t dofSizeLocal() const
+    {
+        return _mesh_component_map.getNLocalUnknowns();
+    }
+
     /// Returns total number of global degrees of freedom for DDC.
     std::size_t dofSizeGlobal() const
```
endJunction: A question for understanding: what is the difference between `dofSize` and `dofSizeGlobal` in case of domain decomposition? One is the mesh component's dictionary size, the other one is like number of global mesh nodes used in the mesh component map...

AssemblerLib/MeshComponentMap.cpp
wenqing: Corrected.

AssemblerLib/MeshComponentMap.cpp
wenqing: Replaced.

AssemblerLib/LocalToGlobalIndexMap.h
wenqing: When PETSc is on, dofSize() returns the number of the unknowns of the rank that include those for  ghost nodes. While  dofSizeLocal() returns only the number of the unknowns that are belong to this present partition.

ProcessLib/Process.h
wenqing: Function ``_x->getValues(&(*result)[0])`` just fetches the values from the equation solutions, and it does not distinguish the solutions for components.  For the case of  component results, we have to modify this function as
1. Consider component in the following piece of code for variable name and the cache for the solutions to be output 
```
		// Get or create a property vector for results.
		boost::optional<MeshLib::PropertyVector<double>&> result;
		if (_mesh.getProperties().hasPropertyVector(property_name))
		{
			result = _mesh.getProperties().template getPropertyVector<double>(
			    property_name);
		}
		else
		{
			result =
			    _mesh.getProperties().template createNewPropertyVector<double>(
			        property_name, MeshLib::MeshItemType::Node);
			result->resize(_x->size());
		}
```
2. Fill the cache ``result`` for output with a similar piece of code to e previous version of the petsc code.
3. Possibly, modify vtu_interface.writeToFile(file_name) for multiple solution arrays.

I will check this issue in another PR.

AssemblerLib/LocalToGlobalIndexMap.h
endJunction: I see, thank you for explanation. Would you please add to the documentation of `dofSizeLocal` that it is not counting the ghost nodes; I think it will assist understanding.

wenqing: Added in the amended commit bf1f9dd

endJunction:
@chleh Could you please check if the changes fulfill the requirements for TES local assembly?

AssemblerLib/MeshComponentMap.h
```diff
@@ -126,12 +126,25 @@ class MeshComponentMap
...
 
+    /// Get the number of local unknowns excluding those associated
+    /// with ghost nodes (for DDC with node-wise mesh partitioning).
+    std::size_t getNLocalUnknowns() const
+    {
+        return _num_local_dof;
+    }
+
+    /// Get ghost indices (for DDC).
+    std::vector<GlobalIndexType> getGhostIndices() const
```
chleh: Returning a `const&` would be better.

AssemblerLib/LocalToGlobalIndexMap.h
```diff
@@ -94,6 +102,12 @@ class LocalToGlobalIndexMap
         return _mesh_component_map.getGlobalIndices(l);
     }
 
+    /// Get ghost indices, forwarded from MeshComponentMap.
+    std::vector<GlobalIndexType> getGhostIndices() const
```
chleh: Returning a `const&` would be better.

AssemblerLib/MeshComponentMap.cpp
```diff
@@ -34,18 +34,32 @@ MeshComponentMap::MeshComponentMap(
...
 {
+    // get number of unknows
+    GlobalIndexType num_unknowns = 0;
+    for (auto const c : components)
+    {
+        assert (c != nullptr);
+        for (unsigned mesh_subset_index = 0; mesh_subset_index < c->size(); mesh_subset_index++)
+        {
+            MeshLib::MeshSubset const& mesh_subset = c->getMeshSubset(mesh_subset_index);
+           const MeshLib::NodePartitionedMesh &mesh
```
chleh: indentation.

AssemblerLib/MeshComponentMap.cpp
```diff
@@ -34,18 +34,32 @@ MeshComponentMap::MeshComponentMap(
...
+                   = static_cast<const MeshLib::NodePartitionedMesh&>(mesh_subset.getMesh());
```
chleh: Should there be a comment that with Petsc we always work with `NodePartitionedMesh`? In a way that's natural.

AssemblerLib/MeshComponentMap.cpp
```diff
@@ -55,12 +69,23 @@ MeshComponentMap::MeshComponentMap(
...
                 {
-                    const GlobalIndexType global_id = static_cast<GlobalIndexType>(components.size()
+                    GlobalIndexType global_id = static_cast<GlobalIndexType>(components.size()
                                                * mesh.getGlobalNodeID(j) + comp_id);
-                    const GlobalIndexType signed_global_id = mesh.isGhostNode( mesh.getNode(j)->getID() )
-                                                    ? -global_id : global_id;
+                    const bool is_ghost = mesh.isGhostNode( mesh.getNode(j)->getID() );
+                    if (is_ghost)
+                    {
+                        _ghosts_indices.push_back(global_id);
```
chleh: So the values in `_ghost_indices` are all negative. Naively I'd expect them to be real ids, i.e. positive. Do they need to be negative?

AssemblerLib/MeshComponentMap.cpp
```diff
@@ -55,12 +69,23 @@ MeshComponentMap::MeshComponentMap(
                 // mesh items are ordered first by node, cell, ....
                 for (std::size_t j=0; j<mesh_subset.getNNodes(); j++)
                 {
-                    const GlobalIndexType global_id = static_cast<GlobalIndexType>(components.size()
+                    GlobalIndexType global_id = static_cast<GlobalIndexType>(components.size()
```
chleh: This was not introduced in this PR, but the formula for `global_id` looks like it makes some assumption about the component order. Btw: The argument `order` seems to be unused.

MathLib/LinAlg/Dense/DenseVector.h
```diff
@@ -75,6 +75,14 @@ class DenseVector : public std::valarray<T>
 		}
 	}
 
+	/// get entry values to an array
+	void getValues(T u[])
```
chleh: I'd strongly prefer `std::vector<T>&` or `std::valarray<T>&`.

AssemblerLib/MeshComponentMap.cpp
```diff
@@ -75,12 +100,23 @@ MeshComponentMap::MeshComponentMap(
...
-                                                    ? -global_id : global_id;
+                    const bool is_ghost = mesh.isGhostNode( mesh.getNode(j)->getID() );
+                    if (is_ghost)
+                    {
+                        _ghosts_indices.push_back(global_id);
+                        global_id = -global_id;
+                        // If the ghost entry has an index of 0,
+                        // its index is set to the negative value of unknowns.
+                        if (global_id == 0)
+                             global_id = -num_unknowns;
```
chleh: Must the `_ghost_indices` entry be altered then, too?

AssemblerLib/MeshComponentMap.cpp
```diff
@@ -75,12 +100,23 @@ MeshComponentMap::MeshComponentMap(
                 // mesh items are ordered first by node, cell, ....
```
chleh: This is the beginning of an else clause. It probably differs from the if-clause above only by the index calculation, i.e., a single line. Can't the commen parts of the two cases be merged somehow?

MathLib/LinAlg/PETSc/PETScMatrix.cpp
```diff
@@ -57,6 +54,7 @@ void PETScMatrix::setRowsColumnsZero(std::vector<PetscInt> const& row_pos)
     const PetscScalar one = 1.0;
     const PetscInt nrows = static_cast<PetscInt> (row_pos.size());
 
+    MatSetOption(_A, MAT_NO_OFF_PROC_ZERO_ROWS, PETSC_TRUE);
```
chleh: Could you add a comment on what that does, please?

MathLib/LinAlg/PETSc/PETScVector.cpp
```diff
@@ -36,14 +36,30 @@ PETScVector::PETScVector(const PetscInt vec_size, const bool is_global_size)
...
+PETScVector::PETScVector(const PetscInt vec_size,
+                         const std::vector<PetscInt>& ghost_ids,
+                         const bool is_global_size) :
+                         _size_ghosts(ghost_ids.size()), has_ghost_id(true)
+{
+    PetscInt nghosts = static_cast<PetscInt>( ghost_ids.size() );
+    if ( is_global_size )
+    {
+        VecCreateGhost(PETSC_COMM_WORLD, PETSC_DECIDE, vec_size, nghosts,
+                       &ghost_ids[0], &_v);
```
chleh: `ghost_ids.data()` should work, too.

MathLib/LinAlg/PETSc/PETScVector.h
```diff
@@ -48,6 +48,16 @@ class PETScVector
         PETScVector(const PetscInt vec_size, const bool is_global_size = true);
 
         /*!
+            \brief Constructor
+            \param vec_size       The size of the vector, either global or local
+            \param ghost_ids      The global indices of ghost entries
+            \param is_global_size The flag of the type of vec_size, i.e. whether it is a global size
+                                  or local size. The default is true.
```
chleh: I don't see what the meaning of this parameter is. I'll discuss it with you.

MathLib/LinAlg/PETSc/PETScVector.h
```diff
@@ -248,6 +257,11 @@ class PETScVector
         PetscInt _size;
         /// Size of local entries
         PetscInt _size_loc;
+        /// Size of local ghost entries
+        PetscInt _size_ghosts = 0;
+
+        /// Flag to indicate whether the vector is created with ghost entry indices
+        bool has_ghost_id = false;
```
chleh: underscore missing at the beginning.

MathLib/LinAlg/PETSc/PETScVector.cpp
```diff
@@ -116,6 +150,42 @@ void PETScVector::getGlobalVector(PetscScalar u[])
...
+       VecGetArray(_v, &loc_array);
+    return loc_array;
+}
+
+void PETScVector::restoreArray(PetscScalar* array) const
+{
+    if (has_ghost_id)
+    {
+        VecRestoreArray(_v_loc, &array);
+        //   VecGhostRestoreLocalForm(_v, &_v_loc);
```
chleh: Why is it commented? Isn't that an important call?

ProcessLib/Process.h
```diff
@@ -164,8 +164,11 @@ class Process
 		{
 			MeshLib::Location const l(_mesh.getID(),
 			                          MeshLib::MeshItemType::Node, i);
-			auto const global_index = std::abs(
+			auto global_index = std::abs(
 			    _local_to_global_index_map->getGlobalIndex(l, component_id));
+			// Ghost entry and its original index is 0
+			if ( global_index == _x->size() )
```
chleh: I don't like it, that now this check is scattered all across. But that can be changed later on. Could you add a TODO for it somewhere, please?

ProcessLib/Process.h
```diff
@@ -214,15 +217,23 @@ class Process
 		    static_cast<const MeshLib::NodePartitionedMesh&>(_mesh);
 		mat_opt.d_nz = pmesh.getMaximumNConnectedNodesToNode();
 		mat_opt.o_nz = mat_opt.d_nz;
+		mat_opt.is_global_size = false;
 		const std::size_t num_unknowns =
-		    _local_to_global_index_map->dofSizeGlobal();
+		    _local_to_global_index_map->dofSizeLocal();
```
chleh: When assembling a nonlinear PDE, one also needs access to the ghost entries, but here we only have non-ghost entries. Can that be changes easily, e.g., by using `dofSize()`?

Tests/MathLib/TestGlobalVectorInterface.cpp
```diff
@@ -187,6 +189,75 @@ void checkGlobalVectorInterfaceMPI()
...
+         rank1: 0 9
+         rank2: 3 5
+
+         Assuming the values of the entries are just their global indices,
+         we have local arrays as:
+         rank0: 0 1 2 3     6 8 10
+         rank1: 4 5 6 7 8   0 9
+         rank2: 9 10 11     3 5
+
+         The above ghost entry embeded vector is realized by the following
```
chleh: Typo: embedded.
Good description :cake:

Tests/MathLib/TestGlobalVectorInterface.cpp
```diff
@@ -187,6 +189,75 @@ void checkGlobalVectorInterfaceMPI()
...
+    MathLib::finalizeVectorAssembly(x_with_ghosts);
+
+    ASSERT_EQ(12u, x_with_ghosts.size());
+
+    std::vector<double> loc_v1(  x_with_ghosts.getLocalSize()
+                               + x_with_ghosts.getGhostSize() );
+    x_with_ghosts.getValues(&loc_v1[0]);                          
+    for (std::size_t i=0; i<expected.size(); i++)
+    {
+         ASSERT_NEAR(expected[i], loc_v1[i], 1.e-10);
```
chleh: Those small integers sould be exactly representable as double values. So `EXPECT_EQ` would be more appropriate, I think. Or is there some argument against it?

Tests/MathLib/TestGlobalVectorInterface.cpp
```diff
@@ -187,6 +189,75 @@ void checkGlobalVectorInterfaceMPI()
...
+    }
+    T_VECTOR x_with_ghosts(local_vec_size, ghost_ids, is_global_size);
+    x_with_ghosts.set(non_ghost_ids, non_ghost_vals);
+    MathLib::finalizeVectorAssembly(x_with_ghosts);
+
+    ASSERT_EQ(12u, x_with_ghosts.size());
+
+    std::vector<double> loc_v1(  x_with_ghosts.getLocalSize()
+                               + x_with_ghosts.getGhostSize() );
+    x_with_ghosts.getValues(&loc_v1[0]);
```
chleh: I'd prefer `loc_v1.data()`

MathLib/LinAlg/Dense/DenseVector.h
chleh: I see that the current interface is probably due to the interface of `PropertyVector`. So maybe that can be changed later on.

chleh:
Apart from my questions :+1:
Good work in general. Sadly the Petsc vectors have to be copied to access their values, but I don't see a simple solution for that for the time being.

MathLib/LinAlg/PETSc/PETScVector.cpp
wenqing: Activated.

ProcessLib/Process.h
wenqing: Added ifdef PETSc here and more comments as
```
			// The global indices of the ghost entries of the global
			// matrix or the global vectors need to be set as negative values
			// for equation assembly, however the global indices start from zero.
			// Therefore, any ghost entry with zero index is assigned an negative
			// value of the vector size or the matrix dimension.  
			// To assign the initial value for the ghost entries,
			// the negative indices of the ghost entries are restored to zero.
			// checked hereby
```

ProcessLib/Process.h
wenqing: Explained in our discussion.

Tests/MathLib/TestGlobalVectorInterface.cpp
wenqing: Corrected.

Tests/MathLib/TestGlobalVectorInterface.cpp
wenqing: Changed accordingly.

Tests/MathLib/TestGlobalVectorInterface.cpp
wenqing: Changed accordingly.

MathLib/LinAlg/PETSc/PETScVector.h
wenqing: Corrected.

MathLib/LinAlg/PETSc/PETScVector.h
wenqing: Added more words in the explanation.

MathLib/LinAlg/PETSc/PETScVector.cpp
wenqing: Changed.

MathLib/LinAlg/PETSc/PETScMatrix.cpp
wenqing: Added a comment.

AssemblerLib/MeshComponentMap.cpp
wenqing: Merged the common part with two if-conditions. Since this function is called once before the time stepping and the increasing of partitions decreases the size of the node loop, the two if-conditions inside the node loop should not slow down the computation distinctively.

AssemblerLib/MeshComponentMap.cpp
wenqing: No, it is not needed because ``_ghost_indices`` only  has positive values.

MathLib/LinAlg/Dense/DenseVector.h
wenqing: Yes, it is the reason. It will be changed later with a possible change in ``PropertyVector``.

AssemblerLib/MeshComponentMap.cpp
wenqing: The values in ``_ghost_indices`` are all positive. The next line set the global index negative for the equation assembly.

AssemblerLib/MeshComponentMap.cpp
wenqing: Added a comment above the line.

AssemblerLib/MeshComponentMap.cpp
wenqing: Adjusted.

AssemblerLib/LocalToGlobalIndexMap.h
wenqing: Changed.

AssemblerLib/MeshComponentMap.h
wenqing: Changed.

AssemblerLib/MeshComponentMap.cpp
```diff
@@ -34,67 +34,86 @@ MeshComponentMap::MeshComponentMap(
...
+             // the following global cell index must be reconsidered
+             // according to the employed cell indexing method.
+             for (std::size_t j=0; j<mesh_subset.getNElements(); j++)
+                 _dict.insert(Line(Location(mesh_id, MeshLib::MeshItemType::Cell, j),
+                              comp_id, cell_index++));
+
+             if (order == ComponentOrder::BY_COMPONENT)
+             {
+                  // Include base nodes.
+                  global_index_offset += mesh.getNGlobalNodes();
```
chleh: It looks like `global_index_offset` could be replaced by `comp_id * mesh.getNGlobalNodes()` or by `_num_global_dof`. That would simplify the code a bit.

chleh:
:+1: again.
I just found one very minor thing when reading through @wenqing's new changes.

AssemblerLib/MeshComponentMap.cpp
wenqing: Thanks to find this. Use ``_num_global_dof``.

wenqing:
@endJunction Please don't merge before I check the results with your PR.

wenqing:
@endJunction  Now the tests work again. You can merge it if you think it is good. 
 @bilke Would you please merge  ufz/ogs-data#1 once this PR is merged?

wenqing:
@endJunction Works

endJunction:
OK. I'll rebase and merge.

wenqing:
@endJunction  It has been rebased.

endJunction:
Great!
I'll extend the output to multicomponent process variables in next PR.

PR #1010, opened 2016-02-02
Tools for Neumann boundary condition preparation

### LINKED ISSUE #992: Bug in ComputeSurfaceNodeIDsInPolygonalRegion
The tool is used to compute a text file containing a table with two columns. The first column contains the node id of the node within the subsurface mesh, the second column contains the associated area. The file is used as an input (direct boundary condition) in the OGS-5 simulator.

The bug is: A incorrect area is computed for border nodes.

TomFischer:
This PR consists of two new simple tools
 - ExtractSurface
 - ComputeNodeAreasFromSurfaceMesh
and some additional functionality to the already existing tool ResetPropertyInPolygonalRegion.

The tools can be used in a workflow setting Neumann boundary conditions for different regions of the top surface (for instance groundwater recharge):
 1. Extract top surface (ExtractSurface)
 2. Mark the surface elements in the region the Neumann conditions should be set: ResetPropertyInPolygonalRegion
 3. Remove those parts of the surface mesh the Neumann condition shouldn't be set to deploying the tool removeMeshElements which results in a surface patch
 4. For OGS-5 modelling use the surface patch to generate Neumann condition using the tool ComputeNodeAreasFromSurfaceMesh

Using this workflow the bug described in issue #992 should be corrected.
Fixes #992.

### DIFF
--- a/Applications/Utils/MeshEdit/ExtractSurface.cpp
+++ b/Applications/Utils/MeshEdit/ExtractSurface.cpp
@@ -1,0 +1,80 @@
+/**
+ * @brief Extracts the surface from the given mesh.
+ *
+ * @copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/LICENSE.txt
+ */
+
+#include <algorithm>
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "tclap/CmdLine.h"
+
+#include "Applications/ApplicationsLib/LogogSetup.h"
+
+#include "BaseLib/StringTools.h"
+#include "BaseLib/FileTools.h"
+
+#include "FileIO/readMeshFromFile.h"
+#include "FileIO/writeMeshToFile.h"
+
+#include "MathLib/Vector3.h"
+
+#include "MeshLib/Mesh.h"
+#include "MeshLib/MeshSurfaceExtraction.h"
+
+int main (int argc, char* argv[])
+{
+	ApplicationsLib::LogogSetup logog_setup;
+
+	TCLAP::CmdLine cmd("Tool extracts the surface of the given mesh.", ' ',
+	                   "0.1");
+	TCLAP::ValueArg<std::string> mesh_in(
+	    "i", "mesh-input-file",
+	    "the name of the file containing the input mesh", true, "",
+	    "file name of input mesh");
+	cmd.add(mesh_in);
+	TCLAP::ValueArg<std::string> mesh_out(
+	    "o", "mesh-output-file",
+	    "the name of the file the surface mesh should be written to", false, "",
+	    "file name of output mesh");
+	cmd.add(mesh_out);
+	TCLAP::ValueArg<double> x("x", "x-component", "x component of the normal",
+	                          false, 0, "floating point value");
+	cmd.add(x);
+	TCLAP::ValueArg<double> y("y", "y-component", "y component of the normal",
+	                          false, 0, "floating point value");
+	cmd.add(y);
+	TCLAP::ValueArg<double> z("z", "z-component", "z component of the normal",
+	                          false, -1.0, "floating point value");
+	cmd.add(z);
+	TCLAP::ValueArg<double> angle_arg(
+	    "a", "angle", "angle between given normal and element normal", false,
+	    90, "floating point value");
+	cmd.add(angle_arg);
+
+	cmd.parse(argc, argv);
+
+	std::unique_ptr<MeshLib::Mesh const> mesh(
+	    FileIO::readMeshFromFile(mesh_in.getValue()));
+	INFO("Mesh read: %u nodes, %u elements.", mesh->getNNodes(), mesh->getNElements());
+
+	// extract surface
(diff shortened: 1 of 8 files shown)


Applications/Utils/MeshEdit/ExtractSurface.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+	cmd.parse(argc, argv);
+
+	std::unique_ptr<MeshLib::Mesh const> mesh(
+	    FileIO::readMeshFromFile(mesh_in.getValue()));
+	INFO("Mesh read: %u nodes, %u elements.", mesh->getNNodes(), mesh->getNElements());
+
+	// extract top surface
+	MathLib::Vector3 const dir(0.0, 0.0, -1.0);
+	double const angle(90);
```
endJunction: I'd add the direction (with default 0,0,-1) and the angle (with default value 90) to the command line arguments. Otherwise the tool should be named 'extractTopSurface'.

Applications/Utils/MeshEdit/ResetPropertiesInPolygonalRegion.cpp
```diff
@@ -224,6 +224,18 @@ int main (int argc, char* argv[])
...
 	if (int_property_arg.isSet()) {
 		int int_property_val(int_property_arg.getValue());
+
+		// check if PropertyVector exists
+		boost::optional<MeshLib::PropertyVector<int> &> opt_pv(
+			mesh->getProperties().getPropertyVector<int>(property_name)
+		);
+		if (!opt_pv) {
+			opt_pv = mesh->getProperties().createNewPropertyVector<int>(
+				property_name, MeshLib::MeshItemType::Cell, 1);
```
endJunction: The tuple size is 1 by default...
I had to ask myself what is so special about the 1 in the code...

Applications/Utils/ModelPreparation/ComputeNodeAreasFromSurfaceMesh.cpp
```diff
@@ -0,0 +1,126 @@
...
+		std::abort();
+	}
+
+	ids_and_area_out << std::setprecision(20);
+	csv_out << std::setprecision(20);
+
+	ids_and_area_out << ids_and_areas[0].first << " " << ids_and_areas[0].second;
+	csv_out << "ID x y z area node_id\n"; // CSV header
+	csv_out << 0 << " " << *(mesh_nodes[0])
+			<< ids_and_areas[0].second << " " << ids_and_areas[0].first;
```
endJunction: I do overlook something probably. Why is the output for id=0 not in the for-loop below? The loop starts from 1.

Applications/Utils/ModelPreparation/ComputeNodeAreasFromSurfaceMesh.cpp
```diff
@@ -0,0 +1,126 @@
...
+#include "BaseLib/FileTools.h"
+
+#include "FileIO/readMeshFromFile.h"
+#include "FileIO/readGeometryFromFile.h"
+
+#include "MeshLib/Mesh.h"
+#include "MeshLib/Node.h"
+#include "MeshLib/MeshSurfaceExtraction.h"
+
+void writeToFile(std::string const& id_area_fname, std::string const& csv_fname,
```
endJunction: Please add static to this function. Otherwise there is a missing prototype warning.

Applications/Utils/ModelPreparation/ComputeNodeAreasFromSurfaceMesh.cpp
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	std::vector<double> areas(
+	    MeshLib::MeshSurfaceExtraction::getSurfaceAreaForNodes(*surface_mesh));
+
+	// pack area and node id together
+	std::vector<std::pair<std::size_t, double>> ids_and_areas;
+	for (std::size_t k(0); k<areas.size(); ++k) {
+		ids_and_areas.push_back(std::make_pair((*orig_node_ids)[k], areas[k]));
+	}
```
endJunction: Can be written w/o for-loop:
```c++
std::transform(orig_node_ids->begin(), orig_node_ids->end(), areas.begin(),
	               std::back_inserter(ids_and_areas),
	               std::make_pair<std::size_t const&, double const&>);
```

MeshLib/MeshGenerators/VtkMeshConverter.cpp
```diff
@@ -455,6 +456,12 @@ void VtkMeshConverter::convertArray(vtkDataArray &array, MeshLib::Properties &pr
 		return;
 	}
 
+	if (vtkUnsignedLongArray::SafeDownCast(&array))
+	{
+		VtkMeshConverter::convertTypedArray<std::size_t>(array, properties, type);
```
endJunction: std::size_t -> unsigned long.
I'm not sure about the correctness of the assumption that std::size_t is always an unsigned long.

TomFischer: The assumption that `std::size_t` is equal to `unsigned long` is **not** always fulfilled. As far as I know the C++ standard does not specify the size of `std::size_t` nor does it specify the size of `unsigned long`. Is an vtk unsigned long garantied to be equal to a compiler and architecture dependend `unsigned long`? What is the correct vtk type for storing `std::size_t`?

bilke:
Maybe you can add a documentation page for the described workflow on docs.ogs.

PR #1012, opened 2016-02-03
[Fix] correct allocation of mat array for AddLayerToMesh

rinkk:


MeshLib/MeshEditing/AddLayerToMesh.cpp
```diff
@@ -174,12 +174,13 @@ MeshLib::Mesh* addLayerToMesh(MeshLib::Mesh const& mesh, double thickness,
 		if (!new_opt_materials) {
 			ERR("Can not set material properties for new layer");
 		} else {
+			new_opt_materials->resize(subsfc_elements.size());
 			unsigned new_layer_id(*(std::max_element(
 				opt_materials->cbegin(), opt_materials->cend()))+1);
 			std::copy(opt_materials->cbegin(), opt_materials->cend(),
```
endJunction: Using `reserve` above and `back_inserter` here (and below in `fill_n`) would save writing to the memory one time while initialization of `new_opt_materials`.

endJunction:
:fast_forward:

endJunction:
@rinkk There is a improvement suggestion above... Otherwise OK.

PR #1013, opened 2016-02-03
Property info display

rinkk:
generalised the methods for retrieving and displaying value bounds for all scalar arrays in a mesh


bilke:
:+1:

endJunction:
The (changed) code in `ElementTreeModel.cpp` and in `checkMesh.cpp` look very similar. Is it possible to avoid the code duplication?

rinkk:
i don't think so ... once i put the information into an output stream and once into a QVariant for populating QTreeItems.

PR #1015, opened 2016-02-03
Coverage fixes

bilke:
I refactored some code coverage logic, also thanks to remarks from @daviddoria.

Basically the CMake scripts for code coverage are now a bit more flexible. I also added Jenkins script for running the coverage in a Docker container. See the [results of it](https://svn.ufz.de:8443/job/OGS-6/job/Docker/job/coverage-publisher/). Coverage is run nightly.

- [x] Adapt Jenkins job to ufz-repo
- [x] Run nightly


endJunction:
Could you explain in short what the "Don't run test tester on OGS_COVERAGE." commit is doing? Just for understanding.

If the two bullet points are not yet resolved, set WIP flag; otherwise somebody could merge it....

bilke:
The diff comparison fails when building with coverage flags therefore I disabled the tester in this case. I haven't investigated this further. Coverage will be run in addition to normal tests (with testers enabled) so this should be ok.

endJunction:
Ahhh, the test tester. Got it!

:fast_forward:

bilke:
Jenkins test this please!

bilke:
Jenkins test this please!

PR #1021, opened 2016-02-04
Separate GroundwaterFlow process parser from its ctor.

endJunction:
Review commit-wise. There should be no behavioural changes.

~~@chleh Needs discussion on what is going into the base Process class...~~
Update: Extract common functions for
- [x] find variable
- [x] find parameter.

### DIFF
--- a/Applications/ApplicationsLib/ProjectData.h
+++ b/Applications/ApplicationsLib/ProjectData.h
@@ -99,8 +99,8 @@ public:
 				// several meshes. Then we have to assign the referenced mesh
 				// here.
 				_processes.emplace_back(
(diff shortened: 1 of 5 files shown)


bilke:
:+1:

wenqing:
Looks better. :+1:

chleh:
:+1: in general. Now GWFPcs is at its bare minimum.
The documentation of the `find...()` functions can be extended. However, for the time being this is not so important. Just add some TODO notes to the corresponding sections. I guess, in a few weeks we will have to do some documentation work anyway, and this point can be addressed then, too.

endJunction:
Fixed doxygen docs...

chleh:
:fast_forward:

PR #1025, opened 2016-02-05
added urls of util documentation to -h output

waltherm:
as titled


TomFischer:
Good idea! :fast_forward:

PR #1022, opened 2016-02-05
Tiny fixes in GeoLib, MeshLib and Tests

TomFischer:
For further details see the commit messages.

PR #1026, opened 2016-02-08
GML reader: Explicitly ignore polylines/surfaces without name

chleh:
This PR should fix the failing tests. It partially supersedes #1022.


endJunction:
@chleh Now the behaviour btw. qt and boost xml is different. One could split the tests for qt from boost but then there is a chance that two compilations do read different things (depending on qt's presence).
Or update qt tests...

endJunction:
To get the tests up and running fast I think it is best to add names to those polylines/surfaces for now. If there is a need to improve the tests regarding ignored polylines/surfaces (which is also affecting the qt parser), it should be done in different PR.

chleh:
@TomFischer's fix rulez!

endJunction:
@rinkk  Please have a look on this and merge, or mark for rebase&merge.

rinkk:
i have no idea what happens here but it looks okay?

PR #1030, opened 2016-02-09
[MGTL] Fix GeoMapper tri/line intersection memleak

endJunction:
Fix a memory leak detected by clang's address sanitizer in the testrunner.

### DIFF
--- a/MeshGeoToolsLib/GeoMapper.cpp
+++ b/MeshGeoToolsLib/GeoMapper.cpp
@@ -201,7 +201,7 @@ double GeoMapper::getMeshElevation(
 	    _grid->getNearestPoint(MathLib::Point3d{{{x, y, 0}}});
 	const std::vector<MeshLib::Element*> elements(
 	    _surface_mesh->getNode(pnt->getID())->getElements());
-	GeoLib::Point* intersection(nullptr);
+	std::unique_ptr<GeoLib::Point> intersection;
 
 	for (auto const & element : elements)
 	{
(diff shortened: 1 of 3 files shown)


MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -201,7 +201,7 @@ double GeoMapper::getMeshElevation(
 	    _grid->getNearestPoint(MathLib::Point3d{{{x, y, 0}}});
 	const std::vector<MeshLib::Element*> elements(
 	    _surface_mesh->getNode(pnt->getID())->getElements());
-	GeoLib::Point* intersection(nullptr);
+	std::unique_ptr<GeoLib::Point> intersection;
```
rinkk: is this automatically a nullptr? (otherwise this will fail a few lines below)

endJunction: Yes. The default ctor constructs a std::unique_ptr that owns nothing. [cppref](http://en.cppreference.com/w/cpp/memory/unique_ptr/unique_ptr)

chleh:
:fast_forward: although it's not good that you mixed refactorings and other changes in a single commit. ;-)

endJunction:
@chleh I must object. Only unique_ptr is changed.

chleh:
@endJunction: I must admit, you are right. And I must realize that Github's color-diff sucks! ^^

chleh:
Jenkins, test this please.

PR #1033, opened 2016-02-11
Various small dialog fixes

rinkk:
I don't think there is much need for a review here. This PR contains only cosmetic changes (size constraints, fixing typos, fixing dialog logic).


endJunction:
:fast_forward:

PR #1039, opened 2016-02-12
Multicomponent process variables and their output.

endJunction:
Process variable is now aware of its number of components. This is reflected in the tests by new `<components>` tag: [Tests data PR](https://github.com/ufz/ogs-data/pull/3/files).

Creation of the mesh property for each process variable is now handled by the process variable itself.

And finally the output is now generalized to all process variables of a process and is independent of number of components.

### DIFF
--- a/AssemblerLib/MeshComponentMap.cpp
+++ b/AssemblerLib/MeshComponentMap.cpp
@@ -285,4 +285,44 @@ std::vector<GlobalIndexType> MeshComponentMap::getGlobalIndicesByComponent(
     return global_indices;
 }
 
+GlobalIndexType MeshComponentMap::getLocalIndex(
+    Location const& l,
+    std::size_t const comp_id,
+    std::size_t const range_begin,
+    std::size_t const range_end) const
+{
+    GlobalIndexType const global_index = getGlobalIndex(l, comp_id);
+#ifndef USE_PETSC
+    return global_index;
+#else
+    if (global_index >= 0)    // non-ghost location.
+        return global_index - range_begin;
+
+    //
+    // For a ghost location look up the global index in ghost indices.
+    //
+    GlobalIndexType index;
+
+    // A special case for a ghost location with global index equal to the size
+    // of the local vector:
+    if (-global_index == _num_global_dof) return 0;
+
+    // TODO Find in ghost indices is O(n^2/2) for n being the length of
+    // _ghosts_indices. Providing an inverted table would be faster.
+    auto const ghost_index_it = std::find(_ghosts_indices.begin(),
+                                          _ghosts_indices.end(), -global_index);
+    if (ghost_index_it == _ghosts_indices.end())
+    {
+        ERR("index %d not found in ghost_indices", -global_index);
+        std::abort();
+    }
+
+    // Using std::distance on a std::vector is O(1). As long as _ghost_indices
+    // remains of std::vector type, this shall be fast.
+    return range_end - range_begin +
+           std::distance(_ghosts_indices.begin(), ghost_index_it);
+
+#endif
+}
+
 }   // namespace AssemblerLib
--- a/MeshLib/PropertyVector.h
+++ b/MeshLib/PropertyVector.h
@@ -70,14 +70,15 @@ protected:
 	/// @brief The constructor taking meta information for the data.
 	/// @param property_name a string describing the property
 	/// @param mesh_item_type the values of the property are either assigned to
-	/// nodes or cells (see enumeration MeshItemType) (default:
-	/// MeshItemType::Cell)
-	/// @param n_components the number of elements of a tuple (default: 1)
+	/// nodes or cells (see enumeration MeshItemType)
+	/// @param n_components the number of components of a property
 	explicit PropertyVector(std::string const& property_name,
-		MeshItemType mesh_item_type = MeshItemType::Cell,
-		std::size_t n_components = 1)
-		: std::vector<PROP_VAL_TYPE>(), _n_components(n_components),
-		_mesh_item_type(mesh_item_type), _property_name(property_name)
+	                        MeshItemType mesh_item_type,
+	                        std::size_t n_components)
+	    : std::vector<PROP_VAL_TYPE>(),
+	      _n_components(n_components),
+	      _mesh_item_type(mesh_item_type),
+	      _property_name(property_name)
 	{}
 
 	/// @brief The constructor taking meta information for the data.
@@ -85,15 +86,15 @@ protected:
 	/// with several entries)
 	/// @param property_name a string describing the property
 	/// @param mesh_item_type the values of the property are either assigned to
-	/// nodes or cells (see enumeration MeshItemType) (default:
-	/// MeshItemType::Cell)
-	/// @param n_components the number of elements of a tuple (default: 1)
+	/// nodes or cells (see enumeration MeshItemType)
+	/// @param n_components the number of components of a property
 	PropertyVector(std::size_t n_property_values,
-		std::string const& property_name,
-		MeshItemType mesh_item_type = MeshItemType::Cell,
-		std::size_t n_components = 1)
-		: std::vector<PROP_VAL_TYPE>(n_property_values*n_components),
-		_mesh_item_type(mesh_item_type), _property_name(property_name)
+	               std::string const& property_name,
+	               MeshItemType mesh_item_type,
+	               std::size_t n_components)
+	    : std::vector<PROP_VAL_TYPE>(n_property_values * n_components),
+	      _mesh_item_type(mesh_item_type),
+	      _property_name(property_name)
 	{}
 
 	std::size_t const _n_components;
@@ -195,19 +196,18 @@ protected:
 	/// \f$[0, \text{n_prop_groups})\f$.
 	/// @param property_name a string describing the property
 	/// @param mesh_item_type the values of the property are either assigned to
-	/// nodes or cells (see enumeration MeshItemType) (default:
-	/// MeshItemType::Cell)
-	/// @param n_components the number of elements of a tuple (default: 1)
+	/// nodes or cells (see enumeration MeshItemType)
+	/// @param n_components the number of elements of a tuple
 	PropertyVector(std::size_t n_prop_groups,
-		std::vector<std::size_t> const& item2group_mapping,
-		std::string const& property_name,
-		MeshItemType mesh_item_type = MeshItemType::Cell,
-		std::size_t n_components = 1)
-		: std::vector<std::size_t>(item2group_mapping),
-		_n_components(n_components),
-		_mesh_item_type(mesh_item_type),
-		_property_name(property_name),
-		_values(n_prop_groups * n_components)
+	               std::vector<std::size_t> const& item2group_mapping,
+	               std::string const& property_name,
+	               MeshItemType mesh_item_type,
+	               std::size_t n_components)
+	    : std::vector<std::size_t>(item2group_mapping),
+	      _n_components(n_components),
+	      _mesh_item_type(mesh_item_type),
+	      _property_name(property_name),
+	      _values(n_prop_groups * n_components)
 	{}
 
 protected:
--- a/MeshLib/PropertyVector.h
+++ b/MeshLib/PropertyVector.h
@@ -44,10 +44,10 @@ class PropertyVector : public std::vector<PROP_VAL_TYPE>,
 friend class Properties;
 
 public:
-	std::size_t getTupleSize() const { return _tuple_size; }
+	std::size_t getNumberOfComponents() const { return _n_components; }
 	std::size_t getNumberOfTuples() const
 	{
-		return std::vector<PROP_VAL_TYPE>::size() / _tuple_size;
+		return std::vector<PROP_VAL_TYPE>::size() / _n_components;
 	}
 	MeshItemType getMeshItemType() const { return _mesh_item_type; }
 	std::string const& getPropertyName() const { return _property_name; }
@@ -55,7 +55,7 @@ public:
 	PropertyVectorBase* clone(std::vector<std::size_t> const& exclude_positions) const
 	{
 		PropertyVector<PROP_VAL_TYPE> *t(new PropertyVector<PROP_VAL_TYPE>(_property_name,
-			_mesh_item_type, _tuple_size));
+			_mesh_item_type, _n_components));
 		BaseLib::excludeObjectCopy(*this, exclude_positions, *t);
 		return t;
 	}
@@ -72,11 +72,11 @@ protected:
 	/// @param mesh_item_type the values of the property are either assigned to
 	/// nodes or cells (see enumeration MeshItemType) (default:
 	/// MeshItemType::Cell)
-	/// @param tuple_size the number of elements of a tuple (default: 1)
+	/// @param n_components the number of elements of a tuple (default: 1)
 	explicit PropertyVector(std::string const& property_name,
 		MeshItemType mesh_item_type = MeshItemType::Cell,
-		std::size_t tuple_size = 1)
-		: std::vector<PROP_VAL_TYPE>(), _tuple_size(tuple_size),
+		std::size_t n_components = 1)
+		: std::vector<PROP_VAL_TYPE>(), _n_components(n_components),
 		_mesh_item_type(mesh_item_type), _property_name(property_name)
 	{}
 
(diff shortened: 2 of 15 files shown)


ProcessLib/InitialCondition.cpp
```diff
@@ -21,7 +21,7 @@
 namespace ProcessLib
 {
 std::unique_ptr<InitialCondition> createUniformInitialCondition(
-    BaseLib::ConfigTree const& config)
+    BaseLib::ConfigTree const& config, int const tuple_size)
```
chleh: Would `num_components` be a better name?

ProcessLib/InitialCondition.h
```diff
@@ -35,7 +35,7 @@ class InitialCondition
 {
 public:
 	virtual ~InitialCondition() = default;
-	virtual double getValue(MeshLib::Node const&) const = 0;
+	virtual double getValue(MeshLib::Node const&, int const) const = 0;
```
chleh: Can you add variable names, so one knows, what the arguments mean? Also to the already existing arg.

ProcessLib/InitialCondition.h
```diff
@@ -57,7 +57,7 @@ class UniformInitialCondition : public InitialCondition
 
 /// Construct a UniformInitialCondition from configuration.
 std::unique_ptr<InitialCondition> createUniformInitialCondition(
-    BaseLib::ConfigTree const& config);
+    BaseLib::ConfigTree const& config, int const tuple_size);
```
chleh: So the number of components of the IC has to match that of the process variable, I guess.
Makes sense.
Does it have to be documented?

ProcessLib/InitialCondition.h
```diff
@@ -70,7 +70,8 @@ class MeshPropertyInitialCondition : public InitialCondition
 		assert(_property.getMeshItemType() == MeshLib::MeshItemType::Node);
 	}
 
-	virtual double getValue(MeshLib::Node const& n) const override
+	virtual double getValue(MeshLib::Node const& n,
+	                        int const component_id) const override
```
chleh: Is the new argument only there for some overloads?

ProcessLib/Process.h
```diff
@@ -259,32 +260,34 @@ class Process
...
-		_x->copyValues(*result);
+		_x->copyValues(x_copy);
 
+		std::size_t const n = _mesh.getNNodes();
+		for (ProcessVariable& pv : _process_variables)
+		{
+			auto& output_data = pv.getOrCreateMeshProperty();
+
+			int const n_components = pv.getTupleSize();
+			for (std::size_t i = 0; i < n; ++i)
```
chleh: compare int to size_t

ProcessLib/Process.h
```diff
@@ -259,32 +260,34 @@ class Process
...
+			{
+				MeshLib::Location const l(_mesh.getID(),
+				                          MeshLib::MeshItemType::Node, i);
+				for (int component_id = 0; component_id < n_components;
+				     ++component_id)
+				{
+					auto const global_index =
+					    std::abs(_local_to_global_index_map->getGlobalIndex(
+					        l, component_id));
+					output_data[i] = (x_copy)[global_index];
```
chleh: no need for parens here.

ProcessLib/ProcessVariable.cpp
```diff
@@ -113,4 +114,24 @@ MeshLib::Mesh const& ProcessVariable::getMesh() const
 	return _mesh;
 }
 
+MeshLib::PropertyVector<double>& ProcessVariable::getOrCreateMeshProperty()
+{
+	boost::optional<MeshLib::PropertyVector<double>&> result;
```
chleh: Too lazy to check by myself: Why do we need `boost::optional` here? The code below looks like it worked well without it.

ProcessLib/ProcessVariable.h
```diff
@@ -87,14 +90,20 @@ class ProcessVariable
...
 
+	// Get or create a property vector for results.
+	// The returned mesh property size is number of mesh nodes times tuple size.
+	MeshLib::PropertyVector<double>& getOrCreateMeshProperty();
+
 private:
 	std::string const _name;
-	MeshLib::Mesh const& _mesh;
+	MeshLib::Mesh& _mesh;
+	int _tuple_size;
```
chleh: `int` or `unsigned` or `size_t`, what would be best?

chleh: I currently don't see, why the tests fail (no access to Jenkins). From my side only minor things. Apart from them :+1:
Have a nice weekend.

ProcessLib/InitialCondition.cpp
endJunction: It's not consistent now. (I'll check consistency related to this PR one more time.) 
I'd prefer tuple size here, so there is no ambiguity to number of components in a process...
MeshProperty uses tuple_size, same as VTK output.

ProcessLib/InitialCondition.h
endJunction: Getting values in process variable (see [here](ffbb9ac68f450ebef691cc97b4d0d2b6ec81e176)) for each component when called from [Process::setInitialConditions()](https://github.com/endJunction/ogs/blob/MulticomponentProcessVars/ProcessLib/Process.h#L183).

ProcessLib/Process.h
endJunction: Good point. I managed to shoot myself in the foot while changing this :-o

ProcessLib/InitialCondition.h
```diff
@@ -45,8 +45,8 @@ class UniformInitialCondition : public InitialCondition
 	UniformInitialCondition(double const value) : _value(value)
 	{
 	}
-
-	virtual double getValue(MeshLib::Node const&) const override
+	virtual double getValue(MeshLib::Node const&,
+	                        int const /* tuple_size */) const override
```
chleh: I guess, `tuple_size` is the wrong name. Maybe that is what you meant in your comment above.
Anyway, I think, also for uniform ICs, different components can have different values. Abd that should be reflected in this class.

ProcessLib/InitialCondition.h
chleh: I oppose: Not resolved. I think, it is also important to have argument names here in the interface.

ProcessLib/InitialCondition.h
chleh: But `component_id` is not used inside this method. That's why I asked. Maybe I should have asked "overrides".

chleh: Probably the same applies as for the Uúniform IC: It should be possible that different components have different values.

ProcessLib/ProcessVariable.cpp
endJunction: Both `getPropertyVector()` and `createNewPropertyVector()` return `boost::optional`. So it is convinient, although because it is checked if the property exists one could do without...
This way is a little safer, because you get a chance to check if everything went well before dereferencing.

ProcessLib/ProcessVariable.h
endJunction: On cppcon15 it was recommended to use int. If there is need for large numbers, use long. And only if doing bit-arithmetic (or some other thing I forgot now, but probably smth. with addresses) use unsigned types.

ProcessLib/InitialCondition.h
endJunction: Have a look on the freshly pushed code....
There will be more race conditions otherwise...

ProcessLib/InitialCondition.h
endJunction: Yes. It's not used yet, because, well, it's not implemented. Will make another PR, because it needs a config tree's ability to read tuples...

MeshLib/PropertyVector.h
```diff
@@ -72,11 +72,11 @@ friend class Properties;
 	/// @param mesh_item_type the values of the property are either assigned to
 	/// nodes or cells (see enumeration MeshItemType) (default:
 	/// MeshItemType::Cell)
-	/// @param tuple_size the number of elements of a tuple (default: 1)
+	/// @param n_components the number of elements of a tuple (default: 1)
 	explicit PropertyVector(std::string const& property_name,
 		MeshItemType mesh_item_type = MeshItemType::Cell,
-		std::size_t tuple_size = 1)
-		: std::vector<PROP_VAL_TYPE>(), _tuple_size(tuple_size),
+		std::size_t n_components = 1)
```
chleh: Could you try to also get rid of all those default values, or add a todo, if it is too much work for now.

chleh: Maybe `operator[]` in this class now does not make sense anymore.

endJunction: Agree. Another PR.

endJunction: Actually the default arguments were not used anyway, so removed them in this PR. The `operator[]` needs discussion.

endJunction:
I messed up output for the PETSc configuration. Needs rework.

ProcessLib/Process.h
```diff
@@ -259,32 +260,34 @@ class Process
...
+
+			int const n_components = pv.getNComponents();
+			for (std::size_t node_id = 0; node_id < n_nodes; ++node_id)
+			{
+				MeshLib::Location const l(_mesh.getID(),
+				                          MeshLib::MeshItemType::Node, node_id);
+				for (int component_id = 0; component_id < n_components;
+				     ++component_id)
+				{
+					auto const global_index =
```
endJunction: Note to myself: Using global index is wrong here.
```
error: 0 | n 432, c 0, gi 432 -> 432
error: 0 | n 433, c 0, gi 1190 -> 433
```
Seems like when starting accessing ghost nodes the global index can jump, like 433 is ghost node and is way back in the dof table. (1190 is one of the ghost indices). But for other ranks it's slightly different:
```
error: 2 | 626 456 170 626
error: 2 | n 0, c 0, gi 875 -> 0
```
and
```
error: 1 | n 185, c 0, gi 618 -> 185
error: 1 | n 186, c 0, gi 619 -> 186
```

wenqing: Yes, using the global index is wrong inside node and component loops. The global index is for the index of any entry of the global matrix and global vector, and it already contains the component id. Here a function  to fetch the value for this node with a known of global id and a known of component id is needed.

endJunction:
OK two more changes were needed to fixup the PETSc configuration.
Main change there is the computation function of the local index in a (PETSc)vector for a given location and component. The implementation is not the nicest one, I admit. If you have better ideas, you are welcome, otherwise it will be future todo.

@bilke Could you please have a look on hanging Jenkins...

AssemblerLib/MeshComponentMap.cpp
```diff
@@ -285,4 +285,39 @@ std::vector<GlobalIndexType> MeshComponentMap::getGlobalIndicesByComponent(
...
+    //
+    // For a ghost location look up the global index in ghost indices.
+    //
+    GlobalIndexType index;
+
+    // A special case for a ghost location with global index equal to the size
+    // of the local vector:
+    if (-global_index == _num_global_dof) return 0;
+
+    auto const ghost_index_it = std::find(_ghosts_indices.begin(),
```
chleh: Just asking myself: Are ghost indices sorted? Then you could do a binary search here,

chleh: Maybe it would be more efficient to implement it the other way around: Save local indices in the `_dict` and provide a method that converts local to global indices; since local indices are contiguous (are they?), they can be directly used as an index in a `ghost_index_local_to_global_index_mapping`.

AssemblerLib/MeshComponentMap.cpp
```diff
@@ -285,4 +285,39 @@ std::vector<GlobalIndexType> MeshComponentMap::getGlobalIndicesByComponent(
...
+                                          _ghosts_indices.end(), -global_index);
+    if (ghost_index_it == _ghosts_indices.end())
+    {
+        ERR("index %d not found in ghost_indices", -global_index);
+        std::abort();
+    }
+    return range_end - range_begin +
+           std::distance(_ghosts_indices.begin(), ghost_index_it);
```
chleh: Should you comment that `std::distance` really does do only pointer arithmetic in the case of `std::vector` iterators?

ProcessLib/InitialCondition.h
chleh: OK, I thought you dropped the need for reading tuples from your concept, but we can stick to that need.

chleh:
Maybe I'll have a detailed look on it tomorrow. For now only the general remarks.

AssemblerLib/MeshComponentMap.cpp
endJunction: The ghost indices are not sorted.
The second idea looks worthwhile. @wenqing What do you think?

wenqing: For the case of ``ComponentOrder::BY_COMPONENT`` and the number of components being one , the indices of entries of the local array are identical to the  indices of the nodes of the local partitioned mesh. For such case, the fetched local array from the global PETSc vector via the member of PETScVector can be directly used for output.  If `ComponentOrder::BY_COMPONENT`` and the number of components being large than one, to fetch a solution array of a component from the local solution array is also easy. 

For the case of ``ComponentOrder::BY_LOCATION``, the index mapping is not so straightforward. 

@endJunction I will prepare a PR to your PR.

endJunction:
I added the three discussed todos for later. Thanks to @chleh and @wenqing.

The automatic merge is not working only because of the Test/Data, so there will be no changes but in Test/Data.

MeshLib/PropertyVector.h
```diff
@@ -70,14 +70,15 @@ friend class Properties;
 	/// @brief The constructor taking meta information for the data.
 	/// @param property_name a string describing the property
 	/// @param mesh_item_type the values of the property are either assigned to
-	/// nodes or cells (see enumeration MeshItemType) (default:
-	/// MeshItemType::Cell)
-	/// @param tuple_size the number of elements of a tuple (default: 1)
+	/// nodes or cells (see enumeration MeshItemType)
+	/// @param n_components the number of elements of a tuple
```
chleh: Minor wording mismatch:
> n_components the number of elements of a tuple

Better:
> n_components the number of components of the property

MeshLib/PropertyVector.h
```diff
@@ -85,18 +86,18 @@ friend class Properties;
 	/// with several entries)
 	/// @param property_name a string describing the property
 	/// @param mesh_item_type the values of the property are either assigned to
-	/// nodes or cells (see enumeration MeshItemType) (default:
-	/// MeshItemType::Cell)
-	/// @param tuple_size the number of elements of a tuple (default: 1)
+	/// nodes or cells (see enumeration MeshItemType)
+	/// @param n_components the number of elements of a tuple
```
chleh: same here

ProcessLib/InitialCondition.h
```diff
@@ -45,8 +45,8 @@ class UniformInitialCondition : public InitialCondition
 	UniformInitialCondition(double const value) : _value(value)
 	{
 	}
-
-	virtual double getValue(MeshLib::Node const&) const override
+	virtual double getValue(MeshLib::Node const&,
+	                        int const /* n_components */) const override
```
chleh: wrong argument name `component_id`
Please add `TODO implement`.

ProcessLib/InitialCondition.h
```diff
@@ -70,9 +73,11 @@ class MeshPropertyInitialCondition : public InitialCondition
 		assert(_property.getMeshItemType() == MeshLib::MeshItemType::Node);
 	}
 
-	virtual double getValue(MeshLib::Node const& n) const override
+	virtual double getValue(MeshLib::Node const& n,
+	                        int const component_id) const override
 	{
-		return _property[n.getID()];
+		return _property[n.getID() * _property.getNumberOfComponents() +
+		                 component_id];
```
chleh: In a subsequent PR I'd like to see the operator[] dropped. Please add a todo somewhere.

ProcessLib/ProcessVariable.h
```diff
@@ -57,6 +57,9 @@ class ProcessVariable
 	/// Returns a mesh on which the process variable is defined.
 	MeshLib::Mesh const& getMesh() const;
 
+	/// Returns the number of components of the process variable.
+	int getNComponents() const { return _n_components; }
```
chleh: In the property vector the method is called `getNumberOfComponents()`. Maybe you want to use the longer name here, too.

chleh:
Only minor remarks. Afterwards :+1:.

wenqing:
@endJunction :+1: . I will prepare another PR once this one is merged and once I get time.

ProcessLib/InitialCondition.h
endJunction: https://github.com/ufz/ogs/issues/1055

ProcessLib/ProcessVariable.h
```diff
@@ -87,14 +90,21 @@ class ProcessVariable
...
+	// Get or create a property vector for results.
+	// The returned mesh property size is number of mesh nodes times number of
+	// components.
+	MeshLib::PropertyVector<double>& getOrCreateMeshProperty();
+
 private:
 	std::string const _name;
-	MeshLib::Mesh const& _mesh;
+	MeshLib::Mesh& _mesh;
+	int _n_components;
```
chleh: Can probably be made const in a follow-up PR.

endJunction: How could one forget to write const-correct code? Will do.

AssemblerLib/MeshComponentMap.cpp
```diff
@@ -285,4 +285,44 @@ std::vector<GlobalIndexType> MeshComponentMap::getGlobalIndicesByComponent(
...
 
+GlobalIndexType MeshComponentMap::getLocalIndex(
+    Location const& l,
+    std::size_t const comp_id,
+    std::size_t const range_begin,
+    std::size_t const range_end) const
+{
+    GlobalIndexType const global_index = getGlobalIndex(l, comp_id);
+#ifndef USE_PETSC
+    return global_index;
```
chleh: TODO: add `(void) range_begin; (void) range_end` in order to silence compiler warnings.

PR #1041, opened 2016-02-15
Remove ctest output files

### LINKED ISSUE #1038: Remove ctest results before running the actual tests.
Currently the result files are overwritten, but if, because of some error, no output is written, the previous ctest results are used.

@bilke Could you add something to AddTest cmake function or something?

bilke:
Fixes #1038.


endJunction:
Works for me. :+1: 
@bilke If you think that deleting the whole Tests/Data dir is easier/more stable/nicer go for it, otherwise merge.

Issue #1023, opened 2016-02-05
XML GML Test disabled when QT4 is not found


chleh:
Because of that this test case is never run in our CI system.

```
if(QT4_FOUND)
     set(TEST_SOURCES ${TEST_SOURCES} FileIO/TestXmlGmlReader.cpp)
endif()
```

rinkk:
PR #1042 should solve this

TomFischer:
Solved.

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/1023).

PR #1042, opened 2016-02-16
Splitting GML tests into boost and qt part

rinkk:
this should solve #1023


Tests/CMakeLists.txt
```diff
@@ -15,12 +16,8 @@ APPEND_SOURCE_FILES(TEST_SOURCES MeshLib)
 APPEND_SOURCE_FILES(TEST_SOURCES MeshGeoToolsLib)
 APPEND_SOURCE_FILES(TEST_SOURCES NumLib)
 
-set(TEST_SOURCES ${TEST_SOURCES}
-	FileIO/TestGLIReader.cpp
-	FileIO/TestCsvReader.cpp
-)
 if(QT4_FOUND)
-	set(TEST_SOURCES ${TEST_SOURCES} FileIO/TestXmlGmlReader.cpp)
+	set(TEST_SOURCES ${TEST_SOURCES} FileIO_Qt/TestQtGmlInterface.cpp)
```
bilke: Just use `APPEND_SOURCE_FILES` here as well (for consistency and the case there will be more test files in the future).

TomFischer:
Is it possible to use exact the same 'FileIOGml' class in both tests? At least the surface test seems to be different.

rinkk:
@TomFischer : No idea, Lars also didn't know a way to do this...

bilke:
Jenkins test this please!

TomFischer:
The classes `FileIOBoostGml` and `FileIOQtGml` are (almost?) the same. I would extract the class and reuse it in both tests. This would save ~200 lines.

rinkk:
moved identical classes to header

TomFischer:
:fast_forward:

PR #1040, opened 2016-02-12
ODE Solver for transient and nonlinear processes

chleh:
I took the last few days to take an epic attempt on reinventing the wheel. Here's the result: A new ODE solver structure. It is not yet feature-complete—like small child—but it is shiny and new, call it iSolver if you like.

I will try to finish the work on this PR until Tuesday, 16 Feb. Do not review yet (although the structures will not change anymore).

TODO:
* [x] replace present timeloop
* [x] iron out some edges with linear solvers and matrix classes (separate PR)
* [x] header guards, copyright texts
* [x] API docs (mostly done)
* [ ] Petsc matrix and vector (stub done, todo: implementation, tests)
* [x] :no_entry_sign: LIS matrix and vector: I'd like to drop them since BLAS functionality is only provided for dense matrices. Discuss.
* [ ] some cleanup
* [ ] write builders that parse configuration options
* [x] clean up BLAS interface

Also important:
* [ ] Compare test cases to actual data. Currently the tests only "run", but are not compared to analytical solutions. It would be nice if some of the mathematicians could do that comparison. Everything is prepared, only the right tolerances have to be given.

Future work:
* [ ] save some matrices (i.e. save space)
* [ ] adaptive time stepping
* [ ] different error norms
* [ ] numerical Jacobians
* [ ] process coupling
* [ ] ...

PR #1045, opened 2016-02-18
ODE Solver for transient and nonlinear processes – core functionality

chleh:
This is a splitoff from #1040, i.e., it contains the parts that have been reviewed yesterday by @wenqing and @endJunction. The comments from the review are in d6131c9, all new modifications come after them. I hope I did not forget to address any of them.

One comment I did not address is that about the ordering of BLAS function arguments. Currently that follows the ordering of the respective PETSc functions. I propose to keep it like that for the time being. Maybe we can change it later on.

As with #1040 the following is the state:
* LisMatrix/LisVector is not supported
* The PETSc build compiles, but many methods are only stubs. @wenqing, maybe we can work on that later on.
* The unit test only makes sure that things compile. Currently it only produces result files that are not compared to the analytical solutions. A strict test has to be implemented later on.


chleh:
Btw. I made a small design improvement: In c1d6cdc a very basic EquationSystem interface is introduced.

endJunction:
Travis takes too long to run
```
Running test GMatrix_GVector_ODE1_BackwardEuler_Picard with 1000 timesteps of size 0.03 s.
```
and then it quits because of timeout.

chleh:
I think it quits because of
```

The log length has exceeded the limit of 4 Megabytes (this usually means that test suite is raising the same exception over and over).

The build has been terminated.
```

chleh:
I'll reduce the number of timesteps for now.

MathLib/LinAlg/BLAS.cpp
```diff
@@ -0,0 +1,204 @@
...
+}
+
+// y = a*x + y
+void axpby(PETScVector& y, double const a, double const b, PETScVector const& x)
+{
+    // TODO check sizes
+    VecAXPBY(y.getRawVector(), a, b, x.getRawVector());
+}
+
+
```
wenqing: http://ufz.github.io/styleguide/cppguide.xml#Vertical_Whitespace

MathLib/LinAlg/PETSc/PETScVector.h
```diff
@@ -171,6 +177,14 @@ class PETScVector
             VecGetValues(_v, e_idxs.size(), &e_idxs[0], &sub_vec[0]);
         }
 
+        // TODO preliminary
+        double operator[] (PetscInt idx) const
```
wenqing: In PETSc vector, to fetch a single element is relatively expensive.

NumLib/ODESolver/MatrixTranslator.h
```diff
@@ -0,0 +1,435 @@
...
+
+namespace NumLib
+{
+
+//! \addtogroup ODESolver
+//! @{
+
+
+/*! Translates matrices assembled by a provided ODE (or other equation)
+ *  to some other matrices suitable to be passed on to nonlinear solvers.
```
wenqing: to the stiffness matrix and the right hand side vector of a system of linear equations for linear solver.

NumLib/ODESolver/MatrixTranslator.h
```diff
@@ -0,0 +1,435 @@
...
+ *
+ * \tparam Matrix the type of matrices occuring in the linearization of the equation.
+ * \tparam Vector the type of the solution vector of the ODEequation
+ * \tparam ODETag a tag indicating the type of equation.
+ */
+template<typename Matrix, typename Vector, ODESystemTag ODETag>
+class MatrixTranslator;
```
wenqing: Might be ``MatrixVectorTranslator`` or keep it like this if you prefer to.

NumLib/ODESolver/ODETypes.h
```diff
@@ -0,0 +1,210 @@
...
+#include "MathLib/LinAlg/Eigen/EigenVector.h"
+#include "MathLib/LinAlg/Eigen/EigenMatrix.h"
+#include "MathLib/LinAlg/Eigen/EigenLinearSolver.h"
+
+
+namespace NumLib
+{
+inline void oneShotLinearSolve(MathLib::EigenMatrix& A, MathLib::EigenVector& rhs, MathLib::EigenVector& x)
+{
+    MathLib::EigenLinearSolver slv(A);
```
wenqing: To create a solver instance each time before it is used is not good for PETSc ksp solver. But just keep it like this, I will test it once this PR is merged.

NumLib/ODESolver/ODETypes.h
```diff
@@ -0,0 +1,210 @@
...
+#include "MathLib/LinAlg/PETSc/PETScVector.h"
+
+
+namespace NumLib
+{
+inline void oneShotLinearSolve(MathLib::PETScMatrix& A, MathLib::PETScVector& rhs,
+                               MathLib::PETScVector& x)
+{
+    (void) A; (void) rhs; (void) x;
+    // TODO implement
```
wenqing: To create a solver instance each time before it is used is not good for PETSc ksp solver. But just keep it like this, I will test it once this PR is merged.

MathLib/LinAlg/BLAS.cpp
chleh: Nevertheless I'd like to keep the two empty lines as they are, because the Matrix section starts here.

NumLib/ODESolver/ODETypes.h
chleh: Probably I will drop the `oneShotLinearSolve()` functions in the follow-up PR that integrates the ODE lib into the rest of OGS6.

MathLib/LinAlg/PETSc/PETScVector.h
chleh: I know. The problem is that the interfaces how to access values in EigenVector, Eigen::VectorXd, and PETScVector all are different. I need the access in `Tests/NumLib/ODEs.h` and I did not want to add another method like `setMatrix()` to `ODETypes.h`. It seemed to be the minimal change to just add `operator[]` to the PETScVector. That's not a final decision, that's why it's marked preliminary.
I don't want to clutter this PR with more of this interface unification stuff. Once we design a uniform matrix and vector interface (hopefully soon), we will remove that method. But for the time being I'd like to keep it in this PR.

NumLib/ODESolver/MatrixTranslator.h
chleh: I'll keep the old name, because it's less work ^^ , and the typenames, hence the lines of code, are already long enough (e.g. `MatrixTranslatorCrankNicolson`).

wenqing:
:+1:
:fast_forward:

NumLib/ODESolver/MatrixTranslator.h
```diff
@@ -0,0 +1,436 @@
...
+namespace NumLib
+{
+
+//! \addtogroup ODESolver
+//! @{
+
+
+/*! Translates matrices and vectors assembled by a provided ODE (or other equation)
+ *  to the stiffness matrix and right-hand side vector of a linear equation system
+ *  that can solved by a linear equation system solver.
```
wenqing: can be

MathLib/LinAlg/Eigen/EigenMatrix.h
```diff
@@ -123,11 +128,13 @@ class EigenMatrix final
         return _mat.coeff(row, col);
     }
 
+    /*
```
endJunction: Please add a short comment on why or a todo.

NumLib/ODESolver/EquationSystem.h
```diff
@@ -0,0 +1,42 @@
...
+class EquationSystem
+{
+public:
+    //! Return the number of equations.
+    virtual std::size_t getNumEquations() const = 0;
+
+    /*! Check whether this is actually a linear equation system.
+     *
+     * \remark
+     * Depending on its parameters an in general nonlinear equation system
```
endJunction: 'an in' -> 'a'

NumLib/ODESolver/MatrixTranslator.h
```diff
@@ -0,0 +1,436 @@
...
+//! \addtogroup ODESolver
+//! @{
+
+
+/*! Translates matrices and vectors assembled by a provided ODE (or other equation)
+ *  to the stiffness matrix and right-hand side vector of a linear equation system
+ *  that can be solved by a linear equation system solver.
+ *
+ * \tparam Matrix the type of matrices occuring in the linearization of the equation.
+ * \tparam Vector the type of the solution vector of the ODEequation
```
endJunction:
```vim
%s/ODEequation/ODE/g
```

NumLib/ODESolver/ODETypes.h
```diff
@@ -0,0 +1,210 @@
...
+
+    auto it = values.begin();
+    for (IndexType r=0; r<rows; ++r) {
+        for (IndexType c=0; c<cols; ++c) {
+            m(r, c) += *(it++);
+        }
+    }
+}
+
+namespace NumLib
```
endJunction: Please check the namespaces in this file; it seems when USE_PETSC is set, the NumLib namespace is opened twice.
Above the functions live in the global namespace. Though it is very temporary, put them in NumLib namespace for example.

NumLib/ODESolver/EquationSystem.h
chleh: No, that's what i mean: The equation system is nonlinear "in general", but in some special case it can be linear, e.g., for constant coefficients.

NumLib/ODESolver/ODETypes.h
chleh: Actually it would be better to put all of this into the MathLib namespace in the end, because all matrices, vectors and linear solvers reside there.
However, I will put everything into the NumLib namespace since the file is located there. I will add a comment that ths will have to change.

NumLib/ODESolver/ODETypes.h
```diff
@@ -0,0 +1,212 @@
...
+#include <Eigen/LU>
+
+namespace NumLib
+{
+
+inline void setMatrix(Eigen::MatrixXd& m,
+                      Eigen::MatrixXd::Index const rows, Eigen::MatrixXd::Index const cols,
+                      std::initializer_list<double> values)
+{
+    using IndexType = Eigen::MatrixXd::Index;
```
endJunction: This index type does not correspond to the MathLib::EigenMatrix index type.
Also see a note in [EigenVector.h](https://github.com/ufz/ogs/blob/master/MathLib/LinAlg/Eigen/EigenVector.h#L34).

In general the matrix and vector types (and their indices) are in inconsistent state. I think the `MatrixTraits` shall be defined only for MathLib::XXXMatrix/Vector types.

endJunction:
Don't forget to rebase and to check if jenkins is OK. (Hopefully Jenkins will be in better state in the morning.)

bilke:
I restarted Jenkins and it should be ok.

NumLib/ODESolver/ODETypes.h
chleh: I don't know what you mean. ^^

> This index type does not correspond to the MathLib::EigenMatrix index type.

That's true, but that's fine, because this method is dealing with `Eigen::MatrixXd`, so the index type must be consistent with that.

> I think the MatrixTraits shall be defined only for MathLib::XXXMatrix/Vector types.

That's not true. (1) `MatrixTraits` are not used in this function. (2) `MatrixTraits` are used in other occasions in the ODESolver lib, thus they must provide the right index type for any matrix type used. Admittedly, currently the `MatrixTraits` are only used in `ODESystem.h`, but the use might be extended, e.g., one could replace many return types of `std::size_t` by `MatrixTraits<...>::Index` in a future PR.

chleh:
Jenkins looks fine, so I'm gonna push the green button now.

chleh:
@endJunction, @wenqing thanks again for your review. More coming soon. ^^

PR #1047, opened 2016-02-19
Check for time binary

### LINKED ISSUE #1044: ctest wrapper command
I just found out after quite some time that ArchLinux is not shipping a /usr/bin/time util (i.e. timing that is not a shell builtin) by default anymore. Should there be a check in cmake if `which time` succeeds?

bilke:
The check was there but disabled because it did not work in the past on Travis. But now it works. If `time` is not found there is CMake warning that corresponding are disabled.

Closes #1044.


chleh:
:fast_forward:

PR #1048, opened 2016-02-19
initializeMeshSubsets() across process vars and components.

endJunction:
One step towards multi-variable processes
MeshSubset initialization across all process variables.

### DIFF
--- a/ProcessLib/Process.h
+++ b/ProcessLib/Process.h
@@ -76,7 +76,8 @@ public:
 		DBUG("Initialize process.");
 
 		DBUG("Construct dof mappings.");
-		initializeMeshSubsets();
+		for (auto const& pv : _process_variables)
+			initializeMeshSubsets(pv);
 
 		_local_to_global_index_map.reset(
 		    new AssemblerLib::LocalToGlobalIndexMap(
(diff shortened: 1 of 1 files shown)


ProcessLib/Process.h
```diff
@@ -143,15 +144,21 @@ class Process
...
 
 		// Collect the mesh subsets in a vector.
-		_all_mesh_subsets.push_back(
-		    new MeshLib::MeshSubsets(_mesh_subset_all_nodes));
+		std::generate_n(
+		    std::back_inserter(_all_mesh_subsets),
+		    variable.getNumberOfComponents(),
+		    [&]()
+		    {
+			    return new MeshLib::MeshSubsets(_mesh_subset_all_nodes);
```
chleh: can we now also switch to unique_ptr?

chleh:
:fast_forward: 
Could you please at some point (e.g., in a new issue) collect, what has to be done until multi-component process-variables will be completely supported (e.g., reading BCs, ICs, output (there was something still missing)).

ProcessLib/Process.h
endJunction: This is a tricky one; Who has the ownership of the MeshSubsets, the Process or the DOF table?
It seems, like it should be the DOF table.

wenqing: Yes, should be DOF table.

chleh: But currently obviously `_all_mesh_subsets` owns them. So in the simplest case it is the question if you want to change the type of `_all_mesh_subsets` to `std::vector<std::unique_ptr<...>>`. Of course, then you also have to change some method signatures accordingly. Later on the meshsubsets may be owned by the DOF table. I don't care for now.
The problem I see is that most algorithms don't care if they work on a `std::vector<T*>` or on a `std::vector<std::unique_ptr<T>>` so maybe we should think about how to handle that without templating everything, but maybe I'm thinking too far.

bilke:
:+1:

endJunction:
Rebasing and merging then...

ProcessLib/Process.h
endJunction: This is WIP @ https://github.com/endJunction/ogs/tree/UniquePtrMeshSubsets

PR #1053, opened 2016-02-23
Avoid warning/error/debug messages in SurfaceGrid

TomFischer:
The construction of the class `SurfaceGrid` within the `Surface` is moved to the method `isPntInSfc()`. This avoids the repeatedly update of the surface grid and the warning/error/debug messages connected to the update. Since `SurfaceGrid` does not need the `Surface` as friend class a better encapsulation is realized.

**Update** In a run time test the new impl. took only ~77 % of the time the previous code needed.


rinkk:
:+1:

endJunction:
:fast_forward:

PR #1054, opened 2016-02-23
Fix clang sanitizer issues

TomFischer:

### DIFF
--- a/BaseLib/FileTools.cpp
+++ b/BaseLib/FileTools.cpp
@@ -56,6 +56,9 @@ void truncateFile( std::string const& filename)
 	ofs.close();
 }
 
+namespace
+{
+
 /** Finds the position of last file path separator.
  * Checks for unix or windows file path separators in given path and returns the
  * position of the last one or std::string::npos if no file path separator was
@@ -75,6 +78,7 @@ std::string::size_type findLastDot(std::string const& path)
 {
 	return path.find_last_of(".");
 }
+} // end namespace
 
 std::string dropFileExtension(std::string const& filename)
 {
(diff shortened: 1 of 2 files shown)


BaseLib/FileTools.cpp
```diff
@@ -62,7 +62,7 @@ void truncateFile( std::string const& filename)
  * found.
  */
 static
-std::size_t findLastPathSeparator(std::string const& path)
+std::string::size_type findLastPathSeparator(std::string const& path)
```
chleh: I think that function should not be used outside of this file since it is rather low-level. Is it used somewhere else?
If not, could you move it to an anonymous namespace?

BaseLib/FileTools.cpp
```diff
@@ -79,12 +79,12 @@ std::size_t findLastDot(std::string const& path)
 std::string dropFileExtension(std::string const& filename)
 {
 	// Look for dots in filename.
-	const std::size_t p = findLastDot(filename);
+	const std::string::size_type p = findLastDot(filename);
```
chleh: I'd prefer `auto` in all occasions.

chleh:
Only small things. Work on them or don't, it's up to you. :+1:

BaseLib/FileTools.cpp
rinkk: I don't like auto all that much because it obfuscates the real type (especially when the real type is not straightforward as in this case).

rinkk:
:+1:

BaseLib/FileTools.cpp
TomFischer: This function is declared/defined only in the cpp file. I moved it to an anonymous namespace.

PR #1056, opened 2016-02-24
Set log level at runtime

bilke:
- Added `-l` command line arguments to OGS cli and the testrunner. Possible values: `none`, `error`, `warn`, `info`, `debug`, `all`.
- `make tests` log-level is now `warn` (this suppresses verbose logging in [recent Jenkins builds](https://svn.ufz.de:8443/job/OGS-6/job/Tests_Linux/511/consoleFull)).

### DIFF
--- a/Applications/ApplicationsLib/LogogSetup.h
+++ b/Applications/ApplicationsLib/LogogSetup.h
@@ -12,6 +12,9 @@
 
 #include <logog/include/logog.hpp>
 
+#include <map>
+#include <memory>
+
 #include "BaseLib/LogogSimpleFormatter.h"
 
 namespace ApplicationsLib
@@ -25,21 +28,28 @@ public:
 	LogogSetup()
 	{
 		LOGOG_INITIALIZE();
-		fmt = new BaseLib::LogogSimpleFormatter;
-		logog_cout = new logog::Cout;
-		logog_cout->SetFormatter(*fmt);
+		logog_cout = std::unique_ptr<logog::Cout>(new logog::Cout);
+		SetFormatter(std::unique_ptr<BaseLib::LogogSimpleFormatter>
+			(new BaseLib::LogogSimpleFormatter));
 	}
 
 	~LogogSetup()
 	{
-		delete fmt;
-		delete logog_cout;
+		// Objects have to be deleted before shutdown
+		fmt.reset(nullptr);
+		logog_cout.reset(nullptr);
 		LOGOG_SHUTDOWN();
 	}
 
+	void SetFormatter(std::unique_ptr<logog::Formatter> formatter)
+	{
+		fmt = std::move(formatter);
+		logog_cout->SetFormatter(*fmt);
+	}
+
 private:
-	BaseLib::LogogSimpleFormatter* fmt;
-	logog::Cout* logog_cout;
+	std::unique_ptr<logog::Formatter> fmt;
+	std::unique_ptr<logog::Cout> logog_cout;
 };
 
 }	// ApplicationsLib
--- a/Applications/ApplicationsLib/LogogSetup.h
+++ b/Applications/ApplicationsLib/LogogSetup.h
@@ -47,6 +47,37 @@ public:
 		logog_cout->SetFormatter(*fmt);
 	}
 
+	void SetLevel(LOGOG_LEVEL_TYPE level)
+	{
+		logog::SetDefaultLevel(level);
+	}
+
+	void SetLevel(std::string const & level)
+	{
+		std::map<std::string, LOGOG_LEVEL_TYPE> foo =
+		{
+			{ "none", LOGOG_LEVEL_NONE },
+			{ "emergency", LOGOG_LEVEL_EMERGENCY },
+			{ "alert", LOGOG_LEVEL_ALERT},
+			{ "critical", LOGOG_LEVEL_CRITICAL },
+			{ "error", LOGOG_LEVEL_ERROR },
+			{ "warn", LOGOG_LEVEL_WARN },
+			{ "info", LOGOG_LEVEL_INFO },
+			{ "debug", LOGOG_LEVEL_DEBUG },
+			{ "all", LOGOG_LEVEL_ALL }
+		};
+
+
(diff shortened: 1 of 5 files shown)


Applications/CLI/ogs.cpp
```diff
@@ -79,7 +79,8 @@ int main(int argc, char *argv[])
 			"(http://www.opengeosys.org) "
 			"Distributed under a Modified BSD License. "
 			"See accompanying file LICENSE.txt or "
-			"http://www.opengeosys.org/project/license",
+			"http://www.opengeosys.org/project/license\n"
+			"version: " + BaseLib::BuildInfo::git_describe,
 		' ',
 		BaseLib::BuildInfo::git_describe);
```
endJunction: Just checking. Will the version number displayed only once?

bilke: Yes, the second `BaseLib::BuildInfo::git_describe` is only shown when running `ogs --version`.

Applications/ApplicationsLib/LogogSetup.h
```diff
@@ -25,21 +28,59 @@ class LogogSetup final
...
 	{
-		delete fmt;
-		delete logog_cout;
+		// Objects have to be deleted before shutdown
+		fmt.reset(nullptr);
+		logog_cout.reset(nullptr);
 		LOGOG_SHUTDOWN();
 	}
 
+	void SetFormatter(std::unique_ptr<logog::Formatter> formatter)
```
chleh: There should be `&&` added. Otherwise it will not work if you pass non-temporaries to that method.

Applications/ApplicationsLib/LogogSetup.h
```diff
@@ -25,21 +28,59 @@ class LogogSetup final
...
+		};
+
+
+		//LOGOG_LEVEL_TYPE level_type;
+		if(foo.find(level) != foo.end())
+			SetLevel(foo[level]);
+		else
+		{
+			ERR("%s is not a valid log level! Setting log level to all.", level.c_str());
+			SetLevel(LOGOG_LEVEL_ALL);
```
chleh: I'd prefer `std::abort()`.

Tests/testrunner.cpp
```diff
@@ -39,62 +39,70 @@
 /// Implementation of the googletest testrunner
 int main(int argc, char* argv[])
 {
+    std::string logLevel("all");
```
chleh: Maybe it would be better to read this setting from the environment, cf. `std::getenv`. This way we would not interfere with gtest's parameters.

chleh: Or: Why can't you use TCLAP here, too?

Tests/testrunner.cpp
```diff
@@ -39,62 +39,70 @@
...
+    for (int i = 1; i < argc; i++)
+    {
+        if(i + 1 == argc)
+            break;
+        if(std::strcmp(argv[i], "-l") == 0)
+            logLevel = argv[i + 1];
```
chleh: Don't we have to remove the two entries from `argv` now?

Tests/testrunner.cpp
```diff
@@ -39,62 +39,70 @@
...
+            (new BaseLib::TemplateLogogFormatterSuppressedGCC
+            <TOPIC_LEVEL_FLAG | TOPIC_FILE_NAME_FLAG | TOPIC_LINE_NUMBER_FLAG>()));
+    logog_setup.SetLevel(logLevel);
+
 
 #ifdef USE_PETSC
-        char help[] = "ogs6 with PETSc \n";
-        PetscInitialize(&argc,&argv,(char *)0,help);
+    char help[] = "ogs6 with PETSc \n";
+    PetscInitialize(&argc,&argv,(char *)0,help);
```
chleh: Btw: `const_cast<char*>("")` is better than `(char *)0`.

Tests/testrunner.cpp
```diff
@@ -39,62 +39,70 @@
...
-            ERR("Unknown exception occurred!");
-        }
-        // finalize libraries
+        // start google test
+        testing::InitGoogleTest ( &argc, argv );
+        ret = RUN_ALL_TESTS();
+    }
+    catch (char* e)
+    {
+        ERR(e);
```
chleh: Should be `ERR("%s", e);` since `e` could contain `%`.

Tests/testrunner.cpp
```diff
@@ -39,62 +39,70 @@
...
+    }
+    catch (std::exception& e)
+    {
+        ERR(e.what());
```
chleh: Same here.

Applications/CLI/ogs.cpp
```diff
@@ -99,6 +100,14 @@ int main(int argc, char *argv[])
 		"output directory");
 	cmd.add(outdir_arg);
 
+	TCLAP::ValueArg<std::string> log_level_arg(
+		"l", "log-level",
```
chleh: Do we need the shortcut `-l`? In general I'd not introduce them until OGS6 is kind of feature-complete.

chleh:
But in general I like the idea!

Tests/testrunner.cpp
bilke: TCLAP does not allow undefined parameters. gtest parameters then have to be passed after `--` which I don't like, e.g.:

```
./testrunner -l warn -- --gtest_filter=Foo
```

Tests/testrunner.cpp
bilke: I just changed indentation here.

Tests/testrunner.cpp
bilke: I don't know... I guess not because gtest does not have a `-l`-parameter.

Applications/CLI/ogs.cpp
bilke: I like the shortcut, I propose to eventually change if it clashes with some other parameter in the future. Btw. OGS will never be feature-complete...

TomFischer:
:+1:

PR #1057, opened 2016-02-25
Two small improvements/fixes

TomFischer:
- Fix clang sanitizer errors in TetGen interface
- Make code readable in MeshRevision

### DIFF
--- a/FileIO/TetGenInterface.cpp
+++ b/FileIO/TetGenInterface.cpp
@@ -132,7 +132,8 @@ bool TetGenInterface::parseSmeshFacets(std::ifstream &input,
 	const unsigned offset = (_zero_based_idx) ? 0 : 1;
 	std::vector<std::size_t> idx_map;
 
-	for (std::size_t k(0); k<nFacets && !input.fail(); k++)
+	std::size_t k(0);
+	while (k<nFacets && !input.fail())
 	{
 		getline (input, line);
 		if (input.fail())
@@ -144,7 +145,6 @@ bool TetGenInterface::parseSmeshFacets(std::ifstream &input,
 		BaseLib::simplify(line);
 		if (line.empty() || line.compare(0,1,"#") == 0)
 		{
-			k--;
 			continue;
 		}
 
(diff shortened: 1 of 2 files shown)


bilke:
:+1:

FileIO/TetGenInterface.cpp
```diff
@@ -132,7 +132,7 @@ bool TetGenInterface::parseSmeshFacets(std::ifstream &input,
 	const unsigned offset = (_zero_based_idx) ? 0 : 1;
 	std::vector<std::size_t> idx_map;
 
-	for (std::size_t k(0); k<nFacets && !input.fail(); k++)
+	for (int k(0); k<nFacets && !input.fail(); k++)
```
chleh: Don't you get a compiler waring now, because of `const std::size_t nFacets`?

FileIO/TetGenInterface.cpp
```diff
@@ -313,7 +313,7 @@ bool TetGenInterface::parseNodes(std::ifstream &ins,
 	double* coordinates (new double[dim]);
 	nodes.reserve(n_nodes);
 
-	for (std::size_t k(0); k < n_nodes && !ins.fail(); k++)
+	for (int k(0); k < n_nodes && !ins.fail(); k++)
```
chleh: same here.

chleh:
If no new compiler warnings then :fast_forward:

FileIO/TetGenInterface.cpp
TomFischer: Unfortunately a new warning occurs but the run time errors has gone. I will reconsider the simple fix again.

endJunction: I'd explicitly cast the `nFacets` to an `int` in the for-loop.

TomFischer: I tried to avoid cast. 

The problem of this specific code was that the variable k was used to control the number of read faceset which should be obvously not negative. But k was incremented in every iteration, even if not a faceset was read. I substituted the for loop by a while loop and only increment k when a faceset was read. I hope this fixes the problem.

FileIO/TetGenInterface.cpp
TomFischer: Substituted for loop by a while loop.

FileIO/TetGenInterface.cpp
chleh: That looks like a proper solution! :+1:

endJunction:
Much nicer solution indeed!

PR #1058, opened 2016-02-29
Added some conceptual documentation for the ODE solver lib

chleh:


TomFischer:
Thanks for docu. I wonder why the pdf and png files are part of the PR when they can be created from the tikz picture?

chleh:
I think, we don't want to require a complete LaTeX installation in order to be able to generate the documentation.

endJunction:
Rebase and merge after accepting my PR.

PR #1050, opened 2016-02-22
ODE Solver for transient and nonlinear processes – organ transplantation

chleh:
TODO
* [x] cleanup
* [x] docs
* [x] adapt input files

Future work:
* save some matrices (i.e. save space)
* adaptive time stepping
* different error norms
* numerical Jacobians
* process coupling
* ...

PR #1059, opened 2016-02-29
Added a matrix pool

### LINKED ISSUE #1067: GWF cube_5e4 ctest failing. std::bad_alloc
I still don't know why this happens. Here is the data:

In the ctor of `TimeDiscretizedODESystem` [line 226](https://github.com/ufz/ogs/blob/master/NumLib/ODESolver/TimeDiscretizedODESystem.h#L226) the `_M` matrix is constructed with size n=54872  and n_nonzero_columns=54872. The matrix is of type `MathLib::EigenMatrix::EigenMatrix`; see [line 51](https://github.com/ufz/ogs/blob/master/MathLib/LinAlg/Eigen/EigenMatrix.h#L51), where
```c++
_mat.reserve(Eigen::VectorXi::Constant(n, n_nonzero_columns));
```
is called. The reserve call goes into Eigen CompressedStorage, reserveInnerVectors, etc. until a bad_alloc:
```
 Eigen::internal::CompressedStorage<double, int>::reallocate (this=0x194e6808, size=3010936384) at /usr/include/eigen3/Eigen/src/SparseCore/CompressedStorage.h:207
```

The `VectorXi` sizes seems a little to large to me; 54872^2 = 3010936384 ~ 3e9 bytes...

chleh:
This matrix pool will make it possible to use global matrices and vectors in different processes. Furthermore, this PR adds builders for global matrices, which abstract away, e.g., the differences between PETSc and Eigen matrix creation. However, for PETSc there will be a follow-up PR.

Maybe we should discuss this PR as we did with #1040 and #1050.

Note: This is a follow-up PR of #1050!

To do:
* [x] add docu (for the members "matrix id")
* [x]  '(minor) cleanup
* [x] cpp file for UnifiedMatrixSetters
* [x] ~~discuss disabled unit tests put them to a separate PR~~ EDIT: this applies to #1060 
* [x] fix #1067 – fixed in ba4cc25c, @endJunction please have a look.
* [x] initialize matrix id members
* [x] make matrix and vector providers global objects, not members.

### DIFF
--- a/Applications/ApplicationsLib/UncoupledProcessesTimeLoop.h
+++ b/Applications/ApplicationsLib/UncoupledProcessesTimeLoop.h
@@ -31,6 +31,8 @@ class UncoupledProcessesTimeLoop
 public:
     bool loop(ProjectData& project, std::string const& outdir);
 
+    ~UncoupledProcessesTimeLoop();
+
 private:
     //! An abstract nonlinear solver
     using AbstractNLSolver = NumLib::NonlinearSolverBase<Matrix, Vector>;
@@ -40,6 +42,8 @@ private:
     using Process          = ProcessLib::Process<GlobalSetupType>;
     //! An abstract time discretization
     using TimeDisc         = NumLib::TimeDiscretization<Vector>;
+
+    std::vector<Vector*> _process_solutions;
 
     struct SingleProcessData
     {
@@ -88,7 +92,6 @@ private:
      *       to be extended slightly.
      */
     template<NumLib::ODESystemTag ODETag>
-    static
     SingleProcessData
     makeSingleProcessData(
             AbstractNLSolver& nonlinear_solver,
@@ -111,7 +114,8 @@ private:
             // because the Newton ODESystem derives from the Picard ODESystem.
             // So no further checks are needed here.
 
-            return SingleProcessData{ *nonlinear_solver_picard, time_disc, ode_sys };
+            return SingleProcessData{
+                *nonlinear_solver_picard, time_disc, ode_sys };
         }
         else if (auto* nonlinear_solver_newton =
                  dynamic_cast<NonlinearSolverNewton*>(&nonlinear_solver))
@@ -136,13 +140,12 @@ private:
         }
     }
 
-    //! Constructs and returns a \c vector of PerProcessData from the given
-    //! ProjectData.
+    //! Constructs and returns a \c vector of SingleProcessData from the given
+    //! \c project.
     std::vector<SingleProcessData> initInternalData(ProjectData& project);
 
-    //! Sets and returns initial conditions for the given ProjectData and
-    //! PerProcessData.
-    std::vector<Vector> setInitialConditions(
+    //! Sets initial conditions for the given \c project and \c per_process_data.
+    void setInitialConditions(
             ProjectData& project, double const t0,
             std::vector<SingleProcessData>& per_process_data);
 
@@ -238,7 +241,7 @@ initInternalData(ProjectData& project)
 
 
 template<typename Matrix, typename Vector>
-std::vector<Vector>
+void
 UncoupledProcessesTimeLoop<Matrix, Vector>::
 setInitialConditions(ProjectData& project,
                      double const t0,
@@ -247,8 +250,7 @@ setInitialConditions(ProjectData& project,
     auto const num_processes = std::distance(project.processesBegin(),
                                              project.processesEnd());
 
-    std::vector<Vector> process_solutions;
-    process_solutions.reserve(num_processes);
+    _process_solutions.reserve(num_processes);
 
     unsigned pcs_idx = 0;
     for (auto p = project.processesBegin(); p != project.processesEnd();
@@ -261,13 +263,11 @@ setInitialConditions(ProjectData& project,
         auto& ode_sys     =  *ppd.tdisc_ode_sys;
         auto const nl_tag =   ppd.nonlinear_solver_tag;
 
-        auto const num_eqs = ode_sys.getNumEquations();
+        // append a solution vector of suitable size
+        _process_solutions.emplace_back(
+                    &MathLib::GlobalVectorProvider<Vector>::provider.getVector(ode_sys));
 
-        // TODO maybe more is required for PETSc
-        // append a solution vector of size num_eqs
-        process_solutions.emplace_back(num_eqs);
-
-        auto& x0 = process_solutions[pcs_idx];
+        auto& x0 = *_process_solutions[pcs_idx];
         pcs.setInitialConditions(x0);
 
         time_disc.setInitialState(t0, x0); // push IC
@@ -282,8 +282,6 @@ setInitialConditions(ProjectData& project,
             time_disc.pushState(t0, x0, mat_strg); // TODO: that might do duplicate work
         }
     }
-
-    return process_solutions;
 }
 
 
@@ -328,7 +326,7 @@ loop(ProjectData& project, std::string const& outdir)
     auto const t0 = 0.0; // time of the IC
 
     // init solution storage
-    auto process_solutions = setInitialConditions(project, t0, per_process_data);
+    setInitialConditions(project, t0, per_process_data);
 
     auto const delta_t = 1.0;
 
@@ -353,7 +351,7 @@ loop(ProjectData& project, std::string const& outdir)
                     // + "_t_"   + std::to_string(t) // TODO: add that later
                     + ".vtu";
 
-            auto& x = process_solutions[pcs_idx];
+            auto& x = *_process_solutions[pcs_idx];
 
             nonlinear_solver_succeeded = solveOneTimeStepOneProcess(
                         timestep, x, t, delta_t,
@@ -374,6 +372,15 @@ loop(ProjectData& project, std::string const& outdir)
     return nonlinear_solver_succeeded;
 }
 
+
+template<typename Matrix, typename Vector>
+UncoupledProcessesTimeLoop<Matrix, Vector>::
+~UncoupledProcessesTimeLoop()
+{
+    for (auto * x : _process_solutions)
+        MathLib::GlobalVectorProvider<Vector>::provider.releaseVector(*x);
 }
+
+} // namespace ApplicationsLib
 
 #endif // APPLICATIONSLIB_UNCOUPLED_PROCESSES_TIMELOOP
--- a/MathLib/LinAlg/MatrixProviderUser.h
+++ b/MathLib/LinAlg/MatrixProviderUser.h
@@ -1,0 +1,131 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef MATHLIB_MATRIX_PROVIDER_USER_H
+#define MATHLIB_MATRIX_PROVIDER_USER_H
+
+#include <cstddef>
+
+namespace MathLib
+{
+
+struct MatrixSpecifications
+{
+    std::size_t const nrows;
+    std::size_t const ncols;
+};
+
+class MatrixSpecificationsProvider
+{
+public:
+    virtual MatrixSpecifications getMatrixSpecifications() const = 0;
+
+    virtual ~MatrixSpecificationsProvider() = default;
+};
+
+
+/*! Manages storage for vectors.
+ *
+ * This interface provides storage management semantics for vectors, which
+ * can be acquired at a certain point in time, be released later on and
+ * be acquired again etc.
+ *
+ * "Released" means that the caller indicates, that he currently does not need
+ * that \c Vector instance, i.e, Either it is not needed anymore for the entire
+ * program run, or it is temporarily not needed, but might be aquired again
+ * later on. Thereby the implementation of this interface can decide, whether
+ * the storage for the specific vector can be freed or reused in the meantime.
+ *
+ * All get-methods of this class come in two variants: One with an \c id argument,
+ * and one without. The latter makes it possible to temporarily release a vector
+ * during the program run and re-acquire it again. The former variant is intended
+ * as a short-cut that simplifies vector acquisition for callers that will use
+ * the same vector throughout all of their lifetime without releasing it
+ * intermediately.
+ *
+ * \attention
+ * The first time a vector is acquired by a method with \c id argument, the \c id
+ * has to be initialized with zero. The respective method then will set the \c id
+ * to the specific \c id of the returned vector.
+ */
+template<typename Vector>
+class VectorProvider
+{
+public:
+    using MSP = MatrixSpecificationsProvider;
+
+    //! Get an uninitialized vector.
+    virtual Vector& getVector() = 0;
+
+    //! Get an uninitialized vector with the given \c id.
+    virtual Vector& getVector(std::size_t& id) = 0;
+
+    //! Get a copy of \c x.
+    virtual Vector& getVector(Vector const& x) = 0;
+
+    //! Get a copy of \c x in the storage of the vector with the given \c id.
+    virtual Vector& getVector(Vector const& x, std::size_t& id) = 0;
+
+    //! Get a vector according to the given specifications.
+    virtual Vector& getVector(MSP const& msp) = 0;
+
+    //! Get a vector according to the given specifications in the storage
+    //! of the vector with the given \c id.
+    virtual Vector& getVector(MSP const& msp, std::size_t& id) = 0;
+
+    //! Release the given vector.
+    //!
+    //! \pre \c x must have been acquired before, i.e., you must not call this
+    //! method twice in a row in the same \c x!
+    virtual void releaseVector(Vector const& x) = 0;
+
+    virtual ~VectorProvider() = default;
+};
+
+/*! Manages storage for matrices.
+ *
+ * This the matrix-analog of VectorProvider. The same notes apply to this class.
+ */
+template<typename Matrix>
+class MatrixProvider
+{
+public:
+    using MSP = MatrixSpecificationsProvider;
+
+    //! Get an uninitialized matrix.
+    virtual Matrix& getMatrix() = 0;
+
+    //! Get an uninitialized matrix with the given \c id.
+    virtual Matrix& getMatrix(std::size_t& id) = 0;
+
+    //! Get a copy of \c A.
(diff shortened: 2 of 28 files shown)


MathLib/LinAlg/SimpleMatrixProvider-impl.h
```diff
@@ -0,0 +1,266 @@
...
+template<typename MatVec>
+void
+transfer(std::map<MatVec*, std::size_t>& from_used,
+         std::map<std::size_t, MatVec*>& to_unused,
+         typename std::map<MatVec*, std::size_t>::iterator it)
+{
+    auto& ptr = it->first;
+    auto const id = it->second;
+
+    auto res = to_unused.emplace(id, std::move(ptr));
```
endJunction: res is unused when compiling with NDEBUG.

MathLib/LinAlg/SimpleMatrixProvider.h
```diff
@@ -0,0 +1,78 @@
...
+    Vector& getVector() override;
+    Vector& getVector(std::size_t& id) override;
+
+    Vector& getVector(Vector const& x) override;
+    Vector& getVector(Vector const& x, std::size_t& id) override;
+
+    Vector& getVector(MSP const& msp) override;
+    Vector& getVector(MSP const& msp, std::size_t& id) override;
+
+    void releaseVector(Vector const& x);
```
endJunction: override

MathLib/LinAlg/SimpleMatrixProvider-impl.h
chleh: Note taken. But remember: This is WIP.

Applications/ApplicationsLib/UncoupledProcessesTimeLoop.h
```diff
@@ -142,7 +148,7 @@ class UncoupledProcessesTimeLoop
 
     //! Sets and returns initial conditions for the given ProjectData and
     //! PerProcessData.
-    std::vector<Vector> setInitialConditions(
+    void setInitialConditions(
```
TomFischer: Since now the method does not return anything the comment above should be changed, too.

MathLib/LinAlg/MatrixVectorTraits.h
```diff
@@ -0,0 +1,61 @@
+// TODO
```
TomFischer: What do you want "TODO"?

chleh:
```

++ git merge-base origin/master HEAD
+ git diff --check 284b10acc0a7d578415ee5afd55c45fa5384c676 HEAD
MathLib/LinAlg/GlobalMatrixProviders.cpp:46: new blank line at EOF.
Build step 'Execute shell' marked build as failure
ERROR: Failed to set Jenkins Identity header on email.
```
If the style check is causing Jenkins trouble, screw it!

endJunction:
@chleh Looks good. The petsc unit test fails...
```
[----------] 1 test from NumLibODEIntTyped/16, where TypeParam = TestCase<MathLib::PETScMatrix, MathLib::PETScVector, ODE1, NumLib::BackwardEuler, (NumLib::NonlinearSolverTag)1>
[ RUN      ] NumLibODEIntTyped/16.T1
[0]PETSC ERROR: ------------------------------------------------------------------------
[0]PETSC ERROR: Caught signal number 11 SEGV: Segmentation Violation, probably memory access out of range
```

chleh:
I think the problem is, that with the new global matrix and vector provider, PETSc is shut down (at the end of `main()`) before the destructor of the matrix provider is run (after the end of `main()`).
There are four possible solutions:
  1. A pre-shutdown hook that is called at the end of `main()` before PETSc shuts down
  2. The matrix provider is allocated inside `main()` and automatically cleaned up. The global matrix provider thus is initialized at the beginning of `main()`
  3. The matrix provider destructor for PETSc shuts down PETSc.
  4. Our PETSc setup object destroys the global matrix provider for PETSc matrices. Maybe then the PETSc setup should own the global PETSc matrix provider.

Case (4) is similar to (2). The pre-shutdown hook is probably implemented easiest. However, it can also be forgotten easiest.
Case (3) smells like bad design.
Cases (2) and (4) make the use of the global providers slightly more ugly: The won't be references anymore, but pointers.

Currently I think (1) is the best option. Going to implement it.

chleh:
Now LinearSolverLibrarySetup cleans up the global matrix providers.
Lesson learned: Global objects can lead to better design, however, they come with ugly side effects.
Still I think the current solution is better than passing the matrix providers manually everywhere.

chleh:
OK, that didn't fix this error. So I'm going to disable the tests for PETSc now. They'll be re-enabled in #1060.

chleh:
Rebased this on top of #1082. So only merge after that!

MathLib/LinAlg/MatrixVectorTraits.cpp
```diff
@@ -0,0 +1,166 @@
...
+#ifdef USE_PETSC
+
+namespace MathLib
+{
+
+std::unique_ptr<PETScMatrix>
+MatrixVectorTraits<PETScMatrix>::
+newInstance()
+{
+    return std::unique_ptr<PETScMatrix>(new PETScMatrix(0, 0)); // TODO default constructor
```
wenqing: The default constructor could generate nothing, e.g. skipping call MatCreate.

chleh:
Going to rebase again. Dear reviewers, sorry for that.

MathLib/LinAlg/MatrixVectorTraits.cpp
chleh: That's what will be done in #1060. Please don't be so strict with PETSc things in this PR.

chleh:
That was hopefully the last rebase!

chleh:
PETSc ctests fail. That's OK.

MathLib/LinAlg/MatrixProviderUser.h
```diff
@@ -0,0 +1,79 @@
...
+};
+
+template<typename Matrix, typename Vector>
+class MatrixProvider
+{
+public:
+    using MSP = MatrixSpecificationsProvider;
+
+    //! get an uninitialized Matrix (or the one with the given id)
+    virtual Matrix& getMatrix() = 0;
```
endJunction: virtual dtor missing for `MatrixProvider`.

chleh: Ah, missed it when removing the inheritance.

NumLib/ODESolver/NonlinearSolver-impl.h
```diff
@@ -149,6 +165,10 @@ solve(Vector &x)
         }
     }
 
+    MathLib::GlobalMatrixProvider<Matrix, Vector>::provider.releaseMatrix(J);
```
wenqing: These releasing functions appear more than once in different functions. Might be good to put them together in a single function.

MathLib/LinAlg/MatrixProviderUser.h
```diff
@@ -0,0 +1,79 @@
+/**
+ * \copyright
```
wenqing: A brief induction about the provider is needed. For example, it provides instances of global matrix for the system of linear equations.

MathLib/LinAlg/SimpleMatrixProvider.h
```diff
@@ -0,0 +1,90 @@
...
+#include<map>
+#include<memory>
+
+#include "MatrixProviderUser.h"
+
+namespace MathLib
+{
+
+template<typename Matrix, typename Vector>
+class SimpleMatrixProvider final
```
wenqing: It provides both matrix and vector. As I commented before, the class name doe not give the full meaning of the class.

wenqing:
:+1:

NumLib/ODESolver/NonlinearSolver-impl.h
chleh: Every matrix or vector is released as soon as it is (temporarily) not used anymore. Therefore the calls happen where they naturally fit, and not in a single place. Or am I not getting your point?

MathLib/LinAlg/SimpleMatrixProvider.h
chleh: Changed to `SimpleMatrixVectorProvider`

MathLib/LinAlg/MatrixVectorTraits.cpp
```diff
@@ -0,0 +1,166 @@
...
+    return std::unique_ptr<PETScVector>(new PETScVector(spec.nrows));
+}
+
+} // namespace MathLib
+
+
+#elif defined(OGS_USE_EIGEN)
+
+// #include "MathLib/LinAlg/Eigen/EigenMatrix.h"
+// #include "MathLib/LinAlg/Eigen/EigenVector.h"
```
TomFischer: Is there a special reason the commented lines should be kept?

MathLib/LinAlg/MatrixVectorTraits.cpp
```diff
@@ -0,0 +1,166 @@
...
+MatrixVectorTraits<PETScVector>::
+newInstance(MatrixSpecifications const& spec)
+{
+    return std::unique_ptr<PETScVector>(new PETScVector(spec.nrows));
+}
+
+} // namespace MathLib
+
+
+#elif defined(OGS_USE_EIGEN)
```
TomFischer: There are two section "#ifdef OGS_USE_EIGEN" in this file. Maybe it is possible and useful to merge the sections together?

MathLib/LinAlg/MatrixVectorTraits.h
```diff
@@ -0,0 +1,72 @@
...
+#include "MathLib/LinAlg/PETSc/PETScVector.h"
+
+namespace MathLib
+{
+SPECIALIZE_MATRIX_VECTOR_TRAITS(PETScMatrix, PETScMatrix::IndexType);
+SPECIALIZE_MATRIX_VECTOR_TRAITS(PETScVector, PETScVector::IndexType);
+}
+
+
+#elif defined(OGS_USE_EIGEN)
```
TomFischer: Same as in the implementation: Why not merge the two OGS_USE_EIGEN sections into one?

MathLib/LinAlg/MatrixVectorTraits.cpp
chleh: Forgot to remove them.

MathLib/LinAlg/SimpleMatrixVectorProvider-impl.h
```diff
@@ -0,0 +1,281 @@
...
+    return *getMatrix_<false>(id, mat_spec).first;
+    // TODO assert that the returned object always is of the right size
+}
+
+template<typename Matrix, typename Vector>
+Matrix&
+SimpleMatrixVectorProvider<Matrix, Vector>::
+getMatrix(MatrixSpecificationsProvider const& msp, std::size_t& id)
+{
+    auto mat_spec = msp.getMatrixSpecifications();
```
TomFischer: Should `mat_spec` be `const`?

MathLib/LinAlg/SimpleMatrixVectorProvider-impl.h
```diff
@@ -0,0 +1,281 @@
...
+    return *getMatrix_<true>(id, mat_spec).first;
+    // TODO assert that the returned object always is of the right size
+}
+
+template<typename Matrix, typename Vector>
+Matrix&
+SimpleMatrixVectorProvider<Matrix, Vector>::
+getMatrix(Matrix const& A)
+{
+    std::size_t id = 0u;
```
TomFischer: Is it possible to reuse the implementation below, i.e. `getMatrix(A, 0)`?

MathLib/LinAlg/MatrixVectorTraits.cpp
chleh: The first ifdef is for dense Eigen matrices. That case will be always built when we use the Eigen library (that's de facto always).
The second ifdef, or series of ifdefs (for PETSc or Eigen) resembles the ifdefs in NumericsConfig.h, i.e., only those methods that work with the global matrix and vector types are enabled.
Thus, there is an advantage from merging those sections (fewer ifdefs), but there's also a disadvantage: EigenMatrix case is always compiled. Hence, I'd like to keep the status quo.

MathLib/LinAlg/MatrixVectorTraits.h
chleh: Same answer as above.

MathLib/LinAlg/SimpleMatrixVectorProvider-impl.h
```diff
@@ -0,0 +1,281 @@
...
+        BLAS::copy(A, *res.first);
+    return *res.first;
+}
+
+template<typename Matrix, typename Vector>
+void
+SimpleMatrixVectorProvider<Matrix, Vector>::
+releaseMatrix(Matrix const& A)
+{
+    auto it = _used_matrices.find(const_cast<Matrix*>(&A));
```
TomFischer: Why is the `const_cast` necessary?

TomFischer: Why is the `const_cast` necessary?

MathLib/LinAlg/SimpleMatrixVectorProvider-impl.h
```diff
@@ -0,0 +1,281 @@
...
+    return *getVector_<true>(id).first;
+}
+
+template<typename Matrix, typename Vector>
+Vector&
+SimpleMatrixVectorProvider<Matrix, Vector>::
+getVector(MatrixSpecificationsProvider const& msp)
+{
+    std::size_t id = 0u;
+    auto const mat_spec = msp.getMatrixSpecifications();
```
TomFischer: Is `mat_spec` a copy or a reference?

MathLib/LinAlg/SimpleMatrixVectorProvider-impl.h
```diff
@@ -0,0 +1,281 @@
...
+    return *getVector_<false>(id, mat_spec).first;
+    // TODO assert that the returned object always is of the right size
+}
+
+template<typename Matrix, typename Vector>
+Vector&
+SimpleMatrixVectorProvider<Matrix, Vector>::
+getVector(MatrixSpecificationsProvider const& msp, std::size_t& id)
+{
+    auto mat_spec = msp.getMatrixSpecifications();
```
TomFischer: If my understanding is correct the `mat_spec` contains matrix/vector meta data. So `getVector()` should return a vector with the size specified in `mat_spec`. Why is this specification not const? For me the comment 2 lines below also indicates that `mat_spec` should be `const`?!

MathLib/LinAlg/SimpleMatrixVectorProvider-impl.h
chleh: Can be done, although `msp.getMatrixSpecifications();` returns a new object.

MathLib/LinAlg/SimpleMatrixVectorProvider-impl.h
chleh: No, it cannot be reused. But there's a bug in the implementation of `getMatrix(id, A);`

MathLib/LinAlg/SimpleMatrixVectorProvider-impl.h
chleh: Now you almost got me. ^^
Taking the address of a const ref returns a pointer to a const, but `_used_matrices` stores pointers to non-const, so `find` expects the latter. The automatic conversion between them fails.

MathLib/LinAlg/SimpleMatrixVectorProvider-impl.h
chleh: Going to turn all of this into `auto const&`.

MathLib/LinAlg/SimpleMatrixVectorProvider-impl.h
chleh: Going to constify.

MathLib/LinAlg/MatrixVectorTraits.cpp
TomFischer: Ok, I see.

MathLib/LinAlg/SimpleMatrixVectorProvider-impl.h
chleh: Cf. e8a6756415d8940bf9152d33043515177ba1e022

MathLib/LinAlg/UnifiedMatrixSetters.cpp
```diff
@@ -0,0 +1,179 @@
...
+#ifdef OGS_USE_EIGEN
+
+// Dense Eigen matrix/vector //////////////////////////////////////////
+
+namespace MathLib
+{
+
+void setMatrix(Eigen::MatrixXd& m,
+               Eigen::MatrixXd::Index const rows, Eigen::MatrixXd::Index const cols,
+               std::initializer_list<double> values)
```
TomFischer: I wonder why the values are not given by (`const`) reference?

chleh: I also wondered at first. I think, I even tried it, but it did not work. But the STL has similar signatures, cf. http://en.cppreference.com/w/cpp/container/vector/vector.

chleh: Probably it would work, but
see also http://stackoverflow.com/a/21868629.

TomFischer: I also looked at cppreference and found out the same.

NumLib/ODESolver/NonlinearSolver.h
```diff
@@ -167,20 +167,20 @@ class NonlinearSolver<Matrix, Vector, NonlinearSolverTag::Picard> final
...
  *
- * \param linear_solver the linear solver that will be used by the nonlinear
- *        solver
+ * \param linear_solver the linear solver that will be used by the nonlinear solver
  * \param config configuration settings
  *
- * \return a pair <tt>(nl_slv, tag)</tt> where \c nl_slv is the generated
- *         nonlinear solver instance and the \c tag indicates if it uses the
- *         Picard or Newton-Raphson method
+ * \return a pair <tt>(nl_slv, tag)</tt> where \c nl_slv is the generatored nonlinear
```
TomFischer: Did you mean generated?

TomFischer:
:+1:

NumLib/ODESolver/NonlinearSolver.h
chleh: No "generatored" sounds much more epic. ^^

chleh:
Gonna squash now.

chleh:
Thanks for your review!

Issue #1046, opened 2016-02-18
ctests: Allow multiple file to be compared with tester


endJunction:
Currently the `TESTER` (diff, numdiff, vtkdiff) can accept only one file for testing like
```cmake
AddTest(
    ...
    TESTER vtkdiff
    DIFF_DATA line_${mesh_size}_neumann_pcs_0_ts_1.vtu D1_left_N1_right pressure
    ...
)
```

It would be good to add support for multiple files (like for MPI) maybe by adding multiple `DIFF_DATA` lines...

chleh:
Closed by #1239.

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/1046).

PR #1060, opened 2016-02-29
Fix ODE solver lib with PETSc

chleh:
This is a follow-up PR of #1059.

To do
* [x] ~~docu~~
* [x] discuss ctest results
* [x] update test input files
* [x] put disabled unit tests to a separate PR
* [x] linear solver names
```xml
<petsc>
    <prefix>gw</prefix>
    <params>-ksp_type cg -pc_type bjacobi -ksp_rtol 1e-16 -ksp_max_it 10000</params>
 </petsc>
```

Subsequent PRs:
* remove name argument from linear solver constructors
* clean up matrix and vector classes

Biggest changes:
* implemented missing functionality for PETSc
* PETScMatrix/Vector now stores a pointer to matrix/vector. Thereby it is possible to default construct them.
* The option prefix for PETSc linear solvers is now directly obtained from the `<petsc>` config tree section.
* Extended matrix specifications, s.t. now they include the DOF table, the mesh and the sparsity pattern.

### DIFF
--- a/MathLib/LinAlg/BLAS.cpp
+++ b/MathLib/LinAlg/BLAS.cpp
@@ -50,28 +50,28 @@ void copy(PETScVector const& x, PETScVector& y)
 
 void scale(PETScVector& x, double const a)
 {
-    VecScale(x.getRawVector(), a);
+    VecScale(*x.getRawVector(), a);
 }
 
 // y = a*y + X
 void aypx(PETScVector& y, double const a, PETScVector const& x)
 {
     // TODO check sizes
-    VecAYPX(y.getRawVector(), a, x.getRawVector());
+    VecAYPX(*y.getRawVector(), a, *x.getRawVector());
 }
 
 // y = a*x + y
 void axpy(PETScVector& y, double const a, PETScVector const& x)
 {
     // TODO check sizes
-    VecAXPY(y.getRawVector(), a, x.getRawVector());
+    VecAXPY(*y.getRawVector(), a, *x.getRawVector());
 }
 
-// y = a*x + y
+// y = a*x + b*y
 void axpby(PETScVector& y, double const a, double const b, PETScVector const& x)
 {
     // TODO check sizes
-    VecAXPBY(y.getRawVector(), a, b, x.getRawVector());
+    VecAXPBY(*y.getRawVector(), a, b, *x.getRawVector());
 }
 
 // Explicit specialization
@@ -122,7 +122,8 @@ void matMult(PETScMatrix const& A, PETScVector const& x, PETScVector& y)
 {
     // TODO check sizes
     assert(&x != &y);
-    MatMult(A.getRawMatrix(), x.getRawVector(), y.getRawVector());
+    if (!y.getRawVector()) y.shallowCopy(x);
+    MatMult(A.getRawMatrix(), *x.getRawVector(), *y.getRawVector());
 }
 
 // v3 = A*v1 + v2
@@ -131,7 +132,18 @@ void matMultAdd(PETScMatrix const& A, PETScVector const& v1,
 {
     // TODO check sizes
     assert(&v1 != &v3);
-    MatMultAdd(A.getRawMatrix(), v1.getRawVector(), v2.getRawVector(), v3.getRawVector());
+    if (!v3.getRawVector()) v3.shallowCopy(v1);
+    MatMultAdd(A.getRawMatrix(), *v1.getRawVector(), *v2.getRawVector(), *v3.getRawVector());
+}
+
+void finalizeAssembly(PETScMatrix& A)
+{
+    A.finalizeAssembly(MAT_FINAL_ASSEMBLY);
+}
+
+void finalizeAssembly(PETScVector& x)
+{
+    x.finalizeAssembly();
 }
 
 }} // namespaces
--- a/MathLib/LinAlg/MatrixVectorTraits.cpp
+++ b/MathLib/LinAlg/MatrixVectorTraits.cpp
@@ -7,6 +7,7 @@
  *
  */
 
+#include "AssemblerLib/LocalToGlobalIndexMap.h"
 #include "MatrixVectorTraits.h"
 
 #ifdef OGS_USE_EIGEN
@@ -32,7 +33,10 @@ std::unique_ptr<Eigen::MatrixXd>
 MatrixVectorTraits<Eigen::MatrixXd>::
 newInstance(MatrixSpecifications const& spec)
 {
-    return std::unique_ptr<Eigen::MatrixXd>(new Eigen::MatrixXd(spec.nrows, spec.ncols));
+    auto const nrows = spec.dof_table ? spec.dof_table->dofSize() : spec.nrows;
+    auto const ncols = spec.dof_table ? nrows : spec.ncols;
+
+    return std::unique_ptr<Eigen::MatrixXd>(new Eigen::MatrixXd(nrows, ncols));
 }
 
 std::unique_ptr<Eigen::VectorXd>
@@ -53,7 +57,9 @@ std::unique_ptr<Eigen::VectorXd>
 MatrixVectorTraits<Eigen::VectorXd>::
 newInstance(MatrixSpecifications const& spec)
 {
-    return std::unique_ptr<Eigen::VectorXd>(new Eigen::VectorXd(spec.nrows));
+    auto const nrows = spec.dof_table ? spec.dof_table->dofSize() : spec.nrows;
+
+    return std::unique_ptr<Eigen::VectorXd>(new Eigen::VectorXd(nrows));
 }
 
 } // namespace MathLib
@@ -63,6 +69,8 @@ newInstance(MatrixSpecifications const& spec)
 
 #ifdef USE_PETSC
 
+#include "MeshLib/NodePartitionedMesh.h"
+
 namespace MathLib
 {
 
@@ -70,7 +78,7 @@ std::unique_ptr<PETScMatrix>
 MatrixVectorTraits<PETScMatrix>::
 newInstance()
 {
-    return std::unique_ptr<PETScMatrix>(new PETScMatrix(0, 0)); // TODO default constructor
+    return std::unique_ptr<PETScMatrix>(new PETScMatrix);
 }
 
 std::unique_ptr<PETScMatrix>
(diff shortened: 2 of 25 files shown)


endJunction:
Jenkins, test this please.

Jenkins was stuck, stopped the job.

chleh:
For the MacPR: I probalby messed up Tests/CMakeLists.txt. @bilke, could you please tell me how to do it properly?

MathLib/LinAlg/BLAS.cpp
```diff
@@ -50,28 +50,28 @@ void copy(PETScVector const& x, PETScVector& y)
...
 
 // y = a*x + y
 void axpy(PETScVector& y, double const a, PETScVector const& x)
 {
     // TODO check sizes
-    VecAXPY(y.getRawVector(), a, x.getRawVector());
+    VecAXPY(*y.getRawVector(), a, *x.getRawVector());
 }
 
 // y = a*x + y
```
TomFischer: `b` is missing in the comment

MathLib/LinAlg/MatrixVectorTraits.cpp
```diff
@@ -105,7 +138,16 @@ std::unique_ptr<PETScVector>
 MatrixVectorTraits<PETScVector>::
 newInstance(MatrixSpecifications const& spec)
 {
-    return std::unique_ptr<PETScVector>(new PETScVector(spec.nrows));
+    auto const is_global_size = false;
+
+    if (spec.dof_table) {
+        auto& dt = *spec.dof_table;
```
TomFischer: Why not `auto const& dt = ...`?

chleh: Actually, it will be const automatically, because we dereference a `const*const`. But I can change it.

MathLib/LinAlg/PETSc/PETScVector.h
```diff
@@ -225,29 +223,32 @@ class PETScVector
...
+            VecAXPY(*_v, -1.0, *v_in._v);
         }
 
 
         // TODO preliminary
-        PETSc_Vec& getRawVector() {return _v; }
+        PETSc_Vec* getRawVector() { return _v.get(); }
 
         // TODO preliminary
         // this method is dangerous insofar you can do arbitrary things also
```
TomFischer: Why not make this a doxygen comment?

MathLib/LinAlg/UnifiedMatrixSetters.cpp
```diff
@@ -83,25 +83,35 @@ double norm(PETScVector const& x)
...
 
 void setMatrix(PETScMatrix& m,
                PETScMatrix::IndexType const rows,
                PETScMatrix::IndexType const cols,
                std::initializer_list<double> values)
 {
-    (void) m; (void) rows; (void) cols; (void) values;
-    // TODO implement
+    m.setZero();
+    addToMatrix(m, rows, cols, values);
```
TomFischer: Is there somewhere a check that the number/size of values is equal to `rows*cols` and this fits to the matrix dimensions?

MathLib/LinAlg/UnifiedMatrixSetters.cpp
```diff
@@ -83,25 +83,35 @@ double norm(PETScVector const& x)
...
 {
+    m.setZero();
+    std::vector<PETScVector::IndexType> rows(tmp.rows());
+    std::vector<PETScVector::IndexType> cols(tmp.cols());
+
+    std::iota(rows.begin(), rows.end(), 0);
+    std::iota(cols.begin(), cols.end(), 0);
+
+    // PETSc wants row-major
+    Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> tmp_ = tmp;
```
TomFischer: Maybe it is possible to check if `tmp` is already in `RowMajor`? Then the copy would not be necessary.

chleh: AFAIK `Eigen::MatrixXd` is __always__ column major.

MathLib/LinAlg/UnifiedMatrixSetters.cpp
chleh: There is an assertion in `addToMatrix()`. However, that's disabled in release mode. Does that suffice?

MathLib/LinAlg/UnifiedMatrixSetters.cpp
```diff
@@ -83,25 +83,35 @@ double norm(PETScVector const& x)
...
+}
 
 void setMatrix(PETScMatrix& m,
                PETScMatrix::IndexType const rows,
                PETScMatrix::IndexType const cols,
                std::initializer_list<double> values)
 {
-    (void) m; (void) rows; (void) cols; (void) values;
-    // TODO implement
+    m.setZero();
```
chleh: TODO: Check if after `setZero()` one has to call `finalizeAssembly()`.

MathLib/LinAlg/UnifiedMatrixSetters.cpp
chleh: There is no check if it fits the matrix dimensions, though. Do we have to check that?

MathLib/LinAlg/UnifiedMatrixSetters.cpp
TomFischer: You are right. The Eigen docu states that the default is column major.

MathLib/LinAlg/UnifiedMatrixSetters.cpp
TomFischer: > There is no check if it fits the matrix dimensions, though. Do we have to check that?

Is it possible that the values for `rows` and `cols` can be set independently from the matrix dimensions? How big is the potential to misuse the interface?

> There is an assertion in addToMatrix(). However, that's disabled in release mode. Does that suffice?

Do you expect enabling the check in release mode will be run time critical? If not, I would have a better feeling with enable check.

MathLib/LinAlg/UnifiedMatrixSetters.cpp
chleh: No problem, it is not obvious from the type name.

MathLib/LinAlg/UnifiedMatrixSetters.cpp
chleh: First of all, those methods are currently intended for use in tests. And I think, there is the only place where they are used.
Yes, matrix dims can be set independently.
The potential for misuse is currently restricted to tests.
Since currently only tests are affected, the check is not time-critical. However, it is common that bounds checks are disabled in release mode. So I think assertions suffice.
I'll add checks for the matrix bounds, then the size arguments can also be dropped.

chleh:
I expect everything to run, now.
@wenqing, thanks for pointing me to the wrongly set PETSc options.

MathLib/LinAlg/UnifiedMatrixSetters.cpp
chleh: The PETSc docu does not say much about that. Cf. http://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Mat/MatZeroEntries.html.
So let's hope for the best. ^^

endJunction:
Looks OK.

Remark: The output of the unit-tests could be reduced, 628KiB is way too much for 417 tests.
The diff of the petsc tests fails for me, but this is related to ufz/ogs#1046; vtkdiff/numdiff would be more suitable for the comparison than diff.

MathLib/LinAlg/UnifiedMatrixSetters.cpp
wenqing: It is not needed. finalizeAssembly() has to be called only if entries are changed.

MathLib/LinAlg/MatrixVectorTraits.cpp
```diff
@@ -84,7 +92,32 @@ std::unique_ptr<PETScMatrix>
 MatrixVectorTraits<PETScMatrix>::
 newInstance(MatrixSpecifications const& spec)
 {
-    return std::unique_ptr<PETScMatrix>(new PETScMatrix(spec.nrows)); // TODO sparsity pattern
+    auto const nrows = spec.dof_table ? spec.dof_table->dofSizeLocal() : spec.nrows;
+    auto const ncols = spec.dof_table ? nrows : spec.ncols;
+
+    // TODO I guess it is not hard to make AssemblerLib::computeSparsityPattern()
```
wenqing: In PETSc, sparsity of the global matrix can be precisely or roughly allocated (e.g. by calling MatMPIAIJSetPreallocation, MatSeqAIJSetPreallocation).

MathLib/LinAlg/MatrixVectorTraits.cpp
```diff
@@ -84,7 +92,32 @@ std::unique_ptr<PETScMatrix>
...
+    //      work also for NodePartitionedMesh'es.
+    assert(((!spec.sparsity_pattern) || spec.sparsity_pattern->size() == 0u) &&
```
wenqing: Would like to remove this assert here.

MathLib/LinAlg/PETSc/PETScLinearSolver.cpp
```diff
@@ -21,29 +21,42 @@
...
+PETScLinearSolver::PETScLinearSolver(const std::string /*prefix*/,
                                      BaseLib::ConfigTree const* const option)
-    : _elapsed_ctime(0.)
 {
-    // Insert options into petsc database if any.
-    if (option)
-    {
+    // Insert options into petsc database. Default options are given in the string below.
+    std::string petsc_options
+            = "-ksp_type cg -pc_type bjacobi -ksp_rtol 1e-16 -ksp_max_it 10000";
```
wenqing: Would like to use bicg as default because `` MatZeroRows() `` is used for applying Dirichlet BC. Only for some simple case, cg can give converged solution. Or we use more complicated cg setting for `` MatZeroRows() ``
``` 
 -ksp_type preonly -pc_type redistribute -redistribute_ksp_type cg -redistribute_pc_type bjacobi -redistribute_sub_pc_type icc 
```
In future, we could consider an option for either using `` MatZeroRows() `` or ``MatZeroRowsColumns()  ``

MathLib/LinAlg/PETSc/PETScVector.cpp
```diff
@@ -217,18 +212,34 @@ void PETScVector::viewer(const std::string &file_name, const PetscViewerFormat v
...
 
+void PETScVector::shallowCopy(const PETScVector &v)
+{
+    destroy();
+
+    _v.reset(new PETSc_Vec);
+
+    VecDuplicate(*v._v, _v.get());
+
+    // TODO can't that be copied from v?
```
wenqing: VecDuplicate is called, then the storage structures of two vectors are the same. The following variables can be assigned to the new vector directly.

MathLib/LinAlg/UnifiedMatrixSetters.cpp
```diff
@@ -83,34 +90,68 @@ double norm(PETScVector const& x)
...
+
+    auto const rows = m.getNRows();
+    auto const cols = m.getNCols();
+
+    assert((IndexType) values.size() == rows*cols);
+
+    Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> tmp(rows, cols);
+
+    auto it = values.begin();
+    for (IndexType r=0; r<rows; ++r) {
```
wenqing: Would it be better to skip this sort of copying.

wenqing:
:+1:

MathLib/LinAlg/MatrixVectorTraits.cpp
chleh: I think, we should go for precise preallocation in the future, since in our FEM processes we can easily access that information anyway.

MathLib/LinAlg/MatrixVectorTraits.cpp
chleh: Currently, if I remember correctly, we don't precompute a sparsity pattern in the PETSc build. Rather we get it from the node partitioned mesh in this method. I think this assertion should stay as long as we don't precompute precise sparsity patterns for PETSc.

MathLib/LinAlg/PETSc/PETScLinearSolver.cpp
chleh: Note taken.

MathLib/LinAlg/PETSc/PETScVector.cpp
chleh: Note taken.

MathLib/LinAlg/UnifiedMatrixSetters.cpp
chleh: Sure, it would be. But this file is currently only used within very small test cases. So, performance is not an issue. When reworking the matrix/vector classes there will have to be a different solution for this anyway.

endJunction:
Don't forget to merge Tests/Data repo.

Issue #1064, opened 2016-03-01
error in downloading boost


norihiro-w:
Hi, I tried to build the latest ogs but failed due to the below error. 

```
[  0%] Performing download step (download, verify and extract) for 'Boost'
-- downloading...
     src='http://sourceforge.net/projects/boost/files/boost/1.56.0/boost_1_56_0.tar.bz2/download'
     dst='/home/localadmin/ogs/ogs6/ogs6-nw-merged/BuildAllRelease/External/boost/src/boost_1_56_0.tar.bz2'
     timeout='none'
-- [download 100% complete]
CMake Error at Boost-stamp/download-Boost.cmake:27 (message):
  error: downloading
  'http://sourceforge.net/projects/boost/files/boost/1.56.0/boost_1_56_0.tar.bz2/download'
  failed

    status_code: 1
    status_string: "Unsupported protocol"
    log: Hostname was NOT found in DNS cache
    Trying 216.34.181.60...

  Connected to sourceforge.net (216.34.181.60) port 80 (#0)

  GET /projects/boost/files/boost/1.56.0/boost_1_56_0.tar.bz2/download
  HTTP/1.1

  User-Agent: curl/7.38.0

  Host: sourceforge.net

  Accept: */*

  

  HTTP/1.1 301 Moved Permanently

  Server nginx is not blacklisted

  Server: nginx

  Date: Tue, 01 Mar 2016 13:22:42 GMT

  Location:
  https://sourceforge.net/projects/boost/files/boost/1.56.0/boost_1_56_0.tar.bz2/download


  X-Content-Type-Options: nosniff

  Content-Length: 386

  Content-Type: text/html; charset=UTF-8

  Via: 1.1 wsa.ufz.de:80 (Cisco-WSA/9.0.1-162)

  Connection: keep-alive

  

  Ignoring the response-body

  <html>

   <head>
    <title>301 Moved Permanently</title>
   </head>
   <body>
    <h1>301 Moved Permanently</h1>
    The resource has been moved to <a href="https://sourceforge.net/projects/boost/files/boost/1.56.0/boost_1_56_0.tar.bz2/download">https://sourceforge.net/projects/boost/files/boost/1.56.0/boost_1_56_0.tar.bz2/download</a>;

  you should be redirected automatically.

  

  

   </body>

  </html>Connection #0 to host sourceforge.net left intact

  Issue another request to this URL:
[...]

PR #1065, opened 2016-03-01
Use https for thirdparty downloads.

endJunction:
~~I hope this will fix the download issues.~~
Using https for thirdparty downloads. This avoids some redirects.

Does not fix issue https://github.com/ufz/ogs/issues/1064, because the problem is not the link, but the downloader.

### DIFF
--- a/scripts/cmake/ThirdPartyLibVersions.cmake
+++ b/scripts/cmake/ThirdPartyLibVersions.cmake
@@ -1,13 +1,13 @@
 set(OGS_BOOST_VERSION 1.56.0)
 string(REPLACE "." "_" OGS_BOOST_VERSION_UNDERLINE ${OGS_BOOST_VERSION})
-set(OGS_BOOST_URL "http://sourceforge.net/projects/boost/files/boost/${OGS_BOOST_VERSION}/boost_${OGS_BOOST_VERSION_UNDERLINE}.tar.bz2/download")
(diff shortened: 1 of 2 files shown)


chleh:
WTF?

```
  error: downloading
  'https://opengeosys.s3.amazonaws.com/ogs6-lib-sources/eigen-3.2.5.tar.gz'
  failed

    status_code: 1
    status_string: "unsupported protocol"
    log: libcurl was built with SSL disabled, https: not supported!

  unsupported protocol
```
from https://svn.ufz.de:8443/job/OGS-6/job/Linux-PRs/1345/consoleFull#-666270100568672ba-e572-4760-810b-dec0ad03aa21

norihiro-w:
It seems the error happens because CMake file(DOWNLOAD) command doesn't support https by default. CMake doc says "For https:// URLs CMake must be built with OpenSSL support. TLS/SSL certificates are not checked by default. Set TLS_VERIFY to ON to check certificates and/or use EXPECTED_HASH to verify downloaded content. If neither TLS option is given CMake will check variables CMAKE_TLS_VERIFY and CMAKE_TLS_CAINFO, respectively." 

see https://cmake.org/cmake/help/v3.2/command/file.html

Someone says the behaviour changed since v3.2.1...
 http://stackoverflow.com/questions/29816529/unsupported-protocol-while-downlod-tar-gz-package

endJunction:
Correct fix is ufz/ogs#1086

PR #1066, opened 2016-03-01
Unclutter Jenkins error parser

chleh:
With the changes introduces Jenkins should no longer point us to, e.g.
```
***error: det|J|=-1.000000e+00 is not positive.
```


TomFischer:
:+1:

PR #1068, opened 2016-03-02
Generalized removeMeshElements tool.

TomFischer:

### DIFF
--- a/Applications/Utils/MeshEdit/removeMeshElements.cpp
+++ b/Applications/Utils/MeshEdit/removeMeshElements.cpp
@@ -28,6 +28,19 @@
 #include "MeshLib/MeshSearch/ElementSearch.h"
 #include "MeshLib/MeshEditing/RemoveMeshComponents.h"
 
+template <typename PROPERTY_TYPE>
+void searchByProperty(std::string const& property_name,
+                      std::vector<PROPERTY_TYPE> const& property_values,
+                      MeshLib::ElementSearch& searcher)
+{
+	for (auto const& property_value : property_values) {
+		const std::size_t n_marked_elements =
+		    searcher.searchByPropertyValue(property_value, property_name);
+		INFO("%d elements with property value %s found.", n_marked_elements,
+		     std::to_string(property_value).c_str());
+	}
+}
+
 int main (int argc, char* argv[])
 {
 	ApplicationsLib::LogogSetup logog_setup;
@@ -60,9 +73,15 @@ int main (int argc, char* argv[])
 	TCLAP::MultiArg<std::string> eleTypeArg("t", "element-type",
 	                                      "element type to be removed", false, "element type");
 	cmd.add(eleTypeArg);
-	TCLAP::MultiArg<unsigned> matIDArg("m", "material-id",
-	                                      "material id", false, "material id");
-	cmd.add(matIDArg);
+
+	TCLAP::MultiArg<int> int_property_arg("", "int-property-value",
+	                                      "new property value (data type int)",
+	                                      false, "number");
+	cmd.add(int_property_arg);
+	TCLAP::ValueArg<std::string> property_name_arg(
+	    "n", "property-name", "name of property in the mesh", false,
+	    "MaterialIDs", "string");
+	cmd.add(property_name_arg);
 
 	// I/O params
 	TCLAP::ValueArg<std::string> mesh_out("o", "mesh-output-file",
@@ -95,12 +114,10 @@ int main (int argc, char* argv[])
 			INFO("%d %s elements found.", n_removed_elements, typeName.c_str());
 		}
 	}
-	if (matIDArg.isSet()) {
-		const std::vector<unsigned> vec_matID = matIDArg.getValue();
-		for (auto matID : vec_matID) {
-			const std::size_t n_removed_elements = ex.searchByMaterialID(matID);
-			INFO("%d elements with material ID %d found.", n_removed_elements, matID);
-		}
+
+	if (int_property_arg.isSet()) {
+		searchByProperty(property_name_arg.getValue(),
+		                 int_property_arg.getValue(), searcher);
 	}
 
 	if (xSmallArg.isSet() || xLargeArg.isSet() ||
--- a/Applications/Utils/MeshEdit/removeMeshElements.cpp
+++ b/Applications/Utils/MeshEdit/removeMeshElements.cpp
@@ -11,33 +11,26 @@
  *              http://www.opengeosys.org/LICENSE.txt
  */
 
+#include <memory>
+
 // TCLAP
 #include "tclap/CmdLine.h"
 
-// ThirdParty/logog
-#include "logog/include/logog.hpp"
+#include "Applications/ApplicationsLib/LogogSetup.h"
 
-// BaseLib
-#include "LogogSimpleFormatter.h"
+#include "FileIO/readMeshFromFile.h"
+#include "FileIO/writeMeshToFile.h"
 
-// FileIO
-#include "Legacy/MeshIO.h"
-#include "readMeshFromFile.h"
-
-// MeshLib
-#include "Mesh.h"
+#include "MeshLib/Mesh.h"
 #include "MeshLib/Node.h"
-#include "Elements/Element.h"
-#include "MeshEnums.h"
-#include "MeshSearch/ElementSearch.h"
-#include "MeshEditing/RemoveMeshComponents.h"
+#include "MeshLib/Elements/Element.h"
+#include "MeshLib/MeshEnums.h"
+#include "MeshLib/MeshSearch/ElementSearch.h"
+#include "MeshLib/MeshEditing/RemoveMeshComponents.h"
 
 int main (int argc, char* argv[])
 {
-	LOGOG_INITIALIZE();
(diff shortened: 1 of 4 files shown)


MeshLib/MeshSearch/ElementSearch.h
```diff
@@ -13,7 +13,10 @@
 #include <limits>
 #include <vector>
 
+#include <boost/optional.hpp>
```
chleh: Do you have to include that manually? Isn't that implicitly done from the definition of `_mesh.getProperties().getPropertyVector<PROPERTY_TYPE>()` ?

MeshLib/MeshSearch/ElementSearch.h
```diff
@@ -31,8 +34,41 @@ class ElementSearch final
...
+	    std::string const& property_name = "MaterialIDs")
+	{
+		boost::optional<MeshLib::PropertyVector<PROPERTY_TYPE> const&> opt_pv(
+		    _mesh.getProperties().getPropertyVector<PROPERTY_TYPE>(
+		        property_name));
+		if (!opt_pv)
+			return 0;
+
+		MeshLib::PropertyVector<PROPERTY_TYPE> const& pv(opt_pv.get());
+		if (pv.getMeshItemType() != MeshLib::MeshItemType::Cell)
```
chleh: Isn't that worth an error message?

Applications/Utils/MeshEdit/removeMeshElements.cpp
```diff
@@ -11,35 +11,44 @@
...
+#include "MeshLib/MeshEnums.h"
+#include "MeshLib/MeshSearch/ElementSearch.h"
+#include "MeshLib/MeshEditing/RemoveMeshComponents.h"
+
+template <typename PROPERTY_TYPE>
+void searchByProperty(std::string const& property_name,
+                      std::vector<PROPERTY_TYPE> const& property_values,
+                      MeshLib::ElementSearch& searcher)
+{
+	for (auto property_value : property_values) {
```
chleh: `auto const&` in order to avoid copies.

Applications/Utils/MeshEdit/removeMeshElements.cpp
```diff
@@ -11,35 +11,44 @@
...
+		const std::size_t n_removed_elements =
```
chleh: why `n_removed_elements` if the function name is `searchByProperty`?

Applications/Utils/MeshEdit/removeMeshElements.cpp
```diff
@@ -11,35 +11,44 @@
...
+		    searcher.searchByPropertyValue(property_value, property_name);
+		INFO("%d elements with property value %d found.", n_removed_elements,
```
chleh: `%d` is problematic since `std::size_t` can be `unsigned long`, then one would need `%lu` in a strict sense. However, that problem is also present in many other locations. Ignore it here, but we need some type-safe logger.
One could workaround this by writing
```
INFO("...%s ...", std::to_string(n_removed_elements));
```
But that's ugly.

Applications/Utils/MeshEdit/removeMeshElements.cpp
```diff
@@ -11,35 +11,44 @@
...
+		     property_value);
```
chleh: Moreover, `%d` does not work if `property_value` is, e.g. a double.

Applications/Utils/MeshEdit/removeMeshElements.cpp
```diff
@@ -67,9 +76,19 @@ int main (int argc, char* argv[])
 	TCLAP::MultiArg<std::string> eleTypeArg("t", "element-type",
 	                                      "element type to be removed", false, "element type");
 	cmd.add(eleTypeArg);
-	TCLAP::MultiArg<unsigned> matIDArg("m", "material-id",
-	                                      "material id", false, "material id");
-	cmd.add(matIDArg);
+
+	TCLAP::MultiArg<char> char_property_arg(
```
chleh: Only out of curiosity: Are those single character properties? What are they good for?

Applications/Utils/MeshEdit/removeMeshElements.cpp
```diff
@@ -101,12 +121,14 @@ int main (int argc, char* argv[])
...
-	if (matIDArg.isSet()) {
-		const std::vector<unsigned> vec_matID = matIDArg.getValue();
-		for (auto matID : vec_matID) {
-			const std::size_t n_removed_elements = ex.searchByMaterialID(matID);
-			INFO("%d elements with material ID %d found.", n_removed_elements, matID);
-		}
+
+	if (char_property_arg.isSet()) {
+		searchByProperty(property_name_arg.getValue(),
+		                 char_property_arg.getValue(), ex);
```
chleh: Not this PRs fault, but `ex` is a poor name for an instance of `MeshLib::ElementSearch`.

chleh:
Only small things. Afterwards :+1:

Applications/Utils/MeshEdit/removeMeshElements.cpp
TomFischer: Character properties can be written from the tool `ResetPropertyInPolygonalRegion`. For this reason I added these property type here.

chleh: But only for my understanding: Are they strings or single characters?

MeshLib/MeshSearch/ElementSearch.h
TomFischer: I think a warning is enough here. Added **warning** message.

MeshLib/MeshSearch/ElementSearch.h
TomFischer: Removed the include.

Applications/Utils/MeshEdit/removeMeshElements.cpp
TomFischer: Changed to `n_marked_elements` what does better refect the purpose of the variable.

Applications/Utils/MeshEdit/removeMeshElements.cpp
TomFischer: Fixed using `std::to_string`.

Applications/Utils/MeshEdit/removeMeshElements.cpp
TomFischer: Single characters.

rahv:
Two things:
* the utility specifies params for int and char and it would probably be minimal additional work to add double-params to that list. Since double props are effectively used, I would appreciate if you could add such a param.
* there is some logic error in the bounding box calculation. this has nothing to do with the recent changes but you might want to fix it or we'd have to create an issue on that. it goes like this: you don't need to specify a complete bounding box for removing elements but can instead specify to only remove all elements with ```z_min <  n``` or sth like that. the searcher should now check if all params for aabb have been set and if not it should use the current aabb-value. if i see this correctly, the utility would currently crash if you didn't specify all six aabb params. (also keep in mind that user shouldn't need to specify all values as it makes the handling very difficult. so replacing the ```||``` with ```&&``` would technically solve the problem but make tool essentially unusable for removing elements based on aabb.)

chleh:
@rinkk, please save your note in an issue.

TomFischer:
Rebased to master, merged review changes and removed the possibilty to give char arguments on the command line.

chleh:
@TomFischer, please try to not rebase during review. Especially, please do not squash any commits when you want others to review the changes you made in response to reviewers' comments. Because: After rebasing it is not visible anymore, what you did in response, so we will have to read everything again.
Of course, there might be cases where it does not hurt to squash things during review, e.g., when the post-comment changes are so many that they fundamentally change the PR. However, I don't know it that applies to this PR.

chleh:
> removed the possibilty to give char arguments on the command line.
Could you please point me to where you removed that? I cannot find the line by myself.
You addressed all of my questions. I believe you that you actually did, what you answered. :+1:

endJunction:
@chleh Sorry about rebase, I thought it can be merged. Mea culpa.

chleh:
@endJunction I thought "rebase and merge" is for marking that case. ^^

TomFischer:
Is there something open in this PR?

chleh:
I guess not. Do we need to save @rinkk's comment?

> Two things:
* the utility specifies params for int and char and it would probably be minimal additional work to add double-params to that list. Since double props are effectively used, I would appreciate if you could add such a param.
* there is some logic error in the bounding box calculation. this has nothing to do with the recent changes but you might want to fix it or we'd have to create an issue on that. it goes like this: you don't need to specify a complete bounding box for removing elements but can instead specify to only remove all elements with ```z_min <  n``` or sth like that. the searcher should now check if all params for aabb have been set and if not it should use the current aabb-value. if i see this correctly, the utility would currently crash if you didn't specify all six aabb params. (also keep in mind that user shouldn't need to specify all values as it makes the handling very difficult. so replacing the ```||``` with ```&&``` would technically solve the problem but make tool essentially unusable for removing elements based on aabb.)

TomFischer:
@chleh I saved the second point of @rinkk 's comment in issue #1089. The first point is a feature request that might be interesting in the future. I would postpone an implementation the the point in time the feature will be needed.

PR #1075, opened 2016-03-03
Fix surface creation bugs in SHPInterface

TomFischer:
- First bug was that empty polygons were created and the AABB of an empty point set is not allowed.
- Second bug was inproper handling parallel line segments.

### DIFF
--- a/GeoLib/AnalyticalGeometry.cpp
+++ b/GeoLib/AnalyticalGeometry.cpp
@@ -126,51 +126,42 @@ bool lineSegmentIntersect(
 	MathLib::Vector3 const qp(a, c);
 	MathLib::Vector3 const pq(c, a);
 
+	auto isLineSegmentIntersectingAB = [&v](MathLib::Vector3 const& ap,
+	                                        std::size_t i)
+	{
+		// check if p is located at v=(a,b): (ap = t*v, t in [0,1])
+		if (0.0 <= ap[i] / v[i] && ap[i] / v[i] <= 1.0) {
+			return true;
+		}
+		return false;
+	};
+
 	if (parallel(v,w)) { // original line segments (a,b) and (c,d) are parallel
 		if (parallel(pq,v)) { // line segment (a,b) and (a,c) are also parallel
-			if (v[0] != 0.0) { // segment (a,b) non perpendicular to x-axis
-				// check if c is located at v (c-a = t (b-a), t in [0,1])
-				if (0.0 <= qp[0] / v[0] && qp[0] / v[0] <= 1.0) {
-					s = c;
-					return true;
-				}
-				// check if d is located at v (d-a = t (b-a), t in [0,1])
-				if (0.0 <= MathLib::Vector3(a, d)[0] / v[0] &&
-				    MathLib::Vector3(a, d)[0] / v[0] <= 1.0) {
-					s = d;
-					return true;
-				}
-				return false;
-			} else if (v[1] != 0.0) {
-				// (a,b) perpendicular to x-axis and not perpendicular to y-axis
-				// check if c is located at v (c-a = t (b-a), t in [0,1])
-				if (0.0 <= qp[1] / v[1] && qp[1] / v[1] <= 1.0) {
-					s = c;
-					return true;
-				}
-				// check if d is located at v (d-a = t (b-a), t in [0,1])
-				if (0.0 <= MathLib::Vector3(a, d)[1] / v[1] &&
-				    MathLib::Vector3(a, d)[1] / v[1] <= 1.0) {
-					s = d;
-					return true;
-				}
-				return false;
-			} else if (v[2] != 0.0) {
-				// (a,b) perpendicular to x- and y-axis and not perpendicular
-				// to z-axis
-				// check if c is located at v (c-a = t (b-a), t in [0,1])
-				if (0.0 <= qp[2] / v[2] && qp[2] / v[2] <= 1.0) {
-					s = c;
-					return true;
-				}
-				// check if d is located at v (d-a = t (b-a), t in [0,1])
-				if (0.0 <= MathLib::Vector3(a, d)[2] / v[2] &&
-				    MathLib::Vector3(a, d)[2] / v[2] <= 1.0) {
-					s = d;
-					return true;
-				}
-				return false;
+			// Here it is already checked that the line segments (a,b) and (c,d)
+			// are parallel. At this point it is also known that the line
+			// segment (a,c) is also parallel to (a,b). In that case it is
+			// possible to express c as c(t) = a + t * (b-a) (analog for the
+			// point d). Since the evaluation of all three coordinate equations
+			// (x,y,z) have to lead to the same solution for the parameter t it
+			// is sufficient to evaluate t only once.
+
+			// Search id of coordinate with largest absolute value which is will
+			// be used in the subsequent computations. This prevents division by
+			// zero in case the line segments are parallel to one of the
+			// coordinate axis.
+			std::size_t i_max(std::abs(v[0]) <= std::abs(v[1]) ? 1 : 0);
+			i_max = std::abs(v[i_max]) <= std::abs(v[2]) ? 2 : i_max;
+			if (isLineSegmentIntersectingAB(qp, i_max)) {
+				s = c;
+				return true;
 			}
+			MathLib::Vector3 const ad(a, d);
+			if (isLineSegmentIntersectingAB(ad, i_max)) {
+				s = d;
+				return true;
+			}
+			return false;
 		}
 		return false;
 	}
--- a/GeoLib/AnalyticalGeometry.cpp
+++ b/GeoLib/AnalyticalGeometry.cpp
@@ -121,27 +121,63 @@ bool lineSegmentIntersect(
 		return true;
 	}
 
-	// general case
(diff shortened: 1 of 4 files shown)


TomFischer:
Jenkins, test this please.

endJunction:
:fast_forward:

rinkk:
I have no idea what you're doing but it look okay and it works.

GeoLib/AnalyticalGeometry.cpp
```diff
@@ -133,11 +133,16 @@ bool lineSegmentIntersect(
 	if (parallel(v,w)) {
 		if (parallel(pq,v)) {
 			// check if c is located at v (c-a = t (b-a), t in [0,1])
-			if (qp[0] / v[0] <= 1.0)
+			if (0 <= qp[0] / v[0] && qp[0] / v[0] <= 1.0) {
```
chleh: Can `v[0]` be 0.0?
Btw: `0 <= qp[0] / v[0]` essentially checks if ` qp[0] ` and `v[0]` have the same sign. Is that intended?

GeoLib/Polygon.cpp
```diff
@@ -368,43 +368,34 @@ void Polygon::splitPolygonAtIntersection (std::list<Polygon*>::iterator polygon_
...
+			polyline0.addPoint (intersection_pnt_id);
 			for (std::size_t k(idx1 + 1); k < (*polygon_it)->getNumberOfPoints(); k++)
-				polygon0->addPoint ((*polygon_it)->getPointID (k));
-			if (!polygon0->initialise())
-			{
-				ERR("Polygon::splitPolygonAtIntersection(): Initialization of polygon0 failed.");
-				exit (1);
-			}
+				polyline0.addPoint ((*polygon_it)->getPointID (k));
+			GeoLib::Polygon *polygon0(new Polygon(polyline0));
```
chleh: I'd prefer unique_ptr here, and release if you transfer ownership.

GeoLib/Polygon.cpp
```diff
@@ -368,43 +368,34 @@ void Polygon::splitPolygonAtIntersection (std::list<Polygon*>::iterator polygon_
...
-				polygon1->addPoint ((*polygon_it)->getPointID (k));
-			polygon1->addPoint (intersection_pnt_id);
-			if (!polygon1->initialise())
-			{
-				ERR("Polygon::splitPolygonAtIntersection(): Initialization of polygon1 failed.");
-				exit (1);
-			}
+				polyline1.addPoint ((*polygon_it)->getPointID (k));
+			polyline1.addPoint (intersection_pnt_id);
+			GeoLib::Polygon *polygon1(new Polygon(polyline1));
```
chleh: Same ownership thing here.

GeoLib/Polygon.cpp
TomFischer: I found an other solution. See the comment in the commit e10ef00.

GeoLib/Polygon.cpp
TomFischer: Same solution as above.

GeoLib/AnalyticalGeometry.cpp
TomFischer: > Can v[0] be 0.0?

Yes, this could be the case. This was a serious problem. I hopefully fixed it in commit 6e75c9e?! Please have a look.

TomFischer: > Btw: 0 <= qp[0] / v[0] essentially checks if qp[0] and v[0] have the same sign. Is that intended?

The intention here is to check if the quotient `qp[0] / v[0]` is in [0,1]. Maybe there is a better solution?

TomFischer:
@rinkk I can not find your `while` loop comment anymore. Your point was valid. I removed the while loop in the commit e10ef00.

chleh:
Your solution in e10ef00 looks valid. It looks much better than before, indeed. :+1:

chleh:
If I see it correctly, you still divide by some vector components very often in 6e75c9e532cc4f50b3c0fbd784638b75ecfc21f2. And probably you always want to check if `0.0 <= a/b <= 1.0`.

> The intention here is to check if the quotient qp[0] / v[0] is in [0,1]. Maybe there is a better solution?

Let `a := qp[0]; b := v[0]`, then you could check:
(i) `0.0 <= a*b` That checks if a and b have the same sign and is logically the same as `0.0 <= a/b`, only it is guaranteed that you don't divide by 0.0. However, I don't know if that has any implications if `b == 0.0`, because then the new check would succeed whereas the old check would not, since `a/0.0 == NaN` which AFAIK compares `false` to everything.
(ii) `std::abs(a) <= std::abs(b)` is __always__ equivalent to `a/b <= 1.0`. There are no special considerations that must be taken for this case.

Let's discuss vis-a-vis.

rinkk:
:+1:

GeoLib/Polygon.cpp
```diff
@@ -348,64 +348,46 @@ void Polygon::ensureCWOrientation ()
...
+
+	GeoLib::Polyline polyline1{(*polygon_it)->getPointsVec()};
+	polyline1.addPoint(intersection_pnt_id);
+	for (std::size_t k(idx0 + 1); k <= idx1; k++)
+		polyline1.addPoint((*polygon_it)->getPointID(k));
+	polyline1.addPoint(intersection_pnt_id);
+
+	// remove original polyline and add two new polylines
+	std::list<GeoLib::Polygon*>::const_iterator polygon1_it =
+	    _simple_polygon_list.insert(_simple_polygon_list.erase(polygon_it),
```
endJunction: Unfortunately the STL version shipped with gcc-4.8.1 does not implement the c++11 version of `std::list::erase` which is taking a const iterator. See http://en.cppreference.com/w/cpp/container/list/erase for details. This was only fixed with  gcc-4.9.0. Opening new issues to set minimum gcc version to 4.9.0.

Until the new gcc is required I would add an #if based on gcc version like this:
```c++
/// \todo Use const_iterator after requiring gcc >= 4.9.0
/// stl::erase takes const iterator required by c++11 since gcc-4.9.0
#if __GNUC__ <= 4 && (__GNUC_MINOR__ < 9)
			polygon1_it =
			    _simple_polygon_list.insert(_simple_polygon_list.erase(
			        (std::list<GeoLib::Polygon*>::iterator)polygon_it),
			        polygon1);
#else
			polygon1_it = _simple_polygon_list.insert(
			    _simple_polygon_list.erase(polygon_it), polygon1);
#endif
```

chleh: Rather than introducing an `#if` here, I'd just go with a non-const iterator. I think using preprocessor macros here is worse than using non-const iterators. Remember: This is not an interface but an internal detail. Or is the build failing because of that?
Btw: you should probably use `const_cast<>`.

endJunction: It's temporary solution until gcc 4.9 is used in CI environment.
Non-const iterator is visible in the interface, which is not good.
`const_cast<>` will probably not work because its type change from `std::list::iterator` to `std::list::const_iterator`

TomFischer: Using the C style cast makes clang not happy.

endJunction: You can make a new temporary std::list::iterator...

GeoLib/AnalyticalGeometry.cpp
```diff
@@ -121,28 +121,55 @@ bool lineSegmentIntersect(
...
+		if (parallel(pq,v)) { // line segment (a,b) and (a,c) are also parallel
+			// Here it is already checked that the line segments (a,b) and (c,d)
+			// are parallel. At this point it is also known that the line
+			// segment (a,c) is also parallel to (a,b). In that case it is
+			// possible to express c as c(t) = a + t * (b-a) (analog for the
+			// point d). Since the evaluation of all three coordinate equations
+			// (x,y,z) have to lead to the same solution for the parameter t it
+			// is sufficient to evaluate t only once.
+
+			// Search id of coordinate with largest absolute value which is will
```
chleh: ~~is~~

GeoLib/AnalyticalGeometry.cpp
```diff
@@ -121,28 +121,55 @@ bool lineSegmentIntersect(
...
+			// be used in the subsequent computations. This prevents division by
+			// zero in case the line segments are parallel to one of the
+			// coordinate axis.
+			std::size_t i_max(std::abs(v[0]) <= std::abs(v[1]) ? 1 : 0);
+			i_max = std::abs(v[i_max]) <= std::abs(v[2]) ? 2 : i_max;
+			if (isLineSegmentIntersectingAB(qp, i_max)) {
+				s = c;
```
chleh: Just wondering: Possibly ab and cd overlap in an entire range. Does the specific choice of s have any consequences? In principle you could use any point in `ab intersect cd`.

chleh:
I don't know why, but I like your new intersection solution :laughing: 
:+1:

GeoLib/AnalyticalGeometry.cpp
TomFischer: In my opinion in this case the complete line segment `ab` and `cd` are intersecting each other should be returned. But this would be a separate PR.

endJunction:
@chleh  This is much better solution using the `auto` then the proposed #if. Thanks!

PR #1082, opened 2016-03-07
Disable Segfaulting PETSc unit tests one by one

chleh:
This PR
* removes the catch-all regular expression filters that disable unit tests for PETSc
* disables only the segfaulting tests for PETSc one by one

Thus, new unit tests will by default also be run in the PETSc build. Furthermore, it will be possible to run `testrunner` also for PETSc without having to pass filters.

### DIFF
--- a/Tests/AssemblerLib/TestMeshComponentMap.cpp
+++ b/Tests/AssemblerLib/TestMeshComponentMap.cpp
@@ -67,7 +67,11 @@ class AssemblerLibMeshComponentMapTest : public ::testing::Test
 
 };
 
+#ifndef USE_PETSC
 TEST_F(AssemblerLibMeshComponentMapTest, CheckOrderByComponent)
+#else
+TEST_F(AssemblerLibMeshComponentMapTest, DISABLED_CheckOrderByComponent)
+#endif
 {
     // - Entries in the vector are arranged in the order of a component type and then node ID
     // - For example, x=[(node 0, comp 0) (node 1, comp 0) ... (node n, comp0), (node 0, comp1) ... ]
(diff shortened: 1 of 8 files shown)


wenqing:
@chleh :+1: . It is better to use `` DISABLED_`` prefix (for gtest ver>1.7).
:fast_forward:

chleh:
I disabled a further test and removed a blank line.

chleh:
Hmm, more segfaults ahead. So I'll have to make some additions to this one.

Tests/AssemblerLib/TestVectorMatrixBuilder.cpp
```diff
@@ -17,6 +17,9 @@
 
 #include "AssemblerLib/VectorMatrixBuilder.h"
 
+#if 1
```
chleh: What's that? Going to remove it.

chleh:
That should be all for now.

chleh:
Only PETSc ctests failing. That's OK. I don't expect problems from the windows and mac builds.

Tests/AssemblerLib/TestMeshComponentMap.cpp
```diff
@@ -130,7 +136,11 @@ TEST_F(AssemblerLibMeshComponentMapTest, OutOfRangeAccess)
         Location(mesh->getID(), MeshItemType::Node, 0), 10));
 }
 
+#ifndef USE_PETSC
+TEST_F(AssemblerLibMeshComponentMapTest, SubsetOfNodesByComponent)
+#else
 TEST_F(AssemblerLibMeshComponentMapTest, SubsetOfNodesByComponent)
```
TomFischer: DISABLED_

PR #1083, opened 2016-03-07
Build support for VTK 7

bilke:
Some modifications to pick up [VTK 7](https://blog.kitware.com/vtk-7-0-0/) from CMake and build against it. This VTK release includes a new and faster rendering backend which may improve Data Explorer performance.


endJunction:
Works for me; VTK-7 is found. Can not check the compilation with gui, because of qt5 requirement of my precompiled VTK version and qt4 in OGS.
:fast_forward:

rinkk:
tested with gui using qt 4.8 and vtk 6.1. everything works fine :+1:

PR #1085, opened 2016-03-09
Clang sanitizer cleanup

endJunction:
__1.__ I have checked all of the undefined behaviour checked by clang's sanitizer and think that the blacklisted are not caused by wrong usage from our side.

Because the names in the blacklist are mangled, they could be different on the envinf1 machine; Then the blacklist shall be updated.

 - [ ] Update blacklist for envinf1.
 - [x] Add the blacklist to CXXFLAGS in cmake.

__2.__ Removed remaining memory leaks in the testrunner.

@chleh There are no more memleaks as of now.

### DIFF
--- a/Tests/AssemblerLib/LocalToGlobalIndexMapMultiComponent.cpp
+++ b/Tests/AssemblerLib/LocalToGlobalIndexMapMultiComponent.cpp
@@ -74,6 +74,12 @@ public:
 		    mesh_items_all_nodes->getIntersectionByNodes(nodes));
 	}
 
+	~AssemblerLibLocalToGlobalIndexMapMultiDOFTest()
+	{
+		for (auto e : boundary_elements)
+			delete e;
+	}
+
 	void initComponents(const unsigned num_components, const unsigned selected_component,
 						const AL::ComponentOrder order)
 	{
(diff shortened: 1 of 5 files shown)


TomFischer:
:fast_forward:

chleh:
:fast_forward:
The blacklist is a good solution for the time being. In the future, that might be changed to some diff-based approach, because, as you pointed out, the blacklist will probably be different on different machines, and it will be different for different library versions. But that's only for some point in the future.

bilke:
Just a suggestion: I would put `blacklist.txt` either maybe somewhere into scripts-directory or as a hidden file (`.blacklist.txt`) and maybe add a prefix, e.g. `sanitizer-blacklist.txt`.

chleh:
I thought, @endJunction has move the blacklist already to scripts/test, and has given it a different name (clang_santize_blacklist.txt or so).
I vote against hidden files under source code management.

endJunction:
I finally managed to blacklist all of "not-of-our-concern" undefined behaviour reported by clang's sanitizer.
The blacklist is placed in scripts/test/clang_sanitizer_blacklist.txt.
I also added an environment variable enabling printing of a stack trace in case any of the undefined behaviour is caught.

@bilke If OK, please merge. It's rebased and ready.

Issue #999, opened 2016-01-28
Crash when building GeoLib::Grid from mesh nodes


rinkk:
This is based on a mesh from @waltherm 
This is the way I can reproduce the crash:
Open DataExplorer, load mesh, call "Analyze mesh..." and press "Start".
The application crashes in MeshLib/MeshEditing/MeshRevision.cpp in `collapseNodeIndices()` when a grid is build based on the mesh nodes (line 195).

Error msg:

```
Unhandled exception at 0x000007FEFCE3ADCD in DataExplorer.exe: Microsoft C++ exception: std::bad_alloc at memory location 0x00000000002F4568.
```

The mesh is located in //visdata/karsten/DE_analyse_mesh_bug/

TomFischer:
I can reproduce the behaviour using the ogs mesh. I can **not** reproduce the behaviour using the original gmsh mesh. There is a difference between nodes of the gmsh mesh and the ogs mesh.

 - ogs node 4 (indexing starts with zero) has z-coordinate 1e-14
 - gmsh node 5 (indexing starts with one) has z-coordinate 0.0

Question: Why is the node of the ogs mesh modified? If there is a special reason we should handle this case or is it modified by mistake?

waltherm:
@TomFischer sorry, Tom. I should have mentioned, that I maninpulated the ogs-mesh on purpose to produce the error. You can easily do the same with the gmsh msh and will produce the bug. 
I changed the z coordinate of the 4th point from 0 to 1e-14.

endJunction:
Those who will look into the DE writing routine, please have a look on the precision of the output. The provided test_ogs.msh file has at least one decimal place less than the test.msh. In my opinion such precision loss must not happen.

waltherm:
:+1: for @endJunction 's comment

TomFischer:
@waltherm Can you produce a gmsh mesh (without manipulating it manually) that produces this behaviour?

TomFischer:
@endJunction The number of digit of the ogs output stream is set to `std::numeric_limits<double>::digits10`, see FileIO/Writer.cpp line 27. According to the [documentation](http://en.cppreference.com/w/cpp/types/numeric_limits/digits10) the number is written in such a precision that reading it again leads to exact the same number. I wonder if that high number of decimal digit of the gmsh mesh coordinates can be reasonable read into a double? What do you suggest to improve in ogs?

rinkk:
Two things:
[...]

PR #1088, opened 2016-03-11
Fix Grid Bug

TomFischer:
This PR fixes issue #999.

This PR substitutes the absolute criterion for the decision which spatial dimension for the setup of the Grid should be taken into account to an relative one (commit 5ed53fd). Furthermore the manual resize of the bounding box is deleted since this is already done in the AABB class in a previous PR.

### DIFF
--- a/GeoLib/Grid.h
+++ b/GeoLib/Grid.h
@@ -198,9 +198,9 @@ Grid<POINT>::Grid(InputIterator first, InputIterator last,
 {
 	auto const n_pnts(std::distance(first,last));
 
-	std::array<double,3> delta = {{this->_max_pnt[0]-this->_min_pnt[0],
-		 this->_max_pnt[1]-this->_min_pnt[1],
-		 this->_max_pnt[2]-this->_min_pnt[2]}};
+	std::array<double, 3> delta = {{_max_pnt[0] - _min_pnt[0],
+	                                _max_pnt[1] - _min_pnt[1],
+	                                _max_pnt[2] - _min_pnt[2]}};
 
 	assert(n_pnts > 0);
 	initNumberOfSteps(max_num_per_grid_cell, static_cast<std::size_t>(n_pnts), delta);
@@ -289,8 +289,8 @@ void Grid<POINT>::createGridGeometry(GeoLib::GEOObjects* geo_obj) const
 {
 	std::vector<std::string> grid_names;
 
-	GeoLib::Point const& llf (this->getMinPoint());
-	GeoLib::Point const& urb (this->getMaxPoint());
+	GeoLib::Point const& llf (getMinPoint());
+	GeoLib::Point const& urb (getMaxPoint());
 
 	const double dx ((urb[0] - llf[0]) / _n_steps[0]);
 	const double dy ((urb[1] - llf[1]) / _n_steps[1]);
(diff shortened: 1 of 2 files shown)


GeoLib/Grid.h
```diff
@@ -185,19 +185,11 @@ Grid<POINT>::Grid(InputIterator first, InputIterator last,
...
-	std::array<double,3> delta = {{this->_max_pnt[0]-this->_min_pnt[0],
-		 this->_max_pnt[1]-this->_min_pnt[1],
-		 this->_max_pnt[2]-this->_min_pnt[2]}};
+	std::array<double, 3> delta = {{_max_pnt[0] - _min_pnt[0],
+	                                _max_pnt[1] - _min_pnt[1],
+	                                _max_pnt[2] - _min_pnt[2]}};
 
-	assert(n_pnts >= 0);
+	assert(n_pnts > 0);
 	initNumberOfSteps(max_num_per_grid_cell, static_cast<std::size_t>(n_pnts), delta);
```
chleh: Why the static cast here? It looks unnecessary.

GeoLib/Grid.h
```diff
@@ -486,9 +480,11 @@ template <typename POINT>
 void Grid<POINT>::initNumberOfSteps(std::size_t n_per_cell,
 	std::size_t n_pnts, std::array<double,3> const& delta)
 {
+	double const max(* std::max_element(delta.cbegin(), delta.cend()));
+
 	std::bitset<3> dim; // all bits set to zero
 	for (std::size_t k(0); k<3; ++k) {
-		if (std::abs(delta[k]) >= std::numeric_limits<double>::epsilon()) {
+		if (delta[k] >= 1e-4 * max) {
```
chleh: I think, this "hack" should be documented: (i) here, as a doxygen comment, and (ii) probably also on the website.

Tests/GeoLib/TestGrid.cpp
```diff
@@ -133,11 +133,11 @@ TEST(GeoLib, SearchNearestPointsInDenseGrid)
 	ASSERT_EQ(j_max*k_max*(i_max/2 + 1) - 1, res->getID());
 	ASSERT_NEAR(sqrt(2.0)/50.0, sqrt(MathLib::sqrDist(*res, search_pnt)), std::numeric_limits<double>::epsilon());
 
-	for (std::size_t i(0); i < i_max; i++) {
+	for (std::size_t i(0); i < i_max; i=i+4) {
```
chleh: Why now += 4?

chleh:
:+1:

GeoLib/Grid.h
```diff
@@ -506,8 +502,8 @@ void Grid<POINT>::initNumberOfSteps(std::size_t n_per_cell,
 
 	switch (dim.count()) {
 	case 3: // 3d case
-		_n_steps[0] = sc_ceil(std::cbrt(
-			n_pnts*delta[0]*delta[0]/(n_per_cell*delta[1]*delta[2])));
+		_n_steps[0] = sc_ceil(std::cbrt(n_pnts * (delta[0] / delta[1]) *
+			                            (delta[0] / delta[2]) / n_per_cell));
```
rinkk: mixed tabs + whitespaces.
also, could you please explain why the computation has been changed (and why it has only been changed for n_steps[0])

TomFischer: > mixed tabs + whitespaces

I thought the rule is: identing by tabs, align by spaces?!

> also, could you please explain why the computation has been changed (and why it has only been changed for n_steps[0])

If the `extensions[0]` (`delta` will be renamed to `extensions`) is very small then computing the square is even smaller. There are some cases the result can not represented by a double. In 3d case `extensions[1]` and `extensions[2]` are also very small. The quotients `extensions[0]/extensions[1]` and `extensions[0]/extensions[2]` are probably rather representable than the result of `extension[0]*extension[0]`.

The computation has been changed only in this case because only here a multiplication occurs.

GeoLib/Grid.h
TomFischer: Documented the method. But doxygen doesn't process the documentation? Any ideas?

What do you mean by documentation on the website?

chleh: For me doxygen produces a correct documentation from your comment.

chleh: > What do you mean by documentation on the website?

I confused it. I thought, this PR was about some preprocessing tool. So I thought, there was some wiki page about it.

TomFischer: @chleh Thanks for investigating the problem. I rebuild the docu and now I can see it, too. Is your original remark addressed with the doxygen comment?

chleh: Yes, it is. :+1: again.

PR #1091, opened 2016-03-11
DataExplorer MainWindow clean up

rinkk:
* fixing a few mem-leaks
* using a few unique pointers
* dereferencing a few pointers for dialog

### DIFF
--- a/Applications/DataExplorer/mainwindow.h
+++ b/Applications/DataExplorer/mainwindow.h
@@ -15,6 +15,8 @@
 #ifndef MAINWINDOW_H
 #define MAINWINDOW_H
 
+#include <memory>
+
 #ifndef Q_MOC_RUN  // See: https://bugreports.qt-project.org/browse/QTBUG-22829
 #include "Applications/ApplicationsLib/ProjectData.h"
 #endif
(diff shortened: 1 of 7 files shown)


Applications/DataExplorer/mainwindow.h
```diff
@@ -125,21 +126,17 @@ protected slots:
...
-
-	QString curFile;
-
-	MshModel* _meshModels;
-	ElementTreeModel* _elementModel;
-	TreeModel* _processModel;
+	
 	ProjectData _project;
-	VtkVisPipeline* _vtkVisPipeline;
+	std::unique_ptr<MshModel> _meshModels;
```
chleh: Why the plural `_meshModels`?

Applications/DataExplorer/mainwindow.h
```diff
@@ -47,7 +49,6 @@ class MainWindow : public QMainWindow, public Ui_MainWindowClass
 
 public:
 	MainWindow(QWidget* parent = 0);
-	~MainWindow();
```
chleh: Regarding the compilation errors: Maybe they go away, if you keep the destructor declaration here and define an empty destructor in `mainwindow.cpp`

chleh:
Looks good :+1:.

Applications/DataExplorer/mainwindow.h
rinkk: No, re-including the destructor didn't help.
We tested a bit on Lars' machine and the only way we could fix it was to remove the forward declarations and really include the headers to the models here. I'm currently rebasing the changes and update asap.

chleh: Next time it would be really nice if you could wait with the rebasing until after code review, s.t. it is clearly visible which changes you made later on.

TomFischer:
Looks good. :+1:

PR #1092, opened 2016-03-11
Select element types/dimension/order to be built via cmake

chleh:
This PR adds the following cmake options, which enable selection of element types to be built via cmake. That possibly saves compile time. The effect will be more significant for fixed-size shape matrices, not so much for dynamically allocated ones.

```
OGS_ENABLE_ELEMENT_CUBOID
OGS_ENABLE_ELEMENT_PRISM
OGS_ENABLE_ELEMENT_PYRAMID
OGS_ENABLE_ELEMENT_SIMPLEX
OGS_MAX_ELEMENT_DIM
OGS_MAX_ELEMENT_ORDER
```

### DIFF
--- a/scripts/cmake/OGSEnabledElements.cmake
+++ b/scripts/cmake/OGSEnabledElements.cmake
@@ -1,0 +1,24 @@
+# Definitions controlling which FEM elements will be compiled
+if(NOT OGS_MAX_ELEMENT_DIM MATCHES "^[0-3]$")
+  message(FATAL_ERROR "OGS_MAX_ELEMENT_DIM must be an integer between 0 and 3.")
+endif()
+add_definitions(-DOGS_MAX_ELEMENT_DIM=${OGS_MAX_ELEMENT_DIM})
+
+if(NOT OGS_MAX_ELEMENT_ORDER MATCHES "^[0-9]$")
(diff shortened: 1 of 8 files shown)


chleh:
Fixed it.

chleh:
At least for non-GUI. ^^

endJunction:
Be aware that when for example only pyramids are enabled the triangles and quads should also be available because of the boundary.

chleh:
Gonna rethink it.

wenqing:
Also line element should be enabled by default because it is used as edge type to assign BCs for all types of element.

chleh:
The macro `ENABLED_ELEMENT_TYPE_LINE` should make sure that lines and points are effectively always enabled.

chleh:
Note: Just added an error message for nonexistent mesh element types.  50269d1

scripts/cmake/OGSEnabledElements.cmake
```diff
@@ -0,0 +1,24 @@
+# Definitions controlling which FEM elements will be compiled
+if(NOT OGS_MAX_ELEMENT_DIM MATCHES "^[0-9]$")
```
endJunction: 0-3 is sufficient.

endJunction:
:fast_forward: from my side.

wenqing:
:+1:

PR #1094, opened 2016-03-14
Pre post timestep for local asm

endJunction:
- Splitting of `postTimestep()` and `output()`. (No changes, just copy&paste).
 - Implementation of preTimestep and postTimestep executors for local asm.
    
To use in particular process one needs an implementation in FEM part of the process of
```c++
void FEM::preTimestep(std::vector<double> const& local_x) override;
```
and an implementation in the Process calling `FEM::preTimestep`
    
```c++
void SpecificProcess::preTimestep() override
{
    Base::_global_setup.execute(
        [&](std::size_t const id, LocalAssembler* const local_assembler, GlobalVector const& x) -> void
        {
            Base::_global_assembler->preTimestep(id, local_assembler, x);
        },
        _local_assemblers, x);
}
```

### DIFF
--- a/AssemblerLib/VectorMatrixAssembler.h
+++ b/AssemblerLib/VectorMatrixAssembler.h
@@ -14,6 +14,49 @@
 
 #include "NumLib/ODESolver/Types.h"
 
+namespace
+{
+inline AssemblerLib::LocalToGlobalIndexMap::RowColumnIndices
+getRowColumnIndices(std::size_t const id,
+                    AssemblerLib::LocalToGlobalIndexMap const& dof_table,
+                    std::vector<GlobalIndexType>& indices)
+{
+    assert(dof_table.size() > id);
+    assert(indices.empty());
+
+    // Local matrices and vectors will always be ordered by component,
+    // no matter what the order of the global matrix is.
+    for (unsigned c = 0; c < dof_table.getNumComponents(); ++c)
+    {
+        auto const& idcs = dof_table(id, c).rows;
+        indices.reserve(indices.size() + idcs.size());
+        indices.insert(indices.end(), idcs.begin(), idcs.end());
+    }
+
+    return AssemblerLib::LocalToGlobalIndexMap::RowColumnIndices(indices,
+                                                                 indices);
+}
+
+template <typename Callback, typename GlobalVector, typename... Args>
+void passLocalVector(Callback& cb, std::size_t const id,
+                     AssemblerLib::LocalToGlobalIndexMap const& dof_table,
+                     GlobalVector const& x, Args&&... args)
+{
+    std::vector<GlobalIndexType> indices;
+    auto const r_c_indices = getRowColumnIndices(id, dof_table, indices);
+
+    std::vector<double> local_x;
+    local_x.reserve(indices.size());
+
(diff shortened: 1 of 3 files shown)


AssemblerLib/VectorMatrixAssembler.h
```diff
@@ -80,6 +80,46 @@ class VectorMatrixAssembler<GlobalMatrix, GlobalVector,
         local_assembler->addToGlobal(r_c_indices, M, K, b);
     }
 
+    /// Executes preTimestep call for each of the local assemblers for the
+    /// given mesh item.
+    /// The positions in the global matrix/vector are taken from
+    /// the LocalToGlobalIndexMap provided in the constructor at index \c id.
+    /// \attention The index \c id is not necesserily the mesh item's id.
+    template <typename LocalAssembler_>
```
chleh: Do you need the trailing underscore?

AssemblerLib/VectorMatrixAssembler.h
```diff
@@ -80,6 +80,46 @@ class VectorMatrixAssembler<GlobalMatrix, GlobalVector,
...
+                     GlobalVector const& x) const
+    {
+        auto cb = [local_assembler](
+            std::vector<double> local_x,
+            LocalToGlobalIndexMap::RowColumnIndices const& r_c_indices)
+        {
+            local_assembler->preTimestep(local_x);
+        };
+
+        passLocalVector(cb, id, _data_pos, x);
```
chleh: Maybe I'm blind, but where is the definition of this method?

chleh: Btw: `operator()` has not been adapted to use the `passLocalVector()` function.

chleh:
Very important comment, though easy to fix. Afterwards :fast_forward:.

ProcessLib/Process.h
```diff
@@ -72,14 +72,59 @@ class Process
 	/// Process specific initialization called by initialize().
 	virtual void createLocalAssemblers() = 0;
 
-	/// Postprocessing after solve().
+	/// Preprocessing before starting assembly for new timestep.
+	virtual void preTimestep(GlobalVector const& x) {}
```
TomFischer: Will `preTimestep()`, `postTimestep()` or `output()` change the internals of the process object or will they work on external data? If they only work on external data the methods could be `const`.

chleh: They do change the internals. E.g., they could save some secondary variables from the previous timeste.

TomFischer: Thanks for explanation. I understand that `preTimestep()` and `postTimestep()` should be able change internal things. Will `output()` change internals, too?

TomFischer:
Looks good. :+1:

AssemblerLib/VectorMatrixAssembler.h
endJunction: Forgot to pull your changes, but tested them in different branch. Stupid me.

I'll have to include the missing commit before other changes, so there will be a rebase, but the extra changes will be on the top.

chleh:
:+1:
I guess, there is no need for further review, and I really need this piece of code.

Issue #663, opened 2015-04-27
[DE-GMSH] saving wrong x-z gmsh geo file coordinates + missing surface


waltherm:
saving the following .gli with the current DE will not save the correct coordinates and will miss the second surface

```
#POINTS
0 0 0 -10.5
1 1000 0 -10.5
2 1000 0 -0.5
3 1000 0 0
4 0 0 0
5 0 0 -0.5
#POLYLINE
 $NAME
  TORF
 $POINTS
  5
  2
  3
  4
  5
#POLYLINE
 $NAME
  SAND
 $POINTS
  0
  1
  2
  5
  0
#SURFACE
 $NAME
  TORF_S
 $POLYLINES
  TORF
 $TYPE
  0
#SURFACE
 $NAME
  SAND_S
 $POLYLINES
  SAND
 $TYPE
  0
#STOP
```

will give this: 
```
// GMSH input file created by OpenGeoSys 6.0.1-289-g3246733

Point(0) = {0,0, 0.0, 300};
Point(1) = {1000,0, 0.0, 300};
Point(2) = {1000,0, 0.0, 300};
Point(3) = {1000,0, 0.0, 300};
Point(4) = {0,0, 0.0, 300};
Point(5) = {0,0, 0.0, 300};
Line(0) = {5,2};
Line(1) = {2,3};
Line(2) = {3,4};
Line(3) = {4,5};
Line Loop(4) = {0,1,2,3};
Plane Surface(0) = {4};
Line(5) = {0,1};
Line { 5 } In Surface { 0 };
Line(6) = {1,2};
Line { 6 } In Surface { 0 };
Line(7) = {2,5};
Line { 7 } In Surface { 0 };
Line(8) = {5,0};
Line { 8 } In Surface { 0 };
```

but should give something like this:
```
Mesh.Algorithm = 1;
Mesh.Optimize = 1;
lc=1;
Point(0)={0,0,-10.5,lc};
Point(1)={1000,0,-10.5,lc};
Point(2)={1000,0,-.5,lc};
Point(3)={1000,0,0,lc};
Point(4)={0,0,0,lc};
Point(5)={0,0,-.5,lc};
Line (1) ={5,2};
Line (2) ={2,3};
Line (3) ={3,4};
Line (4) ={4,5};
Line (5) ={0,1};
Line (6) ={1,2};
Line (7) ={5,0};
Line Loop(1) ={1,2,3,4};
Line Loop(2) ={5,6,-1,7};
[...]

PR #1095, opened 2016-03-14
[Fix] Generate Gmsh mesh in xz plane

TomFischer:
This fixes issue #663.

The original input data is rotated to the x-y-plane. All intermediate computations are done and finally to data is rotated back to the original location.

- [x] check with more test examples.

### DIFF
--- a/FileIO/GMSHInterface.cpp
+++ b/FileIO/GMSHInterface.cpp
@@ -29,12 +29,15 @@
 #include "FileIO/GmshIO/GMSHFixedMeshDensity.h"
 #include "FileIO/GmshIO/GMSHNoMeshDensity.h"
 
+#include "GeoLib/AnalyticalGeometry.h"
 #include "GeoLib/GEOObjects.h"
 #include "GeoLib/Point.h"
 #include "GeoLib/Polygon.h"
 #include "GeoLib/Polyline.h"
 #include "GeoLib/PolylineWithSegmentMarker.h"
 #include "GeoLib/QuadTree.h"
+
+#include "MathLib/LinAlg/Solvers/GaussAlgorithm.h"
 
 #include "MeshLib/Elements/Elements.h"
 #include "MeshLib/Mesh.h"
@@ -371,12 +374,13 @@ void GMSHInterface::writeGMSHInputFile(std::ostream& out)
 	if (! merged_pnts) {
 		ERR("GMSHInterface::writeGMSHInputFile(): Did not found any points.");
 		return;
-	} else {
-		const std::size_t n_pnts(merged_pnts->size());
-		for (std::size_t k(0); k<n_pnts; k++) {
-			(*((*merged_pnts)[k]))[2] = 0.0;
-		}
 	}
+
+	// Rotate points to the x-y-plane.
+	_inverse_rot_mat = GeoLib::rotatePointsToXY(*merged_pnts);
+	// Compute inverse rotation matrix to reverse the rotation later on.
+	_inverse_rot_mat.transposeInPlace();
+
 	std::vector<GeoLib::Polyline*> const* merged_plys(_geo_objs.getPolylineVec(_gmsh_geo_name));
 	DBUG("GMSHInterface::writeGMSHInputFile(): \t ok.");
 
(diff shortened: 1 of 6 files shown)


waltherm:
looks good, works with my tested examples, which failed before. thanks, @TomFischer !

FileIO/GMSHInterface.cpp
```diff
@@ -371,12 +374,13 @@ void GMSHInterface::writeGMSHInputFile(std::ostream& out)
...
 		return;
-	} else {
-		const std::size_t n_pnts(merged_pnts->size());
-		for (std::size_t k(0); k<n_pnts; k++) {
-			(*((*merged_pnts)[k]))[2] = 0.0;
-		}
 	}
+
+	// Rotate points to the x-y-plane.
+	MathLib::DenseMatrix<double> rot_mat = rotatePointsToXY(*merged_pnts);
```
endJunction: Maybe this needs a translation to the origin too?

FileIO/GMSHInterface.cpp
```diff
@@ -463,12 +467,54 @@ void GMSHInterface::writeGMSHInputFile(std::ostream& out)
...
-		if (_gmsh_pnts[k]) {
-			out << *(_gmsh_pnts[k]) << "\n";
+	for (auto & gmsh_pnt : _gmsh_pnts) {
+		// reverse rotation
+		if (gmsh_pnt) {
+			double* tmp = _inverse_rot_mat * gmsh_pnt->getCoords();
+			(*gmsh_pnt)[0] = tmp[0];
+			(*gmsh_pnt)[1] = tmp[1];
+			(*gmsh_pnt)[2] = tmp[2];
+			delete [] tmp;
```
rinkk: why can you not simply assign ``` gmsh_pnt = _inverse_rot_mat * gmsh_pnt->getCoords();``` ?

FileIO/GMSHInterface.cpp
```diff
@@ -463,12 +467,54 @@ void GMSHInterface::writeGMSHInputFile(std::ostream& out)
...
+			out << *gmsh_pnt << "\n";
 		}
 	}
 }
 
+MathLib::DenseMatrix<double> GMSHInterface::rotatePointsToXY(
```
rinkk: i might be wrong but wasn't there a similar method somewhere else? you definitely needed sth like this in the TB example. if so, we should unify both methods...

rinkk:
pls check for code duplication concerning rotation, otherwise :+1:

FileIO/GMSHInterface.cpp
TomFischer: This can be done. But I don't see a great benefit.

FileIO/GMSHInterface.cpp
TomFischer: The types do not match. The matrix multiplication is not overloaded for `GMSHPoint`s.

FileIO/GMSHInterface.cpp
TomFischer: Very good point. The only difference between `GMSHInterface::rotatePointsToXY()` and the `rotatePointsToXY()` from GeoLib/AnalyticalGeometry* is the return value. I'll try to change the return value of the GeoLib version and remove the code above.

FileIO/GMSHInterface.cpp
```diff
@@ -371,12 +374,13 @@ void GMSHInterface::writeGMSHInputFile(std::ostream& out)
...
-		const std::size_t n_pnts(merged_pnts->size());
-		for (std::size_t k(0); k<n_pnts; k++) {
-			(*((*merged_pnts)[k]))[2] = 0.0;
-		}
 	}
+
+	// Rotate points to the x-y-plane.
+	MathLib::DenseMatrix<double> rot_mat = rotatePointsToXY(*merged_pnts);
+	// Compute inverse rotation matrix to reverse the rotation later on.
+	computeInverseRotationMatrix(rot_mat);
```
norihiro-w: why not simply use a transpose of the rotation matrix? usually R^T = R^-1

TomFischer: You are right! Thanks. The matrix has already a transpose operation.

norihiro-w:
A basic question: why do you need to rotate points to xy? I thought Gmsh can handle points in xz

TomFischer:
@norihiro-w 
> A basic question: why do you need to rotate points to xy? I thought Gmsh can handle points in xz 

You are right: Gmsh can handle input data located not only in the x-y-plane.

The pre-processing of the data (topological computations, intersections, ...) are performed under the assumption the data is located in the x-y-plane. For this reason the workflow implemented in this PR rotates the input data, perform these computations and afterwards rotates the data back (for this I need the inverse (transposed) rotation matrix).

norihiro-w:
i see. thank you for the explanation!

PR #1097, opened 2016-03-14
LocalAssemberTraits

chleh:
The struct introduced makes it possible to use the right version of Eigen's `block()` methods transparently for both fixed-size and dynamically allocated shape matrices.

Currently this code is not used anywhere, but will be in the TES process.

Caveat: The given implementation only works if all process variables are single-component and use the same shape functions!

### DIFF
--- a/ProcessLib/LocalAssemblerTraits.h
+++ b/ProcessLib/LocalAssemblerTraits.h
@@ -1,0 +1,234 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef PROCESSLIB_LOCALASSEMBLERTRAITS_H
+#define PROCESSLIB_LOCALASSEMBLERTRAITS_H
+
+namespace ProcessLib
+{
+
+namespace detail
+{
+
+/*! This class makes it possible to handle Eigen's block expressions transparently,
+ * both for fixed-size and dynamically allocated Eigen matrices.
+ *
+ * \tparam ShpPol   the shape matrix policy used
+ * \tparam NIntPts  the number of integration points for the FEM element
+ * \tparam NodalDOF the number of d.o.f. per node.
+ *                  If zero, this struct's methods will determine the size of
+ *                  the generated Eigen::block's at runtime. If non-zero, the
+ *                  size will be determined at compile-time.
+ * \tparam Dim      global spatial dimension
+ */
+template<typename ShpPol, unsigned NIntPts, unsigned NodalDOF, unsigned Dim>
+struct LocalAssemblerTraitsFixed
+{
+private:
+    template<int N, int M>
+    using Matrix = typename ShpPol::template MatrixType<N, M>;
+    template<int N>
+    using Vector = typename ShpPol::template VectorType<N>;
+
+public:
+    using ShapeMatrices = typename ShpPol::ShapeMatrices;
+
+    //! Square matrix of the given space dimensionality
+    using MatrixDimDim = Matrix<Dim, Dim>;
+
+    // TODO That only works if all process variables are single component
+    //      and use the same shape fucntions.
+    //! Local matrix for the given number of d.o.f.\ per node and number of
+    //! integration points
+    using LocalMatrix = Matrix<NIntPts*NodalDOF, NIntPts*NodalDOF>;
+    //! Local vector for the given number of d.o.f.\ per node and number of
+    //! integration points
+    using LocalVector = Vector<NIntPts*NodalDOF>;
+
+    //! Local vector for one component of one process variable.
+    //! The size is the number of nodes in the element.
+    using Vector1Comp = typename ShapeMatrices::ShapeType;
+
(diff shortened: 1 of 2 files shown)


endJunction:
Could you give an example? Not necessarily code, but just two snippets before and after this PR.

chleh:
```{.cpp}
// Note: Use LocalAssemblerTraits here (not ...Fixed)!
using Traits = LocalAssemblerTraits<ShapeMatricesType, ShapeFunction::NPOINTS,
        NODAL_DOF, GlobalDim>;
auto const N = /* number of integration points */;

Traits::blockShpShp(local_K, N*r, N*c, N, N);

// Will either translate to
local_K.block(N*r, N*c, N, N);
// or to
local_K.block<N, N>(N*r, N*c);
// depending on EIGEN_DYNAMIC_SHAPE_MATRICES (first case) or not (second case).
```

ProcessLib/LocalAssemblerTraits.h
```diff
@@ -0,0 +1,211 @@
...
+    using LocalMatrix = Matrix<NIntPts*NodalDOF, NIntPts*NodalDOF>;
+    using LocalVector = Vector<NIntPts*NodalDOF>;
+
+    using Vector1Comp = typename ShapeMatrices::ShapeType;
+
+    using LaplaceMatrix = Matrix<Dim*NodalDOF, Dim*NodalDOF>;
+
+
+    // block dim x dim, fixed-size const matrix
+    // TODO: swap variable names width <--> height
```
endJunction: Its rather good point to do so...

ProcessLib/LocalAssemblerTraits.h
```diff
@@ -0,0 +1,211 @@
...
+    template<typename Mat>
+    static
+    typename std::enable_if<NodalDOF != 0,
+        decltype(std::declval<const Mat>().template block<Dim, Dim>(0u, 0u))
+    >::type
+    blockDimDim(Mat const& mat,
+                unsigned top, unsigned left, unsigned width, unsigned height)
+    {
+        assert(width==Dim && height==Dim);
+        (void) width; (void) height;
```
endJunction: The width and height variables are not referenced in the documentation, so it's better to comment them in the arguments' list

ProcessLib/LocalAssemblerTraits.h
```diff
@@ -0,0 +1,211 @@
...
+    }
+};
+
+
+#ifndef EIGEN_DYNAMIC_SHAPE_MATRICES
+
+template<typename ShpPol, unsigned NIntPts, unsigned NodalDOF, unsigned Dim>
+using LocalAssemblerTraits = LocalAssemblerTraitsFixed<ShpPol, NIntPts, NodalDOF, Dim>;
+
+static_assert(EIGEN_DYNAMIC_SHAPE_MATRICES_FLAG == 0, "inconsistent use of macros");
```
endJunction: Please add the macro name and reason for the failure in the message.

ProcessLib/LocalAssemblerTraits.h
```diff
@@ -0,0 +1,210 @@
...
+    //      and use the same shape fucntions.
+    using LocalMatrix = Matrix<NIntPts*NodalDOF, NIntPts*NodalDOF>;
+    using LocalVector = Vector<NIntPts*NodalDOF>;
+
+    using Vector1Comp = typename ShapeMatrices::ShapeType;
+
+    using LaplaceMatrix = Matrix<Dim*NodalDOF, Dim*NodalDOF>;
+
+
+    // block dim x dim, fixed-size const matrix
```
wenqing: Doc. for doxygen.

chleh:
I added commit bdb0b08164a0846eeec1420c49cb4e45aaa8c769, which is a prerequisite that enables use of the traits struct.

chleh:
The failing windows ctest is due to some numerical issues in MathLib.DenseGaussAlgorithm. That's probably not related to this PR.

ProcessLib/LocalAssemblerTraits.h
```diff
@@ -0,0 +1,223 @@
...
+
+    //! Laplace matrix for the given space dimensionality and number of d.o.f
+    //! per node.
+    using LaplaceMatrix = Matrix<Dim*NodalDOF, Dim*NodalDOF>;
+
+
+    //! Get a block \c Dim x \c Dim whose upper left corner is at \c top and \c left.
+    template<typename Mat>
+    static
+    typename std::enable_if<NodalDOF != 0,
```
TomFischer: Only for my understanding: If the template parameter `NodelDOF` is not equal zero, i.e. the size is known at compile time, `blockDimDim` returns a fixed size Eigen matrix? Otherwise the implementation below is activated?

TomFischer:
:fast_forward:

ProcessLib/LocalAssemblerTraits.h
chleh: > Only for my understanding: If the template parameter NodelDOF is not equal zero, i.e. the size is known at compile time, blockDimDim returns a fixed size Eigen matrix? Otherwise the implementation below is activated?

Yes, if `NodalDOF != 0`, an Eigen block of fixed size is returned.
Admittedly, this is a bit hackish. I will improve the somewhat cryptc hint in the Doxygen comment of the struct

> This class takes advantage of the instantiations given below its definition.

And I will move the struct to a `detail` namespace, since it should not be used directly, but only via the `LocalAssemblerTraits` typedef.

PR #1098, opened 2016-03-14
[PL] shape matrices created in separate function

chleh:
As titled. Nothing fancy.


endJunction:
Could you please describe why we need the changes?

chleh:
Because those lines will be needed in every single local assembler that will be written. There are no changes, only a move of code.

wenqing:
I see.

chleh:
Do you have any further comments? This is a really small PR, so I hope, it can be merged fast.

TomFischer:
:fast_forward:

PR #1099, opened 2016-03-15
Fixing Mem-leaks in DataExplorer

rinkk:
this should fix the remaining leaks after PR #1091 on DataExplorer startup.

### DIFF
--- a/Applications/DataExplorer/VtkAct/VtkCustomInteractorStyle.h
+++ b/Applications/DataExplorer/VtkAct/VtkCustomInteractorStyle.h
@@ -72,13 +72,13 @@ protected:
 	virtual ~VtkCustomInteractorStyle();
 
 	/// @brief The vtk object to pick.
-	vtkDataObject* Data;
+	vtkDataObject* _data;
 
 	/// @brief The mapper for highlighting the selected cell.
-	vtkDataSetMapper* selectedMapper;
+	vtkDataSetMapper* _selectedMapper;
 
 	/// @brief The actor for highlighting the selected cell.
-	vtkActor* selectedActor;
+	vtkActor* _selectedActor;
 
 private:
 	bool _highlightActor;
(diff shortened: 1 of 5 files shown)


Applications/DataExplorer/VtkVis/VisualizationWidget.cpp
```diff
@@ -45,7 +45,7 @@
 #include <QString>
 
 VisualizationWidget::VisualizationWidget( QWidget* parent /*= 0*/ )
-	: QWidget(parent)
+	: _vtkRender(nullptr), _markerWidget(nullptr), QWidget(parent)
```
endJunction: warning: field '_markerWidget' will be initialized after base 'QWidget'
The order either in declarations or in the initialization list should be change and be the same.

chleh:
Looks good. :+1:

Applications/DataExplorer/VtkAct/VtkCustomInteractorStyle.h
```diff
@@ -72,13 +72,13 @@ public slots:
...
-	vtkDataObject* Data;
+	vtkDataObject* _data;
 
 	/// @brief The mapper for highlighting the selected cell.
-	vtkDataSetMapper* selectedMapper;
+	vtkDataSetMapper* _selectedMapper;
 
 	/// @brief The actor for highlighting the selected cell.
-	vtkActor* selectedActor;
+	vtkActor* _selectedActor;
```
TomFischer: Maybe it is possible to use `vtkSmartPointer` or `std::unique_ptr` for `_data` (line 75), `_selectedMapper` (line 78) or `_selectedActor`?

TomFischer:
One tiny question to answer, then :fast_forward:

Applications/DataExplorer/VtkAct/VtkCustomInteractorStyle.h
rinkk: I don't think so. Whenever the variables are used, VTK expects "normal" pointers, so I'd have to use ```.get()``` which makes the whole concept debatable in my opinion. On top of that I get error messages from stdlib when the unique_ptr should be deleted but VTK still uses the memory. If it's possible to do it, it's at least not straightforward.

PR #1100, opened 2016-03-15
ODESolver: Added pre and post iteration hooks

chleh:
Said hooks are called before and after each iteration of the nonlinear solver.

I need that functionality for TES in order to be able to drop some nonsensible iteration results.

### DIFF
--- a/NumLib/ODESolver/NonlinearSolver-impl.h
+++ b/NumLib/ODESolver/NonlinearSolver-impl.h
@@ -71,17 +71,17 @@ solve(Vector &x)
         {
             switch(sys.postIteration(x_new))
             {
-            case IterationResult::OK:
+            case IterationResult::SUCCESS:
                 // Don't copy here. The old x might still be used further below.
                 // Although currently it is not.
                 break;
-            case IterationResult::FAILED:
+            case IterationResult::FAILURE:
                 iteration_succeeded = false;
                 // Copy new solution to x.
                 // Thereby the failed solution can be used by the caller for debugging purposes.
                 BLAS::copy(x_new, x);
                 break;
-            case IterationResult::AGAIN:
+            case IterationResult::REPEAT_ITERATION:
                 continue; // That throws the iteration result away.
             }
         }
@@ -190,12 +190,12 @@ solve(Vector &x)
 
             switch(sys.postIteration(x_new))
             {
-            case IterationResult::OK:
+            case IterationResult::SUCCESS:
                 break;
-            case IterationResult::FAILED:
+            case IterationResult::FAILURE:
                 iteration_succeeded = false;
                 break;
-            case IterationResult::AGAIN:
+            case IterationResult::REPEAT_ITERATION:
                 // TODO introduce some onDestroy hook.
                 MathLib::GlobalVectorProvider<Vector>::provider.releaseVector(x_new);
                 continue; // That throws the iteration result away.
--- a/NumLib/ODESolver/NonlinearSolver-impl.h
+++ b/NumLib/ODESolver/NonlinearSolver-impl.h
@@ -43,13 +43,15 @@ solve(Vector &x)
     auto& rhs   = MathLib::GlobalVectorProvider<Vector>::provider.getVector(_rhs_id);
     auto& x_new = MathLib::GlobalVectorProvider<Vector>::provider.getVector(_x_new_id);
 
-    bool success = false;
+    bool error_norms_met = false;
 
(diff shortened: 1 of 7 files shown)


NumLib/ODESolver/NonlinearSolver-impl.h
```diff
@@ -43,13 +43,15 @@ solve(Vector &x)
...
+    bool error_norms_met = false;
 
     BLAS::copy(x, x_new); // set initial guess, TODO save the copy
 
     for (unsigned iteration=1; iteration<_maxiter; ++iteration)
     {
-        sys.assembleMatricesPicard(x_new);
+        sys.preIteration(iteration, x);
+
+        sys.assembleMatricesPicard(x);
```
endJunction: Why is `assembleMatricesPicard` now called with `x` and not with `x_new`?

NumLib/ODESolver/NonlinearSolver.h
```diff
@@ -122,6 +122,7 @@ class NonlinearSolver<Matrix, Vector, NonlinearSolverTag::Newton> final
     std::size_t _res_id = 0u;           //!< ID of the residual vector.
     std::size_t _J_id = 0u;             //!< ID of the Jacobian matrix.
     std::size_t _minus_delta_x_id = 0u; //!< ID of the \f$ -\Delta x\f$ vector.
+    std::size_t _x_new_id = 0u;         //!< ID of the vector storing \f$ x - (-\Delta x) \f$.
```
endJunction: How is the new variable used in pre/postIteration calls?

NumLib/ODESolver/NonlinearSolver-impl.h
chleh: At that point those two vectors have exactly the same values. But I think, it is more natural to call the assembly with something old (x) and not with something to be computed (x_new).
That's also more consistent with the Newton case.

NumLib/ODESolver/NonlinearSystem.h
```diff
@@ -20,6 +20,46 @@ namespace NumLib
 //! \addtogroup ODESolver
 //! @{
 
+//! Status flags telling the NonlinearSolver if an iteration succeeded.
+enum class IterationResult : char
+{
+    OK,     //!< Iteration succeeded.
+    FAILED, //!< Iteration failed irrecoverably.
+    AGAIN   //!< Iteration has to be repeated.
```
endJunction: Better to choose SUCCESS, FAILURE, and REPEAT_ITERATION, instead of comments.

NumLib/ODESolver/NonlinearSolver.h
chleh: Sorry, I don't understand the question. The use of the variable is up to the one who implements a process.
For `postIteration()` it is natural to implement some checks.
I don't have a good example for `preIteration()`. But you could precompute some cached values or initialize something, e.g. external software.
Passing `x` to `preIteration()` probably overlaps pretty much with what can be done in the `assemble...()` call.

chleh: Btw: the __new__ variable is only used in the `postIteration()` call.

chleh:
Sorry for the big change, but I noticed that the pre and post iteration hooks must be put in the `EquationSystem` interface s.t. both `ODESystem` and `NonlinearSystem` inherit them.

TomFischer:
:+1:

NumLib/ODESolver/NonlinearSolver-impl.h
```diff
@@ -146,21 +174,53 @@ solve(Vector &x)
...
+
+        if (!iteration_succeeded)
+        {
             ERR("The linear solver failed.");
-            BLAS::axpy(x, -_alpha, minus_delta_x);
-            success = false;
+        }
+        else
+        {
+            // TODO could be solved in a better way
```
wenqing: Since there are many todos (in this PR also the merged PR), I think it is better to use doxygen command ``\todo``, which can generates a todo list, to remind one what have to do.

wenqing:
:+1:

NumLib/ODESolver/NonlinearSolver-impl.h
chleh: I suggest to change all TODO tags to \todo in a separate PR, and then all at once for all of OGS6, if that's important.

endJunction:
Looks OK.

chleh:
Sorry, I forgot to check that I had to squash commits before merging.

PR #1102, opened 2016-03-15
[AsmL] extended VectorMatrixAssembler

chleh:
* renamed the standalone method
* added a method doing essentially the same as a member to the class

I use this method in the TES process to check if the primary variables are within given bounds, i.e., if they are positive.


TomFischer:
:+1:

PR #1104, opened 2016-03-16
Cleanup and mem leak fixing

TomFischer:
[GeoLib]
 - QuadTree: correction of destructor fixes mem leaks  

[FileIO]
- Initial implementation of destructor for
  * GMSHInterface
  * GMSHPolygonTree

### DIFF
--- a/FileIO/GmshIO/GMSHPolygonTree.cpp
+++ b/FileIO/GmshIO/GMSHPolygonTree.cpp
@@ -55,7 +55,7 @@ bool GMSHPolygonTree::insertStation(GeoLib::Point const* station)
 			if (((*it)->getPolygon())->isPntInPolygon (*station)) {
 				bool rval(dynamic_cast<GMSHPolygonTree*>((*it))->insertStation (station));
 				// stop recursion if sub SimplePolygonTree is a leaf
-				if (rval && (*it)->getNChilds() == 0)
+				if (rval && (*it)->getNChildren() == 0)
 					_stations.push_back (station);
 				return rval;
 			}
--- a/FileIO/GmshIO/GMSHPolygonTree.cpp
+++ b/FileIO/GmshIO/GMSHPolygonTree.cpp
@@ -42,6 +42,8 @@ GMSHPolygonTree::~GMSHPolygonTree()
 		for (auto * polyline : _plys)
 			delete polyline;
 	}
+	// member of GeoLib::SimplePolygonTree, but the ownership is not transmitted
+	delete _node_polygon;
 }
 
 bool GMSHPolygonTree::insertStation(GeoLib::Point const* station)
--- a/FileIO/GmshIO/GMSHPolygonTree.cpp
+++ b/FileIO/GmshIO/GMSHPolygonTree.cpp
@@ -50,8 +50,8 @@ bool GMSHPolygonTree::insertStation(GeoLib::Point const* station)
 {
 	if (_node_polygon->isPntInPolygon(*station)) {
 		// try to insert station into the child nodes
-		for (std::list<SimplePolygonTree*>::const_iterator it (_childs.begin());
-			 it != _childs.end(); ++it) {
+		for (std::list<SimplePolygonTree*>::const_iterator it (_children.begin());
+			 it != _children.end(); ++it) {
 			if (((*it)->getPolygon())->isPntInPolygon (*station)) {
 				bool rval(dynamic_cast<GMSHPolygonTree*>((*it))->insertStation (station));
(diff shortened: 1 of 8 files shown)


FileIO/GmshIO/GMSHPolygonTree.cpp
```diff
@@ -35,7 +35,16 @@ GMSHPolygonTree::GMSHPolygonTree(GeoLib::Polygon* polygon, GMSHPolygonTree* pare
 {}
 
 GMSHPolygonTree::~GMSHPolygonTree()
-{}
+{
+	// the polylines are processed also by the childs, but the root is
```
rinkk: "children"

FileIO/GmshIO/GMSHPolygonTree.cpp
```diff
@@ -35,7 +35,16 @@ GMSHPolygonTree::GMSHPolygonTree(GeoLib::Polygon* polygon, GMSHPolygonTree* pare
...
+	// responsible to cleanup up
+	if (_parent == nullptr) { // root
```
rinkk: is this *really* supposed to be ```==``` ?

GeoLib/QuadTree.h
```diff
@@ -71,9 +71,9 @@ template <typename POINT> class QuadTree
 	 */
 	~QuadTree()
 	{
-		if (_is_leaf)
-			for (std::size_t k(0); k < 4; k++)
-				delete _childs[k];
+		for (std::size_t k(0); k < 4; k++) {
+			delete _childs[k];
```
rinkk: technically, it doesn't make any difference, but it'd be good if you could rename "_childs" to "_children".

GeoLib/QuadTree.h
```diff
@@ -381,16 +384,20 @@ template <typename POINT> class QuadTree
 		POINT mid_point(_ll);
 		mid_point[0] += (_ur[0] - _ll[0]) / 2.0;
 		mid_point[1] += (_ur[1] - _ll[1]) / 2.0;
+		assert(_childs[0] == nullptr);
```
rinkk: it would be clearer if you could check for nullpointers at the beginning of the method in a for-loop or something.
also, is there a reason why you check this?

FileIO/GmshIO/GMSHPolygonTree.cpp
TomFischer: If `_parent` is a `nullptr` the node of the tree is the root. Only the root node is responsible for cleanup.

GeoLib/QuadTree.h
TomFischer: You are right. @endJunction mentioned this also. I'll try to fix this asap.

GeoLib/QuadTree.h
TomFischer: The assertions were quite useful to track down (my) programming errors. I would not gather them together at the beginning of the method, because I want to keep the assertions as near as possible at the location the condition should be valid.

norihiro-w:
Looks good apart from one minor comment

TomFischer:
> Maybe you can extract the checks into a function and use it from here and AABB?

- This check is for 2d, class AABB performs those checks in 3d.
- I don't know if it is a good idea to enlarge the interface of AABB to handle the special 2d xy case.
- Using a separate function needs to give `min` and `max` to this function. Using this separate function in the  AABB would be additional overhead. I expect that `AABB::containsPoint()` is called very often.

GeoLib/QuadTree.h
rinkk: My arguments would be that they are interupting the flow when reading the code and that code is executed even though you could already have known that something is wrong. If something is actually wrong you need to debug anyway. But it's no big issue, do it however you like.

endJunction:
Looks good, though I'd have tried to use unique_ptr's instead of dtor's deletes.

TomFischer:
> Looks good, though I'd have tried to use unique_ptr's instead of dtor's deletes.

- I tried to use `std::unique_ptr` with `GMSHInterface::_mesh_density_strategy`. Unfortunately  the `GMSHInterface::_mesh_density_strategy` is handed over to `GMSHPolygonTree`. Maybe a bigger re-design is necessary.
- The `_node_polygon` that is deleted in the destructor of `GMSHPolygonTree`. `GMSHPolygonTree` is derived from `SimplePolygonTree` and `_node_polygon` is a member from `SimplePolygonTree`. But `SimplePolygonTree` should not take responsibility to remove the `_node_polygon`. At the moment I don't see that `unique_ptr` is useful here.

PR #1105, opened 2016-03-16
Output Config

chleh:
This PR:
* adds a class for writing basic PVD files
* adds a class that handles output configuration

Output files now have names like `square_1e4_pcs_0_ts_1_t_1.000000.vtu`.
The output is configured by a snippet like the following in the prj file:
```
    <output>
        <type>VTK</type>
        <prefix>line_1</prefix>
        <timesteps>
            <pair><repeat> 10 </repeat><each_steps>  1 </each_steps></pair>
            <pair><repeat> 10 </repeat><each_steps>  1 </each_steps></pair>
            <pair><repeat> 10 </repeat><each_steps>  1 </each_steps></pair>
            <pair><repeat> 10 </repeat><each_steps>  1 </each_steps></pair>
            <pair><repeat> 10 </repeat><each_steps>  1 </each_steps></pair>
            <pair><repeat> 10 </repeat><each_steps>  1 </each_steps></pair>
            <pair><repeat>  1 </repeat><each_steps> 10 </each_steps></pair>
        </timesteps>
    </output>
```
where the `<timesteps>` part is optional and defaults to doing output in every timestep.

The input files changed accordingly: E.g. `<file>line_1</file>` → `<prefix>line_1</prefix>`.

### DIFF
--- a/Applications/ApplicationsLib/ProjectData.h
+++ b/Applications/ApplicationsLib/ProjectData.h
@@ -22,6 +22,7 @@
 #include "ProcessLib/ProcessVariable.h"
 #include "ProcessLib/Process.h"
 #include "ProcessLib/Parameter.h"
+#include "ProcessLib/Output.h"
 
 #include "NumLib/ODESolver/Types.h"
 
@@ -60,10 +61,13 @@ public:
 	ProjectData();
 
 	/// Constructs project data by parsing provided configuration.
-	/// The additional  path is used to find files referenced in the
-	/// configuration.
+	///
+	/// \param config_tree Configuration as read from the prj file.
+	/// \param project_directory Where to look for files referenced in the \c config_tree.
+	/// \param output_directory  Where to write simulation output files to.
 	ProjectData(BaseLib::ConfigTree const& config_tree,
-	            std::string const& path);
+	            std::string const& project_directory,
+	            std::string const& output_directory);
 
 	ProjectData(ProjectData&) = delete;
 
@@ -131,10 +135,16 @@ public:
 		return _processes.end();
 	}
 
-	std::string const&
-	getOutputFilePrefix() const
+	ProcessLib::Output<GlobalSetupType> const&
+	getOutputControl() const
 	{
-		return _output_file_prefix;
+		return *_output;
+	}
+
+	ProcessLib::Output<GlobalSetupType>&
+	getOutputControl()
+	{
+		return *_output;
 	}
 
 	TimeLoop& getTimeLoop()
@@ -174,7 +184,8 @@ private:
 
 	/// Parses the output configuration.
 	/// Parses the file tag and sets output file prefix.
-	void parseOutput(BaseLib::ConfigTree const& output_config, std::string const& path);
+	void parseOutput(BaseLib::ConfigTree const& output_config,
+	                 std::string const& output_directory);
 
 	void parseTimeStepping(BaseLib::ConfigTree const& timestepping_config);
 
@@ -195,8 +206,7 @@ private:
 	/// Buffer for each parameter config passed to the process.
 	std::vector<std::unique_ptr<ProcessLib::ParameterBase>> _parameters;
 
-	/// Output file path with project prefix.
-	std::string _output_file_prefix;
+	std::unique_ptr<ProcessLib::Output<GlobalSetupType>> _output;
 
 	/// The time loop used to solve this project's processes.
 	std::unique_ptr<TimeLoop> _time_loop;
--- a/Applications/ApplicationsLib/UncoupledProcessesTimeLoop.h
+++ b/Applications/ApplicationsLib/UncoupledProcessesTimeLoop.h
@@ -29,7 +29,7 @@ template<typename Matrix, typename Vector>
 class UncoupledProcessesTimeLoop
 {
 public:
-    bool loop(ProjectData& project, std::string const& outdir);
+    bool loop(ProjectData& project);
 
     ~UncoupledProcessesTimeLoop();
 
@@ -151,11 +151,9 @@ private:
 
     //! Solves one timestep for the given \c process.
     bool solveOneTimeStepOneProcess(
-            unsigned const timestep,
             Vector& x, double const t, double const delta_t,
             SingleProcessData& process_data,
-            Process& process,
-            std::string const& output_file_name);
+            Process& process);
 
     //! Sets the EquationSystem for the given nonlinear solver,
     //! which is Picard or Newton depending on the NLTag.
@@ -290,11 +288,9 @@ template<typename Matrix, typename Vector>
 bool
 UncoupledProcessesTimeLoop<Matrix, Vector>::
 solveOneTimeStepOneProcess(
-        unsigned const timestep,
         Vector& x, double const t, double const delta_t,
         SingleProcessData& process_data,
-        typename UncoupledProcessesTimeLoop<Matrix, Vector>::Process& process,
-        std::string const& output_file_name)
+        typename UncoupledProcessesTimeLoop<Matrix, Vector>::Process& process)
 {
     auto& time_disc        =  process.getTimeDiscretization();
     auto& ode_sys          = *process_data.tdisc_ode_sys;
@@ -314,7 +310,6 @@ solveOneTimeStepOneProcess(
     time_disc.pushState(t, x, mat_strg);
 
     process.postTimestep(x);
-    process.output(output_file_name, timestep, x);
 
     return nonlinear_solver_succeeded;
 }
@@ -323,54 +318,79 @@ solveOneTimeStepOneProcess(
 template<typename Matrix, typename Vector>
 bool
 UncoupledProcessesTimeLoop<Matrix, Vector>::
-loop(ProjectData& project, std::string const& outdir)
+loop(ProjectData& project)
 {
     auto per_process_data = initInternalData(project);
+
+    auto& out_ctrl = project.getOutputControl();
+    out_ctrl.init(project.processesBegin(), project.processesEnd());
 
     auto const t0 = 0.0; // time of the IC
 
     // init solution storage
     setInitialConditions(project, t0, per_process_data);
 
-    auto const delta_t = 1.0;
-
-    // TODO only for now
-    // Make sure there will be exactly one iteration of the loop below.
-    auto const t_end   = 1.5;
-
-    double t;
-    unsigned timestep = 1; // the first timestep really is number one
-    bool nonlinear_solver_succeeded = true;
-    for (t=t0+delta_t; t<t_end+delta_t; t+=delta_t, ++timestep)
+    // output initial conditions
     {
         unsigned pcs_idx = 0;
         for (auto p = project.processesBegin(); p != project.processesEnd();
              ++p, ++pcs_idx)
         {
-            std::string const& outpref = project.getOutputFilePrefix();
-            std::string const  output_file_name =
-                    BaseLib::joinPaths(outdir, outpref)
-                    + "_pcs_" + std::to_string(pcs_idx)
-                    + "_ts_"  + std::to_string(timestep)
-                    // + "_t_"   + std::to_string(t) // TODO: add that later
-                    + ".vtu";
+            auto const& x0 = *_process_solutions[pcs_idx];
+            out_ctrl.doOutput(**p, 0, t0, x0);
+        }
+    }
 
+    auto const delta_t = 1.0;
+
+    // TODO only for now
+    // Make sure there will be exactly one iteration of the loop below.
+    auto const t_end = 0.5;
+
+    double t, t_last;
+    unsigned timestep = 1; // the first timestep really is number one
+    bool nonlinear_solver_succeeded = true;
+
+    for (t=t0+delta_t, t_last=t0;
+         t<t_end+delta_t;
+         t_last=t, t+=delta_t, ++timestep)
+    {
+        // TODO use process name
+        unsigned pcs_idx = 0;
+        for (auto p = project.processesBegin(); p != project.processesEnd();
+             ++p, ++pcs_idx)
+        {
             auto& x = *_process_solutions[pcs_idx];
 
             nonlinear_solver_succeeded = solveOneTimeStepOneProcess(
-                        timestep, x, t, delta_t,
-                        per_process_data[pcs_idx], **p, output_file_name);
+                        x, t, delta_t, per_process_data[pcs_idx], **p);
 
             if (!nonlinear_solver_succeeded) {
                 ERR("The nonlinear solver failed in timestep #%u at t = %g s"
                     " for process #%u.", timestep, t, pcs_idx);
+
+                // save unsuccessful solution
+                out_ctrl.doOutputAlways(**p, timestep, t, x);
+
                 break;
+            } else {
+                out_ctrl.doOutput(**p, timestep, t, x);
             }
         }
 
         if (!nonlinear_solver_succeeded) break;
+    }
 
-        break; // TODO only do a single timestep for now
+    // output last timestep
+    if (nonlinear_solver_succeeded)
+    {
+        unsigned pcs_idx = 0;
+        for (auto p = project.processesBegin(); p != project.processesEnd();
+             ++p, ++pcs_idx)
+        {
+            auto const& x = *_process_solutions[pcs_idx];
+            out_ctrl.doOutputLastTimestep(**p, timestep-1, t_last, x);
+        }
     }
 
     return nonlinear_solver_succeeded;
--- a/FileIO/VtkIO/PVDFile.cpp
+++ b/FileIO/VtkIO/PVDFile.cpp
@@ -7,9 +7,9 @@
  *
  */
 
-#include <fstream>
 #include "PVDFile.h"
 
+#include <fstream>
 #include <iomanip>
 #include <limits>
 #include <logog/include/logog.hpp>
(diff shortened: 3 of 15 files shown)


chleh:
Looks like g++ 4.8's lib is too old. ^^ Gonna workaround.

ProcessLib/Output.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+    auto const pcs = 0u; // TODO fix
+
+    std::string const output_file_name =
+            _output_file_prefix + "_pcs_" + std::to_string(pcs)
+            + "_ts_" + std::to_string(timestep)
+            + "_t_"  + std::to_string(t)
+            + ".vtu";
+    DBUG("output to %s", output_file_name.c_str());
+    // auto const& p = _processes[pcs];
```
chleh: TODO remove.

FileIO/VtkIO/PVDFile.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+PVDFile::PVDFile(const std::string &fn)
+    : _fh{fn.c_str()}
+{
+    if (!_fh) {
+        ERR("could not open file `%s'", fn.c_str());
+        std::abort();
+    }
+
+    _fh << std::setprecision(16);
```
TomFischer: Why "16"? Maybe you could use `std::numeric_limits<double>::digits10`?

FileIO/VtkIO/PVDFile.cpp
```diff
@@ -0,0 +1,45 @@
...
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#include <fstream>
+#include <iomanip>
+#include "PVDFile.h"
+
+#include "logog/include/logog.hpp"
```
chleh: TODO use <>

chleh:
Fixing PETScVector in this PR needs too many changes. So I removed one commit.

Applications/ApplicationsLib/ProjectData.h
```diff
@@ -63,7 +64,8 @@ class ProjectData final
 	/// The additional  path is used to find files referenced in the
 	/// configuration.
 	ProjectData(BaseLib::ConfigTree const& config_tree,
-	            std::string const& path);
+	            std::string const& path,
+	            std::string const& output_directory);
```
TomFischer: - Maybe `path` should be renamed to `input_path` or `input_directory` to reflect better that this variable is used for input in contrast to the `output_directory`? 
- I would prefer `output_path` instead of `output_directory`.
- For consistency I would either document all parameters or none. At the moment `path` is documented, `output_directory` is not.

FileIO/VtkIO/PVDFile.cpp
```diff
@@ -0,0 +1,45 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2015, OpenGeoSys Community (http://www.opengeosys.org)
```
TomFischer: 2012-2016

norihiro-w:
General comments
- why create vtk file for each process? This is inconvenient and using more disk space
- about time information in vtk file names: what is format? Scientific or fixed? How many digits? Can Paraview recognize their sequence without a PVD file?
- It's convenient if a PVD file is updated in a valid form at each output. You don't want to check results during simulations?

norihiro-w:
For config, I would use repeat instead of count

chleh:
> General comments
- why create vtk file for each process? This is inconvenient and using more disk space

When writing only one VTK file there could be name clashes for different outputs, especially with secondary process variables. Having one file per process does not have such problems. Actually I don't even know, if this approach will work for multiple processes. Currently there is only one. Probably with the presented implementation every two processes on the same mesh will effectively write almost identical VTK files (I won't fix that in this PR!). However different processes could live on different meshes. So probably each process should have their own output.

> - about time information in vtk file names: what is format? Scientific or fixed? How many digits? 

`std::to_string` generates fixed output with 6 decimal places. Cf. http://en.cppreference.com/w/cpp/string/basic_string/to_string and http://en.cppreference.com/w/cpp/io/c/fprintf

> Can Paraview recognize their sequence without a PVD file?

I don't know. But why can't you simply open the PVD file? (Probably not during simulation...)

> - It's convenient if a PVD file is updated in a valid form at each output. You don't want to check results during simulations?

I think that can be done in a subsequent PR. This PR is meant as a first attempt to encapsulate output control in a separate class. If you need more, nobody keeps you from extending this functionality.

Applications/ApplicationsLib/ProjectData.h
chleh: I prefer the suffix `_directory`, because the arguments refer to directories.
From `..._path` it is not directly visible if files or dirs are meant.

norihiro-w:
First of all, I'm not familiar with current implementation (in simulation part) any more so that I can comment only on what features we need and what are missing. Depending on the amount of works needed to implement the requirement, you guys can decide if it's worth to implement them in this PR or separate PRs. The followings are my reply to each comment you gave above:
- You are thinking output is actually dependent on a mesh, which I also agree. then why not having a such interface instead of process-dependent? Since the two approaches need different code structure, it's better if we implement it as soon as possible before another process type is coming in. Concerning variable names, we just need to give different names (e.g. (pcs name)_(var name)) or ask a user to name them. 
- I would remove time information from the file names. Though exceeding 6 decimal places can merely happen, there is no convincing reason to include such an unnecessary limitation which we many need to fix in the future. Having a time step number in the file name is enough to distinguish files from different times. and anyway time information is stored in a PVD file.
- It helps when you don't have a PVD or when you lost it, or when it is broken.
- That can be done in a separate PR if it will not change code structure or interface a lot.

chleh:
I'd like to make as few changes as possible to this PR now, because there are only today and tomorrow left before my vacation. Sure, that's a lame excuse. But (i) I need this PR for the upcoming TES process, and (ii) I won't have time in april for doing such things.

Your comments are important for the long run, but currently I simply don't have time.

> I would remove time information from the file names. [...] Having a time step number in the file name is enough to distinguish files from different times. [...]

I'd keep it there. Especially since the filenames are already unique without the time information.
Adding time information makes it easy to see from the file name with which time you are dealing. I think, it's a rather handy feature when handling output files.

> It helps when you don't have a PVD or when you lost it, or when it is broken.

What? The time in the file name? Then I agree.

norihiro-w:
Sorry I don't totally get it. Maybe thre is something I don't know about current development way. In general, if you need some code to run your application or you don't have time for improvement, why do you need to make a pull request? You just need to stick with your branch, don't you? If we merge PRs because of these reasons, there is no point to review the codes.

chleh:
The merge of the TES process into master has been planned since last september. Furthermore: The inner parts of OGS6 are constantly moving. I don't have the time to maintain my branches all the time.

Furthermore: I think it is much better to include a simple implementation of some feature early and extend it later on, than to wait until that feature is perfect. If we would do the latter, OGS6 maybe wouldn't have moved in any direction in the last six months.

Also: I think the quality of my code is not bad. It is just not feature-complete.

norihiro-w:
> The merge of the TES process into master has been planned since last september. Furthermore: The inner parts of OGS6 are constantly moving. I don't have the time to maintain my branches all the time.

That's your problem and doesn't give any reason to merge PRs because you don't have time or you need it asap. 

> Furthermore: I think it is much better to include a simple implementation of some feature early and extend it later on, than to wait until that feature is perfect. If we would do the latter, OGS6 maybe wouldn't have moved in any direction in the last six months.

This way of answering is much better than above. Simpler is better, but too simple is not god.

> Also: I think the quality of my code is not bad. It is just not feature-complete.

This is what code reviewers decide for each PR.

norihiro-w:
@endJunction  to be discussed

norihiro-w:
> I'd keep it there. Especially since the filenames are already unique without the time information.
Adding time information makes it easy to see from the file name with which time you are dealing. I think, it's a rather handy feature when handling output files.

Go ahead if you insist it. 

> What? The time in the file name? Then I agree.

What? I don't understand your reply.

Applications/ApplicationsLib/UncoupledProcessesTimeLoop.h
```diff
@@ -29,7 +29,7 @@ template<typename Matrix, typename Vector>
 class UncoupledProcessesTimeLoop
 {
 public:
-    bool loop(ProjectData& project, std::string const& outdir);
+    bool loop(ProjectData& project);
```
norihiro-w: Doesn't it makes sense to pass project in constructor?

FileIO/VtkIO/PVDFile.cpp
```diff
@@ -0,0 +1,46 @@
...
+}
+
+void PVDFile::addVTUFile(const std::string &fn, double timestep)
+{
+    _fh << "    <DataSet timestep=\"" << timestep << "\" group=\"\" part=\"0\" file=\"" << fn << "\"/>\n";
+}
+
+PVDFile::~PVDFile()
+{
+    _fh << "  </Collection>\n</VTKFile>\n";
```
norihiro-w: I don't think having an output in destructor is a good idea. Destructor should be used only for finalizing resources. Moreover, it's not an intuitive interface. A user of this class has to delete an object to get a complete output.

FileIO/VtkIO/PVDFile.cpp
```diff
@@ -0,0 +1,46 @@
...
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#include <fstream>
+#include <iomanip>
+#include <limits>
+#include <logog/include/logog.hpp>
+
+#include "PVDFile.h"
```
norihiro-w: should be included at the top 

see http://ufz.github.io/styleguide/cppguide.xml?showone=Names_and_Order_of_Includes#Names_and_Order_of_Includes

ProcessLib/Output.cpp
```diff
@@ -0,0 +1,146 @@
...
+ */
+
+#include<cassert>
+#include<fstream>
+#include<vector>
+
+#include <logog/include/logog.hpp>
+
+#include "BaseLib/FileTools.h"
+#include "Output.h"
```
norihiro-w: include at the top

ProcessLib/Output.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+namespace
+{
+//! Determines if there should be output at the given \c timestep.
+template<typename CountsSteps>
+bool doOutput(std::size_t timestep, CountsSteps const& repeats_each_steps)
```
norihiro-w: i would name it needOutput(). doOutput sounds something is written into a file

ProcessLib/Output.cpp
```diff
@@ -0,0 +1,146 @@
...
+        for (auto pair : timesteps->getConfSubtreeList("pair"))
+        {
+            auto repeat      = pair.getConfParam<std::size_t>("repeat");
+            auto each_steps = pair.getConfParam<std::size_t>("each_steps");
+
+            assert(repeat != 0 && each_steps != 0);
+            out->_repeats_each_steps.emplace_back(repeat, each_steps);
+        }
+
+        if (out->_repeats_each_steps.size() == 0) {
```
norihiro-w: recommend to use `empty()`

Applications/ApplicationsLib/UncoupledProcessesTimeLoop.h
chleh: But this way the time loop can be constructed independently from the project.
Also: It has been like that before this PR.

FileIO/VtkIO/PVDFile.cpp
chleh: I don't see any problem.

> Destructor should be used only for finalizing resources.

It only finalizes the XML data.

> Moreover, it's not an intuitive interface. A user of this class has to delete an object to get a complete output.

I think, there's no problem. Actually: That's rather intuitive. Since objects are destructed automatically, this implementation asserts that you cannot forget to call some `close()` routine.

chleh:
@norihiro-w Regarding your fundamental objections against this PR: Will you be in the office today? Otherwise I cannot discuss that with you for two weeks, now.

One plea: If you have very substantial comments to a PR (as you have here), please just come over to my office and discuss them. I think that is much easier and more successful than having a discussion on github.

ProcessLib/Output.cpp
chleh: Renamed to `shallDoOutput()`.

FileIO/VtkIO/PVDFile.cpp
chleh: I pulled it further to the top. But it must be after <fstream> since the ofstream is only forward-declared in PVDFile.h.

ProcessLib/Output.cpp
chleh: Also: must come after <fstream>

norihiro-w:
I'm not sure if I can come to the office today. My daughter is sick and has to be taken care. I have just brought her to a doctor.

norihiro-w:
@chleh I don't think I can work on this today. If you really think merging this PR is urgent, please discuss with others, e.g. @endJunction. If they think no further modification is required, go ahead.

chleh:
I hope your daughter will get well soon.

Applications/ApplicationsLib/UncoupledProcessesTimeLoop.h
norihiro-w: i see

FileIO/VtkIO/PVDFile.cpp
norihiro-w: From my point of view, the XML data is not resource, rather outcome of this functionality. 

I think the interface is not intuitive because usually we (maybe only me?) do not expect a destructor doing something beyond resource finalization (completing the XML data is not included here). Besides, one needs to write the following codes to make it sure the XML is complete at some point. 

```
{
PVDFile pvd("test.pvd");
pvd.addVTUFile();
}

// do something using the PVD file
```

The following codes looks more readable 
```
PVDFile pvd("test.pvd");
pvd.addVTUFile();

// do something using the PVD file
```

or 

```
PVDFile pvd("test.pvd");
pvd.addVTUFile();
pvd.close();

// do something using the PVD file
```

ProcessLib/Output.cpp
```diff
@@ -0,0 +1,147 @@
...
+{
+    auto spd_it = _single_process_data.find(&process);
+    if (spd_it == _single_process_data.end()) {
+        ERR("The given process is not contained in the output configuration."
+            " Aborting.");
+        std::abort();
+    }
+    auto& spd = spd_it->second;
+
+    std::string const output_file_name =
```
norihiro-w: I changed my mind and got idea that it might be better to remove a time step number from the VTU file name, i.e. only including only time information. There is a problem in ogs5 benchmarking that benchmarks using adaptive time stepping can produce result files with different names because the file name includes a time step number and the number can change when someone changed numerics.

FileIO/VtkIO/PVDFile.cpp
```diff
@@ -0,0 +1,46 @@
...
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#include <fstream>
+#include "PVDFile.h"
```
wenqing: ``#include "PVDFile.h"`` should appear as the first including.

norihiro-w:
@chleh I sent a PR to your branch, which improves PVDFile class and change a VTU file name.

ProcessLib/Output.cpp
chleh: I think the timestep number should be kept in the output file name since it guarantees unique file names; in contrast `std::to_string(t)` does not (probably it will give unique file names for all practical situations, but what if 6 decimal places don't suffice?).
Regarding your adaptive timestepping argument: Probably there should be different timestep numbers only if we change the adaptivity algorithm; and if we do so, not only the name but also the contents of the files will change. Thus, a change in timestep number would mean a rather fundamental change in our code and there will be manual work anyway, so we can also change the timestep numbers in the filename. Or did you get different timestep numbers with the same configuration on different machines? If so: Might that point to a serious problem, since adaptivity is not stable when changing machines?
Furthermore: The output is currently controlled by timestep number, not by time, so changing the former would also change tha latter. Sure, for adaptivity that will have to change (in a future PR).
To summarize: I think we should keep the timestep number for the time being.

chleh:
@norihiro-w Thanks for your PR-PR. It pretty much reflects what you've written in your comments above :smile:
I will merge it when I'm back from vacation.

ProcessLib/Output.cpp
norihiro-w: no we didn't change the algorithm. Whatever changes (bugfix, change of arithmetic operation orders) in assembly can cause different evolution of time stepping.

norihiro-w:
@chleh No further comments from me at the moment. Remaining issues (mesh-dependent output and the naming of VTU files) can be fixed later. From next time, please make a PR smaller, focusing only one feature or a topic for discussion in each PR. It makes reviewing and discussions easier and better.

norihiro-w:
i will check jenkins tomorrow

chleh:
All major discussions settled, now. So, please review this.

norihiro-w:
:+1 from me

ProcessLib/Output.cpp
```diff
@@ -0,0 +1,147 @@
...
+            _output_file_prefix + "_pcs_" + std::to_string(spd.process_index)
+            + "_ts_" + std::to_string(timestep)
+            + "_t_"  + std::to_string(t)
+            + ".vtu";
+    DBUG("output to %s", output_file_name.c_str());
+    process.output(output_file_name, timestep, x);
```
endJunction: Not important notice: In Process::output() the timestep is an unsigned int... Could be consistent

endJunction:
I used this branch together with the time loop PR in M-process and it worked.

ProcessLib/Output.cpp
chleh: Changed the signatures to `unsigned` in the `Output` class.

endJunction:
Jenkins, test this please.

endJunction:
@chleh The ufz/ogs/master needs a reference to the new ufz/ogs-data/master commit for Test/Data submodule. Now it's pointing to ufz/ogs-data/changed-process-output-config...

endJunction:
@chleh Now some of the ctest's are failing. Please have a look into https://svn.ufz.de:8443/job/OGS-6/job/Tests_Linux_Large/318/
There are additional commits on top, but touching only the Changelog.

PR #1106, opened 2016-03-16
Fix PETScVector

chleh:
* fixes copy ctor and makes it explicit
* adds move ctor

### DIFF
--- a/MathLib/LinAlg/PETSc/PETScVector.cpp
+++ b/MathLib/LinAlg/PETSc/PETScVector.cpp
@@ -77,6 +77,17 @@ PETScVector::PETScVector(const PETScVector &existing_vec, const bool deep_copy)
     }
 }
 
+PETScVector::PETScVector(PETScVector &&other)
+    : _v{std::move(other._v)}
+    , _v_loc{std::move(other._v_loc)}
+    , _start_rank{other._start_rank}
+    , _end_rank{other._end_rank}
+    , _size{other._size}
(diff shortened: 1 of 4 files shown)


endJunction:
A note aside: is there an obvious reason not to implement the assignment operators?

MathLib/LinAlg/PETSc/PETScVector.cpp
```diff
@@ -77,6 +77,17 @@ PETScVector::PETScVector(const PETScVector &existing_vec, const bool deep_copy)
     }
 }
 
+PETScVector::PETScVector(PETScVector &&other)
+    : _v{std::move(other._v)}
```
norihiro-w: Why are using {} instead of  ()? Make it consistent with another constructor

chleh: Isn't it consistent within this file?

norihiro-w: if you ask me so, my answer is "yes i do". Loosing consistency is not very minor thing for me.

chleh: Now all initializer lists in `PETScVector.*` use braces for construction.

chleh:
@endJunction I deleted the move assignment operator explicitly as discussed.

norihiro-w:
:+1:

PR #1107, opened 2016-03-17
[FileIO] Gmsh: Respect mesh density scaling factor at stations

TomFischer:
Changes in PR opens the possibility to refine a mesh near stations.


FileIO/GmshIO/GMSHNoMeshDensity.h
```diff
@@ -37,6 +37,14 @@ class GMSHNoMeshDensity: public FileIO::GMSH::GMSHMeshDensityStrategy {
 		(void)(pnt);
 		return 0.0;
 	}
+
+	double getMeshDensityAtStation(GeoLib::Point const*const pnt) const
```
endJunction: just comment /*pnt*/

FileIO/GmshIO/GMSHFixedMeshDensity.cpp
```diff
@@ -36,5 +36,12 @@ double GMSHFixedMeshDensity::getMeshDensityAtPoint(GeoLib::Point const*const pnt
 	return _mesh_density;
 }
 
+double GMSHFixedMeshDensity::getMeshDensityAtStation(GeoLib::Point const*const pnt) const
```
endJunction: /*pnt*/

FileIO/GMSHInterface.cpp
```diff
@@ -411,22 +412,30 @@ void GMSHInterface::writeGMSHInputFile(std::ostream& out)
...
 		if (stations) {
-			const std::size_t n_stations(stations->size());
-			for (std::size_t k(0); k < n_stations; k++) {
+			for (auto * station : *stations) {
 				bool found(false);
 				for (std::list<GMSH::GMSHPolygonTree*>::iterator it(_polygon_tree_list.begin());
 					it != _polygon_tree_list.end() && !found; ++it) {
-					if ((*it)->insertStation((*stations)[k])) {
+					GeoLib::Station *station_pnt(new GeoLib::Station(* static_cast<GeoLib::Station*>(station)));
+					gmsh_stations->push_back(station_pnt);
```
endJunction: `emplace_back` as discussed...

endJunction:
:fast_forward:

FileIO/GmshIO/GMSHNoMeshDensity.h
```diff
@@ -31,12 +31,16 @@ class GMSHNoMeshDensity: public FileIO::GMSH::GMSHMeshDensityStrategy {
 		(void)(vec);
 	}
 
-	double getMeshDensityAtPoint(GeoLib::Point const*const pnt) const
+	double getMeshDensityAtPoint(GeoLib::Point const*const) const
+	{
+		return 0.0;
```
rinkk: Why do these automatically return 0?

rinkk:
Does this PR actually change anything in the mesh generation or is it just bug fixing / rewriting?

Mesh desity around stations would be respected before so I'm not sure what changed. Are GeoLib::Points now included in mesh generation if they are not part of a polyline or surface?

FileIO/GmshIO/GMSHNoMeshDensity.h
TomFischer: Good point. The implementation was required since the virtual base class was extented. Today I discovered that the class `GMSHNoMeshDensity` is not used. I will delete it.

TomFischer:
> Does this PR actually change anything in the mesh generation ...

Yes, it does.

> ... or is it just bug fixing / rewriting?

Yes, it is.

> Mesh desity around stations would be respected before so I'm not sure what changed.

Before the wrong scaling parameter was used.

> Are GeoLib::Points now included in mesh generation if they are not part of a polyline or surface?

No, `GeoLib::Point`s are only include if they are part of a polyline or surface. If the mesh should be refined to a special point the point has to be included as `GeoLib::Station`.

rinkk:
Thanks for all the info. :+1:

PR #1108, opened 2016-03-17
Using existing FixedTimeStepper in time loop

chleh:
This is a follow-up PR of #1105. Only the last four commits matter.

* Extends FixedTimeStepper builder to support different timestep sizes.
* Uses FixedTimeStepper in the time loop

### DIFF
--- a/Applications/ApplicationsLib/UncoupledProcessesTimeLoop.h
+++ b/Applications/ApplicationsLib/UncoupledProcessesTimeLoop.h
@@ -17,6 +17,7 @@
 #include "BaseLib/ConfigTree.h"
 #include "NumLib/ODESolver/TimeDiscretizedODESystem.h"
 #include "NumLib/ODESolver/NonlinearSolver.h"
+#include "NumLib/TimeStepping/Algorithms/FixedTimeStepping.h"
 
 #include "ProjectData.h"
 
@@ -29,6 +30,11 @@ template<typename Matrix, typename Vector>
 class UncoupledProcessesTimeLoop
 {
 public:
+    explicit UncoupledProcessesTimeLoop(
+            std::unique_ptr<NumLib::ITimeStepAlgorithm>&& timestepper)
+        : _timestepper{std::move(timestepper)}
+    {}
+
     bool loop(ProjectData& project);
 
     ~UncoupledProcessesTimeLoop();
@@ -44,6 +50,7 @@ private:
     using TimeDisc         = NumLib::TimeDiscretization<Vector>;
 
     std::vector<Vector*> _process_solutions;
+    std::unique_ptr<NumLib::ITimeStepAlgorithm> _timestepper;
 
     struct SingleProcessData
     {
@@ -199,16 +206,20 @@ createUncoupledProcessesTimeLoop(BaseLib::ConfigTree const& conf)
 {
     auto const type = conf.getConfParam<std::string>("type");
 
-    if (type == "SingleStep")
-    {
-        using TimeLoop = UncoupledProcessesTimeLoop<Matrix, Vector>;
-        return std::unique_ptr<TimeLoop>(new TimeLoop);
-    }
-    else
-    {
+    std::unique_ptr<NumLib::ITimeStepAlgorithm> timestepper;
+
+    if (type == "SingleStep") {
+        conf.ignoreConfParam("type");
+        timestepper.reset(new NumLib::FixedTimeStepping(0.0, 1.0, 1.0));
+    } else if (type == "FixedTimeStepping") {
+        timestepper = NumLib::FixedTimeStepping::newInstance(conf);
+    } else {
             ERR("Unknown timestepper type: `%s'.", type.c_str());
             std::abort();
     }
+
+    using TimeLoop = UncoupledProcessesTimeLoop<Matrix, Vector>;
+    return std::unique_ptr<TimeLoop>{new TimeLoop{std::move(timestepper)}};
 }
 
 
--- a/Applications/ApplicationsLib/UncoupledProcessesTimeLoop.h
+++ b/Applications/ApplicationsLib/UncoupledProcessesTimeLoop.h
@@ -204,7 +204,7 @@ template<typename Matrix, typename Vector>
 std::unique_ptr<UncoupledProcessesTimeLoop<Matrix, Vector> >
 createUncoupledProcessesTimeLoop(BaseLib::ConfigTree const& conf)
 {
-    auto const type = conf.getConfParam<std::string>("type");
+    auto const type = conf.peekConfParam<std::string>("type");
 
     std::unique_ptr<NumLib::ITimeStepAlgorithm> timestepper;
 
@@ -336,7 +336,7 @@ loop(ProjectData& project)
     auto& out_ctrl = project.getOutputControl();
     out_ctrl.init(project.processesBegin(), project.processesEnd());
 
-    auto const t0 = 0.0; // time of the IC
+    auto const t0 = _timestepper->getTimeStep().current(); // time of the IC
 
     // init solution storage
     setInitialConditions(project, t0, per_process_data);
@@ -352,20 +352,17 @@ loop(ProjectData& project)
         }
     }
 
-    auto const delta_t = 1.0;
-
-    // TODO only for now
-    // Make sure there will be exactly one iteration of the loop below.
-    auto const t_end = 0.5;
-
-    double t, t_last;
-    unsigned timestep = 1; // the first timestep really is number one
+    double t = t0;
+    std::size_t timestep = 1; // the first timestep really is number one
     bool nonlinear_solver_succeeded = true;
 
-    for (t=t0+delta_t, t_last=t0;
-         t<t_end+delta_t;
-         t_last=t, t+=delta_t, ++timestep)
+    while (_timestepper->next())
     {
+        auto const ts = _timestepper->getTimeStep();
+        auto const delta_t = ts.dt();
+        t        = ts.current();
+        timestep = ts.steps();
+
         // TODO use process name
         unsigned pcs_idx = 0;
         for (auto p = project.processesBegin(); p != project.processesEnd();
@@ -400,7 +397,7 @@ loop(ProjectData& project)
              ++p, ++pcs_idx)
         {
             auto const& x = *_process_solutions[pcs_idx];
-            out_ctrl.doOutputLastTimestep(**p, timestep-1, t_last, x);
+            out_ctrl.doOutputLastTimestep(**p, timestep, t, x);
         }
     }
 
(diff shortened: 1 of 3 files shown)


Applications/ApplicationsLib/UncoupledProcessesTimeLoop.h
```diff
@@ -29,7 +30,12 @@ template<typename Matrix, typename Vector>
 class UncoupledProcessesTimeLoop
 {
 public:
-    bool loop(ProjectData& project, std::string const& outdir);
+    bool loop(ProjectData& project);
+
+    explicit UncoupledProcessesTimeLoop(
```
norihiro-w: Constructors and Destructor should come before declaration of methods

see http://ufz.github.io/styleguide/cppguide.xml?showone=Declaration_Order#Declaration_Order

Applications/ApplicationsLib/UncoupledProcessesTimeLoop.h
```diff
@@ -29,7 +30,12 @@ template<typename Matrix, typename Vector>
 class UncoupledProcessesTimeLoop
 {
 public:
-    bool loop(ProjectData& project, std::string const& outdir);
+    bool loop(ProjectData& project);
```
norihiro-w: Don't it makes sense to pass `project` in constructor?

Applications/ApplicationsLib/UncoupledProcessesTimeLoop.h
```diff
@@ -199,18 +204,22 @@ template<typename Matrix, typename Vector>
...
-    if (type == "SingleStep")
-    {
-        using TimeLoop = UncoupledProcessesTimeLoop<Matrix, Vector>;
-        return std::unique_ptr<TimeLoop>(new TimeLoop);
-    }
-    else
-    {
+    std::unique_ptr<NumLib::ITimeStepAlgorithm> timestepper;
+
+    if (type == "SingleStep") {
```
norihiro-w: Do you still need SingleStep? you can do it with FixiedTimeStepping now.

FileIO/VtkIO/PVDFile.cpp
```diff
@@ -0,0 +1,46 @@
...
+}
+
+void PVDFile::addVTUFile(const std::string &fn, double timestep)
+{
+    _fh << "    <DataSet timestep=\"" << timestep << "\" group=\"\" part=\"0\" file=\"" << fn << "\"/>\n";
+}
+
+PVDFile::~PVDFile()
+{
+    _fh << "  </Collection>\n</VTKFile>\n";
```
norihiro-w: I don't think having an output in destructor is a good idea. Destructor should be used only for finalizing resources. Moreover, it's not an intuitive interface. A user of this class have to delete an object to get a complete output.

NumLib/TimeStepping/Algorithms/FixedTimeStepping.cpp
```diff
@@ -9,16 +9,16 @@
...
 #include <cmath>
 #include <numeric>
 #include <limits>
 #include <cassert>
 
+#include <logog/include/logog.hpp>
+
 #include "BaseLib/ConfigTree.h"
-#include "logog/include/logog.hpp"
+#include "FixedTimeStepping.h"
```
norihiro-w: should be included at the top

NumLib/TimeStepping/Algorithms/FixedTimeStepping.cpp
```diff
@@ -31,16 +31,54 @@ FixedTimeStepping::FixedTimeStepping(double t0, double tn, double dt)
...
+    for (auto const pair : range)
+    {
+        auto const repeat = pair.getConfParam<std::size_t>("repeat");
+        delta_t           = pair.getConfParam<double>("delta_t");
+
+        if (repeat == 0) {
+            ERR("<repeat> is zero.");
+            std::abort();
+        }
+        if (delta_t < 0.0) {
```
norihiro-w: is dt == 0 acceptable?

NumLib/TimeStepping/Algorithms/FixedTimeStepping.cpp
```diff
@@ -31,16 +31,54 @@ FixedTimeStepping::FixedTimeStepping(double t0, double tn, double dt)
...
+            ERR("timestep <delta_t> is smaller than zero.");
+            std::abort();
+        }
+
+        if (t_curr <= t_end) {
```
norihiro-w: i guess `<` is correct here. e.g. in case `t_initial` == `t_end`

NumLib/TimeStepping/Algorithms/FixedTimeStepping.cpp
```diff
@@ -31,16 +31,54 @@ FixedTimeStepping::FixedTimeStepping(double t0, double tn, double dt)
...
+            timesteps.resize(timesteps.size() + repeat, delta_t);
+
+            t_curr += repeat * delta_t;
+        }
+    }
+
+    // append last delta_t until t_end is reached
+    if (t_curr < t_end)
```
norihiro-w: have yon considered if `t_curr > t_end`?

ProcessLib/Output.cpp
```diff
@@ -0,0 +1,146 @@
...
+ */
+
+#include<cassert>
+#include<fstream>
+#include<vector>
+
+#include <logog/include/logog.hpp>
+
+#include "BaseLib/FileTools.h"
+#include "Output.h"
```
norihiro-w: include at the top

ProcessLib/Output.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+namespace
+{
+//! Determines if there should be output at the given \c timestep.
+template<typename CountsSteps>
+bool doOutput(std::size_t timestep, CountsSteps const& repeats_each_steps)
```
norihiro-w: i would name it `needOutput()`. `doOutput` sounds something is written into a file

ProcessLib/Output.cpp
```diff
@@ -0,0 +1,146 @@
...
+        for (auto pair : timesteps->getConfSubtreeList("pair"))
+        {
+            auto repeat      = pair.getConfParam<std::size_t>("repeat");
+            auto each_steps = pair.getConfParam<std::size_t>("each_steps");
+
+            assert(repeat != 0 && each_steps != 0);
+            out->_repeats_each_steps.emplace_back(repeat, each_steps);
+        }
+
+        if (out->_repeats_each_steps.size() == 0) {
```
norihiro-w: recommend to use `empty()`

norihiro-w:
Sorry I made some comments here but they should be in another PR. Those comments were already moved to the right PR.

Applications/ApplicationsLib/UncoupledProcessesTimeLoop.h
chleh: But "SingleStep" is more explicit than having to configure `FixiedTimeStepping` with suitable t0, t_end, delta_t s.t. only exactly one timestep is computed.

NumLib/TimeStepping/Algorithms/FixedTimeStepping.cpp
chleh: Probably not. Thanks for pointing this out.

NumLib/TimeStepping/Algorithms/FixedTimeStepping.cpp
chleh: I think that is handled in the `FixedTimeStepping::next()` method.

NumLib/TimeStepping/Algorithms/FixedTimeStepping.cpp
chleh: I'd leave it as `<=`, because here we use `t_curr += repeat * delta_t`, but the time loop later on will use `t_curr += delta_t` repeatedly. Due to rounding errors those two methods might get to different results. Maybe one should even add a small epsilon in order to make sure that the generated timestep list always covers the entire simulation time.

NumLib/TimeStepping/Algorithms/FixedTimeStepping.cpp
chleh: Related to the note above: Changed the condition to `t_curr <= t_end`.

Applications/ApplicationsLib/UncoupledProcessesTimeLoop.h
norihiro-w: ok

NumLib/TimeStepping/Algorithms/FixedTimeStepping.cpp
norihiro-w: i mean, e.g. if a user accidentally sets `t_initial` is same as `t_end` in a config.

NumLib/TimeStepping/Algorithms/FixedTimeStepping.cpp
norihiro-w: Can `FixedTimeStepping::next()`change its behaviour after this function? If it doesn't change, I'd handle the case here and complete all required time step sizes within this function.

chleh:
Rebased and squashed, ready for review, I think.

NumLib/TimeStepping/Algorithms/FixedTimeStepping.h
```diff
@@ -3,7 +3,7 @@
  * \date   2012-08-03
  *
  * \copyright
- * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ * Copyright (c) 2012-2015, OpenGeoSys Community (http://www.opengeosys.org)
```
endJunction: 16

norihiro-w:
:+1:

NumLib/TimeStepping/Algorithms/FixedTimeStepping.h
```diff
@@ -59,32 +61,26 @@ class FixedTimeStepping : public ITimeStepAlgorithm
...
-    /**
-     * @brief Create timestepper from the given configuration
-     *
-     * Currently this function only covers uniform timestep size.
-     */
-    static FixedTimeStepping* newInstance(BaseLib::ConfigTree const& config);
-
-    virtual ~FixedTimeStepping() {}
+    /// Create timestepper from the given configuration
+    static std::unique_ptr<ITimeStepAlgorithm> newInstance(BaseLib::ConfigTree const& config);
```
endJunction: Just a note not related to this PR. Sometimes there is a standalone method for a class `A`  `createA(ConfigTree, ...)`, sometimes there is a static member `A::newInstance(ConfigTree, ...)`. Having both "patterns" is potentially confusing.

endJunction:
The previous version of this branch was successfully used in M-process implemenatation.

Fix the copyright year and :fast_forward:

PR #1109, opened 2016-03-18
Ode solver with sundials third party library.

endJunction:
This is an adaptation and cleanup of Christophs ode solver interface.

Todo (unordered list):

 - [x] _(non-monadic)_ check error
 - [x] Fix memleaks
 - [x] Test suitability for TES
 - [x] Documentation
 - [x] Changlog
 - [x] Check CMake for sundials.
 - [x] Rework getYDot


Tests/MathLib/TestCVode.cpp
```diff
@@ -121,7 +121,7 @@ TEST(MathLibCVodeTest, ZeoliteAdsorption)
 
 	ode_solver->setFunction(f_zeolite, nullptr);
 
-	ode_solver->setIC(0.0, {pV, C});
+	ode_solver->setIC(0.0, {{pV, C}});
```
chleh: Only for my understanding: That's due to some c++11 std::array constructor implementation "bug", right?

MathLib/ODE/OdeSolverFactory.h
```diff
@@ -148,7 +145,7 @@ class ConcreteOdeSolver final
 	void init() override
 	{
 		Implementation::init(NumEquations);
-		Implementation::setFunction(&_handles);
+		Implementation::setFunction(_handles.get());
```
chleh: Wouldn't `*_handles` be even better? Or do we need to allow nullptr? Sorry, I cannot remember this detail.

Tests/MathLib/TestCVode.cpp
```diff
@@ -116,11 +116,10 @@ TEST(MathLibCVodeTest, ZeoliteAdsorption)
 	config.put("linear_multistep_method", "BDF");
 	auto ode_solver = MathLib::createOdeSolver<2>(config);
 
+	ode_solver->setFunction(f_zeolite, nullptr);
```
chleh: Why the reordering? Maybe there will be more changes later on. Let's talk about it.

MathLib/ODE/declarations.h
```diff
@@ -15,19 +15,30 @@
 namespace MathLib
 {
 
+template <int M, int N>
+using MappedMatrix = Eigen::Map<Eigen::Matrix<double, M, N>>;
```
chleh: Nice!

MathLib/ODE/CVodeSolver.cpp
```diff
@@ -278,15 +278,13 @@ void CVodeSolverImpl::solve(const double t_end)
...
+double* CVodeSolverImpl::getYDot(const double t, double const* const y)
 {
-	if (_f != nullptr)
-	{
-		return _f->call(t, y, ydot);
-	}
+	if (!_f) return nullptr;
 
-	return false;
+	double* ydot = nullptr;
```
chleh: That doesn't segfault?

MathLib/ODE/OdeSolverFactory.h
```diff
@@ -162,24 +161,24 @@ class ConcreteOdeSolver final
 		Implementation::setTolerance(abstol, reltol);
 	}
 
-	void setIC(const double t0, const Arr& y0) override
+	void setIC(const double t0,
+	           const MappedConstVector<NumEquations>& y0) override
```
chleh: Only curious: Why do you replace `Arr` with `MappedConstVector` and not with some `Eigen::Matrix<double,N,1>`?

Tests/MathLib/TestCVode.cpp
```diff
@@ -159,7 +159,7 @@ TEST(MathLibCVodeTest, Exponential)
 	ode_solver->setFunction(f, nullptr);
 	ode_solver->setTolerance(1e-8, 1e-6);
 
-	ode_solver->setIC(t0, {{y0}});
+	ode_solver->setIC(t0, {&y0});
```
chleh: What's that?

MathLib/ODE/CVodeSolver.h
```diff
@@ -12,7 +12,7 @@
 
 #include "BaseLib/ConfigTree.h"
 
-#include "declarations.h"
+#include "OdeSolver-fwd.h"
```
chleh: I think the new name is a bit misleading: You don't fwd-declare `ODESolver`.
When compared to `MathLib/ODESolver/Types.h`, I would use `Types.h` also here.

MathLib/ODE/CVodeSolver.cpp
```diff
@@ -303,59 +303,59 @@ CVodeSolverImpl::~CVodeSolverImpl()
 	}
 }
 
-CVodeSolverInternal::CVodeSolverInternal(BaseLib::ConfigTree const& config,
+CVodeSolver::CVodeSolver(BaseLib::ConfigTree const& config,
                                          unsigned const num_equations)
```
chleh: Check indentation.

Tests/MathLib/TestCVode.cpp
endJunction: Can be. I do remember something about std::array. On the other hand std::array is just struct, so it could be related to struct's initialization.

MathLib/ODE/OdeSolverFactory.h
endJunction: This is resolved by passing a unique_ptr now.

Tests/MathLib/TestCVode.cpp
endJunction: The reordering happened historically when `init` and `setFunction` were merged.

Tests/MathLib/TestCVode.cpp
endJunction: It would be nice if all of the y's would be in the same domain (like Eigen::Matrix). On the other hand creating of the Eigen::Matrices is not possible with initializer lists, if I'm not mistaken.
Switching back to std::array allows usage of initializer lists, but is not consistent with the space where the y's do live.

MathLib/ODE/CVodeSolver.cpp
```diff
@@ -171,6 +171,18 @@ CVodeSolverImpl::CVodeSolverImpl(const BaseLib::ConfigTree& config,
...
+
+	auto f_wrapped = [](const realtype t, const N_Vector y, N_Vector ydot,
+	                    void* function_handles) -> int
+	{
+		bool successful = static_cast<FunctionHandles*>(function_handles)
+		                      ->call(t, NV_DATA_S(y), NV_DATA_S(ydot));
+		return successful ? 0 : 1;
+	};
+
+	// TODO: check not run twice! move this call somewhere else
```
chleh: Now the comment is not needed anymore.

chleh:
@endJunction, I checked your branch yesterday evening at commit 77c2de343c5f855696581a8ade81368ef59be000.
You didn't do any major API changes, so this branch—as expected—still suits my needs.

I only have two simple suggestions:
* Please add overloads to `setIC()` and `getYDot()` which take `std::array<double, N> const&`, because then one can use those functions like `ode_solver.setIC(1.0, { 2.0, 3*x, cos(y) })`.
* I don't know if it is somewhere documented in our code that sundials expects column-major matrix storage order. At least it should be explicitly documented. Furthermore, it would probably be good if you included the storage order parameter `Eigen::ColMajor` explicitly into the `MappedMatrix` and `MappedConstMatrix` typedefs.

bilke:
@endJunction Feel free to `git cherry-pick` bilke/ogs@2d6221bc7a5

MathLib/CMakeLists.txt
```diff
@@ -63,12 +63,14 @@ add_library(MathLib STATIC ${SOURCES})
...
-target_link_libraries(MathLib INTERFACE
-	GeoLib
-	logog
-	sundials_cvode
-	sundials_nvecserial
-)
+if (CVODE_FOUND)
+	target_link_libraries(MathLib INTERFACE
+		GeoLib
+		logog
```
endJunction: Still needs GeoLib and logog

MathLib/ODE/ConcreteOdeSolver.h
```diff
@@ -121,6 +124,7 @@ std::unique_ptr<OdeSolver<NumEquations, FunctionArguments...>> createOdeSolver(
 	    new ConcreteOdeSolver<NumEquations, CVodeSolver, FunctionArguments...>(
```
endJunction: return nullptr by default.

Tests/MathLib/TestCVode.cpp
```diff
@@ -0,0 +1,337 @@
...
+
+	ode_solver->preSolve();
+
+	const double dt = 1e-1;
+
+	for (unsigned i = 1; i <= 10; ++i)
+	{
+		const double time = dt * i;
+
+		ode_solver->solve(time);
```
endJunction: check inside `solve` if time is reached, return error code. Maybe covered by sundials `check_error`

Tests/MathLib/TestCVode.cpp
```diff
@@ -0,0 +1,337 @@
...
+		auto const y = ode_solver->getSolution();
+		double time_reached = ode_solver->getTime();
+		auto const y_dot = ode_solver->getYDot(time_reached, y);
+
+		std::printf(
+		    "t: %14.7g, y: %14.7g, diff: %14.7g, y_dot: %14.7g, diff: %14.7g\n",
+		    time_reached, y[0], y[0] - exp(-15.0 * time_reached), y_dot[0],
+		    y_dot[0] + 15.0 * exp(-15.0 * time_reached));
+
+		ASSERT_NEAR(time, time_reached, std::numeric_limits<double>::epsilon());
```
endJunction: ``EXPECT...``
also check if solution is right.

MathLib/ODE/OdeSolverTypes.h
```diff
@@ -0,0 +1,67 @@
...
+
+template <int N>
+using MappedConstVector = MappedConstMatrix<N, 1>;
+
+// y read only
+// ydot is written in the function to be called.
+template <unsigned N, typename... FunctionArguments>
+using Function = bool (*)(const double t,
+                          MappedConstVector<N> const y,
+                          MappedVector<N> ydot,
```
endJunction: Why not pass by reference?

MathLib/ODE/ConcreteOdeSolver.h
```diff
@@ -0,0 +1,131 @@
...
+	    BaseLib::ConfigTree const& config);
+};
+
+#ifdef CVODE_FOUND
+template <unsigned NumEquations, typename... FunctionArguments>
+std::unique_ptr<OdeSolver<NumEquations, FunctionArguments...>> createOdeSolver(
+    BaseLib::ConfigTree const& config)
+{
+	return std::unique_ptr<OdeSolver<NumEquations, FunctionArguments...>>(
+	    new ConcreteOdeSolver<NumEquations, CVodeSolver, FunctionArguments...>(
```
endJunction: template param order could be same

MathLib/ODE/Handles.h
```diff
@@ -0,0 +1,104 @@
...
+namespace detail
+{
+
+/// Function handles for N equations and arbitrary arguments.
+template <unsigned N, typename... FunctionArguments>
+struct Handles;
+
+/// Function handles for N equations and single argument.
+template <unsigned N, typename FunctionArgument>
+struct Handles<N, FunctionArgument> : public MathLib::FunctionHandles
```
endJunction: Rename to FunctionHandlesImpl.

MathLib/ODE/Handles.h
```diff
@@ -0,0 +1,104 @@
...
+			          MappedVector<N>{ydot},
+			          MappedMatrix<N, N>{jac},
+			          _data);
+		return true;
+	}
+
+	bool hasJacobian() const override { return df != nullptr; }
+	unsigned getNumEquations() const override { return N; }
+private:
+	Function f = nullptr;
```
endJunction: make Function a std::function?

MathLib/ODE/ConcreteOdeSolver.h
```diff
@@ -0,0 +1,131 @@
...
+		Eigen::Matrix<double, NumEquations, 1, Eigen::ColMajor> y_dot;
+		Implementation::getYDot(t, y.data(), y_dot.data());
+		return y_dot;
+	}
+
+private:
+	/// instances of this class shall only be constructed by
+	/// the friend function listed below
+	ConcreteOdeSolver(BaseLib::ConfigTree const& config)
+	    : Implementation{config, NumEquations}
```
endJunction: = default

MathLib/ODE/CVodeSolver.cpp
```diff
@@ -0,0 +1,353 @@
...
+	}
+
+	_t = t0;
+}
+
+void CVodeSolverImpl::preSolve()
+{
+	assert(_f != nullptr && "ode function handle was not provided");
+
+	check_error("CVodeReInit", CVodeReInit, _cvode_mem, _t, _y);
```
endJunction: comment "setIC"

endJunction: maybe change ``check_error`` signature.

MathLib/ODE/CVodeSolver.cpp
```diff
@@ -0,0 +1,353 @@
...
+			std::abort();
+		}
+	}
+
+	_y = N_VNew_Serial(num_equations);
+	_abstol = N_VNew_Serial(num_equations);
+	_num_equations = num_equations;
+
+	_cvode_mem =
+	    CVodeCreate(_linear_multistep_method, _nonlinear_solver_iteration);
```
endJunction: check_error

wenqing:
@chleh Using pimp idiom is a great idea for this case. :fast_forward:

MathLib/CMakeLists.txt
```diff
@@ -60,25 +63,29 @@ add_library(MathLib STATIC ${SOURCES})
 
 set_target_properties(MathLib PROPERTIES LINKER_LANGUAGE CXX)
 
-target_link_libraries(MathLib INTERFACE
-	GeoLib
-	logog
-)
+if (CVODE_FOUND)
```
norihiro-w: suggest not always include found libraries

MathLib/ODE/CVodeSolver.cpp
```diff
@@ -0,0 +1,353 @@
...
+
+extern "C" {
+#include <cvode/cvode.h>             /* prototypes for CVODE fcts., consts. */
+#include <nvector/nvector_serial.h>  /* serial N_Vector types, fcts., macros */
+#include <cvode/cvode_dense.h>       /* prototype for CVDense */
+#include <sundials/sundials_dense.h> /* definitions DlsMat DENSE_ELEM */
+#include <sundials/sundials_types.h> /* definition of type realtype */
+}
+
+#include "CVodeSolver.h"
```
norihiro-w: is there any reason you cannot include this before extern?

MathLib/ODE/OdeSolverTypes.h
```diff
@@ -0,0 +1,67 @@
...
+
+#ifndef MATHLIB_ODE_ODESOLVERTYPES_H
+#define MATHLIB_ODE_ODESOLVERTYPES_H
+
+#include <Eigen/Core>
+
+namespace MathLib
+{
+template <int M, int N>
+using MappedMatrix = Eigen::Map<Eigen::Matrix<double, M, N, Eigen::ColMajor>>;
```
norihiro-w: please rename to something like MappedColMajorMatrix, because you define it as general type in MathLib. same applied to below types

norihiro-w:
I'm not asking to change it now, but have a general comment that I don't get a point of using template for this because sundials library takes raw pointers and function pointers. so it doesn't help usability and optimization. maybe i miss some points you intended. to be discussed in future

endJunction:
I'll keep a long answer for discussion. Some of the templates ensure correct array size at compile time, some are needed for correct casts when going back and forth between `double`-/function-pointers and `Eigen::Map`/function objects.

chleh:
@norihiro-w Only some remarks.
In
```
template <int M, int N>
using MappedMatrix = ...
```
we can get rid of the template params when we use a dynamic-size eigen matrix here. Same applies for the `OdeSolver` class.
The variadic template param `FunctionArgs...` can be avoided if we use `std::function<>` internally rather than raw function pointers.
The templated base class `Implementation` of `ConcreteOdeSolver` is there to slightly reduce code duplication if we later on use solvers from additional 3rd party libs.

norihiro-w:
thanks both for the quick response. let's come
back to this after Christoph s branch is merged

MathLib/ODE/OdeSolverTypes.h
chleh: Indeed, it's a general type. But it's only visible if you include the particular header (Hopefully it won't pollute all of our program.). I'd prefer to keep the shorter name, because in the ODE solver setting most users will probably not care if the matrix is row or column major; usually they'll just write to it using `operator()` or assignment to the whole matrix.
Additionally, I think we also do not indicate the storage order in the names of our other matrix types, neither for dense nor for sparse ones.

chleh:
@wenqing, thank you.

MathLib/ODE/OdeSolverTypes.h
norihiro-w: - it's visible also when someone tries to define MappedMatrix at other place in MathLib. It's not easy for others to know that general type is already reserved because it's given as a part of specific functionality.
- i asked about column major because all shape matrices are/must be row major. if we use both ordering types, we should rename them to indicate the type. Sparse matrices are ordering type independent.

norihiro-w: forgot to say. this issue is not urgent and can be fixed later if you like.

chleh: Valid points. I suggest to talk about it vis-a-vis. Maybe on monday, maybe later.

chleh:
@endJunction, FYI: I'm going to work on this PR now.

chleh:
@endJunction g++ fails with ambiguous function call at setIC(). Does it compile fine for you? Strange. Gonna resolve it.

MathLib/CMakeLists.txt
chleh: done.

MathLib/ODE/CVodeSolver.cpp
chleh: Fixed the order. Btw: extern is not needed, already done by sundials headers.

MathLib/ODE/ConcreteOdeSolver.h
chleh: cannot be defaulted: there is an init list.

MathLib/ODE/CVodeSolver.cpp
chleh: cannot be checked, because a pointer is returned. However, I added an error message.

Tests/MathLib/TestCVode.cpp
chleh: Covered by sundials.

MathLib/ODE/OdeSolverTypes.h
chleh:
```
	DBUG("sizes: map %lu, double* %lu, int %lu",
		 sizeof(Eigen::Map<Eigen::Matrix<double, 20, 20, Eigen::ColMajor>>),
		 sizeof(double*), sizeof(int));
```
says Eigen::Map is 16 bytes, i.e. probably one double* and two ints or so. Anyway, I'll change to pass by reference.

chleh:
@norihiro-w's comments are still open.

Tests/MathLib/TestODESolver.cpp
```diff
@@ -0,0 +1,291 @@
...
+#endif  // CVODE_FOUND
+}
+
+template<unsigned NumEquations>
+std::unique_ptr<MathLib::ODESolver<NumEquations>>
+make_ode_solver(boost::property_tree::ptree const& conf)
+{
+    // Make sure testrunner does not crash if we haven't built with support for
+    // any external ODE solver lib.
+    if (!any_ode_solver_libs_available()) return nullptr;
```
chleh: TODO add error message.

MathLib/ODE/OdeSolverTypes.h
chleh: After discussion with @norihiro-w I'll put everything in MathLib/ODESolver to the separate namespace `MathLib::ODESolver`.

chleh:
This PR looks stalled again. I talked to @norihiro-w last week. He said, he does not request any changes to this PR. They can be done later. Does anybody else want to make some comments or request some changes? Or may I squash and merge this PR now?

MathLib/CMakeLists.txt
```diff
@@ -60,25 +63,34 @@ add_library(MathLib STATIC ${SOURCES})
...
+	target_link_libraries(MathLib INTERFACE
+		GeoLib
+		logog
+		sundials_cvode
+		sundials_nvecserial
+	)
+else()
```
endJunction: - Skip the else branch and add sundials libraries as for example lapack libraries after the default target_link_libs(GeoLib logog)
 - If the sundials libraries' names are available in a variable like SUNDIALS_LIBRARIES, it would be nicer to use that variable instead of naming them explicitly

MathLib/ODE/ConcreteODESolver.h
```diff
@@ -0,0 +1,125 @@
...
+                new detail::FunctionHandlesImpl<NumEquations>{f, df}});
+    }
+
+    void setTolerance(const std::array<double, NumEquations>& abstol,
+                      const double reltol) override
+    {
+        Implementation::setTolerance(abstol.data(), reltol);
+    }
+
+    void setTolerance(const double abstol, const double reltol) override
```
endJunction: Could you please add the difference btw. the two `setTolerance` interfaces to documentation?

MathLib/ODE/ConcreteODESolver.h
```diff
@@ -0,0 +1,125 @@
...
+    Eigen::Matrix<double, NumEquations, 1, Eigen::ColMajor> getYDot(
+        const double t, const MappedConstVector<NumEquations>& y) const override
+    {
+        Eigen::Matrix<double, NumEquations, 1, Eigen::ColMajor> y_dot;
+        Implementation::getYDot(t, y.data(), y_dot.data());
+        return y_dot;
+    }
+
+private:
+    //! Instances of this class shall only be constructed by createODESolver().
```
endJunction: shall -> should? or even "must"?

MathLib/ODE/FunctionHandles.h
```diff
@@ -0,0 +1,115 @@
...
+            MappedVector<N> ydot_mapped{ydot};
+            return f(t, MappedConstVector<N>{y}, ydot_mapped);
+        }
+        return false;
+    }
+
+    /*! Calls the stored function computing
+     *  \f$\mathtt{jac} := \partial \dot y/\partial y\f$.
+     *
+     * \returns true or false indicating whethere the function succeeded.
```
endJunction: typo: whether

endJunction:
Nice documentation!
Please fix the CMakeFile, and if you like the minor documentation comments and squash/rebase.

MathLib/ODE/ConcreteODESolver.h
chleh: I think "shall" is exactly the right English word: "Instanzen dieser Klasse _sollen_ nur von `createODESolver()` erzeugt werden."

MathLib/CMakeLists.txt
chleh: It would be nice if you could fix the CMakeLists.txt, since I'm pretty short on time, and you already know what's to be done.

endJunction:
Jenkins, test this please!

PR #1111, opened 2016-03-18
Make geometry search strict

chleh:
Until now, if a geometry, e.g., for a BC, is not found, a debug (!) message is printed and a `nullptr` is returned from the respective method. If I remember correctly, this `nullptr` effectively causes that a Dirichlet BC is set not only on the boundary but on the entire domain (at least I observed such a behaviour some time ago).
The search for geometries is one (maybe the only?) weak point in OGS6's project file parsing.
This PR changes the behaviour in such a way that an __error__ message is printed if a geometry is not found and the program being aborted afterwards. That's consistent with what we do for all other project file errors.
I am aware of the fact that this PR can cause serious trouble for the DataExplorer: Nobody wants it to crash on wrong input. I just want to point to the problem. My solution is simple and effective (too effective for the DE). If you find a better solution than mine, I would appreciate that.

If you want to take over this PR, please do so.

For the time being, here's my PR.

ToDo:
* [ ] fix input files of present end-to-end tests

### DIFF
--- a/GeoLib/GEOObjects.cpp
+++ b/GeoLib/GEOObjects.cpp
@@ -63,8 +63,8 @@ const std::vector<Point*>* GEOObjects::getPointVec(const std::string &name) cons
 	if (idx != std::numeric_limits<std::size_t>::max())
 		return _pnt_vecs[idx]->getVector();
 
-	DBUG("GEOObjects::getPointVec() - No entry found with name \"%s\".", name.c_str());
-	return nullptr;
+	ERR("GEOObjects::getPointVec() - No entry found with name \"%s\".", name.c_str());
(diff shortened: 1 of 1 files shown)


rinkk:
This will definitely break the GUI.
The Data Model doesn't know what components of a geometry (pnts, plys, sfcs) exist and need to test for those. Also, you can add/remove component from a geometry, merge multiple geometries and so other things that will result in ```abort()``` with these changes.

In general, I am of the opinion that an auxiliary library such as GeoLib should not be able to decide if the application should crash.

chleh:
The presented solution only masks the descibed error. Closing this now and going to come up with a different solution.

PR #1112, opened 2016-03-21
Static string references for MshModel

rinkk:
This saves ~10% memory when loading a mesh.


TomFischer:
Nice improvement! :+1:

bilke:
:+1:

PR #1114, opened 2016-03-22
Changelog for preparation of version 6.0.5

TomFischer:
as titled


chleh:
@TomFischer, just made a PR to your's with my changes.
The following points should also go into the rease notes:

 - New configuration tree parser. Checks configuration parameters more strictly, automatically prints error/warning messages.
 - added an ODE solver library that can solve transient and nonlinear processes
   (see http://doxygen.opengeosys.org/df/d35/group__ODESolver.html)

PR #1115, opened 2016-03-22
Fix color and import leaks

rinkk:
95% of this PR is just swapping pointers to ```Color```s for actual values or references.

### DIFF
--- a/Applications/DataExplorer/Base/Color.cpp
+++ b/Applications/DataExplorer/Base/Color.cpp
@@ -25,7 +25,7 @@ namespace GeoLib {
 
 Color getRandomColor()
 {
-	std::array<unsigned char, 3> col;
+	Color col;
 	col[0] = static_cast<unsigned char>((rand()%5)*50);
 	col[1] = static_cast<unsigned char>((rand()%5)*50);
 	col[2] = static_cast<unsigned char>((rand()%5)*50);
@@ -64,7 +64,7 @@ int readColorLookupTable(std::map<std::string, Color> &colors, const std::string
 	return 1;
 }
 
-Color const& getColor(const std::string &id, std::map<std::string, Color> &colors)
+Color const getColor(const std::string &id, std::map<std::string, Color> &colors)
 {
 	for (std::map<std::string, Color>::const_iterator it=colors.begin(); it !=colors.end(); ++it)
 	{
@@ -75,13 +75,6 @@ Color const& getColor(const std::string &id, std::map<std::string, Color> &color
 	Color c = getRandomColor();
 	colors.insert(std::pair<std::string, Color>(id, c));
 	return c;
-}
-
-Color const& getColor(double id, std::map<std::string, GeoLib::Color> &colors)
-{
-	std::ostringstream stream;
-	stream << id;
-	return getColor(stream.str(), colors);
 }
 
 }
--- a/Applications/DataExplorer/Base/Color.cpp
+++ b/Applications/DataExplorer/Base/Color.cpp
@@ -23,18 +23,17 @@
 
 namespace GeoLib {
 
-Color* getRandomColor()
+Color getRandomColor()
 {
 	std::array<unsigned char, 3> col;
 	col[0] = static_cast<unsigned char>((rand()%5)*50);
 	col[1] = static_cast<unsigned char>((rand()%5)*50);
 	col[2] = static_cast<unsigned char>((rand()%5)*50);
-	return new Color(col);
+	return col;
 }
 
-int readColorLookupTable(std::map<std::string, Color*> &colors, const std::string &filename)
+int readColorLookupTable(std::map<std::string, Color> &colors, const std::string &filename)
 {
-
 	std::ifstream in( filename.c_str() );
 
 	if (!in)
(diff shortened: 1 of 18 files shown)


Applications/DataExplorer/Base/Color.cpp
```diff
@@ -50,35 +49,35 @@ int readColorLookupTable(std::map<std::string, Color*> &colors, const std::strin
...
 	WARN("Key \"%s\" not found in color lookup table.", id.c_str());
-	Color* c = getRandomColor();
-	colors.insert(std::pair<std::string, Color*>(id, c));
+	Color c = getRandomColor();
+	colors.insert(std::pair<std::string, Color>(id, c));
 	return c;
 }
 
-const Color* getColor(double id, std::map<std::string, GeoLib::Color*> &colors)
+Color const& getColor(double id, std::map<std::string, GeoLib::Color> &colors)
```
TomFischer: I know this is not the part of the PR, but just out of curiosity: Why is `id` a double? I would expect either an integral type or a std::string.

Applications/DataExplorer/Base/Color.cpp
```diff
@@ -50,35 +49,35 @@ int readColorLookupTable(std::map<std::string, Color*> &colors, const std::strin
...
 {
 	std::ostringstream stream;
 	stream << id;
```
TomFischer: Also not necessarily the topic of this PR: Is it possible to use `std::to_string(...)` instead of
```
std::ostringstream stream;
stream << id
```
?

TomFischer:
:+1:

Applications/DataExplorer/Base/Color.cpp
```diff
@@ -50,35 +49,35 @@ int readColorLookupTable(std::map<std::string, Color*> &colors, const std::strin
...
 	{
 		if (id.compare(it->first) == 0)
 			return it->second;
 	}
 	WARN("Key \"%s\" not found in color lookup table.", id.c_str());
-	Color* c = getRandomColor();
-	colors.insert(std::pair<std::string, Color*>(id, c));
+	Color c = getRandomColor();
+	colors.insert(std::pair<std::string, Color>(id, c));
 	return c;
```
TomFischer: A reference of a local variable is returned. This will be a problem later on!

Applications/DataExplorer/Base/Color.cpp
```diff
@@ -23,18 +23,17 @@
 
 namespace GeoLib {
 
-Color* getRandomColor()
+Color getRandomColor()
 {
 	std::array<unsigned char, 3> col;
```
norihiro-w: minor comment: Isn't it possible to write here `Color col;`? just to improve readability bit... At the beginning I thought Color has a constructor accepting std::array.

Applications/DataExplorer/DataView/DiagramView/DetailWindow.cpp
```diff
@@ -124,9 +124,8 @@ void DetailWindow::resizeWindow()
 
 void DetailWindow::addList(DiagramList* list)
 {
-	GeoLib::Color* c = GeoLib::getRandomColor();
-	QColor colour((*c)[0], (*c)[1], (*c)[2]);
-	delete c;
+	GeoLib::Color const c = GeoLib::getRandomColor();
```
norihiro-w: though I'm not sure if there is any difference, `GeoLib::Color const c(GeoLib::getRandomColor());` may be safer, i.e. avoiding copy.

Applications/DataExplorer/DataView/StationTreeView.cpp
```diff
@@ -237,38 +230,28 @@ void StationTreeView::showDiagramPrefsDialog()
...
 	std::vector<ModelTreeItem*> lists = static_cast<StationTreeModel*>(model())->getLists();
 	std::size_t nLists = lists.size();
 	for (std::size_t i = 0; i < nLists; i++)
-		if ( listName.toStdString().compare( lists[i]->data(0).toString().toStdString() )
-		     == 0 )
-		{
-			const std::vector<GeoLib::Point*>* stations =
-			        dynamic_cast<BaseItem*>(lists[i]->getItem())->getStations();
+	{
+		if ( listName.toStdString().compare( lists[i]->data(0).toString().toStdString() ) != 0 )
```
norihiro-w: isn't it possible to compare QString directly without converting to std::string?

norihiro-w:
:+1:  only minor comments you can take if you like

Applications/DataExplorer/Base/Color.cpp
rinkk: I have no answer to either of your questions. This code is very old and I have no idea what it was used for. Apparently it is not used anymore so I simply removed the method.

Applications/DataExplorer/DataView/DiagramView/DetailWindow.cpp
chleh: Above line does not copy. I cannot tell the respective line of the C++ standard, but I am very sure, that
```
GeoLib::Color const c = GeoLib::getRandomColor();
```
and
```
GeoLib::Color const c(GeoLib::getRandomColor());
```
are equivalent. Though, I think the variant with `=` is more readable (albeit the constructor call is somewhat obfuscated).

Applications/DataExplorer/DataView/StationTreeView.cpp
rinkk: Of course it's possible to compare QStrings directly. I have no idea what I was thinking 6 years ago :)

Applications/DataExplorer/Base/Color.cpp
rinkk: Fixed ... it *did* work, though. I was probably just lucky...

Applications/DataExplorer/DataView/DiagramView/DetailWindow.cpp
chleh: Here is something about it: http://en.cppreference.com/w/cpp/language/copy_initialization
> * If T is a class type and the type of other is cv-unqualified version of T or a class derived from T, __the converting constructors of T are examined and the best match is selected by overload resolution.__ The constructor is then called to initialize the object. 
[...]

> Copy-initialization is less permissive than direct-initialization: explicit constructors are not converting constructors and are not considered for copy-initialization.

norihiro-w: thanks for the explanation!

PR #1116, opened 2016-03-24
Tiny fix for wrong behaviour of tool ResetPropertiesInPolygonalRegion

TomFischer:
Resets properties inside instead outside.


endJunction:
:fast_forward:

Issue #1117, opened 2016-03-30
Tool `removeMeshElements` crashs.


TomFischer:
Under the condition that a mesh has both cell and node properties the method `Properties::excludeCopyProperties()` (used in removeMeshElements) can/will crash. 

The bug can be reproduced with the following mesh: [MeshReproducingBug.zip](https://github.com/ufz/ogs/files/194859/MeshReproducingBug.zip). The mesh contains the cell property "MarkedCells" and the node property "OriginalSubsurfaceIDs". Command line that will cause the crash is:
`removeMeshElements -i hex_3x3x2_sfc-marked_cells.vtu -o hex_3x3x2_sfc-marked_cells-removed.vtu -n MarkedCells --int-property-value 1`.

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/1117).

PR #1118, opened 2016-03-30
Fix Properties::excludeCopyProperties()

TomFischer:
This PR adds a second argument to the method `Properties::excludeCopyProperties()` such that both property types, cell and node properties can be correctly copied. Furthermore it was necessary to move some attributes / methods from `PropertyVector` to the abstract  base class `PropertyVectorBase`.

This fixes issue #1117 .

### DIFF
--- a/MeshLib/PropertyVector.h
+++ b/MeshLib/PropertyVector.h
@@ -31,6 +31,23 @@ public:
 		std::vector<std::size_t> const& exclude_positions
 	) const = 0;
 	virtual ~PropertyVectorBase() = default;
+
+	MeshItemType getMeshItemType() const { return _mesh_item_type; }
+	std::string const& getPropertyName() const { return _property_name; }
+	std::size_t getNumberOfComponents() const { return _n_components; }
+
+protected:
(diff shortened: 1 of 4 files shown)


endJunction:
:fast_forward:

MeshLib/PropertyVector.h
```diff
@@ -31,6 +31,23 @@ class PropertyVectorBase
 		std::vector<std::size_t> const& exclude_positions
 	) const = 0;
 	virtual ~PropertyVectorBase() = default;
+
+	MeshItemType getMeshItemType() const { return _mesh_item_type; }
+	std::string const& getPropertyName() const { return _property_name; }
+	std::size_t getNumberOfComponents() const { return _n_components; }
+
+protected:
+	explicit PropertyVectorBase(std::string const& property_name,
```
norihiro-w: no need to set explicit keyword, because the constructor requires three arguments

norihiro-w:
looks good. you can ignore the minor comment if you like

TomFischer:
Jenkins, test this please.

PR #1120, opened 2016-03-30
Parsing of a space delimited config values

endJunction:
Provides reading of list of values from a config file.
```xml
<values>0 1 2 5</values>
```
can now be read with
```c++
auto v = config.getConfParam<std::vector<int>>("values");
```

The idea behind this is to overload the implementation on `*T` parameter. (The idea is from [stackoverflow](http://stackoverflow.com/a/8315197).)

### DIFF
--- a/BaseLib/ConfigTree-impl.h
+++ b/BaseLib/ConfigTree-impl.h
@@ -81,6 +81,17 @@ boost::optional<std::vector<T>> ConfigTree::getConfParamOptionalImpl(
         T value;
         while (sstr >> value)
             result.push_back(value);
+        if (!sstr.eof())  // The stream is not read until the end, must be an
+                        // error. result contains number of read values.
+        {
+            error("Value for key <" + param + "> `" +
+                  shortString(sstr.str()) +
+                  "' not convertible to a vector of the desired type."
+                  " Could not convert token no. " +
+                  std::to_string(result.size() + 1) + ".");
+            return boost::none;
+        }
+
         return boost::make_optional(result);
     }
 
--- a/BaseLib/ConfigTree-impl.h
+++ b/BaseLib/ConfigTree-impl.h
@@ -9,6 +9,8 @@
 
 #include "ConfigTree.h"
 
+#include <sstream>
+
 namespace BaseLib
 {
 
@@ -50,15 +52,37 @@ getConfParam(std::string const& param, T const& default_value) const
     return default_value;
 }
 
-template<typename T>
-boost::optional<T>
-ConfigTree::
-getConfParamOptional(std::string const& param) const
+template <typename T>
+boost::optional<T> ConfigTree::getConfParamOptional(
+    std::string const& param) const
 {
     checkUnique(param);
 
+    return getConfParamOptionalImpl(param, static_cast<T*>(nullptr));
+}
+
+template <typename T>
+boost::optional<T> ConfigTree::getConfParamOptionalImpl(
+    std::string const& param, T*) const
+{
+    if (auto p = getConfSubtreeOptional(param)) return p->getValue<T>();
+
+    return boost::none;
+}
+
+template <typename T>
+boost::optional<std::vector<T>> ConfigTree::getConfParamOptionalImpl(
+    std::string const& param, std::vector<T>*) const
+{
     if (auto p = getConfSubtreeOptional(param))
-        return p->getValue<T>();
+    {
+        std::istringstream sstr{p->getValue<std::string>()};
+        std::vector<T> result;
+        T value;
+        while (sstr >> value)
+            result.push_back(value);
+        return boost::make_optional(result);
+    }
 
     return boost::none;
 }
--- a/BaseLib/ConfigTree.h
+++ b/BaseLib/ConfigTree.h
@@ -15,6 +15,7 @@
 
 #include <functional>
 #include <memory>
+#include <vector>
 
 #include <boost/property_tree/ptree.hpp>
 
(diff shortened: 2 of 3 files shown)


BaseLib/ConfigTree.h
```diff
@@ -498,6 +499,15 @@ class ConfigTree final
                           std::string const& message);
 
 private:
+    //! Default implementation of reading a value of type T.
+    template<typename T> boost::optional<T>
+    getConfParamOptionalImpl(std::string const& param, T*) const;
```
TomFischer: For what is the second parameter?

BaseLib/ConfigTree.h
```diff
@@ -498,6 +499,15 @@ class ConfigTree final
...
+
+    //! Implementation of reading a vector of values of type T.
+    template<typename T> boost::optional<std::vector<T>>
+    getConfParamOptionalImpl(std::string const& param, std::vector<T>*) const;
```
TomFischer: For what is the second parameter?

TomFischer:
Jenkins, test this please.

TomFischer:
:+1:

BaseLib/ConfigTree-impl.h
```diff
@@ -50,15 +52,37 @@ getConfParam(std::string const& param, T const& default_value) const
...
+boost::optional<T> ConfigTree::getConfParamOptional(
+    std::string const& param) const
 {
     checkUnique(param);
 
+    return getConfParamOptionalImpl(param, static_cast<T*>(nullptr));
+}
+
+template <typename T>
+boost::optional<T> ConfigTree::getConfParamOptionalImpl(
```
norihiro-w: can't we simply write `getConfParamOptionalImpl<T>(param);` by removing the second parameter?

endJunction:
The second parameter in `getConfParamOptionalImpl<T>()` is used to disambiguate between vector and non-vector variants. The correct type is triggered by the `static_cast<T*>(nullptr)` which is the second parameter. More specialized overload is considered first.

bilke:
Jenkins test this please.

norihiro-w:
i see- thanks!

endJunction:
There were compilation errors before because on eve the `icu` library changed from version 1.51 to 1.55 which is requiring an STL implementation with `regex_t`. The `icu` library is loaded by boost.

For now there is explicit unloading of the `icu-1.55` and loading of the previous version in the script. This might be reverted when we switch to a newer STL implementation (better than the gcc's 4.8.1).

Unclear remains the question why the compilation errors happend only for petsc builds, but as long as there is a working configuration we don't care too much.

chleh:
Nori is right: you don't need the second argument. If you introduce a 
second template param and put them in the right order, only the one that 
cannot be inferred from the argument list hast to be given. (When 
overthinking that again, maybe my comment is not valid, decide by 
yourself). However probably it would be much better to use two functions 
with distinct name, like: `...Single` and `...Vector`.
And shouldn't there be a variant returning std::array or a variant where 
one can pass the expected vector size to configtree?
Furthermore I think sstreem has very poor error checking. So you have to 
provide that manually, e.g., what happens when you want to make a 
vector<int> out of `0 0.1 true a`? There should probide a test case for 
that.

Best regards, christoph

Am 30.03.2016 22:26 schrieb Dmitry Yu. Naumov:

endJunction:
We shall discuss the technical details of the implementation; I'm not fully understanding how to remove the second parameter. Renaming of functions is also not possible, because current solution relies on overloading.
If the need arises we can introduce reading of fixed-size arrays.
sstream has simple behaviour and sufficient for now. Because it is implementation detail and is not changing the interface, improvements can be done later when needed.

BaseLib/ConfigTree-impl.h
chleh: I think @endJunction's solution is the best one. The problem is that you cannot have partial specializations if templated __methods__. So, e.g., if you declare a template method
```
    template<typename T>
    void a();
```
and a specialization
```
    template<typename T>
    void a<vector<T>>();
```
you will get a compilation error like
`non-type partial specialization »a<std::vector<T, std::allocator<_CharT> > >« is not allowed`.

BaseLib/ConfigTree-impl.h
```diff
@@ -50,15 +52,40 @@ getConfParam(std::string const& param, T const& default_value) const
...
+{
     if (auto p = getConfSubtreeOptional(param))
-        return p->getValue<T>();
+    {
+        std::stringstream ss{p->getValue<std::string>()};
+        std::vector<T> result;
+        T value;
+        while (ss >> value)
+            result.push_back(value);
+        if (!ss.eof()) // The stream is not read until the end, must be an error
```
chleh: I see you have some error checking here, I overlooked that last week. The problem with your solution is, that `boost::none` in the default case means "parameter has not been found"—cf. the implementation of `getConfParam<>()`—, i.e., it probably triggers confusing error messages.
Thus, you should print a proper error message here and abort rather than returning `none`. I think it would be good if that error message included the index of the value at which the error occured:
```
error("Value for key <" + param + "> `" + shortString(p->data())
+ "' not convertible to a vector of the desired type."
" Could not convert token no. " + std::to_string(result.size()+1) + ".");
```

endJunction:
@chleh Not sure if I got the error message right (is using _tree correct there?). If OK, this is rebased and checked.

chleh:
@endJunction using `_tree` is not correct. I made a PR to your branch. Now there will be error message in the failing tests:
```
error <> : Value for key <vector_bad1> `x 1 2a' not convertible to a vector of the desired type. Could not convert token no. 1.
error <> : Value for key <vector_bad2> `0 1 2a' not convertible to a vector of the desired type. Could not convert token no. 4.
```
The second line might be a bit confusing (cf. "token no. 4"), but I don't see any simple way to give a better message.

chleh:
:+1: after merging my PR.

PR #1123, opened 2016-04-04
Use non-default constructors for local assemblers

chleh:
The changes in this PR are the following:
* the `LocalDataInitializer` now supports local assemblers with non-default constructors
* therefore the `init`-method of the local assemblers can be dropped. Its functionality is merged into the constructor

This way, const correctness of the local assemblers can be improved. All data members can be initialized at construction time, which means that one probably needs less heap allocations (for more complicated processes) and that most data can be passed around as references at construction time.

### DIFF
--- a/AssemblerLib/LocalDataInitializer.h
+++ b/AssemblerLib/LocalDataInitializer.h
@@ -111,6 +111,20 @@ static_assert(false, "The macro OGS_MAX_ELEMENT_ORDER is undefined.");
 #endif
 
 
+// Generates a lambda that creates a new LocalAssembler of type LAData<SHAPE_FCT>
+#define OGS_NEW_LOCAL_ASSEMBLER(SHAPE_FCT) \
+    [](MeshLib::Element const& e, \
+       std::size_t const local_matrix_size, \
+       unsigned const integration_order, \
+       ConstructorArgs&&... args) \
+    { \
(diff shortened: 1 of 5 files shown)


TomFischer:
:+1:

AssemblerLib/LocalDataInitializer.h
```diff
@@ -146,19 +160,19 @@ class LocalDataInitializer
 #if (OGS_ENABLED_ELEMENTS & ENABLED_ELEMENT_TYPE_LINE) != 0 \
     && OGS_MAX_ELEMENT_DIM >= 0 && OGS_MAX_ELEMENT_ORDER >= 1
         _builder[std::type_index(typeid(MeshLib::Point))] =
-            [](){ return new LAData<NumLib::ShapePoint1>; };
+                OGS_NEW_LOCAL_ASSEMBLER(NumLib::ShapePoint1);
```
endJunction: Actually the lambda was used here, because it is so short. Now the function is a little longer and is moved out. But then there is no reason to use a macro; a template function will do as well.

chleh:
Wtf. I really just merged that PR? Don't know, how that happened. Sorry for that.

chleh:
@endJunction I'll incorporate your comment into #1125.

PR #1127, opened 2016-04-05
Pass time to ODESystem::getKnownSolutions();

endJunction:
For time-variable boundary conditions the current time is needed.

~~I'm not sure if this is the right solution, therefor it's for discussion.~~
The time is injected in the `TimeDiscretizedODESystem` now and the nonlinear solver is not touched.
The dirichlet bcs needs an update too, but this the point of this PR.

### DIFF
--- a/NumLib/ODESolver/ODESystem.h
+++ b/NumLib/ODESolver/ODESystem.h
@@ -65,8 +65,12 @@ public:
 
     using Index = typename MathLib::MatrixVectorTraits<Matrix>::Index;
 
-    virtual std::vector<ProcessLib::DirichletBc<Index> > const* getKnownSolutions() const
+    //! Provides known solutions (Dirichlet boundary conditions) vector for
+    //! the ode system at the given time \c t.
+    virtual std::vector<ProcessLib::DirichletBc<Index>> const*
+    getKnownSolutions(double const t) const
     {
+        (void)t;
         return nullptr; // by default there are no known solutions
     }
 };
--- a/ProcessLib/Process.h
+++ b/ProcessLib/Process.h
@@ -213,8 +213,8 @@ public:
 		//      VectoMatrixAssembler since that will work for all kinds of processes.
 	}
 
-	std::vector<DirichletBc<Index> > const* getKnownSolutions()
-	const override final
+	std::vector<DirichletBc<Index>> const* getKnownSolutions(
+	    double const /*t*/) const override final
 	{
 		return &_dirichlet_bcs;
 	}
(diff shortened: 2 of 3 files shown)


chleh:
I think in the nonlinear solver there should be no reference to time, because time is a concept completely unknown to a general nonlinear system.
Instead I suggest that time is being stored as a data member in the individual process (or even in the base class `Process`). From the individual process time can be passed as an additional argument to the Dirichlet/Neumann BC.
I also see the downside of that approach: It adds state to the process. Still I prefer that over extending some APIs in non-natural way.

endJunction:
OK. I understand the reasons. I'd ask @chleh then for possible solution w/o disturbing the existing concepts. Note, similar requirements are posed for pre/post/assemble functions, where the user might want to access t/dt/ts/... .

chleh:
Only one short comment:

> Note, similar requirements are posed for pre/post/assemble functions, where the user might want to access t/dt/ts/... .

I think t and dt should be passed to the `preTimestep()` method, which can then store those values for any further use by the processes.

NumLib/ODESolver/ODESystem.h
```diff
@@ -65,7 +65,8 @@ class ODESystem<Matrix, Vector,
 
     using Index = typename MathLib::MatrixVectorTraits<Matrix>::Index;
 
-    virtual std::vector<ProcessLib::DirichletBc<Index> > const* getKnownSolutions() const
+    virtual std::vector<ProcessLib::DirichletBc<Index>> const*
+    getKnownSolutions(double const t) const
```
chleh: t is not used, I guess there will be a compiler warning.

ProcessLib/Process.h
```diff
@@ -213,8 +213,8 @@ class Process
 		//      VectoMatrixAssembler since that will work for all kinds of processes.
 	}
 
-	std::vector<DirichletBc<Index> > const* getKnownSolutions()
-	const override final
+	std::vector<DirichletBc<Index>> const* getKnownSolutions(
+	    double const t) const override final
```
chleh: Same here: Compiler warning.

NumLib/ODESolver/ODESystem.h
```diff
@@ -65,7 +65,8 @@ class ODESystem<Matrix, Vector,
 
     using Index = typename MathLib::MatrixVectorTraits<Matrix>::Index;
 
-    virtual std::vector<ProcessLib::DirichletBc<Index> > const* getKnownSolutions() const
+    virtual std::vector<ProcessLib::DirichletBc<Index>> const*
```
chleh: Is it necessary to provide some doc? If so, could you please add a line of doxygen doc, please. I forgot that when writing this code.

chleh:
:+1: That's what we discussed after lunch.

endJunction:
Cleaned up warnings, added documentation, as suggested in the comments.

TomFischer:
:fast_forward:

PR #1130, opened 2016-04-07
Apply Dirichlet BCs in the Newton scheme

chleh:
The `applyKnownSolutions()` methods in the time loop do safe downcasting, similar to the `setEquationSystem()` methods that are already present.

The provided implementations of `applyKnownSolutions()` in NumLib are definitely not perfect, but they should work. Performance optimizations to be done later on.

This is a follow-up PR of #1127.

### DIFF
--- a/Applications/ApplicationsLib/UncoupledProcessesTimeLoop.h
+++ b/Applications/ApplicationsLib/UncoupledProcessesTimeLoop.h
@@ -197,6 +197,33 @@ private:
             break;
         }
     }
+
+    //! Applies known solutions to the solution vector \c x, transparently
+    //! for equation systems linearized with either the Picard or Newton method.
+    template<NumLib::NonlinearSolverTag NLTag>
+    static void applyKnownSolutions(
+            EquationSystem const& eq_sys, Vector& x)
+    {
+        using EqSys = NumLib::NonlinearSystem<Matrix, Vector, NLTag>;
+        assert(dynamic_cast<EqSys const*> (&eq_sys) != nullptr);
+        auto& eq_sys_ = static_cast<EqSys const&> (eq_sys);
+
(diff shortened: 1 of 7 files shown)


wenqing:
I think the function of applying Dirichlet BC for the Picard or the Newton-Raphson method should be the same. The only difference is the value of BC as
Picard:   u_b = u_{given}
Newton: u_b = u_{given}^{k+1}-u_{given}^{k},  u_b = 0 if the Dirichlet BC being independent of time.

norihiro-w:
and set zero to corresponding residuals if you haven't implemented it yet.

TomFischer:
It seems that @endJunction commit has changed a little bit. Maybe its possible to rebase to his branch?

wenqing:
After discussed with @chleh , I see his idea of the implementation: :+1:.

endJunction:
"Pass time to ODESystem::getKnownSolutions();"  PR is merged with some more comments, therefore the "merge conflict".

chleh:
@norihiro-w 

> and set zero to corresponding residuals if you haven't implemented it yet.

clarified in discussion.

Applications/ApplicationsLib/UncoupledProcessesTimeLoop.h
```diff
@@ -310,11 +337,16 @@ solveOneTimeStepOneProcess(
 
     setEquationSystem(nonlinear_solver, ode_sys, nl_tag);
 
-    process.preTimestep(x, t, delta_t);
+    // Note: Order matters!
+    // nextTimestep(); applyKnownSolutions(); preTimestep();
```
endJunction: Instead of repeating the code, I would suggest to write the reason, why the order is important.

endJunction:
Aside of possible comment improvement :fast_forward:

PR #1131, opened 2016-04-07
Use norm of delta x within Newton scheme

chleh:
The old approach of taking the residual norm was erroneous: It was computed too early (before applying Dirichlet BCs), thus it did not converge.

This PR changes that. And it switches to norm of delta x for now. That's then consistent with the way the Picard solver currently computes errors. Of course, later on there will be more error calculation methods to choose from.

### DIFF
--- a/NumLib/ODESolver/NonlinearSolver-impl.h
+++ b/NumLib/ODESolver/NonlinearSolver-impl.h
@@ -160,17 +160,11 @@ solve(Vector &x)
         sys.assembleResidualNewton(x);
         sys.getResidual(x, res);
 
-        // std::cout << "  res:\n" << res << std::endl;
-
-        // TODO streamline that, make consistent with Picard.
-        if (BLAS::norm2(res) < _tol) {
-            error_norms_met = true;
-            break;
-        }
-
         sys.assembleJacobian(x);
         sys.getJacobian(J);
         sys.applyKnownSolutionsNewton(J, res, minus_delta_x);
+
+        auto const error_res = BLAS::norm2(res);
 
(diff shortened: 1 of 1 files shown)


norihiro-w:
definitely we need a check based on residual norm (can be done later).

NumLib/ODESolver/NonlinearSolver-impl.h
```diff
@@ -160,18 +160,12 @@ solve(Vector &x)
...
-        if (BLAS::norm2(res) < _tol) {
-            error_norms_met = true;
-            break;
-        }
-
         sys.assembleJacobian(x);
         sys.getJacobian(J);
         sys.applyKnownSolutionsNewton(J, res, minus_delta_x);
 
+        auto const error_res = BLAS::norm2(res);
```
TomFischer: Maybe `norm2_res` is a better name?

TomFischer: Maybe `norm2_res` is a better name?

chleh: For now, maybe yes. But this entire line will move to some new ConvergenceCriterion class in some months. Thus, I'd like to keep the name: It's less work, and the old name doesn't hurt.

TomFischer: Okay, if the code will be moved later we can keep it for the moment.

:fast_forward:

endJunction:
Looks harmless ;)

chleh:
@endJunction That's what makes it dangerous :smile:

PR #1132, opened 2016-04-08
Set global vector entries for different types

chleh:
The provided functions allow setting vector entries using the same interface for
`Eigen::VectorXd`, `EigenVector` and `PETScVector`.

This is a precursor to #1130.

### DIFF
--- a/MathLib/LinAlg/UnifiedMatrixSetters.cpp
+++ b/MathLib/LinAlg/UnifiedMatrixSetters.cpp
@@ -67,6 +67,12 @@ void setVector(Eigen::VectorXd& v, std::initializer_list<double> values)
     for (std::size_t i=0; i<values.size(); ++i) v[i] = *(it++);
 }
 
+void setVector(Eigen::VectorXd& v, MatrixVectorTraits<Eigen::VectorXd>::Index const index,
+               double const value)
+{
+    v[index] = value;
+}
+
 } // namespace MathLib
 
 #endif // OGS_USE_EIGEN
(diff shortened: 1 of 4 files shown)


MathLib/LinAlg/UnifiedMatrixSetters.cpp
```diff
@@ -97,6 +103,12 @@ void setVector(PETScVector& v,
     v.set(idcs, vals);
 }
 
+void setVector(PETScVector& v, MatrixVectorTraits<PETScVector>::Index const index,
```
wenqing: Looks good, except that this function might be seldom used.

MathLib/LinAlg/UnifiedMatrixSetters.cpp
```diff
@@ -97,6 +103,12 @@ void setVector(PETScVector& v,
...
+               double const value)
+{
+    v.set(index, value);
```
chleh: TODO: handle negative indices specially; and look for negative indices in PR #1130.

MathLib/LinAlg/UnifiedMatrixSetters.cpp
chleh: Indeed, thie functions introduced won't be used in many places. Furthermore, for performance reasons there maybe should be setters that take vectors if indices and values and set them all at once (future work, I commented on that in #1130).

MathLib/LinAlg/UnifiedMatrixSetters.cpp
TomFischer: I am unsure. Do you want to handle negative indices in this PR or later on?

chleh:
Sorry guys, I accidentally already merged #1130. So this PR now is obsolete.
I think there are no critical points in here, except for the nagative index thing.
I'm going to open an issue for that.

MathLib/LinAlg/UnifiedMatrixSetters.cpp
chleh: "Treated" in bug report #1140.

PR #1133, opened 2016-04-08
Enable shared linking

endJunction:
~~Fixing metis linkage;~~ Remove STATIC from add_library().

With this changes and configuring the build for debug and with `-DBUILD_SHARED_LIBS=On` the compile + linking time for the ogs target (from scratch, no ccache) reduces (for that particular branch endJunction/M2 with dynamic eigen matrices) from 3:45 to 2:55.

The default build process should not change with this PR, because the static linking is chosen by default.

__Update:__ On another machine there is no change, unfortunately, which I don't understand yet.

### DIFF
--- a/MathLib/CMakeLists.txt
+++ b/MathLib/CMakeLists.txt
@@ -56,7 +56,7 @@ if(METIS_FOUND)
 endif()
 
 # Create the library
-add_library(MathLib STATIC ${SOURCES})
+add_library(MathLib ${SOURCES})
 
 set_target_properties(MathLib PROPERTIES LINKER_LANGUAGE CXX)
 
--- a/MathLib/CMakeLists.txt
+++ b/MathLib/CMakeLists.txt
@@ -56,12 +56,12 @@ if(METIS_FOUND)
 endif()
 
 # Create the library
-add_library(MathLib ${SOURCES})
+add_library(MathLib STATIC ${SOURCES})
 
 set_target_properties(MathLib PROPERTIES LINKER_LANGUAGE CXX)
 
 target_link_libraries(MathLib
-    logog
+    AssemblerLib
 )
 
 if(METIS_FOUND)
--- a/scripts/cmake/CompilerSetup.cmake
+++ b/scripts/cmake/CompilerSetup.cmake
@@ -22,6 +22,12 @@ if(NOT COMPILER_IS_MSVC)
 	set( CMAKE_INCLUDE_SYSTEM_FLAG_CXX "-isystem" CACHE STRING "" FORCE )
 endif()
 
+# When static libraries are used in some shared libraries it is required that
+# also the static libraries have position independent code.
+if(BUILD_SHARED_LIBS)
+    set(CMAKE_POSITION_INDEPENDENT_CODE TRUE)
+endif()
+
 # Set additional user-given compiler flags
 set(CMAKE_CXX_FLAGS ${OGS_CXX_FLAGS})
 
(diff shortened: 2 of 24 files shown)


chleh:
Having the additional option of shared linking is good, in general.
However, could you please give some information on:
* whether debugger starts faster/slower (just run ogs in the debugger w/o args)
* some details of the build that gave the build time improvement from 3:45 to 2:55: Was it a clean build? Did you use ccache? (Probably the answers to these two questions are yes and no, respectively.)

At least this PR seems to clean up the cmake scripts a bit, so :fast_forward:

norihiro-w:
I couldn't compile this branch on my Mac with clang. So far I found two reasons
- `Undefined symbols` : Setting `INTERFACE` is inappropriate for some libraries in `target_link_libraries`. PUBLIC or PRIVATE should be used. see http://stackoverflow.com/questions/26037954/cmake-target-link-libraries-interface-dependencies
- Circular dependency: e.g. MathLib and GeoLib. CMake detects it for building shared lib.

MathLib/CMakeLists.txt
```diff
@@ -65,6 +65,10 @@ target_link_libraries(MathLib INTERFACE
 	logog
 )
 
+if(METIS_FOUND)
+    target_link_libraries(MathLib INTERFACE metis)
```
norihiro-w: use ${METIS_LIBRARY}

endJunction:
@norihiro-w The undefined symbols and INTERFACE vs PRIVATE/PUBLIC is easily fixable. The circular dependencies are more difficult, I'm working.

I suggest to fix the first point, but leave the circular dependencies for further PRs.

MathLib/LinAlg/Solvers/BiCGStab.cpp
```diff
@@ -13,6 +13,8 @@
  */
 
 #include "BiCGStab.h"
+#include "../Sparse/CRSMatrix.h"
```
chleh: Do we allow `..` in an include path?

endJunction:
Before this PR is forgotten I propose following compromise: Static linking is removed for all but circularly dependent libraries until we are working on the resolution of those circular dependencies.

When a circular dependency is broken, the STATIC flag must be removed.

Further I propose to change one (or multiple) of the jenkins linux builds to dynamic linking, such that any new circular library dependencies can be recognized early on.

scripts/cmake/CompilerSetup.cmake
```diff
@@ -22,6 +22,12 @@ if(NOT COMPILER_IS_MSVC)
 	set( CMAKE_INCLUDE_SYSTEM_FLAG_CXX "-isystem" CACHE STRING "" FORCE )
 endif()
 
+# When static libraries are going to be used in other libraries position
```
chleh: > static libraries

I thought, this commit is about shared libs.

endJunction: "... in other libraries..." means the shared libraries which are dependent on the static libraries.
Comment improvements are welcome.

chleh:
```
When static libraries are used in some shared libraries it is required that
also the static libraries have position independent code.
```

endJunction:
I have added a Linux-PRs-dynamic test on jenkins building a debug config with shared libs. Currently testing, so if some of your PRs rise a red flag for Linux-PRs-dynamic ignore it for now.

endJunction:
Jenkins, test this please!

TomFischer:
Ready for merge?!

endJunction:
Jenkins, test this please!

endJunction:
Rebased on ufz/master. It's ready if the tests are all green. Maybe one more time "Jenkins ... this please."...

PR #1135, opened 2016-04-09
Added executeMemberDereferenced() + (minor) cleanup

chleh:
Changes:
1. `operator()()` in `VectorMatrixAssembler` renamed to `assemble()`.
That change makes the assembly routine an "ordinary" member, like all the other methods.
2. `LocalAssembler` now is a tpl param of the class `VectorMatrixAssembler`, not of its methods anymore (needed for the next point)
3. added `executeMemberDereferenced()` to `SerialExecutor`. Thereby one does not have to use lambdas when executing member methods other than `operator()()`, cf. https://github.com/ufz/ogs/compare/master...chleh:execute-overload?expand=1#diff-f2f37a7122ad44875cbf665ce72096deR153.
4. the `GlobalAssembler` now is owned by the concrete process (since it now depends on the specific `LocalAssembler`).

Point (3) makes `GlobaLSetup::execute...()` calls more readable in the specific process. I admit, that it does not help much (but maybe a little bit) to make the control flow easier to understand. But, this change hides the inner details better from process developers. I assume, that it makes the code look a little bit simpler to new developers (who don't care about inner details), since they now are not confronted with lambdas immediately.

### DIFF
--- a/AssemblerLib/SerialExecutor.h
+++ b/AssemblerLib/SerialExecutor.h
@@ -18,37 +18,28 @@ namespace AssemblerLib
 
 struct SerialExecutor
 {
-    /// Executes a \c f for each element from the input container.
+    /// Executes \c f for each element from the input container.
+    ///
+    /// The elements of \c c are dereferenced before being passed to \c f.
     /// Return values of the function call are ignored.
     ///
-    /// \tparam F   \c f type.
-    /// \tparam C   input container type.
+    /// \note
+    /// This function is intended to be used if \c f is a plain function,
+    /// a static class method, a lambda or an object having a call operator
+    /// defined. For non-static member functions you are encouraged to use
+    /// executeMemberDereferenced(), if applicable.
     ///
-    /// \param f    a function that accepts a pointer to container's elements and
-    ///             an index as arguments.
+    /// \tparam F    type of the callback function \c f.
+    /// \tparam C    input container type.
+    /// \tparam Args types additional arguments passed to \c f.
+    ///
+    /// \param f    a function that accepts a a reference to container's elements,
+    ///             an index as arguments (and possibly further arguments).
     /// \param c    a container supporting access over operator[].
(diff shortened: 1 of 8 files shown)


chleh:
Maybe you disagree with that approach, so let's discuss.

ProcessLib/Process.h
```diff
@@ -60,18 +60,16 @@ class Process
 	Process(MeshLib::Mesh& mesh,
 	        NonlinearSolver& nonlinear_solver,
 	        std::unique_ptr<TimeDiscretization>&& time_discretization)
-	    : _mesh(mesh), _nonlinear_solver(nonlinear_solver)
+	    : _nonlinear_solver(nonlinear_solver)
 	    , _time_discretization(std::move(time_discretization))
+		, _mesh(mesh)
```
chleh: indentation.

endJunction:
Jenkins, test this please!

AssemblerLib/SerialExecutor.h
```diff
@@ -62,6 +62,24 @@ struct SerialExecutor
...
 
+    /// Executes the given \c method of the given \c object for each element
+    /// from the input \c container.
+    ///
+    /// This method is very similar to executeDereferenced().
+    ///
+    /// \see executeDereferenced()
+    template <typename Container, typename Object, typename... MethodArgs, typename... Args>
+    static void
+    executeMemberDereferenced(
```
endJunction: please check, if `executeDereferenced` is still needed.

endJunction:
After answering the comments :fast_forward:

norihiro-w:
looks good

AssemblerLib/SerialExecutor.h
chleh: removed `execute()`. `executeDereferenced()` is still needed for, e.g., for lambdas or plain functions.

PR #1136, opened 2016-04-09
Process cleanup

chleh:
Changes
* some more private instead of protected members
* `GlobalSetup` now does not have any instances anymore.

follow-up of #1135

### DIFF
--- a/ProcessLib/ProcessVariable.h
+++ b/ProcessLib/ProcessVariable.h
@@ -75,10 +75,9 @@ public:
 		}
 	}
 
-	template <typename OutputIterator, typename GlobalSetup, typename... Args>
+	template <typename GlobalSetup, typename OutputIterator, typename... Args>
 	void createNeumannBcs(OutputIterator bcs,
 	                      MeshGeoToolsLib::BoundaryElementsSearcher& searcher,
-	                      GlobalSetup const&,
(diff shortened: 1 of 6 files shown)


chleh:
Jenkins, test this please.

endJunction:
Jenkins, test this please!

endJunction:
Looks good. :fast_forward: 

Note on word choice: disabling GlobalSetup ctor does not make it a singleton. Singletons are instantiated giving always at most one object, so they do have a constructor.

ProcessLib/ProcessVariable.h
```diff
@@ -75,10 +75,9 @@ class ProcessVariable
 		}
 	}
 
-	template <typename OutputIterator, typename GlobalSetup, typename... Args>
+	template <typename GlobalSetup, typename OutputIterator, typename... Args>
```
norihiro-w: do you still need to set GlobalSetup as a template parameter?

chleh: Not more and not less than before this PR. Within our current usage, we can easily provide GlobalSetup solely as global typedef. I think that would be a good thing, but it would would introduce changes in quite some places (I guess; easy changes, but at many places). So I'd like to leave that change for a future PR.

PR #1138, opened 2016-04-12
Break fileIO ApplicationsLib dependency

endJunction:
~~Follow up of ufz/ogs#1133.~~ This PR adds only the one commit [[DE] Break ApplicationLib dependency.](https://github.com/ufz/ogs/pull/1138/commits/7cf6b6068e9984949d35a41ce81d8558bfe7cf4d)

The dependency is caused by usage of ProjectData::addMesh by XmlGspInterface.
Unfortunately the programming of that class and its parents, the XMLInterface and XMLQtInterface is not as separate as it could be, therefore the current solution can be seen as a quickfix.

Clean solution would be, for example, a separation of the reader and the writer parts of the XMLInterface(s) and to use standalone functions for qt's xml validation.

### DIFF
--- a/FileIO/XmlIO/Qt/XmlGspInterface.cpp
+++ b/FileIO/XmlIO/Qt/XmlGspInterface.cpp
@@ -19,7 +19,6 @@
 
 #include <logog/include/logog.hpp>
 
-#include "Applications/ApplicationsLib/ProjectData.h"
 #include "GeoLib/GEOObjects.h"
 
 #include "XmlGmlInterface.h"
@@ -39,9 +38,15 @@
 
 namespace FileIO
 {
-XmlGspInterface::XmlGspInterface(ProjectData& project) :
-	XMLInterface(), XMLQtInterface(BaseLib::FileFinder().getPath("OpenGeoSysProject.xsd")),
-	_project(project)
+XmlGspInterface::XmlGspInterface(
+    GeoLib::GEOObjects& geoObjects,
+    std::vector<MeshLib::Mesh*> const& mesh_vector,
+    std::function<void(MeshLib::Mesh* const)>&& add_mesh_callback)
+    : XMLInterface(),
+      XMLQtInterface(BaseLib::FileFinder().getPath("OpenGeoSysProject.xsd")),
+      _geoObjects(geoObjects),
+      _mesh_vector(mesh_vector),
+      _add_mesh_callback(std::move(add_mesh_callback))
 {
 }
 
@@ -69,7 +74,7 @@ int XmlGspInterface::readFile(const QString &fileName)
 		const QString file_node(fileList.at(i).nodeName());
 		if (file_node.compare("geo") == 0)
 		{
-			XmlGmlInterface gml(*(_project.getGEOObjects()));
+			XmlGmlInterface gml(_geoObjects);
 			const QDomNodeList childList = fileList.at(i).childNodes();
 			for(int j = 0; j < childList.count(); j++)
 			{
@@ -86,7 +91,7 @@ int XmlGspInterface::readFile(const QString &fileName)
 		}
 		else if (file_node.compare("stn") == 0)
 		{
-			XmlStnInterface stn(*(_project.getGEOObjects()));
+			XmlStnInterface stn(_geoObjects);
 			const QDomNodeList childList = fileList.at(i).childNodes();
 			for(int j = 0; j < childList.count(); j++)
 				if (childList.at(j).nodeName().compare("file") == 0)
@@ -98,8 +103,7 @@ int XmlGspInterface::readFile(const QString &fileName)
 			const std::string msh_name = path.toStdString() +
 			                             fileList.at(i).toElement().text().toStdString();
 			MeshLib::Mesh* msh = FileIO::readMeshFromFile(msh_name);
-			if (msh)
-				_project.addMesh(msh);
+			if (msh) _add_mesh_callback(msh);
 		}
 	}
 
@@ -114,7 +118,6 @@ int XmlGspInterface::writeToFile(const std::string& filename)
 
 bool XmlGspInterface::write()
 {
-	GeoLib::GEOObjects* geoObjects = _project.getGEOObjects();
 	QFileInfo fi(QString::fromStdString(_filename));
 	std::string path((fi.absolutePath()).toStdString() + "/");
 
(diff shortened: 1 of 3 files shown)


FileIO/XmlIO/Qt/XmlGspInterface.cpp
```diff
@@ -39,9 +38,15 @@
 
 namespace FileIO
 {
-XmlGspInterface::XmlGspInterface(ProjectData& project) :
-	XMLInterface(), XMLQtInterface(BaseLib::FileFinder().getPath("OpenGeoSysProject.xsd")),
-	_project(project)
+XmlGspInterface::XmlGspInterface(
+    GeoLib::GEOObjects& geoObjects,
+    std::vector<MeshLib::Mesh*> const& mesh_vector,
+    std::function<void(MeshLib::Mesh* const)> add_mesh_callback)
```
chleh: move it: `&&`

chleh:
Last commit looks good. Just make it compile. :+1:

FileIO/XmlIO/Qt/XmlGspInterface.cpp
rinkk: This prevents the Data Explorer from loading anything except geometries and meshes from a project?!

rinkk:
As a general question: Why does the DataExplorer need to be decoupled from FileIO?

FileIO/XmlIO/Qt/XmlGspInterface.cpp
```diff
@@ -153,8 +156,9 @@ bool XmlGspInterface::write()
 	}
 
 	// MSH
-	const std::vector<MeshLib::Mesh*> msh_vec = _project.getMeshObjects();
-	for (std::vector<MeshLib::Mesh*>::const_iterator it(msh_vec.begin()); it != msh_vec.end(); ++it)
+	for (std::vector<MeshLib::Mesh*>::const_iterator it(_mesh_vector.begin());
```
TomFischer: Why not shorter and more readable: `for (auto const* mesh : _mesh_vector)`?

FileIO/XmlIO/Qt/XmlGspInterface.cpp
endJunction: No, it does not. It is only separating circular dependency of the FileIO and ApplicationsLib. The code for reading is not touched, so everything what was read should behave exactly the same. Only the place where the mesh is added to project has moved.

endJunction:
The data explorer is not decoupled from FileIO, and actually nothing is decoupled. The circular dependency of ApplicationLib and FileIO is removed.

FileIO/XmlIO/Qt/XmlGspInterface.cpp
rinkk: But if the project file reader has no knowledge of ProjectData but only of GeoObjects + the mesh vector it can only store the related information. I don't doubt that everything works fine at the moment but the next step after reorganising the project file reader would be reading/writing boundary conditions which would then require another parameter, I think? Maybe we could talk about this tomorrow or on friday? (maybe together with Tom as would probably be involved in the next stage)

chleh: @rinkk, I can understand your concerns. But if we keep all parts of `ProjectData` in here, soon `FileIO` will also depend on `ProcessLib` (for BCs). Probably that shouldn't be.
Maybe a different approach would be to move some of the parsing code to the DE directory. Do you think, that's feasible?
E.g., the following case checking code probably should be done in some DE class.
```
if (file_node.compare("geo") == 0)
```
Maybe at some point one could even use the constructor that is provided with `ProjectData` to read the `*.prj` (if we remove `std::abort()` :smile:).
That's at least what I can say after a discussion with @endJunction.

norihiro-w: maybe it's not good idea to mix core parts and application layers together. How about provide separate FileIO for ApplicationLib? Current FileIO can be used to provide read and write of basic data, e.g. geo and mesh.

endJunction: Agree with Nori. If planning a major cleanup of the FileIO/ApplicationsLib separation of the business logic and the actual reading of the data would be a good goal.

We shall have a 30min discussion tomorrow.

norihiro-w: or to make it consistent for all libs, we can create FileIo for each lib, e.g. geolib/fileio, meshlib/fileio

bilke:
Jenkins test this please.

rinkk:
Would it help to create a seperate data holder object for the Data Explorer in, e.g. BaseLib, and keep everything else as it is?  
What I need (and what ProjectData originally was) is a container for storing all data relevant to the currently processed project. OGS, and especially the Data Explorer has 20 or more input formats and I would like to keep these seperate from the application itself. In addition, some of said formats can hold geometry *or* meshes so that a clear dinstinction what library they'd belong to is not necessarily clear.

endJunction:
In the discussion we came to the following two steps:
 - This PR will be merged in order to be able to use shared libraries asap.
 - A "project data" structure---different from the current ProjectData---suitable for the data explorer and used by the FileIO will be introduced, thus breaking the dependency of the FileIO and the ApplicationsLib.

MathLib/CMakeLists.txt
```diff
@@ -56,25 +56,29 @@ if(METIS_FOUND)
...
 
-target_link_libraries(MathLib INTERFACE
-	GeoLib
-	logog
+target_link_libraries(MathLib
+    #AssemblerLib   # Needed, but causes circular dependency.
+    logog
 )
 
+if(METIS_FOUND)
```
norihiro-w: TOdo for later, not now: we shouldnt always link metis and lapack just because Cmake could find them. there should be a flag like OGS_USE_METIS.

norihiro-w:
i haven't checked ProjectData stuff, but other changes look nice. just a question: you removed logog dependency from most of libs. is it because BaseLib depends on it?

endJunction:
The logog is removed as dependency from most libs but BaseLib, because the default behaviour for linking is transitive. See [cmake doc](https://cmake.org/cmake/help/v3.3/command/target_link_libraries.html#libraries-for-both-a-target-and-its-dependents).

I'm yet figuring out, what the minimum specification is, such that all links fine w/o overhead. But first resolving the circular dependencies bugs.

norihiro-w:
thanks for the explanation!

endJunction:
__Update:__ Initially I thought the EnableSharedLibs PR would be much easier, but first I need to break the circular deps. So this is now a stand-alone PR not based on shared libs but set up as a prerequisite for them.
This is actually reviewed and all previous comments are incorporated.

norihiro-w:
:+1:

PR #1139, opened 2016-04-13
[GL] Introducing class LineSegment.

TomFischer:
PR improves the existing class `LineSegment` and establishs it at some places in the `GeoLib`. The changes should improve the readability.

Furthermore the PR introduces a `SegmentIterator` in class `Polyline`. The iterator class allows to use algorithms from the standard library and makes the code more readable at some places.

### DIFF
--- a/FileIO/GmshIO/GMSHPolygonTree.cpp
+++ b/FileIO/GmshIO/GMSHPolygonTree.cpp
@@ -86,7 +86,11 @@ void GMSHPolygonTree::insertPolyline (GeoLib::PolylineWithSegmentMarker * ply)
 			if (! ply->isSegmentMarked(k)) {
 				std::size_t seg_num(0);
 				GeoLib::Point *intersection_pnt(new GeoLib::Point);
-				while (_node_polygon->getNextIntersectionPointPolygonLine(*(ply->getPoint(k)), *(ply->getPoint(k+1)), intersection_pnt, seg_num)) {
+				while (_node_polygon->getNextIntersectionPointPolygonLine(
+				    {const_cast<GeoLib::Point*>(ply->getPoint(k)),
+				     const_cast<GeoLib::Point*>(ply->getPoint(k + 1))},
+				    *intersection_pnt, seg_num))
+				{
 					// insert the intersection point to point vector of GEOObjects instance
 					const std::size_t pnt_vec_size(pnt_vec.size());
 					// appendPoints deletes an already existing point
@@ -110,7 +114,11 @@ void GMSHPolygonTree::insertPolyline (GeoLib::PolylineWithSegmentMarker * ply)
 					}
 
 					std::size_t tmp_seg_num(seg_num+1);
-					if (! _node_polygon->getNextIntersectionPointPolygonLine(*(ply->getPoint(k)), *(ply->getPoint(k+1)), &tmp_pnt, tmp_seg_num)) {
+					if (!_node_polygon->getNextIntersectionPointPolygonLine(
+					        {const_cast<GeoLib::Point*>(ply->getPoint(k)),
+					         const_cast<GeoLib::Point*>(ply->getPoint(k + 1))},
+					        tmp_pnt, tmp_seg_num))
+					{
 						// check a point of the segment except the end points
 						for (std::size_t i(0); i<3; i++) {
 							tmp_pnt[i] = ((*(ply->getPoint(k)))[i] + (*(ply->getPoint(k+1)))[i]) / 2;
--- a/GeoLib/AnalyticalGeometry.cpp
+++ b/GeoLib/AnalyticalGeometry.cpp
@@ -215,26 +215,22 @@ bool lineSegmentIntersect(
 }
 
 bool lineSegmentsIntersect(const GeoLib::Polyline* ply,
-                           std::size_t &idx0,
-                           std::size_t &idx1,
+                           GeoLib::Polyline::SegmentIterator &seg_it0,
+                           GeoLib::Polyline::SegmentIterator &seg_it1,
                            GeoLib::Point& intersection_pnt)
 {
-	std::size_t n_segs(ply->getNumberOfPoints() - 1);
-	/**
-	 * computing the intersections of all possible pairs of line segments of the given polyline
-	 * as follows:
-	 * let the segment \f$s_1 = (A,B)\f$ defined by \f$k\f$-th and \f$k+1\f$-st point
-	 * of the polyline and segment \f$s_2 = (C,B)\f$ defined by \f$j\f$-th and
-	 * \f$j+1\f$-st point of the polyline, \f$j>k+1\f$
-	 */
-	for (std::size_t k(0); k < n_segs - 2; k++) {
-		for (std::size_t j(k + 2); j < n_segs; j++) {
-			if (k != 0 || j < n_segs - 1) {
-				if (lineSegmentIntersect(*(ply->getPoint(k)), *(ply->getPoint(k + 1)),
-				                         *(ply->getPoint(j)), *(ply->getPoint(j + 1)),
-				                         intersection_pnt)) {
-					idx0 = k;
-					idx1 = j;
+	std::size_t const n_segs(ply->getNumberOfSegments());
+	// Neighbouring segments always intersects at a common vertex. The algorithm
+	// checks for intersections of non-neighbouring segments.
+	for (seg_it0 = ply->begin(); seg_it0 != ply->end() - 2; ++seg_it0)
+	{
+		seg_it1 = seg_it0+2;
+		std::size_t const seg_num_0 = seg_it0.getSegmentNumber();
+		for ( ; seg_it1 != ply->end(); ++seg_it1) {
+			// Do not check first and last segment, because they are
+			// neighboured.
+			if (!(seg_num_0 == 0 && seg_it1.getSegmentNumber() == n_segs - 1)) {
+				if (lineSegmentIntersect(*seg_it0, *seg_it1, intersection_pnt)) {
 					return true;
 				}
 			}
--- a/GeoLib/AnalyticalGeometry.cpp
+++ b/GeoLib/AnalyticalGeometry.cpp
@@ -99,6 +99,14 @@ bool parallel(MathLib::Vector3 v, MathLib::Vector3 w)
 	return parallel;
 }
 
+bool lineSegmentIntersect(GeoLib::LineSegment const& s0,
+                          GeoLib::LineSegment const& s1,
+                          GeoLib::Point& s)
+{
+	return lineSegmentIntersect(s0.getBeginPoint(), s0.getEndPoint(),
+	                            s1.getBeginPoint(), s1.getEndPoint(), s);
+}
+
 bool lineSegmentIntersect(
 	GeoLib::Point const& a,
 	GeoLib::Point const& b,
--- a/GeoLib/AnalyticalGeometry.cpp
+++ b/GeoLib/AnalyticalGeometry.cpp
@@ -103,17 +103,11 @@ bool lineSegmentIntersect(GeoLib::LineSegment const& s0,
                           GeoLib::LineSegment const& s1,
                           GeoLib::Point& s)
 {
-	return lineSegmentIntersect(s0.getBeginPoint(), s0.getEndPoint(),
-	                            s1.getBeginPoint(), s1.getEndPoint(), s);
-}
+	GeoLib::Point const& a{s0.getBeginPoint()};
+	GeoLib::Point const& b{s0.getEndPoint()};
+	GeoLib::Point const& c{s1.getBeginPoint()};
+	GeoLib::Point const& d{s1.getEndPoint()};
 
-bool lineSegmentIntersect(
-	GeoLib::Point const& a,
-	GeoLib::Point const& b,
-	GeoLib::Point const& c,
-	GeoLib::Point const& d,
-	GeoLib::Point& s)
-{
 	if (!isCoplanar(a, b, c, d))
 		return false;
 
--- a/GeoLib/AnalyticalGeometry.cpp
+++ b/GeoLib/AnalyticalGeometry.cpp
@@ -518,21 +518,30 @@ bool isCoplanar(const MathLib::Point3d& a, const MathLib::Point3d& b,
 void computeAndInsertAllIntersectionPoints(GeoLib::PointVec &pnt_vec,
 	std::vector<GeoLib::Polyline*> & plys)
 {
+	auto computeSegmentIntersections = [&pnt_vec](GeoLib::Polyline& poly0,
+	                                              GeoLib::Polyline& poly1)
+	{
+		for (auto seg0_it(poly0.begin()); seg0_it != poly0.end(); ++seg0_it)
+		{
+			for (auto seg1_it(poly1.begin()); seg1_it != poly1.end(); ++seg1_it)
+			{
+				GeoLib::Point s(0.0, 0.0, 0.0, pnt_vec.size());
+				if (lineSegmentIntersect(*seg0_it, *seg1_it, s))
+				{
+					std::size_t const id(
+					    pnt_vec.push_back(new GeoLib::Point(s)));
+					poly0.insertPoint(seg0_it.getSegmentNumber() + 1, id);
+					poly1.insertPoint(seg1_it.getSegmentNumber() + 1, id);
+				}
+			}
+		}
+	};
+
 	for (auto it0(plys.begin()); it0 != plys.end(); ++it0) {
 		auto it1(it0);
 		++it1;
 		for (; it1 != plys.end(); ++it1) {
-			for (std::size_t i(0); i<(*it0)->getNumberOfPoints()-1; i++) {
-				for (std::size_t j(0); j<(*it1)->getNumberOfPoints()-1; j++) {
-					GeoLib::Point s(0.0, 0.0, 0.0, pnt_vec.size());
-					if (lineSegmentIntersect(*(*it0)->getPoint(i), *(*it0)->getPoint(i+1),
-						*(*it1)->getPoint(j), *(*it1)->getPoint(j+1), s)) {
-						std::size_t const id(pnt_vec.push_back(new GeoLib::Point(s)));
-						(*it0)->insertPoint(i+1, id);
-						(*it1)->insertPoint(j+1, id);
-					}
-				}
-			}
+			computeSegmentIntersections(*(*it0), *(*it1));
 		}
 	}
 }
--- a/GeoLib/LineSegment.cpp
+++ b/GeoLib/LineSegment.cpp
@@ -1,0 +1,82 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/LICENSE.txt
+ */
+
+
+#include "LineSegment.h"
+
+#include <iostream>
+
+namespace GeoLib
+{
+LineSegment::LineSegment(Point* const a, Point* const b,
+                         bool point_mem_management_by_line_segment)
+    : _a(a),
+      _b(b),
+      _point_mem_management_by_line_segment(
+          point_mem_management_by_line_segment)
+{}
+
+LineSegment::LineSegment(LineSegment const& line_segment)
+    : _a(new Point(line_segment.getBeginPoint())),
+      _b(new Point(line_segment.getEndPoint())),
+      _point_mem_management_by_line_segment(true)
+{}
+
+LineSegment::LineSegment(LineSegment&& line_segment)
+    : _a(std::move(line_segment._a)),
+      _b(std::move(line_segment._b)),
+      _point_mem_management_by_line_segment(
+          line_segment._point_mem_management_by_line_segment)
+{
+	line_segment._a = nullptr;
+	line_segment._b = nullptr;
+	line_segment._point_mem_management_by_line_segment = false;
+}
+
+LineSegment::~LineSegment()
+{
+	if (_point_mem_management_by_line_segment) {
+		delete _b;
+		delete _a;
+	}
+}
+
+Point const& LineSegment::getBeginPoint() const
+{
+	return *_a;
+}
+
+Point & LineSegment::getBeginPoint()
+{
+	return *_a;
+}
+
+Point const& LineSegment::getEndPoint() const
+{
+	return *_b;
+}
+
+Point & LineSegment::getEndPoint()
+{
+	return *_b;
+}
+
+std::ostream& operator<< (std::ostream& os, LineSegment const& s)
+{
+	os << "{(" << s.getBeginPoint() << "), (" << s.getEndPoint() << ")}";
+	return os;
+}
+
+std::ostream& operator<<(std::ostream& os,
+                         std::pair<GeoLib::LineSegment const&,
+                                   GeoLib::LineSegment const&> const& seg_pair)
+{
+	os << seg_pair.first << " x " << seg_pair.second;
+	return os;
+}
+}
--- a/GeoLib/Polygon.cpp
+++ b/GeoLib/Polygon.cpp
@@ -106,10 +106,10 @@ std::vector<GeoLib::Point> Polygon::getAllIntersectionPoints(
 		GeoLib::LineSegment const& segment) const
 {
 	std::vector<GeoLib::Point> intersections;
-	const std::size_t n_segments(getNumberOfPoints() - 1);
 	GeoLib::Point s;
-	for (std::size_t k(0); k < n_segments; k++) {
-		if (GeoLib::lineSegmentIntersect(*(getPoint(k)), *(getPoint(k+1)), a, b, s)) {
+	for (auto seg_it(begin()); seg_it != end(); ++seg_it)
+	{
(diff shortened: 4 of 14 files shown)


TomFischer:
Jenkins, test this please.

TomFischer:
Just let you know the status of the PR:
- The testrunner is failing under Windows builds.
- The testrunner is running fine under Mac and linux builds.
- The Clang sanitizer under linux reports some mem leaks. These issues seems to be Qt related.
- Valgrind kills the program because of an instruction valgrind does not understand. The place where this is reported is related to the test that is failing under Windows. A snippet of the valgrind message is below. Maybe someone has an idea to find the bug?

valgrind: Unrecognised instruction at address 0x136f6ad5.             
 at 0x136F6AD5: std::(anonymous namespace)::__x86_rdrand() (random.cc:69)
 by 0x136F6C71: std::random_device::_M_getval() (random.cc:130)
 by 0xD40864: std::random_device::operator()() (in /home/fischeth/w/OGS-6/BuildDebugWithCLangGui/bin/testrunner)
 by 0xD40217: autocheck::rng() (generator.hpp:20)
 by 0xE539F7: autocheck::generator<double, void>::operator()(unsigned long) (generator.hpp:162)
 by 0xE53798: autocheck::RandomCirclePointGeneratorXY<autocheck::generator<double, void> >::operator()(unsigned long) (AutoCheckGenerators.h:38)
 by 0xE53537: autocheck::SymmSegmentGeneratorXY<autocheck::RandomCirclePointGeneratorXY<autocheck::generator<double, void> > >::operator()(unsigned long) (AutoCheckGenerato      rs.h:71)
 by 0xE5341A: std::tuple<GeoLib::LineSegment, GeoLib::LineSegment> autocheck::generate<std::tuple<GeoLib::LineSegment, GeoLib::LineSegment>, autocheck::SymmSegmentGenerator      XY<autocheck::RandomCirclePointGeneratorXY<autocheck::generator<double, void> > >, autocheck::SymmSegmentGeneratorXY<autocheck::RandomCirclePointGeneratorXY<autocheck::generator<double      , void> > >, 0, 1>(std::tuple<autocheck::SymmSegmentGeneratorXY<autocheck::RandomCirclePointGeneratorXY<autocheck::generator<double, void> > >, autocheck::SymmSegmentGeneratorXY<autoch      eck::RandomCirclePointGeneratorXY<autocheck::generator<double, void> > > >&, unsigned long, autocheck::range<0, 0, 1> const&) (generator.hpp:28)
 by 0xE52E07: std::tuple<GeoLib::LineSegment, GeoLib::LineSegment> autocheck::generate<std::tuple<GeoLib::LineSegment, GeoLib::LineSegment>, autocheck::SymmSegmentGenerator      XY<autocheck::RandomCirclePointGeneratorXY<autocheck::generator<double, void> > >, autocheck::SymmSegmentGeneratorXY<autocheck::RandomCirclePointGeneratorXY<autocheck::generator<double      , void> > > >(std::tuple<autocheck::SymmSegmentGeneratorXY<autocheck::RandomCirclePointGeneratorXY<autocheck::generator<double, void> > >, autocheck::SymmSegmentGeneratorXY<autocheck::      RandomCirclePointGeneratorXY<autocheck::generator<double, void> > > >&, unsigned long) (in /home/fischeth/w/OGS-6/BuildDebugWithCLangGui/bin/testrunner)
 by 0xE526A0: autocheck::arbitrary<autocheck::SymmSegmentGeneratorXY<autocheck::RandomCirclePointGeneratorXY<autocheck::generator<double, void> > >, autocheck::SymmSegmentG      eneratorXY<autocheck::RandomCirclePointGeneratorXY<autocheck::generator<double, void> > > >::operator()(autocheck::value<std::tuple<GeoLib::LineSegment, GeoLib::LineSegment> >&) (arbit      rary.hpp:50)
by 0xE4D2B1: void autocheck::check<GeoLib::LineSegment, GeoLib::LineSegment, LineSegmentIntersect2dTest_RandomSegmentOrientationIntersecting_Test::TestBody()::$_0, autoche      ck::arbitrary<autocheck::SymmSegmentGeneratorXY<autocheck::RandomCirclePointGeneratorXY<autocheck::generator<double, void> > >, autocheck::SymmSegmentGeneratorXY<autocheck::RandomCircl      ePointGeneratorXY<autocheck::generator<double, void> > > > >(LineSegmentIntersect2dTest_RandomSegmentOrientationIntersecting_Test::TestBody()::$_0, unsigned long, autocheck::arbitrary<      autocheck::SymmSegmentGeneratorXY<autocheck::RandomCirclePointGeneratorXY<autocheck::generator<double, void> > >, autocheck::SymmSegmentGeneratorXY<autocheck::RandomCirclePointGenerato      rXY<autocheck::generator<double, void> > > >, autocheck::reporter const&, autocheck::classifier<GeoLib::LineSegment, GeoLib::LineSegment>, bool) (check.hpp:39)
by 0xE4D1BA: LineSegmentIntersect2dTest_RandomSegmentOrientationIntersecting_Test::TestBody() (TestLineSegmentIntersect2d.cpp:78)
Your program just tried to execute an instruction that Valgrind did not recognise.  There are two possible reasons for this.          
 1. Your program has a bug and erroneously jumped to a non-code location.  If you are running Memcheck and you just saw a warning about a bad jump, it's probably your program's fault.      
 2. The instruction is legitimate but Valgrind doesn't handle it, i.e. it's Valgrind's fault.  If you think this is the case or you are not sure, please let us know and we'll try to fix it.      
 Either way, Valgrind will now raise a SIGILL signal which will        
 probably kill your program.

norihiro-w:
It seems valgrind failed because the test uses RDRAND instruction (__x86_rdrand()) which is on-chip hardware random number generator (see https://en.wikipedia.org/wiki/RdRand). As this is hardware dependent, valgrind may not suppose it in some hardware. see, e.g., https://bugs.launchpad.net/ubuntu/+source/valgrind/+bug/852795

I'm not sure how we can fix it. you may need to change some code in autocheck where it specifies how to generate random numbers, i.e. change it to use software random number generator.

chleh:
Currently I don't feel like I want to contemplate about your changes all by myself. I think this PR is a good candidate for a personal introduction by yourself to the reviewers (less thinking required on my side, and maybe it's faster). Of course, if others volunteered for reviewing without prior presentation, that would be fine, too.

TomFischer:
@chleh The main points of the PR are the changes to class `LineSegment` and the introduction of the class `Polyline::SegmentIterator`. All other changes are related to this.

Are you available tomorrow morning for an explanation?

GeoLib/LineSegment.cpp
```diff
@@ -0,0 +1,88 @@
...
+      _b(std::move(line_segment._b)),
+      _point_mem_management_by_line_segment(
+          line_segment._point_mem_management_by_line_segment)
+{
+	line_segment._a = nullptr;
+	line_segment._b = nullptr;
+	line_segment._point_mem_management_by_line_segment = false;
+}
+
+LineSegment::LineSegment()
```
endJunction: This is no longer needed...

TomFischer:
Commit 2c86725 and d25f04e results of a discussion with @chleh.
- first commit documents/explains the default value of the third argument
- second commit is simply using modern C++

TomFischer:
Jenkins, test this please.

GeoLib/Polygon.h
```diff
@@ -96,33 +94,28 @@ class Polygon : public Polyline
...
 	 * @param intersection_pnt (output) next intersection point
-	 * @param seg_num (input/output) the number of the polygon segment that is intersecting
+	 * @param seg_num (in/out) the number of the polygon segment that is intersecting
 	 */
-	bool getNextIntersectionPointPolygonLine(GeoLib::Point const & a,
-									GeoLib::Point const & b,
-									GeoLib::Point* intersection_pnt,
-									std::size_t& seg_num) const;
-
+	bool getNextIntersectionPointPolygonLine(GeoLib::LineSegment const& seg,
```
chleh: Could you please document the meaning of the Boolean return value, too.

GeoLib/Polygon.cpp
```diff
@@ -191,44 +192,38 @@ bool Polygon::isPartOfPolylineInPolygon(const Polyline& ply) const
...
+bool Polygon::getNextIntersectionPointPolygonLine(
+    GeoLib::LineSegment const& seg, GeoLib::Point & intersection,
+    std::size_t& seg_num) const
 {
 	if (_simple_polygon_list.empty()) {
-		const std::size_t n_segments(getNumberOfPoints() - 1);
-		for (std::size_t k(seg_num); k < n_segments; k++) {
-			if (GeoLib::lineSegmentIntersect(*(getPoint(k)), *(getPoint(k + 1)), a, b, *intersection_pnt)) {
-				seg_num = k;
+		for (auto seg_it(begin()+seg_num); seg_it != end(); ++seg_it) {
```
chleh: I'm just wondering: The search doesn't wrap around at the last segment. What's the use-case of skipping some polygon edges?

GeoLib/Polygon.cpp
```diff
@@ -191,44 +192,38 @@ bool Polygon::isPartOfPolylineInPolygon(const Polyline& ply) const
...
 		for (std::list<Polygon*>::const_iterator it(_simple_polygon_list.begin()); it
 					!= _simple_polygon_list.end(); ++it) {
-			const Polygon* polygon(*it);
-			const std::size_t n_nodes_simple_polygon(polygon->getNumberOfPoints() - 1);
-			for (std::size_t k(0); k < n_nodes_simple_polygon; k++) {
-				if (GeoLib::lineSegmentIntersect(*(polygon->getPoint(k)), *(polygon->getPoint(k + 1)),
-								a, b, *intersection_pnt)) {
-					seg_num = k;
+			Polygon const& polygon(*(*it));
+			for (auto seg_it(polygon.begin()); seg_it != polygon.end(); ++seg_it)
```
chleh: Just a note: Could be `for (auto& seg : polygon)`, like you also wrote it in other places.

Tests/GeoLib/AutoCheckGenerators.h
```diff
@@ -68,10 +68,9 @@ struct SymmSegmentGeneratorXY
 
 	result_type operator()(std::size_t /*size*/ = 0)
```
chleh: Not this PR, but a defaulted commented argument is BAAAD.

Tests/GeoLib/TestLineSegmentIntersect2d.cpp
```diff
@@ -50,18 +50,29 @@ class LineSegmentIntersect2dTest : public testing::Test
 	ac::gtest_reporter gtest_reporter;
 };
 
+#if not defined(_MSC_VER) && _MSC_VER < 1900
+// Compilers of MVS below 2015 do not support unrestricted unions. The
+// unrestricted union is used by autocheck to handle test data. The autocheck
+// workaround for MVS compilers (below version 2015) contains a bug and in the
+// consequence the tests crashes. For this reason the tests are disabled under
+// this environments.
```
chleh: But it at least compiles under old MSVC++, right?
Then I'd prefer not to `#if` out the whole test, but to do
```
#if ...
TEST_F(LineSegmentIntersect2dTest, RandomSegmentOrientationIntersecting)
#else
TEST_F(LineSegmentIntersect2dTest, RandomSegmentOrientationIntersecting_DISABLED)
#endif
```
That's how we treat non-working PETSc tests, too.

Maybe you could also add a comment on the experienced error message, either in the code or in the discussion of the PR, because it took some time to figure out what the error was and that it was not your/pur fault.

Tests/GeoLib/TestPolygon.cpp
```diff
@@ -138,40 +139,45 @@ TEST_F(PolygonTest, isPntInPolygonCheckOuterPoints)
...
+	{
+		EXPECT_TRUE(_polygon->containsSegment(*segment_it));
 	}
 
 	{ // 70
-		GeoLib::Point const& a = *(_polygon->getPoint(7));
-		GeoLib::Point const& b = *(_polygon->getPoint(0));
-		ASSERT_TRUE(_polygon->containsSegment(a, b));
+		GeoLib::LineSegment const segment{
+		    const_cast<GeoLib::Point*>(_polygon->getPoint(7)),
```
chleh: The casts are not so nice. But probably they are a consequence of the design. Let's talk on monday.

Tests/GeoLib/TestPolyline.cpp
```diff
@@ -102,6 +102,16 @@ TEST(GeoLib, PolylineTest)
...
 	ASSERT_TRUE(ply.isClosed());
 
+	std::size_t segment_cnt(0);
+	for (auto seg_it(ply.begin()); seg_it != ply.end(); ++seg_it)
+	{
+		++segment_cnt;
+	}
+	ASSERT_EQ(ply.getNumberOfSegments(), segment_cnt);
+
+	std::copy(ply.begin(), ply.end(),
```
chleh: That could make lots of output. I think we should use `std::cout` as little as possible in the tests. Dou you need the output? Can you do it with `DBUG` (probably not)?

FileIO/GmshIO/GMSHPolygonTree.cpp
```diff
@@ -86,7 +86,11 @@ void GMSHPolygonTree::insertPolyline (GeoLib::PolylineWithSegmentMarker * ply)
 			if (! ply->isSegmentMarked(k)) {
 				std::size_t seg_num(0);
 				GeoLib::Point *intersection_pnt(new GeoLib::Point);
-				while (_node_polygon->getNextIntersectionPointPolygonLine(*(ply->getPoint(k)), *(ply->getPoint(k+1)), intersection_pnt, seg_num)) {
+				while (_node_polygon->getNextIntersectionPointPolygonLine(
+				    {const_cast<GeoLib::Point*>(ply->getPoint(k)),
+				     const_cast<GeoLib::Point*>(ply->getPoint(k + 1))},
```
chleh: As already said: I don't like the casts. Please provide a non-const overload instead:
```Point* Polyline::getPoint(std::size_t i);```

File: FileIO/GmshIO/GMSHPolygonTree.cpp [hunk 10]
```diff
@@ -86,7 +86,11 @@ void GMSHPolygonTree::insertPolyline (GeoLib::PolylineWithSegmentMarker * ply)
 			if (! ply->isSegmentMarked(k)) {
 				std::size_t seg_num(0);
 				GeoLib::Point *intersection_pnt(new GeoLib::Point);
```
chleh: Not, this PR, but anyway: This line could possible introduce a memory leak. I'd feel more comfortable by using `std::unique_ptr` here and `release()`, when you `pnt_vec.push_back()`, because the big loop below is a bit big. :)

GeoLib/AnalyticalGeometry.cpp
```diff
@@ -207,26 +209,19 @@ bool lineSegmentIntersect(
...
-	for (std::size_t k(0); k < n_segs - 2; k++) {
-		for (std::size_t j(k + 2); j < n_segs; j++) {
-			if (k != 0 || j < n_segs - 1) {
-				if (lineSegmentIntersect(*(ply->getPoint(k)), *(ply->getPoint(k + 1)),
-				                         *(ply->getPoint(j)), *(ply->getPoint(j + 1)),
-				                         intersection_pnt)) {
-					idx0 = k;
-					idx1 = j;
+	std::size_t const n_segs(ply->getNumberOfSegments());
+	for (seg_it0 = ply->begin(); seg_it0 != ply->end() - 2; ++seg_it0)
```
chleh: Should there be a comment that you don't check neighbouring segments, because they always intersect (at the common vertex), but that this intersection has no special geometrical meaning (i.e., it does not mean that the polyline is self-intersecting.)?

GeoLib/AnalyticalGeometry.cpp
```diff
@@ -207,26 +209,19 @@ bool lineSegmentIntersect(
 }
 
 bool lineSegmentsIntersect(const GeoLib::Polyline* ply,
-                           std::size_t &idx0,
-                           std::size_t &idx1,
+                           GeoLib::Polyline::SegmentIterator &seg_it0,
+                           GeoLib::Polyline::SegmentIterator &seg_it1,
                            GeoLib::Point& intersection_pnt)
 {
```
chleh: Should there be an assertion that the segment iterators belong to `ply`?

GeoLib/AnalyticalGeometry.cpp
```diff
@@ -514,21 +509,30 @@ bool isCoplanar(const MathLib::Point3d& a, const MathLib::Point3d& b,
...
+		for (auto seg0_it(poly0.begin()); seg0_it != poly0.end(); ++seg0_it)
+		{
+			for (auto seg1_it(poly1.begin()); seg1_it != poly1.end(); ++seg1_it)
+			{
+				GeoLib::Point s(0.0, 0.0, 0.0, pnt_vec.size());
+				if (lineSegmentIntersect(*seg0_it, *seg1_it, s))
+				{
+					std::size_t const id(
+					    pnt_vec.push_back(new GeoLib::Point(s)));
+					poly0.insertPoint(seg0_it.getSegmentNumber() + 1, id);
```
chleh: Will that mess up the segment numbers stored in the iterator when `seg0_it` is incremented afterwards?

GeoLib/Polyline.h
```diff
@@ -141,6 +247,16 @@ class Polyline : public GeoObject
 	 * */
 	const Point* getPoint(std::size_t i) const;
```
chleh: Maybe there should be a non-const overload for this method.

chleh:
Many comments, but only minor issues. Afterwards :+1:

Tests/GeoLib/TestLineSegmentIntersect2d.cpp
```diff
@@ -50,18 +50,29 @@ class LineSegmentIntersect2dTest : public testing::Test
 	ac::gtest_reporter gtest_reporter;
 };
 
+#if not defined(_MSC_VER) && _MSC_VER < 1900
```
endJunction: MSVC does not understand `not` operator. Must be `!`. And maybe parantheses around last expression...

GeoLib/Polyline.h
```diff
@@ -50,6 +51,109 @@ enum class Location
 class Polyline : public GeoObject
 {
 public:
+	class SegmentIterator
+	    : public std::iterator<std::forward_iterator_tag, LineSegment>
```
rinkk: Could you put the iterator class in a seperate file? This takes up almost have of the header file and makes it difficult to look for polyline-specific methods.

GeoLib/Polygon.cpp
TomFischer: @chleh What do you mean by "wrap around at the last segment"?
By `seg_num != 0` it is possible to skip some segments. This can be useful when in a previous run the algorithm checked already `seg_num` line segments of the polyline.

GeoLib/Polygon.cpp
TomFischer: Unfortunately this is not possible, because the `seg` of the type `LineSegment` hasn't any knowledge about their position in the polyline. This information is only avaiable into the `SegmentIterator`. I wonder if I should change my design of `LineSegment` and `SegmentIterator` again?

Tests/GeoLib/TestPolyline.cpp
TomFischer: This was only a test. Will remove the code.

FileIO/GmshIO/GMSHPolygonTree.cpp
TomFischer: This code will change later on. I have already a [branch](https://github.com/TomFischer/ogs/tree/GMSHInterface) where `GMSHPolygonTree` will be reworked.

FileIO/GmshIO/GMSHPolygonTree.cpp
TomFischer: Good point. Will include your remark in this [GMSHInterface-branch](https://github.com/TomFischer/ogs/tree/GMSHInterface).

GeoLib/AnalyticalGeometry.cpp
TomFischer: The segment iterators `seg_it0` and `seg_it1` are employed for returning the information which segments are intersecting. They are set within the function and thus it is ensured that they belong to `ply`.

GeoLib/AnalyticalGeometry.cpp
TomFischer: As far as I see this is not the case. The segment iterator stores only the segment number that is not touched by inserting a point **into** or **after** the segment the `SegmentIterator` points to.

Tests/GeoLib/TestLineSegmentIntersect2d.cpp
TomFischer: - On MSVC++ the tests can be runned only with ` _MSC_VER >= 1900` that supports unrestricted unions.
- I hoped I explained the reason for disabling the tests in the comment below the `#if ...`. What can be improved in the explanation?
- What is the advantage to "#if ... #else ... #endif" each particular test instead of all tests at once?

GeoLib/Polyline.h
TomFischer: The iterator class is part of the interface of the polyline and has to declared in the header file. I moved the implementation to the cpp file. Now the number of lines for the class declaration is only 35 lines compared to 100 lines before (declaration + implementation).

GeoLib/Polygon.cpp
chleh: Thanks, that explains it.

GeoLib/Polygon.cpp
chleh: OK, I see. No, you don't have to change the design again.

Tests/GeoLib/TestLineSegmentIntersect2d.cpp
chleh: > What is the advantage to "#if ... #else ... #endif" each particular test instead of all tests at once?

When doing so, the word "DISABLED_" occurs explicitly in the source code. So it is rather easy to get all disabled tests. Furthermore the number of disabled tests is reported by testrunner. Furthermore it is at least made sure that the test itself can be compiled, even though it might crash at runtime. So as long as it compiles, I'd prefer disabling tests one by one.

Tests/GeoLib/TestLineSegmentIntersect2d.cpp
```diff
@@ -50,7 +50,7 @@ class LineSegmentIntersect2dTest : public testing::Test
 	ac::gtest_reporter gtest_reporter;
 };
 
-#if not defined(_MSC_VER) && _MSC_VER < 1900
+#if !defined(_MSC_VER) || (_MSC_VER < 1900)
```
chleh: `!defined(_MSC_VER)` disables tests most compilers, doesn't it?
I suggest `#if (!defined(_MSC_VER)) && _MSC_VER < 1900`

GeoLib/Polyline.cpp
```diff
@@ -450,6 +450,104 @@ double Polyline::getDistanceAlongPolyline(const MathLib::Point3d& pnt,
...
+{}
+
+Polyline::SegmentIterator::SegmentIterator(SegmentIterator const& src)
+    : _polyline(src._polyline), _segment_number(src._segment_number)
+{}
+
+Polyline::SegmentIterator& Polyline::SegmentIterator::operator=(
+    SegmentIterator const& rhs)
+{
+	assert(&(rhs._polyline) == &_polyline);
```
chleh: you could make `_polyline` a pointer and also copy that.

GeoLib/Polyline.cpp
```diff
@@ -450,6 +450,104 @@ double Polyline::getDistanceAlongPolyline(const MathLib::Point3d& pnt,
...
+	SegmentIterator t(rhs);
+	std::swap(t._segment_number, _segment_number);
```
chleh: Can't you just assign `_segment_number`?

GeoLib/Polyline.cpp
TomFischer: I do not understand. What would be the advantage to make `_polyline` a copy?

Tests/GeoLib/TestLineSegmentIntersect2d.cpp
TomFischer: I checked this. The result is that the code doesn't compile when unrestricted unions aren't supported because then a default constructor is needed. But I don't want to implement the default constructor only for the reason that *disabled* tests can be compiled.

chleh: OK, I don't insist.

Only out of curiosity:
> then a default constructor is needed

A default ctor for what? For your test case or for some OGS6 classes? Or for autocheck?

chleh: Regarding the autocheck issue: I think it's a bit unfortunate that our unit tests have higher demands on the compiler than the code they test...

TomFischer: Autocheck needs the default constructor when it doesn't handle the unrestricted union properly.

> I think it's a bit unfortunate that our unit tests have higher demands on the compiler than the code they test...

Thats true. But I think this was not intended when choosing autocheck.

endJunction:
It is approaching the "good to merge" state.
:fast_forward: from my side.

FileIO/GmshIO/GMSHPolygonTree.cpp
TomFischer: In the future the internal data of the polyline should be accessed via line segment iterator instead of getPoint().

PR #1143, opened 2016-04-14
Split FileIO MeshLib dependency in LayeredMeshGenerator

endJunction:
Move reading of raster files out of layered mesh generator.

Add convenience function to read all raster files given by a list of file names.

### DIFF
--- a/FileIO/AsciiRasterInterface.cpp
+++ b/FileIO/AsciiRasterInterface.cpp
@@ -215,4 +215,15 @@ void AsciiRasterInterface::writeRasterAsASC(GeoLib::Raster const& raster, std::s
     out.close();
 }
 
-} // end namespace GeoLib
+bool allRastersExist(std::vector<std::string> const& raster_paths)
+{
+	for (auto raster = raster_paths.begin(); raster != raster_paths.end();
+	     ++raster)
+	{
+		std::ifstream file_stream(*raster, std::ifstream::in);
+		if (!file_stream.good()) return false;
+		file_stream.close();
+	}
+	return true;
+}
+}
--- a/FileIO/AsciiRasterInterface.cpp
+++ b/FileIO/AsciiRasterInterface.cpp
@@ -216,7 +216,9 @@ void AsciiRasterInterface::writeRasterAsASC(GeoLib::Raster const& raster, std::s
     out.close();
 }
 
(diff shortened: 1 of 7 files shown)


FileIO/AsciiRasterInterface.h
```diff
@@ -49,6 +52,10 @@ class AsciiRasterInterface {
                                  double &min, double &max);
 };
 
+/// Reads a vector of rasters given by file names. On error nothing is returned,
+/// otherwise the returned vector contains pointers to the read rasters.
+boost::optional<std::vector<GeoLib::Raster const*>> readRasters(
```
norihiro-w: it looks bit redundant for me to use `boost::optional` for `std::vector`, because one can do the same thing using `std::vector::empty()`.

norihiro-w:
looks good. I have just one minor comment about usage of boost optional. the comment is not critical so go ahead if you don't like it

endJunction:
@norihiro-w I had the empty vector before and it works. Reasons for optional: When looking on the algorithm it is a one-to-one transformation of strings to rasters, and boost::optional indicates if everything was ok. Therefore one can distinguish if an error happened or nothing was to be read at all.

FileIO/AsciiRasterInterface.cpp
```diff
@@ -215,4 +216,29 @@ void AsciiRasterInterface::writeRasterAsASC(GeoLib::Raster const& raster, std::s
...
+		if (!file_stream.good()) return false;
+		file_stream.close();
+	}
+	return true;
+}
+
+boost::optional<std::vector<GeoLib::Raster const*>> readRasters(
+    std::vector<std::string> const& raster_paths)
+{
+	if (!allRastersExist(raster_paths)) return boost::none;
```
chleh: Only a note: Actually it is dangerous to first check if files exist, then believe that they exist and just try to read them blindly. The files could have vanished in between. @endJunction, you shouldn't spend time on that now, especially since the problem has not been introduced in this PR. But for the future we should keep in mind that files should be read in one go and not be checked for existence prior to opening.

FileIO/AsciiRasterInterface.h
chleh: I agree with @endJunction's answer below.

chleh:
:+1:

PR #1144, opened 2016-04-16
Simplify concrete processes

chleh:
This is a follow-up of #1136.

Changes:
* interpolation (necessary for nonlinear processes) does not require passing arrays anymore
* GWFlow moved to separate subdir, and entirely to separate namespace
* defined common local assembler interface for all processes
* store process data in a separate struct
Thereby it is easier for users to add own properties to a process, to read those properties from file, etc.
* process_variables are passed as a list – first step towards multi-variable processes
* `createLocalAssemblers()` now done by a general function – users see less template stuff.

The aim of this PR is that we can tell users who start implementing their processes: "Copy and paste GWFlow and be happy.". The changes should cover most things needed to implement nonlinear, monolithic processes having only single-component process variables.
A PR regarding secondary variable output will come soon.

PR #1146, opened 2016-04-17
Make local assembler independent of global matrices

chleh:
Follow-up of #1144.

Open issues
* steady diffusion test
* `setupLocalMatrix()` function returns `Eigen::Map` to dynamic-size `Eigen::MatrixXd`.

### DIFF
--- a/Applications/ApplicationsLib/ProjectData.cpp
+++ b/Applications/ApplicationsLib/ProjectData.cpp
@@ -33,7 +33,7 @@
 
 #include "UncoupledProcessesTimeLoop.h"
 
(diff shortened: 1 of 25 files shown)


chleh:
I just noticed that the changes introduced in this PR will make the assembly of analtytical Jacobians less efficient, since the already assembled local matrices/vectors M, K and b are not stored in the local assemblers anymore, hence cannot be reused. (Note: Mechanics is special. There M, K and b need not be reused since the Jacobian is given by an entirely different formula.)
Since the changes in this PR also make the overall code more complicated, I will close this PR now.

PR #1147, opened 2016-04-17
Fix math lib metis linking

endJunction:
This is a split-off from https://github.com/ufz/ogs/pull/1133 having only the MathLib  and metis changes.

This has been reviewed and all changes are included. Look at it as a small MathLib cleanup.


norihiro-w:
:+1:

PR #1149, opened 2016-04-19
Add parsing of piecewise linear curves.

endJunction:
Adds a `_curves` map from names to `MathLib::PiecewiseLinearInterpolation`s to project data when
```xml
<curves>
  <curve>
    <name>curveA</name>
    <coords>0 1 1.5</coords>
    <values>10 11 12.3</values>
  </curve>
</curves>
```
is present in the project file.

I currently use this feature to create time-dependent boundary conditions (but this is too early for PR); and @yonghui56 started work on Richard's flow, where the curves are used for parameters.

### DIFF
--- a/Applications/ApplicationsLib/ProjectData.cpp
+++ b/Applications/ApplicationsLib/ProjectData.cpp
@@ -359,8 +359,8 @@ void ProjectData::parseNonlinearSolvers(BaseLib::ConfigTree const& config)
 static std::unique_ptr<MathLib::PiecewiseLinearInterpolation>
 createPiecewiseLinearInterpolation(BaseLib::ConfigTree const& config)
 {
-	auto const coords = config.getConfParam<std::vector<double>>("coords");
-	auto const values = config.getConfParam<std::vector<double>>("values");
+	auto coords = config.getConfParam<std::vector<double>>("coords");
+	auto values = config.getConfParam<std::vector<double>>("values");
 	if (coords.empty() || values.empty())
 	{
 		ERR("The given co-ordinates or values vector is empty.");
@@ -373,7 +373,8 @@ createPiecewiseLinearInterpolation(BaseLib::ConfigTree const& config)
 	}
 
 	return std::unique_ptr<MathLib::PiecewiseLinearInterpolation>{
-	    new MathLib::PiecewiseLinearInterpolation{coords, values}};
+	    new MathLib::PiecewiseLinearInterpolation{std::move(coords),
+	                                              std::move(values)}};
 }
 
 void ProjectData::parseCurves(
--- a/Applications/ApplicationsLib/ProjectData.cpp
+++ b/Applications/ApplicationsLib/ProjectData.cpp
@@ -21,6 +21,7 @@
 #include "BaseLib/FileTools.h"
 #include "BaseLib/uniqueInsert.h"
 
+#include "MathLib/InterpolationAlgorithms/PiecewiseLinearInterpolation.h"
 #include "MeshLib/Mesh.h"
 
 #include "NumLib/ODESolver/TimeDiscretizationBuilder.h"
@@ -72,8 +73,10 @@ ProjectData::ProjectData(BaseLib::ConfigTree const& project_config,
 	}
 	_mesh_vec.push_back(mesh);
 
-	// process variables
+	// curves
+	parseCurves(project_config.getConfSubtreeOptional("curves"));
 
+	// process variables
 	parseProcessVariables(project_config.getConfSubtree("process_variables"));
 
 	// parameters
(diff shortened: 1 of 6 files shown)


Applications/ApplicationsLib/ProjectData.cpp
```diff
@@ -352,3 +355,41 @@ void ProjectData::parseNonlinearSolvers(BaseLib::ConfigTree const& config)
...
+		std::abort();
+	}
+	if (coords.size() != values.size())
+	{
+		ERR("The given co-ordinates and values vector sizes are different.");
+		std::abort();
+	}
+
+	return std::unique_ptr<MathLib::PiecewiseLinearInterpolation>{
+	    new MathLib::PiecewiseLinearInterpolation{coords, values}};
```
chleh: You could move the vectors.

Applications/ApplicationsLib/ProjectData.h
```diff
@@ -13,7 +13,10 @@
 #ifndef PROJECTDATA_H_
 #define PROJECTDATA_H_
 
+#include <map>
```
chleh: It's interesting that we didn't have that include before...
Same for string below.

Applications/ApplicationsLib/ProjectData.h
```diff
@@ -13,7 +13,10 @@
...
 #include <memory>
+#include <string>
+#include <boost/optional/optional.hpp>
```
chleh: I guess that already comes in via `ConfigTree`.

chleh:
Only minor comments :+1:

Applications/ApplicationsLib/ProjectData.cpp
endJunction: The `PiecewiseLinearInterpolation` ctors need to be modified. There are three possibilities:
 - Duplicate code and replace `const&` parameters with `&&` and `std::move` in initializer list.
 - Use universal reference, _i.e._ `template <typename T> PiecewiseLI(T&& coords, T&& values) : _coords(std::forward<T>(coords)), ...`.
 - Pass-by-value and `std::move` in initializer list.
I suggest to take the third option, because the vector's move operation is cheap and it covers the two cases of passed l-values and r-values w/o code duplication and templates.

cf. to "Effective modern c++" Item 41: "Consider pass by value for copyable parameters that are cheap to move and always copied."

endJunction: Fourth possibility is to implement ctor taking r-values only. Currently there are any calls with l-values.

chleh: > cf. to "Effective modern c++" Item 41: "Consider pass by value for copyable parameters that are cheap to move and always copied."

But we do not always copy the vectors. In fact, they are only copied in the unit tests, aren't they?

My opinion:
1. IMHO the best option
2. Problematic. Maybe then different calls of `std::vector`'s ctor other than copy/move could be generated implicitly, e.g., `vector(std::initializer_list)`, though I'm not sure about it. Anyway in this case the avoided code duplication is neglectable. Furthermore we don't do such things in any other place, do we?
3. There will always be a copy and we rely on the compilers optimizer to perform the copy elision.
4. There are have several other places where we've only implemented r-value reference arguments, so it wouldn't hurt (unless you can give an example that hurts :smile:).

Anyway, that's only a minor issue. Don't waste time on it. If you can't decide on a specific solution, please just leave a TODO comment in the code, so we can find the spot later on. I'd vote for option 1 or 4, choose either one of them, if you like.

MathLib/InterpolationAlgorithms/PiecewiseLinearInterpolation.cpp
```diff
@@ -13,30 +13,35 @@
...
 #include "BaseLib/quicksort.h"
 
 namespace MathLib
 {
-PiecewiseLinearInterpolation::PiecewiseLinearInterpolation(const std::vector<double>& supporting_points,
-                                                           const std::vector<double>& values_at_supp_pnts,
-                                                           bool supp_pnts_sorted) :
-	_supp_pnts(supporting_points), _values_at_supp_pnts(values_at_supp_pnts)
+PiecewiseLinearInterpolation::PiecewiseLinearInterpolation(
+    std::vector<double> supporting_points,
```
chleh: I don't like that solution, because it contradicts our current paradigm. Probably the code works: Move constructor called twice. Still, we should be extremely sure that we know by heart what we are doing here, and that it works as a general approach. It probably does, if Scott Meyers suggests it. But passing by value/ref/ref&& is already tricky enough, and this technique adds another facet. Will we remember in half a year, why we pass by value here? Can we teach others, why that's good? Will those taught be confused and then pass everything by value?

Applications/ApplicationsLib/ProjectData.cpp
endJunction: We always copy the vectors in PiecewiseLinearInterpolation ctor. Before this PR copy is via copy, with this PR copy is via move if r-value is passed.

chleh: So if I get you right, "copy" is everything except storing a reference to an external vector as a data member of the class?
Still I'd like to know the opinion of the others about that "paradigm shift". Maybe I'm the only one who is concerned.

wenqing:
Just one comment for future development: I think we need a class for curves (or functions), which can provide the value and also possibly the tangential of the function at a given variable.

endJunction:
After discussion with Tom, we move to r-values constructor and drop the const& one.

TomFischer:
@chleh Are your minor comments answered? If yes, from my point of view  you could set the "rebase and merge" flag.

PR #1151, opened 2016-04-19
Simplify concrete processes

chleh:
Changes:
* interpolation (necessary for nonlinear processes) does not require passing arrays anymore
* GWFlow moved to separate subdir, and entirely to separate namespace
* defined common local assembler interface for all processes
* store process data in a separate struct
Thereby it is easier for users to add own properties to a process, to read those properties from file, etc.
* process_variables are passed as a list – first step towards multi-variable processes
* `createLocalAssemblers()` now done by a general function – users see less template stuff.

The aim of this PR is that we can tell users who start implementing their processes: "Copy and paste GWFlow and be happy.". The changes should cover most things needed to implement nonlinear, monolithic processes having only single-component process variables.
A PR regarding secondary variable output will come soon.

### DIFF
--- a/AssemblerLib/LocalDataInitializer.h
+++ b/AssemblerLib/LocalDataInitializer.h
@@ -119,35 +119,17 @@ namespace AssemblerLib
 /// For example for MeshLib::Line a local assembler data with template argument
 /// NumLib::ShapeLine2 is created.
 template <
-    template <typename, typename> class LocalAssemblerDataInterface_,
-    template <typename, typename, typename, typename, unsigned> class LocalAssemblerData_,
-    typename GlobalMatrix_,
-    typename GlobalVector_,
+    typename LocalAssemblerInterface,
+    template <typename, typename, typename, typename, unsigned> class LocalAssemblerData,
+    typename GlobalMatrix,
+    typename GlobalVector,
     unsigned GlobalDim,
     typename... ConstructorArgs>
 class LocalDataInitializer
 {
-    template <typename ShapeFunction_>
-    using IntegrationMethod = typename NumLib::GaussIntegrationPolicy<
-                typename ShapeFunction_::MeshElement>::IntegrationMethod;
-
-    template <typename ShapeFunction_>
-    using LAData = LocalAssemblerData_<
-            ShapeFunction_,
-            IntegrationMethod<ShapeFunction_>,
-            GlobalMatrix_, GlobalVector_, GlobalDim>;
-
-    using LADataIntfPtr
-        = std::unique_ptr<LocalAssemblerDataInterface_<GlobalMatrix_, GlobalVector_>>;
-
-    using LADataBuilder = std::function<LADataIntfPtr(
-            MeshLib::Element const& e,
-            std::size_t const local_matrix_size,
-            unsigned const integration_order,
-            ConstructorArgs&&...
-        )>;
-
 public:
+    using LADataIntfPtr = std::unique_ptr<LocalAssemblerInterface>;
+
     LocalDataInitializer()
     {
         // /// Lines and points ///////////////////////////////////
@@ -280,6 +262,23 @@ public:
     }
 
 private:
+    using LADataBuilder = std::function<LADataIntfPtr(
+            MeshLib::Element const& e,
+            std::size_t const local_matrix_size,
+            unsigned const integration_order,
+            ConstructorArgs&&...
+        )>;
+
+    template <typename ShapeFunction>
+    using IntegrationMethod = typename NumLib::GaussIntegrationPolicy<
+                typename ShapeFunction::MeshElement>::IntegrationMethod;
+
+    template <typename ShapeFunction>
+    using LAData = LocalAssemblerData<
+            ShapeFunction,
+            IntegrationMethod<ShapeFunction>,
+            GlobalMatrix, GlobalVector, GlobalDim>;
+
     // Generates a function that creates a new LocalAssembler of type LAData<SHAPE_FCT>
     template<typename ShapeFct>
     static LADataBuilder makeLocalAssemblerBuilder()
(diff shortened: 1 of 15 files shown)


endJunction:
Maybe the repetition `GroundwaterFlow::GroundwaterFlowProcess` is not necessary and `GroundwaterFlow::Process` is sufficient? Now it would be good time to change it.

chleh:
I'd rather stick with verbose names here, because here is the point where users come in, and in some months there will probably be plenty of concrete processes. If every one of them is simply called `Process`, and their base class is called so, too, it will be harder to talk about them. There are not many occurences of that name; in particular almost all occurences are within the GWFlow namespace.

AssemblerLib/LocalDataInitializer.h
```diff
@@ -119,35 +119,17 @@ namespace AssemblerLib
 /// For example for MeshLib::Line a local assembler data with template argument
 /// NumLib::ShapeLine2 is created.
 template <
-    template <typename, typename> class LocalAssemblerDataInterface_,
-    template <typename, typename, typename, typename, unsigned> class LocalAssemblerData_,
-    typename GlobalMatrix_,
-    typename GlobalVector_,
+    class LocalAssemblerInterface,
```
wenqing: Might be better: class --> typename

TomFischer:
:+1:

AssemblerLib/LocalDataInitializer.h
```diff
@@ -119,35 +119,17 @@ namespace AssemblerLib
...
-
-    using LADataBuilder = std::function<LADataIntfPtr(
-            MeshLib::Element const& e,
-            std::size_t const local_matrix_size,
-            unsigned const integration_order,
-            ConstructorArgs&&...
-        )>;
-
 public:
+    using LADataIntfPtr = std::unique_ptr<LocalAssemblerInterface>;
```
endJunction: If I have seen it correctly this type is used only once. If I'm right, then it would be better to write it at that place once.

NumLib/Function/Interpolation.h
```diff
@@ -16,38 +16,69 @@
...
+namespace detail
+{
+
+template<unsigned DOFOffset, typename NodalValues, typename ShapeMatrix>
+void shapeFunctionInterpolate(
+        const NodalValues &/*nodal_values*/,
+        const ShapeMatrix &/*shape_matrix_N*/)
+{}
+
+template<unsigned DOFOffset, typename NodalValues, typename ShapeMatrix, typename... Numbers>
```
endJunction: `Numbers` is not descriptive. Better `ValueType` or `ValueTypes` or whatever the semantic of this is. Documentation would be good too...

AssemblerLib/LocalDataInitializer.h
chleh: This line is only move and not introduced in this PR. And it is used in four places, or so.

NumLib/Function/Interpolation.h
```diff
@@ -16,38 +16,69 @@
...
+template<typename NodalValues, typename ShapeMatrix, typename... Numbers>
 void shapeFunctionInterpolate(
         const NodalValues& nodal_values,
         const ShapeMatrix& shape_matrix_N,
-        std::array<double*, NodalDOF> interpolated_values
+        double& interpolated_value,
+        Numbers&... interpolated_values
         )
 {
+    auto const num_nodal_dof = sizeof...(interpolated_values) + 1;
```
endJunction: About names: not clear to me is how the number of the DOF is related to the number of Numbers aka `interpolated_values`. The `Numbers` name is very confusing to me.

NumLib/Function/Interpolation.h
chleh: This is in detail namespace. The documentation is given below for the function that is to be used by users. The current implementation restricts the type to `double`, as expressed by `Numbers`, i.e., you should not pass a vector. Furthermore, the name of the argument is descriptive, at least I hope so.

NumLib/Function/Interpolation.h
chleh: Actually I though `Numbers` would be less confusing than `ValueTypes`, since only scalars are accepted as parameters. And I thought, the meaning of the parameters of the function would be clear by the name of those parameters, not necessarily by the name of their type.
I also thought, that writing `typename... Doubles` would be more confusing than `typename... Numbers`.
And the relation between the number of d.o.f. and the number of arguments is manifest in the line you commented on.

chleh: Variadic arguments without variadic types would be the best solution. Let's ask Herb Sutter, if he can do anything about that. :smile:

NumLib/Function/Interpolation.h
endJunction: Add documentation link to that function please.
Would ScalarType reflect the semantic of that type?

ProcessLib/GroundwaterFlow/GroundwaterFlowProcessData.h
```diff
@@ -0,0 +1,50 @@
...
+            )
+        : hydraulic_conductivity(hydraulic_conductivity_)
+    {}
+
+    GroundwaterFlowProcessData(GroundwaterFlowProcessData&& other)
+        : hydraulic_conductivity(other.hydraulic_conductivity)
+    {}
+
+    //! Copies are forbidden.
+    GroundwaterFlowProcessData(GroundwaterFlowProcessData const&) = delete;
```
endJunction: Because there is a user defined move ctor, isn't it deleted anyway?

chleh: I don't know by heart. Deleting it explicitly does no harm IMHO.
http://en.cppreference.com/w/cpp/language/copy_constructor says, it is implicitly deleted.

endJunction:
:fast_forward:

ProcessLib/GroundwaterFlow/GroundwaterFlowProcessData.h
chleh: I'll delete the assignment operators, too.

PR #1153, opened 2016-04-19
Remove mesh lib on file io dependency

endJunction:
Somehow this dependency slipped past my mind.
Moving reading of the file into DE.

Use instead layered mesh constructor taking a GeoLib::Raster.
Reading of the raster happens inside the application now.

Typo was "wher". No changes to the comment otherwise.

### DIFF
--- a/Applications/DataExplorer/DataView/MshView.cpp
+++ b/Applications/DataExplorer/DataView/MshView.cpp
@@ -25,6 +25,7 @@
 
 #include "FileIO/SHPInterface.h"
 #include "FileIO/TetGenInterface.h"
+#include "FileIO/AsciiRasterInterface.h"
 
 #include "MeshLib/Mesh.h"
 #include "MeshLib/Node.h"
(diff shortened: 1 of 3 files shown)


Applications/DataExplorer/DataView/MshView.cpp
```diff
@@ -164,7 +165,17 @@ void MshView::openMap2dMeshDialog()
...
 	if (dlg.useRasterMapping())
 	{
-		if (!MeshLib::MeshLayerMapper::layerMapping(*result, dlg.getRasterPath(), dlg.getNoDataReplacement()))
+		std::unique_ptr<GeoLib::Raster> raster{
+		    FileIO::AsciiRasterInterface::readRaster(dlg.getRasterPath())};
+		if (!raster)
+		{
+			OGSError::box(QString::fromStdString(
+			    "Error mapping mesh. Could not read raster file " +
+			    dlg.getRasterPath()));
```
TomFischer: This is a nice side effect: A possible error message can be displayed directly in the DataExplorer.

TomFischer:
:fast_forward:

chleh:
:fast_forward: 
@endJunction, once you have finished disentangling the various OGS6 libs, could you please provide a drawing (maybe hand-drawn) depicting the dependency graph among the libs, because sometimes there could be two options: E.g., either MeshLib depends on FileIO or the other way around.

endJunction:
For uptodate relations see Doxygen, _e.g._ for [MeshLib](https://doxygen.opengeosys.org/dir_cbf36d127bae6c5ddda0a904f4deb7f8.html) where there is MeshLib<->FileIO dependency, but will be FileIO depends on MeshLib only after this is merged.

PR #1154, opened 2016-04-20
Cleanup and remove warnings

TomFischer:
While searching a bug I raised the warning level of the compiler. PR tries to fix some warnings.

### DIFF
--- a/GeoLib/Polyline.cpp
+++ b/GeoLib/Polyline.cpp
@@ -417,15 +417,15 @@ double Polyline::getDistanceAlongPolyline(const MathLib::Point3d& pnt,
 	double dist(-1.0), lambda;
 	bool found = false;
 	// loop over all line segments of the polyline
-	for (std::size_t k = 0; k < this->getNumberOfPoints() - 1; k++) {
+	for (std::size_t k = 0; k < getNumberOfPoints() - 1; k++) {
 		// is the orthogonal projection of the j-th node to the
 		// line g(lambda) = _ply->getPoint(k) + lambda * (_ply->getPoint(k+1) - _ply->getPoint(k))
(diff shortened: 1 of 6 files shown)


endJunction:
:fast_forward:

GeoLib/Polyline.cpp
```diff
@@ -94,7 +94,9 @@ void Polyline::insertPoint(std::size_t pos, std::size_t pnt_id)
 		}
 	}
 
-	std::vector<std::size_t>::iterator it(_ply_pnt_ids.begin() + pos);
+	std::vector<std::size_t>::iterator it(
+	    _ply_pnt_ids.begin() +
+	    static_cast<std::vector<std::size_t>::difference_type>(pos));
```
chleh: Minor thing: I think the code would be much more readable if you you'd provide a `const pos_signed` once in this method and would not cast many times. Or does `pos` change? At least it's not const.

chleh:
:+1:

chleh:
A side point: Maybe we should remove `-Wno-deprecated` from `CompilerSetup.cmake`
```
set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CPU_FLAGS} -std=c++11 -Wno-deprecated -Wall -Wextra")
```
Or will that blow up the warning list?

endJunction:
I've added some more of the explicit ctor/copy/assignment as an addendum here.

@chleh Eigen has some of, now deprecated, `throw()` things attached to some functions; noexcept would be c++11 standard. There are less then 100 warnings or so.

endJunction:
Jenkins, test this please!

Some fetching errors Jenkins' side....

endJunction:
Jenkins, test this please!

Some fetching errors Jenkins' side....

PR #1156, opened 2016-04-21
Tiny fixes in  tools.

TomFischer:
As titled.

### DIFF
--- a/Applications/Utils/MeshEdit/removeMeshElements.cpp
+++ b/Applications/Utils/MeshEdit/removeMeshElements.cpp
@@ -160,6 +160,9 @@ int main (int argc, char* argv[])
 	std::unique_ptr<MeshLib::Mesh const> new_mesh(MeshLib::removeElements(
 	    *mesh, searcher.getSearchedElementIDs(), mesh->getName()));
 
+	if (new_mesh == nullptr)
+		return EXIT_FAILURE;
+
 	// write into a file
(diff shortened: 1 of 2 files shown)


Applications/Utils/MeshEdit/removeMeshElements.cpp
```diff
@@ -160,6 +160,9 @@ int main (int argc, char* argv[])
 	std::unique_ptr<MeshLib::Mesh const> new_mesh(MeshLib::removeElements(
 	    *mesh, searcher.getSearchedElementIDs(), mesh->getName()));
 
+	if (new_mesh == nullptr)
+		return EXIT_FAILURE;
```
endJunction: I'd add a reason for the exit...

endJunction:
:fast_forward:

Applications/Utils/MeshEdit/removeMeshElements.cpp
TomFischer: The function `removeElements` prints already the reason for the exit. Besides the value `nullptr` is returned in two cases (no element or all elements of the mesh will be removed). This makes a specific message at this place difficult. Of course it is possible to add more logic to distinct the cases and print a reason here.

chleh:
:fast_forward:

Applications/Utils/MeshEdit/removeMeshElements.cpp
endJunction: OK.

Issue #1155, opened 2016-04-20
Copy of GeoLib::Polygon


endJunction:
There is an implicitly created copy ctor for GeoLib::Polygon required in
 - GeoLib/AnalyticalGeometry.cpp:561 `return GeoLib::Polygon(rot_polyline);`
 - GeoLib/Polygon.cpp:420 `GeoLib::Polygon* polygon0 (new GeoLib::Polygon(*(*polygon_it)));`

Polygon has a list of pointers, so the default created copy is making only shallow copy. I wonder if this is correct.

Anyway a copy ctor and assignment must be implemented.

TomFischer:
As far as I see this should be a bug.

TomFischer:
I am working on the issue in a [branch](https://github.com/TomFischer/ogs/tree/PolygonCopyConstructorAssignment).

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/1155).

PR #1157, opened 2016-04-22
[GL] Polygon: Impl. of copy constructor and assignment operator

TomFischer:
As titled.

Fixes issue #1155.

**Update (2016-04-25)**
Removed the assignment operator since it isn't used anywhere and marked it explicitly as deleted.

### DIFF
--- a/GeoLib/Polygon.cpp
+++ b/GeoLib/Polygon.cpp
@@ -396,7 +396,7 @@ void Polygon::splitPolygonAtIntersection(
 
 void Polygon::splitPolygonAtPoint (std::list<GeoLib::Polygon*>::iterator polygon_it)
 {
-	std::size_t n ((*polygon_it)->getNumberOfPoints() - 1), idx0 (0), idx1(0);
+	std::size_t n((*polygon_it)->getNumberOfPoints() - 1), idx0(0), idx1(0);
 	std::vector<std::size_t> id_vec(n);
 	std::vector<std::size_t> perm(n);
 	for (std::size_t k(0); k < n; k++)
@@ -417,27 +417,25 @@ void Polygon::splitPolygonAtPoint (std::list<GeoLib::Polygon*>::iterator polygon
 				std::swap (idx0, idx1);
 
 			// create two closed polylines
-			GeoLib::Polygon* polygon0 (new GeoLib::Polygon(*(*polygon_it)));
-			for (std::size_t k(0); k <= idx0; k++)
-				polygon0->addPoint ((*polygon_it)->getPointID (k));
-			for (std::size_t k(idx1 + 1); k < (*polygon_it)->getNumberOfPoints(); k++)
-				polygon0->addPoint ((*polygon_it)->getPointID (k));
-			polygon0->initialise();
+			GeoLib::Polyline polyline0{*(*polygon_it)};
+			for (std::size_t j(0); j <= idx0; j++)
+				polyline0.addPoint((*polygon_it)->getPointID(j));
+			for (std::size_t j(idx1 + 1);
+			     j < (*polygon_it)->getNumberOfPoints(); j++)
+				polyline0.addPoint((*polygon_it)->getPointID(j));
 
-			GeoLib::Polygon* polygon1 (new GeoLib::Polygon(*(*polygon_it)));
-			for (std::size_t k(idx0); k <= idx1; k++)
-				polygon1->addPoint ((*polygon_it)->getPointID (k));
-			polygon1->initialise();
+			GeoLib::Polyline polyline1{*(*polygon_it)};
+			for (std::size_t j(idx0); j <= idx1; j++)
+				polyline1.addPoint((*polygon_it)->getPointID(j));
 
 			// remove original polygon and add two new polygons
-			std::list<GeoLib::Polygon*>::iterator polygon0_it, polygon1_it;
-			polygon1_it =
-			        _simple_polygon_list.insert (_simple_polygon_list.erase (
-			                                             polygon_it), polygon1);
-			polygon0_it = _simple_polygon_list.insert (polygon1_it, polygon0);
+			auto polygon1_it = _simple_polygon_list.insert(
+			    _simple_polygon_list.erase(polygon_it), new Polygon(polyline1));
+			auto polygon0_it = _simple_polygon_list.insert(
+			    polygon1_it, new Polygon(polyline0));
 
-			splitPolygonAtPoint (polygon0_it);
-			splitPolygonAtPoint (polygon1_it);
+			splitPolygonAtPoint(polygon0_it);
+			splitPolygonAtPoint(polygon1_it);
 
 			return;
 		}
--- a/GeoLib/Polygon.cpp
+++ b/GeoLib/Polygon.cpp
@@ -28,13 +28,20 @@ Polygon::Polygon(const Polyline &ply, bool init) :
 {
 	if (init)
 		initialise ();
+	_simple_polygon_list.push_back(this);
 }
 
 Polygon::Polygon(Polygon const& other)
     : Polyline(other), _aabb(other._aabb)
 {
-	for (auto const* sub_polygon : other._simple_polygon_list) {
-		_simple_polygon_list.emplace_back(new Polygon(*sub_polygon));
+	_simple_polygon_list.push_back(this);
+	auto sub_polygon_it(other._simple_polygon_list.begin());
+	for (sub_polygon_it++;  // the first entry is the polygon itself, skip the
+	                        // entry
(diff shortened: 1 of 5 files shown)


GeoLib/Polygon.cpp
```diff
@@ -417,27 +417,25 @@ void Polygon::splitPolygonAtPoint (std::list<GeoLib::Polygon*>::iterator polygon
...
-			std::list<GeoLib::Polygon*>::iterator polygon0_it, polygon1_it;
-			polygon1_it =
-			        _simple_polygon_list.insert (_simple_polygon_list.erase (
-			                                             polygon_it), polygon1);
-			polygon0_it = _simple_polygon_list.insert (polygon1_it, polygon0);
-
-			splitPolygonAtPoint (polygon0_it);
-			splitPolygonAtPoint (polygon1_it);
+			auto polygon1_it = _simple_polygon_list.insert(
+			    _simple_polygon_list.erase(polygon_it), new Polygon(polyline1));
```
chleh: You erase a pointer to a polygon from `_simple_polygon_list`. Does that leak memory?
Just a note: It seems like `*polygon_it = new Polygon(polyline1);` is equivalent to the given statement.

GeoLib/Polygon.cpp
```diff
@@ -38,6 +38,25 @@ Polygon::Polygon(Polygon const& other)
 	}
 }
 
+Polygon& Polygon::operator=(Polygon const& other)
+{
+	if (other._ply_pnts != _ply_pnts) {
+		ERR(
+		    "Polygon::operator=: polygon other and this polygon are not based "
+		    "on the same point vector. Aborting");
```
chleh: Why is that necessary?

GeoLib/Polygon.cpp
```diff
@@ -38,6 +38,25 @@ Polygon::Polygon(Polygon const& other)
...
+		std::abort();
+	}
+	if (other == *this)
```
chleh: You compare for equality (i.e., compare the objects), not for identity (i.e., compare their addresses), here. I'd compare `&other == this`. That should also be faster.

GeoLib/Polygon.cpp
```diff
@@ -38,6 +38,25 @@ Polygon::Polygon(Polygon const& other)
...
+		return *this;
+
+	Polygon tmp(other);
+	std::swap(tmp._aabb, _aabb);
+	std::swap(tmp._simple_polygon_list, _simple_polygon_list);
+	std::swap(tmp._length, _length);
+	std::swap(tmp._ply_pnt_ids, _ply_pnt_ids);
```
chleh: You only have two data members inside `Polygon`: `_simple_polygon_list` and `_aabb`. Are the other two defined in the base class?
Then it would be better to use `operator=` of `Polyline` for them, e.g., `Polyline::operator=(*this, other);` (I'm not sure if it works like this.) If you think, the code becomes too complicated by that, just leave it as it is.

Tests/GeoLib/TestPolygon.cpp
```diff
@@ -199,6 +199,28 @@ TEST_F(PolygonTest, isPolylineInPolygon)
...
+	// create a polygon
+	GeoLib::Polyline polyline(_pnts);
+	polyline.addPoint(0);
+	polyline.addPoint(1);
+	polyline.addPoint(2);
+	polyline.addPoint(0);
+	GeoLib::Polygon polygon(polyline);
+	// assign an other polygon
+	polygon = *_polygon;
+	// ToDo change to segments
```
chleh: Very minor thing: I suggest to agree on one spelling for TODO. E.g., my IDE recognizes `TODO` and gathers all those lines in a nice list.

chleh:
Many comments, nothing big. Afterwards :+1:

GeoLib/Polygon.cpp
endJunction: The polygons do not store the points but have a reference to point vector. So depending on how the points are managed all can go wrong here w/o the if. The details need to be explained anyway; Tom?

GeoLib/Polygon.cpp
```diff
@@ -417,27 +444,25 @@ void Polygon::splitPolygonAtPoint (std::list<GeoLib::Polygon*>::iterator polygon
...
 			// create two closed polylines
-			GeoLib::Polygon* polygon0 (new GeoLib::Polygon(*(*polygon_it)));
+			GeoLib::Polyline polyline0{*(*polygon_it)};
 			for (std::size_t k(0); k <= idx0; k++)
-				polygon0->addPoint ((*polygon_it)->getPointID (k));
-			for (std::size_t k(idx1 + 1); k < (*polygon_it)->getNumberOfPoints(); k++)
-				polygon0->addPoint ((*polygon_it)->getPointID (k));
-			polygon0->initialise();
+				polyline0.addPoint((*polygon_it)->getPointID(k));
+			for (std::size_t k(idx1 + 1);
```
endJunction: Declaration shadows a local variable `k`; see line 402. Maybe not a bug but could be nicer.

GeoLib/Polygon.cpp
```diff
@@ -417,27 +444,25 @@ void Polygon::splitPolygonAtPoint (std::list<GeoLib::Polygon*>::iterator polygon
...
+			     k < (*polygon_it)->getNumberOfPoints(); k++)
+				polyline0.addPoint((*polygon_it)->getPointID(k));
 
-			GeoLib::Polygon* polygon1 (new GeoLib::Polygon(*(*polygon_it)));
+			GeoLib::Polyline polyline1{*(*polygon_it)};
 			for (std::size_t k(idx0); k <= idx1; k++)
```
endJunction: Same here...

GeoLib/Polygon.cpp
```diff
@@ -30,6 +30,33 @@ Polygon::Polygon(const Polyline &ply, bool init) :
...
+		ERR(
+		    "Polygon::operator=: polygon other and this polygon are not based "
+		    "on the same point vector. Aborting");
+		std::abort();
+	}
+	if (other == *this)
+		return *this;
+
+	Polygon tmp(other);
+	std::swap(tmp._aabb, _aabb);
```
endJunction: This can be handled in another PR but the precedence is created here:
`warning: definition of implicit copy assignment operator for 'AABB' is deprecated because it has a user-declared copy constructor [-Wdeprecated]`
... requested in std::swap<GeoLib::AABB> at this line.

GeoLib/Polygon.cpp
chleh: I just saw that `const std::vector<Point*> &_ply_pnts;` is a reference member of `Polyline`. Therefore it cannot be reset by the assignment operator.
Different thing: However, it would probably be cheaper to compare the addresses and not the whole vectors. Furthermore the check `other == *this` should probably be done first.

GeoLib/Polygon.cpp
endJunction: If this code is tested, there is no memleak reported by address sanitizer...

TomFischer: As far as I see the code isn't tested. At the moment I think @chleh is right. This is probably a memory leak.

GeoLib/Polygon.cpp
```diff
@@ -72,7 +72,7 @@ bool Polygon::isPntInPolygon (GeoLib::Point const & pnt) const
 	std::size_t n_intersections (0);
 	GeoLib::Point s;
 
-	if (_simple_polygon_list.empty ()) {
+	if (_simple_polygon_list.front() == this) {
```
chleh: There are two occasions of `_simple_polygon_list.empty()` left in Polygon.cpp. I guess, they also have to be changed.

chleh:
One comment, afterwards :+1:

endJunction:
:fast_forward:
Please make sure after a rebase that all tests pass (it takes sometimes some time).

TomFischer:
At my laptop all tests pass after rebase. Unfortunately I can not see why Jenkins fails because I am not at the UFZ and thus have no access to Jenkins. If somebody see that the failing Jenkins is not caused by my changes feel free to merge.

norihiro-w:
@TomFischer you can go to the Jenkins via https://my.ufz.de/login

TomFischer:
> @TomFischer you can go to the Jenkins via https://my.ufz.de/login

Thanks for the info. Will try it.

bilke:
@TomFischer I played around with some new Jenkins functionality. That is why there is another Jenkins build run (called `Jenkins` in the check list below) is there. I have merged #1170 which should fix this new Jenkins run. The `default` Jenkins job builds fine.

TomFischer:
@bilke Thanks for the info.

PR #1158, opened 2016-04-23
Prevent passing temporaries to ConfigTree ctor

chleh:
This PR explicitly deletes the ctor of ConfigTree that takes an r-value reference of boost::property_tree. That way passing temporaries will lead to a compilation error.

It is tempting to pass temporaries in unit tests, e.g. when there is only an empty config tree needed to construct some objects. However, passing a temporary leads to dangling references and program crash.

The change in this PR of course will only prevent passing temporaries directly to the ctor. If they are passed, e.g., via `const&` to a function that passes them on, you are screwed :smile:

### DIFF
--- a/BaseLib/ConfigTree.h
+++ b/BaseLib/ConfigTree.h
@@ -256,9 +256,16 @@ public:
      * i.e., warnings will also result in program abortion!
      */
     explicit ConfigTree(PTree const& tree,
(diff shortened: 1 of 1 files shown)


endJunction:
:fast_forward:

TomFischer:
:fast_forward:

PR #1159, opened 2016-04-23
Deleted LocalAssemblerBuilder

chleh:
Said class only meant one additional indirection when creating local assemblers. I think, we don't need it, so I removed it. Thereby our code should be a bit easier to understand now.

Changes:
* delete said class
* changed method transform to transformDereferenced
* LocalDataInitializer is final now


AssemblerLib/LocalDataInitializer.h
```diff
@@ -240,18 +242,19 @@ class LocalDataInitializer
 
     /// Sets the provided data_ptr to the newly created local assembler data.
     void operator()(
-            const MeshLib::Element& e,
+            std::size_t const element_id,
```
endJunction: Elements have an id internally: `MeshLib::Element::getID()`. Or is it a different id?

AssemblerLib/LocalDataInitializer.h
```diff
@@ -240,18 +242,19 @@ class LocalDataInitializer
...
+            MeshLib::Element const*const element,
```
endJunction: Why is using pointer here better than a reference?

endJunction:
I can imagine the reason for a class was absence of lambda support in MSVC at that point of time.

endJunction:
After comment (or change) of that element id :fast_forward:

AssemblerLib/LocalDataInitializer.h
chleh: I remember in the docs of the old LocalAssemblerBuilder, there was a comment
> \attention The index \c id is not necesserily the mesh item's id.

I don't know, when they can be different. Let's discuss tomorrow.

AssemblerLib/LocalDataInitializer.h
chleh: Changed in a later commit in this PR.

AssemblerLib/LocalDataInitializer.h
endJunction: Oh, that index. Yes they can be different. Please don't call it element_id. It is the mesh_item_id which can also refer to node ids (or edge, face ids, when implemented).

chleh: Renamed to `mesh_item_id`.

AssemblerLib/LocalDataInitializer.h
```diff
@@ -240,18 +242,19 @@ class LocalDataInitializer
...
-            ConstructorArgs&&... args)
+            ConstructorArgs&&... args) const
     {
-        auto const type_idx = std::type_index(typeid(e));
-        auto it = _builder.find(type_idx);
+        auto const type_idx = std::type_index(typeid(element));
+        auto const it = _builder.find(type_idx);
 
         if (it != _builder.end()) {
+            auto const num_local_dof = _dof_table.getNumElementDOF(mesh_item_id);
```
norihiro-w: It looks `_dof_table` is used only here. If the object is needed only for having the number of local DoF, you can pass the number in the constructor.

norihiro-w:
:+1:

wenqing:
:+1:

AssemblerLib/LocalDataInitializer.h
```diff
@@ -240,18 +242,19 @@ class LocalDataInitializer
 
     /// Sets the provided data_ptr to the newly created local assembler data.
     void operator()(
-            const MeshLib::Element& e,
+            std::size_t const mesh_item_id,
```
endJunction: Please add the previous comment regarding this id, which was:
```
/// The positions in the global matrix/vector are taken from
/// the LocalToGlobalIndexMap provided in the constructor at index \c id.
/// \attention The index \c id is not necesserily the mesh item's id.
```

The reason for not being the mesh element's id is more clear when we have two meshes. Then the second's mesh element ids are offset by first mesh size. They correspond to the entries in the dof table. So one could also call it dof_table_entry_id or sth. along this lines.

chleh: New comment is
```
/// Sets the provided \c data_ptr to the newly created local assembler data.
///
/// \attention
/// The index \c id is not necessarily the mesh item's id. Especially when
/// having multiple meshes it will differ from the latter.
```

PR #1160, opened 2016-04-23
Fixed-size local matrices for multi-variable processes

chleh:
GWFlow shall serve as a template for any other processes, at least for the time being. This PR extends GWFlow s.t. it can easily be extended to multiple variables.

Note: The given changes only allow extension to multi-variables if each of them has one component and uses the same shape functions. Generalizations of that are future work. Probably then we will need to discuss if generalizations can be done using fixed matrices (theoretically, yes, but the code could be complicated).


endJunction:
Looks innocent.

I'm not sure whether it is good idea to "generalize" GWF process instead of having some `DummyProcessTemplateProcess`.

:fast

TomFischer:
:+1:

PR #1145, opened 2016-04-16
Extrapolation of integration point values to nodes

chleh:
Follow-up of #1161.

This PR now compiles also with PETSc. However, I think it currently does not make sense to make it also run with PETSc, there are simply too many odds and ends, e.g.:
* we cannot write to ghost nodes and sum up ghost nodes etc.
* the unit test fails because `NodePartitionedMesh` is hard-wired into `MeshComponentMap` for the PETSc build.

Future work will also be to extend the extrapolator to multi-component variables.

The files related to this PR are all in NumLib/Extrapolation and Tests/Numlib
I think it's best to wait with the review until #1161 is merged and this branch is rebased. 

In the meantime the unit test will give you an impression how the interpolation looks like from the user side: https://github.com/chleh/ogs6/blob/extrapol/Tests/NumLib/TestExtrapolation.cpp
Soon there will be a further PR connecting the Extrapolator to the Process.

Note: With this PR the LocalLinearLeastSquaresExtrapolator from NumLib uses SerialExecutor from AssemblerLib.

PR #1161, opened 2016-04-23
Misc changes

chleh:
This PR is a prerequisite for #1145.

Review component-wise.
Probably the biggest change is that `MatrixVectorProvider` gets not passed a `MatrixSpecificationsProvider` anymore, but rather is called directly with `MatrixSpecifications`.

### DIFF
--- a/MathLib/LinAlg/BLAS.cpp
+++ b/MathLib/LinAlg/BLAS.cpp
@@ -20,11 +20,27 @@ namespace MathLib { namespace BLAS
 {
 
 // Explicit specialization
+// Computes the Manhattan norm of x
+template<>
+double norm1(Eigen::VectorXd const& x)
+{
+    return x.lpNorm<1>();
+}
+
+// Explicit specialization
 // Computes the Euclidean norm of x
 template<>
 double norm2(Eigen::VectorXd const& x)
 {
     return x.norm();
+}
+
+// Explicit specialization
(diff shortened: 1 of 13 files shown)


MathLib/LinAlg/BLAS.cpp
```diff
@@ -20,13 +20,29 @@ namespace MathLib { namespace BLAS
 {
 
 // Explicit specialization
+// Computes the Manhattan norm of x
```
endJunction: In numerics it is usually called p1 norm (or more correctly l1 (like small-ell-one, $\ell_1$) norm). There are small-ell spaces and L-spaces with their p-norms... But Manhatten is not wrong, just unusual to me.

endJunction:
:fast

MathLib/LinAlg/BLAS.cpp
chleh: Just thought that Manhattan is in a line with Euclidean and Maximum, i.e. no numbers included in the name.
BTW: Wikipedia lists Manhattan twice :smile:
> The taxicab metric is also known as rectilinear distance, L1 distance or \ell_1 norm (see Lp space), snake distance, city block distance, Manhattan distance or Manhattan length
> (https://en.wikipedia.org/wiki/Taxicab_geometry)

norihiro-w:
:+1:

PR #1162, opened 2016-04-25
Data explorer project data holder

rinkk:
Removes the changes to Data Explorer classes in PR #1138 by introducing a seperate data holder object for Data Explorer.


DataHolderLib/DataExplorerProject.cpp
```diff
@@ -0,0 +1,116 @@
...
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#include "DataExplorerProject.h"
+
+#include <algorithm>
+
+// ThirdParty/logog
+#include "logog/include/logog.hpp"
```
endJunction: system include

DataHolderLib/DataExplorerProject.h
```diff
@@ -0,0 +1,81 @@
...
+public:
+	/// Constructor
+	DataExplorerProject();
+
+	DataExplorerProject(DataExplorerProject&) = delete;
+
+	~DataExplorerProject();
+
+	/// Returns the GEOObjects containing all points, polylines and surfaces.
+	GeoLib::GEOObjects* getGEOObjects()
```
endJunction: returning a reference would simplify some lines when using this function.

DataHolderLib/DataExplorerProject.cpp
```diff
@@ -0,0 +1,116 @@
...
+		isUnique = true;
+		cpName = name;
+
+		count++;
+		// If the original name already exists we start to add numbers to name for
+		// as long as it takes to make the name unique.
+		if (count > 1)
+			cpName = cpName + "-" + std::to_string(count);
+
+		for (std::vector<MeshLib::Mesh*>::const_iterator it = _mesh_vec.begin();
```
TomFischer: IMHO `for (auto const* mesh : _mesh_vec)` would be more readable. Additionaly in the if condition you can substitute `(*it)->getName()` with `mesh->getName()` which I think is much clearer. Maybe you could also use `std::string::operator==` instead of `std::string::compare` for better readability?

DataHolderLib/DataExplorerProject.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+// ThirdParty/logog
+#include "logog/include/logog.hpp"
+
+#include "BaseLib/FileTools.h"
+#include "BaseLib/uniqueInsert.h"
+
+#include "MeshLib/Mesh.h"
+
+DataExplorerProject::DataExplorerProject() = default;
```
TomFischer: Why not put this in the header?

DataHolderLib/DataExplorerProject.h
```diff
@@ -0,0 +1,81 @@
...
+	bool isMeshNameUniqueAndProvideUniqueName(std::string &name) const;
+
+	/// Returns true if a mesh with the same name exists and false otherwise.
+	bool meshExists(const std::string &name) const;
+
+
+	/// Returns an iterator to the first found mesh with the given name.
+	std::vector<MeshLib::Mesh*>::const_iterator findMeshByName(
+		std::string const& name) const;
+	std::vector<MeshLib::Mesh*>::iterator findMeshByName(
```
TomFischer: Is the non-const version really needed?

DataHolderLib/DataExplorerProject.h
```diff
@@ -0,0 +1,81 @@
...
+namespace MeshLib {
+	class Mesh;
+}
+
+/**
+ * The ProjectData Object contains all the data needed for a certain project, i.e. all
+ * geometric data (stored in a GEOObjects-object), all the meshes, processes,
+ * and process variables.
+ */
+class DataExplorerProject final
```
TomFischer: For at least consistency reasons I would move the class in a namespace `DataHolderLib`.

DataHolderLib/DataExplorerProject.h
```diff
@@ -0,0 +1,81 @@
...
+	bool meshExists(const std::string &name) const;
+
+
+	/// Returns an iterator to the first found mesh with the given name.
+	std::vector<MeshLib::Mesh*>::const_iterator findMeshByName(
+		std::string const& name) const;
+	std::vector<MeshLib::Mesh*>::iterator findMeshByName(
+		std::string const& name);
+
+	GeoLib::GEOObjects *_geoObjects = new GeoLib::GEOObjects();
```
TomFischer: If you really need to store `_geoObjects` as a pointer I would prefer to use `unique_ptr`.

DataHolderLib/DataExplorerProject.h
rinkk: Yes, it is needed whenever a mesh is modified (removing elements, adding scalars, changing mat ids, etc.)

DataHolderLib/DataExplorerProject.h
endJunction: Agree. A namespace is required.

DataHolderLib/Project.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+bool Project::removeMesh(const std::string &name)
+{
+	bool mesh_found = false;
+	std::vector<MeshLib::Mesh*>::iterator it = findMeshByName(name);
+	while (it != _mesh_vec.end())
+	{
+		delete *it;
+		*it = nullptr;
+		it = findMeshByName(name);
```
TomFischer: Why is `findMeshByName()` used a second time? As far as I understand the name of the mesh has to be unique. If the first search some lines above found a mesh with a corresponding name how should the second `findMeshByName()` result in a valid iterator?

DataHolderLib/Project.cpp
```diff
@@ -0,0 +1,116 @@
...
+		mesh_found = true;
+	}
+
+	_mesh_vec.erase(std::remove(_mesh_vec.begin(), _mesh_vec.end(), nullptr),
```
TomFischer: I wonder why the entry isn't erased above in line 69? Maybe we discuss this tomorrow?!

DataHolderLib/Project.h
```diff
@@ -0,0 +1,81 @@
...
+	/// Returns true if a mesh with the same name exists and false otherwise.
+	bool meshExists(const std::string &name) const;
+
+
+	/// Returns an iterator to the first found mesh with the given name.
+	std::vector<MeshLib::Mesh*>::const_iterator findMeshByName(std::string const& name) const;
+	std::vector<MeshLib::Mesh*>::iterator findMeshByName(std::string const& name);
+
+	GeoLib::GEOObjects _geoObjects;
+	std::vector<MeshLib::Mesh*> _mesh_vec;
```
chleh: When implementing new classes, I'd really like to see `std::unique_ptr` to be used for storing heap-allocated objects. I see, you expose `_mesh_vec` above (`getMeshObjects()`). So if it's not too much work, maybe you can change it.

DataHolderLib/Project.cpp
```diff
@@ -0,0 +1,116 @@
...
+	std::string name = mesh->getName();
+	getUniqueName(name);
+	mesh->setName(name);
+	_mesh_vec.push_back(mesh);
+}
+
+std::vector<MeshLib::Mesh*>::const_iterator Project::findMeshByName(
+		std::string const& name) const
+{
+	return const_cast<Project&>(*this).findMeshByName(name);
```
chleh: Nice trick!
But do you actually need both variants of that method? Effectively this method returns a `MeshLib::Mesh* const*` (an iterator to a const pointer to a mesh), so probably you can do with the mesh what you want, even in the const case.

DataHolderLib/Project.cpp
```diff
@@ -0,0 +1,116 @@
...
+}
+
+bool Project::removeMesh(const std::string &name)
+{
+	bool mesh_found = false;
+	std::vector<MeshLib::Mesh*>::iterator it = findMeshByName(name);
+	while (it != _mesh_vec.end())
+	{
+		delete *it;
+		*it = nullptr;
```
chleh: Could you add a comment that you mark `it` for later erasure, please.

DataHolderLib/Project.cpp
```diff
@@ -0,0 +1,116 @@
...
+		cpName = name;
+
+		count++;
+		// If the original name already exists we start to add numbers to name for
+		// as long as it takes to make the name unique.
+		if (count > 1)
+			cpName = cpName + "-" + std::to_string(count);
+
+		for (MeshLib::Mesh* mesh :_mesh_vec)
+			if ( cpName.compare(mesh->getName()) == 0 )
```
chleh: Note: You could check for equality by using `==`.

chleh:
Only minor comments from my side :+1:

DataHolderLib/Project.h
rinkk: All of this is just copy/paste from ProjectData, so it's not new code. Half of the things people are now commenting on haven't even been implemented by me. I'd rather keep it the way it is because otherwise this will be another large commit requiring changes in lots of DE classes. Maybe some other time.

DataHolderLib/Project.cpp
rinkk: again, this was not done by me, it's straight from projectdata. i do need the non-const method. having both is no bother so i'm just keeping it.

rinkk:
@TomFischer : Could you pls check why Jenkins failed (MathLib.DenseGaussAlgorithm)? I don't think this has anything to do with this PR.

TomFischer:
> Could you pls check why Jenkins failed (MathLib.DenseGaussAlgorithm)? I don't think this has anything to do with this PR.

Sometimes randomly generated matrices are not invertible. I guess this is the case here.

chleh:
@endJunction pointed out some time ago that it is nearly impossible (probability zero in R^(n x n)) to randomly generate a singular matrix.

rinkk:
It's okay ... the error popped up and I didn't know what it was about. No worries.

TomFischer:
> @endJunction pointed out some time ago that it is nearly impossible (probability zero in R^(n x n)) to randomly generate a singular matrix.

That is correct as long as we are not restricted to some floating point values of R and as long as the operations can be executed with infinite precision. And maybe the random number generator generates values that are not always that independend as needed?

TomFischer:
I send you a PR.

rinkk:
can this be merged?

PR #1166, opened 2016-04-26
Move AssemblerLib parts to MathLib and ProcessLib

endJunction:
- Move SparsityPattern typedef to MathLib; Breaking one small bit of circular dependency.
 - Those parts dealing with creation of the local assemblers belong to the ProcessLib.
 - The `GlobalSetup` is used in ProcessLib/NumericsConfig.h; so it is moved to ProcessLib too.

The previously ProcessUtils.h is split into two files for `createLocalAssemblers` and `initShapeMatrices` both placed in ProcessLib/Utils/. I can imagine that ProcessLib/LocalAssemblerUtils would be a better directory name.

No code changes, just includes and moves.

### DIFF
--- a/MathLib/LinAlg/SparsityPattern.h
+++ b/MathLib/LinAlg/SparsityPattern.h
@@ -1,0 +1,23 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef MATHLIB_LINALG_SPARSITYPATTERN_H
+#define MATHLIB_LINALG_SPARSITYPATTERN_H
+
+#include <vector>
+
+#include "ProcessLib/NumericsConfig.h"
+
+namespace MathLib
+{
+/// A vector telling how many nonzeros there are in each global matrix row.
+using SparsityPattern = std::vector<GlobalIndexType>;
+}
+
+#endif // MATHLIB_LINALG_SPARSITYPATTERN_H
--- a/ProcessLib/Process.h
+++ b/ProcessLib/Process.h
@@ -20,6 +20,7 @@
 #include "BaseLib/ConfigTree.h"
 #include "FileIO/VtkIO/VtuInterface.h"
 #include "MeshGeoToolsLib/MeshNodeSearcher.h"
+#include "MathLib/LinAlg/SparsityPattern.h"
 #include "MeshLib/MeshSubset.h"
 #include "MeshLib/MeshSubsets.h"
 
@@ -351,7 +352,7 @@ private:
 	std::unique_ptr<AssemblerLib::LocalToGlobalIndexMap>
 	    _local_to_global_index_map;
 
-	AssemblerLib::SparsityPattern _sparsity_pattern;
+	MathLib::SparsityPattern _sparsity_pattern;
 
 	std::vector<DirichletBc<GlobalIndexType>> _dirichlet_bcs;
 	std::vector<std::unique_ptr<NeumannBc<GlobalSetup>>> _neumann_bcs;
--- a/ProcessLib/Process.h
+++ b/ProcessLib/Process.h
@@ -10,35 +10,14 @@
 #ifndef PROCESS_LIB_PROCESS_H_
 #define PROCESS_LIB_PROCESS_H_
 
-#include <memory>
-#include <string>
-
-#include <logog/include/logog.hpp>
-
 #include "AssemblerLib/ComputeSparsityPattern.h"
-#include "AssemblerLib/LocalToGlobalIndexMap.h"
-#include "BaseLib/ConfigTree.h"
 #include "FileIO/VtkIO/VtuInterface.h"
-#include "MeshGeoToolsLib/MeshNodeSearcher.h"
-#include "MathLib/LinAlg/SparsityPattern.h"
-#include "MeshLib/MeshSubset.h"
-#include "MeshLib/MeshSubsets.h"
-
-#include "DirichletBc.h"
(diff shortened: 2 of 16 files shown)


ProcessLib/Process.h
```diff
@@ -20,6 +20,7 @@
 #include "BaseLib/ConfigTree.h"
 #include "FileIO/VtkIO/VtuInterface.h"
 #include "MeshGeoToolsLib/MeshNodeSearcher.h"
+#include "MathLib/LinAlg/SparsityPattern.h"
```
chleh: Is that needed? You already have the compute header.

MathLib/LinAlg/SparsityPattern.h
```diff
@@ -0,0 +1,23 @@
...
+#define MATHLIB_LINALG_SPARSITYPATTERN_H
+
+#include <vector>
+
+#include "ProcessLib/NumericsConfig.h"
+
+namespace MathLib
+{
+/// A vector telling how many nonzeros there are in each global matrix row.
+using SparsityPattern = std::vector<GlobalIndexType>;
```
chleh: Only an note s.t. I dont't forget it: For PETSc we could extend the sparsity pattern, because in PETSc matrices there is a pattern for the "diagonal" part and one for the "off-diagonal" part.
http://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Mat/MatCreateAIJ.html

ProcessLib/ProcessUtil.h
```diff
@@ -12,9 +12,10 @@
 #include <vector>
 #include <logog/include/logog.hpp>
 
-#include "AssemblerLib/LocalDataInitializer.h"
 #include "AssemblerLib/LocalToGlobalIndexMap.h"
 
+#include "Utils/LocalDataInitializer.h"
```
chleh: Suggest to use an absolute path here, i.e. prepend `ProcessLib`. Or what is our rule for that?

chleh: Minor name question: Should we use `Utils` like "utilities" or `Util` like "utility"?

ProcessLib/Utils/InitShapeMatrices.h
```diff
@@ -0,0 +1,52 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ */
+
+#ifndef PROCESSLIB_PROCESSUTIL_H
```
chleh: Wrong header guard :smile:

chleh:
:fast_forward: 

As far as I can remember, you also wanted to
* [ ] change `LocalToGlobalIndexMap` (the `dofSize...()` methods, s.t. it is not necessary anymore to have a pointer to the DOF table in the `MatrixSpecifications`.
* [ ] move `SerialExecutor` to `BaseLib`. Otherwise my extrapolation PR will work against your efforts.

Maybe those changes will entail many small adjustments, so it's OK when you do them in separate PRs. But I find it always hard to see from your individual piece-by-piece PRs, how far your progress is and what you have planned to do until you have finished. So could you please open an issue with a (rather) complete list of what is left to do until the circular dependencies are resolved?

MathLib/LinAlg/SparsityPattern.h
wenqing: Providing exact diagonal and off-diagonal is the best choice for the computational efficiency. However, using d_nz and o_nz solely also do a better job, which has a sort of flexibility.

ProcessLib/Process.h
endJunction: Changed the include list using include-what-you-use tool.

endJunction:
As I have written in the description, there are no changes to the code in this PR.
Where to move the executors is still an open discussion, we didn't make a decision on Monday.

I shall create a list of problems regarding circular dependencies.

chleh:
I admit you cought me out not reading your introductory comment carefully enough :smile: (Although I can vaguely remember your "no code changes" phrase. Anyway, my critisism still holds. :smile:

ProcessLib/Process.h
```diff
@@ -10,34 +10,48 @@
...
 #include "NumLib/ODESolver/NonlinearSolver.h"
 #include "NumLib/ODESolver/ODESystem.h"
 #include "NumLib/ODESolver/TimeDiscretization.h"
+#include "NumLib/ODESolver/Types.h"
 
-namespace MeshLib
+#include "NumericsConfig.h"
+#include "Parameter.h"
+#include "ProcessVariable.h"
+#include "vtkXMLWriter.h"
```
chleh: Did any compilation fail because we didn't have this long list? If not, I'd prefer to keep the list of includes short, because when changing some internals something in `Priocess` now we will always have to check, e.g., if `#include <iterator>` is still needed.
If you want to make something more consistent, I suggest adding the header issue to @norihiro-w's list (#1167) and make some cleanup-days, maybe end of may.

endJunction:
Second attempt for the includes: now absolute minimum strategy.

ProcessLib/Utils/CreateLocalAssemblers.h
```diff
@@ -6,15 +6,16 @@
  *              http://www.opengeosys.org/project/license
  */
 
-#ifndef PROCESSLIB_PROCESSUTIL_H
-#define PROCESSLIB_PROCESSUTIL_H
+#ifndef UTILS_CREATE_LOCAL_ASSEMBLERS_H_
```
norihiro-w: prefix PROCESSLIB_

endJunction:
I had to rebase it, because the extrapolation test was not present. The last three commits do fix the extrapolation test.

norihiro-w:
:+1:

TomFischer:
:+1:

PR #1169, opened 2016-04-28
Fixing several issues with NetCDF-configure dialogue

rinkk:
This would probably require a complete rework but the fixes here will at least fix some pressing issues:
* 2D arrays can be converted into mesh/raster (prev. only 3D)
* the correct scalar array is now transfered onto a mesh even if 0-D/1-D arrays are present
* some dialogue fixes (enabling/disabling widgets depending on scalar array choice)

### DIFF
--- a/Applications/DataExplorer/DataView/NetCdfConfigureDialog.cpp
+++ b/Applications/DataExplorer/DataView/NetCdfConfigureDialog.cpp
@@ -16,23 +16,21 @@
 #include <vtkImageImport.h>
 
 // Constructor
-NetCdfConfigureDialog::NetCdfConfigureDialog(const std::string &fileName, QDialog* parent)
-	: QDialog(parent), _currentFile(new NcFile(fileName.c_str(), NcFile::ReadOnly)),
-	  _currentInitialDateTime(QDateTime()), _currentMesh(NULL), _currentRaster(NULL), _currentPath(fileName)
+NetCdfConfigureDialog::NetCdfConfigureDialog(std::string const& fileName, QDialog* parent)
+: QDialog(parent), _currentFile(new NcFile(fileName.c_str(), NcFile::ReadOnly)),
+  _currentInitialDateTime(QDateTime()), _currentMesh(nullptr), _currentRaster(nullptr), _currentPath(fileName)
 {
 	setupUi(this);
 
-	setVariableSelect(); // set up variables of the file in the combobox
-	comboBoxVariable->setCurrentIndex(valueWithMaxDim()); //pre-select the variable with the biggest number of dimensions...valueWithMaxDim()
-
-	_currentVar = _currentFile->get_var(comboBoxVariable->currentIndex());
+	int const idx = setVariableSelect(); // set up variables of the file in the combobox
+	comboBoxVariable->setCurrentIndex(idx); //pre-select the variable with the biggest number of dimensions...valueWithMaxDim()
+	_currentVar = _currentFile->get_var(idx);
 
 	setDimensionSelect();
 
 	lineEditName->setText(setName());
 
 	this->radioMesh->setChecked(true);
-
 }
 
 NetCdfConfigureDialog::~NetCdfConfigureDialog()
@@ -43,7 +41,7 @@ NetCdfConfigureDialog::~NetCdfConfigureDialog()
 void NetCdfConfigureDialog::accept()
 {
 	QMessageBox valueErrorBox;
-	if (_currentVar->num_dims() < 3){
+	if (_currentVar->num_dims() < 2){
 		valueErrorBox.setText("Selected Variable has not enough dimensions.");
 		valueErrorBox.exec();
 	}else if (doubleSpinBoxDim2Start->value() == doubleSpinBoxDim2Start->maximum()){
@@ -66,19 +64,9 @@ void NetCdfConfigureDialog::reject()
 	this->done(QDialog::Rejected);
 }
 
-int NetCdfConfigureDialog::valueWithMaxDim()
-{
-	int idMaxDim = 0;
-	for (int i=0; i < _currentFile->num_dims(); i++)
-	{
-		if ((_currentFile->get_var(i)->num_dims()) > 0) idMaxDim = i + 1;
-	}
-	return idMaxDim;
-}
-
 void NetCdfConfigureDialog::on_comboBoxVariable_currentIndexChanged(int id)
 {
-	_currentVar = _currentFile->get_var(id);
+	_currentVar = _currentFile->get_var(_id_map[id]);
 	setDimensionSelect();
 }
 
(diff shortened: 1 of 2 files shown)


Applications/DataExplorer/DataView/NetCdfConfigureDialog.cpp
```diff
@@ -43,7 +41,7 @@ NetCdfConfigureDialog::~NetCdfConfigureDialog()
 void NetCdfConfigureDialog::accept()
 {
 	QMessageBox valueErrorBox;
-	if (_currentVar->num_dims() < 3){
+	if (_currentVar->num_dims() < 2){
```
TomFischer: Just out of curiosity: What is `_currentVar` used for and why a return value of `num_dim()` smaller than 2 not sufficient? A suggestion: if you invested some time to understand the class and to solve the issues it would be very nice if you could add some documentation to the class. I hope you will find the time! :smile:

Applications/DataExplorer/DataView/NetCdfConfigureDialog.cpp
```diff
@@ -162,63 +150,67 @@ void NetCdfConfigureDialog::on_comboBoxDim4_currentIndexChanged(int id)
...
-	for (int i=0; i<(_currentFile->num_vars()); i++)
+	std::size_t max_dim(0);
+	std::size_t max_dim_idx(0);
+	std::size_t const n_vars (_currentFile->num_vars());
+	for (int i=0; i<n_vars; i++)
 	{
-		NcVar *focusedVar = _currentFile->get_var(i);
-		if (focusedVar->num_dims() > 0) comboBoxVariable->addItem(focusedVar->name());
+		NcVar const& focusedVar = *_currentFile->get_var(i);
+		std::cout << focusedVar.num_dims() << std::endl;
```
TomFischer: Please use logog here.

Applications/DataExplorer/DataView/NetCdfConfigureDialog.cpp
```diff
@@ -162,63 +150,67 @@ void NetCdfConfigureDialog::on_comboBoxDim4_currentIndexChanged(int id)
...
-		spinBoxDim4->setEnabled(false);spinBoxDim4->setValue(0);
-	}else{  //pre-set dimension selection, typical for 4-d-nc-files:
-		comboBoxDim4->setEnabled(true);
-		spinBoxDim4->setEnabled(true);
-		comboBoxDim1->setCurrentIndex(2);on_comboBoxDim1_currentIndexChanged(2);
-		comboBoxDim2->setCurrentIndex(3);on_comboBoxDim2_currentIndexChanged(3);
-		comboBoxDim3->setCurrentIndex(0);on_comboBoxDim3_currentIndexChanged(0);
-		comboBoxDim4->setCurrentIndex(1);on_comboBoxDim4_currentIndexChanged(1);
+		for (std::size_t j=0; j<_currentVar->num_dims(), j<4; ++j)
+			dim_box[j]->addItem(_currentVar->get_dim(i)->name());
```
TomFischer: I do not understand what is happening here. In the inner loop `j` is required to be less than `_currentVar->num_dims()` but in the access `_currentVar->get_dim(i)` the outer loop-variable `i` is used. Maybe `i` can be greater than `j`. Can this be a potential problem?

TomFischer:
Only minor comments. After answering questions: :fast_forward:

Applications/DataExplorer/DataView/NetCdfConfigureDialog.cpp
chleh: I think `j<_currentVar->num_dims(), j<4` is equivalent to `j<4`.

Applications/DataExplorer/DataView/NetCdfConfigureDialog.cpp
rinkk: this is a debug msg, I'll remove it.

Applications/DataExplorer/DataView/NetCdfConfigureDialog.cpp
TomFischer: `j<_currentVar->num_dims(), j<4` is equivalent to `j<std::min(_currentVar->num_dims(), 4)`.

Assume `dim` is 4 (i.e. the largest `i` is 3) and `_currentVar->getDim()` return a value smaller than `dim`, for instance 2. Then `_currentVar->get_dim(3)` will be possible. I wonder if this will be a problem?

rinkk: each variable in the file has a number of dimension between 0 and 4. we will only import variables with dim>=2. Variables with dim<2 are only scalars, scales, ranges or other support data. the names of the dimensions of each variable have to be inserted into the combobox menus in the dialogue, that is what is happening here. I have 4 comboboxes, not all of which have necessarily associated variables.  So I'm looping over these boxes and fill in the possible variable names for each dimension. The limit for i and j is the same, with a maximum of 4.

rinkk: I will replace ```_currentVar->num_dims()``` for the 2nd loop with ```dim``` so it'll be obvious it's no problem.

TomFischer: Now I understand. Thanks for the explanation. Maybe it is possible to use `j<dim, j<4` in the inner loop instead of `j<_currentVar->num_dim(), j<4` since this was the reason for my confusion.

The code could be easier understandable if you would copy your above explanation to the appropriate place in the code.

TomFischer:
I know it is not your fault, but could you add the OGS license header please.

Applications/DataExplorer/DataView/NetCdfConfigureDialog.cpp
rinkk: NetCDF can store multiple variables of varying dimensions, e.g. x=3 has dim=0, but I can also store time- or space ranges (dim = 1), rasters (dim=2), rasters over time or 3-d rasters (dim=3), etc. If you have a netcdf file containing climate data you can easily have 100 or 200 variables.
These variables are interconnected, e.g. a geo-referenced raster will have its extend in x and y are stored as arrays (so you need not necessarily have equidistant or isotropic pixels).
_currentVar is the variable that is currently selected in the dialogue and which all other input in the dialogue depends on. We ignore selectable variables with dim<2 because we cannot convert those into raster- or mesh objects.

TomFischer: Thank you very much! I think it would be extremly helpful if you could transform the explanation to the code somehow.

rinkk: added a comment

rinkk:
can this be merged?

TomFischer:
:+1:

PR #1171, opened 2016-04-28
Output secondary variables

chleh:
This PR integrates the extrapolator class into the processes.

Changes:
* new classes handling output, especially of 2ndary vars. They configure which data to write to file.
* the output scheme is suitable both for extrapolation of integration point values as well as computation of secondary variables directly on the mesh nodes.
* GWFlow now computes the Darcy velocity, solely for output purposes, as an example.

Problems:
* I think `std::vector<std::vector<double>> _darcy_velocities` is bad. One could do better if `getIntegrationPointValues()` would return a reference to some Eigen type.
* ~~The ctor of GWFlowProcess now demands less encapsulation, breaks our current init scheme and is somewhat fishy.~~ Fixed.

The new `<process>` subtree in the input files looks like:
```
        <process>
            <name>GW23</name>
            <type>GROUNDWATER_FLOW</type>

            <hydraulic_conductivity>K</hydraulic_conductivity>

            <nonlinear_solver>basic_picard</nonlinear_solver>
            <time_discretization><type>BackwardEuler</type></time_discretization>

            <process_variables>
                <process_variable>pressure</process_variable>
            </process_variables>
            <secondary_variables>
                <darcy_velocity_x>v_x</darcy_velocity_x>
            </secondary_variables>

            <output>
                <variables>
                    <variable>pressure</variable>
                    <variable>v_x</variable>
                </variables>
                <output_extrapolation_residuals>true</output_extrapolation_residuals>
            </output>
        </process>
```

### DIFF
--- a/ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
+++ b/ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
@@ -26,14 +26,27 @@ namespace ProcessLib
 namespace GroundwaterFlow
 {
 
+enum class IntegrationPointValue {
+    DarcyVelocityX,
+    DarcyVelocityY,
+    DarcyVelocityZ
+};
+
 const unsigned NUM_NODAL_DOF = 1;
+
+template <typename GlobalMatrix, typename GlobalVector>
+class GroundwaterFlowLocalAssemblerInterface
+        : public ProcessLib::LocalAssemblerInterface<GlobalMatrix, GlobalVector>
+        , public NumLib::Extrapolatable<GlobalVector, IntegrationPointValue>
+{};
 
 template <typename ShapeFunction,
          typename IntegrationMethod,
          typename GlobalMatrix,
          typename GlobalVector,
          unsigned GlobalDim>
-class LocalAssemblerData : public ProcessLib::LocalAssemblerInterface<GlobalMatrix, GlobalVector>
+class LocalAssemblerData
+        : public GroundwaterFlowLocalAssemblerInterface<GlobalMatrix, GlobalVector>
 {
     using ShapeMatricesType = ShapeMatrixPolicyType<ShapeFunction, GlobalDim>;
     using ShapeMatrices = typename ShapeMatricesType::ShapeMatrices;
@@ -64,7 +77,7 @@ public:
         assert(local_matrix_size == ShapeFunction::NPOINTS * NUM_NODAL_DOF);
     }
 
-    void assemble(double const /*t*/, std::vector<double> const& /*local_x*/) override
+    void assemble(double const /*t*/, std::vector<double> const& local_x) override
     {
         _localA.setZero();
         _localRhs.setZero();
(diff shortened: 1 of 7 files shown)


ProcessLib/Process.h
```diff
@@ -27,9 +27,11 @@
 #include "NeumannBc.h"
 #include "NeumannBcAssembler.h"
 #include "Parameter.h"
+#include "ProcessOutput.h"
 #include "ProcessVariable.h"
 #include "UniformDirichletBoundaryCondition.h"
 
+#include "NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.h"
```
chleh: TODO remove.

ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
```diff
@@ -64,21 +77,31 @@ class LocalAssemblerData : public ProcessLib::LocalAssemblerInterface<GlobalMatr
...
             auto const& sm = _shape_matrices[ip];
             auto const& wp = integration_method.getWeightedPoint(ip);
-
             auto const k = _process_data.hydraulic_conductivity(_element);
+
             _localA.noalias() += sm.dNdx.transpose() * k * sm.dNdx *
                                  sm.detJ * wp.getWeight();
+
+            // Darcy velocity only computed for output.
+            auto const darcy_velocity = (k * sm.dNdx *
```
wenqing: Better to move the computation of the velocity to another function, and also call this function in a standalone function outside of the global assembly function. If we deal with non-linear problems, we only need to compute the velocity after a linearization step is converged.

ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
```diff
@@ -99,6 +150,10 @@ class LocalAssemblerData : public ProcessLib::LocalAssemblerInterface<GlobalMatr
     NodalVectorType _localRhs;
 
     unsigned const _integration_order;
+
+    std::vector<std::vector<double>> _darcy_velocities
```
wenqing: vector in vector could be slower than an array, especially it is used at the element level.

ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
chleh: > If we deal with non-linear problems, we only need to compute the velocity after a linearization step is converged. 

That's an interesting fact, thanks. I didn't know it.

Regarding the GWFlow files in this PR: Any changes made are primarily for illustrative purposes. Since the velocity is not needed in GWFlow it probably should not be computed in production code at all. Besides, the current way of computing it produces completely meaningless results, since the velocity is computed before solving the equation system.

I think it doesn't make much sense to extend GWFlow by a `postTimestep()` hook that computes the velocity which is needed only for output. I suggest that I keep the GWFlow changes during the review, but will remove it before merging. This way you'll see how secondary variables can be used and the current code does not get cluttered in the end.

wenqing: The velocity is not only computed for output, it also has to be computed for the governing equations for heat or mass transport. Therefore, the computation can not be in postTimestep().

chleh: Maybe then I misunderstood you. But anyway, I think we should discuss that after this PR, shouldn't we? And 'll remove the offending code before merge.

ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
TomFischer: Is it possible to use `std::array<std::array<double, ShapeFunction::NPOINTS>, GlobalDim>`?

TomFischer:
After resolving merge conflicts :+1:

chleh:
May I squash and merge this PR? (provided the tests work)

endJunction:
:fast_forward:

PR #1177, opened 2016-04-29
one dimensional root finding

chleh:
This PR adds a class implementing the regula falsi method for one-dimensional root finding.

The idea is to use this class in order to become a rough estimate of a root in only very few iterations with only very few calls of the user-provided function. Additionally, no tangents are needed.

Hence, this method should be suitable for being used inside local assemblers.

### DIFF
--- a/MathLib/Nonlinear/Root1D.h
+++ b/MathLib/Nonlinear/Root1D.h
@@ -1,0 +1,165 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef MATHLIB_NONLINEAR_ROOT1D_H
+#define MATHLIB_NONLINEAR_ROOT1D_H
+
+#include <cassert>
+#include <cmath>
+#include <limits>
+#include <type_traits>
+#include <logog/include/logog.hpp>
+
+namespace MathLib
+{
+namespace Nonlinear
+{
+
+namespace detail
+{
+
+//! Tells if \c a and \c b have the same sign.
+inline bool same_sign(double a, double b)
+{
+    // note: signbit() casts integers to double and distinguishes between +0.0 and
+    // -0.0. However, the latter is not a problem, because if f(x0) == 0, we've
+    // already found the root x0 that we are searching for.
+    return std::signbit(a) == std::signbit(b);
+}
+
+inline bool almost_zero(double a)
+{
+    return std::abs(a) <= std::numeric_limits<double>::epsilon();
+}
+
+}
+
+
+//! Use the regula falsi method to find the root of some scalar function of one
+//! variable.
+template<typename SubType, typename Function>
+class RegulaFalsi
+{
+public:
+    //! Initializes finding a root of the \c Function \c f in the interval
+    //! [\c a, \c b].
+    RegulaFalsi(Function const& f, double a, double b)
+        : _f(f), _a(a), _b(b), _fa(f(a)), _fb(f(b))
+    {
+        static_assert(
+                std::is_same<double, decltype(f(0.0))>::value,
+                "Using this class for functions that do not return double"
+                " involves a lot of casts. Hence it is disabled.");
+
+        if (detail::almost_zero(_fa)) {
+            _b = _a;
+        } else if (detail::almost_zero(_fb)) {
+            _a = _b;
+        } else if (detail::same_sign(_fa, _fb)) {
+            ERR("Regula falsi cannot be done, because the function values"
+                " at the interval ends have the same sign.");
+            std::abort();
+        }
+    }
+
+    //! Do \c num_steps iteration of regula falsi.
+    void step(const unsigned num_steps)
+    {
+        for (unsigned i=0; i<num_steps; ++i)
+        {
+            if (_a == _b) return;
+
+            const double s = (_fb - _fa)/(_b - _a);
+            const double c = _a - _fa/s;
+            const double fc = _f(c);
+
+            if (detail::almost_zero(fc)) {
+                _a = _b = c;
+                return;
+            } else if (!detail::same_sign(fc, _fb)) {
+                _a = _b; _fa = _fb;
+                _b =  c; _fb =  fc;
+            } else {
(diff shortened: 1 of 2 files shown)


endJunction:
I'd strongly recommend to remove all output from tests.

Tests/MathLib/TestNonlinear1D.cpp
```diff
@@ -0,0 +1,63 @@
+#include <gtest/gtest.h>
+
+#include <cstdio>
+
+#include "MathLib/Nonlinear/Root1D.h"
+
+namespace
+{
+
+double f(double x)
```
endJunction: Unused so far.

Tests/MathLib/TestNonlinear1D.cpp
```diff
@@ -0,0 +1,63 @@
...
+};
+
+
+using namespace MathLib::Nonlinear;
+typedef ::testing::Types<Unmodified, Illinois, Pegasus, AndersonBjorck> RegulaFalsiTypes;
+TYPED_TEST_CASE(RegulaFalsiTest, RegulaFalsiTypes);
+
+
+
+TYPED_TEST(RegulaFalsiTest, QuadraticFunction)
```
endJunction: Missing prefix for MathLib. Otherwise inconvenient to test everything from MathLib

MathLib/Nonlinear/Root1D.h
```diff
@@ -0,0 +1,174 @@
...
+namespace Nonlinear
+{
+
+namespace detail
+{
+
+inline
+bool same_sign(const double a, const double b)
+{
+    return (a < 0.0) == (b < 0.0);
```
endJunction: Could also be using the [signbit](http://en.cppreference.com/w/cpp/numeric/math/signbit):
return std::signbit(a) == std::signbit(b);

endJunction:
Also be aware, that if Function f returns a non-double, there will be conversion. I'd prefer to have that parameter as a template.

MathLib/Nonlinear/Root1D.h
chleh: http://en.cppreference.com/w/cpp/numeric/math/signbit
looks a bit esoteric in some cases. So I'd rather keep the comparison to zero.

endJunction: So if a = -0 and b = 1, same_sign(-0, 1) = false, that's exactly the signbit(a) == signbit(b)...

chleh:
I now restrict `f` to functions returning double via a static assert.

chleh:
I also squashed it because of many changes.

MathLib/Nonlinear/Root1D.h
```diff
@@ -0,0 +1,156 @@
...
+template<typename SubType, typename Function>
+RegulaFalsi<SubType, Function>
+makeRegulaFalsi(Function const& f, double const a, double const b)
+{
+    return RegulaFalsi<SubType, Function>(f, a, b);
+}
+
+
+//! Used by \c RegulaFalsi in the original regula falsi algorithm.
+struct Unmodified
```
endJunction: Maybe those subtypes can go into some scope more local to RegulaFalsi?
I'd suggest RegulaFalsi namespace.

chleh: Maybe instead of `Nonlinear`? The types are already in a pretty nested namespace. Furthermore the other members of MathLib/Nonlinear look pretty obsolete.

endJunction: My fault. Didn't see the Nonlinear namespace.

MathLib/Nonlinear/Root1D.h
```diff
@@ -0,0 +1,156 @@
...
+    //! [\c a, \c b].
+    RegulaFalsi(Function const& f, double a, double b)
+        : _f(f), _a(a), _b(b), _fa(f(a)), _fb(f(b))
+    {
+        static_assert(
+                std::is_same<double, decltype(f(0.0))>::value,
+                "Using this class for functions that do not return double"
+                " involves a lot of casts. Hence it is disabled.");
+
+        if (_fa == 0.0) {
```
norihiro-w: why not use epsilon here?

MathLib/Nonlinear/Root1D.h
```diff
@@ -0,0 +1,156 @@
...
+            _b = _a;
+        } else if (_fb == 0.0) {
+            _a = _b;
+        } else {
+            assert(!detail::same_sign(_fa, _fb));
```
norihiro-w: What do you intend here with assert? Is there no need to check it in release mode?

MathLib/Nonlinear/Root1D.h
```diff
@@ -0,0 +1,156 @@
...
+
+    //! Do \c num_steps iteration of regula falsi.
+    void step(const unsigned num_steps)
+    {
+        for (unsigned i=0; i<num_steps; ++i)
+        {
+            if (_a == _b) return;
+
+            const double s = (_fb - _fa)/(_b - _a);
+            const double c = _a - _fa/s;
```
norihiro-w: check if s is zero

MathLib/Nonlinear/Root1D.h
```diff
@@ -0,0 +1,156 @@
...
+        }
+    }
+
+    //! Returns the current estimate of the root.
+    double getResult() const
+    {
+        if (_a == _b) return _a;
+
+        const double s = (_fb - _fa)/(_b - _a);
+        const double c = _a - _fa/s;
```
wenqing: Here the same for s=0.

MathLib/Nonlinear/Root1D.h
chleh: `s` cannot be zero here since `_fa` and `_fb` have oppisite sign.

MathLib/Nonlinear/Root1D.h
chleh: Same answer as above.

MathLib/Nonlinear/Root1D.h
chleh: If `_fa` and `_fb` have the same sign, it is not guaranteed that there is a root in [a,b].
Will enable it in release mode, too.

MathLib/Nonlinear/Root1D.h
chleh: I can replace such checks in all of this class.

wenqing:
:+1:

TomFischer:
:+1:

PR #1179, opened 2016-04-30
Removed custom pow() implementation

chleh:
The runtime pow() is slow (i.e. it does not do exponentiation by squaring). I think the template-meta-POW would only work for integral bases and exponents.

Both things are used nowhere.


endJunction:
If not used, remove!
:fast_forward:

TomFischer:
:+1:

chleh:
Travis does not know conan?
`scripts/travis/conan.sh: line 5: conan: command not found`

PR #1193, opened 2016-05-12
[NL] fixed error computation

### LINKED ISSUE #1180: Nonlinear solver bug
The Picard method currently fails. I could locate the problem: It is caused by not calling `makeCompressed()` on sparse Eigen matrices after they have been assembled (at least that is my current guess).
This problem can be fixed by adding the following overload to BLAS.cpp/.h:
```
void finalizeAssembly(EigenMatrix& x)
{
    x.getRawMatrix().makeCompressed();
}
```
However, the issue needs a bit of further investigation, which I cannot do at the moment.
Maybe this very problem is what made Yonghui considerable headache last week.

chleh:
Picard iteration convergence is dramatically improved when applying Dirichlet BCs before computing errors but after solving the linearized system. I guess, the same holds for Newton.
I hope that will finally fix #1180.

### DIFF
--- a/NumLib/ODESolver/NonlinearSolver-impl.h
+++ b/NumLib/ODESolver/NonlinearSolver-impl.h
@@ -70,6 +70,8 @@ solve(Vector &x)
         }
         else
         {
+            sys.applyKnownSolutions(x_new);
+
             switch(sys.postIteration(x_new))
             {
             case IterationResult::SUCCESS:
@@ -193,6 +195,7 @@ solve(Vector &x)
             auto& x_new =
                     MathLib::GlobalVectorProvider<Vector>::provider.getVector(x, _x_new_id);
             BLAS::axpy(x_new, -_alpha, minus_delta_x);
+            sys.applyKnownSolutions(x_new);
 
             switch(sys.postIteration(x_new))
             {


norihiro-w:
@chleh Can you give us more explanations? I don't get what's happening in the implementations. In general, there is no need to apply Dirichlet BCs before computing errors in Picard. One should apply the BCs to A, b, and x before solving Ax=b. Calculated x should keep the BC values. 

In Newton-Raphson, one should apply the BCs to x before assembly of r and J.  then dx=0 for the prescribed values should be applied to J, dx, and r before solving J dx = -r. Again, updated x should already include the prescribed values.

chleh:
That was an ad-hoc fix I did in the last days. I only observed that my TES process did not converge when undoing the changes in this PR, but I didn't dig any further into this. I can examine it better when I will be back at UFZ.
The only thing that comes to my mind: Do we apply Dirichlet BCs to also x before the linear solver in the case of EigenLIS (and also plain Eigen)? Maybe we don't and therefore the linear solver fails to exactly reproduce the BC values. But that's only a guess. Sorry for the lame excuse.

chleh:
Obviously this PR is not needed.

PR #1194, opened 2016-05-13
[GL] Changes in GeoLib needed for GeoMapper Refactorization.

TomFischer:
See commit messages for details.

### DIFF
--- a/GeoLib/AnalyticalGeometry.cpp
+++ b/GeoLib/AnalyticalGeometry.cpp
@@ -568,10 +568,14 @@ GeoLib::Polygon rotatePolygonToXY(GeoLib::Polygon const& polygon_in,
     return GeoLib::Polygon(rot_polyline);
 }
 
-std::vector<MathLib::Point3d>
-lineSegmentIntersect2d(MathLib::Point3d const& a, MathLib::Point3d const& b,
-    MathLib::Point3d const& c, MathLib::Point3d const& d)
+std::vector<MathLib::Point3d> lineSegmentIntersect2d(
+    GeoLib::LineSegment const& ab, GeoLib::LineSegment const& cd)
 {
+    GeoLib::Point const& a{ab.getBeginPoint()};
+    GeoLib::Point const& b{ab.getEndPoint()};
+    GeoLib::Point const& c{cd.getBeginPoint()};
+    GeoLib::Point const& d{cd.getEndPoint()};
+
     double const orient_abc(ExactPredicates::getOrientation2d(a, b, c));
     double const orient_abd(ExactPredicates::getOrientation2d(a, b, d));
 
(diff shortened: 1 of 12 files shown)


GeoLib/AnalyticalGeometry.cpp
```diff
@@ -583,12 +587,85 @@ lineSegmentIntersect2d(MathLib::Point3d const& a, MathLib::Point3d const& b,
...
+						// check if b in (cd)
+						double const t2((b[0]-c[0])/(d[0]-c[0]));
+						if (0 <= t2 && t2 <= 1) {
+							return {{b, c}};
+						} else {
+							// check d in (ab)
+							double const t3((d[0]-a[0])/(b[0]-a[0]));
+							if (0 <= t3 && t3 <= 1) {
+								return {{c, d}};
+							} else {
```
endJunction: After a return there is no need for the else; this can reduce the depth.

GeoLib/AnalyticalGeometry.h
```diff
@@ -384,6 +384,12 @@ void computeAndInsertAllIntersectionPoints(GeoLib::PointVec &pnt_vec,
 GeoLib::Polygon rotatePolygonToXY(GeoLib::Polygon const& polygon_in,
 	MathLib::Vector3 & plane_normal);
 
+/// Sorts the vector of segments such that the \f$i\f$-th segment is connected
+/// with the \f$i+1\f$st segment, i.e. the end point of the \f$i\f$-th segment
+/// is the start point of the \f$i+1\f$st segment.
+void sortSegments(MathLib::Point3d const& seg_beg_pnt,
```
norihiro-w: Two questions. maybe you can answer them in the documentation
- why do you need to pass `seg_beg_pnt`? I thought you can automatically determine the beginning point from connectivity of segments.
- do you consider a case when some segments are not connected?

GeoLib/Polyline.h
```diff
@@ -113,16 +113,18 @@ class Polyline : public GeoObject
 	/**
 	 * Adds an id of a point at the end of the polyline. The id have to be inside
 	 * the (internal) _ply_pnts vector the polyline is based on.
+	 * @return true if the point could be added, else false
```
norihiro-w: can you explain in the document that when the point addition could fail?

norihiro-w:
@TomFischer can you please add tests for `sortSegments()` and `lineSegmentIntersect2d()`?

TomFischer:
> @TomFischer can you please add tests for sortSegments() and lineSegmentIntersect2d()?

For the function `lineSegmentIntersect2d()` there exists already a test. For `sortSegments()` I will provide a test soon.

bilke:
Jenkins test this please.

TomFischer:
The test for `sortSegments()` runs.

norihiro-w:
:+1:

PR #1196, opened 2016-05-13
Partitioning tool

wenqing:
A utility to process mesh partition data (currently from METIS) for ogs6


wenqing:
Please do not review it.

PR #1197, opened 2016-05-15
fix FindCVODE.cmake 

norihiro-w:
This fixes linking errors with sundias libraries


endJunction:
My fault :)
After introducing the `CVODE_LIBRARY / _NVECSERIAL`  forgot to use them...

PR #1201, opened 2016-05-16
Tabs to whitespaces

endJunction:
As discussed about one week ago here is a PR changing tabs to 4 whitespaces almost everywhere. The configs are updated too; add `?ws=1` to the address line when viewing the diff like here: [diff ignoring whitespace changes](https://github.com/ufz/ogs/pull/1201/files?w=1).

Almost all of your branches will have merge/rebase conflicts. To resolve those as painless as possible there is a small script and some command with explanation, but before you make any changes make a backup. Like real backup, not a shallow clone, because if you destroy you .git/* the behaviour is undefined. Don't ask me how I know.

Plan (which didn't work like expected. See another plan in the comments.):
for any branch:

1. First rebase to the current ufz/ogs/master (one before this merge).
2. Replace all tabs in all modified/added files in all commits to whitespaces using `git filter-branch`
3. Rebase onto the merged PR ignoring whitespace changes.

You will need a small script named `t2ws.sh` from this [gist](https://gist.github.com/endJunction/bd1e8bade148b5d6e25dbb9d46bc671f)

Let's assume you have a branch **A** somewhere. The current master is **M**. The merged TabsToWhitespaces PR is **T**:
```
T
| \
|  * replace tabs.
|  .
| /
M
|   A
| /
M'
```

1. `git rebase M A` (this is your usual business). The new branch is called **A'**:
```
T
| \
|  * replace tabs.
|  .   A'
| /  /
M----
```
2. Rewrite history for the new branch **A'**: `git filter-branch -f --tree-filter "git diff --name-only --diff-filter=AM ${GIT_COMMIT}^ | xargs t2ws.sh" M..A'` Let's call the rewritten history of **A'** as **A''**:
```
T
| \
|  * replace tabs.
|  .  A'  A''
| /  /   /
M--------
```
3. `git rebase -Xignore-all-space T A''` yields the following graph ignoring the whitespace changes:

```
   A'''
  /
T
| \
|  * replace tabs.
|  .  A'  A''
| /  /   /
M--------
```

The filter-branch works like this: the --tree-filter checks out every commit (specified by the range **M..A** above) and runs a given command. The `git diff --name-only --diff-filter=AM ${GIT_COMMIT}^` print file names which were added/modified since previous commit, the `${GIT_COMMIT}^`. (Try it at any of your branches to see the output. The command is harmless.) This files are then processed by the script, which ignores empty input and non-regular-file input (like directories _e.g._ `Tests/Data`). Then filter-branch's tree-filter makes a new commit and goes to the next one.

Don't forget to backup!

**Addendum:** Sometimes (when skipping the first step for example), after the step 3 there could be new tabs in the code. One can check for this with `git diff --name-only --diff-filter=AM ufz/master | xargs grep '\t'` or something like this.


norihiro-w:
:+1:

TomFischer:
:fast_forward:

endJunction:
I must admit, the proposed plan is not quite working on non-simplistic branches.

endJunction:
More (computationally) expensive, but working for me plan. (A, M, T branches as above)

1. Rebase **A** to **M** yielding **A1**. This resolves the non-tab_to_whitespace changes:```git rebase M A```
2. Rebase **A1** to **T**  ignoring whitespace changes yielding **A2**. Here you _might_ still need to resolve some of the conflicts because of the whitespaces. ``` git rebase -Xignore-all-space T A2```
3. Cleanup the whitespace introduced in **A2** yielding **A3**: ```git filter-branch -f --tree-filter t2ws.sh T..A2```
4. Check the difference of **A3** to **A2**, which should be only in whitespaces (so empty output) with ```git diff --ignore-all-space A2```

Questions and improvements are welcome.

endJunction:
Tom has reported a successful result using the second approach for the updating of the branches.

PR #1208, opened 2016-05-21
Fix circular dependency between modules

norihiro-w:
fix circular dependency between 
- MathLib and AssemblerLib
- AssemblerLib and NumLib
- NumLib and ProcessLib
- MeshLib and InsituLib

see current dependency graph https://doxygen.opengeosys.org/dir_6ccf91d5adff89be0d7026e347471e07.html

To fix them, I did the followings and removed AssemblerLib and InsituLib:
- merge AssemblerLib to NumLib (DOF and Assembler dirs)
- move DOF related stuff in MathLib/LinAlg to NumLib/DOF
- move NumericsConfig.h and GlobalSetup.h in ProcessLib to NumLib
- move InsituLib to MeshLib/Vtk


norihiro-w:
and this is the new dependency graph
module

Applications/DataExplorer/DataView/ElementTreeModel.cpp
```diff
@@ -21,7 +21,7 @@
 
 #include "GeoLib/AABB.h"
 
-#include "InSituLib/VtkMappedMeshSource.h"
+#include "MeshLib/vtk/VtkMappedMeshSource.h"
```
bilke: `vtk` should be upper-case: `Vtk`

bilke:
👍 for InSituLib-move.

Applications/DataExplorer/DataView/ElementTreeModel.cpp
chleh: Why not all uppercase `VTK`?

MathLib/LinAlg/SparsityPattern.h
```diff
@@ -12,12 +12,11 @@
...
 #include <vector>
 
-#include "ProcessLib/NumericsConfig.h"
-
 namespace MathLib
 {
 /// A vector telling how many nonzeros there are in each global matrix row.
-using SparsityPattern = std::vector<GlobalIndexType>;
+template <typename IndexType>
+using SparsityPattern = std::vector<IndexType>;
```
chleh: I'm not immediately 100% sure if that should be a templated type alias...

NumLib/NumericsConfig.h
```diff
@@ -12,6 +12,8 @@
 
 #include <type_traits>
 
+#include "MathLib/LinAlg/SparsityPattern.h"
```
chleh: Why not move it to MathLib alongside with GlobalSetup? The two files only reference MathLib stuff, or am I wrong?

ProcessLib/DirichletBc.h
```diff
@@ -12,19 +12,15 @@
...
 
 /// A dirichlet boundary condition is represented by a list of global indices
 /// with corresponding values.
 template <typename IndexType>
-struct DirichletBc final
-{
-    std::vector<IndexType> global_ids;
-    std::vector<double> values;
-};
+using DirichletBc = NumLib::IndexValueVector<IndexType>;
```
chleh: Is it worth keeping the file only for one typedef?

NumLib/ODESolver/NonlinearSolver-impl.h
```diff
@@ -14,11 +14,11 @@
 
 #include "BaseLib/ConfigTree.h"
 #include "MathLib/LinAlg/BLAS.h"
-#include "AssemblerLib/GlobalMatrixProviders.h"
+#include "MathLib/LinAlg/VectorNorms.h"
```
chleh: Is that needed?

NumLib/IndexValueVector.h
```diff
@@ -0,0 +1,27 @@
...
+#ifndef NUM_LIB_INDEXVALUEVECTOR_H
+#define NUM_LIB_INDEXVALUEVECTOR_H
+
+#include <vector>
+
+namespace NumLib
+{
+
+template <typename IndexType>
+struct IndexValueVector final
```
chleh: I wouldn't call it vector, since it's two vectors, but I don't have a better name for it right now.

NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.h
```diff
@@ -10,7 +10,7 @@
 #ifndef NUMLIB_LOCAL_LLSQ_EXTRAPOLATOR_H
 #define NUMLIB_LOCAL_LLSQ_EXTRAPOLATOR_H
 
-#include "AssemblerLib/GlobalMatrixProviders.h"
+#include "NumLib/DOF/GlobalMatrixProviders.h"
```
chleh: I'd leave it in `MathLib`, since it only copes with Mathlib stuff. Or is that impossible?
In particular I wouldn't put it into the DOF directory, because it merely provides matrices and vectors.

Applications/ApplicationsLib/LinearSolverLibrarySetup.h
```diff
@@ -19,7 +19,7 @@
 /// The default implementation is empty providing polymorphic behaviour when
 /// using this class.
 
-#include "MathLib/LinAlg/GlobalMatrixProviders.h"
+#include "AssemblerLib/GlobalMatrixProviders.h"
```
chleh: This issue can be fixed IMHO in a better way by consistently naming the member functions of the DOF table. Then the DOF related methods of the matrix and vector providers would only get integer arguments and not the whole DOF table.

chleh:
I think there should be a different "fix" for the matrix/vector related issues. I don't like to have them in AssemblerLib. Then everything you do with matrices and vectors will pull in AssemblerLib. I already discussed that with @endJunction some weeks ago.
But I'm mostly OK with your other changes.

Applications/DataExplorer/DataView/ElementTreeModel.cpp
norihiro-w: i followed MeshLib/IO/VtkIO

norihiro-w:
@chleh There is no AssemblerLib anymore and they are moved to NumLib/DOF and NumLib/Assembler. 
- NumLib/DOF includes DOF stuff and mat/vec files depending on DOF (moved from MathLib/LinAlg)
- NumLib/Assembler includes Executor.

Some functions in mat/vec files which are independent of DOF can be extracted from the files and can be moved back to MathLib/LinAlg. However the refactoring will not be a part of this PR.

chleh:
OK, I can do the moving back to MathLib later on. I think the mat/vec code really belongs there.

norihiro-w:
I will rework on this once Christoph's branch is merged, but comments are always welcome

NumLib/NumericsConfig.h
norihiro-w: Basically I would like to keep a module as general as possible. It's hard to do it for NumLib but MathLib is still possible with this way. My understanding is that these configuration are specifying how we use or specialize MathLib for our purpose, i.e. ogs6.

ProcessLib/DirichletBc.h
norihiro-w: I personally don't care it. If you don't like it, may I ask you to do it in another PR?

NumLib/IndexValueVector.h
norihiro-w: me neither

norihiro-w:
rebased

NumLib/IndexValueVector.h
endJunction: Just an idea: `KnownSolutions` or `KnownSolutionsVector` -- plays well with `applyKnownSolution`.

NumLib/NumericsConfig.h
chleh: OK, let's discuss that after this PR.

chleh:
:+1:

NumLib/IndexValueVector.h
norihiro-w: I think this data type can also be used for Neumann or other conditions. In that case, the values are not known solutions any more.

PR #1210, opened 2016-05-24
set false to AllowShortIfStatementsOnASingleLine in clang-format 

norihiro-w:
I would like to suggest setting `AllowShortIfStatementsOnASingleLine: false` in clang-format. Without this, 
```
    if (!someVar) {
        return;
    }
```
will be changed to 
```
    if (!someVar) return;
```

However I intentionally use the former because I can set a breakpoint to the line `return;`.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -3,6 +3,7 @@ Language:        Cpp
 BasedOnStyle:    Google
 
 AccessModifierOffset: -4
+AllowShortIfStatementsOnASingleLine: false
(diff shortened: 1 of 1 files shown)


endJunction:
:fast_forward: 
Maybe you could check if this is default Google style, then the line can be removed (with idea that only differences to the Google style are visible).

TomFischer:
:+1:

norihiro-w:
@endJunction the option is set true in the Google style.

PR #1211, opened 2016-05-25
Small fixes

norihiro-w:
changes
- include string in LogogSetup
- automatically detect source files in FileIO as we do in other lib
- git ignore CMakeLists.txt.user used in QtCreator

### DIFF
--- a/FileIO/CMakeLists.txt
+++ b/FileIO/CMakeLists.txt
@@ -1,17 +1,10 @@
-set(SOURCES
-    CsvInterface.h
-    CsvInterface.cpp
-    GMSInterface.h
-    GMSInterface.cpp
-    GMSHInterface.h
-    GMSHInterface.cpp
-    PetrelInterface.h
-    PetrelInterface.cpp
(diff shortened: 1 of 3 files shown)


FileIO/CMakeLists.txt
```diff
@@ -1,17 +1,10 @@
...
-    PetrelInterface.cpp
-    TetGenInterface.h
-    TetGenInterface.cpp
-)
-if(Shapelib_FOUND)
-    list(APPEND SOURCES SHPInterface.h SHPInterface.cpp)
+GET_SOURCE_FILES(SOURCES_FileIO)
+set(SOURCES ${SOURCES_FileIO})
+if(NOT Shapelib_FOUND)
+    list(REMOVE_ITEM SOURCES SHPInterface.h SHPInterface.cpp)
```
chleh: Would it be a better solution to just #ifdef out everything in these two files if shapelib is not found rather than removing the files here?

chleh:
:+1:

PR #1214, opened 2016-05-25
Extend local asm interface

endJunction:
This adds missing functions to the interface for the jacobian assembly and the pre/post timestep operations.

### DIFF
--- a/AssemblerLib/VectorMatrixAssembler.h
+++ b/AssemblerLib/VectorMatrixAssembler.h
@@ -136,16 +136,20 @@ public:
     /// \attention The index \c id is not necessarily the mesh item's id.
     void preTimestep(std::size_t const id,
                      LocalAssembler& local_assembler,
-                     GlobalVector const& x) const
+                     GlobalVector const& x,
+                     double const t,
+                     double const delta_t) const
     {
         auto cb = [&local_assembler](
             std::vector<double> const& local_x,
-            LocalToGlobalIndexMap::RowColumnIndices const& /*r_c_indices*/)
+            LocalToGlobalIndexMap::RowColumnIndices const& /*r_c_indices*/,
+            double const t,
+            double const delta_t)
         {
-            local_assembler.preTimestep(local_x);
+            local_assembler.preTimestep(local_x, t, delta_t);
         };
 
-        passLocalVector_(cb, id, _data_pos, x);
+        passLocalVector_(cb, id, _data_pos, x, t, delta_t);
     }
 
     /// Executes the postTimestep() method of the local assembler for the
--- a/ProcessLib/LocalAssemblerInterface.h
+++ b/ProcessLib/LocalAssemblerInterface.h
@@ -26,8 +26,34 @@ public:
 
     virtual void assemble(double const t, std::vector<double> const& local_x) = 0;
 
-    virtual void addToGlobal(AssemblerLib::LocalToGlobalIndexMap::RowColumnIndices const&,
-            GlobalMatrix& M, GlobalMatrix& K, GlobalVector& b) const = 0;
+    virtual void addToGlobal(
+        AssemblerLib::LocalToGlobalIndexMap::RowColumnIndices const&,
+        GlobalMatrix& M, GlobalMatrix& K, GlobalVector& b) const = 0;
+
+    virtual void assembleJacobian(double const /*t*/,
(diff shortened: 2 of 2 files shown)


ProcessLib/LocalAssemblerInterface.h
```diff
@@ -26,8 +26,20 @@ class LocalAssemblerInterface
...
     virtual void assemble(double const t, std::vector<double> const& local_x) = 0;
 
-    virtual void addToGlobal(AssemblerLib::LocalToGlobalIndexMap::RowColumnIndices const&,
-            GlobalMatrix& M, GlobalMatrix& K, GlobalVector& b) const = 0;
+    virtual void assembleJacobian(double const t,
+                                  std::vector<double> const& local_x) = 0;
+
+    virtual void preTimestep(std::vector<double> const& local_x) = 0;
+    virtual void postTimestep(std::vector<double> const& local_x,
+                              double const t) = 0;
```
chleh: IMHO implementing these methods are optional. I'd appreciate empty default implementations.

ProcessLib/LocalAssemblerInterface.h
```diff
@@ -26,8 +26,20 @@ class LocalAssemblerInterface
...
+
+    virtual void addToGlobal(
+        AssemblerLib::LocalToGlobalIndexMap::RowColumnIndices const&,
+        GlobalMatrix& M, GlobalMatrix& K, GlobalVector& b) const = 0;
+
+    virtual void addJacobianToGlobal(
+        AssemblerLib::LocalToGlobalIndexMap::RowColumnIndices const& indices,
+        GlobalMatrix& Jac) const = 0;
```
chleh: Likewise.

ProcessLib/LocalAssemblerInterface.h
```diff
@@ -26,8 +26,20 @@ class LocalAssemblerInterface
 
     virtual void assemble(double const t, std::vector<double> const& local_x) = 0;
 
-    virtual void addToGlobal(AssemblerLib::LocalToGlobalIndexMap::RowColumnIndices const&,
-            GlobalMatrix& M, GlobalMatrix& K, GlobalVector& b) const = 0;
+    virtual void assembleJacobian(double const t,
+                                  std::vector<double> const& local_x) = 0;
```
chleh: Also optional.

chleh:
In general :+1:
Maybe we have to discuss whether to use empty default implementations or not.
Maybe for assembleJacobian/addJacobianToGlobal a failing default implementation would be best.

ProcessLib/LocalAssemblerInterface.h
endJunction: Yes I thought about it, but the default implementation might have a drawback that these functions are called for every element. I'll need to check that before providing the defaults here. Tonight.

endJunction:
Passing t to the `Process::postTimestep` is needed because the `local assembler's::postTimestep(x, t)` has t in its signature and the straight forward way to get the time there is over `Process::postTimestep(x, t)`.

endJunction:
@Yonghui56 Please tell us if the t and delta_t are sufficient for your particular process. You can store them in the shared process data for later use (for example in the postTimestep()).

AssemblerLib/VectorMatrixAssembler.h
```diff
@@ -136,16 +136,20 @@ class VectorMatrixAssembler<GlobalMatrix, GlobalVector, LocalAssembler,
     /// \attention The index \c id is not necessarily the mesh item's id.
     void preTimestep(std::size_t const id,
                      LocalAssembler& local_assembler,
-                     GlobalVector const& x) const
+                     GlobalVector const& x,
+                     double const t,
```
norihiro-w: why not pass a TimeStep object which contain both information?

chleh: I'd prefer two doubles for the time being. However, then one could also discuss their order. Cf. assemble(t,x,M,K,b) vs preTimestep(x,t,dt).

endJunction: Agree with Christoph, as long as we don't need the whole TimeStep object (e.g. for the multi-step methods storing multiple dt's) the two doubles are good.

chleh:
:+1:

PR #1216, opened 2016-05-26
Splitting FileIO part 3

norihiro-w:
update: move some FileIO stuff to GeoLib and NumLib


norihiro-w:
todo
- extract change in DE

norihiro-w:
travis failed due to style check of *.xsd files

bilke:
Please remove the trailing whitespace in the xsd files to get the green check mark.

norihiro-w:
done

NumLib/IO/VtkIO/PVDFile.cpp
```diff
@@ -14,7 +14,9 @@
 #include <limits>
 #include <logog/include/logog.hpp>
 
-namespace FileIO
+namespace NumLib
```
chleh: I'd put that file where also VTK output resides, because this class only makes sense at places where you do vtk output.

chleh:
:+1: but please consider the PVD remark.

norihiro-w:
moved PVD

PR #1218, opened 2016-05-26
remove #ifdef OGS_USE_EIGEN from local matrix related files

norihiro-w:
as titled. #ifdef OGS_USE_EIGEN is still used for global matrix, vector, and linear solver.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,7 +73,7 @@ option(OGS_USE_PETSC "Use PETSc routines" OFF)
 option(OGS_USE_MPI "Use MPI" OFF)
 
 # Eigen
-option(OGS_USE_EIGEN "Use EIGEN for local matrix and vector" ON)
+option(OGS_USE_EIGEN "Use Eigen linear solver" ON)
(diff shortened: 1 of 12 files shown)


CMakeLists.txt
```diff
@@ -134,15 +134,16 @@ if(OGS_USE_MPI)
     add_definitions(-DUSE_MPI)
 endif()
 
+add_definitions(-DEIGEN_INITIALIZE_MATRICES_BY_ZERO) # TODO check if needed
+if (EIGEN_NO_DEBUG)
+    add_definitions(-DEIGEN_NO_DEBUG)
```
chleh: Shall we really keep that? AFAIK it disables Eigen checks even in debug mode.

chleh:
:+1:

CMakeLists.txt
norihiro-w: I would keep it as it can speed up debugging when you are sure that Eigen related parts are fine and there will be a bug in other places. Note that EIGEN_NO_DEBUG is OFF by default.

PR #1221, opened 2016-05-27
define shape_matrices in win

Yonghui56:


endJunction:
Jenkins, test this please!

chleh:
https://eigen.tuxfamily.org/dox/group__TopicStlContainers.html says that also the header `Eigen/StdVector` has to be incuded, if I read that page correctly.

Furthermore we could add a typedef (e.g. `VectorOfShapeMatrices`) s.t. we don't always have to write the Eigen allocator stuff explicitly.

chleh:
@endJunction, regarding our discussion yesterday: Finally I tested the alignment of double, and it is 8 bytes on my machine, which certainly differs from the 16 byte requirement for Eigen vectorizable types. Thus, the general problem addressed in this PR is not only restricted to 32 bit machines.

[The Eigen vectorizable types page](https://eigen.tuxfamily.org/dox/group__TopicFixedSizeVectorizable.html) says
> Now, vectorization (both SSE and AltiVec) works with 128-bit packets. Moreover, for performance reasons, these packets need to be have 128-bit alignment.

So maybe we _only_ sacrifice speed when not being 16 bytes aligned. And maybe @Yonghui56's program failed because the CPU's vector instruction failed to read the 4 bytes aligned double on his windows 32 bit build.

Anyway, I think we should keep this PR. I mainly see one problem: The custom Eigen allocator and the StlVector header can be easily forgotten.

chleh:
Some notes:
Eigen auto-aligns (AutoAlign) vectorizable data types by default. This can be overridden in the Options template param passed to the Matrix class, cf.
https://eigen.tuxfamily.org/dox/group__enums.html#ga8193de13223530e9ad888f78f2c2b5b2
https://eigen.tuxfamily.org/dox/classEigen_1_1Matrix.html
Core/DenseStorage.h (the type plain_array)

However, C++ operator new still does not necessarily allocate aligned storage, cf.
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=65122
So if we want to align our shape matrices vectors, we'll probably have to use the custom allocator.

endJunction:
Some tests: mechanics branch (heavy on Eigen), cube with 1e3 elements, nonlinear local newton, 100 time steps.
Disabling the Eigen alignement for the ShapeMatrices is fastest:
26.804170524 seconds time elapsed                                          ( +-  0.14% )

Without any modifications like (ufz/master), i.e. Eigen alignment for ShapeMatrices:
26.811209107 seconds time elapsed                                          ( +-  0.05% )

With this PR and appropriate fixes in the mechanics:
26.873654409 seconds time elapsed                                          ( +-  0.14% )


I opt for removal of that explicit Eigen alignment, especially because now the Win32 builds are harder to create.

PR #1222, opened 2016-05-29
Remove iostream include

endJunction:
Replace cout/cerr output with hopefully semantically equal logog calls.

### DIFF
--- a/BaseLib/IO/Writer.cpp
+++ b/BaseLib/IO/Writer.cpp
@@ -12,12 +12,12 @@
  *
  */
 
-// ** INCLUDES **
 #include "Writer.h"
 
-#include <iostream>
 #include <fstream>
 #include <limits>
+
+#include <logog/include/logog.hpp>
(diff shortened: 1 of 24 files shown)


BaseLib/IO/Writer.cpp
```diff
@@ -12,13 +12,13 @@
...
-// ** INCLUDES **
-#include "Writer.h"
-
-#include <iostream>
 #include <fstream>
 #include <limits>
 
+#include <logog/include/logog.hpp>
+
+#include "Writer.h"
```
norihiro-w: why include its header here?

chleh:
:+1:

BaseLib/IO/Writer.cpp
endJunction: It was a momentary weakness ;)

TomFischer:
:fast_forward:

Issue #1164, opened 2016-04-26
Moving FileIO to each module


norihiro-w:
(I'm discussing the future direction and not immediate changes in the source code.) 

I would like to suggest here splitting FileIO and move them to each corresponding module. This seems to be the only solution to work out the circular dependency issue around FileIO. Using a lambda function or adding a module just to hold a data class ( #1162) are temporal solutions and are not really solving the problem. Also, FileIO actually is breaking modularity of OGS codes, e.g. Mesh module should provide file IO stuff by itself (current codes requires MeshLib + FileIO + all other modules FileIO depends). 

To implement it, changes we need are
1. move common writer classes to BaseLib
2. move specific IO classes to corresponding module, e.g. GeoLib/FileIO, MeshLib/FileIO
3. modify codes where the IO classes are used

Let me know if you have any objections to the suggestion. As I'm suggesting this, I'll do the code modification (of course your help is welcome).

endJunction:
Looks like a good plan to me!

TomFischer:
@norihiro-w I can support you. How do we organize the work?

norihiro-w:
@TomFischer Thanks! I think we should wait until we finish merging Christoph's version and also ask other guys about a right timing to do it, because it may change many places in the code. But if you have time, you can start Step 1 (see above). I guess this gives little influence to others.

TomFischer:
I moved Writer.{h,cpp} to BaseLib [here](https://github.com/TomFischer/ogs/tree/RestructureFileIO). The second step will be more complicated since in some cases geometry and mesh related stuff is coupled together. Examples are the GMSHInterface (can read GMSH meshes and write GMSH geometries), Feflow interface, TetGenInterface or the ShapeInterface. My first idea is to splitt those Interfaces in mesh related and geometry related parts. Are there other ideas or remarks?

norihiro-w:
Not from me. I think that's the best way to go.

chleh:
I think your work on this issue won't influence my TES process much, since I don't do any I/O directly by myself. So from my point of view you can well go ahead.

@TomFischer maybe you can simply leave the "problematic" classes untouched. Then we will have solved the problem at least to 80% or so.

TomFischer:
For discussion: I would introduce new namespace like `MeshLib::FileIO` and `GeoLib::FileIO`?!

norihiro-w:
Do we really need it? I don't think there is a naming conflict.

TomFischer:
[...]

PR #1217, opened 2016-05-26
Splitting FileIO part 4

norihiro-w:
follow up of #1216 

This PR targets the following files in FileIO, which IMO are more related to applications and shouldn't go to basic modules such as GeoLib and MeshLib. 
- GMSInterface
- PetrelInterface
- SHPInterface
- XmlGspInterface
- XmlLutInterface
- XmlNumInterface
- ImportFileTypes.h

ImportFileTypes.h is moved to Base dir in DE because only DE is using it. Other files are moved to the new directory Applications/FileIO (ApplicationsFileIOLib). Having these in ApplicationsLib/IO is also possible but I'm not sure yet what is a purpose of ApplicationsLib.

PR #1225, opened 2016-05-30
[TetGenInterface] read Smesh as Mesh and move it to MeshLib/IO

norihiro-w:
as titled and see comments in #1164


Applications/DataExplorer/mainwindow.cpp
```diff
@@ -624,18 +624,22 @@ void MainWindow::loadFile(ImportFileType::type t, const QString &fileName)
     }
     else if (t == ImportFileType::TETGEN)
     {
-        if (fi.suffix().toLower().compare("poly") == 0 || fi.suffix().toLower().compare("smesh") == 0)
+        if (fi.suffix().toLower().compare("smesh") == 0)
```
TomFischer: What about poly input format?

norihiro-w: i excluded it because They the Tetgen interface is not supporting poly format yet.

rinkk:
I have a number of issues with this PR. I'm going to list these here and maybe we can discuss how to proceed:

* First, a general thing: I don't see the benefit in moving import/export file formats to GeoLib/MeshLib. I can understand that GL/ML should be independent libraries and as such they need to read/write their respective file formats. External formats however are application-specific and the seperation between geometry and mesh is often debatable, as in this case here. If you were to move the OGS-GeoLib into a, say, mechanics code, there'd be no need to read shp-files or GMS-files as these are typical formats for hydrogeological data. If these interfaces are located in FileIO they can work like add-ons for GL/ML and don't clutter up the respective libraries. Also, there are no circular dependencies concerning these interfaces so there is also no reason from a software-engineering POV to move them.

* Concerning this PR: It's debatable if a TetGen - SMesh is a mesh or a geometry file. These files are defining Piecewise Linear Complex (PLC) functions and these could either be viewed as 2D meshes or as triangulated geometric surfaces. Since pre-processing data in OGS is usually geometry and smesh-files *cannot* contain 3D "pieces" I opted to put them into GeoLib. Putting them into MeshLib is the same as putting ArcGIS TIN files into ML or GOCAD TSurfaces. It's certainly possible but our geometric structures represent much better what these files describe.

* Removing the writeSmesh-method for geometries: You removed this but it is actually useful functionality if the input data consists of multiple geometric surfaces, such as GOCAD TSurface files. We do not currently allow to use this in the DataExplorer because it would require stable methods to check if multiple surfaces form a closed volume, but it is usable in scripts/tools where you can assume the user knows what they are doing. In contrast, we do currently not have the functionality to combine multiple 2D meshes into a similar structure.

Sorry for the wall of text.

norihiro-w:
TODO: move tests

norihiro-w:
@rinkk Didn't see #1164 and #1217? Application specific IOs were moved to Applications/FileIO. 
- With current usage, I would keep SMesh in MeshLib because SMesh is indeed with point and topological definitions (i.e. elements). Since TIN doesn't include topological information, it cannot fully represent SMesh, but a mesh can do. 
- If you need Surface object from SMesh, you can do it by using `convertMeshToGeo()` in DE. Opposite conversion is also possible with `convertSurfaceToMesh()`.

If you still don't get it, let's have a short meeting next week. sorry I'm occupied this week.

rinkk:
I am referring to *multiple* surfaces forming *one* geometry and TetGen supporting exactly that.  
Yes, you can convert *one* surface into *one* mesh and vice versa but that doesn't solve the problem of having multiple surface input data from VTK / GOCAD / ArcGIS and converting this.

norihiro-w:
That is something should be done in the conversion. A mesh imported from Smsh has MaterialIDs to distinguish different surfaces.

norihiro-w:
@rinkk I added GeoLib/IO/TetGenInterface for direct IO between Surface objects and a Tetgen smesh file.

norihiro-w:
@rinkk I added also TetGenInterface to GeoLib/IO which can directly make a conversion between Surface objects and Smesh file as it was done previously. I hope this makes you happy

norihiro-w:
can someone review this?

norihiro-w:
modified as discussed in the meeting.

endJunction:
Looks good to me.
:fast_forward:

rinkk:
:+1:

PR #1226, opened 2016-05-31
Split Gmsh Interface into GeoLib-part and MeshLib-part

TomFischer:

This is one step to resolve the issue https://github.com/ufz/ogs/issues/1164.


MeshLib/IO/GmshReader.h
```diff
@@ -0,0 +1,53 @@
...
+class Node;
+class Element;
+class Mesh;
+
+namespace IO
+{
+namespace GMSH
+{
+
+void readNodeIDs(std::ifstream& in, unsigned n_nodes,
```
norihiro-w: Is it right that  this function should not be called from outside? If that's the case, do you still need the declaration here? same applies to the below function

MeshLib/IO/GmshReader.h
```diff
@@ -0,0 +1,53 @@
+/**
+ *
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ */
+
+#include <fstream>
```
norihiro-w: maybe you can include the forward deceleration

MeshLib/IO/GmshReader.h
```diff
@@ -0,0 +1,40 @@
+/**
+ *
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ */
+
+#include <iosfwd>
```
norihiro-w: no need to include this.

MeshLib/IO/GmshReader.h
```diff
@@ -0,0 +1,40 @@
...
+
+namespace MeshLib
+{
+class Node;
```
norihiro-w: Forward declaration of Node and Element can be removed

GeoLib/IO/GMSHInterface.cpp
```diff
@@ -16,12 +16,12 @@
...
 #include "BaseLib/FileTools.h"
 
-#include "FileIO/GMSHInterface.h"
-#include "FileIO/GmshIO/GMSHAdaptiveMeshDensity.h"
-#include "FileIO/GmshIO/GMSHFixedMeshDensity.h"
-#include "FileIO/GmshIO/GMSHPoint.h"
-#include "FileIO/GmshIO/GMSHPolygonTree.h"
-#include "FileIO/GmshIO/GMSHMeshDensityStrategy.h"
+#include "GeoLib/IO/GMSHInterface.h"
+#include "GeoLib/IO/GmshIO/GMSHAdaptiveMeshDensity.h"
```
chleh: folder could be renamed to GeoLib/IO/Gmsh

Applications/DataExplorer/mainwindow.cpp
```diff
@@ -430,10 +430,10 @@ void MainWindow::save()
         std::vector<std::string> selected_geometries;
         _project.getGEOObjects().getGeometryNames(selected_geometries);
 
-        GMSHInterface gmsh_io(
+        GeoLib::IO::GMSHInterface gmsh_io(
             _project.getGEOObjects(), true,
-            FileIO::GMSH::MeshDensityAlgorithm::AdaptiveMeshDensity,
-            0.05, 0.5, 2, selected_geometries);
+            GeoLib::IO::GMSH::MeshDensityAlgorithm::AdaptiveMeshDensity, 0.05,
```
chleh: The directory is named GeoLib/IO/GmshIO, but the namespace is GeoLib::IO::GMSH. Maybe the two things should match?

chleh:
:+1:

endJunction:
I also would prefer not to repeat "IO" in the "GeoLib/IO/GmshIO" second time.

PR #1228, opened 2016-06-01
Copied physical constants from OGS5

chleh:
... and use them in TES and for adsorption.


TomFischer:
Looks useful. :fast_forward:

PR #1229, opened 2016-06-01
Duplicate geometry

rinkk:
This adds a class for creating a deep copy of an existing geometry. I needed this for a number of external tools and we discussed that it might make sense to put this into GeoLib.

@TomFischer : maybe you can use the surface-copy routine as a blueprint for the surface copy ctor.

### DIFF
--- a/GeoLib/DuplicateGeometry.cpp
+++ b/GeoLib/DuplicateGeometry.cpp
@@ -94,7 +94,7 @@ std::unique_ptr<std::vector<Surface*>> DuplicateGeometry::copySurfacesVector(
             continue;
         (*new_surfaces)[i] = new GeoLib::Surface(*_geo_objects.getPointVec(_output_name));
 
-        std::size_t n_tris (surfaces[i]->getNTriangles());
+        std::size_t n_tris (surfaces[i]->getNumberOfTriangles());
         for (std::size_t j=0; j<n_tris; ++j)
         {
             GeoLib::Triangle const* t = (*surfaces[i])[j];
--- a/GeoLib/DuplicateGeometry.cpp
+++ b/GeoLib/DuplicateGeometry.cpp
@@ -61,9 +61,9 @@ void DuplicateGeometry::duplicate(std::string const& input_name)
 }
 
 std::unique_ptr<std::vector<GeoLib::Polyline*>> DuplicateGeometry::copyPolylinesVector(
-    std::vector<GeoLib::Polyline*> const& polylines)
+    std::vector<GeoLib::Polyline*> const& polylines) const
 {
-    std::size_t n_plys = polylines.size();
+    std::size_t const n_plys = polylines.size();
     std::unique_ptr<std::vector<GeoLib::Polyline*>> new_lines
         (new std::vector<Polyline*>(n_plys, nullptr));
 
@@ -72,7 +72,7 @@ std::unique_ptr<std::vector<GeoLib::Polyline*>> DuplicateGeometry::copyPolylines
         if (polylines[i] == nullptr)
             continue;
         (*new_lines)[i] = new GeoLib::Polyline(*_geo_objects.getPointVec(_output_name));
-        std::size_t nLinePnts (polylines[i]->getNumberOfPoints());
+        std::size_t const nLinePnts (polylines[i]->getNumberOfPoints());
         for (std::size_t j=0; j<nLinePnts; ++j)
             (*new_lines)[i]->addPoint(polylines[i]->getPointID(j));
     }
@@ -80,9 +80,9 @@ std::unique_ptr<std::vector<GeoLib::Polyline*>> DuplicateGeometry::copyPolylines
 }
 
 std::unique_ptr<std::vector<Surface*>> DuplicateGeometry::copySurfacesVector(
-    std::vector<Surface*> const& surfaces)
+    std::vector<Surface*> const& surfaces) const
 {
-    std::size_t n_sfc = surfaces.size();
+    std::size_t const n_sfc = surfaces.size();
     std::unique_ptr<std::vector<GeoLib::Surface*>> new_surfaces
         (new std::vector<Surface*>(n_sfc, nullptr));
 
@@ -92,7 +92,7 @@ std::unique_ptr<std::vector<Surface*>> DuplicateGeometry::copySurfacesVector(
             continue;
         (*new_surfaces)[i] = new GeoLib::Surface(*_geo_objects.getPointVec(_output_name));
 
-        std::size_t n_tris (surfaces[i]->getNumberOfTriangles());
+        std::size_t const n_tris (surfaces[i]->getNumberOfTriangles());
         for (std::size_t j=0; j<n_tris; ++j)
         {
             GeoLib::Triangle const* t = (*surfaces[i])[j];
--- a/GeoLib/DuplicateGeometry.cpp
+++ b/GeoLib/DuplicateGeometry.cpp
@@ -9,12 +9,10 @@
 
 #include "DuplicateGeometry.h"
 
-#include <memory>
-#include <vector>
-
 // ThirdParty/logog
 #include <logog/include/logog.hpp>
 
+#include "GeoLib/GEOObjects.h"
 #include "GeoLib/Point.h"
 #include "GeoLib/Polyline.h"
 #include "GeoLib/Surface.h"
--- a/GeoLib/DuplicateGeometry.cpp
+++ b/GeoLib/DuplicateGeometry.cpp
@@ -1,0 +1,123 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#include "DuplicateGeometry.h"
+
+#include <memory>
+#include <vector>
+
+// ThirdParty/logog
+#include "logog/include/logog.hpp"
+
+#include "GeoLib/Point.h"
+#include "GeoLib/Polyline.h"
+#include "GeoLib/Surface.h"
+#include "GeoLib/Triangle.h"
+
+
+namespace GeoLib
+{
+
+DuplicateGeometry::DuplicateGeometry(GeoLib::GEOObjects &geo_objects,
+	std::string const& input_name,
+	std::string const& output_name)
+: _output_name(output_name), _geo_objects(geo_objects)
+{
+	duplicate(input_name);
+}
+
+void DuplicateGeometry::duplicate(std::string const& input_name)
+{
+	std::vector<GeoLib::Point*> const*const pnts (_geo_objects.getPointVec(input_name));
+	if (pnts == nullptr)
+	{
+		ERR("Geometry \"%s\" not found.", input_name.c_str());
+		return;
+	}
+
+	std::unique_ptr< std::vector<GeoLib::Point*> > new_pnts (new std::vector<GeoLib::Point*>);
+	new_pnts->reserve(pnts->size());
+	std::transform(pnts->cbegin(), pnts->cend(), std::back_inserter(*new_pnts),
+		[](GeoLib::Point* point) { return new GeoLib::Point(*point); });
+	_geo_objects.addPointVec(std::move(new_pnts), _output_name);
+
+	std::vector<GeoLib::Polyline*> const* plys (_geo_objects.getPolylineVec(input_name));
+	if (plys)
+	{
+		auto new_plys = copyPolylinesVector(*plys);
+		_geo_objects.addPolylineVec(std::move(new_plys), _output_name);
+	}
+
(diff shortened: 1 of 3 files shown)


GeoLib/DuplicateGeometry.cpp
```diff
@@ -0,0 +1,123 @@
...
+ *
+ */
+
+#include "DuplicateGeometry.h"
+
+#include <memory>
+#include <vector>
+
+// ThirdParty/logog
+#include "logog/include/logog.hpp"
```
endJunction: System include

Tests/GeoLib/TestDuplicateGeometry.cpp
```diff
@@ -0,0 +1,150 @@
...
+#include "GeoLib/Surface.h"
+#include "GeoLib/Triangle.h"
+
+TEST(GeoLib, DuplicateGeometry)
+{
+	GeoLib::GEOObjects geo;
+	std::string input_name ("input");
+
+	// generate points
+	srand ( static_cast<unsigned>(time(nullptr)) );
```
endJunction: All of the system calls do live in the std:: namespace. Please explicitly specify it.

Tests/GeoLib/TestDuplicateGeometry.cpp
```diff
@@ -0,0 +1,150 @@
...
+	std::size_t n_pnts (rand() % 1000 + 100);
+	int box_size (rand());
+	double half_box_size(box_size/2);
+	std::unique_ptr<std::vector<GeoLib::Point*>> pnts (new std::vector<GeoLib::Point*>);
+	pnts->reserve(n_pnts);
+	for (int k(0); k<n_pnts; k++) {
+		pnts->push_back(new GeoLib::Point(rand() % box_size - half_box_size, rand() % box_size - half_box_size, rand() % box_size - half_box_size));
```
endJunction: please format the long lines. clang-format.
Tabs are forbidden.

Tests/GeoLib/TestDuplicateGeometry.cpp
```diff
@@ -0,0 +1,150 @@
...
+		std::vector<GeoLib::Surface*> const* new_sfcs (geo.getSurfaceVec(output2));
+		ASSERT_EQ(n_sfcs, new_sfcs->size());
+
+		for (std::size_t i=0; i<n_sfcs; ++i)
+		{
+			std::size_t const n_tris ((*new_sfcs)[i]->getNTriangles());
+			ASSERT_EQ(n_tris, (*sfcs)[i]->getNTriangles());
+			for (std::size_t j=0; j<n_tris; ++j)
+				for (std::size_t k=0; k<3; ++k)
+					ASSERT_EQ((*(*(*sfcs)[i])[j])[k], (*(*(*new_sfcs)[i])[j])[k]);
```
endJunction: This is lisp. In c++ usually a comparison operator is provided, such that higher level abstractions are not mixed with lowest level accesses.

Tests/GeoLib/TestDuplicateGeometry.cpp
bilke: You mean spaces are used!

Tests/GeoLib/TestDuplicateGeometry.cpp
rinkk: An == op on this level doesn't make sense except for theoretical test cases. Surfaces should be evalutated as identical if they are topological identical. This includes variances in id's, triangle sequences or even slight differences in elevation because for all practical purposes these values will vary due to floating point inaccuracies, mesh algorithms and the like.

I agree that the above line doesn't look nice but it gets the job done and it's just for a test.

GeoLib/DuplicateGeometry.h
```diff
@@ -0,0 +1,69 @@
...
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef DUPLICATEGEOMETRY_H_
+#define DUPLICATEGEOMETRY_H_
+
+#include <string>
+
+#include "GeoLib/GEOObjects.h"
```
norihiro-w: this can also be forward declaration

GeoLib/DuplicateGeometry.h
```diff
@@ -0,0 +1,69 @@
...
+     * \param geo_objects The container for geometries
+     * \param input_name  The geometry to be copied
+     * \param output_name The name of the copy (note: this might be modified by GEOObjects)
+     */
+    DuplicateGeometry(GeoLib::GEOObjects &geo_objects,
+        std::string const& input_name,
+        std::string const& output_name);
+
+    // Returns the (possibly modified) output name of the new geometry.
+    std::string const& getFinalizedOutputName() { return _output_name; }
```
norihiro-w: can be const function

GeoLib/DuplicateGeometry.cpp
```diff
@@ -0,0 +1,123 @@
...
+            (*new_surfaces)[i]->addTriangle(
+                t->getPoint(0)->getID(), t->getPoint(1)->getID(), t->getPoint(2)->getID());
+        }
+    }
+    return new_surfaces;
+}
+
+std::vector<GeoLib::Point*>& DuplicateGeometry::getPointVectorCopy()
+{
+    return const_cast<std::vector<GeoLib::Point*>&>(*_geo_objects.getPointVec(_output_name));
```
chleh: Rather than having const casts here, I'd prefer a non-const overloads of `getPointVec()` etc.

Tests/GeoLib/TestDuplicateGeometry.cpp
```diff
@@ -0,0 +1,153 @@
...
+        {
+            std::size_t const n_ply_pnts ((*new_plys)[i]->getNumberOfPoints());
+            ASSERT_EQ(n_ply_pnts, (*plys)[i]->getNumberOfPoints());
+            ASSERT_EQ((*new_plys)[i]->getNumberOfSegments(), (*plys)[i]->getNumberOfSegments());
+            for (std::size_t j=0; j<n_ply_pnts; ++j)
+                ASSERT_EQ((*plys)[i]->getPointID(j), (*new_plys)[i]->getPointID(j));
+        }
+
+        std::vector<GeoLib::Polyline*>& mod_plys (dup.getPolylineVectorCopy());
+        mod_plys.push_back(new GeoLib::Polyline(*(*new_plys)[0]));
```
chleh: Only curious: Is there a reason for that?

GeoLib/DuplicateGeometry.cpp
TomFischer: Non-const overloads of `getPointVec()` should not be provided. The internal vector of points must not be changed outside since there are internal data structures (point id map) that have to stay in sync with the vector of points.

chleh:
don't like the const-cast apart from that :+1:

GeoLib/DuplicateGeometry.h
rinkk: cannot be a forward declaration, we don't use pointers of  geoobjects here.

Tests/GeoLib/TestDuplicateGeometry.cpp
rinkk: adding a polyline to the copied vector to test if it's possible and if the correct result is returned.

GeoLib/DuplicateGeometry.cpp
chleh: But then this const_cast is also wrong, isn't it?

Tests/GeoLib/TestDuplicateGeometry.cpp
chleh: OK, thanks for the explanation.

GeoLib/DuplicateGeometry.cpp
rinkk: No, it isn't. At this point to geometry has been "cloned". This is useless in itself and only done if data needs to be modified for mapping, etc. So once you created the copy you are allowed to modify exactly this copy and nothing else.

GeoLib/DuplicateGeometry.h
chleh: But can't references bve fwd-declared, too?

rinkk: I'm getting a lot of compile errors if I do it so at least for me it doesn't seem to work.

rinkk: Scratch my last comment, it does work, there was another thing causing compile errors...

rinkk: Argh .... I accidently wrote a forward declaration *and* a normal declaration in the header file. Using *only* the forward declaration the compile errors are back and it really doesn't seem to work.

rinkk: final comment on this (sorry for the spam): I shuffeld some system-includes around and now it *does* work. sorry for the confusion.

TomFischer:
I will set it to "rebase and merge". `rand` in the test lives in `std` namespace.

GeoLib/DuplicateGeometry.cpp
chleh: I still don't like it, but I won't block this PR because of that.

rinkk: I don't care if this is done here or in GEOObjects but I do *need* an option to modify geometry. If it'd be available from GEOObjects would be easier and more widely applicable but having this here is better than nothing...

GeoLib/DuplicateGeometry.cpp
```diff
@@ -0,0 +1,121 @@
...
+    {
+        auto new_sfcs = copySurfacesVector(*sfcs);
+        _geo_objects.addSurfaceVec(std::move(new_sfcs), _output_name);
+    }
+}
+
+std::unique_ptr<std::vector<GeoLib::Polyline*>> DuplicateGeometry::copyPolylinesVector(
+    std::vector<GeoLib::Polyline*> const& polylines)
+{
+    std::size_t n_plys = polylines.size();
```
endJunction: const

GeoLib/DuplicateGeometry.cpp
```diff
@@ -0,0 +1,121 @@
...
+    std::vector<GeoLib::Surface*> const* sfcs (_geo_objects.getSurfaceVec(input_name));
+    if (sfcs)
+    {
+        auto new_sfcs = copySurfacesVector(*sfcs);
+        _geo_objects.addSurfaceVec(std::move(new_sfcs), _output_name);
+    }
+}
+
+std::unique_ptr<std::vector<GeoLib::Polyline*>> DuplicateGeometry::copyPolylinesVector(
+    std::vector<GeoLib::Polyline*> const& polylines)
```
endJunction: Not sure here, but if the method is not modifying anything it could be declared const.

GeoLib/DuplicateGeometry.cpp
```diff
@@ -0,0 +1,121 @@
...
+    std::size_t n_plys = polylines.size();
+    std::unique_ptr<std::vector<GeoLib::Polyline*>> new_lines
+        (new std::vector<Polyline*>(n_plys, nullptr));
+
+    for (std::size_t i=0; i<n_plys; ++i)
+    {
+        if (polylines[i] == nullptr)
+            continue;
+        (*new_lines)[i] = new GeoLib::Polyline(*_geo_objects.getPointVec(_output_name));
+        std::size_t nLinePnts (polylines[i]->getNumberOfPoints());
```
endJunction: const

GeoLib/DuplicateGeometry.cpp
```diff
@@ -0,0 +1,121 @@
...
+        for (std::size_t j=0; j<nLinePnts; ++j)
+            (*new_lines)[i]->addPoint(polylines[i]->getPointID(j));
+    }
+    return new_lines;
+}
+
+std::unique_ptr<std::vector<Surface*>> DuplicateGeometry::copySurfacesVector(
+    std::vector<Surface*> const& surfaces)
```
endJunction: possibly const?

GeoLib/DuplicateGeometry.h
```diff
@@ -0,0 +1,69 @@
...
+    // Returns a reference to the copied point vector for modification
+    std::vector<GeoLib::Point*>& getPointVectorCopy();
+    // Returns a reference to the copied polyline vector for modification
+    std::vector<GeoLib::Polyline*>& getPolylineVectorCopy();
+    // Returns a reference to the copied surface vector for modification
+    std::vector<GeoLib::Surface*>& getSurfaceVectorCopy();
+
+private:
+    // creates a deep copy of the designated geometry
+    void duplicate(std::string const& input_name);
```
endJunction: This method (and the two following) cannot be called from outside (private and class is final) and are used only in the implementation part. I'd suggest to move them into the implementation instead of adding them to the private interface part.

rinkk: All private methods cannot be called from outside and are only used for the implementation part of a class. That's the very definition of "private". Personally I dislike having methods only in the cpp file because it's very difficult to find out these methods exist at all. Therefore I'd rather keep it as it is.

chleh: Maybe it's not so important here: The class is final...

PR #1230, opened 2016-06-01
Disallow 32-bit Windows builds.

bilke:
Can be forced by setting `OGS_32_BIT=ON`.

In general Windows 32-bit builds are not supported but often are accidentally configured (because CMake defaults to a 32-bit generator). Beside running OGS on Win XP there is no reason to build for 32-bit Windows.

### DIFF
--- a/scripts/cmake/ProjectSetup.cmake
+++ b/scripts/cmake/ProjectSetup.cmake
@@ -1,3 +1,10 @@
+# Check requirements / supported configurations
+if(MSVC AND NOT HAVE_64_BIT AND NOT OGS_32_BIT)
+    message(FATAL_ERROR "Building OGS on Windows with 32-bit is not supported! \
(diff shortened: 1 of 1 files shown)


endJunction:
:fast_forward:

TomFischer:
Looks good. :+1:

chleh:
:+1: You could also disallow 32 bit builds on other platforms, or are they supported better on, e.g. linux?
`OGS_32_BIT` looks very innocent. I'd call it `OGS_I_REALLY_DESPERATELY_NEED_TO_BUILD_32_BIT` or something similar.

bilke:
On Linux-based platforms [it is much harder](http://stackoverflow.com/questions/5805874/the-proper-way-of-forcing-a-32-bit-compile-using-cmake) to configure for 32-bit so in this case I assume the person knows what he is doing...

PR #1235, opened 2016-06-03
[GL/IO] Increase output precision for gml output.

TomFischer:
The increase of the output precision solves problem that sometimes not all boundary conditions set via a geometry are found. This can be caused by the different precision the mesh and the geometry are written. This PR tries to resolve the problem.


chleh:
:fast_forward:

bilke:
👍

PR #1236, opened 2016-06-03
Simplified VecMatAssembler

chleh:
@endJunction suggested that some lambdas can be removed. That might increase performance a bit, and it makes the code easier to understand.

There are no logic changes in this PR, it is merely a rewrite of existing code.

### DIFF
--- a/NumLib/Assembler/VectorMatrixAssembler.h
+++ b/NumLib/Assembler/VectorMatrixAssembler.h
@@ -13,10 +13,13 @@
 #include "NumLib/ODESolver/Types.h"
 #include "NumLib/DOF/LocalToGlobalIndexMap.h"
 
-namespace
+namespace NumLib
+{
+
+namespace detail
 {
 inline std::vector<GlobalIndexType>
-getRowColumnIndices(std::size_t const id,
+getIndices(std::size_t const id,
                     NumLib::LocalToGlobalIndexMap const& dof_table)
 {
     assert(dof_table.size() > id);
@@ -41,7 +44,7 @@ prepareLocalVector(
         NumLib::LocalToGlobalIndexMap const& dof_table,
         GlobalVector const& x)
 {
-    auto const indices = getRowColumnIndices(id, dof_table);
+    auto const indices = getIndices(id, dof_table);
 
     std::vector<double> local_x;
     local_x.reserve(indices.size());
@@ -55,9 +58,6 @@ prepareLocalVector(
     return std::make_tuple(local_x, indices);
 }
 }
-
-namespace NumLib
-{
 
 /*! Calls the local assemblers of FEM processes and assembles
  *  \c GlobalMatrix'es and \c GlobalVector's.
@@ -95,7 +95,7 @@ public:
         const double t, GlobalVector const& x,
         GlobalMatrix& M, GlobalMatrix& K, GlobalVector& b) const
     {
-        auto const x_and_indices = prepareLocalVector(id, _data_pos, x);
+        auto const x_and_indices = detail::prepareLocalVector(id, _data_pos, x);
         auto const& local_x = std::get<0>(x_and_indices);
         auto const& indices = std::get<1>(x_and_indices);
         auto const r_c_indices =
@@ -114,7 +114,7 @@ public:
                           GlobalVector const& x,
                           GlobalMatrix& Jac) const
     {
-        auto const x_and_indices = prepareLocalVector(id, _data_pos, x);
+        auto const x_and_indices = detail::prepareLocalVector(id, _data_pos, x);
         auto const& local_x = std::get<0>(x_and_indices);
         auto const& indices = std::get<1>(x_and_indices);
         auto const r_c_indices =
@@ -133,7 +133,7 @@ public:
                      double const t,
                      double const delta_t) const
     {
-        auto const x_and_indices = prepareLocalVector(id, _data_pos, x);
+        auto const x_and_indices = detail::prepareLocalVector(id, _data_pos, x);
         auto const& local_x = std::get<0>(x_and_indices);
 
         local_assembler.preTimestep(local_x, t, delta_t);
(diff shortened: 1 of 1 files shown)


endJunction:
@chleh I think `std::tie` will resolve the return value problems. It should avoid the copy construction; `std::tuple` has copy ctor. http://stackoverflow.com/a/19800472

chleh:
I think the errors were related to tuple's ctor being explicit. Let' see...

NumLib/Assembler/VectorMatrixAssembler.h
```diff
@@ -32,27 +31,28 @@ getRowColumnIndices(std::size_t const id,
...
+    return indices;
 }
 
-template <typename Callback, typename GlobalVector, typename... Args>
-void passLocalVector_(Callback& cb, std::size_t const id,
-                      NumLib::LocalToGlobalIndexMap const& dof_table,
-                      GlobalVector const& x, Args&&... args)
+template <typename GlobalVector>
+std::tuple<std::vector<double>, std::vector<GlobalIndexType>>
+prepareLocalVector(
```
wenqing: Nice to use std::tuple hereby. :+1:

NumLib/Assembler/VectorMatrixAssembler.h
```diff
@@ -95,17 +95,14 @@ class VectorMatrixAssembler<GlobalMatrix, GlobalVector, LocalAssembler,
...
-                const double t,
-                GlobalMatrix& M, GlobalMatrix& K, GlobalVector& b)
-        {
-            local_assembler.assemble(t, local_x);
-            local_assembler.addToGlobal(r_c_indices, M, K, b);
-        };
-
-        passLocalVector_(cb, id, _data_pos, x, t, M, K, b);
+        auto const x_and_indices = prepareLocalVector(id, _data_pos, x);
+        auto const& local_x = std::get<0>(x_and_indices);
```
wenqing: std:tie can also get references. But is it not clear, which is the faster, by using std::tie or by using std::get.

chleh: I think there is not much benefit from using `std::tie` here: prepareLocalVector() has to return by value, the involved copy might be optimized out by the compiler. `local_x` and `indices` are only references, they will probably be optimized out anyway.

wenqing: thanks.

endJunction:
Baseline 26.772320207 seconds time elapsed                                          ( +-  0.05% )
---
Using tuple  26.971311014 seconds time elapsed                                          ( +-  0.13% )
Using tie 27.006360965 seconds time elapsed                                          ( +-  0.19% )

Using tie means:
```c++
        std::vector<double> local_x;
        std::vector<GlobalIndexType> indices;
        std::tie(local_x, indices) = prepareLocalVector(id, _data_pos, x);
```
and
```c++
        std::vector<double> local_x;
        std::tie(local_x, std::ignore) = prepareLocalVector(id, _data_pos, x);
```
if the indices are not used.

So overall there is no difference between tuple and tie solutions, but avoiding a named temporary when using tie is more readable. There is some difference to the previous code, but I think readability wins here.

@chleh Independent of your decision of tuple or tie :fast_forward:

NumLib/Assembler/VectorMatrixAssembler.h
```diff
@@ -32,27 +31,28 @@ getRowColumnIndices(std::size_t const id,
         indices.insert(indices.end(), idcs.begin(), idcs.end());
     }
 
-    return NumLib::LocalToGlobalIndexMap::RowColumnIndices(indices,
-                                                                 indices);
+    return indices;
```
norihiro-w: why not rename this function if it will no longer return row and column indices?

NumLib/Assembler/VectorMatrixAssembler.h
```diff
@@ -95,17 +95,14 @@ class VectorMatrixAssembler<GlobalMatrix, GlobalVector, LocalAssembler,
...
-                LocalToGlobalIndexMap::RowColumnIndices const& r_c_indices,
-                const double t,
-                GlobalMatrix& M, GlobalMatrix& K, GlobalVector& b)
-        {
-            local_assembler.assemble(t, local_x);
-            local_assembler.addToGlobal(r_c_indices, M, K, b);
-        };
-
-        passLocalVector_(cb, id, _data_pos, x, t, M, K, b);
+        auto const x_and_indices = prepareLocalVector(id, _data_pos, x);
```
norihiro-w: i don't see a good reason for creating indices within this function. Instead, this brings `std::get<0>` and std::get<0>` which mess up usability and readability. maybe I miss some point

chleh: Indices are needed for addToGlobal() below. They have been created inside passLocalVector_() before, too. I think the changes introduced make the code easier to understand. What do you want to change? Use std::tie? I think that wouldn't improve readability.

norihiro-w: i mean, why not have `auto const indices = getRowColumnIndices(id, dof_table);` here, not in `prepareLocalVector()`. then the function returns only one variable and std::get is not necessary anymore.  Is there any reason you need to create indices in that function?

chleh: I don't need an explicit indices vector in prepareLocalVector(), but I need to know about the element's indices there. So the indices vector could be omitted in prepareLocalVector(), but then I'd need to query the dof table twice for the element's indices.
Furthermore the local DOFs and the corresponding indices are linked. Computing them at the same spot emphasizes that, and makes sure they always match (I admit, the chance of a mismatch is not very high).
I'd like to keep the return value of prepareLocalVector() as it is.

norihiro-w: still i don't get it. why can't you just pass an indices vector to prepareLocalVector()? then the query is made only once. And I also don't get when the local DOFs and the corresponding indices don't match.

chleh: I didn't think of that before, and I also didn't get it from your questions. If that makes you happy, I can implement it like that in order to settle this discussion now.

norihiro-w: yes please if you don't have a reason for it

chleh: Your proposed change makes the code even shorter. :smile:

norihiro-w:
👍

TomFischer:
:+1:

PR #1237, opened 2016-06-03
[GL] Impl. of copy constructor of Surface and minor improvements

TomFischer:
- Impl. of Surface copy construtor
- Cleanup unused code
- Clang format
- Make classes Triangle and Surface final
- Readability improvements

### DIFF
--- a/GeoLib/Surface.cpp
+++ b/GeoLib/Surface.cpp
@@ -92,51 +92,49 @@ Surface* Surface::createSurface(const Polyline& ply)
         return nullptr;
     }
 
-    if (ply.getNumberOfPoints() > 2)
-    {
-        // create empty surface
-        Surface* sfc(new Surface(ply.getPointsVec()));
-
-        Polygon* polygon(new Polygon(ply));
-        polygon->computeListOfSimplePolygons();
-
-        // create surfaces from simple polygons
-        const std::list<GeoLib::Polygon*>& list_of_simple_polygons(
-            polygon->getListOfSimplePolygons());
-        for (std::list<GeoLib::Polygon*>::const_iterator simple_polygon_it(
-                 list_of_simple_polygons.begin());
-             simple_polygon_it != list_of_simple_polygons.end();
-             ++simple_polygon_it)
-        {
-            std::list<GeoLib::Triangle> triangles;
-            GeoLib::EarClippingTriangulation(*simple_polygon_it, triangles);
-
-            // add Triangles to Surface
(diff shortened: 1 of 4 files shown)


GeoLib/Surface.h
```diff
@@ -36,31 +32,36 @@ class SurfaceGrid;
...
-    Surface(const std::vector<Point*> &pnt_vec);
-    virtual ~Surface ();
+    explicit Surface(const std::vector<Point*>& pnt_vec);
+    explicit Surface(Surface const& src);
+    ~Surface();
 
-    /// return a geometry type
-    virtual GEOTYPE getGeoType() const {return GEOTYPE::SURFACE;}
+    Surface(Surface && src) = delete;
+    Surface& operator=(Surface const& src) = delete;
```
endJunction: Are there reasons not to implement the copy assignment? If not I'd suggest to implement it too along with the copy ctor.

GeoLib/Surface.cpp
```diff
@@ -31,20 +27,37 @@
...
+}
+
+Surface::Surface(Surface const& src)
+    : _sfc_pnts(src._sfc_pnts),
+      _bounding_volume(new AABB(*(src._bounding_volume))),
+      _surface_grid(nullptr)
+{
+    std::transform(src._sfc_triangles.cbegin(),
+                   src._sfc_triangles.cend(),
+                   std::back_inserter(_sfc_triangles),
```
endJunction: Adding a reserve for _sfc_triangles would save few vector reallocations.

GeoLib/Surface.h
TomFischer: I implemented the copy constructor since @rinkk needs it. The only reason the move constructor and the assignment and assignment move operators are not implemented is that they are not needed until now.

endJunction:
:fast_forward:

GeoLib/Surface.h
```diff
@@ -36,31 +32,36 @@ class SurfaceGrid;
...
  * the Triangles consisting of points from _sfc_pnts.
  * */
-class Surface : public GeoObject
+class Surface final : public GeoObject
 {
 public:
-    Surface(const std::vector<Point*> &pnt_vec);
-    virtual ~Surface ();
+    explicit Surface(const std::vector<Point*>& pnt_vec);
+    explicit Surface(Surface const& src);
```
norihiro-w: no need to set explicit for a copy constructor

norihiro-w:
👍

endJunction:
@TomFischer The `explicit` ctor could be fixed...

PR #1239, opened 2016-06-03
Made tests more flexible

chleh:
This PR
* provides tolerances for individual ctests
* extends tests s.t. vtkdiff can compare multiple files and multiple properties
* switches PETSc tests to vtkdiff.


bilke:
👍

PR #1240, opened 2016-06-06
Move MatrixVectorTraits back to MathLib

chleh:
Some code from NumLib/DOF has been moved to MathLib/LinAlg including:
* MatrixVectorTraits
* MatrixSpecifications
* A part of NumericsConfig

Furthermore MatrixSpecifications have been simplified s.t. they don't reference the DOF table and the mesh anymore. In order to do that I found it necessary to use the GlobalSparsityPattern for the PETSc case, too.


MathLib/LinAlg/MatrixVectorTraits.cpp
```diff
@@ -7,8 +7,8 @@
  *
  */
 
-#include "LocalToGlobalIndexMap.h"
-#include "MatrixProviderUser.h"
+#include "NumLib/DOF/LocalToGlobalIndexMap.h"
```
norihiro-w: remove these

norihiro-w:
👍

Issue #1220, opened 2016-05-26
Consistent naming


norihiro-w:
- getN, getNum, getNumberOf
- ~~ID, Index, IDs, Indices~~ 
- ~~BaseLib/quicksort.h -> Quicksort.h~~
- In ConfigTree: Conf -> Config, Param -> Parameter
- init, initialize

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/1220).

PR #1243, opened 2016-06-10
Consistent naming

norihiro-w:
As written in #1220. Let me know if there is something else to be renamed
- use only getNumberOfXXX() instead of getN, getNum
- use only initialize() instead of init() (private methods are not subject of this change)
- no use of short names in ConfigTree

### DIFF
--- a/Applications/ApplicationsLib/ProjectData.cpp
+++ b/Applications/ApplicationsLib/ProjectData.cpp
@@ -57,13 +57,13 @@ ProjectData::ProjectData(BaseLib::ConfigTree const& project_config,
 {
     std::string const geometry_file = BaseLib::copyPathToFileName(
             //! \ogs_file_param{prj__geometry}
-            project_config.getConfParam<std::string>("geometry"), project_directory
+            project_config.getParameter<std::string>("geometry"), project_directory
         );
     detail::readGeometry(geometry_file, *_geoObjects);
 
     std::string const mesh_file = BaseLib::copyPathToFileName(
             //! \ogs_file_param{prj__mesh}
-            project_config.getConfParam<std::string>("mesh"), project_directory
+            project_config.getParameter<std::string>("mesh"), project_directory
         );
 
     MeshLib::Mesh* const mesh = MeshLib::IO::readMeshFromFile(mesh_file);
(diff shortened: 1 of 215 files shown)


Applications/ApplicationsLib/ProjectData.cpp
```diff
@@ -57,13 +57,13 @@ ProjectData::ProjectData(BaseLib::ConfigTree const& project_config,
 {
     std::string const geometry_file = BaseLib::copyPathToFileName(
             //! \ogs_file_param{prj__geometry}
-            project_config.getConfParam<std::string>("geometry"), project_directory
+            project_config.getParameter<std::string>("geometry"), project_directory
```
chleh: Please keep the prefix `getConf` in all methods of `ConfigTree`. You might change it to `getConfig` in order to improve naming consistency. The reason behind using `getConf` as a prefix for those methods is to be able to easily write scirpts that grep all configuration parameters from the source code. Thereby we are able to provide crosslinking between the documentation and the source code and to check (to a certain extent) if everything is documented.

BaseLib/ConfigTree.h
```diff
@@ -338,24 +338,6 @@ class ConfigTree final
...
     //!\{
 
-    /*! Get parameter \c param from the configuration tree.
-     *
-     * \return the subtree representing the requested parameter
-     *
-     * \pre \c param must not have been read before from this ConfigTree.
-     */
-    ConfigTree
-    getConfParam(std::string const& param) const;
```
chleh: Please keep these methods.
They are here on purpose: You might have something like '<tag attr="xyz">5.0</tag>` in an input file. I think it's more natural to parse that using `getConfParam()` rather than using `getConfSubtree()`.

scripts/doc/get-project-params.sh
```diff
@@ -27,13 +27,13 @@ cat <<"EOF" \
...
-ignoreConfAttribute.*)
-ignoreConfParam.*)
-peekConfParam.*)
+checkParameter.*)
+getAttribute.*)
+getParameter.*)
+getSubtree.*)
+ignoreAttribute.*)
+ignoreParameter.*)
+peekParameter.*)
```
chleh: Note: normalize-param-cache.py also contains some affected reg exps.

Applications/ApplicationsLib/ProjectData.cpp
norihiro-w: ok, i will keep the prefix

BaseLib/ConfigTree.h
norihiro-w: I deleted because nobody use these methods now. anyway if you have a reason for it, I keep them.

scripts/doc/get-project-params.sh
norihiro-w: thanks. I will check it

endJunction:
Nice touch to the readability!
:fast_forward:

Issue #1165, opened 2016-04-26
Error handling


norihiro-w:
I'd like to suggest stop using `std::abort()` and instead using `exception` for error handling. Using `abort` is not a good way for us because
- destructors of objects are not called. Some resource has to be properly released.
- OGS cannot be used as a library, e.g. use in Data Explorer.

Indeed I was against using `exception` before, because I thought it is computationally expensive. After checking it again, I found out that recent compilers implement it quite efficiently, i.e. zero cost if an exception doesn't occur. It seems there is no reason anymore to avoid using exception.

endJunction:
`ag std::abort -l --ignore ThirdParty | xargs sed -i 's/std::abort();/std::runtime_error("ERROR");/'`

https://gist.github.com/endJunction/0adef9c317a6607a0ab0b54c4ef0ff67
clang: 0.78% slower, gcc: 0.56% slower when using exceptions instead of std::abort. Measurement variance ~0.15%.

Personal note: have in mind that the exceptions were allowed even when not used; Somewhat related to this PR https://github.com/ufz/ogs/pull/758

TomFischer:
@endJunction Nice work! For me it looks like we can switch to exceptions. But for a final decision it would be helpful to see measurements for larger examples. Could you also provide some numbers for larger examples?

norihiro-w:
@endJunction is `std::runtime_error();` same as `throw std::runtime_error();`?

chleh:
When switching to throw std::runtime_error() we could consider always providing a message argument. Thereby we would not need to write an additional ERR() line together with every throw but could just catch the exception at the top level and log there.

endJunction:
@TomFischer Larger examples are usually busy with linear solver system, not assembly, but I'll try to run 10 tests tonight.
@norihiro-w Should be `throw std::runtime_error("ERROR");` of course.
@chleh I would also at least distinguish between logic and runtime or memory allocation errors (use what STL has to offer w/o adding own exceptions) and include `__LINE__` and `__FILE__` into the message (where appropriate).

endJunction:
cube_1e6_neumann.prj 10 runs


      | with std::abort | with exceptions | difference
--------|--------------------|-------------------|--------------
clang |   46.750198191s +-  0.11% | 46.710338681s +-  0.10% | 0.09% (less than accuracy)
gcc    |  45.677585534s +-  0.16%  | 45.525131356s +-  0.13% | 0.03% (less than accuracy)

[...]

PR #1245, opened 2016-06-10
Replace std::abort() with std::runtime_error

norihiro-w:
as issued in #1165. I added ERRMSG macro (see 74bd882ff513eeb8ebdbe68762e87e2bbb34a83a) in StringTools.h so that one can write as

```
throw std::runtime_error(ERRMSG("Some error with type %d", -1));
```

which generates an error message "Some error with type -1 at (file name), line (line no)".

### DIFF
--- a/BaseLib/ConfigTree.cpp
+++ b/BaseLib/ConfigTree.cpp
@@ -7,8 +7,11 @@
  *
  */
 
-#include <logog/include/logog.hpp>
 #include "ConfigTree.h"
+
+#include <logog/include/logog.hpp>
+
+#include "Error.h"
 
 // Explicitly instantiate the boost::property_tree::ptree which is a typedef to
 // the following basic_ptree.
@@ -30,12 +33,10 @@ ConfigTree(PTree const& tree,
     : _tree(&tree), _filename(filename), _onerror(error_cb), _onwarning(warning_cb)
 {
     if (!_onerror) {
-        ERR("ConfigTree: No valid error handler provided.");
-        std::abort();
+        OGS_FATAL("ConfigTree: No valid error handler provided.");
     }
     if (!_onwarning) {
-        ERR("ConfigTree: No valid warning handler provided.");
-        std::abort();
+        OGS_FATAL("ConfigTree: No valid warning handler provided.");
     }
 }
 
@@ -193,7 +194,6 @@ void ConfigTree::ignoreConfParamAll(const std::string &param) const
 void ConfigTree::error(const std::string& message) const
 {
     _onerror(_filename, _path, message);
-    std::abort();
 }
 
 void ConfigTree::warning(const std::string& message) const
@@ -205,9 +205,8 @@ void ConfigTree::warning(const std::string& message) const
 void ConfigTree::onerror(const std::string& filename, const std::string& path,
                             const std::string& message)
 {
-    ERR("ConfigTree: In file `%s' at path <%s>: %s",
+    OGS_FATAL("ConfigTree: In file `%s' at path <%s>: %s",
         filename.c_str(), path.c_str(), message.c_str());
-    std::abort();
 }
 
 void ConfigTree::onwarning(const std::string& filename, const std::string& path,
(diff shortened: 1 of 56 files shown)


chleh:
Wouldn't it be better to introduce a macro `OGS_ERROR(...)` or `OGS_FATAL(...)` instead of always having to write `throw std::runtime_error(ERRMSG(...))`?

BaseLib/StringTools.cpp
```diff
@@ -141,3 +142,20 @@ void correctScientificNotation(std::string filename, std::size_t precision)
...
+{
+    char buffer[256];
+    int size_of_buffer = sizeof(buffer) / sizeof(char);
+    memset(buffer, 0, sizeof(buffer));
+    va_list args;
+    va_start(args, format_str);
+#ifdef MSVC
+    vsnprintf_s(buffer, size_of_buffer, _TRUNCATE, format_str.data(), args);
+#else
+    vsnprintf(buffer, size_of_buffer, format_str, args );
```
chleh: Error messages of more than 256 characters are pretty long, but I think there should be a check if the buffer was long enough, and if not a bigger buffer should be allocated dynamically (cf. "Notes" in http://en.cppreference.com/w/cpp/io/c/fprintf).

BaseLib/StringTools.cpp
```diff
@@ -141,3 +142,20 @@ void correctScientificNotation(std::string filename, std::size_t precision)
...
+
+
+std::string format(const char* format_str, ... )
+{
+    char buffer[256];
+    int size_of_buffer = sizeof(buffer) / sizeof(char);
+    memset(buffer, 0, sizeof(buffer));
+    va_list args;
+    va_start(args, format_str);
+#ifdef MSVC
```
chleh: Only wondering: Is this necessary?

BaseLib/ConfigTree.cpp
```diff
@@ -7,9 +7,12 @@
  *
  */
 
-#include <logog/include/logog.hpp>
 #include "ConfigTree.h"
 
+#include <logog/include/logog.hpp>
+
+#include "StringTools.h"
```
chleh: I would like a separate `BaseLib/Logger.h` header, which includes logog and StringTools, better.

chleh:
Will the error message of the exception always be printed? (I.e., in debug and release builds?)
Or do we need a dedicated try/catch block in the main function for that?

chleh:
After the questions are answered :+1:

norihiro-w:
- Having `OGS_ERROR` macro is also fine, though explicitly writing std::runtime_error makes it clear what is really happening. I would like to hear how others think about it.
- I will add a try/catch block in the main function. IMO those error messages should be printed also in release mode.

BaseLib/StringTools.cpp
norihiro-w: will check it

BaseLib/StringTools.cpp
norihiro-w: will check it.TCLAP has also this condition

BaseLib/ConfigTree.cpp
norihiro-w: i am not sure if this is a good idea. exceptions are different from logging.

endJunction:
At first I agreed with Christoph having a single macro like OGS_FATAL. This could be also replaced with something different than throw. On the other hand there are more exception types than the runtime error; having a single macro takes the opportunity to have different error handling...

chleh:
But catching exceptions occasionally is discouraged, AFAIK...

norihiro-w:
how about a macro `OGS_RUNTIME_ERROR`? In the future, someone could add `OGS_LOGICAL_ERROR` or something else.

chleh:
OGS_RUNTIME_ERROR is OK.
But is there the chance that anyone of us will ever make a distinction between different kinds of exceptions thrown?

norihiro-w:
no idea yet

norihiro-w:
or shall we use `OGS_FATAL` now and rename it later if someone wants to use different exceptions?

norihiro-w:
updated. please check each commit

chleh:
perfect. :+1:

PR #1246, opened 2016-06-10
Remove std::unique_ptr for PETSc vec mat

norihiro-w:
It doesn't make sense to use std::unique_ptr for PETSc vec mat objects because Vec and Mat types are actually pointers, i.e. `typedef struct _p_Mat* Mat`. 

and it can mislead people that std::unique_ptr can delete a PETSc Mat object, which is not true.


chleh:
:fast_forward:

wenqing:
:+1:

Issue #1167, opened 2016-04-27
TODO list after merging Christoph's branch


norihiro-w:
Sorted by priority

(1) Code modularity
- [x] fix circular dependency #1200 (PR #1208)
- [ ] split FileIO #1164 (PR #1182 #1202 #1216 #1217 #1225)

(2) Better interface in core modules
- [x] remove #ifdef USE_EIGEN. (DenseMatrix will be replaced later by Tom) (PR #1218)
- [x] remove unused Sparse stuff in MathLib/LinAlg (PR #1219)
- [x] consistent function names, e.g. getN and getNum #1220 (PR #1243)
- [ ] make GeoLib related stuff in MeshLib to MeshGeoToolsLib into MeshLib #1215 (PR #1242 and more )
- [x] Error handling: Replace `std::abort()` with `exception` #1165 (PR #1245)
- [ ] Cleanup mat/vec interfaces (see also #1077). Currently algebraic operations are given in both BLAS namespace and in mat/vec classes, which is confusing users. The functions should be gathered in one place, either as class functions or in BLAS namesake (The name BLAS is confusing and should be renamed). (PR #1247 and more)

(3) Cleanup for faster compiling, etc.
- [x] Global typedef for sparse matrix and vectors #1134
- [ ] Minimize usage of template techniques for better developer friendly API and shorter compiling time (To be specified. Removing some templates is already given above. ) #1137
- [ ] Customize selection of libraries in cmake #1142
- [ ] rename `NumLib/ODESolver` to `NumLib/TimeIntegration`.
- [ ] clean up includes. @endJunction has some fancy tool for it. Maybe by that we already gain some compilation speed (at least if not building from scratch).

(4) Essential feature addition / deletion 
- [ ] Data output for each mesh, not for each process (I guess this influences the process interface)
- [ ] Global typedef for integer and real, e.g. `OgsInt`, `OgsReal`. This is necessary for supporting `long long` (e.g. more than 2 billion DoFs (e.g. 20 million HEX27 elements for mechanics in 3D) ) and quad precision, which are also supported by PETSc, Lis, and PARDISO.
- [ ] Introducing Parameters class (I am thinking something like TCLAP but parsing ConfigTree). This enables us to include documentation of parameters into the executable (and force developers to write short documentation for each parameter) so that we can print help of all parameters with `ogs --help-parameters`.

(5) Misc. feature addition / deletion
[...]

PR #1247, opened 2016-06-10
Clean up global Mat Vec class interface

norihiro-w:
follow up of #1246 

moved all linear algebraic operations in global mat vec classes to BLAS namespace. The next step is renaming "BLAS" to something else.

### DIFF
--- a/MathLib/LinAlg/UnifiedMatrixSetters.cpp
+++ b/MathLib/LinAlg/UnifiedMatrixSetters.cpp
@@ -89,9 +89,27 @@ void setVector(Eigen::VectorXd& v, MatrixVectorTraits<Eigen::VectorXd>::Index co
 namespace MathLib
 {
 
-double norm(PETScVector const& x)
+double norm(PETScVector const& x, MathLib::VecNormType nmtype = MathLib::VecNormType::NORM2)
 {
-    return x.getNorm();
+    NormType petsc_norm = NORM_1;
+    switch(nmtype)
+    {
+        case MathLib::VecNormType::NORM1:
+            petsc_norm = NORM_1;
+            break;
+        case MathLib::VecNormType::NORM2:
+            petsc_norm = NORM_2;
+            break;
+        case MathLib::VecNormType::INFINITY_N:
+            petsc_norm = NORM_INFINITY;
+            break;
+        default:
+            break;
+    }
+
+    PetscScalar norm = 0.;
+    VecNorm(x.getRawVector(), petsc_norm, &norm);
+    return norm;
 }
 
 void setVector(PETScVector& v,
--- a/MeshLib/IO/VtkIO/VtuInterface.cpp
+++ b/MeshLib/IO/VtkIO/VtuInterface.cpp
@@ -84,9 +84,9 @@ bool VtuInterface::writeToFile(std::string const &file_name)
 
     const std::string file_name_rank = file_name_base + "_"
                                        + std::to_string(mpi_rank) + ".vtu";
-    const bool vtu_status_i = writeVTU<vtkXMLUnstructuredGridWriter>(file_name_rank);
+    bool vtu_status_i = writeVTU<vtkXMLUnstructuredGridWriter>(file_name_rank);
     bool vtu_status = false;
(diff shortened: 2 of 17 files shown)


chleh:
Before going on we should coordinate things, because I aslo have a private branch with 1000+ lines of changes in matrix/vector classes.

norihiro-w:
@chleh what have you changed? My changes are as already explained in #1167

chleh:
This PR is OK, but before renaming BLAS, please let's talk at first.
Actually I removed nearly all methods from the global matrix and vector classes. Maybe that's not the best solution. But I thought, if algebra is done by stand-alone functions, reading and writing could also be done by those.

MathLib/LinAlg/UnifiedMatrixSetters.cpp
```diff
@@ -89,9 +89,27 @@ void setVector(Eigen::VectorXd& v, MatrixVectorTraits<Eigen::VectorXd>::Index co
 namespace MathLib
 {
 
-double norm(PETScVector const& x)
+double norm(PETScVector const& x, MathLib::VecNormType nmtype = MathLib::VecNormType::NORM2)
```
chleh: Suggest to remove this method from this file (maybe in a later PR).
I am against default arguments. They only make the code more error-prone.
I have some code regarding norms in a private branch. I'll make a PR in the next few weeks.

Tests/MathLib/TestGlobalMatrixInterface.cpp
```diff
@@ -94,24 +94,12 @@ void checkGlobalMatrixInterfaceMPI(T_MATRIX &m, T_VECTOR &v)
 
     MathLib::finalizeMatrixAssembly(m);
 
-    // Multiply by a vector
-    v = 1.;
-    const bool deep_copy = false;
-    T_VECTOR y(v, deep_copy);
-    m.multiply(v, y);
-
-    ASSERT_EQ(sqrt(3*(3*3 + 7*7)), y.getNorm());
```
chleh: Shall we better keep the test and rewrite it in terms of `BLAS::`? Or shall we remove it entirely?

MeshLib/IO/VtkIO/VtuInterface.cpp
```diff
@@ -86,7 +86,7 @@ bool VtuInterface::writeToFile(std::string const &file_name)
                                        + std::to_string(mpi_rank) + ".vtu";
     const bool vtu_status_i = writeVTU<vtkXMLUnstructuredGridWriter>(file_name_rank);
     bool vtu_status = false;
-    MPI_Allreduce(&vtu_status_i, &vtu_status, 1, MPI_C_BOOL, MPI_LAND, PETSC_COMM_WORLD);
+    MPI_Allreduce(const_cast<bool*>(&vtu_status_i), &vtu_status, 1, MPI_C_BOOL, MPI_LAND, PETSC_COMM_WORLD);
```
chleh: I think it would be better to declare `vtu_status_i` non-const above.

chleh:
after questions have been answered :+1:

Tests/MathLib/TestGlobalMatrixInterface.cpp
norihiro-w: These tests have been reactivated using BLAS

MathLib/LinAlg/UnifiedMatrixSetters.cpp
norihiro-w: changing the behavior is not a part of this PR, though I agree having a default value here is no help.

TomFischer:
Looks good. :fast_forward:

MathLib/LinAlg/UnifiedMatrixSetters.cpp
chleh: Sorry, I missed that.

Tests/MathLib/TestGlobalMatrixInterface.cpp
chleh: Sorry, I missed that.

PR #1248, opened 2016-06-13
[NL] Apply all Dirichlet BCs at once.

TomFischer:
Saves time.

### DIFF
--- a/NumLib/ODESolver/TimeDiscretizedODESystem.h
+++ b/NumLib/ODESolver/TimeDiscretizedODESystem.h
@@ -359,10 +359,14 @@ public:
             _ode.getKnownSolutions(_time_disc.getCurrentTime());
 
         if (known_solutions) {
+            using IndexType = typename MathLib::MatrixVectorTraits<Matrix>::Index;
+            std::vector<IndexType> ids;
+            std::vector<double> values;
             for (auto const& bc : *known_solutions) {
-                // TODO maybe it would be faster to apply all at once
-                MathLib::applyKnownSolution(A, rhs, x, bc.ids, bc.values);
+                std::copy(bc.ids.cbegin(), bc.ids.cend(), std::back_inserter(ids));
+                std::copy(bc.values.cbegin(), bc.values.cend(), std::back_inserter(values));
             }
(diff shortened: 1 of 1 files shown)


endJunction:
Looks good. Could you post some timings, like for the "Unstrut part" master takes ~x secs... (Only what you already have.)
Would you try to sort the ids (simultaneously with the values) and measure the times one more time please?

NumLib/ODESolver/TimeDiscretizedODESystem.h
```diff
@@ -359,10 +359,13 @@ class TimeDiscretizedODESystem<Matrix, Vector,
             _ode.getKnownSolutions(_time_disc.getCurrentTime());
 
         if (known_solutions) {
+            decltype((*known_solutions)[0].ids) ids;
```
chleh: That looks very awkward.

chleh:
A different option is to extend `MathLib::applyKnownSolution` s.t. it accepts the whole vector `known_solutions` and loop over it there.

TomFischer:
As @endJunction suggested I will try the sort the ids and values and will produce timings.

TomFischer:
I did some time measurements.
Without the commit 6de4fb2 `applyKnownSolution`  took 37078 seconds (10:20 h). The reduce the time to circa 8 seconds. The sorting and application of the id and value vector the time for `applyKnownSolution` is 12 seconds.

TomFischer:
> A different option is to extend MathLib::applyKnownSolution s.t. it accepts the whole vector known_solutions and loop over it there.

The linear algebra functionality deals with matrices and vectors. I don't like the idea to offer an interface that accepts boundary condition object in the linear algebra data structures. As long as this is not a bottleneck anymore I would keep the implementation as simple as possible.

endJunction:
:fast_forward: 
I'd replace the `decltype` things with explicit types, which are known and fixed. The `decltype` is usually used in template metaprogramming where the types can be different for the same algorithm...

chleh:
:+1:

wenqing:
:+1:  It is also better for MPI stuff.

PR #1252, opened 2016-06-15
Quickfix for confusing error messages from ConfigTree

chleh:
The problem is the following:
1. Something goes wrong → exception thrown
2. stack unwinds
3. ConfigTree gets destroyed
4. The check detects that it has not been processed entirely → throws again, leading to an error message from a point that would have been OK if the (1.) would not have occured.

This PR fixes point no. (4.).


norihiro-w:
:+1:

TomFischer:
:+1:

PR #1254, opened 2016-06-15
Output iteration results

chleh:
... after every iteration of the nonlinear solver.

This adds PR adds a callback to the nonlinear solvers' `solve()` method that can be used for output.
The output file name is the same as usual, but with `_iter_#ITER` appended.

The places in the code are not performance-critical, so it is runtime-checked if there shall be output. It can bve configured with the tag `<output_iteration_results>` in the output config.

### DIFF
--- a/Applications/ApplicationsLib/UncoupledProcessesTimeLoop.h
+++ b/Applications/ApplicationsLib/UncoupledProcessesTimeLoop.h
@@ -155,7 +155,7 @@ private:
 
     //! Solves one timestep for the given \c process.
     bool solveOneTimeStepOneProcess(
-            GlobalVector& x, std::size_t timestep, double const t, double const delta_t,
+            GlobalVector& x, std::size_t const timestep, double const t, double const delta_t,
             SingleProcessData& process_data,
             Process& process, ProcessLib::Output const& output_control);
 
--- a/Applications/ApplicationsLib/UncoupledProcessesTimeLoop.h
+++ b/Applications/ApplicationsLib/UncoupledProcessesTimeLoop.h
@@ -155,9 +155,9 @@ private:
 
     //! Solves one timestep for the given \c process.
     bool solveOneTimeStepOneProcess(
-            GlobalVector& x, double const t, double const delta_t,
+            GlobalVector& x, std::size_t timestep, double const t, double const delta_t,
             SingleProcessData& process_data,
-            Process& process);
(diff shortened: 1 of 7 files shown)


ProcessLib/Output.h
```diff
@@ -58,6 +58,15 @@ class Output
             const double t,
             GlobalVector const& x);
 
+    //! Writes output for the given \c process.
+    //! To be used for debug output after an iteration of the nonlinear solver.
+    void doOutputIteration(
```
norihiro-w: I would specify this is for nonlinear iterations, i.e. `doOutputNonlinearIteration`

ProcessLib/Output.h
```diff
@@ -80,11 +89,13 @@ class Output
         MeshLib::IO::PVDFile pvd_file;
     };
 
-    Output(std::string const& prefix)
+    Output(std::string const& prefix, bool output_iteration_results)
```
norihiro-w: output_nonlinear_iteration_results

ProcessLib/Output.h
chleh: Should the suffix then be `_nliter_#ITER`?

Applications/ApplicationsLib/UncoupledProcessesTimeLoop.h
```diff
@@ -156,9 +156,9 @@ class UncoupledProcessesTimeLoop
 
     //! Solves one timestep for the given \c process.
     bool solveOneTimeStepOneProcess(
-            Vector& x, double const t, double const delta_t,
-            SingleProcessData& process_data,
-            Process& process);
+        Vector& x, std::size_t timestep, double const t, double const delta_t,
```
endJunction: timestep could be const, isn't it?

ProcessLib/Output.cpp
```diff
@@ -137,4 +142,27 @@ doOutputLastTimestep(Process const& process,
         doOutputAlways(process, timestep, t, x);
 }
 
+void Output::doOutputIteration(
+    Process const& process, unsigned timestep, const double t,
```
endJunction: timestep should be const

endJunction:
After answering the questions :fast

endJunction:
In the "[AppL] cleanup" a file Applications/ApplicationsLib/UncoupledProcessesTimeLoop.h.orig is removed. Please remove it from the repository. It is introduced in the first commit "optionally do output ..."

PR #1255, opened 2016-06-15
add missing dependency for dynamic linking

norihiro-w:
fix dynamic linking errors on mac. Note that STATIC is explicitly given to some DE libraries due to their circular dependency.

### DIFF
--- a/Applications/DataExplorer/Base/CMakeLists.txt
+++ b/Applications/DataExplorer/Base/CMakeLists.txt
@@ -41,7 +41,6 @@ add_library(QtBase
 
 target_link_libraries(QtBase
     Qt4::QtGui
-    logog
     GeoLib)
 
 set_property(TARGET QtBase PROPERTY FOLDER "DataExplorer")
--- a/Applications/DataExplorer/Base/CMakeLists.txt
+++ b/Applications/DataExplorer/Base/CMakeLists.txt
@@ -39,6 +39,9 @@ add_library(QtBase
     ${HEADERS}
 )
 
-target_link_libraries(QtBase PUBLIC Qt4::QtGui)
+target_link_libraries(QtBase
+    Qt4::QtGui
+    logog
+    GeoLib)
 
 set_property(TARGET QtBase PROPERTY FOLDER "DataExplorer")
--- a/Applications/DataExplorer/DataView/DiagramView/CMakeLists.txt
+++ b/Applications/DataExplorer/DataView/DiagramView/CMakeLists.txt
@@ -39,6 +39,12 @@ add_library(QtDiagramView
     ${UIS}
 )
 
-target_link_libraries(QtDiagramView Qt4::QtGui)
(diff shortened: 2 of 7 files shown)


endJunction:
:fast_forward:

norihiro-w:
it seems setting `BUILD_SHARED_LIBS=ON` is not enough to generate and use DLLs with MSVC. One needs to introduce EXPORT keywords in the code ...

bilke:
Yes to create DLLs on MSVC you have to add lots of ugly export statements...
If it is not required because of very important reasons I would skip DLL building on MSVC..

norihiro-w:
We wanted to allow dynamic linking to reduce compiling time. if it's not a realistic option on MSVC, we can forget about it.

norihiro-w:
https://blog.kitware.com/create-dlls-on-windows-without-declspec-using-new-cmake-export-all-feature/

chleh:
:+1: from my side. @bilke, what do you say about @norihiro-w's cmake export comment?

Applications/DataExplorer/DataView/DiagramView/CMakeLists.txt
```diff
@@ -39,6 +39,12 @@ add_library(QtDiagramView
     ${UIS}
 )
 
-target_link_libraries(QtDiagramView Qt4::QtGui)
+target_link_libraries(QtDiagramView
+    Qt4::QtGui
+    logog
+    BaseLib
+    GeoLib
```
TomFischer: `logog` is linked in `BaseLib` which is linked in `MathLib` which is linked in `GeoLib`. Maybe it is enough to link against `GeoLib`?

Applications/DataExplorer/Base/CMakeLists.txt
```diff
@@ -39,6 +39,9 @@ add_library(QtBase
     ${HEADERS}
 )
 
-target_link_libraries(QtBase PUBLIC Qt4::QtGui)
+target_link_libraries(QtBase
+    Qt4::QtGui
+    logog
+    GeoLib)
```
TomFischer: I think `logog` can be removed here, since `GeoLib` is linked with `MathLib` which is linked with `BaseLib` which is linked with `logog`.

MaterialsLib/CMakeLists.txt
```diff
@@ -4,3 +4,7 @@ GET_SOURCE_FILES(SOURCES_ADSORPTION Adsorption)
 set(SOURCES ${SOURCES} ${SOURCES_ADSORPTION})
 
 add_library(MaterialsLib ${SOURCES} )
+target_link_libraries(MaterialsLib
+    logog
+    BaseLib
```
TomFischer: `BaseLib` is linked with `logog` already. So there is no need to link with `logog`.

TomFischer:
When minor issues are resolved: :+1:

PR #1264, opened 2016-06-17
Cleanup

endJunction:
Some cleanups mostly suggested by clang-tidy.
I'll add the used clang-tidy tests shortly; waiting for compilation result.

I think commit-wise review is better.

Nice side effect is that the code is a little faster and the binary a little smaller.

### DIFF
--- a/Applications/FileIO/GMSInterface.cpp
+++ b/Applications/FileIO/GMSInterface.cpp
@@ -140,10 +140,10 @@ void GMSInterface::writeBoreholesToGMS(const std::vector<GeoLib::Point*>* statio
     out << "name" << "\t" << std::fixed << "X" << "\t" << "Y"  << "\t" << "Z" <<  "\t" <<
     "soilID" << "\n";
 
-    for (std::size_t j = 0; j < stations->size(); j++)
+    for (auto station_as_point : *stations)
     {
         GeoLib::StationBorehole* station =
-                static_cast<GeoLib::StationBorehole*>((*stations)[j]);
+                static_cast<GeoLib::StationBorehole*>(station_as_point);
         std::vector<GeoLib::Point*> profile = station->getProfile();
         std::vector<std::string> soilNames  = station->getSoilNames();
         //std::size_t idx = 0;
--- a/GeoLib/Polygon.cpp
+++ b/GeoLib/Polygon.cpp
@@ -359,10 +359,10 @@ void Polygon::ensureCWOrientation ()
 
 #if __GNUC__ <= 4 && (__GNUC_MINOR__ < 9)
 void Polygon::splitPolygonAtIntersection(
-    std::list<Polygon*>::iterator polygon_it)
+    const std::list<Polygon*>::iterator& polygon_it)
 #else
 void Polygon::splitPolygonAtIntersection(
-    std::list<Polygon*>::const_iterator polygon_it)
+    const std::list<Polygon*>::const_iterator& polygon_it)
 #endif
 {
     GeoLib::Polyline::SegmentIterator seg_it0((*polygon_it)->begin());
@@ -409,7 +409,7 @@ void Polygon::splitPolygonAtIntersection(
     splitPolygonAtIntersection(polygon1_it);
 }
 
-void Polygon::splitPolygonAtPoint (std::list<GeoLib::Polygon*>::iterator polygon_it)
+void Polygon::splitPolygonAtPoint (const std::list<GeoLib::Polygon*>::iterator& polygon_it)
 {
     std::size_t n((*polygon_it)->getNumberOfPoints() - 1), idx0(0), idx1(0);
     std::vector<std::size_t> id_vec(n);
--- a/GeoLib/Polygon.cpp
+++ b/GeoLib/Polygon.cpp
@@ -125,9 +125,9 @@ std::vector<GeoLib::Point> Polygon::getAllIntersectionPoints(
 {
     std::vector<GeoLib::Point> intersections;
     GeoLib::Point s;
-    for (auto seg_it(begin()); seg_it != end(); ++seg_it)
+    for (auto&& seg_it : *this)
     {
-        if (GeoLib::lineSegmentIntersect(*seg_it, segment, s)) {
+        if (GeoLib::lineSegmentIntersect(seg_it, segment, s)) {
             intersections.push_back(s);
         }
     }
@@ -308,8 +308,8 @@ void Polygon::ensureCWOrientation ()
     // rotate copied points into x-y-plane
     GeoLib::rotatePointsToXY(tmp_polygon_pnts);
 
-    for (std::size_t k(0); k < tmp_polygon_pnts.size(); k++)
-        (*(tmp_polygon_pnts[k]))[2] = 0.0; // should be -= d but there are numerical errors
+    for (auto & tmp_polygon_pnt : tmp_polygon_pnts)
+        (*tmp_polygon_pnt)[2] = 0.0; // should be -= d but there are numerical errors
 
     // *** get the left most upper point
     std::size_t min_x_max_y_idx (0); // for orientation check
--- a/GeoLib/Polygon.h
+++ b/GeoLib/Polygon.h
@@ -134,12 +134,13 @@ private:
     void ensureCWOrientation ();
 
 #if __GNUC__ <= 4 && (__GNUC_MINOR__ < 9)
-    void splitPolygonAtIntersection(std::list<Polygon*>::iterator polygon_it);
+    void splitPolygonAtIntersection(
+        const std::list<Polygon*>::iterator& polygon_it);
 #else
     void splitPolygonAtIntersection(
-        std::list<Polygon*>::const_iterator polygon_it);
+        const std::list<Polygon*>::const_iterator& polygon_it);
 #endif
-    void splitPolygonAtPoint (std::list<Polygon*>::iterator polygon_it);
+    void splitPolygonAtPoint (const std::list<Polygon*>::iterator& polygon_it);
     std::list<Polygon*> _simple_polygon_list;
     AABB _aabb;
 };
--- a/GeoLib/Station.cpp
+++ b/GeoLib/Station.cpp
@@ -15,6 +15,7 @@
 #include "Station.h"
 
 #include <cstdlib>
+#include <utility>
 
 #include <logog/include/logog.hpp>
 
@@ -23,12 +24,12 @@
 namespace GeoLib
 {
 Station::Station(double x, double y, double z, std::string name) :
-    Point (x,y,z), _name(name), _type(Station::StationType::STATION),
+    Point (x,y,z), _name(std::move(name)), _type(Station::StationType::STATION),
     _station_value(0.0), _sensor_data(nullptr)
 {}
 
 Station::Station(Point* coords, std::string name) :
-    Point (*coords), _name(name), _type(Station::StationType::STATION),
+    Point (*coords), _name(std::move(name)), _type(Station::StationType::STATION),
     _station_value(0.0), _sensor_data(nullptr)
 {}
 
--- a/GeoLib/Station.cpp
+++ b/GeoLib/Station.cpp
@@ -23,13 +23,13 @@
 
 namespace GeoLib
 {
-Station::Station(double x, double y, double z, std::string name) :
-    Point (x,y,z), _name(std::move(name)), _type(Station::StationType::STATION),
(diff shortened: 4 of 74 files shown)


TomFischer:
Looks good. :+1:

GeoLib/Station.cpp
```diff
@@ -23,12 +24,12 @@
 namespace GeoLib
 {
 Station::Station(double x, double y, double z, std::string name) :
-    Point (x,y,z), _name(name), _type(Station::StationType::STATION),
+    Point (x,y,z), _name(std::move(name)), _type(Station::StationType::STATION),
```
norihiro-w: is move really needed here?

GeoLib/Polygon.cpp
```diff
@@ -125,9 +125,9 @@ std::vector<GeoLib::Point> Polygon::getAllIntersectionPoints(
 {
     std::vector<GeoLib::Point> intersections;
     GeoLib::Point s;
-    for (auto seg_it(begin()); seg_it != end(); ++seg_it)
+    for (auto && seg_it : *this)
```
chleh: I'm wondering about the `&&`...

GeoLib/Polygon.h
```diff
@@ -134,12 +134,12 @@ class Polygon : public Polyline
     void ensureCWOrientation ();
 
 #if __GNUC__ <= 4 && (__GNUC_MINOR__ < 9)
-    void splitPolygonAtIntersection(std::list<Polygon*>::iterator polygon_it);
+    void splitPolygonAtIntersection(const std::list<Polygon*>::iterator& polygon_it);
 #else
     void splitPolygonAtIntersection(
         std::list<Polygon*>::const_iterator polygon_it);
```
chleh: That case goes unmodified?

GeoLib/Station.cpp
```diff
@@ -23,12 +24,12 @@
 namespace GeoLib
 {
 Station::Station(double x, double y, double z, std::string name) :
```
chleh: Why not `const&` here?

MathLib/Integration/GaussLegendre.h
```diff
@@ -29,6 +29,15 @@ struct GaussLegendre {
     static const double W[Order];
 };
 
+template <> double const GaussLegendre<1>::X[1];
```
chleh: Looks like that should be in a cpp file.

MathLib/Integration/GaussLegendrePyramid.h
```diff
@@ -42,6 +42,12 @@ struct GaussLegendrePyramid<3> {
     static const double W[NPoints];
 };
 
+template <>
+const std::array<std::array<double, 3>, GaussLegendrePyramid<1>::NPoints>
+    GaussLegendrePyramid<1>::X;
+template <>
+double const GaussLegendrePyramid<1>::W[1];
```
chleh: mv to cpp file

MeshLib/MeshGenerators/MeshGenerator.cpp
```diff
@@ -36,9 +36,9 @@ std::vector<MeshLib::Node*> MeshGenerator::generateRegularNodes(
         for (std::size_t j = 0; j < vec_xyz_coords[1]->size(); j++)
         {
             const double y ((*vec_xyz_coords[1])[j]+origin[1]);
-            for (std::size_t k = 0; k < vec_xyz_coords[0]->size(); k++)
+            for (double k : *vec_xyz_coords[0])
```
chleh: k is a bad name for a double.

ProcessLib/Output.h
```diff
@@ -33,7 +33,7 @@ class Output
                         ::const_iterator;
 
     //! Opens a PVD file for each process.
-    void initialize(ProcessIter first, ProcessIter last);
+    void initialize(ProcessIter first, const ProcessIter& last);
```
chleh: I don't like the asymmetry.

ProcessLib/ProcessVariable.cpp
```diff
@@ -116,7 +116,7 @@ ProcessVariable::ProcessVariable(BaseLib::ConfigTree const& config,
 }
 
 ProcessVariable::ProcessVariable(ProcessVariable&& other)
-    : _name(std::move(other._name)),
+    : _name(other._name),
```
chleh: That's curious. Can you explain?

Tests/MeshLib/TestAddLayerToMesh.cpp
```diff
@@ -31,9 +31,9 @@ namespace AddLayerValidation
         ElementErrorFlag const flags[nErrorFlags] = {ElementErrorFlag::ZeroVolume,
         ElementErrorFlag::NonCoplanar, ElementErrorFlag::NonConvex,  ElementErrorFlag::NodeOrder};
         std::vector<ElementErrorCode> const codes (MeshLib::MeshValidation::testElementGeometry(mesh));
-        for (std::size_t i=0; i<codes.size(); ++i)
+        for (auto code : codes)
```
chleh: No ref here?

chleh:
from my side only minor things. afterwards :+1:

MathLib/Integration/GaussLegendre.h
endJunction: This is a declaration. In the cpp we have a definition...
It is resolving the warning (this one is for `W` but same is for `X`):
```
ogs/NumLib/Fem/Integration/IntegrationGaussRegular-impl.h:82:27: warning: instantiation of variable 'MathLib::GaussLegendre<4>::W' required here, but no definition is available [-Wundefined-var-template]
        weight *= Method::W[pos[d]];
                          ^
/home/naumov/w/ogs/MathLib/Integration/GaussLegendre.h:29:25: note: forward declaration of template entity is here
    static const double W[Order];
                        ^
/home/naumov/w/ogs/NumLib/Fem/Integration/IntegrationGaussRegular-impl.h:82:27: note: add an explicit instantiation declaration to suppress this warning if 'MathLib::GaussLegendre<4>::W' is explicitly instantiated in another translation unit
        weight *= Method::W[pos[d]];
```

Tests/MeshLib/TestAddLayerToMesh.cpp
endJunction: It's a the `code` is a `std::bitset<5>`, so no reference like for doubles or ints.

GeoLib/Station.cpp
endJunction: It was not before... but see the answer for the question on the next line (in the diff).

GeoLib/Station.cpp
endJunction: Because `name` is passed by value it is cheaper to move and we know that `std::string` has a move-ctor.
Maybe passing by const ref in first place would have been a better idea...

Changing parameter to const ref, because this is something we all understand well.

MathLib/Integration/GaussLegendre.h
chleh: I see.
Just curious: Would it suffice to instantiate via:
`template<unsigned ORDER> double const GaussLegendre<ORDER>::X[ORDER];`

GeoLib/Station.cpp
endJunction: Not changing. There is a default value for the string.

chleh: Can't default values be passed by const ref?

endJunction: Yes this didn't come to my mind.

MathLib/Integration/GaussLegendre.h
endJunction:
```
ogs/MathLib/Integration/GaussLegendre.h:32:62: error: default initialization of an object of const type 'const double [2]'
template <unsigned ORDER> double const GaussLegendre<ORDER>::X[Order];
```

In the cpp there are specializations...

endJunction:
I'll answer the move/rvalue related things later.

MathLib/Integration/GaussLegendre.h
```diff
@@ -29,16 +29,27 @@ struct GaussLegendre {
     static const double W[Order];
 };
 
-extern template <> double const GaussLegendre<1>::X[1];
```
chleh: `extern template double const GaussLegendre<1>::X[1];` could work (i.e. no extra `<>`).

endJunction: Nope
```
error: explicit specialization of 'X' after instantiation template <> double const GaussLegendre<1>::X[1] = {0.};
```

I'll set correct `#ifndef _MSC_VER`, and then it's OK. On the other hand one could move all definitions to the headers...

chleh: I'd prefer moving the definitions to the header in order to save the ifdef.

GeoLib/Station.cpp
```diff
@@ -23,13 +23,13 @@
 
 namespace GeoLib
 {
-Station::Station(double x, double y, double z, std::string name) :
-    Point (x,y,z), _name(std::move(name)), _type(Station::StationType::STATION),
+Station::Station(double x, double y, double z, std::string const& name) :
```
rinkk: I am right now preparing a PR such that station names can be changed later (this is a request for the new GW-book). This might be interfering with that request?

MeshLib/MeshEditing/FlipElements.cpp
```diff
@@ -56,7 +56,7 @@ std::unique_ptr<MeshLib::Mesh> createFlippedMesh(MeshLib::Mesh const& mesh)
     for (std::size_t i=0; i<n_elems; ++i)
         new_elems.push_back(createFlippedElement(*elems[i], new_nodes).release());
 
-    MeshLib::Properties new_props (mesh.getProperties());
+    const MeshLib::Properties& new_props (mesh.getProperties());
```
rinkk: A new mesh is generated here. Won't it couse memory issues if you pass a reference here, i.e. both meshes point to the same property vector?

Applications/FileIO/GMSInterface.cpp
```diff
@@ -140,10 +140,10 @@ void GMSInterface::writeBoreholesToGMS(const std::vector<GeoLib::Point*>* statio
     out << "name" << "\t" << std::fixed << "X" << "\t" << "Y"  << "\t" << "Z" <<  "\t" <<
     "soilID" << "\n";
 
-    for (std::size_t j = 0; j < stations->size(); j++)
+    for (auto j : *stations)
```
rinkk: can you put a more useful name here, "station" or so?

MeshLib/MeshEditing/FlipElements.cpp
endJunction: Thanks. I'll check this. But the tests are all passing so far...

GeoLib/Station.cpp
chleh: Shouldn't interfere: It's only about the convention "pass big objects by reference".

endJunction:
Jenkins, test this please!

GeoLib/Polygon.cpp
endJunction: `this` is a Polygon, which is a Polyline, which has begin() and end() returning a Polyline::SegmentIterator, a value, or more correctly a prvalue. This prvalue is captured in the rvalue reference 'seg_it'. This is what clang-tidy saw.

Timings:
original 0.02190
for-range-loop with auto 0.02128
for-range-loop with auto const& or auto&& 0.02111

ProcessLib/ProcessVariable.cpp
endJunction: Removing std::move is wrong here. I have no explanation why clang-tidy suggested this fix.
For LineSegment changes it is OK, because there are only POD's moved, but here it is a std::string, which is expensive to copy...
Reverting this change.

endJunction:
Finished from my side. Rolled back some changes. Learned few new things about moves. Explicit instantiation declaration issue is not fully understood, but will not be handled in this PR.

GeoLib/Polygon.cpp
```diff
@@ -362,7 +362,7 @@ void Polygon::splitPolygonAtIntersection(
     const std::list<Polygon*>::iterator& polygon_it)
 #else
 void Polygon::splitPolygonAtIntersection(
-    std::list<Polygon*>::const_iterator polygon_it)
+    const std::list<Polygon*>::const_iterator& polygon_it)
```
chleh: Really? I'd opt for passing iterators by value. But do what you like.

endJunction:
Windows didn't start build at all...

PR #1265, opened 2016-06-17
Pimpl GroundwaterFlowProcess

norihiro-w:
Tried pimpl in GroundwaterFlowProcess. see only 9ac9f0b 
It makes compiling 5sec faster (33 -> 28s, `touch&make -j4`), because ProjectData is not recompiled.


chleh:
Nice improvement in compilation times. Looking at that one commit, it seems to me that you don't pimpl a lot. Can't the same effect be achieved by moving the entire declaration of GWFlow to the cpp file and making `createGroundwaterFlowProcess` return a `unique_ptr<Process>`? If so I'd prefer that: It saves one level of indirection/splitting.

norihiro-w:
I don't think it has the same effect. The speedup was gained because there is no need to include GroundwaterFlowFEM.h in GroundwaterFlowProcess.h. That became possible by moving member variables of GroundwaterFlowProcess to the impl.

chleh:
> That became possible by moving member variables of GroundwaterFlowProcess to the impl.
But that would be possible, too if yu move the entire `GroundwaterFlowProcess` class to the cpp file (and return `unique_ptr<Process>` from the create function), wouldn't it?

norihiro-w:
i got what you mean. in that case, I would create another file having only the create function. Class declaration should be in a hear file.

norihiro-w:
here comparison: 0m22.975s (pimpl) vs 0m23.099s (create function)

norihiro-w:
compiling GWF process took around 12sec

PR #1267, opened 2016-06-17
Fix OGS_USE_LIS build

norihiro-w:
see commits


TomFischer:
For me lis now works again. Thanks @norihiro-w  :+1:

PR #1271, opened 2016-06-17
[PL] added debug note to ProcessOutput

chleh:
It is rather easy to get the extrapolation calls wrong. @endJunction asked me to provide a debug notice for that.

Please have a look if my note is understandable, and if not, please make some better suggestions.

### DIFF
--- a/NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.h
+++ b/NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.h
@@ -63,8 +63,18 @@ public:
 #endif
           ,
           _local_to_global(dof_table)
-    {}
-
+    {
+        /* Note in case the following assertion fails.
+         * If you copied the extrapolation code, for your processes from
+         * somewhere, note that the code from the groundwater flow process might
+         * not suit your needs: It is a special case and is therefore most
+         * likely too simplistic. You better adapt the extrapolation code from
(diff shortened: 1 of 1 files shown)


chleh:
I added an assertion, maybe that's already enough. @endJunction, does it always hold that `dof_table.getNumComponents() == 1` → one variable with one component? Or can it be different, e.g. for vectorial process variables, like displacement?

chleh:
If the assertion is already enough, maybe we can remove the comment again.

NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.h
```diff
@@ -63,8 +63,11 @@ class LocalLinearLeastSquaresExtrapolator
 #endif
           ,
           _local_to_global(dof_table)
-    {}
-
+    {
+        assert(dof_table.getNumComponents() == 1 &&
```
norihiro-w: getNumberOfComponents

norihiro-w:
👍

endJunction:
@chleh `DOF::getNumComponents()` returns 1 iff the total number of components is 1.

The names of many things in the LocalToGlobalIndexMap are wrong...

endJunction:
:+1: Works for me and checks correctly of the dof is wrong.
I think the assert is sufficient...

PR #1272, opened 2016-06-17
Nothrow ConfigTree destructor

chleh:
With the default strict config checks, the dtor of ConfigTree threw exceptions. There are many problems with throwing dtors. This PR makes sure, `~ConfigTree()` won't throw.
The drawback is, that now we have to manually check from time to time if all config stuff went fine. (And I really like to keep the strict config checks.)

### DIFF
--- a/BaseLib/ConfigTree.cpp
+++ b/BaseLib/ConfigTree.cpp
@@ -73,6 +73,7 @@ ConfigTree::~ConfigTree()
     try {
         checkAndInvalidate();
     } catch (std::exception& e) {
+        ERR("%s", e.what());
         configtree_destructor_error_messages.push_front(e.what());
     }
 }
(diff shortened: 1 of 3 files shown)


chleh:
Jenkins, test this please.

norihiro-w:
why can't you throw an exception while parsing, instead of throwing it at the end?

chleh:
At the end it is checked that all config settings have been read exectly once. I'd really like to keep this check, and the fatal error arising from it: It prevents from typos in optional settings and makes sure that input files don't contain unnecessary settings.

BaseLib/ConfigTree.cpp
```diff
@@ -65,8 +70,12 @@ ConfigTree(ConfigTree && other)
 
 ConfigTree::~ConfigTree()
 {
-    if (!std::uncaught_exception())
```
chleh: TODO: keep that.

endJunction:
:fast_forward: after your TODO update.

norihiro-w:
:+1: though I thinks it's enough to output error message without throwing exceptions.

endJunction:
... and merge after rebase.

PR #1275, opened 2016-06-20
Updating Schema paths

rinkk:


endJunction:
:fast_forward:

BaseLib/FileFinder.h
```diff
@@ -36,7 +36,8 @@ class FileFinder
     FileFinder()
     {
         addDirectory(".");
-        addDirectory(BuildInfo::source_path + "/FileIO");
+        addDirectory(BuildInfo::source_path + "/GeoLib/IO/XmlIO");
```
norihiro-w: I would move these setting to where calling this function (with extending this class). `FileFinder` is a general class and should not include setting for specific usage.

rinkk: That is *not* specific usage. FileFinder is designed so config files can be found even though they might be at different locations at a different time. XSD files are such config files, so they are added here. If you need to search different directories, feel free to add them.

chleh:
Different thing: Does it make sense to search files in the `BuildInfo::source_path`? The sources might have been deleted when you run the program, or the user might not have access to the sources at all (e.g., because you give somebody the executable)...

rinkk:
That's why "." is added as path. One is for distributed builds, one for development builds (well, two now after interfaces have been distributed among different libraries).

BaseLib/FileFinder.h
norihiro-w: i don't agree with you. Configuration files are application specific. If FileFinder is only used in that way, I would move this to ApplicationsLib and rename it to `ConfigFieFinder`.

PR #1276, opened 2016-06-21
Cleanup in MeshLib

TomFischer:
- Remove unused includes.
- Move ElementErrorCode from MeshQuality subfolder to Elements subfolder.


chleh:
:fast_forward:

rinkk:
Why move ElementErrorCode? Wouldn't it be better to have all problem detection functionality in the same directory?

TomFischer:
> Why move ElementErrorCode? Wouldn't it be better to have all problem detection functionality in the same directory?

This is exactly what the changes try to do. The element classes have a `validate` method, i.e., the `ElementErrorCode` is created from the elements. For this reason `ElementErrorCode` is moved from MeshQuality subfolder to Elements subfolder. In my point of view we should not mix validation with element quality. We have to check that an element is valid **before** we use it for quality evaluation (and other compuatitions).

PR #1278, opened 2016-06-21
[WIP] supporting dynamic linking on MSVC

norihiro-w:
Trying to support `cmake -DCMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=TRUE -DBUILD_SHARED_LIBS=TRUE` as explained in https://blog.kitware.com/create-dlls-on-windows-without-declspec-using-new-cmake-export-all-feature/


norihiro-w:
I close the PR. but it is still recored in issue

PR #1279, opened 2016-06-21
move DataHolderLib under Applications

norihiro-w:
as titled. DataHolderLib is used only in Applications


endJunction:
:fast_forward:

chleh:
:+1:

PR #1281, opened 2016-06-21
[NL] Remove unneded template parameter names.

endJunction:
For dynamic shape matrix policy the matrix/vector size parameters are unused.


norihiro-w:
👍

PR #1282, opened 2016-06-22
Change Surface search tolerance.

TomFischer:
Until now the default value of `std::numeric_limits<double>::epsilon()` was used. This PR changes the behaviour, i.e., the user can give some epsilon value different from the default value.

### DIFF
--- a/GeoLib/Surface.cpp
+++ b/GeoLib/Surface.cpp
@@ -153,15 +153,14 @@ bool Surface::isPntInBoundingVolume(MathLib::Point3d const& pnt) const
     return _bounding_volume->containsPoint(pnt);
 }
 
-bool Surface::isPntInSfc(MathLib::Point3d const& pnt) const
+bool Surface::isPntInSfc(MathLib::Point3d const& pnt, double eps) const
 {
     // Mutable _surface_grid is constructed if method is called the first time.
     if (_surface_grid == nullptr)
     {
         _surface_grid.reset(new SurfaceGrid(this));
     }
-    return _surface_grid->isPointInSurface(
-        pnt, std::numeric_limits<double>::epsilon());
+    return _surface_grid->isPointInSurface(pnt, eps);
 }
 
 const Triangle* Surface::findTriangle(MathLib::Point3d const& pnt) const
--- a/GeoLib/Surface.h
+++ b/GeoLib/Surface.h
@@ -71,9 +71,10 @@ public:
     /**
      * is the given point pnt located in the surface
      * @param pnt the point
+     * @param eps geometric tolerance for the search
      * @return true if the point is contained in the surface
      */
-    bool isPntInSfc(MathLib::Point3d const& pnt) const;
+    bool isPntInSfc(MathLib::Point3d const& pnt, double eps) const;
 
     /**
      * find a triangle in which the given point is located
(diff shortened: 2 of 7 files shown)


norihiro-w:
:+1:

GeoLib/Surface.h
```diff
@@ -71,9 +71,11 @@ class Surface final : public GeoObject
     /**
      * is the given point pnt located in the surface
      * @param pnt the point
+     * @param eps geometric tolerance for the search
      * @return true if the point is contained in the surface
      */
-    bool isPntInSfc(MathLib::Point3d const& pnt) const;
+    bool isPntInSfc(MathLib::Point3d const& pnt,
+                    double eps = std::numeric_limits<double>::epsilon()) const;
```
chleh: I am against default values. What would be the effort to remove the default?

GeoLib/Surface.cpp
```diff
@@ -153,15 +153,14 @@ bool Surface::isPntInBoundingVolume(MathLib::Point3d const& pnt) const
     return _bounding_volume->containsPoint(pnt);
 }
 
-bool Surface::isPntInSfc(MathLib::Point3d const& pnt) const
+bool Surface::isPntInSfc(MathLib::Point3d const& pnt, double eps) const
```
chleh: @endJunction would say: "const correctness" :smile:

MeshGeoToolsLib/MeshNodeSearcher.cpp
```diff
@@ -39,10 +39,10 @@ MeshNodeSearcher::MeshNodeSearcher(MeshLib::Mesh const& mesh,
         _mesh(mesh), _mesh_grid(_mesh.getNodes().cbegin(), _mesh.getNodes().cend()),
         _search_length(0.0), _search_all_nodes(search_all_nodes)
 {
-    DBUG("Constructing MeshNodeSearcher obj.");
+    INFO("Constructing MeshNodeSearcher obj.");
```
chleh: Is it useful to raise this loglevel? (The raise below probably is)

MeshGeoToolsLib/MeshNodesAlongSurface.h
```diff
@@ -39,7 +39,8 @@ class MeshNodesAlongSurface
      * @param mesh_nodes Nodes the search will be performed on.
      * @param sfc Along the GeoLib::Surface sfc the mesh nodes are searched.
      */
-    MeshNodesAlongSurface(MeshLib::Mesh const& mesh, GeoLib::Surface const& sfc, bool search_all_nodes = true);
+    MeshNodesAlongSurface(MeshLib::Mesh const& mesh, GeoLib::Surface const& sfc,
+                          double epsilon, bool search_all_nodes = true);
```
chleh: Looks like the docs need some update. There's another default value here, but that's probably for another PR.

GeoLib/Surface.cpp
TomFischer: Yes, I know. But in my point of view `eps` can not be changed anyway.

GeoLib/Surface.h
TomFischer: Removed default value.

chleh:
:+1:

PR #1284, opened 2016-06-22
Move XmlLutReader to DE/VtkVis to avoid circulary dependency

norihiro-w:
Currently Applications/FileIO and DE/VtkVis are dependent each other because XmlLutReader.h requires  VtkColorLookupTable defined in VtkVis. This PR proposes moving XmlLutReader.h to VtkVis/IO. Let me know if there is a better solution for this.


rinkk:
Can you specify how there is a circular dependency?  
I'd rather keep external format IO files in FileIO.

norihiro-w:
@rinkk click numbers next to arrows in https://doxygen.opengeosys.org/dir_9c00a509492fd290454a18350731207e.html

rinkk:
I can fix the reader so it'd not be necessary to move the file.  
However, I would first like to discuss in a larger round if we actually need to remove circular dependencies in Data Explorer because it looks like there'll be some major issues that'd take a lot longer to solve. Maybe we can do that tomorrow when Lars is back?

norihiro-w:
Having a meeting tomorrow is fine for me. 

IMO it doesn't matter if circular dependencies exist in DE, but not with outside of DE, i.e. FileIO.

norihiro-w:
@rinkk can you tell us when you are going to move the file?

rinkk:
I am currently doing some urgent proposal writing. Sorry.

rinkk:
I'll upload my solution tomorrow.

PR #1286, opened 2016-06-22
rename MathLib::BLAS

norihiro-w:
as issued already. I'd like to start implementing it. 
My first suggestion is renaming BLAS namespace to LinAlg. Let me know if you have other ideas. I also think removing the second namespace and keeping only MathLib is ok.


NumLib/ODESolver/MatrixTranslator.h
```diff
@@ -117,25 +117,25 @@ class MatrixTranslatorGeneral<Matrix, Vector, ODESystemTag::FirstOrderImplicitQu
     //! Computes \f$ A = M \cdot \alpha + K \f$.
     void computeA(Matrix const& M, Matrix const& K, Matrix& A) const override
     {
-        namespace BLAS = MathLib::BLAS;
+        using namespace MathLib::LinAlg;
```
chleh: I'd prefere a namespace alias over `using namespace`.

chleh:
I'd definitely keep the second namespace, because it contains lots of standalone functions.

Maybe `MathLib::MatVec`, because the namespace also contains `copy` and `finalizeAssembly`.

norihiro-w:
@endJunction @wenqing you don't have no preference?

norihiro-w:
and @TomFischer ?

wenqing:
@norihiro-w How about a full name, 'LinearAlgbra'?

endJunction:
namespace alias to MathLib::LinAlg or LinearAlgebra or maybe BasicLinearAlgebra (not BLAS) or LinearAlgebraBase. But I'm fine with other names too.

TomFischer:
My personal favourite is `MathLib::LinAlg` which also reflects the folder structure and for this reason is easy to remember.

norihiro-w:
Thanks, guys. Then I would take MathLib::LinAlg if there is no serious objection.

wenqing:
Since most of you prefer to use MathLib::LinAlg. I am fine with it.

norihiro-w:
renaming is done

TomFischer:
Looks good. :+1:

norihiro-w:
@endJunction I don't see a reason why this branch has to be rebased. There is no commit after 5406f523 (base of this branch) touching related parts (MathLib/LinAlg, NumLib, ApplicationsLib/UncoupledProcessesTimeLoop), except for #1296 where only comments in some NumLib code were changed. Can you provide me specific reasons why rebase is required for this branch?

chleh:
@endJunction wants to retain a completely linear history for the master branch. That's why he wants everybody to (in the best case always) rebase before merging a PR.
Maybe we can get both, easy merging and linear history, by introducing a `pullrequest` branch. Then we could make PRs to the `pullrequest` branch, merge them after code review without having to rebase. Afterwards some script in Jenkins could always do a rebase and merge that into `master`. Then `master` would always have an entirely linear history automatically. What do you (especially @endJunction) think?

norihiro-w:
@endJunction may I merge this now or do you have some comments?

PR #1287, opened 2016-06-22
fix internally setting BOOST_ROOT from env var

norihiro-w:
fix a problem that a environmental variable BOOST_ROOT doesn't work

### DIFF
--- a/scripts/cmake/ExternalProjectBoost.cmake
+++ b/scripts/cmake/ExternalProjectBoost.cmake
@@ -13,7 +13,7 @@ if(Boost_FOUND)
 endif()
 
 if(NOT DEFINED BOOST_ROOT AND DEFINED ENV{BOOST_ROOT})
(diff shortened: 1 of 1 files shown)


endJunction:
@norihiro-w Could you give a short description of the problem, or maybe how to reproduce it, or any hint what was going wrong before?

norihiro-w:
Previously in CMake, BOOST_ROOT var was set to "$ENV{BOOST_ROOT}; TRUE".

norihiro-w:
I think TRUE is simply unnecessary there as far as I read the documentation of SET()

bilke:
Yes the `TRUE` is wrong here.

endJunction:
Just a suggestion, although it is disputed, I would have rebased before the merge to check if there are any incompatibilities (on compilation level/test level).

norihiro-w:
generally yes but I don't think that is necessary for this tiny fix.

PR #1289, opened 2016-06-22
Moving CSV-/Raster-interface back to FileIO

rinkk:
As discussed, interfaces for external file formats are moved back into Applications/FileIO.

### DIFF
--- a/MeshGeoToolsLib/GeoMapper.cpp
+++ b/MeshGeoToolsLib/GeoMapper.cpp
@@ -19,7 +19,7 @@
 
 #include <logog/include/logog.hpp>
 
-#include "GeoLib/IO/AsciiRasterInterface.h"
+#include "Applications/FileIO/AsciiRasterInterface.h"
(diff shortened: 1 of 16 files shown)


MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -19,7 +19,7 @@
 
 #include <logog/include/logog.hpp>
 
-#include "GeoLib/IO/AsciiRasterInterface.h"
```
norihiro-w: don't make a general lib dependent on FileIO

rinkk: I'll fix the mapper class.

rinkk: Removed the IO include.

TomFischer:
The last commit can be merged into the first commit in the rebase process. :+1:

norihiro-w:
:+1:

PR #1293, opened 2016-06-22
Move FEFLOW interface to Applications/FileIO

TomFischer:
This removes dependencies of `MeshLib` from `GeoLib`.


norihiro-w:
:+1:

PR #1295, opened 2016-06-23
[GL] Correctly setting point names

rinkk:
This fixes an issue where point names would get lost when merging or saving geometries.

### DIFF
--- a/GeoLib/PointVec.h
+++ b/GeoLib/PointVec.h
@@ -101,6 +101,8 @@ public:
 
     std::string const& getItemNameByID(std::size_t id) const;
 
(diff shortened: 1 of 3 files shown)


TomFischer:
Looks good! :+1:

GeoLib/PointVec.h
```diff
@@ -101,6 +101,8 @@ class PointVec final : public TemplateVec<Point>
 
     std::string const& getItemNameByID(std::size_t id) const;
 
+    void setNameForElement(std::size_t id, std::string const& name);
```
norihiro-w: can you add override keyword?

norihiro-w:
👍

PR #1301, opened 2016-06-29
remove support of EigenXd as global matrix and vector types

norihiro-w:
as titled

### DIFF
--- a/MathLib/LinAlg/UnifiedMatrixSetters.h
+++ b/MathLib/LinAlg/UnifiedMatrixSetters.h
@@ -15,43 +15,14 @@
 #include <initializer_list>
 #include "MatrixVectorTraits.h"
 
-#ifdef OGS_USE_EIGEN
-
-// Dense Eigen matrix/vector //////////////////////////////////////////
-
-#include <Eigen/Core>
-
-namespace MathLib
-{
-
-void setMatrix(Eigen::MatrixXd& m,
-               std::initializer_list<double> values);
-
-void setMatrix(Eigen::MatrixXd& m, Eigen::MatrixXd const& tmp);
-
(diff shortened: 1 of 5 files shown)


MathLib/LinAlg/UnifiedMatrixSetters.h
```diff
@@ -13,45 +13,17 @@
 #define MATHLIB_UNIFIED_MATRIX_SETTERS_H
 
 #include <initializer_list>
+#include <Eigen/Eigen>
```
chleh: I think it would be enough to include <Eigen/Core>. Telling from the <Eigen/Eigen> header on my system it also includes linear solvers and other things.

chleh:
:+1:

MathLib/LinAlg/UnifiedMatrixSetters.h
```diff
@@ -76,11 +48,11 @@ void setMatrix(PETScMatrix& m,
 #elif defined(OGS_USE_EIGEN)
 
 // Sparse global EigenMatrix/EigenVector //////////////////////////////////////////
+#include "Eigen/EigenVector.h"
```
chleh: Why don't you leave it as a forward declaration?

TomFischer:
I understand that for large problems dense matrices can not be used. But is debugging with a sparse matrix just as easy as with a dense matrix?

norihiro-w:
@TomFischer I don't think it's difficult to debug with Eigen SparseMatrix. However if someone really wants to use dense matrices, it should be done within EigenMatrix class, i.e. setting dense matrix type to RowMatrixType.

TomFischer:
@norihiro-w Thanks for explanation. :fast_forward:

PR #1302, opened 2016-06-29
Removing circular dependency with XmlLutReader

rinkk:
Moves Color to DataHolderLib, introduces a new data holder for lookup-tables and removes the circular dependency.

### DIFF
--- a/Applications/DataExplorer/DataView/DiagramView/DetailWindow.cpp
+++ b/Applications/DataExplorer/DataView/DiagramView/DetailWindow.cpp
@@ -12,11 +12,12 @@
  *
  */
 
+#include "DetailWindow.h"
+
 #include <QFileDialog>
 #include <QSettings>
 
 #include "Applications/DataHolderLib/Color.h"
-#include "DetailWindow.h"
 #include "DiagramPrefsDialog.h"
 
 DetailWindow::DetailWindow(QWidget* parent) : QWidget(parent)
--- a/Applications/DataExplorer/DataView/DiagramView/DetailWindow.cpp
+++ b/Applications/DataExplorer/DataView/DiagramView/DetailWindow.cpp
@@ -12,12 +12,12 @@
  *
  */
 
-#include "Color.h"
-#include "DetailWindow.h"
-#include "DiagramPrefsDialog.h"
-
 #include <QFileDialog>
 #include <QSettings>
+
+#include "Applications/DataHolderLib/Color.h"
+#include "DetailWindow.h"
+#include "DiagramPrefsDialog.h"
 
 DetailWindow::DetailWindow(QWidget* parent) : QWidget(parent)
 {
@@ -124,7 +124,7 @@ void DetailWindow::resizeWindow()
 
 void DetailWindow::addList(DiagramList* list)
 {
-    GeoLib::Color const c = GeoLib::getRandomColor();
+    DataHolderLib::Color const c = DataHolderLib::getRandomColor();
     QColor colour(c[0], c[1], c[2]);
     this->addList(list, colour);
     resizeWindow();
--- a/Applications/DataExplorer/VtkVis/VtkAlgorithmProperties.cpp
+++ b/Applications/DataExplorer/VtkVis/VtkAlgorithmProperties.cpp
@@ -88,6 +88,8 @@ void VtkAlgorithmProperties::SetLookUpTable(const QString &array_name, const QSt
         colorLookupTable->setLookupTable(lut);
         SetLookUpTable(array_name, colorLookupTable);
     }
+    else
+        ERR ("Error reading color look-up table.");
 }
 
 void VtkAlgorithmProperties::SetScalarVisibility(bool on)
--- a/Applications/DataExplorer/VtkVis/VtkAlgorithmProperties.cpp
+++ b/Applications/DataExplorer/VtkVis/VtkAlgorithmProperties.cpp
@@ -20,6 +20,7 @@
 #include <vtkProperty.h>
 #include <vtkTexture.h>
 
+#include "Applications/DataHolderLib/ColorLookupTable.h"
 #include "VtkColorLookupTable.h"
 #include "XmlIO/Qt/XmlLutReader.h"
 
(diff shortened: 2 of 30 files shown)


Applications/DataExplorer/VtkVis/VtkStationSource.h
```diff
@@ -36,16 +34,16 @@ class VtkStationSource : public vtkPolyDataAlgorithm, public VtkAlgorithmPropert
...
 
     /// Returns the type of observation site represented in this source object
     GeoLib::Station::StationType getType() const
         { return static_cast<GeoLib::Station*>((*_stations)[0])->type(); };
 
     /// Sets a predefined color lookup table for the colouring of borehole stratigraphies
-    int setColorLookupTable(const std::string &filename)
-        { return GeoLib::readColorLookupTable(_colorLookupTable, filename); }
+//    int setColorLookupTable(const std::string &filename)
+//        { return GeoLib::readColorLookupTable(_colorLookupTable, filename); }
```
TomFischer: Is it possible to remove from 44-46?

Applications/DataExplorer/VtkVis/VtkVisPipeline.h
```diff
@@ -15,17 +15,15 @@
 #ifndef VTKVISPIPELINE_H
 #define VTKVISPIPELINE_H
 
-// ** INCLUDES **
-#include "BaseLib/Histogram.h"
+#include <QMap>
+#include <QVector>
 
-#include "Color.h"
+#include "BaseLib/Histogram.h"
```
TomFischer: Why do you need to include `BaseLib/Histogram.h`?

Applications/DataHolderLib/Color.cpp
```diff
@@ -63,7 +73,7 @@ int readColorLookupTable(std::map<std::string, Color> &colors, const std::string
 
     return 1;
 }
-
+*/
```
TomFischer: Is it possible to remove the commented code? If not, please add a comment why it is necessary to keep the code.

Applications/DataHolderLib/Color.h
```diff
@@ -20,22 +20,26 @@
...
+Color createColor(unsigned char r, unsigned char g, unsigned char b);
+
+Color createColor(unsigned char r, unsigned char g, unsigned char b, unsigned char a);
 
 /// Returns a random RGB colour.
 Color getRandomColor();
 
 /// Reads a color-lookup-table from a file and writes it to a map.
-int readColorLookupTable(std::map<std::string, GeoLib::Color> &colors, const std::string &filename);
+//int readColorLookupTable(std::map<std::string, GeoLib::Color> &colors, const std::string &filename);
```
TomFischer: Lines 34,35 can possibly be removed?!

Applications/DataHolderLib/ColorLookupTable.h
```diff
@@ -0,0 +1,71 @@
...
+ *    colour - RGBA value
+ *    name   - a name referencing the colour (such as a stratigraphic layer)
+ * Colours from the table can then be accessed using either id or name.
+ */
+class ColorLookupTable
+{
+public:
+    ColorLookupTable();
+
+    void setColor(double id, DataHolderLib::Color color);
```
TomFischer: Can `color` be passed by `const reference`? The same holds for `setColor()` in line 45.

Applications/DataExplorer/VtkVis/VtkAlgorithmProperties.cpp
```diff
@@ -81,8 +82,13 @@ void VtkAlgorithmProperties::SetLookUpTable(const QString &array_name, vtkLookup
...
 {
-    VtkColorLookupTable* colorLookupTable = FileIO::XmlLutReader::readFromFile(filename);
-    SetLookUpTable(array_name, colorLookupTable);
+    DataHolderLib::ColorLookupTable lut;
+    if (FileIO::XmlLutReader::readFromFile(filename, lut))
+    {
+        VtkColorLookupTable* colorLookupTable = VtkColorLookupTable::New();
+        colorLookupTable->setLookupTable(lut);
+        SetLookUpTable(array_name, colorLookupTable);
+    }
```
TomFischer: Maybe you can give an error message in case `readFromFile()` returns `false`?

TomFischer:
Apart from small questions: :+1:

Applications/DataExplorer/DataView/DiagramView/DetailWindow.cpp
```diff
@@ -12,13 +12,13 @@
...
 
-#include "Color.h"
-#include "DetailWindow.h"
-#include "DiagramPrefsDialog.h"
-
 #include <QFileDialog>
 #include <QSettings>
 
+#include "Applications/DataHolderLib/Color.h"
+#include "DetailWindow.h"
```
norihiro-w: include this at top

norihiro-w:
Changes in this PR are lots compared to just moving XmlLutReader to somewhere under DE (#1284). What is a reason you have to keep it in FileIO? XmlLutReader is used only in DE, isn't it?

TomFischer:
> Changes in this PR are lots compared to just moving XmlLutReader to somewhere under DE (#1284). What is a reason you have to keep it in FileIO? XmlLutReader is used only in DE, isn't it?

One reason is to keep IO functionality local. Another is that is now possible to use it in other utils without introducing new circular dependencies. External tools like Gocad store specifc color tables which could be reused by paraview or DE.

norihiro-w:
Thanks for the two points. However
- I'm not sure what you mean "local". Moving XmlLutReader under DE actually improves locality because it is only used there.
- For keeping quality and maintainability of codes, we should keep codes as simple as possible based on current usage. When it comes in the future that someone really needs another tool using color tables, the code should be modified accordingly.

TomFischer:
- By "local" I mean that all IO functionality is grouped together. If you search for a specific interface it should be sufficient to look into the FileIO.
- I have already a branch for reading Gocad SGrid data and I will use the color lookup table there in the near future.

norihiro-w:
I don't agree with the first point (because they are not local from a viewpoint of modularity), but the second point gives some reason to keep it in FileIO. thanks

Applications/DataExplorer/VtkVis/VtkStationSource.h
rinkk: I'd rather keep this for now as this is another part of the DE where lookup-tables are used but it would need a complete rewrite. I don't want to make this PR larger than it already is.

Applications/DataExplorer/VtkVis/VtkAlgorithmProperties.cpp
rinkk: added error msg

PR #1303, opened 2016-06-29
[Jenkins] Windows Pipeline job

bilke:
Builds existing Linux (gcc in Docker) build in parallel with new Windows build.

The new Windows build is a replacement for these legacy jobs which can be removed when the new job is working properly:

- OGS-6/Win
- OGS-6/Win-PRs
- OGS-6/Win_Utils
- OGS-6/Gui/Win
- OGS-6/Gui/Win-PRs

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,4 +1,11 @@
 node('master') {
     checkout scm
-    load 'scripts/jenkins/gcc.groovy'
+
+    def builders = [:]
+    builders['gcc'] = { load 'scripts/jenkins/gcc.groovy' }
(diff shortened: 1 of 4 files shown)


bilke:
@endJunction I checked that the pipeline is aborted on a failed step. If e.g. on Windows the CLI build fails, the remaining Windows pipeline is not executed but the Linux pipeline executes normally.

Although the [Stage View](https://svn.ufz.de:8443/job/OGS-6/job/ufz/branch/PR-1303/) is a little bit misleading (see build number 8): If one stage fails (e.g. a stage in the middle) it looks like the last stage has failed too but this is not the case here.

PR #1306, opened 2016-07-04
Removed VectorMatrixAssembler

### LINKED ISSUE #1277: Get rid of VectorMatrixAssembler
... and instead pass the global vectors and matrices and the d.o.f. table directly to the local assembler.

Pros:
* higher flexibility. e.g. it is easier to use auxiliary data, e.g. a global vector from a previous iteration
* (minor) local d.o.f. need not be passed as std::vector, but the local assembler can choose a better type
* no need for addToGlobal() anymore
* one level less indirection

Cons:
* slightly more code duplication. I.e., in every process there will be the lines
```
        auto const indices = detail::getIndices(id, _data_pos);
        auto const local_x = detail::getLocalNodalDOFs(x, indices);
        auto const r_c_indices =
            NumLib::LocalToGlobalIndexMap::RowColumnIndices(indices, indices);
```

chleh:
For some notes see #1277
Closes #1277.


ProcessLib/TES/TESLocalAssembler-impl.h
```diff
@@ -132,13 +133,16 @@ TESLocalAssembler<
...
-                                            std::vector<double> const& local_x)
+void TESLocalAssembler<ShapeFunction_, IntegrationMethod_, GlobalDim>::assemble(
+    std::size_t const id, NumLib::LocalToGlobalIndexMap const& dof_table,
+    double const /*t*/, GlobalVector const& x, GlobalMatrix& M, GlobalMatrix& K,
+    GlobalVector& b)
 {
     _local_M.setZero();
     _local_K.setZero();
     _local_b.setZero();
+    auto const indices = NumLib::getIndices(id, dof_table);
```
wenqing: Not used in this function.

wenqing: Sorry, I made a mistake. They are used after several line below.

NumLib/DOF/DOFTableUtil.h
```diff
@@ -0,0 +1,27 @@
...
+
+namespace NumLib
+{
+//! Returns nodal indices for the item identified by \c id from the given
+//!  \c dof_table.
+std::vector<GlobalIndexType> getIndices(
+    std::size_t const id, NumLib::LocalToGlobalIndexMap const& dof_table);
+
+//! Returns the values for the given \c dof_indices from the vector \c x.
+std::vector<double> getLocalNodalDOFs(
```
norihiro-w: Isn't this a function to create a sub-vector from the given positions? why not have it somewhere in MathLib?

NumLib/DOF/DOFTableUtil.h
```diff
@@ -0,0 +1,27 @@
...
+#define NUMLIB_DOF_DOFTABLEUTIL_H
+
+#include "NumLib/DOF/LocalToGlobalIndexMap.h"
+
+namespace NumLib
+{
+//! Returns nodal indices for the item identified by \c id from the given
+//!  \c dof_table.
+std::vector<GlobalIndexType> getIndices(
+    std::size_t const id, NumLib::LocalToGlobalIndexMap const& dof_table);
```
norihiro-w: is `id` a node ID or a DoF ID or element ID? can you specify it?

ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
```diff
@@ -74,10 +75,16 @@ class LocalAssemblerData
         assert(local_matrix_size == ShapeFunction::NPOINTS * NUM_NODAL_DOF);
     }
 
-    void assemble(double const /*t*/, std::vector<double> const& local_x) override
+    void assemble(std::size_t const id,
```
norihiro-w: If  `id` means an element ID, is it really needed to pass it here? `Element` object is already stored in this class.

NumLib/DOF/DOFTableUtil.h
chleh: It has the same meaning as `id` in the SerialExecutor. Currently it's the element `id`, but @endJunction told me once that in the future it might have broader meaning. That's why I chose the word here.
@endJunction, can you comment, please?

NumLib/DOF/DOFTableUtil.h
chleh: Would probably be the better choice. However, I didn't want to create a second file for that single function. Or do you have an existing file in mind?
May I keep the function here for the time being?

ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
chleh: For the meaning of `id` see my comment above.
I wrote `executeMemberOnDereferenced()` s.t. it passes the id as the first argument, because for future use it might be handy, and because the other `execute...()` methods also pass it. Therefore the `id` parameter is present here and I use it instead of `_element.getID()`.

NumLib/DOF/DOFTableUtil.h
endJunction: All correct. Maybe better to call it `mesh_item_id` for now, like it is in [LocalToGlobalIndexMap](https://github.com/ufz/ogs/blob/master/NumLib/DOF/LocalToGlobalIndexMap.h#L78) for some functions.

If we decide to remove the possibility of non-element-wise (*i.e.* node-wise, or edge-wise) assembly, then we can restrict the name to element_id (and maybe save one or two lines at places where the elements are available.)

NumLib/DOF/DOFTableUtil.h
chleh: Added suitable methods to our global vector classes.

NumLib/DOF/DOFTableUtil.h
chleh: Changed id to mesh_item_id

ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
chleh: Btw: The element is not (yet) stored in the TES local assembler.

MathLib/LinAlg/Eigen/EigenVector.h
```diff
@@ -65,6 +65,19 @@ class EigenVector final
         return _vec[rowId];
     }
 
+    /// get entries
+    std::vector<double> get(std::vector<IndexType> const& indices) const
+    {
+        std::vector<double> local_x;
+        local_x.reserve(indices.size());
```
norihiro-w: not now: allocating memory for each time (for each element) can degrade speed. need to profile later on

norihiro-w:
:+1:

NumLib/Assembler/SerialExecutor.h
```diff
@@ -76,6 +76,26 @@ struct SerialExecutor
...
     }
 
+    /// Executes the given \c method on each element of the input \c container.
+    ///
+    /// This method is very similar to executeMemberDereferenced().
+    ///
+    /// \param container collection of objects having pointer semantics.
+    /// \param method    the method being called, i.e., a member function pointer
+    ///                  to a member function of the \c container's elements.
+    /// \param Args      types of further arguments passed on to the method
```
endJunction: warning: Parameter 'Args' not found in the declaration --> 'args'

Also in other functions in the SerialExecutor, but this one is new...

ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
```diff
@@ -100,14 +107,11 @@ class LocalAssemblerData
...
         }
-    }
 
-    void addToGlobal(NumLib::LocalToGlobalIndexMap::RowColumnIndices const& indices,
-        GlobalMatrix& /*M*/, GlobalMatrix& K, GlobalVector& b)
-        const override
-    {
-        K.add(indices, _localA);
-        b.add(indices.rows, _localRhs);
+        auto const r_c_indices =
```
endJunction: Note for myself: Not sure here.

chleh:
I moved the index and local DOF computation into the general local assembler interface.

PR #1307, opened 2016-07-04
[NL] Unify unsigned type usage in shape functions.

endJunction:
The elements' dimensions (DIM and NPOINTS) are declared
'unsigned' is the MeshLib (MeshElement::dimension and
MeshElement::n_all_nodes) and should be of the same
type for the shape functions.

This is automated change with sed.


norihiro-w:
👍

chleh:
:+1:

wenqing:
:+1:

PR #1308, opened 2016-07-04
Added finalizeAssembly(x) to avoid PETSc errors in the vector copying

wenqing:
as titled. Although the results of the PETSc related tests are correct after introducing BLAS (now LinAlg), PETSc errors occur in the log file. This PR fixed the problem.

### DIFF
--- a/NumLib/ODESolver/TimeDiscretizedODESystem.cpp
+++ b/NumLib/ODESolver/TimeDiscretizedODESystem.cpp
@@ -32,6 +32,7 @@ void applyKnownSolutions(std::vector<Solutions> const* const known_solutions,
             MathLib::setVector(x, bc.ids[i], bc.values[i]);
         }
     }
+    MathLib::LinAlg::finalizeAssembly(x);
 }
 }
 


NumLib/ODESolver/NonlinearSolver.cpp
```diff
@@ -45,6 +45,7 @@ bool NonlinearSolver<NonlinearSolverTag::Picard>::solve(
 
     bool error_norms_met = false;
 
+    MathLib::LinAlg::finalizeAssembly(x);
```
norihiro-w: is here really the right place to finalize `x`? The vector is given as a parameter to this function.

chleh: I agree with @norihiro-w. I guess (didn't check), that the error is caused, because `finalizeAssembly()` is not called after `applyKnownSolutions()` in the time loop. Maybe the errors are fixed if you add `finalizeAssembly()` to `applyKnownSolutions()` in `TimeDiscretizedODESystem.cpp`: https://github.com/wenqing/ogs/blob/1dbf4cd153e7302cfda83d4212164b5990874981/NumLib/ODESolver/TimeDiscretizedODESystem.cpp#L34.

wenqing: @chleh Moving  MathLib::LinAlg::finalizeAssembly(x) to applyKnownSolutions() also solves the problem. By the way, set values individually inside   applyKnownSolutions()  is insufficiency for MPI computation, I want to add a function of setVectorValues in  MathLib

wenqing: Moved to applyKnownSolutions().

chleh:
:+1:

PR #1311, opened 2016-07-05
Allow PETSc run with only one CPU core

wenqing:
Allow PETSc run with only one CPU core with reading original mesh data directly, e.g.

```
orterun -np 1   ogs -o ~/temp ~/code/ogs6/sources/Tests/Data/Elliptic/cube_1x1x1_GroundWaterFlow/cube_1e3_neumann.prj
```
where orterun could also be mpirun.

### DIFF
--- a/Applications/CLI/Tests.cmake
+++ b/Applications/CLI/Tests.cmake
@@ -201,14 +201,158 @@ else()
         cube_1e3_neumann_pcs_0_ts_1_t_1.000000_2.vtu pressure pressure
     )
 
+    # Single core
+    # CUBE 1x1x1 GROUNDWATER FLOW TESTS
+    foreach(mesh_size 1e0 1e1 1e2 1e3)
+        AddTest(
+            NAME GroundWaterFlowProcess_cube_1x1x1_${mesh_size}
+            PATH Elliptic/cube_1x1x1_GroundWaterFlow
+            EXECUTABLE_ARGS cube_${mesh_size}.prj
+            WRAPPER mpirun
+            WRAPPER_ARGS -np 1
+            TESTER vtkdiff
+            ABSTOL 1e-15 RELTOL 1e-15
+            DIFF_DATA cube_${mesh_size}_pcs_0_ts_1_t_1_0.vtu Linear_1_to_minus1 pressure
+        )
+
+        AddTest(
+            NAME GroundWaterFlowProcess_cube_1x1x1_Neumann_${mesh_size}
+            PATH Elliptic/cube_1x1x1_GroundWaterFlow
+            EXECUTABLE_ARGS cube_${mesh_size}_neumann.prj
+            WRAPPER mpirun
+            WRAPPER_ARGS -np 1
+            TESTER vtkdiff
+            ABSTOL 1e-1 RELTOL 1e-1
+            DIFF_DATA cube_${mesh_size}_neumann_pcs_0_ts_1_t_1_0.vtu D1_left_front_N1_right pressure
+        )
+    endforeach()
+
+
+    foreach(mesh_size 1e4 2e4 3e4 4e4 5e4 1e5 1e6)
+        AddTest(
+            NAME LARGE_GroundWaterFlowProcess_cube_1x1x1_${mesh_size}
+            PATH Elliptic/cube_1x1x1_GroundWaterFlow
+            EXECUTABLE_ARGS cube_${mesh_size}.prj
+            WRAPPER mpirun
+            WRAPPER_ARGS -np 1
+            TESTER vtkdiff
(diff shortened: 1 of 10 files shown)


MeshLib/IO/readMeshFromFile.cpp
```diff
@@ -45,9 +45,36 @@ namespace IO
...
+    MPI_Comm_size(MPI_COMM_WORLD, &world_size);
+    if (world_size > 1)
+    {
+        MeshLib::IO::NodePartitionedMeshReader read_pmesh(PETSC_COMM_WORLD);
+        const std::string file_name_base = BaseLib::dropFileExtension(file_name);
+        return read_pmesh.read(file_name_base);
+    }
+    else if (world_size == 1)
+    {
+        MeshLib::Mesh* mesh = nullptr;
```
endJunction: there is MeshLib::IO::readMeshFromFile() function already capturing msh/vtk stuff.

Applications/CLI/Tests.cmake
```diff
@@ -200,4 +200,15 @@ else()
...
+    AddTest(
+        NAME cube_1e3_neumann_np_1
+        PATH Elliptic/cube_1x1x1_GroundWaterFlow
+        EXECUTABLE_ARGS cube_1e3_neumann.prj
+        WRAPPER mpirun
+        WRAPPER_ARGS -np 1
+        TESTER vtkdiff
+        ABSTOL 1e-1 RELTOL 1e-1
+        DIFF_DATA cube_1e3_neumann_pcs_0_ts_1_t_1_0.vtu D1_left_front_N1_right pressure
+    )
```
chleh: Is it necessary to use mpirun here? If not, I'd prefer to run without it. Furthermore it would be nice if all tests now were run with the PETSc configuration, not only a few selected ones.
If there are problems running all ctests with PETSc, this can also be done later.

MeshLib/NodePartitionedMesh.h
```diff
@@ -29,6 +29,32 @@ namespace MeshLib
...
+              _n_active_base_nodes(mesh.getNumberOfBaseNodes()),
+              _n_active_nodes(mesh.getNumberOfNodes())
+        {
+            const auto& mesh_nodes = mesh.getNodes();
+            for (std::size_t i = 0; i < _nodes.size(); i++)
+            {
+                _global_node_ids[i] = _nodes[i]->getID();
+                auto node = _nodes[i];
+                // Copy constructor of Mesh does not copy the connected
+                // nodes to node.
```
chleh: Actually, that looks like an incomplete implementation of the `Node` copy constructor. I'd rather fix it there. Can anyone more involved in MeshLib comment on that?

chleh:
:+1: after questions. Only small things.
It's good to have this feature, now.

MeshLib/IO/readMeshFromFile.cpp
wenqing: Introduced a new function, readMeshFromFileSerial.

Applications/CLI/Tests.cmake
wenqing: Then we should firstly compile PETSc without MPI, i.e. using it in serial way. Otherwise, mpirun should be used to initializes the MPI environment.  Do we need to compile a serial PETSc library?

If we want to compile PETSc serially, MeshComponentMap::MeshComponentMap should be modify to let PETSc vector and matrix use the DOF table of serial computing.

As for the unit test, the DISABLED TESTS are mostly related to MeshComponentMap::MeshComponentMap, which has different definitions for serial and parallel computing.

MeshLib/NodePartitionedMesh.h
wenqing: There are two commented out lines in the copy constructor of class Mesh for calculating the connected nodes as
```
    //this->setNodesConnectedByEdges();
    //this->setNodesConnectedByElements();
```
At the beginning, I wanted to add these lines in the copy constructor of class Mesh, but stopped to do that after see that comments.

wenqing:
@chleh  Added a comment to the copy constructor
```
            // TODO To add copying of the connected nodes (and elements)
            //      in the copying constructor of class Node in order to
            //      drop the following loop.
```

MeshLib/NodePartitionedMesh.h
chleh: I am fine for this PR with the todo comment you added above. But maybe @rinkk or @endJunction can briefly tell why the Node constructor is incomplete?

chleh:
:+1:

MeshLib/NodePartitionedMesh.h
endJunction: I can only guess that one tries to avoid double work for some special cases...

PR #1312, opened 2016-07-06
Added a wrapper for std::bind

chleh:
Follow-up of #1309.

Example:
```{.cpp}

using std::placeholders;
Object some_object;

auto f_bind = std::function<ReturnType(Arg1, Arg2, Arg3)>(
                  std::bind(&Object::methodWithThreeArguments,
                      std::ref(some_object), _1, _2, _3));

auto f_easy = easyBind(&Object::methodWithThreeArguments, some_object);
```

The lines generating `f_bind` and `f_easy` produce equivalent results.

Features:
* less typing
* result type can be used for type deduction (I'll use that in a follow-up PR)
* no accidental copies of the object being bound to.
* easyBind guarantees that all types involved match up.

PR #1314, opened 2016-07-06
Named Functions

chleh:
Follow-up of #1312.

This PR adds classes that enable the construction of expression trees dynamically at runtime.
That roughly goes into the direction discussed recently in the frame of constitutive relations.
However, I want to use the code provided here at first to compute secondary variables.

For a usage example, please have a look at the tests. There will be a follow-up PR coupling this PR with FEM processes, soon.

### DIFF
--- a/NumLib/NamedFunction.cpp
+++ b/NumLib/NamedFunction.cpp
@@ -1,0 +1,198 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#include "NamedFunction.h"
+#include "BaseLib/TMPUtil.h"
+
+//! Helper struct used in conjunction with BaseLib::IntegerSequence to get a
+//! sequence of types, where each type is double.
+template <int>
+struct Double
+{
+    using type = double;
+};
+
+/*! Calls the given \c function with the given  \c arguments.
+ *
+ * \tparam Indices sequence of integers used to expand the \c arguments vector
+ * to individual arguments.
+ */
+template <int... Indices>
+double call_(void* function, std::vector<double> const& arguments)
+{
+    assert(arguments.size() == sizeof...(Indices));
+    auto* fct = reinterpret_cast<
+        std::function<double(typename Double<Indices>::type...)>*>(function);
+    auto* args = arguments.data();
+    return (*fct)(args[Indices]...);
+}
+
+using CallerFunction = decltype(&call_<>);
+
+//! Helps instantiating the call_() function.
+template <int... Indices>
+CallerFunction generateCallerFromIntegerSequence(
+    BaseLib::IntegerSequence<Indices...>)
+{
+    return call_<Indices...>;
+}
+
+//! Instantiates the call_() function for the provided number of arguments.
+template <int NumArguments>
+CallerFunction generateCaller()
+{
+    return generateCallerFromIntegerSequence(
+        typename BaseLib::GenerateIntegerSequence<NumArguments>::type{});
+}
+
+//! Holds instantiations of the call_() function for various numbers of
+//! arguments.
+static const CallerFunction callers[] = {
+    generateCaller<0>(),  generateCaller<1>(),  generateCaller<2>(),
+    generateCaller<3>(),  generateCaller<4>(),  generateCaller<5>(),
+    generateCaller<6>(),  generateCaller<7>(),  generateCaller<8>(),
+    generateCaller<9>(),  generateCaller<10>(), generateCaller<11>(),
+    generateCaller<12>(), generateCaller<13>(), generateCaller<14>(),
+    generateCaller<15>(), generateCaller<16>(), generateCaller<17>(),
+    generateCaller<18>(), generateCaller<19>(), generateCaller<20>(),
+    generateCaller<21>(), generateCaller<22>(), generateCaller<23>(),
+    generateCaller<24>(), generateCaller<25>(), generateCaller<26>(),
+    generateCaller<27>(), generateCaller<28>(), generateCaller<29>(),
+    generateCaller<30>(), generateCaller<31>()};
+static_assert(sizeof(callers) / sizeof(CallerFunction) ==
+                  NumLib::MAX_FUNCTION_ARGS,
+              "You did not instantiate the right number of callers.");
+
+/*! Deletes the given \c function.
+ *
+ * \tparam Indices sequence of integers used to cast the \c function to the
+ * correct type.
+ */
+template <int... Indices>
+void delete_(void* function)
+{
+    auto* fct = reinterpret_cast<
+        std::function<double(typename Double<Indices>::type...)>*>(function);
+    delete fct;
+}
+
+using DeleterFunction = decltype(&delete_<>);
+
+//! Helps instantiating the delete_() function.
+template <int... Indices>
+DeleterFunction generateDeleterFromIntegerSequence(
+    BaseLib::IntegerSequence<Indices...>)
+{
+    return delete_<Indices...>;
+}
+
+//! Instantiates the delete_() function for the provided number of arguments.
+template <int NumArguments>
+DeleterFunction generateDeleter()
+{
+    return generateDeleterFromIntegerSequence(
+        typename BaseLib::GenerateIntegerSequence<NumArguments>::type{});
+}
+
+//! Holds instantiations of the delete_() function for various numbers of
+//! arguments.
+static const DeleterFunction deleters[] = {
+    generateDeleter<0>(),  generateDeleter<1>(),  generateDeleter<2>(),
+    generateDeleter<3>(),  generateDeleter<4>(),  generateDeleter<5>(),
+    generateDeleter<6>(),  generateDeleter<7>(),  generateDeleter<8>(),
+    generateDeleter<9>(),  generateDeleter<10>(), generateDeleter<11>(),
+    generateDeleter<12>(), generateDeleter<13>(), generateDeleter<14>(),
+    generateDeleter<15>(), generateDeleter<16>(), generateDeleter<17>(),
+    generateDeleter<18>(), generateDeleter<19>(), generateDeleter<20>(),
+    generateDeleter<21>(), generateDeleter<22>(), generateDeleter<23>(),
+    generateDeleter<24>(), generateDeleter<25>(), generateDeleter<26>(),
+    generateDeleter<27>(), generateDeleter<28>(), generateDeleter<29>(),
+    generateDeleter<30>(), generateDeleter<31>()};
+static_assert(sizeof(deleters) / sizeof(DeleterFunction) ==
+                  NumLib::MAX_FUNCTION_ARGS,
+              "You did not instantiate the right number of deleters.");
+
+/*! Copies the given \c function.
+ *
+ * \tparam Indices sequence of integers used to cast the \c function to the
+ * correct type.
+ */
+template <int... Indices>
+void* copy_(void* function)
+{
+    auto* fct = reinterpret_cast<
+        std::function<double(typename Double<Indices>::type...)>*>(function);
+    return new std::function<double(typename Double<Indices>::type...)>(*fct);
+}
+
+using CopierFunction = decltype(&copy_<>);
+
+//! Helps instantiating the copy_() function.
+template <int... Indices>
+CopierFunction generateCopierFromIntegerSequence(
+    BaseLib::IntegerSequence<Indices...>)
+{
+    return copy_<Indices...>;
+}
+
+//! Instantiates the copy_() function for the provided number of arguments.
+template <int NumArguments>
+CopierFunction generateCopier()
+{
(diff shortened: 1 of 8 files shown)


NumLib/NamedFunction.h
```diff
@@ -0,0 +1,123 @@
...
+    /*! Constructs a new named function.
+     *
+     * \param name the function's name
+     * \param arguments names  of arguments of the function
+     * \param function the actual function object
+     */
+    template <typename ReturnType, typename... Arguments>
+    NamedFunction(
+        std::string const& name,
+        std::vector<ArgumentInfo>&& arguments,
```
norihiro-w: better to name it `argument_names`

NumLib/NamedFunction.h
```diff
@@ -0,0 +1,123 @@
...
+    NamedFunction(NamedFunction&& other);
+    NamedFunction(NamedFunction const&);
+
+    ~NamedFunction();
+
+    //! Returns the function's name.
+    std::string const& getName() const { return _name; }
+
+    //! Returns information about the function's arguments.
+    std::vector<ArgumentInfo> const& getArgumentInfo() const
```
norihiro-w: I would name it `getArgumentNames()`

NumLib/NamedFunction.h
```diff
@@ -0,0 +1,123 @@
...
+public:
+    using ArgumentInfo = std::string;
+
+    /*! Constructs a new named function.
+     *
+     * \param name the function's name
+     * \param arguments names  of arguments of the function
+     * \param function the actual function object
+     */
+    template <typename ReturnType, typename... Arguments>
```
norihiro-w: question: can this class handle a function without arguments?

NumLib/NamedFunction.h
```diff
@@ -0,0 +1,123 @@
...
+const bool AllTypesSameAs<TRef>::value;
+
+} // namespace detail
+
+
+
+namespace NumLib
+{
+//! Maximum number of function arguments supported by NamedFunction.
+const int MAX_FUNCTION_ARGS = 32;
```
norihiro-w: why not define this within NamedFunction class? The variable name is too general to be placed under NumLib namespace

NumLib/NamedFunctionCaller.h
```diff
@@ -0,0 +1,133 @@
...
+class NamedFunctionCaller final
+{
+public:
+    //! Constructs an instance whose unbound arguments have the given names.
+    explicit NamedFunctionCaller(
+        std::initializer_list<std::string> unbound_arguments);
+
+    //! Adds the given named function
+    void addNamedFunction(NamedFunction&& fct);
+    std::vector<NamedFunction> const& getNamedFunctions() const
```
norihiro-w: need documentation

NumLib/NamedFunctionCaller.h
```diff
@@ -0,0 +1,133 @@
...
+{
+class SpecialFunctionCaller;
+
+//! Builds expression trees of named functions dynamically at runtime.
+class NamedFunctionCaller final
+{
+public:
+    //! Constructs an instance whose unbound arguments have the given names.
+    explicit NamedFunctionCaller(
+        std::initializer_list<std::string> unbound_arguments);
```
norihiro-w: argument_names

NumLib/NamedFunctionCaller.h
```diff
@@ -0,0 +1,133 @@
...
+
+    //! Adds the given named function
+    void addNamedFunction(NamedFunction&& fct);
```
norihiro-w: isn't it possible to pass functions to the constructor?

NumLib/NamedFunctionCaller.h
```diff
@@ -0,0 +1,133 @@
...
+    {
+        return _named_functions;
+    }
+
+    //! Declares that the argument with name \c sink_args of the function \c
+    //! sink_fct is being computed by the function \c source.
+    //!
+    //! The functions involved need not already be known to the
+    //! NamedFunctionCaller.
+    void plug(std::string const& sink_fct, std::string const& sink_arg,
```
norihiro-w: why not merge this into `applyPlugs()`?

NumLib/NamedFunctionCaller.h
```diff
@@ -0,0 +1,133 @@
...
+              std::string const& source);
```
norihiro-w: source_fct

NumLib/NamedFunctionCaller.h
```diff
@@ -0,0 +1,133 @@
...
+
+    //! Actually plug all the plugs previously declared.
+    //!
+    //! \pre All functions involved must have been added.
+    void applyPlugs();
+
+    //! Calls the function with the given name with the given unbound arguments.
+    //!
+    //! \pre applyPlugs() must have been called before.
```
norihiro-w: why not call `applyPlugs()` inside of this function?

NumLib/NamedFunctionCaller.h
```diff
@@ -0,0 +1,133 @@
...
+    double call(std::string const& function_name,
+                std::vector<double> const& unbound_arguments) const;
+
+    //! Creates a function caller that is able to call the function with the
+    //! given name.
+    SpecialFunctionCaller getSpecialFunction(std::string const& function_name);
```
norihiro-w: `getSpecialFunctionCaller`

NumLib/NamedFunctionCaller.h
```diff
@@ -0,0 +1,133 @@
...
+    std::vector<SinkSource> _deferred_plugs;
+
+    friend class SpecialFunctionCaller;
+};
+
+//! A function caller that can call one specific function.
+//!
+//! \todo Use this class to provide some optimizations of the expression
+//! evaluation.
+class SpecialFunctionCaller final
```
norihiro-w: why `special`? it looks `specific` is more appropriate here.

NumLib/NamedFunctionCaller.h
norihiro-w: do we need this function? A user can call it using a result of `getSpecialFunction()`.

NumLib/NamedFunctionProvider.h
```diff
@@ -0,0 +1,32 @@
...
+#include "NamedFunction.h"
+
+namespace NumLib
+{
+
+//! Interface used for providing named functions.
+class NamedFunctionProvider
+{
+public:
+    virtual std::vector<NamedFunction> getNamedFunctions() const
```
norihiro-w: why not make this function pure virtual?

norihiro-w:
i checked only class interface. Comments are mostly about naming of functions etc.

NumLib/NamedFunction.h
chleh: Yes, zero arguments are valid.

NumLib/NamedFunction.h
chleh: It is used in the templated constructor below. However, I can put as a static member into the class body.

NumLib/NamedFunctionCaller.h
chleh: I think that wouldn't be handy in the end. Currently I build a `NamedFunctionCaller` in the constructor of `Process` and add functions to it later on during the initialization.
Passing all functions to the ctor would mean that everything has to be ready from the beginning.

NumLib/NamedFunctionCaller.h
chleh: That's a bit related to your question above.
`plug()` defines edges in the expression call graph. Upon calling plug the associated vertices (i.e. named functions) needn't be there yet (That is, because when parsing the process configuration not all of the inner parts of the process exist yet.).
However, upon calling `applyPlugs()` all needed named function must have been added to the `NamedFunctionCaller`.
So: No I think it's not possible easily.

NumLib/NamedFunctionCaller.h
chleh: > why not call applyPlugs() inside of this function?

`applyPlugs()` belongs to the initialization of the instance, `call()` belongs to its application. So I rather would keep the two parts separated. Furthermore, the latter is const, the former is not. (Sure, those are not very strong arguments.)

> do we need this function? A user can call it using a result of getSpecialFunction().

It's not strictly needed. I can remove it.

NumLib/NamedFunctionCaller.h
chleh: That would be a better name, indeed.

NumLib/NamedFunctionProvider.h
chleh: This interface is meant to be pulled into many different classes, e.g. into the interfaces of (all) material properties. If one does not provide any named function and because there is a default implementation one does not need to bother in a concrete implementation to implement this method again. So I'd like to keep it as it is.

chleh:
I'll do the renaming when the main part of the code review has been finished.

NumLib/NamedFunction.cpp
```diff
@@ -0,0 +1,198 @@
...
+    generateCaller<3>(),  generateCaller<4>(),  generateCaller<5>(),
+    generateCaller<6>(),  generateCaller<7>(),  generateCaller<8>(),
+    generateCaller<9>(),  generateCaller<10>(), generateCaller<11>(),
+    generateCaller<12>(), generateCaller<13>(), generateCaller<14>(),
+    generateCaller<15>(), generateCaller<16>(), generateCaller<17>(),
+    generateCaller<18>(), generateCaller<19>(), generateCaller<20>(),
+    generateCaller<21>(), generateCaller<22>(), generateCaller<23>(),
+    generateCaller<24>(), generateCaller<25>(), generateCaller<26>(),
+    generateCaller<27>(), generateCaller<28>(), generateCaller<29>(),
+    generateCaller<30>(), generateCaller<31>()};
```
chleh: TODO: `MAX_FUNCTION_ARGS` is 32.

NumLib/NamedFunction.h
norihiro-w: thanks!

NumLib/NamedFunctionCaller.h
norihiro-w: i see. thanks for the explanation

NumLib/NamedFunctionProvider.h
norihiro-w: i see

norihiro-w:
i will not review cpp files because there are many things I don't understand and these code will be used only for output. If the renaming is done, this PR is ok from my side.

NumLib/NamedFunctionCaller.cpp
```diff
@@ -0,0 +1,300 @@
...
+              int sink_fct, Callback&& callback)
+{
+    assert(sink_fct < static_cast<int>(map_sink_source.size()));
+    callback(sink_fct, TraversePosition::StartNode);
+
+    if (sink_fct < 0)
+        return;
+
+    auto const& si_so = map_sink_source[sink_fct];
+    int const num_args = si_so.size();
```
endJunction: auto or std::size_t (resolves some of the implicit conversions...)

NumLib/NamedFunctionCaller.cpp
```diff
@@ -0,0 +1,300 @@
...
+        auto const sink_arg_it =
+            std::find(sink_args.begin(), sink_args.end(), sink_arg);
+        if (sink_arg_it == sink_args.end())
+        {
+            OGS_FATAL(
+                "An argument with the name `%s' has not been found for the "
+                "function `%s'.",
+                sink_arg.c_str(), sink_fct.c_str());
+        }
+        std::size_t sink_arg_idx =
```
endJunction: const

NumLib/NamedFunctionCaller.cpp
```diff
@@ -0,0 +1,300 @@
...
+    }
+}
+
+double NamedFunctionCaller::call(
+    const std::string& function_name,
+    const std::vector<double>& unbound_arguments) const
+{
+    assert(unbound_arguments.size() == getNumberOfUnboundArguments());
+
+    auto it = _map_name_idx.find(function_name);
```
wenqing: const?

NumLib/NamedFunctionProvider.h
```diff
@@ -0,0 +1,32 @@
...
+    {
+        return std::vector<NamedFunction>{};
+    }
+
+    virtual ~NamedFunctionProvider() = default;
```
endJunction: No need to default the dtor...

NumLib/NamedFunctionCaller.cpp
```diff
@@ -0,0 +1,318 @@
...
+            }
+        }
+    }
+
+    return num_dependent == 0;
+}
+
+enum class TraversePosition { StartNode, BetweenChildren, EndNode };
+
+template <typename Callback>
```
endJunction: DFS, traverse

NumLib/NamedFunctionCaller.h
chleh: removed.

NumLib/NamedFunctionCaller.h
chleh: Name changed.

NumLib/NamedFunctionCaller.cpp
chleh: method deleted.

PR #1318, opened 2016-07-18
Fix surfaces lying in x-y-plane

chleh:
In our shape matrix computation routines there was some code detecting the special case when a 2D element lies in the x-y plane. This code led to Neumann BCs being added sign-flipped if the surface at which it was applied was in the x-y-plane with surface normal (0, 0, -1).

This PR fixes the special cases in the code. A small end-to-end test example has been added. Cf. image: Groundwaterflow process; upper row before the PR, lower row after the PR; one surface is kept at Dirichlet 1, the other has Neumann 1, i.e., the point at which the Neumann BC is applied must be the maximum over the entire z-axis; that did not always hold before this PR (upper right case).

fixed-surface-orientation

Furthermore the det J > 0 check has been made strict in this PR. I.e., elements with zero volume or wrong orientation are disallowed now.

### DIFF
--- a/Applications/CLI/Tests.cmake
+++ b/Applications/CLI/Tests.cmake
@@ -124,9 +124,31 @@ if(NOT OGS_USE_MPI)
             ABSTOL 1e-14 RELTOL 1e-14
             DIFF_DATA line_${mesh_size}_neumann_pcs_0_ts_1_t_1.000000.vtu D1_left_N1_right pressure
         )
-        endforeach()
+    endforeach()
 
+    # Some Neumann BC tests
+    AddTest(
+        NAME GroundWaterFlowProcess_cube_top
+        PATH Elliptic/cube_1x1x1_GroundWaterFlow
+        EXECUTABLE ogs
+        EXECUTABLE_ARGS cube_1e3_top_neumann.prj
+        WRAPPER time
+        TESTER vtkdiff
+        ABSTOL 1e-16 RELTOL 1e-16
+        DIFF_DATA cube_1e3_top_neumann_pcs_0_ts_1_t_1.000000.vtu pressure pressure
+    )
+    AddTest(
+        NAME GroundWaterFlowProcess_cube_bottom
+        PATH Elliptic/cube_1x1x1_GroundWaterFlow
+        EXECUTABLE ogs
+        EXECUTABLE_ARGS cube_1e3_bottom_neumann.prj
+        WRAPPER time
+        TESTER vtkdiff
+        ABSTOL 1e-16 RELTOL 1e-16
+        DIFF_DATA cube_1e3_bottom_neumann_pcs_0_ts_1_t_1.000000.vtu pressure pressure
+    )
 
+    # TES tests
     AddTest(
         NAME TES_zeolite_discharge_small
         PATH Parabolic/TES/1D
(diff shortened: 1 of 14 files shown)


GeoLib/AnalyticalGeometry-impl.h
```diff
@@ -118,17 +118,31 @@ void computeRotationMatrixToXY(MathLib::Vector3 const& n,
...
+            rot_mat(0,1) = 0.0;
+            rot_mat(0,2) = 0.0;
+            rot_mat(1,0) = 0.0;
+            rot_mat(1,1) = 1.0;
+            rot_mat(1,2) = 0.0;
+            rot_mat(2,0) = 0.0;
+            rot_mat(2,1) = 0.0;
+            rot_mat(2,2) = -1.0;
+            return;
+        }
```
TomFischer: Why not modify the first and third diagonal entries via: `rot_mat(i,i) = n[2]>0 ? 1.0 : -1.0;` (i = 0,2) and not duplicate the setting of the other matrix entries?

Tests/Data
```diff
@@ -1 +1 @@
-Subproject commit 3087421f7ad138c72be7ee85f6872925da48ba7b
+Subproject commit 28d74b31d74aa5f03a418cc5a11f33cd1e2d7662
```
endJunction: In the Tests/Data could the Elliptic/cube_1x1x1_NeumannBC/ could you please cleanup the vtu files? There are unnecessary data arrays in the input vtu and output vtus. (Alternatively one could refer to the original input vtu (without copy) which is one level up).

chleh:
Jenkins, test this please.

chleh:
Jenkins, test this please.

Applications/CLI/Tests.cmake
```diff
@@ -126,7 +126,29 @@ if(NOT OGS_USE_MPI)
...
+    # Some Neumann BC tests
+    AddTest(
+        NAME NeumannBC_cube_top
+        PATH Elliptic/cube_1x1x1_NeumannBC
+        EXECUTABLE ogs
+        EXECUTABLE_ARGS cube_1e3_top_neumann.prj
+        WRAPPER time
+        TESTER vtkdiff
+        ABSTOL 1e-16 RELTOL 1e-16
+        DIFF_DATA cube_1e3_top_neumann_pcs_0_ts_1_t_1.000000.vtu pressure pressure
```
TomFischer: Is it already fixed as in PR "Fix AddTest"?

Tests/Data
endJunction: Looks good. Later when merging the Tests/Data branch into Tests/Data/master please squash the changes, because the data will take space otherwise.

Applications/CLI/Tests.cmake
```diff
@@ -126,7 +126,29 @@ if(NOT OGS_USE_MPI)
         )
         endforeach()
 
+    # Some Neumann BC tests
+    AddTest(
+        NAME NeumannBC_cube_top
+        PATH Elliptic/cube_1x1x1_NeumannBC
```
endJunction: The test is a GroundWaterFlow test and should be placed in Tests/Data/Elliptic/cube_1x1x1_GroundWaterFlow.

endJunction:
After the test is named GroundWaterFlow/neumann... :fast_forward:

Applications/CLI/Tests.cmake
chleh: After the fix AddTest PR cmake will complain about this change. So there will be no silent failure.

PR #1319, opened 2016-07-18
Fix pvtu name

wenqing:
Since SetFileName of ``vtkSmartPointer<UnstructuredGridWriter>`` drops characters starting from '.' in a string, the previous changes in the output file names cause a problem in writing a correct pvtu file. This PR fixes the problem, meanwhile it removes some compilation warnings.

### DIFF
--- a/Applications/CLI/Tests.cmake
+++ b/Applications/CLI/Tests.cmake
@@ -168,9 +168,9 @@ else()
         TESTER vtkdiff
         ABSTOL 1e-16 RELTOL 1e-16
         DIFF_DATA
-        quad_20x10_GroundWaterFlow_result_pcs_0_ts_1_t_1.000000_0.vtu pressure pressure
-        quad_20x10_GroundWaterFlow_result_pcs_0_ts_1_t_1.000000_1.vtu pressure pressure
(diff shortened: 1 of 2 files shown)


wenqing:
Jenkins, test this please.

wenqing:
Jenkins, test this please.

endJunction:
I can not confirm that vtkXMLUnstructuredGridWriter->SetFileName method drops any characters starting from a dot. Actually it does not modify the given file name at all:
The method is implemented in [vtkXMLWriter.h:105]( https://github.com/Kitware/VTK/blob/master/IO/XML/vtkXMLWriter.h#L105) via `vtkSetStringMacro(FileName)` where `FileName` is a `char*`. The  macro itself is implemented in [vtkSetGet.h:104](https://github.com/Kitware/VTK/blob/master/Common/Core/vtkSetGet.h#L104) and only copies the given char array.

The error must be somewhere else, not in the SetFileName method.

chleh:
Maybe the parallel version `vtkXMLPUnstructuredGridWriter` behaves differently?

wenqing:
@chleh You are right.

endJunction:
I have merged the commit fixing some warnings; it has nothing to do with this PR.

PR #1322, opened 2016-07-19
Added time measurement output

chleh:
Sorry @wenqing that we did not respond to your time measurement PR #863. But I think time measurement is important, so here is my attempt on it.

This PR provides wall-time measurement of assembly, linear solvers, nonlinear iterations, dirchlet BCs timesteps, output, and overall runtime.

The time is measured for individual "events" and not accumulated over timesteps.

### DIFF
--- a/Applications/CLI/ogs.cpp
+++ b/Applications/CLI/ogs.cpp
@@ -75,6 +75,9 @@ int main(int argc, char *argv[])
     ApplicationsLib::LogogSetup logog_setup;
     logog_setup.setLevel(log_level_arg.getValue());
 
+    INFO("This is OpenGeoSys-6 version %s.",
+         BaseLib::BuildInfo::git_describe.c_str());
+
     BaseLib::RunTime run_time;
     run_time.start();
 
--- a/Applications/CLI/ogs.cpp
+++ b/Applications/CLI/ogs.cpp
@@ -11,14 +11,12 @@
  */
 
 #include <chrono>
-#include <ctime>
-
-// ThirdParty/tclap
-#include "tclap/CmdLine.h"
+#include <tclap/CmdLine.h>
 
 // BaseLib
 #include "BaseLib/BuildInfo.h"
 #include "BaseLib/ConfigTreeUtil.h"
+#include "BaseLib/DateTools.h"
 #include "BaseLib/FileTools.h"
 #include "BaseLib/RunTime.h"
 
@@ -81,16 +79,10 @@ int main(int argc, char *argv[])
     run_time.start();
 
     {
-        auto const start_time_sys = std::chrono::system_clock::to_time_t(
-            std::chrono::system_clock::now());
-        char time_str[100];
-        if (std::strftime(time_str, sizeof(time_str), "%F %T %z",
-            std::localtime(&start_time_sys))) {
-	    INFO("OGS started on %s.", time_str);
-	}
+        auto const start_time = std::chrono::system_clock::now();
+        auto const time_str = BaseLib::formatDate(start_time);
+        INFO("OGS started on %s.", time_str.c_str());
     }
-
-    std::chrono::steady_clock::now();
 
     auto ogs_status = EXIT_SUCCESS;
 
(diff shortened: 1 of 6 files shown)


Applications/CLI/ogs.cpp
```diff
@@ -74,6 +79,21 @@ int main(int argc, char *argv[])
...
     logog_setup.setLevel(log_level_arg.getValue());
 
+    BaseLib::RunTime run_time;
+    run_time.start();
+
+    {
+        auto const start_time_sys = std::chrono::system_clock::to_time_t(
+            std::chrono::system_clock::now());
+        std::ostringstream sstr;
+        sstr << std::put_time(std::localtime(&start_time_sys), "%F %T %z");
```
wenqing: % before T and Z display in red colour, but the expression looks nothing wrong.

wenqing:
:+1:

Applications/CLI/ogs.cpp
```diff
@@ -74,6 +77,23 @@ int main(int argc, char *argv[])
...
+        auto const start_time_sys = std::chrono::system_clock::to_time_t(
+            std::chrono::system_clock::now());
+        char time_str[100];
+        if (std::strftime(time_str, sizeof(time_str), "%F %T %z",
+            std::localtime(&start_time_sys))) {
+	    INFO("OGS started on %s.", time_str);
+	}
+    }
+
+    std::chrono::steady_clock::now();
```
TomFischer: Since the return value is not stored here I wonder if this line is necessary?

Applications/CLI/ogs.cpp
```diff
@@ -74,6 +77,23 @@ int main(int argc, char *argv[])
...
+
+    {
+        auto const start_time_sys = std::chrono::system_clock::to_time_t(
+            std::chrono::system_clock::now());
+        char time_str[100];
+        if (std::strftime(time_str, sizeof(time_str), "%F %T %z",
+            std::localtime(&start_time_sys))) {
+	    INFO("OGS started on %s.", time_str);
+	}
+    }
```
TomFischer: Maybe lines 84-90 can be moved as a general function into BaseLib?

Applications/CLI/ogs.cpp
```diff
@@ -117,9 +137,22 @@ int main(int argc, char *argv[])
...
     }
+
+    {
+        auto const end_time_sys = std::chrono::system_clock::to_time_t(
+            std::chrono::system_clock::now());
+        char time_str[100];
+        if (std::strftime(time_str, sizeof(time_str), "%F %T %z",
+            std::localtime(&end_time_sys))) {
+	    INFO("OGS terminated on %s.", time_str);
+	}
```
TomFischer: Here you could use the suggested general BaseLib function and avoid code duplication.

TomFischer:
I do not know if it would be a good idea to implement a new loging macro TIME into logog for timing output. What do you think?

TomFischer:
After moving extracting code to BaseLib: :+1:

chleh:
@wenqing: I'd prefer not to accumulate in ogs. I think such accumulation can be easily done by grepping the measured times from the log and summing them up in a script.

chleh:
@TomFischer: Since the logog macros and logog log line prefixes are defined in macro.hpp and formatter.cpp in that third party package I don't want to introduce a custom TIME log macro now.

chleh:
In the last commit I added functionality that prints the OGS version in every run.

BaseLib/DateTools.h
```diff
@@ -55,6 +57,15 @@ int strDate2int(const std::string &s);
  */
 int xmlDate2int(const std::string &s);
 
+/**
+ * Formats the given time point according to RFC 3339 (cf. man-page of the unix
+ * date utility).
+ *
+ * Example: 2006-08-14 02:34:56-06:00
```
endJunction: I'd prefer not to print the year and the timezone, which are very unlikely to change in a single run...

chleh: I'd rather keep both of them. `08-14 02:34:56` doesn't look like a real timestamp, does it? :smile:

Issue #1320, opened 2016-07-18
Mac_utils build fails


endJunction:
Since the https://github.com/ufz/ogs/pull/1238 PR mac-utils build on jenkins is broken (see Jenkins/OGS-6/Mac_Utils there it is red since build number 535, which is the named PR merge). Namely since b7d9fb, where metis cmake is included.

The error is
```
ThirdParty/tetgen/predicates.cxx:123:10: fatal error: 'fpu_control.h' file not found
#include <fpu_control.h>
```

It seems to me that the metis' cmake defines LINUX, wihch is protecting the fpu_control.h include.

wenqing:
@endJunction   Before merging, PR  #1238 looks like passed all tests including those on Jenkins (with :white_check_mark: ). When I compiled the source code on cygwin, gcc also got such error in building GeoLib.

wenqing:
@endJunction Just dig out that it is the reason, and I just sent a PR to METIS repository, scibuilder/metis#2.

With this fixing, ogs now can be compiled under Cygwin with gcc, and it should be compiled under Mac OS too.

endJunction:
Thanks @wenqing! The Mac_Utils build compiles now.

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/1320).

PR #1324, opened 2016-07-20
[Compile] Enabled compilation under Cygwin and fixed the compilation error under Apple

wenqing:
as titled.
My PR to metis has already been merged. Therefore, issue #1320 should be resolved.

### DIFF
--- a/BaseLib/Subdivision.h
+++ b/BaseLib/Subdivision.h
@@ -75,7 +75,7 @@ public:
             const double dL0,
             const double max_dL,
             const double multiplier)
-    : _L(L), _dL0(dL0), _max_dL(max_dL), _multiplier(multiplier) {}
+    : _length(L), _dL0(dL0), _max_dL(max_dL), _multiplier(multiplier) {}
 
     /// Returns a vector of subdivided points
     std::vector<double> operator()() const
@@ -88,20 +88,20 @@ public:
             vec_x.push_back(x);
             x += std::min(_max_dL, _dL0*std::pow(_multiplier, static_cast<double>(i)));
             i++;
-        } while (x<_L);
+        } while (x<_length);
 
-        if (vec_x.back() < _L) {
+        if (vec_x.back() < _length) {
             double last_dx = vec_x[vec_x.size()-1] - vec_x[vec_x.size()-2];
-            if (_L-vec_x.back()<last_dx)
-                vec_x[vec_x.size()-1] = _L;
+            if (_length-vec_x.back()<last_dx)
+                vec_x[vec_x.size()-1] = _length;
             else
-                vec_x.push_back(_L);
+                vec_x.push_back(_length);
         }
         return vec_x;
     }
 
 private:
-    const double _L;
+    const double _length;
     const double _dL0;
     const double _max_dL;
     const double _multiplier;
--- a/scripts/cmake/CompilerSetup.cmake
+++ b/scripts/cmake/CompilerSetup.cmake
@@ -53,7 +53,9 @@ if(COMPILER_IS_GCC)
         endif()
     endif()
     if(CYGWIN)
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CPU_FLAGS} -std=gnu++0x -Wa,-mbig-obj -Wno-deprecated -Wall -Wextra")
+        # option "-Wa,-mbig-obj" is used to avoid the error of objects having "too many sections".
+        # c.f. (http://stackoverflow.com/questions/16596876/object-file-has-too-many-sections).
+        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CPU_FLAGS} -std=gnu++11 -Wa,-mbig-obj -Wno-deprecated -Wall -Wextra")
     else()
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CPU_FLAGS} -std=c++11 -Wno-deprecated -Wall -Wextra")
     endif()
--- a/scripts/cmake/CompilerSetup.cmake
+++ b/scripts/cmake/CompilerSetup.cmake
@@ -52,7 +52,11 @@ if(COMPILER_IS_GCC)
             set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -D_GLIBCXX_DEBUG -D_GLIBCXX_DEBUG_ASSERT -D_GLIBCXX_DEBUG_PEDASSERT -D_GLIBCXX_DEBUG_VERIFY")
         endif()
     endif()
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CPU_FLAGS} -std=c++11 -Wno-deprecated -Wall -Wextra")
+    if(CYGWIN)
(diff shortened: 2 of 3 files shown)


BaseLib/Subdivision.h
```diff
@@ -88,20 +88,20 @@ class GradualSubdivision : public ISubdivision
...
+                vec_x[vec_x.size()-1] = _length;
             else
-                vec_x.push_back(_L);
+                vec_x.push_back(_length);
         }
         return vec_x;
     }
 
 private:
-    const double _L;
```
wenqing: The name of ``_L`` caused a compilation error under Cygwin with gcc.

scripts/cmake/CompilerSetup.cmake
```diff
@@ -52,7 +52,11 @@ if(COMPILER_IS_GCC)
             set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -D_GLIBCXX_DEBUG -D_GLIBCXX_DEBUG_ASSERT -D_GLIBCXX_DEBUG_PEDASSERT -D_GLIBCXX_DEBUG_VERIFY")
         endif()
     endif()
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CPU_FLAGS} -std=c++11 -Wno-deprecated -Wall -Wextra")
+	if(CYGWIN)
+		set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CPU_FLAGS} -std=gnu++0x -Wa,-mbig-obj -Wno-deprecated -Wall -Wextra")
```
wenqing: Under Cygwin, g++ gives  big object for some heavy template embedded classes that exceeds the ccache limits. Therefore gcc option of  ``-Wa,-mbig-obj`` is added.

endJunction: In cases where something goes wrong for some reasons under some circumstances a version/exact error message/locations etc. are helpful.
 - What is the version of g++
 - Which classes are heavy templated
 - Where are those classes embedded (which library or obj-file)
 - Which limits of ccache are exceeded
???

Also why there is `-std=c++0x` instead of `-std=c++11`?

wenqing:
@endJunction 
* gcc version is  5.4.0 (under Cygiwn). 
* Using -std=c++0x instead of -std=c++11 to avoid compilation errors in gtest such as (c.f http://stackoverflow.com/questions/18784112/googletest-1-6-with-cygwin-1-7-compile-error-fileno-was-not-declared-in-this)
 
```
error: ‘fileno’ was not declared in this scope inline int FileNo(FILE* file) { return fileno(file);
error: ‘strdup’ was not declared in this scope  inline char* StrDup(const char* src) { return strdup(src); 
error: ‘fdopen’ was not declared in this scope inline FILE* FDOpen(int fd, const char* mode) { return fdopen(fd, mode); 
error: ‘mkstemp’ was not declared in this scope const int captured_fd = mkstemp(name_template);
```
*  Using gcc option ``-Wa,-mbig-obj `` under Cygwin to avoid the compilation error of  ``too many sections`` in compiling CoordinatesMapping.cpp. See the error message below.
```
[ 50%] Building CXX object NumLib/CMakeFiles/NumLib.dir/Fem/CoordinatesMapping/NaturalCoordinatesMapping.cpp.o
/usr/lib/gcc/x86_64-pc-cygwin/5.4.0/../../../../x86_64-pc-cygwin/bin/as: /home/wwang/.ccache/4/e/4953d7977279db69a9814a19aeb0c2-3791103.o.tmp.VDIPERFORM024.10344: too many sections (35272)
/tmp/cc74HGQJ.s: Assembler messages:
/tmp/cc74HGQJ.s: Fatal error: can't write /home/wwang/.ccache/4/e/4953d7977279db69a9814a19aeb0c2-3791103.o.tmp.VDIPERFORM024.10344: File too big
/usr/lib/gcc/x86_64-pc-cygwin/5.4.0/../../../../x86_64-pc-cygwin/bin/as: /home/wwang/.ccache/4/e/4953d7977279db69a9814a19aeb0c2-3791103.o.tmp.VDIPERFORM024.10344: too many sections (35272)
/tmp/cc74HGQJ.s: Fatal error: can't close /home/wwang/.ccache/4/e/4953d7977279db69a9814a19aeb0c2-3791103.o.tmp.VDIPERFORM024.10344: File too big
make[2]: *** [NumLib/CMakeFiles/NumLib.dir/build.make:207: NumLib/CMakeFiles/NumLib.dir/Fem/CoordinatesMapping/NaturalCoordinatesMapping.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:3515: NumLib/CMakeFiles/NumLib.dir/all] Error 2
make: *** [Makefile:161: all] Error 2
```

endJunction:
Thanks for the additional information!

The standard should be set with `-std=gnu++11` option which is supported by the gcc-5, because (from gcc man-page) "the name c++0x is deprecated." (Same for gnu).

chleh:
@wenqing
Could you please add a comment including the stackoverflow link and a comment saying that `-Wa,-mbig-obj` prevents the `too many sections` error to the cmake file?
You changed the test input files. Is there a special reason why groundwater flow now uses BiCGStab and why you altered the linear solver tolerances?
Apart from that :+1: from my side.

@endJunction 
> The standard should be set with -std=gnu++11 option which is supported by the gcc-5

What about gcc  < 5?

wenqing:
@chleh  A comment has been added. 
As for the updates of KSP setting for some tests (ufz/ogs-data#12), it has nothing to do with any source code PR but is aimed to enhance the computation for those with big mesh and improve the accuracy for TES tests.

wenqing:
-std=gnu++11 is used because the latest Cygwin has gcc 5.4.

endJunction:
@chleh c++0x is depricated since 4.8.3.

wenqing:
There are two issue for run the complied commends, ogs, testrunner, etc, as
1. Since rpath is does not work under Cygwin, one has to copy shared dlls to the directory
    where the executable files are. For example, in the building directory, run command 
   ``` cp External/vtk/src/vtk-build/bin/cygvtk* bin/ ``` 
2. Three tests fail under Cygwin
```
[  FAILED  ] 3 tests, listed below:
[  FAILED  ] FileIO.TetGenSmeshReader, unknown file: Failure
[  FAILED  ] FileIO.TetGenMeshReaderWithMaterials
   A property with the specified name "MaterialIDs" is not available.
[  FAILED  ] FileIO.TestGmsInterface
   GMSInterface::readGMS3DMMesh(): Could not read expected file header.
```

endJunction:
The cygwin support looks promising but will take some more time to resolve the test issues.
Until then I suggest to take changes to BaseLib/Subdivision.* (make another PR).

For the cygwin support also please keep in mind, there are no jenkins tests for this. I'm also not sure if the support is needed at all (for discussion).

wenqing:
@endJunction  Created a PR #1329 for BaseLib/Subdivision.h

wenqing:
@ There seems be possible to run Jenkens under Cygwin. I agree that let's leave this stuff for future discussion.

endJunction:
Jenkins, skip this build.

endJunction:
No activity for more than 30 days.

endJunction:
The open PR causes Jenkins rebuilds after restarts etc.

PR #1325, opened 2016-07-20
Fix AddTest

chleh:
In #1239 I messed up the vtkdiff tests: One filename and two fieldnames are given, e.g.,
`tes_zeolite_discharge_small_pcs_0_ts_19_t_0.100000.vtu pressure pressure`.
If the two field names are equal, that field is compared to itself yielding zero error norms.

This PR fixes that.

~~TODO: Parallel PETSc tests.~~

### DIFF
--- a/Applications/CLI/Tests.cmake
+++ b/Applications/CLI/Tests.cmake
@@ -13,7 +13,8 @@ if(NOT OGS_USE_MPI)
             WRAPPER time
             TESTER vtkdiff
             ABSTOL 1e-15 RELTOL 1e-15
-            DIFF_DATA cube_${mesh_size}_pcs_0_ts_1_t_1.000000.vtu Linear_1_to_minus1 pressure
+            DIFF_DATA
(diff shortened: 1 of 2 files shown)


chleh:
I had to update the PETSc ctest reference solutions since the mesh partitioning changed in @wenqing's partitioning PR. The absolute differences between the old and the new reference solutions are on the order of 1.e-9.

Applications/CLI/Tests.cmake
```diff
@@ -171,12 +183,12 @@ if(NOT OGS_USE_MPI)
         EXECUTABLE_ARGS tes-1D-zeolite-discharge-large.prj
         WRAPPER time
         TESTER vtkdiff
-        ABSTOL 1e-16 RELTOL 1e-16
+        ABSTOL 1e-3 RELTOL 1e-4
```
TomFischer: Maybe this tolerance is a liitle to sloppy...

chleh: I'll fix it as soon as I can strictly assess the correctness of the TES process, i.e., as soon as the surface extraction & output is ready.

PR #1326, opened 2016-07-21
SWMM Interface

rinkk:
First part of the interface. The committed code is for reading SWMM simulation input files (*.inp) and converting those into geometry or mesh structures.

SWMM is available from the Environmental Protection Agency, the source code is available here:  
https://www.epa.gov/water-research/storm-water-management-model-swmm#downloads.

I'm not sure what the license is. The source code used here is SWMM 5.1.010 and I can't find any licensing information anywhere. SWMM 5.1.009 is [available from Github](https://github.com/OpenWaterAnalytics/SWMM) and the license file looks like BSD although no license name is given.

Edit: [SWMM 5.1.10 is also available on Github](https://github.com/OpenWaterAnalytics/Stormwater-Management-Model) but no licensing information is given there either.


wenqing:
Is it possible to add the SWMM as a submodule?

rinkk:
I don't know. I talked this over with Lars and he said this'd probably the best solution.

Applications/Utils/SWMMConverter/SWMMInterface.cpp
```diff
@@ -0,0 +1,36 @@
+
+
+#include <string>
+#include <iostream>
+#include "ThirdParty/SWMMInterface/swmm5_iface.h"
+
+int main(int argc, char *argv[])
+{
+   // Open outfile as a SWMM output file
+    std::string outfile = argv[1];
```
TomFischer: If no command line argument is given the tool will probably crash without a useful error message.

Applications/Utils/SWMMConverter/SWMMInterface.cpp
```diff
@@ -0,0 +1,36 @@
+
```
TomFischer: OGS license header is missing.

Applications/Utils/SWMMConverter/SWMMInterface.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include "ThirdParty/SWMMInterface/swmm5_iface.h"
+
+int main(int argc, char *argv[])
+{
+   // Open outfile as a SWMM output file
+    std::string outfile = argv[1];
+    int r = OpenSwmmOutFile(const_cast<char*>(outfile.c_str()));
+    if (r == 1)
+    {
+        printf("\nInvalid results in SWMM output file.\n");
```
TomFischer: You could use logog for output.

Applications/Utils/SWMMConverter/SWMMConverter.cpp
```diff
@@ -0,0 +1,60 @@
+
```
TomFischer: OGS license header.

Applications/Utils/SWMMConverter/SWMMInterface.h
```diff
@@ -0,0 +1,136 @@
...
+        std::vector<MeshLib::Node*> const& nodes, std::map<std::string, std::size_t> const& name_id_map);
+
+    /// Reading polygons from SWMM input file and converting them into OGS polyline vector.
+    static int readPolygons(std::ifstream &in, std::vector<GeoLib::Polyline*> &lines,
+        std::vector<std::string> &line_names, std::vector<GeoLib::Point*> &points,
+        std::vector<std::string> &pnt_names);
+
+    //variables
+    std::string const _base_name;
+    MeshLib::Mesh * _mesh;
```
TomFischer: Can `_mesh` be a reference?

Applications/Utils/SWMMConverter/SWMMInterface.cpp
rinkk: changed in later commit

Applications/Utils/SWMMConverter/SWMMInterface.cpp
rinkk: changed in later commit

Applications/Utils/SWMMConverter/SWMMInterface.h
```diff
@@ -0,0 +1,136 @@
...
+    static int SwmmInputToGeometry(std::string const& inp_file_name, GeoLib::GEOObjects &geo_objects);
+
+
+private:
+    /// Checking if file is a SWMM input file
+    static int isSwmmFile(std::string const& inp_file_name, std::ifstream &in, std::string const& ext);
+
+    /// Reading points from SWMM input file and converting them into OGS point vector.
+    template <typename T>
+    static int readCoordinates(std::ifstream &in, std::vector<T*> &points, std::vector<std::string> &names)
```
TomFischer: From the implementation it would be possible to return a bool.

Applications/Utils/SWMMConverter/SWMMInterface.h
rinkk: Since the mesh will/may be created by the interface this cannot be a reference.

wenqing:
Since SWMM is available in github, I think it can be added as a submodule. By this way, you will not lose  the updates from the SWMM source code, and the updates are applied automatically.

rinkk:
While SWMM is available on GitHub, the interface (swmm5_iface) is not. Not sure how to deal with that.

endJunction:
Isn't [swmm5.h](https://github.com/OpenWaterAnalytics/Stormwater-Management-Model/blob/master/src/swmm5.h) not providing the same things (at least the functions look very similar)? Or maybe it is the [iface.c](https://github.com/OpenWaterAnalytics/Stormwater-Management-Model/blob/master/src/iface.c)?

Applications/Utils/SWMMConverter/SWMMInterface.h
rinkk: I'd rather keep it int for all functions so I can give meaningful error messages from gui.

rinkk:
swmm5 *does* provide the underlying functionality, but that'd mean I'd have to write the 300 LoC in swmm_iface.c myself (i.e. copy it somewhere) to make it actually usable. Which is fishy concerning license-issues...

PR #1327, opened 2016-07-22
GeoMapper improvements

TomFischer:
This PR contains a reimplementation of  `GeoMapper::advancedMapOnMesh` because the old implementation had some issues, see issue #944. For the new implementation three new tests are added.


MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -352,6 +356,342 @@ void GeoMapper::advancedMapOnMesh(
...
+///
+/// The algorithm projects every element of the elements vector and the point
+/// \c p orthogonal to the \f$x\f$-\f$y\f$ plane. In the \f$x\f$-\f$y\f$ plane
+/// it is checked if the projeted point is in the projected element.
+static MeshLib::Element const* findElementContainingPoint2d(
+    std::vector<MeshLib::Element const*> const& elements,
+    MathLib::Point3d const& p)
+{
+    for (auto const elem : elements) {
+        std::unique_ptr<MeshLib::Element> elem_2d(elem->clone());
```
chleh: Must that be a unique_ptr?

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -352,6 +356,342 @@ void GeoMapper::advancedMapOnMesh(
...
+        GeoLib::LineSegment elem_segment{
+            new GeoLib::Point(*dynamic_cast<MathLib::Point3d*>(
+                const_cast<MeshLib::Node*>(edge->getNode(0))), 0),
+            new GeoLib::Point(*dynamic_cast<MathLib::Point3d*>(
+                const_cast<MeshLib::Node*>(edge->getNode(1))), 0),
+            false};
+        std::vector<MathLib::Point3d> const intersections(
+            GeoLib::lineSegmentIntersect2d(segment, elem_segment));
+        for (auto const& p : intersections)
+            element_intersections.emplace_back(p);
```
chleh: Points could possibly be moved.

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -352,6 +356,342 @@ void GeoMapper::advancedMapOnMesh(
...
+    if (intersections.size() > 2) {
+        std::stringstream out;
+        out << "element with id " << elem->getID() << " and seg "
+                  << " intersecting at more than two edges\n";
+        for (std::size_t k(0); k<intersections.size(); ++k)
+            out << k << " " << intersections[k]
+                      << "\n";
+        out << "Could not map segment on element. Aborting.\n";
+        ERR("%s", out.str().c_str());
+        std::abort();
```
chleh: OGS_FATAL?

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -352,6 +356,342 @@ void GeoMapper::advancedMapOnMesh(
...
+            std::numeric_limits<double>::epsilon())
+            sub_segments.emplace_back(
+                GeoLib::LineSegment{new GeoLib::Point{intersections[0], 0},
+                                    new GeoLib::Point{end_pnt, 0}, true});
+    }
+
+    if (intersections.size() == 1 && (elem != beg_elem && elem != end_elem))
+    {
+        // Since the line segment enters and leaves the element in the same
+        // point there isn't any need to insert a new sub line segment.
```
chleh: Can that happen? If a line intersects an element in a point, i.e., in a corner of the element, it always intersects two edges, doesn't it? Or am I misunderstanding the entire matter?

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -352,6 +356,342 @@ void GeoMapper::advancedMapOnMesh(
...
+        sub_segments.emplace_back(
+            GeoLib::LineSegment{new GeoLib::Point{intersections[0], 0},
+                                new GeoLib::Point{intersections[1], 0}, true});
+    }
+    return sub_segments;
+}
+
+static std::vector<GeoLib::LineSegment> mapLineSegment(
+    GeoLib::LineSegment const& segment,
+    std::vector<MeshLib::Element const*> const& surface_elements,
```
chleh: Is that restricted to elements lying at the surface of a domain?

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -352,6 +356,342 @@ void GeoMapper::advancedMapOnMesh(
...
+        if (element_intersections.empty())
+            continue;
+
+        BaseLib::makeVectorUnique(element_intersections);
+
+        std::vector<GeoLib::LineSegment> sub_seg_elem(
+            createSubSegmentsForElement(element_intersections, beg_elem,
+                                        end_elem, beg_pnt, end_pnt, elem));
+        sub_segments.insert(sub_segments.end(), sub_seg_elem.begin(),
+                            sub_seg_elem.end());
```
chleh: Don't you also have to make the sub_segments unique?

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -352,6 +356,342 @@ void GeoMapper::advancedMapOnMesh(
...
+    }
+
+    // beg_elem == nullptr means there isn't any element corresponding to the
+    // beg_pnt and as a consequence the above algorithm doesn't insert a sub
+    // segment
```
chleh: If the above loop does nothing in this case, shouldn't it then be inside the else-clause of this if statement?

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -352,6 +356,342 @@ void GeoMapper::advancedMapOnMesh(
...
+                                     MathLib::Point3d& q)
+{
+    // create plane equation: n*p = d
+    MathLib::Vector3 const& p(*(elem.getNode(0)));
+    MathLib::Vector3 const n(MeshLib::FaceRule::getSurfaceNormal(&elem));
+    if (n[2] == 0.0) { // vertical plane, z coordinate is arbitrary
+        q[2] = p[2];
+    } else {
+        double const d(MathLib::scalarProduct(n, p));
+        q[2] = (d - n[0]*q[0] - n[1]*q[1])/n[2];
```
chleh: You only modify z coordinates. I.e., that's not an orthogonal projection of the point `q` onto the element. Is that what you want?

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -352,6 +356,342 @@ void GeoMapper::advancedMapOnMesh(
...
+    return sub_segments;
+}
+
+static void mapPointOnSurfaceElement(MeshLib::Element const& elem,
+                                     MathLib::Point3d& q)
+{
+    // create plane equation: n*p = d
+    MathLib::Vector3 const& p(*(elem.getNode(0)));
+    MathLib::Vector3 const n(MeshLib::FaceRule::getSurfaceNormal(&elem));
+    if (n[2] == 0.0) { // vertical plane, z coordinate is arbitrary
```
chleh: x and y coordinates are arbitrary, you fix z, don't you?

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -352,6 +356,342 @@ void GeoMapper::advancedMapOnMesh(
...
+
+    return candidate_elements;
+}
+
+static bool snapPointToElementNode(MathLib::Point3d& p,
+                                   MeshLib::Element const& elem, double eps)
+{
+    for (std::size_t k(0); k<elem.getNumberOfNodes(); ++k) {
+        double const sqr_dist_2d(MathLib::sqrDist2d(p, *(elem.getNode(k))));
+        if (sqr_dist_2d < eps) {
```
chleh: the epsilon here (and in other places) is given for the squared distance. Is that consistent everywhere? Personally, I'd expect that tolerances are given w.r.t. the distance, not to its square.

MeshGeoToolsLib/GeoMapper.cpp
TomFischer: You are correct, I don't want an orthogonal projection of the point `q` onto the element. It is intended only to modify the `z` coordinate.

MeshGeoToolsLib/GeoMapper.cpp
TomFischer: As far as I can see `eps` is always used in squared distance comparisons in my code.

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -352,6 +356,342 @@ void GeoMapper::advancedMapOnMesh(
...
+    {
+        auto candidate_elements(getCandidateElementsForLineSegmentIntersection(
+            mesh_element_grid, *segment_it));
+
+        auto mapPoint = [&candidate_elements](MathLib::Point3d& p) {
+            auto const* elem(
+                findElementContainingPoint2d(candidate_elements, p));
+            if (elem)
+            {
+                if (!snapPointToElementNode(p, *elem, 1e-3))
```
chleh: Why the hard-coded tolerance?

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -352,6 +356,342 @@ void GeoMapper::advancedMapOnMesh(
...
+                    mapPointOnSurfaceElement(*elem, p);
+            }
+            return elem;
+        };
+
+        // map segment begin and end point
+        auto const* beg_elem(mapPoint((*segment_it).getBeginPoint()));
```
chleh: Can't you use `->` here?

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -352,6 +356,342 @@ void GeoMapper::advancedMapOnMesh(
...
+{
+    // 1. extract surface
+    if (_surface_mesh)
+        delete _surface_mesh;
+    if (mesh.getDimension()<3) {
+        _surface_mesh = new MeshLib::Mesh(mesh);
+    } else {
+        const MathLib::Vector3 dir(0,0,-1);
+        _surface_mesh =
+            MeshLib::MeshSurfaceExtraction::getMeshSurface(mesh, dir, 91);
```
chleh: 91?

chleh:
Code looks good :+1:

MeshGeoToolsLib/GeoMapper.cpp
TomFischer: I will change variable name to better reflect that the comparsion is done with squared tolerance.

MeshGeoToolsLib/GeoMapper.cpp
TomFischer: Thanks, I'll move the points.

MeshGeoToolsLib/GeoMapper.cpp
TomFischer: I can't, since `->` is not overloaded yet for the SegmentIterator class.

MeshGeoToolsLib/GeoMapper.cpp
TomFischer: This value is an angle. I want an angle a little bit greater than 90. Maybe 90 + eps should be enough. I will change this.

MeshGeoToolsLib/GeoMapper.cpp
TomFischer: Geometric tolerances are almost always hard to define. `1e-3` means that the *squared* distance between the `p` and a node of the element are snapped when is smaller than this tolerance value. I think this is a strong criterion.

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -212,208 +217,339 @@ double GeoMapper::getMeshElevation(
...
+{
+    for (std::size_t k(0); k<elem.getNumberOfNodes(); ++k) {
+        double const sqr_dist_2d(MathLib::sqrDist2d(p, *(elem.getNode(k))));
+        if (sqr_dist_2d < sqr_eps) {
+#ifndef NDEBUG
+            std::stringstream out;
+            out.precision(std::numeric_limits<double>::digits10);
+            out << "Segment point snapped from " << p;
+#endif
+            p[0] = (*(elem.getNode(k)))[0];
```
wenqing: Use a loop or introduce a constant, e.g., auto const* node =  elem.getNode(k)?

endJunction: ... immideatly after the for loop begin.

MeshGeoToolsLib/GeoMapper.cpp
chleh: I'd say that depends on the order of magnitude of your coordinates. sqrt(1e-3) ~ 0.03, that's several elements wide for my meshes.
The best solution solution would be to have an argument `snap_distance`. If nobody else complains, however, then you might keep this magic number.

MeshGeoToolsLib/GeoMapper.cpp
chleh: OK, I also assumed something like that. You needn't change it, it only looks a bit odd. Can the `getMeshSurface()` method be adapted to have some tolerance (optional)?

TomFischer: The `angle` is already a tolerance. A face of an mesh element belongs to the (geometric) surface described analytical by its normal - in the code the `dir` vector - iff the scalar product between the face normal of the mesh element and the vector `dir`  is smaller than the given `angle`.

Please correct me, if I misunderstood you.

MeshGeoToolsLib/GeoMapper.cpp
TomFischer: 1. The *squared* distance have to smaller than `1e-3`, i.e., the distance has to be smaller than `1e-6`. I hope this is acceptable in your applications?
2. Since in adaptive meshes the "mesh density" can vary a lot, there can not be a single value for this tolerance. So far I haven't any idea to solve this issue in general.
3. In this particular case a possible solution would be to use the mesh element size.

endJunction:
CTest failing because you need to rebase Tests/Data branch...

MeshGeoToolsLib/GeoMapper.cpp
TomFischer: Changed the implementation of `snapPointToElementNode` using the hard coded "eps" as a relative tolerance.

MeshGeoToolsLib/GeoMapper.cpp
chleh: So this line effectively returns a `_surface_mesh` containing all those faces (of elements at the surface) whose normal has an angle of less than 91 degrees with the vector `dir = (0, 0, -1)`?

TomFischer: Yes, it returns such a surface mesh.

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -433,8 +433,13 @@ getCandidateElementsForLineSegmentIntersection(
...
 
 static bool snapPointToElementNode(MathLib::Point3d& p,
-                                   MeshLib::Element const& elem, double sqr_eps)
+                                   MeshLib::Element const& elem, double rel_eps)
 {
+    // values will be initialized within computeSqrNodeDistanceRange
+    double sqr_min, sqr_max;
+    elem.computeSqrNodeDistanceRange(sqr_min, sqr_max);
+
+    double const sqr_eps(rel_eps*rel_eps * sqr_min);
```
chleh: So now if e.g. `rel_eps = 1e-3`, then the tolerance in distances is 1e-3 times the minimum edge length of the element? Sounds good.
Note: Now rel_eps is not a tolerance of "distance squared" anymore, but of "distance".

chleh:
:+1:

Applications/DataExplorer/mainwindow.cpp
```diff
@@ -857,7 +858,7 @@ void MainWindow::mapGeometry(const std::string &geo_name)
     else // use mesh from ProjectData
         mesh = _project.getMeshObjects()[choice-2].get();
 
-    std::string const& new_geo_name = dlg.getNewGeoName();
+    std::string new_geo_name = dlg.getNewGeoName();
```
endJunction: Just a note: Better to keep this one const and make another one in lines 870s to clarify the dependencies of the names...

BaseLib/makeVectorUnique.h
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#pragma once
```
endJunction: For now we stick to the include guards, pragma once are not yet welcome...

GeoLib/LineSegment.cpp
```diff
@@ -93,4 +93,15 @@ std::ostream& operator<<(std::ostream& os,
...
+
+bool operator==(LineSegment const& s0, LineSegment const& s1)
+{
+    double const tol(std::numeric_limits<double>::epsilon());
+    if ((MathLib::sqrDist(s0.getBeginPoint(), s1.getBeginPoint()) < tol &&
+         MathLib::sqrDist(s0.getEndPoint(), s1.getEndPoint()) < tol) ||
+        (MathLib::sqrDist(s0.getBeginPoint(), s1.getEndPoint()) < tol &&
+         MathLib::sqrDist(s0.getEndPoint(), s1.getBeginPoint()) < tol))
+        return true;
+    return false;
```
endJunction:
```c++
return MathLib::sqrDist(..) && ...
```
instead of if (true) return true else return false

GeoLib/LineSegment.h
```diff
@@ -50,5 +50,8 @@ std::ostream& operator<< (std::ostream& os, LineSegment const& s);
 std::ostream& operator<<(std::ostream& os,
                          std::pair<GeoLib::LineSegment const&,
                                    GeoLib::LineSegment const&> const& seg_pair);
+
+bool operator==(LineSegment const& s0, LineSegment const& s1);
```
endJunction: Please mention in the comment that the equality is not strict...

GeoLib/PointVec.cpp
```diff
@@ -249,4 +249,25 @@ void PointVec::setNameForElement(std::size_t id, std::string const& name)
...
+    MathLib::Point3d const& max(_aabb.getMaxPoint());
+    double const rel_eps(_rel_eps / std::sqrt(MathLib::sqrDist(min, max)));
+
+    _aabb = GeoLib::AABB(_data_vec->begin(), _data_vec->end());
+
+    _rel_eps = rel_eps * std::sqrt(MathLib::sqrDist(_aabb.getMinPoint(),
+                                                    _aabb.getMaxPoint()));
+
+    _oct_tree.reset(OctTree<GeoLib::Point, 16>::createOctTree(
+        _aabb.getMinPoint(), _aabb.getMaxPoint(), _rel_eps));
```
endJunction: one can use `min` and `max` to make the code shorter, but there will be no difference in the binary...

GeoLib/PointVec.h
```diff
@@ -103,6 +103,7 @@ class PointVec final : public TemplateVec<Point>
 
     void setNameForElement(std::size_t id, std::string const& name) override;
 
+    void resetInternalDataStructures();
```
endJunction: Please add a comment what is rebuild and if something is invalidated after this function (like pointers...).

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -212,208 +217,343 @@ double GeoMapper::getMeshElevation(
...
 
-std::unique_ptr<std::vector<GeoLib::Polyline*>> copyPolylinesVector(
-    std::vector<GeoLib::Polyline*> const& polylines,
-    std::vector<GeoLib::Point*> const& points)
+/// Find the 2d-element within the \c elements that contains the given point \c p.
+///
+/// The algorithm projects every element of the elements vector and the point
+/// \c p orthogonal to the \f$x\f$-\f$y\f$ plane. In the \f$x\f$-\f$y\f$ plane
+/// it is checked if the projeted point is in the projected element.
+static MeshLib::Element const* findElementContainingPoint2d(
```
endJunction: Maybe better call this one ...XY instead of 2d, because the latter is not unique...

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -212,208 +217,343 @@ double GeoMapper::getMeshElevation(
...
 }
 
-std::unique_ptr<std::vector<GeoLib::Polyline*>> copyPolylinesVector(
-    std::vector<GeoLib::Polyline*> const& polylines,
-    std::vector<GeoLib::Point*> const& points)
+/// Find the 2d-element within the \c elements that contains the given point \c p.
+///
+/// The algorithm projects every element of the elements vector and the point
+/// \c p orthogonal to the \f$x\f$-\f$y\f$ plane. In the \f$x\f$-\f$y\f$ plane
+/// it is checked if the projeted point is in the projected element.
```
endJunction: typo: projected

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -212,208 +217,343 @@ double GeoMapper::getMeshElevation(
...
+        if (sqr_dist_2d < sqr_eps) {
+#ifndef NDEBUG
+            std::stringstream out;
+            out.precision(std::numeric_limits<double>::digits10);
+            out << "Segment point snapped from " << p;
+#endif
+            p = node;
+#ifndef NDEBUG
+            out << "to " << p;
+            DBUG("%s", out.str().c_str());
```
endJunction: Not sure here, but if this output is being showed in debug build of the testrunner, then better to comment it...

MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -212,208 +217,343 @@ double GeoMapper::getMeshElevation(
...
+        auto const* beg_elem(mapPoint((*segment_it).getBeginPoint()));
+        auto const* end_elem(mapPoint((*segment_it).getEndPoint()));
+
+        // Since the mapping of the segment begin and end points the coordinates
+        // changed. The internal data structures of PointVec are possibly
+        // invalid and hence it is necessary to re-create them.
+        orig_points.resetInternalDataStructures();
+
+        if (beg_elem == end_elem)  // TODO: handle case: beg_elem == end_elem == nullptr
+            continue;
```
endJunction: I assume, that the mentioned todo case does not happen (yet); Then it would be helpful if such a case happens, to have an assert inside the if.
```c++
if (b == e) {assert(b != nullptr}; continue;}
```

MeshLib/MeshSurfaceExtraction.cpp
```diff
@@ -70,11 +70,13 @@ MeshLib::Mesh* MeshSurfaceExtraction::getMeshSurface(
     const MeshLib::Mesh& mesh, const MathLib::Vector3& dir, double angle,
     std::string const& subsfc_node_id_backup_prop_name)
 {
+/*
```
endJunction: Needs a comment why the block is still needed; remove it otherwise...

endJunction:
Add the comments, fix typo, and optionally other things, then merge, when ready.

GeoLib/PointVec.cpp
TomFischer: `min` and `max` are the min point and max point *before* the reset, i.e., before the recomputation of the aabb object. I need the "old" values for the computation of `_rel_eps`. After the reset min and max point will have changed.

MeshGeoToolsLib/GeoMapper.cpp
TomFischer: Changed to `#ifdef DEBUG_GEOMAPPER`.

MeshGeoToolsLib/GeoMapper.cpp
TomFischer: We have to distingush two cases (needs some work):
1.  `beg_elem == end_elem` and the segment intersects elements
2.  `beg_elem == end_elem` and the segment does not intersect any element, i.e., the segment is located outside of the mesh area.

The handling of the first case needs more additional work.

PR #1329, opened 2016-07-26
[Compile] Fixed a potential compilation error

wenqing:
The name of _L causes a compilation error by using gcc under Cygwin.


endJunction:
I tried to force travis to recompile it, but their cmake crashes with a segv; other checks are all OK.

PR #1331, opened 2016-07-27
Fix dots in VTU file names

chleh:
#1319 fixed pvtu file names. But the solution causes an issue if some directories in the file path contains dots in their name. This PR makes sure that only dots in the basename of the pvtu file are replaced. For the rest see the commit messages.

### DIFF
--- a/BaseLib/FileTools.cpp
+++ b/BaseLib/FileTools.cpp
@@ -141,11 +141,11 @@ std::string extractPath(std::string const& pathname)
         return "";
     return pathname.substr(0, pos + 1);
 }
-static const char * pathSeparator =
+static const char pathSeparator =
 #ifdef _WIN32
-                            "\\";
+                            '\\';
 #else
-                            "/";
+                            '/';
 #endif
 
 std::string appendPathSeparator(std::string const& path)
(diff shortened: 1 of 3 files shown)


endJunction:
:fast_forward:

BaseLib/FileTools.cpp
```diff
@@ -141,11 +141,11 @@ std::string extractPath(std::string const& pathname)
         return "";
     return pathname.substr(0, pos + 1);
 }
-static const char * pathSeparator =
+static const char pathSeparator =
```
wenqing: can it be "/\\" without the following #ifdef?

MeshLib/IO/VtkIO/VtuInterface.cpp
```diff
@@ -77,12 +77,17 @@ bool VtuInterface::writeToFile(std::string const &file_name)
     // and PETSC_COMM_WORLD should be replaced with the argument.
     int mpi_rank;
     MPI_Comm_rank(PETSC_COMM_WORLD, &mpi_rank);
-    std::string file_name_base = boost::erase_last_copy(file_name, ".vtu");
+    auto const file_name_base = boost::erase_last_copy(file_name, ".vtu");
+
+    auto const dirname = BaseLib::extractPath(file_name_base);
+    auto basename = BaseLib::extractBaseName(file_name_base);
```
wenqing: basename-->basename_without_path

wenqing:
:+1:

BaseLib/FileTools.cpp
wenqing: Display problem of "/\\" in my comment, which should be ``"/\\"``.

chleh: A backslash as path separator would definitely be wrong for unix. Furthermore having a string of length > 1 would make some functions in this file slightly more complicated. So I don't want to change it, and the ifdef has been there also before this PR.

MeshLib/IO/VtkIO/VtuInterface.cpp
chleh: I chose the variable names in correspondence with the unix cmdline tools `dirname` and `basename`. So the basename does not contain any leading directories.

PR #1332, opened 2016-07-28
Functionality for writing CSV files

rinkk:
Vectors are added to object and then written to the file in the sequence they have been added in.
Currently only double/int/string vectors are supported.

### DIFF
--- a/Applications/FileIO/CsvInterface.cpp
+++ b/Applications/FileIO/CsvInterface.cpp
@@ -22,6 +22,12 @@
 
 namespace FileIO {
 
+CsvInterface::CsvInterface()
+: _writeCsvHeader(true)
+{
+}
+
+
 int CsvInterface::readPoints(std::string const& fname, char delim,
                              std::vector<GeoLib::Point*> &points)
 {
@@ -198,31 +204,26 @@ bool CsvInterface::write()
         return false;
     }
 
-    std::size_t const vec_size (getVectorSize(0));
     std::size_t const n_vecs (_data.size());
-    for (std::size_t i=1; i<n_vecs; ++i)
+    std::size_t const vec_size (getVectorSize(0));
+
+    if (_writeCsvHeader)
     {
-        if (getVectorSize(i) != vec_size)
-        {
-            ERR ("CsvInterface::write() - Data vectors are not equally long.");
-            return false;
-        }
+        _out << _vec_names[0];
+        for (std::size_t i=1; i<n_vecs; ++i)
+            _out << "\t" << _vec_names[i];
+        _out << "\n";
     }
 
-    _out << _vec_names[0];
-    for (std::size_t i=1; i<n_vecs; ++i)
-        _out << "\t " << _vec_names[i];
-    _out << std::endl;
-
-    for (std::size_t j=1; j<vec_size; ++j)
+    for (std::size_t j=0; j<vec_size; ++j)
     {
         writeValue(0,j);
         for (std::size_t i=1; i<n_vecs; ++i)
         {
-            _out << "\t ";
+            _out << "\t";
             writeValue(i,j);
         }
-        _out << std::endl;
+        _out << "\n";
     }
     return true;
 }
--- a/Applications/FileIO/CsvInterface.cpp
+++ b/Applications/FileIO/CsvInterface.cpp
@@ -14,6 +14,8 @@
 #include "CsvInterface.h"
 
 #include <algorithm>
+#include <iostream>
+#include <numeric>
 #include <stdexcept>
 
 #include "GeoLib/Point.h"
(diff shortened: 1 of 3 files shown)


Applications/FileIO/CsvInterface.cpp
```diff
@@ -181,4 +183,69 @@ std::size_t CsvInterface::findColumn(std::string const& line, char delim, std::s
...
+        {
+            ERR ("CsvInterface::write() - Data vectors are not equally long.");
+            return false;
+        }
+    }
+
+    _out << _vec_names[0];
+    for (std::size_t i=1; i<n_vecs; ++i)
+        _out << "\t " << _vec_names[i];
+    _out << std::endl;
```
endJunction: Better to use "\n" here and below because std::endl flushes the buffer, which can result in worse performance than necessary.

Applications/FileIO/CsvInterface.h
```diff
@@ -36,9 +39,36 @@ namespace FileIO {
...
 
 public:
+    /// Contructor (only needed for writing files)
+    CsvInterface() {};
+
+    /// Adds an index vector of size s to the CSV file
+    void addIndexVectorForWriting(std::size_t s);
+
+    /// Adds a data vector to the CSV file. All data vectors have to have the same size.
+    /// Vectors will be written in the same sequence they have been added to the interfaceW.
```
endJunction: The sizes are not checked here, but in the write function, which might be too late already (e.g. to be able to trace where the data comes from). Better to have the "equal size" invariant in the class.

Typo? interfaceW. ?

endJunction:
Round-trip test?

Applications/FileIO/CsvInterface.h
```diff
@@ -36,9 +39,36 @@ namespace FileIO {
...
+    template<typename T>
+    bool addVectorForWriting(std::string const& vec_name, std::vector<T> const& vec)
+    {
+        if (typeid(vec) == typeid(std::vector<std::string>) ||
+            typeid(vec) == typeid(std::vector<double>) ||
+            typeid(vec) == typeid(std::vector<int>))
```
chleh: You check that at runtime is there a reason why you don't do it at compile time?

Applications/FileIO/CsvInterface.cpp
```diff
@@ -181,4 +183,69 @@ std::size_t CsvInterface::findColumn(std::string const& line, char delim, std::s
...
+            writeValue(i,j);
+        }
+        _out << std::endl;
+    }
+    return true;
+}
+
+std::size_t CsvInterface::getVectorSize(std::size_t idx) const
+{
+    if (_data[idx].type() == typeid(std::vector<std::string>))
```
chleh: Note: You could probably do something like
```
if (auto const* vec_string = boost::any_cast<std::vector<std::string>*>(&_data[idx]))
    return vec_string->size();
```

Applications/FileIO/CsvInterface.cpp
```diff
@@ -181,4 +183,69 @@ std::size_t CsvInterface::findColumn(std::string const& line, char delim, std::s
...
+        return boost::any_cast<std::vector<double>>(_data[idx]).size();
+    else if (_data[idx].type() == typeid(std::vector<int>))
+        return boost::any_cast<std::vector<int>>(_data[idx]).size();
+    return 0;
+}
+
+void CsvInterface::writeValue(std::size_t vec_idx, std::size_t in_vec_idx)
+{
+    if (_data[vec_idx].type() == typeid(std::vector<std::string>))
+        _out << boost::any_cast<std::vector<std::string>>(_data[vec_idx])[in_vec_idx];
```
chleh: Is the precision of `_out` correctly set somewhere?

chleh:
After questions :+1:

Applications/FileIO/CsvInterface.h
rinkk: I don't understand what you mean...

Applications/FileIO/CsvInterface.cpp
rinkk: What'd be the advantage?

Applications/FileIO/CsvInterface.cpp
rinkk: It's set in the Writer base class.

Applications/FileIO/CsvInterface.cpp
chleh: Only one casting/typeiding around. It's just a note: Annoying people in order to educate them. ^^

Applications/FileIO/CsvInterface.h
chleh: The checks in the if clause are evaluated at run-time. However, it is possible to rewrite the `addVectorForWriting()` method s.t. the compiler will raise an error if you provide data of the wrong type.
Is there a reason that the compiler should not quit with such an error?

rinkk: I'm still not sure what you mean. The interface should be general so the programme shouldn't crash/exit if a vector of the wrong type is given.

Example: You load a bunch of data sets into the data explorer and decide you want to write all the parameters of *one* data set as a CSV file.  You cannot know at compile time which type of params the data set contains so you need to check at run-time. If a param is in there whose type is not supported, it should not crash the programme but instead be able to write all other params and give an error message that one param could not be written. 

Does that help?

chleh: Let's talk on monday.

chleh: As just discussed I sent you a small PR that implements the compile-time type check. https://github.com/rinkk/ogs/pull/15

Tests/FileIO/TestCsvWriter.cpp
```diff
@@ -0,0 +1,80 @@
...
+TEST(CsvWriter, WriteReadTest)
+{
+    std::string test_file(BaseLib::BuildInfo::tests_tmp_path + "TestData.csv");
+
+    std::vector<std::string> str_vec {"Red", "Orange", "Yellow", "Green", "Blue", "Indigo", "Violet" };
+    std::vector<int> int_vec { 1, 2, 4, 8, 16, 32, 64 };
+    std::vector<double> dbl_vec;
+    std::srand ( static_cast<unsigned>(std::time(nullptr)) );
+    for (std::size_t i=0; i<int_vec.size(); ++i)
+        dbl_vec.push_back(static_cast<double>(rand()) / RAND_MAX);
```
endJunction: Better to prefix the STL functions; std::rand().

Tests/FileIO/TestCsvWriter.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+    std::vector<int> idx_result;
+    result = FileIO::CsvInterface::readColumn<int>(test_file, '\t', idx_result, "Index");
+    ASSERT_EQ(0, result);
+    ASSERT_EQ(dbl_vec.size(), idx_result.size());
+
+    std::vector<int> int_result;
+    result = FileIO::CsvInterface::readColumn<int>(test_file, '\t', int_result, vec_names[1]);
+    ASSERT_EQ(0, result);
+    ASSERT_EQ(int_vec.size()-1, int_result.size());
```
endJunction: Could you explain why the `-1`? It is not obvious to me.

endJunction:
The definition of CsvInterface ctor is missing; cannot link:
```
[ 71%] Linking CXX executable ../bin/testrunner
CMakeFiles/testrunner.dir/FileIO/TestCsvWriter.cpp.o: In function
    `CsvWriter_WriteReadTest_Test::TestBody()':
ogs/Tests/FileIO/TestCsvWriter.cpp:32:
    undefined reference to `FileIO::CsvInterface::CsvInterface()'
```

I'm not sure why this error didn't appear in other compilations. Clang-3.8, dynamic linking...

Tests/FileIO/TestCsvWriter.cpp
rinkk: Because I pushed an additional element into int_vector in line 41 to test if a vector of different length would be rejected. So now int_vector is one element longer than the other ones.

rinkk:
I don't understand the error, the ctor is there (CsvInterface.cpp, Line 25).

Tests/FileIO/TestCsvWriter.cpp
endJunction: OK. I see. Please add a short comment in the code.

endJunction:
You are right. Ctor is there. I'm looking into this issue, for now please ignore it...

endJunction:
General note: I think here two things are mixed up in a single class, one is a table for arbitrary types (or selection of them like int, double, string) and IO interface for that table. If somebody needs a table this class could be a possible starting point.

PR #1336, opened 2016-08-03
Added uniform Robin BC

chleh:
This is a successor of #1337.

Missing features:
* The Robin BC currently cannot be used with the Newton-Raphson scheme out of the box. Currently there is no process ready for Newton, so I postponed this feature.

PR #1337, opened 2016-08-03
Restructured bcs

chleh:
This is a predecessor of #1336.

Most of the code changes are simple moves from old BC classes to new ones.
Major changes:
* Got rid of different BC-config classes. Instead there is now one simple BC config class used in `ProcessVariable`
* All BCs now have a common base class. (The concept of it is not perfect since the Dirichlet BCs don't quite fit, but it can be used as a starting point for extensions/generalizations)
* There is a generic implementation for "generic natural BCs" (like Neumann and Robin; please suggest a better name) that omits code duplication pretty much. New BCs essentially only need to implement an assembly method and a create function.

Btw: This PR is #1337 :smile_cat:


chleh:
Just to clarify: There are many lines of changes in this PR, but there are only very few new features.
Interesting might be:
* the general BC interface `ProcessLib/BoundaryCondition/BoundaryCondition.h`
* the Dirichlet BC interface `ProcessLib/BoundaryCondition/DirichletBoundaryCondition.h` which misuses the general BC interface a bit.
* the BC collection `ProcessLib/BoundaryCondition/BoundaryConditionCollection.*`
* the generic Neumann/Robin/... BC `ProcessLib/BoundaryCondition/GenericNaturalBoundaryCondition*` (note: the code is essentially moved from the old Neumann BC classes)

The other changes are mainly code moves with minor changes, and dependent changes in order to cope with the moved code and to use the new classes.

ProcessLib/BoundaryCondition/BoundaryCondition.h
```diff
@@ -0,0 +1,46 @@
...
+}
+
+namespace ProcessLib
+{
+struct BoundaryConditionConfig;
+
+class BoundaryCondition
+{
+public:
+    virtual void apply(const double t, GlobalVector const& x, GlobalMatrix& K,
```
endJunction: Needs a short comment...

ProcessLib/BoundaryCondition/UniformDirichletBoundaryCondition.cpp
```diff
@@ -11,70 +11,57 @@
...
+createUniformDirichletBoundaryCondition(
+    BoundaryConditionConfig const& config,
+    MeshGeoToolsLib::MeshNodeSearcher& searcher,
+    NumLib::LocalToGlobalIndexMap const& dof_table,
+    int const variable_id)
 {
     DBUG("Constructing UniformDirichletBoundaryCondition from config.");
     //! \ogs_file_param{boundary_condition__type}
-    config.checkConfigParameter("type", "UniformDirichlet");
+    config.config.checkConfigParameter("type", "UniformDirichlet");
```
endJunction: Just an idea: Would deriving BCConfig from ConfigTree be possible, and would it be nice too?

ProcessLib/ProcessVariable.cpp
```diff
@@ -155,4 +136,19 @@ MeshLib::PropertyVector<double>& ProcessVariable::getOrCreateMeshProperty()
     }
 }
 
+std::vector<std::unique_ptr<BoundaryCondition>>
+ProcessVariable::getBoundaryConditions(
```
endJunction: Better to name it `create...` because it of what it does.

ProcessLib/BoundaryCondition/UniformDirichletBoundaryCondition.cpp
chleh: ConfigTree is rather fragile (that's why there are so many test cases for it): All of its methods have to interact nicely. So when deriving from ConfigTree one probably would have to make most methods of ConfigTree `final` in order to not accidentally break things or accidentally forget some checks in the derived implementation.
Furthermore derivation is not necessarily good design, and in this special case BCConfig comprises a ConfigTree, a geometry and a component index, which are three entirely distinct things, so I'd favour composition anyway.

ProcessLib/BoundaryCondition/GenericNaturalBoundaryCondition-impl.h
```diff
@@ -0,0 +1,90 @@
...
+        NumLib::LocalToGlobalIndexMap const& dof_table_bulk,
+        int const variable_id, int const component_id,
+        unsigned const global_dim,
+        std::vector<MeshLib::Element*> const& elements, Data&& data)
+    : _data(std::forward<Data>(data)), _integration_order(integration_order)
+{
+    assert(component_id <
+           static_cast<int>(dof_table_bulk.getNumberOfComponents()));
+
+    // deep copy because the neumann bc config destroys the elements.
```
endJunction: Is the reference to "neumann bc config" still correct?

ProcessLib/ProcessVariable.cpp
endJunction: Or even to `createBoundaryConditionsFromConfigs` to be explicit.

ProcessLib/BoundaryCondition/BoundaryConditionCollection.h
```diff
@@ -0,0 +1,46 @@
...
+#ifndef PROCESSLIB_BOUNDARYCONDITIONCOLLECTION_H
+#define PROCESSLIB_BOUNDARYCONDITIONCOLLECTION_H
+
+#include "DirichletBoundaryCondition.h"
+#include "ProcessLib/ProcessVariable.h"
+
+namespace ProcessLib
+{
+
+class BoundaryConditionCollection : public BoundaryCondition
```
endJunction: Needs discussion: I'm not very happy with the statement "A boundary conditions collection is a boundary condition."
There must be an anchor for the recursive definition...

ProcessLib/BoundaryCondition/BoundaryConditionCollection.cpp
```diff
@@ -0,0 +1,57 @@
...
+    for (int variable_id = 0;
+         variable_id < static_cast<int>(process_variables.size());
+         ++variable_id)
+    {
+        ProcessVariable& pv = process_variables[variable_id];
+        auto bcs = pv.getBoundaryConditions(dof_table, variable_id,
+                                            integration_order);
+
+        std::move(bcs.begin(), bcs.end(),
+                  std::back_inserter(_boundary_conditions));
```
endJunction: The move algorithm is cool! Thanks.

ProcessLib/BoundaryCondition/UniformDirichletBoundaryCondition.cpp
```diff
@@ -11,70 +11,57 @@
...
     bc.values.reserve(bc.values.size() + ids.size());
-    for (auto& id : ids)
-    {
+    for (auto& id : ids) {
         MeshLib::Location l(searcher.getMeshId(), MeshLib::MeshItemType::Node,
                             id);
         // TODO: that might be slow, but only done once
         const auto g_idx =
-            dof_table.getGlobalIndex(l, variable_id, component_id);
+            dof_table.getGlobalIndex(l, variable_id, config.component_id);
```
endJunction: For discussion: It seems to me that the component_id is misplaced, and should go together with the variable_id.

endJunction:
Few small things to correct and the others are for design discussion anyway.
:fast_forward:

ProcessLib/BoundaryCondition/UniformDirichletBoundaryCondition.cpp
chleh: Looking at this issue from the create methods side: var_id comes from the process, which only knows about the order of its ProcessVariables, but not, which component gets which BC; component_id on the other hand comes directly from the BC config.

TomFischer:
Discussion idea: Move up the element/node searchers.... 12:15

ProcessLib/ProcessVariable.cpp
chleh: I think the first one is sufficient. The caller doesn't need to know that there are underlying configs.

ProcessLib/BoundaryCondition/BoundaryConditionCollection.h
chleh: I could remove the inheritance, since this class is not meant to be used polymorphically.
The inheritance is there to stress that a BC collection in the end is only a BC (a special one, which happens to be a collection of several individual BCs), so I'd like to keep it.

ProcessLib/BoundaryCondition/GenericNaturalBoundaryCondition-impl.h
chleh: Moved upwards along with the searcher.

chleh:
The last commit moves the mesh element/node searchers.

ProcessLib/BoundaryCondition/BoundaryConditionCollection.h
chleh: Removed the inheritance.

endJunction:
Looks good! Not sure if someone else will review this; rebase and merge in the afternoon otherwise.

wenqing:
:+1:

PR #1340, opened 2016-08-08
Mechanics Process

endJunction:
... including linear and Burger's materials.

Todo:
- [ ] Resolve circular deps SolidModels <-> ProcessLib; There is header file dependency because in SolidModels the ProcessLib::Parameters are requiered and the SolidModels are required by the Process itself. There are several ideas how to resolve this type of issues. To be discussed ASAP.

- [ ] Documentation of benchmarks
- [x] Code documentation, especially:
 - [x] Interface of the `SolidsBase`
- [x] Update Traction and Displacement BCs (use Neumann and Dirichlet instead)
- [x] Fix compilation issues.
- [x] MathLib unit tests.
- [x] Resolve Eigen/MSVC compilation issue.

### DIFF
--- a/Applications/ApplicationsLib/ProjectData.cpp
+++ b/Applications/ApplicationsLib/ProjectData.cpp
@@ -32,6 +32,7 @@
 #include "ProcessLib/UncoupledProcessesTimeLoop.h"
 
 #include "ProcessLib/GroundwaterFlow/CreateGroundwaterFlowProcess.h"
+#include "ProcessLib/SmallDeformation/CreateSmallDeformationProcess.h"
 #include "ProcessLib/TES/CreateTESProcess.h"
 #include "ProcessLib/HeatConduction/CreateHeatConductionProcess.h"
 
@@ -268,6 +269,28 @@ void ProjectData::parseProcesses(BaseLib::ConfigTree const& processes_config)
         {
             process = ProcessLib::HeatConduction::createHeatConductionProcess(
                 *_mesh_vec[0], _process_variables, _parameters, process_config);
+        }
+        else if (type == "SMALL_DEFORMATION")
+        {
+            switch (process_config.getConfigParameter<int>("dimension"))
+            {
+                case 2:
+                    process = ProcessLib::SmallDeformation::
+                        createSmallDeformationProcess<2>(
+                            *_mesh_vec[0], _process_variables, _parameters,
+                            process_config);
+                    break;
+                case 3:
+                    process = ProcessLib::SmallDeformation::
+                        createSmallDeformationProcess<3>(
+                            *_mesh_vec[0], _process_variables, _parameters,
+                            process_config);
+                    break;
+                default:
+                    OGS_FATAL(
+                        "SMALL_DEFORMATION process does not support "
+                        "given dimension");
+            }
         }
         else
         {
--- a/Applications/CLI/Tests.cmake
+++ b/Applications/CLI/Tests.cmake
@@ -243,6 +243,52 @@ if(NOT OGS_USE_MPI)
          temperature_analytical.vtu line_60_heat_pcs_0_ts_405_t_31640625.000000.vtu Temperature_Analytical_1year temperature
     )
 
+    # Mechanics; Small deformations, linear (SDL)
+    AddTest(
+        NAME Mechanics_SDL_square_1e0_displacementBC
+        PATH Mechanics/Linear
+        EXECUTABLE ogs
+        EXECUTABLE_ARGS square_1e0.prj
+        WRAPPER time
+        TESTER vtkdiff
+        ABSTOL 1e-16 RELTOL 1e-16
+        DIFF_DATA
+        square_1e0_expected_pcs_0_ts_4_t_1.000000.vtu square_1e0_pcs_0_ts_4_t_1.000000.vtu displacement displacement
+    )
+    AddTest(
+        NAME Mechanics_SDL_square_1e2_tractionBC
+        PATH Mechanics/Linear
+        EXECUTABLE ogs
+        EXECUTABLE_ARGS square_1e2.prj
+        WRAPPER time
+        TESTER vtkdiff
+        ABSTOL 1e-16 RELTOL 1e-16
+        DIFF_DATA
+        square_1e2_expected_pcs_0_ts_4_t_1.000000.vtu square_1e2_pcs_0_ts_4_t_1.000000.vtu displacement displacement
+    )
+    AddTest(
+        NAME LARGE_Mechanics_SDL_disc_with_hole
+        PATH Mechanics/Linear
+        EXECUTABLE ogs
+        EXECUTABLE_ARGS disc_with_hole.prj
+        WRAPPER time
+        TESTER vtkdiff
+        ABSTOL 1e-16 RELTOL 1e-16
+        DIFF_DATA
+        disc_with_hole_expected_pcs_0_ts_4_t_1.000000.vtu disc_with_hole_pcs_0_ts_4_t_1.000000.vtu displacement displacement
+    )
+    AddTest(
+        NAME LARGE_Mechanics_SDL_square_1e5_tractionBC
+        PATH Mechanics/Linear
+        EXECUTABLE ogs
+        EXECUTABLE_ARGS square_1e5.prj
+        WRAPPER time
+        TESTER vtkdiff
+        ABSTOL 1e-16 RELTOL 1e-16
+        DIFF_DATA
+        square_1e5_expected_pcs_0_ts_4_t_1.000000.vtu square_1e5_pcs_0_ts_4_t_1.000000.vtu displacement displacement
+    )
+
 else()
     # MPI groundwater flow tests
     AddTest(
--- a/Documentation/ProjectFile/process/SMALL_DEFORMATION/c_SMALL_DEFORMATION.md
+++ b/Documentation/ProjectFile/process/SMALL_DEFORMATION/c_SMALL_DEFORMATION.md
@@ -1,0 +1,1 @@
+Deformation process with linear kinematics.
--- a/Documentation/ProjectFile/process/SMALL_DEFORMATION/constitutive_relation/LinearElasticIsotropic/t_poissons_ratio.md
+++ b/Documentation/ProjectFile/process/SMALL_DEFORMATION/constitutive_relation/LinearElasticIsotropic/t_poissons_ratio.md
@@ -1,0 +1,1 @@
+Poisson's ratio of a linear elastic material.
--- a/MaterialLib/SolidModels/CreateLinearElasticIsotropic.h
+++ b/MaterialLib/SolidModels/CreateLinearElasticIsotropic.h
@@ -1,0 +1,49 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef MATERIALLIB_SOLIDMODELS_CREATELINEARELASTICISOTROPIC_H_
+#define MATERIALLIB_SOLIDMODELS_CREATELINEARELASTICISOTROPIC_H_
+
+#include "ProcessLib/Utils/ProcessUtils.h"  // required for findParameter
+#include "LinearElasticIsotropic.h"
+
+namespace Solids
+{
+
+template <int DisplacementDim>
+std::unique_ptr<LinearElasticIsotropic<DisplacementDim>>
+createLinearElasticIsotropic(
+    std::vector<std::unique_ptr<ProcessLib::ParameterBase>> const& parameters,
+    BaseLib::ConfigTree const& config)
+{
+    config.checkConfigParameter("type", "LinearElasticIsotropic");
+    DBUG("Create LinearElasticIsotropic material");
+
+    // Youngs modulus
+    auto& youngs_modulus = ProcessLib::findParameter<double>(
+        config, "youngs_modulus", parameters, 1);
+
+    DBUG("Use '%s' as youngs_modulus parameter.", youngs_modulus.name.c_str());
+
+    // Poissons ratio
+    auto& poissons_ratio = ProcessLib::findParameter<double>(
+        config, "poissons_ratio", parameters, 1);
+
+    DBUG("Use '%s' as poissons_ratio parameter.", poissons_ratio.name.c_str());
+
+    typename LinearElasticIsotropic<DisplacementDim>::MaterialProperties mp{
+        youngs_modulus, poissons_ratio};
+
+    return std::unique_ptr<LinearElasticIsotropic<DisplacementDim>>{
+        new LinearElasticIsotropic<DisplacementDim>{mp}};
+}
+
+}  // namespace Solids
+
+#endif  // MATERIALLIB_SOLIDMODELS_CREATELINEARELASTICISOTROPIC_H_
--- a/MaterialLib/SolidModels/LinearElasticIsotropic.h
+++ b/MaterialLib/SolidModels/LinearElasticIsotropic.h
@@ -1,0 +1,112 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef MATERIALLIB_SOLIDMODELS_LINEARELASTICISOTROPIC_H_
+#define MATERIALLIB_SOLIDMODELS_LINEARELASTICISOTROPIC_H_
+
+#include "MechanicsBase.h"
+
+namespace Solids
+{
+template <int DisplacementDim>
+class LinearElasticIsotropic final : public MechanicsBase<DisplacementDim>
+{
+public:
+    /// Variables specific to the material model
+    class MaterialProperties
+    {
+        using P = ProcessLib::Parameter<double>;
+        using X = ProcessLib::SpatialPosition;
+
+    public:
+        MaterialProperties(P const& youngs_modulus, P const& poissons_ratio)
+            : _youngs_modulus(youngs_modulus), _poissons_ratio(poissons_ratio)
+        {
+        }
+
+        /// Lamé's first parameter.
+        double lambda(double const t, X const& x) const
+        {
+            return _youngs_modulus(t, x)[0] * _poissons_ratio(t, x)[0] /
+                   (1 + _poissons_ratio(t, x)[0]) /
+                   (1 - 2 * _poissons_ratio(t, x)[0]);
+        }
+
+        /// Lamé's second parameter, the shear modulus.
+        double mu(double const t, X const& x) const
+        {
+            return _youngs_modulus(t, x)[0] /
+                   (2 * (1 + _poissons_ratio(t, x)[0]));
+        }
+
+    private:
+        P const& _youngs_modulus;
+        P const& _poissons_ratio;
+    };
+
+    struct MaterialStateVariables
+        : public MechanicsBase<DisplacementDim>::MaterialStateVariables
+    {
+        void pushBackState() {}
+    };
+
+    std::unique_ptr<
+        typename MechanicsBase<DisplacementDim>::MaterialStateVariables>
+    createMaterialStateVariables() override
+    {
+        return std::unique_ptr<
+            typename MechanicsBase<DisplacementDim>::MaterialStateVariables>{
+            new MaterialStateVariables};
+    }
+
+public:
+    static int const KelvinVectorSize =
+        ProcessLib::KelvinVectorDimensions<DisplacementDim>::value;
+    using KelvinVector = ProcessLib::KelvinVectorType<DisplacementDim>;
+    using KelvinMatrix = ProcessLib::KelvinMatrixType<DisplacementDim>;
+
+    explicit LinearElasticIsotropic(
+        MaterialProperties const& material_properties)
+        : _mp(material_properties)
+    {
+    }
+
+    void computeConstitutiveRelation(
+        double const t,
+        ProcessLib::SpatialPosition const& x,
+        double const /*dt*/,
+        KelvinVector const& eps_prev,
+        KelvinVector const& eps,
+        KelvinVector const& sigma_prev,
+        KelvinVector& sigma,
+        KelvinMatrix& C,
+        typename MechanicsBase<DisplacementDim>::MaterialStateVariables&
+        /*material_state_variables*/) override
+    {
+        C.setZero();
+
+        C.template topLeftCorner<3, 3>().setConstant(_mp.lambda(t, x));
+        C.noalias() += 2 * _mp.mu(t, x) * KelvinMatrix::Identity();
+
+        sigma.noalias() = sigma_prev + C * (eps - eps_prev);
+    }
+
+private:
+    MaterialProperties _mp;
+};
+
+}  // namespace Solids
+
+namespace Solids
+{
+extern template class LinearElasticIsotropic<2>;
+extern template class LinearElasticIsotropic<3>;
+}  // namespace Solids
+
+#endif  // MATERIALLIB_SOLIDMODELS_LINEARELASTICISOTROPIC_H_
--- a/MaterialLib/SolidModels/MechanicsBase.h
+++ b/MaterialLib/SolidModels/MechanicsBase.h
@@ -22,6 +22,11 @@ class Element;
 
 namespace Solids
 {
+/// Interface for mechanical solid material models. Provides updates of the
+/// stress for a given current state and also a tangent at that position. If the
+/// implemented material model stores an internal state, the nested
+/// MaterialStateVariables class should be used; it's only responsibility is to
+/// provide state's push back possibility.
 template <int DisplacementDim>
 struct MechanicsBase
 {
@@ -36,9 +41,16 @@ struct MechanicsBase
         virtual void pushBackState() = 0;
     };
 
+    /// Polymorphic creator for MaterialStateVariables objects specific for a
+    /// material model.
+    virtual std::unique_ptr<MaterialStateVariables>
+    createMaterialStateVariables() = 0;
+
     using KelvinVector = ProcessLib::KelvinVectorType<DisplacementDim>;
     using KelvinMatrix = ProcessLib::KelvinMatrixType<DisplacementDim>;
 
+    /// Dynamic size Kelvin vector and matrix wrapper for the polymorphic
+    /// constitutive relation compute function.
     void computeConstitutiveRelation(
         double const t,
         ProcessLib::SpatialPosition const& x,
@@ -76,6 +88,10 @@ struct MechanicsBase
         C = C_;
     }
 
+    /// Computation of the constitutive relation for specific material model.
+    /// This should be implemented in the derived model. Fixed Kelvin vector and
+    /// matrix size version; for dynamic size arguments there is an overloaded
+    /// wrapper function.
     virtual void computeConstitutiveRelation(
         double const t,
         ProcessLib::SpatialPosition const& x,
@@ -86,9 +102,6 @@ struct MechanicsBase
         KelvinVector& sigma,
         KelvinMatrix& C,
         MaterialStateVariables& material_state_variables) = 0;
-
-    virtual std::unique_ptr<MaterialStateVariables>
-    createMaterialStateVariables() = 0;
 
     virtual ~MechanicsBase() = default;
 };
--- a/MaterialLib/SolidModels/MechanicsBase.h
+++ b/MaterialLib/SolidModels/MechanicsBase.h
@@ -1,0 +1,97 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef MATERIALLIB_SOLIDMODELS_MECHANICSBASE_H_
+#define MATERIALLIB_SOLIDMODELS_MECHANICSBASE_H_
+
+#include <memory>
+
+#include "ProcessLib/Deformation/BMatrixPolicy.h"
+#include "ProcessLib/Parameter/Parameter.h"
+
+namespace MeshLib
+{
+class Element;
+}
+
+namespace Solids
+{
+template <int DisplacementDim>
+struct MechanicsBase
+{
+    /// The MaterialStateVariables may store material model specific state
+    /// (other than sigma and eps), which are usually material history
+    /// dependent. The objects are stored by the user (usually in assembly per
+    /// integration point) and are created via \ref
+    /// createMaterialStateVariables().
+    struct MaterialStateVariables
+    {
+        virtual ~MaterialStateVariables() = default;
+        virtual void pushBackState() = 0;
+    };
+
+    using KelvinVector = ProcessLib::KelvinVectorType<DisplacementDim>;
+    using KelvinMatrix = ProcessLib::KelvinMatrixType<DisplacementDim>;
+
+    void computeConstitutiveRelation(
+        double const t,
+        ProcessLib::SpatialPosition const& x,
+        double const dt,
+        Eigen::Matrix<double, Eigen::Dynamic, 1> const& eps_prev,
+        Eigen::Matrix<double, Eigen::Dynamic, 1> const& eps,
+        Eigen::Matrix<double, Eigen::Dynamic, 1> const& sigma_prev,
+        Eigen::Matrix<double, Eigen::Dynamic, 1>& sigma,
+        Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>&
+            C,
+        MaterialStateVariables& material_state_variables)
+    {
+        // TODO Avoid copies of data:
+        // Using MatrixBase<Derived> not possible because template functions
+        // cannot be virtual. Maybe there is a workaround for this.  Using
+        // Map<Matrix<double, ...>> makes the interface (for the material model
+        // implementation) unnecessary difficult.
+        KelvinVector const eps_prev_{eps_prev};
+        KelvinVector const eps_{eps};
+        KelvinVector const sigma_prev_{sigma_prev};
+        KelvinVector sigma_{sigma};
+        KelvinMatrix C_{C};
+
+        computeConstitutiveRelation(t,
+                                    x,
+                                    dt,
+                                    eps_prev_,
+                                    eps_,
+                                    sigma_prev_,
+                                    sigma_,
+                                    C_,
+                                    material_state_variables);
+
+        sigma = sigma_;
+        C = C_;
+    }
+
+    virtual void computeConstitutiveRelation(
+        double const t,
+        ProcessLib::SpatialPosition const& x,
+        double const dt,
+        KelvinVector const& eps_prev,
+        KelvinVector const& eps,
+        KelvinVector const& sigma_prev,
+        KelvinVector& sigma,
+        KelvinMatrix& C,
+        MaterialStateVariables& material_state_variables) = 0;
+
+    virtual std::unique_ptr<MaterialStateVariables>
+    createMaterialStateVariables() = 0;
+
+    virtual ~MechanicsBase() = default;
+};
+}
+
+#endif  // MATERIALLIB_SOLIDMODELS_MECHANICSBASE_H_
--- a/ProcessLib/SmallDeformation/CreateLocalAssemblers.h
+++ b/ProcessLib/SmallDeformation/CreateLocalAssemblers.h
@@ -1,0 +1,109 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ */
+
+#ifndef PROCESSLIB_SMALLDEFORMATION_CREATE_LOCAL_ASSEMBLERS_H_
+#define PROCESSLIB_SMALLDEFORMATION_CREATE_LOCAL_ASSEMBLERS_H_
+
+#include <vector>
+
+#include <logog/include/logog.hpp>
+
+#include "NumLib/DOF/LocalToGlobalIndexMap.h"
+
+#include "LocalDataInitializer.h"
+
+namespace ProcessLib
+{
+namespace SmallDeformation
+{
+namespace detail
+{
+template <unsigned GlobalDim, int DisplacementDim,
+          template <typename, typename, unsigned, int>
+          class LocalAssemblerImplementation,
+          typename LocalAssemblerInterface, typename... ExtraCtorArgs>
+void createLocalAssemblers(
+    NumLib::LocalToGlobalIndexMap const& dof_table,
+    std::vector<MeshLib::Element*> const& mesh_elements,
+    unsigned const integration_order,
+    std::vector<std::unique_ptr<LocalAssemblerInterface>>& local_assemblers,
+    ExtraCtorArgs&&... extra_ctor_args)
+{
+    // Shape matrices initializer
+    using LocalDataInitializer =
+        LocalDataInitializer<LocalAssemblerInterface,
+                             LocalAssemblerImplementation, GlobalDim,
+                             DisplacementDim, ExtraCtorArgs...>;
+
+    DBUG("Create local assemblers.");
+    // Populate the vector of local assemblers.
+    local_assemblers.resize(mesh_elements.size());
+
+    LocalDataInitializer initializer(dof_table);
+
+    DBUG("Calling local assembler builder for all mesh elements.");
+    GlobalExecutor::transformDereferenced(
+        initializer,
+        mesh_elements,
+        local_assemblers,
+        integration_order,
+        std::forward<ExtraCtorArgs>(extra_ctor_args)...);
+}
+
+}  // namespace detail
+
+/*! Creates local assemblers for each element of the given \c mesh.
+ *
+ * \tparam LocalAssemblerImplementation the individual local assembler type
+ * \tparam LocalAssemblerInterface the general local assembler interface
+ * \tparam ExtraCtorArgs types of additional constructor arguments.
+ *         Those arguments will be passed to the constructor of
+ *         \c LocalAssemblerImplementation.
+ *
+ * The first two template parameters cannot be deduced from the arguments.
+ * Therefore they always have to be provided manually.
+ */
+template <int DisplacementDim,
+          template <typename, typename, unsigned, int>
+          class LocalAssemblerImplementation,
+          typename LocalAssemblerInterface, typename... ExtraCtorArgs>
+void createLocalAssemblers(
+    const unsigned dimension,
+    std::vector<MeshLib::Element*> const& mesh_elements,
+    NumLib::LocalToGlobalIndexMap const& dof_table,
+    unsigned const integration_order,
+    std::vector<std::unique_ptr<LocalAssemblerInterface>>& local_assemblers,
+    ExtraCtorArgs&&... extra_ctor_args)
+{
+    DBUG("Create local assemblers.");
+
+    switch (dimension)
+    {
+        case 2:
+            detail::createLocalAssemblers<2, DisplacementDim,
+                                          LocalAssemblerImplementation>(
+                dof_table, mesh_elements, integration_order, local_assemblers,
+                std::forward<ExtraCtorArgs>(extra_ctor_args)...);
+            break;
+        case 3:
+            detail::createLocalAssemblers<3, DisplacementDim,
+                                          LocalAssemblerImplementation>(
+                dof_table, mesh_elements, integration_order, local_assemblers,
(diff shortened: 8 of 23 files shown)


chleh:
4k lines of changes. Can't you split that into smaller pieces? :smile:

endJunction:
@chleh Down to 3k and will be less with new bcs. I might be splitting the MathLib/KelvinVector things while preparing this to its final glance.

endJunction:
Jenkins, test this please!

endJunction:
FYI
There is a problem with the combination of VisualStudio12 2013 ( compiler identification is MSVC 18.0.31101.0) and Eigen-3.2.5 resulting in compilation error:
```
c:\libs\msvc2013-x64\include\eigen3\eigen\src/Core/DenseBase.h(505): error C2280: 'std::unique_ptr<Solids::SolidsBase<1>::InternalVariables,std::default_delete<_Ty>>::unique_ptr(const std::unique_ptr<_Ty,std::default_delete<_Ty>> &)' : attempting to reference a deleted function [F:\jenkins\workspace\OGS-6\ufz\PR-1340\build\ProcessLib\ProcessLib.vcxproj]
          with
          [
              _Ty=Solids::SolidsBase<1>::InternalVariables
          ] (F:\jenkins\workspace\OGS-6\ufz\PR-1340\ogs\ProcessLib\SmallDeformation\SmallDeformationProcess.cpp)
          C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\include\memory(1486) : see declaration of 'std::unique_ptr<Solids::SolidsBase<1>::InternalVariables,std::default_delete<_Ty>>::unique_ptr'
          with
          [
              _Ty=Solids::SolidsBase<1>::InternalVariables
          ]
          This diagnostic occurred in the compiler generated function 'IntegrationPointData<EigenDynamicBMatrixPolicy<ShapeFunction,1>,1>::IntegrationPointData(const IntegrationPointData<EigenDynamicBMatrixPolicy<ShapeFunction,1>,1> &)'
          with
          [
              ShapeFunction=NumLib::ShapeLine2
          ]
```

This is a [fixed bug](http://eigen.tuxfamily.org/bz/show_bug.cgi?id=920) in Eigen, though I'm not sure in which version it is included.

Applications/ApplicationsLib/ProjectData.cpp
```diff
@@ -186,6 +187,33 @@ void ProjectData::buildProcesses()
                     *_mesh_vec[0], *nl_slv, std::move(time_disc),
                     std::move(conv_crit), _process_variables, _parameters, pc));
         }
+        else if (type == "SMALL_DEFORMATION")
+        {
+            switch (pc.getConfigParameter<int>("dimension"))
```
norihiro-w: isn't it possible to guess the dimension from a mesh?

MaterialLib/SolidModels/SolidsBase.h
```diff
@@ -0,0 +1,87 @@
...
+
+namespace MeshLib
+{
+class Element;
+}
+
+namespace Solids
+{
+template <int DisplacementDim>
+struct SolidsBase
```
norihiro-w: are you going to add other solid properties, e.g. density, thermal properties, here?

MaterialLib/SolidModels/SolidsBase.h
```diff
@@ -0,0 +1,87 @@
...
+{
+    struct InternalVariables
```
norihiro-w: can you document this struct?

MaterialLib/SolidModels/SolidsBase.h
```diff
@@ -0,0 +1,87 @@
...
+    // Using Map<Matrix<double, ...>> makes the interface (for the material
+    // model implementation) unnecessary difficult.
+    void computeConstitutiveRelation(
+        double const t,
+        ProcessLib::SpatialPosition const& x,
+        double const dt,
+        Eigen::Matrix<double, -1, 1> const& eps_prev,
+        Eigen::Matrix<double, -1, 1> const& eps,
+        Eigen::Matrix<double, -1, 1> const& sigma_prev,
+        Eigen::Matrix<double, -1, 1>& sigma,
```
norihiro-w: Why does this function need to return stress?

MaterialLib/SolidModels/SolidsBase.h
```diff
@@ -0,0 +1,87 @@
...
+
+    using KelvinVector = ProcessLib::KelvinVectorType<DisplacementDim>;
+    using KelvinMatrix = ProcessLib::KelvinMatrixType<DisplacementDim>;
+
+    // TODO Avoid copies of data:
+    // Using MatrixBase<Derived> not possible because template functions cannot
+    // be virtual. Maybe there is a workaround for this.
+    // Using Map<Matrix<double, ...>> makes the interface (for the material
+    // model implementation) unnecessary difficult.
+    void computeConstitutiveRelation(
```
norihiro-w: how are you thinking to pass other variables such as temperature? i mean the future extension

Applications/ApplicationsLib/ProjectData.cpp
endJunction: In general all possibilities can occur, like 2D displacement in 3D, or 2D movement of line elements, but the non-matching dimensions needs special models.
For now, indeed the displacement and the mesh dimensions are equal.
@nagelt Can you comment on whether we need that flexibility?

MaterialLib/SolidModels/SolidsBase.h
endJunction: sigma is also update together with C. See [linear elastic](https://github.com/endJunction/ogs/blob/M5_full/MaterialLib/SolidModels/LinearElasticIsotropic-impl.h#L35) and the [Burgers](https://github.com/endJunction/ogs/blob/M5_full/MaterialLib/SolidModels/ViscoElasticBurgers-impl.h#L178) material implementations.
The Burgers material model will be pull-requested as follow up.

MaterialLib/SolidModels/SolidsBase.h
endJunction: I'll add a comment along these lines:
"InternalVariables store material model specific state (other than sigma and eps), which are usually material history dependent."

The  class name comes, because this state is internal to the material model, but now this being visible in the interface, maybe MaterialState, or MaterialInternalState, or InternalState would be a better name.

MaterialLib/SolidModels/SolidsBase.h
endJunction: For sharing of data with Process there is MaterialProperties class (e.g. for [linear elastic](https://github.com/endJunction/ogs/blob/M5_full/MaterialLib/SolidModels/LinearElasticIsotropic.h#L29)) where the Youngs modulus and the Poisson ratio are stored. In case of [Burgers model](https://github.com/endJunction/ogs/blob/M5_full/MaterialLib/SolidModels/LinearElasticIsotropic.h#L29) There are many more parameters.
The MaterialProperties are created from the config, where one can refer to other's processes `Parameter`s.

MaterialLib/SolidModels/SolidsBase.h
endJunction: Have not thought too much about this so far. But yes, it seems a proper place for having a solid material model which is temperature dependent. For solid properties only. The interface should remain as it is in general.
Maybe I'm missing something---which should be discussed then.

endJunction: Well, how is it actually modeled, when rubber is stretched it is warming up and when it is released then cooling down? Where does the heat contribution from mechanical stress go to?

norihiro-w: i don't get what you want to say in the last questions. anyway, if the class is used only for mechanics, I suggest making the namespace more specific, e.g. `SolidMechanics`.

MaterialLib/SolidModels/SolidsBase.h
norihiro-w: i'm asking why the function named `computeConstitutiveRelation()` should also return stress. why you cannot return only a constitutive matrix? what is an idea behind?

MaterialLib/SolidModels/SolidsBase.h
norihiro-w: then things are hidden in Parameter and not obvious from the class interface. not nice

chleh:
Maybe it would be good to postpone the discussions about "extended" material models to end of this year, when HM, HT and density driven flow processes are ready and we see more of the overall picture?

MaterialLib/SolidModels/SolidsBase.h
endJunction: If there are no objections, I rename this file/namespace to MechanicsBase.h such that we have
MaterialLib/SolidModels/MechanicsBase.h w/o repititions.

Applications/ApplicationsLib/ProjectData.cpp
nagelt: My current understanding is that we would only need that for structural elements (plates, shells, membranes, etc.). These elements require special element formulations resting on physical input, i.e. postulated deformation modes, that would have to be implemented separately. As we're dealing only with 3D continuum and 2D plane strain continuum elements in formulations, where the deformation takes place only in the mesh domain (e.g. in-plane), I'd say that indeed we can derive the dimension from the mesh.

Applications/ApplicationsLib/ProjectData.cpp
```diff
@@ -186,6 +187,33 @@ void ProjectData::buildProcesses()
...
+                            createSmallDeformationProcess<3>(
+                                *_mesh_vec[0], *nl_slv, std::move(time_disc),
+                                std::move(conv_crit), _process_variables,
+                                _parameters, pc));
+                    break;
+                default:
+                    ERR(
+                        "SMALL_DEFORMATION process does not support "
+                        "given dimension");
+                    std::abort();
```
chleh: OGS_FATAL.

MaterialLib/SolidModels/SolidsBase.h
nagelt: No objections. Rubber, if considered elastic, will cool if you stretch it, and heat when you compress it. If it is modelled as viscoelastic (usually more appropriate), it will heat due to dissipative effects regardless of the deformation mode.

MaterialLib/SolidModels/SolidsBase.h
nagelt: I suggest either MaterialInternalVariables or MaterialStateVariables.

Applications/CLI/Tests.cmake
```diff
@@ -217,6 +217,53 @@ if(NOT OGS_USE_MPI)
...
+        EXECUTABLE ogs
+        EXECUTABLE_ARGS disc_with_hole.prj
+        WRAPPER time
+        TESTER vtkdiff
+        ABSTOL 1e-16 RELTOL 1e-16
+        DIFF_DATA
+        disc_with_hole_expected_pcs_0_ts_4_t_1.000000.vtu disc_with_hole_pcs_0_ts_4_t_1.000000.vtu displacement displacement
+    )
+    AddTest(
+        NAME Mechanics_SDL_square_1e0_displacementBC
```
chleh: Just wondering: You really want to test a one element test case?

Documentation/ProjectFile/process/SMALL_DEFORMATION/c_SMALL_DEFORMATION.md
```diff
@@ -0,0 +1 @@
+Deformation process with linear kinematics.
```
chleh: dynamics rather than kinematics? (since forces are involved.)

Documentation/ProjectFile/process/SMALL_DEFORMATION/constitutive_relation/LinearElasticIsotropic/t_poissons_ratio.md
```diff
@@ -0,0 +1 @@
+Poissons ratio of a linear elastic material.
```
chleh: "Poisson's", and "Young's" below.

Documentation/ProjectFile/process/SMALL_DEFORMATION/constitutive_relation/ViscoElasticBurgers/c_ViscoElasticBurgers.md
```diff
@@ -0,0 +1,3 @@
+Visco elastic material model based on Burgers' rheological model. Viscosities
+and Kelvin shear modulus are stress dependent according to LUBBY2
+model. Heusermann 1983.
```
chleh: Maybe \cite?

MaterialLib/SolidModels/SolidsBase.h
nagelt: The idea was this: Our constitutive relation is given as a differential-algebraic equation system. Differential for evolution equations, algebraic for constraints and absolute stress formulations. The function computeConstitutiveRelation() integrates this DAE (the constitutive relation) from one time step to the next. Hence the name. The result is the stress (and the updated internal/state variables). As a by-product from this integration, we obtain also C (algorithmically consistent). Stress is needed for the residual, C for the Jacobian.

Documentation/ProjectFile/process/SMALL_DEFORMATION/constitutive_relation/ViscoElasticBurgers/t_kelvin_shear_modulus.md
```diff
@@ -0,0 +1 @@
+Initial shear modulus of the Kelvin element in the Burgers' model.
```
chleh: I think it should be either "in the Burgers model" or "in Burgers' model".

MaterialLib/SolidModels/LinearElasticIsotropic.h
```diff
@@ -0,0 +1,104 @@
...
+{
+template <int DisplacementDim>
+class LinearElasticIsotropic final : public SolidsBase<DisplacementDim>
+{
+public:
+    /// Variables specific to the material model
+    class MaterialProperties
+    {
+        using P = ProcessLib::Parameter<double>;
+        using X = ProcessLib::SpatialPosition;
```
chleh: Minor: I'd prefer to omit these aliases.

MaterialLib/SolidModels/LinearElasticIsotropic.h
```diff
@@ -0,0 +1,104 @@
...
+    };
+
+    std::unique_ptr<typename SolidsBase<DisplacementDim>::InternalVariables>
+    createInternalVariables() override
+    {
+        return std::unique_ptr<
+            typename SolidsBase<DisplacementDim>::InternalVariables>{
+            static_cast<
+                typename SolidsBase<DisplacementDim>::InternalVariables*>(
+                new InternalVariables)};
```
chleh: Can't you just `return std::unique_ptr<InternalVariables>(new InternalVariables);`, the unique_ptr does the upcasting automatically.

MaterialLib/SolidModels/LinearElasticIsotropic-impl.h
```diff
@@ -0,0 +1,40 @@
...
+    typename SolidsBase<
+        DisplacementDim>::InternalVariables& /*internal_variables*/)
+{
+    C.setZero();
+
+    C.template topLeftCorner<3, 3>().setConstant(_mp.lambda(t, x));
+    C.noalias() += 2 * _mp.mu(t, x) * KelvinMatrix::Identity();
+
+    sigma.noalias() = sigma_prev + C * (eps - eps_prev);
+}
```
chleh: Minor: You could just inline this one method, too.

Documentation/ProjectFile/process/SMALL_DEFORMATION/c_SMALL_DEFORMATION.md
nagelt: linear kinematics implies small deformations.

Documentation/ProjectFile/process/SMALL_DEFORMATION/constitutive_relation/ViscoElasticBurgers/t_kelvin_shear_modulus.md
nagelt: the former

MaterialLib/SolidModels/SolidsBase.h
```diff
@@ -0,0 +1,87 @@
...
+        double const t,
+        ProcessLib::SpatialPosition const& x,
+        double const dt,
+        Eigen::Matrix<double, -1, 1> const& eps_prev,
```
chleh: I'd prefer Eigen::Dynamic instead of -1.

MaterialLib/SolidModels/SolidsBase.h
```diff
@@ -0,0 +1,87 @@
...
+{
+    struct InternalVariables
+    {
+        virtual void pushBackState() = 0;
+    };
+
+    using KelvinVector = ProcessLib::KelvinVectorType<DisplacementDim>;
+    using KelvinMatrix = ProcessLib::KelvinMatrixType<DisplacementDim>;
+
+    // TODO Avoid copies of data:
```
chleh: Where is data copied? On the caller side?

chleh: Found the copy, maybe you could move the comment down to the place of copying.

chleh: Btw. this method is not virtual.

MaterialLib/SolidModels/createLinearElasticIsotropic.h
```diff
@@ -0,0 +1,51 @@
...
+    BaseLib::ConfigTree const& config)
+{
+    config.checkConfigParameter("type", "LinearElasticIsotropic");
+    DBUG("Create LinearElasticIsotropic material");
+
+    // Youngs modulus
+    auto& youngs_modulus = ProcessLib::findParameter<double>(
+        config, "youngs_modulus", parameters, 1);
+
+    DBUG("Use \'%s\' as youngs_modulus parameter.",
```
chleh: no backslashes necessary.

MaterialLib/SolidModels/LinearElasticIsotropic.h
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+public:
+    static int const KelvinVectorSize =
+        ProcessLib::KelvinVectorDimensions<DisplacementDim>::value;
+    using KelvinVector = ProcessLib::KelvinVectorType<DisplacementDim>;
+    using KelvinMatrix = ProcessLib::KelvinMatrixType<DisplacementDim>;
+
+    LinearElasticIsotropic(MaterialProperties& material_properties)
```
chleh: explicit.

MaterialLib/SolidModels/LinearElasticIsotropic.h
```diff
@@ -0,0 +1,104 @@
...
+#include "SolidsBase.h"
+
+namespace Solids
+{
+template <int DisplacementDim>
+class LinearElasticIsotropic final : public SolidsBase<DisplacementDim>
+{
+public:
+    /// Variables specific to the material model
+    class MaterialProperties
```
chleh: Minor: Maybe nested classes shouldn't become the default way of grouping things together.

MaterialLib/SolidModels/createLinearElasticIsotropic.h
```diff
@@ -0,0 +1,51 @@
...
+        config, "poissons_ratio", parameters, 1);
+
+    DBUG("Use \'%s\' as poissons_ratio parameter.",
+         poissons_ratio.name.c_str());
+
+    typename LinearElasticIsotropic<DisplacementDim>::MaterialProperties mp{
+        youngs_modulus, poissons_ratio};
+
+    return std::unique_ptr<LinearElasticIsotropic<DisplacementDim>>{
+        new LinearElasticIsotropic<DisplacementDim>{mp}};
```
chleh: If I see it correctly, the created LinearElasticIsotropic instance will reference the local variable `mp`.

ProcessLib/SmallDeformation/CreateSmallDeformationProcess.h
```diff
@@ -0,0 +1,108 @@
...
+        {//! \ogs_file_param_special{process__SMALL_DEFORMATION__process_variables__process_variable}
+         "process_variable"});
+
+    DBUG("Associate displacement with process variable \'%s\'.",
+         process_variables.back().get().getName().c_str());
+
+    if (process_variables.back().get().getNumberOfComponents() !=
+        DisplacementDim)
+    {
+        ERR("Wrong number of components");
```
chleh: OGS_FATAL.

ProcessLib/SmallDeformation/CreateSmallDeformationProcess.h
```diff
@@ -0,0 +1,108 @@
...
+            type.c_str());
+    }
+
+    SmallDeformationProcessData<DisplacementDim> process_data{
+        std::move(material)};
+
+    SecondaryVariableCollection secondary_variables;
+
+    NumLib::NamedFunctionCaller named_function_caller(
+        {"SmallDeformation_displacement"});
```
chleh: Does that work? I don't expect it, since the displacement is a vectorial quantity.

ProcessLib/SmallDeformation/LocalDataInitializer.h
```diff
@@ -0,0 +1,291 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef PROCESSLIB_SMALLDEFORMATION_LOCALDATAINITIALIVER_H_
```
chleh: Typo: LOCALDATAINITIALIZER

ProcessLib/SmallDeformation/LocalDataInitializer.h
```diff
@@ -0,0 +1,291 @@
...
+/// The LocalDataInitializer is a functor creating a local assembler data with
+/// corresponding to the mesh element type shape functions and calling
+/// initialization of the new local assembler data.
+/// For example for MeshLib::Quad a local assembler data with template argument
+/// NumLib::ShapeQuad4 is created.
+template <
+    typename LocalAssemblerInterface,
+    template <typename, typename, unsigned, unsigned> class LocalAssemblerData,
+    unsigned GlobalDim, unsigned DisplacementDim, typename... ConstructorArgs>
+class LocalDataInitializer final
```
chleh: Can you please remind me why you have to exclude lines and points here? Or is it only for performance reasons now? (Sure, lines and points wouldn't make sense with the current 2D and 3D KelvinVectors, but there was some reason due to the compilation...)

ProcessLib/SmallDeformation/SmallDeformationFEM.h
```diff
@@ -0,0 +1,356 @@
...
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef PROCESS_LIB_SMALLDEFORMATION_FEM_H_
+#define PROCESS_LIB_SMALLDEFORMATION_FEM_H_
+
+#include <fstream>
+#include <memory>
+#include <sstream>
```
chleh: sstream and fstream, really?

ProcessLib/SmallDeformation/SmallDeformationFEM.h
```diff
@@ -0,0 +1,356 @@
...
+          _internal_variables(std::move(other._internal_variables)),
+          _C(std::move(other._C)),
+          _detJ(std::move(other._detJ))
+    {
+    }
+#endif  // _MSC_VER
+
+    using BMatrixType = typename BMatricesType::BMatrixType;
+    using StressVectorType = typename BMatricesType::StressVectorType;
+    using StrainVectorType = typename BMatricesType::StressVectorType;
```
chleh: Minor: You probably could save one typedef.

ProcessLib/SmallDeformation/SmallDeformationFEM.h
```diff
@@ -0,0 +1,356 @@
...
+        _sigma_prev = _sigma;
+        _internal_variables->pushBackState();
+    }
+};
+
+namespace ProcessLib
+{
+namespace SmallDeformation
+{
+enum class IntegrationPointValue
```
chleh: Still needed? At least the extrapolation in general doesn't.

ProcessLib/SmallDeformation/SmallDeformationFEM.h
```diff
@@ -0,0 +1,356 @@
...
+    SmallDeformationLocalAssembler(SmallDeformationLocalAssembler&&) = delete;
+
+    SmallDeformationLocalAssembler(
+        MeshLib::Element const& e,
+        std::size_t const local_matrix_size,
+        unsigned const integration_order,
+        SmallDeformationProcessData<DisplacementDim>& process_data)
+        : _process_data(process_data),
+          _localA(local_matrix_size, local_matrix_size),
+          _localRhs(local_matrix_size),
```
chleh: Just wondering: I thought, you always assemble the residuum and the Jacobian...
But no rename needed. The variables will be gone after the new Jacobian assembly.

ProcessLib/SmallDeformation/SmallDeformationFEM.h
```diff
@@ -0,0 +1,356 @@
...
+
+#include "SmallDeformationProcessData.h"
+
+template <typename BMatricesType, int DisplacementDim>
+struct IntegrationPointData final
+{
+    IntegrationPointData(Solids::SolidsBase<DisplacementDim>& solid_material)
+        : _solid_material(solid_material),
+          _internal_variables(
+              std::move(_solid_material.createInternalVariables()))
```
chleh: is move necessary here?

ProcessLib/SmallDeformation/SmallDeformationFEM.h
```diff
@@ -0,0 +1,356 @@
...
+
+        unsigned const n_integration_points =
+            _integration_method.getNumberOfPoints();
+
+        NodalDisplacementVectorType local_displacement(ShapeFunction::NPOINTS *
+                                                       DisplacementDim);
+
+        for (int i = 0; i < ShapeFunction::NPOINTS * DisplacementDim; ++i)
+        {
+            local_displacement[i] = local_x[i];
```
chleh: Eigen::Map would save the copy.

ProcessLib/SmallDeformation/SmallDeformationFEM.h
```diff
@@ -0,0 +1,356 @@
...
+        NumLib::LocalToGlobalIndexMap::RowColumnIndices const& indices,
+        GlobalMatrix& Jac) override
+    {
+        _localA.setZero();
+
+        NodalDisplacementVectorType local_displacement(ShapeFunction::NPOINTS *
+                                                       DisplacementDim);
+        for (int i = 0; i < ShapeFunction::NPOINTS * DisplacementDim; ++i)
+        {
+            local_displacement(i) = local_x[i];
```
chleh: Unneeded?

ProcessLib/SmallDeformation/SmallDeformationFEM.h
```diff
@@ -0,0 +1,356 @@
...
+            _ip_data[ip]._solid_material.computeConstitutiveRelation(
+                t, x_position, _process_data.dt, eps_prev, eps, sigma_prev,
+                sigma, C, internal_variables);
+
+            _localRhs.noalias() -=
+                B.transpose() * sigma * detJ * wp.getWeight();
+
+            _secondary_data._sigmaXX[ip] = sigma[0];
+            _secondary_data._sigmaYY[ip] = sigma[1];
+            _secondary_data._sigmaXY[ip] = sigma[3];
```
chleh: Note: Since you store sigma in the `_ip_data`, it is not necessary to save them separately only for extrapolation/output.

ProcessLib/SmallDeformation/SmallDeformationProcess.h
```diff
@@ -0,0 +1,176 @@
...
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef PROCESS_LIB_SMALLDEFORMATIONPROCESS_H_
+#define PROCESS_LIB_SMALLDEFORMATIONPROCESS_H_
+
+#include <cassert>
+
+#include "NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.h"
```
chleh: I guess that's not necessary anymore.

ProcessLib/SmallDeformation/SmallDeformationProcess.h
```diff
@@ -0,0 +1,176 @@
...
+                &Base::getTimeDiscretization()) != nullptr)
+        {
+            ERR("The SmallDeformationProcess can not be solved with the "
+                "ForwardEuler time discretization scheme. Aborting");
+            // Because the M matrix is not assembled. Thus, the linearized
+            // system would be singular. The same applies to CrankNicolson with
+            // theta = 0.0, but this case is not checked here.  Anyway, the
+            // SmallDeformationProcess shall be transferred to a simpler
+            // ODESystemTag in the future.
+            std::abort();
```
chleh: OGS_FATAL. Btw: I think that doesn't apply to Newton schemes...

ProcessLib/SmallDeformation/SmallDeformationProcess.h
```diff
@@ -0,0 +1,176 @@
...
+                                                            LocalAssemblerData>(
+            mesh.getDimension(), mesh.getElements(), dof_table,
+            integration_order, _local_assemblers, _process_data);
+
+        // TODO move the two data members somewhere else.
+        // for extrapolation of secondary variables
+        std::vector<std::unique_ptr<MeshLib::MeshSubsets>>
+            all_mesh_subsets_single_component;
+        all_mesh_subsets_single_component.emplace_back(
+            new MeshLib::MeshSubsets(this->_mesh_subset_all_nodes.get()));
```
chleh: `this->` needed?

chleh:
Mostly minor things from my side. Afterwards :+1:.

MaterialLib/SolidModels/SolidsBase.h
norihiro-w: ok

MaterialLib/SolidModels/SolidsBase.h
norihiro-w: maybe I'm missing a full picture... does it make sense to have two separate functions to get stress and C? in newton raphson, we evaluate residual and Jacobian in separate functions. hence we don't need both stress and C at the same time.

endJunction: We cannot pull the computation of sigma and C apart. In the non-linear Burgers model, for example, sigma and C are the result of the same algorithm.
Also in the local assmebler part the residual and Jacobian are evaluated together, especially after the Newton-Raphson PR #1352.

In principle it is possible to separate those computations, but additional storage will be required to store the result of the local Newton iterations.

Applications/CLI/Tests.cmake
endJunction: Why not? It has also some degrees of freedom...

MaterialLib/SolidModels/LinearElasticIsotropic.h
```diff
@@ -0,0 +1,104 @@
...
+        using P = ProcessLib::Parameter<double>;
+        using X = ProcessLib::SpatialPosition;
+
+    public:
+        MaterialProperties(P const& youngs_modulus, P const& poissons_ratio)
+            : _youngs_modulus(youngs_modulus), _poissons_ratio(poissons_ratio)
+        {
+        }
+
+        double lambda(double const t, X const& x) const
```
norihiro-w: does it make sense to pass time? i can't imagine a case lambda needs this.

MaterialLib/SolidModels/LinearElasticIsotropic.h
```diff
@@ -0,0 +1,104 @@
...
+        {
+            return _youngs_modulus(t, x)[0] * _poissons_ratio(t, x)[0] /
```
norihiro-w: remark for future optimisation: if parameter`(t, x)` is expensive, we need to find a way to minimize the number of calling.

MaterialLib/SolidModels/SolidsBase.h
norihiro-w: i see. thanks for the explanation.

MaterialLib/SolidModels/LinearElasticIsotropic.h
endJunction: This just the interface of Parameter class. Would be writing 0 instead of time otherwise, but that is even more questionable.

MaterialLib/SolidModels/LinearElasticIsotropic.h
endJunction: Simple one: One could simply pass lambda and mu as already computed parameters. Young's modulus and Poisson's ratio are now in the config because these are readily available quantities.

Simplifying such expressions of Parameters would require a (JIT) compiler.

MaterialLib/SolidModels/SolidsBase.h
endJunction: Yes. That one cannot be virtual because of the template... But the fixed size one is virtual.
But when the DisplacementDim is removed, all of this might change.

MaterialLib/SolidModels/LinearElasticIsotropic.h
endJunction: For discussion... Can be created from different configs with same result. Are shared as a single quantity between the local assemblers.

ProcessLib/SmallDeformation/LocalDataInitializer.h
endJunction: #pragma once? ;)

MaterialLib/SolidModels/createLinearElasticIsotropic.h
endJunction: Clarified in discussion. Not a temporary, because of copy in LinearElastic...() ctor.

MaterialLib/SolidModels/LinearElasticIsotropic.h
endJunction: OK. The suggestion was to make a **not** nested class LinearElasticIsotropicMaterialProperties (or something with shorter name).

ProcessLib/SmallDeformation/LocalDataInitializer.h
endJunction: B-matrices for points are 0-sized. In shape matrix policy we have a special case. Fixing that was finicky and led to strange construct for Kelvin vector dimensions. Maybe when DisplacementDim is removed, these copies of LocalDataInitializer and creator could be removed.

chleh: Thanks.

Applications/ApplicationsLib/ProjectData.cpp
wenqing: I think, under the mentioned cases, we only need dimension of element and the dimension of space (plates, shells, etc,), which are already kept in mesh object.

MaterialLib/SolidModels/LinearElasticIsotropic.h
wenqing: Document it as 'Lame parameter', or change the function name to lame_parameter.

MaterialLib/SolidModels/LinearElasticIsotropic.h
```diff
@@ -0,0 +1,104 @@
...
+                   (1 + _poissons_ratio(t, x)[0]) /
+                   (1 - 2 * _poissons_ratio(t, x)[0]);
+        };
+
+        double mu(double const t, X const& x) const
```
wenqing: Shear modulus.

ProcessLib/SmallDeformation/SmallDeformationFEM.h
wenqing: Maybe one like TensorVectorType

ProcessLib/SmallDeformation/SmallDeformationProcess.h
wenqing: For dynamic deformation, M is from the second derivative of displacement with respect to time. Therefore if-condition of (dynamic_cast<NumLib::ForwardEuler*> is not adequate for this case.

MaterialLib/SolidModels/LinearElasticIsotropic.h
endJunction: I'd only document both lambda and mu as Lame parameter, don't rename one to firstLameParameter and the other to shearModulus. OK?

wenqing: Documentations for the two parameter are OK (Especially for mu, the second Lame parameter, known as shear modulus in the physical meaning),

endJunction:
I'll try to remove the DisplacementDim in another PR, otherwise it'll take even more time.

If there are no more review comments coming, I will merge this tomorrow.

MaterialLib/SolidModels/CreateLinearElasticIsotropic.h
```diff
@@ -0,0 +1,49 @@
...
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef MATERIALLIB_SOLIDMODELS_CREATELINEARELASTICISOTROPIC_H_
+#define MATERIALLIB_SOLIDMODELS_CREATELINEARELASTICISOTROPIC_H_
+
+#include "ProcessLib/Utils/ProcessUtils.h"  // required for findParameter
```
norihiro-w: the circular dependency still exists here.

MaterialLib/SolidModels/MechanicsBase.h
```diff
@@ -0,0 +1,94 @@
...
+    /// createMaterialStateVariables().
+    struct MaterialStateVariables
+    {
+        virtual void pushBackState() = 0;
+    };
+
+    using KelvinVector = ProcessLib::KelvinVectorType<DisplacementDim>;
+    using KelvinMatrix = ProcessLib::KelvinMatrixType<DisplacementDim>;
+
+    void computeConstitutiveRelation(
```
norihiro-w: please document this function

MaterialLib/SolidModels/MechanicsBase.h
```diff
@@ -0,0 +1,94 @@
...
+        ProcessLib::SpatialPosition const& x,
+        double const dt,
+        KelvinVector const& eps_prev,
+        KelvinVector const& eps,
+        KelvinVector const& sigma_prev,
+        KelvinVector& sigma,
+        KelvinMatrix& C,
+        MaterialStateVariables& material_state_variables) = 0;
+
+    virtual std::unique_ptr<MaterialStateVariables>
```
norihiro-w: please document this function. Should one always call this function to generate a MaterialStateVariables object?

ProcessLib/SmallDeformation/CreateLocalAssemblers.h
```diff
@@ -0,0 +1,108 @@
...
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ */
+
+#ifndef PROCESSLIB_SMALLDEFORMATION_CREATE_LOCAL_ASSEMBLERS_H_
+#define PROCESSLIB_SMALLDEFORMATION_CREATE_LOCAL_ASSEMBLERS_H_
+
+#include <logog/include/logog.hpp>
+#include <vector>
```
norihiro-w: include std header first

ProcessLib/SmallDeformation/CreateLocalAssemblers.h
```diff
@@ -0,0 +1,108 @@
...
+ * \tparam LocalAssemblerImplementation the individual local assembler type
+ * \tparam LocalAssemblerInterface the general local assembler interface
+ * \tparam ExtraCtorArgs types of additional constructor arguments.
+ *         Those arguments will be passed to the constructor of
+ *         \c LocalAssemblerImplementation.
+ *
+ * The first two template parameters cannot be deduced from the arguments.
+ * Therefore they always have to be provided manually.
+ */
+template <unsigned DisplacementDim,
```
norihiro-w: at other places, `int` is used for `DisplacementDim`

ProcessLib/SmallDeformation/CreateLocalAssemblers.h
```diff
@@ -0,0 +1,108 @@
...
+            break;
+        case 3:
+            detail::createLocalAssemblers<3, DisplacementDim,
+                                          LocalAssemblerImplementation>(
+                dof_table, mesh_elements, integration_order, local_assemblers,
+                std::forward<ExtraCtorArgs>(extra_ctor_args)...);
+            break;
+        default:
+            OGS_FATAL(
+                "Meshes with dimension different than one and three are not "
```
norihiro-w: one -> two

ProcessLib/SmallDeformation/SmallDeformationFEM.h
```diff
@@ -0,0 +1,335 @@
...
+#include "ProcessLib/LocalAssemblerTraits.h"
+#include "ProcessLib/Parameter/Parameter.h"
+#include "ProcessLib/Utils/InitShapeMatrices.h"
+
+#include "SmallDeformationProcessData.h"
+
+template <typename BMatricesType, int DisplacementDim>
+struct IntegrationPointData final
+{
+    IntegrationPointData(Solids::MechanicsBase<DisplacementDim>& solid_material)
```
norihiro-w: explicit

ProcessLib/SmallDeformation/SmallDeformationFEM.h
```diff
@@ -0,0 +1,335 @@
...
+    }
+};
+
+namespace ProcessLib
+{
+namespace SmallDeformation
+{
+
+template <typename ShapeMatrixType>
+struct SecondaryData
```
norihiro-w: what is secondary data? why are these not included in IntegrationPointData?

ProcessLib/SmallDeformation/SmallDeformationFEM.h
```diff
@@ -0,0 +1,335 @@
...
+{
+    std::vector<ShapeMatrixType> N;
```
norihiro-w: why are shape functions stored here?

ProcessLib/SmallDeformation/SmallDeformationFEM.h
```diff
@@ -0,0 +1,335 @@
...
+            eps.noalias() =
+                B *
+                Eigen::Map<typename BMatricesType::NodalForceVectorType const>(
+                    local_x.data(), ShapeFunction::NPOINTS * DisplacementDim);
+
+            _ip_data[ip]._solid_material.computeConstitutiveRelation(
+                t, x_position, _process_data.dt, eps_prev, eps, sigma_prev,
+                sigma, C, material_state_variables);
+
+            _localRhs.noalias() -=
```
norihiro-w: why negative sign here?

ProcessLib/SmallDeformation/SmallDeformationFEM.h
```diff
@@ -0,0 +1,335 @@
...
+                B.transpose() * sigma * detJ * wp.getWeight();
+
+            // TODO: Reuse _ip_data[ip]._sigma
+            _secondary_data._sigmaXX[ip] = sigma[0];
+            _secondary_data._sigmaYY[ip] = sigma[1];
+            _secondary_data._sigmaXY[ip] = sigma[3];
+        }
+
+        b.add(indices.rows, _localRhs);
```
norihiro-w: is it true that body fore has not been included yet?

ProcessLib/SmallDeformation/SmallDeformationFEM.h
```diff
@@ -0,0 +1,335 @@
...
+    StiffnessMatrixType _localA;
+    NodalForceVectorType _localRhs;
+
+    IntegrationMethod _integration_method;
+    MeshLib::Element const& _element;
+    SecondaryData<typename ShapeMatrices::ShapeType> _secondary_data;
+};
+
+template <typename ShapeFunction, typename IntegrationMethod,
+          unsigned GlobalDim, unsigned DisplacementDim>
```
norihiro-w: int DisplacementDim ?

ProcessLib/SmallDeformation/SmallDeformationFEM.h
wenqing: I think It is for the residual for the Newton-Raphson.

ProcessLib/SmallDeformation/SmallDeformationFEM.h
endJunction: Yes, no body forces.

ProcessLib/SmallDeformation/SmallDeformationFEM.h
endJunction: Outer forces **-** inner forces.

endJunction: @nagelt Can you answer Norihiro's question please, I've tried to repeat, but my knowledge is limited...

norihiro-w: @nagelt What I don't understand is inconsistency of signs in residuals and Jacobian. if the residual is given as `r = external force - inner force`, Jacobian should be give as `J=dr/du=- d(inner force)/dt` with a negative sign. 

but if `_localRhs` is actually used as `-r`, the expression here is correct. I thought the assembly should return `r` and not `-r`.

norihiro-w: @endJunction @nagelt I solved it by myself. `MatrixTranslatorGeneral::computeResidual()` calculates residual as `r = M du/dt + K u - b`. This assembly is setting `b`. At the end, the residual becomes `r = inner force - external force` and the sign is consistent with the Jacobian. I don't think current approach is easy to understand and it should be improved in the future.

norihiro-w:
:+1: from my side

PR #1341, opened 2016-08-09
[NL] Fix Newton max iterations counter comparison.

endJunction:
Otherwise it behaves different than the Picard iterations: [NonlinearSolver.cpp, Picard solve method](https://github.com/ufz/ogs/blob/master/NumLib/ODESolver/NonlinearSolver.cpp#L53)

### DIFF
--- a/NumLib/ODESolver/NonlinearSolver.cpp
+++ b/NumLib/ODESolver/NonlinearSolver.cpp
@@ -189,7 +189,7 @@ bool NonlinearSolver<NonlinearSolverTag::Newton>::solve(
     minus_delta_x.setZero();
 
     unsigned iteration = 1;
-    for (; iteration < _maxiter; ++iteration)
+    for (; iteration <= _maxiter; ++iteration)
     {
         BaseLib::RunTime time_iteration;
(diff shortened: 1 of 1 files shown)


endJunction:
Also I find the iterations count starting from 1 rather unusual for the c-language family, but the TES process depends on the numbering of non-linear iterations...

chleh:
:fast_forward:

Concerning the start from one: I find it rather awkward to talk about a 0th iteration when talking to somebody. That's why I chose it the way it is.

nagelt:
0th iteration implies for me the initial guess of the residual.

norihiro-w:
+1

norihiro-w:
related to TN comment, i have a remark for the future. it looks current implementation computes residual and Jacobian at the same time, though only residual shoud be evaluated in some cases such as with line search, jagged Jacobian evaluation.

chleh:
@norihiro-w In connection with the numerical Jacobian assembly I created an interface that will provide different possibilities for Jacobian/residual computations. Maybe this interface will already suffice for the use-case you mentioned.

PR #1315, opened 2016-07-07
Make use of named functions to compute secondary variables

chleh:
Follow-up of #1314

TODO
* [x] docu

PR #1342, opened 2016-08-09
Added flexible nonlinear solver convergence criteria

### LINKED ISSUE #1339: Convergence criteria
I have some code adding flexible convergence criteria to our nonlinear solver in my private branches.
To be PR'ed soon.

chleh:
~~Follow-up of #1315 and of #1349.~~
Closes #1339.

Core things:
* The [ConvergenceCriterion](https://github.com/chleh/ogs6/blob/conv-crit/NumLib/ODESolver/ConvergenceCriterion.h) interface
* Its usage in the [nonlinear solvers](https://github.com/ufz/ogs/pull/1342/files#diff-bf2c8423f053d67a6d9d9a91d8ac0a87) 
* The changes to the input files ufz/ogs-data#17

TODO
* [x] squash
* [x] docu

For later:
* There is currently no residual-norm check class implemented. I think that an be done later on as soon as it is needed.

### DIFF
--- a/NumLib/ODESolver/ConvergenceCriterion.cpp
+++ b/NumLib/ODESolver/ConvergenceCriterion.cpp
@@ -12,6 +12,7 @@
 #include "BaseLib/Error.h"
 
 #include "ConvergenceCriterionDeltaX.h"
+#include "ConvergenceCriterionResidual.h"
 #include "ConvergenceCriterionPerComponentDeltaX.h"
 
 namespace NumLib
@@ -24,6 +25,8 @@ std::unique_ptr<ConvergenceCriterion> createConvergenceCriterion(
 
     if (type == "DeltaX") {
         return createConvergenceCriterionDeltaX(config);
+    } else if (type == "Residual") {
+        return createConvergenceCriterionResidual(config);
     } else if (type == "PerComponentDeltaX") {
         return createConvergenceCriterionPerComponentDeltaX(config);
     }
--- a/NumLib/ODESolver/ConvergenceCriterion.cpp
+++ b/NumLib/ODESolver/ConvergenceCriterion.cpp
@@ -1,0 +1,34 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#include "ConvergenceCriterion.h"
+#include "BaseLib/ConfigTree.h"
+#include "BaseLib/Error.h"
+
+#include "ConvergenceCriterionDeltaX.h"
+#include "ConvergenceCriterionPerComponentDeltaX.h"
+
+namespace NumLib
+{
+std::unique_ptr<ConvergenceCriterion> createConvergenceCriterion(
+    const BaseLib::ConfigTree& config)
+{
+    //! \ogs_file_param{process__convergence_criterion__type}
+    auto const type = config.peekConfigParameter<std::string>("type");
+
+    if (type == "DeltaX") {
+        return createConvergenceCriterionDeltaX(config);
+    } else if (type == "PerComponentDeltaX") {
+        return createConvergenceCriterionPerComponentDeltaX(config);
+    }
+
+    OGS_FATAL("There is no convergence criterion of type `%s'.", type.c_str());
+}
+
+}  // NumLib
--- a/NumLib/ODESolver/ConvergenceCriterion.cpp
+++ b/NumLib/ODESolver/ConvergenceCriterion.cpp
@@ -14,6 +14,7 @@
 #include "ConvergenceCriterionDeltaX.h"
 #include "ConvergenceCriterionResidual.h"
 #include "ConvergenceCriterionPerComponentDeltaX.h"
+#include "ConvergenceCriterionPerComponentResidual.h"
 
 namespace NumLib
 {
@@ -29,6 +30,8 @@ std::unique_ptr<ConvergenceCriterion> createConvergenceCriterion(
         return createConvergenceCriterionResidual(config);
     } else if (type == "PerComponentDeltaX") {
         return createConvergenceCriterionPerComponentDeltaX(config);
+    } else if (type == "PerComponentResidual") {
+        return createConvergenceCriterionPerComponentResidual(config);
     }
 
     OGS_FATAL("There is no convergence criterion of type `%s'.", type.c_str());
--- a/NumLib/ODESolver/ConvergenceCriterion.cpp
+++ b/NumLib/ODESolver/ConvergenceCriterion.cpp
@@ -37,4 +37,12 @@ std::unique_ptr<ConvergenceCriterion> createConvergenceCriterion(
     OGS_FATAL("There is no convergence criterion of type `%s'.", type.c_str());
 }
 
+bool checkRelativeTolerance(const double reltol, const double numerator,
+                            const double denominator)
+{
+    auto const eps = std::numeric_limits<double>::epsilon();
+    return std::abs(numerator) <
+           std::abs(reltol) * (std::abs(denominator) + eps);
+}
+
 }  // NumLib
--- a/NumLib/ODESolver/ConvergenceCriterion.h
+++ b/NumLib/ODESolver/ConvergenceCriterion.h
@@ -19,10 +19,11 @@ class ConfigTree;
 
 namespace NumLib
 {
-/*! Convergence criterion for nonlinear solvers.
+/*! Convergence criterion for iterative algorithms, like nonlinear solvers.
  *
  * It is able to check the difference of the solution vector between iterations
- * and the residual of the nonlinear equation system.
+ * and the residual of the equation being solved (e.g. the nonlinear equation
+ * system).
  */
 class ConvergenceCriterion
 {
--- a/NumLib/ODESolver/ConvergenceCriterion.h
+++ b/NumLib/ODESolver/ConvergenceCriterion.h
@@ -77,6 +77,13 @@ public:
 std::unique_ptr<ConvergenceCriterion> createConvergenceCriterion(
     BaseLib::ConfigTree const& config);
 
+//! Returns if |numerator/denominator| < |reltol|.
+//! This method copes with the case that denominator = 0 by always adding
+//! epsilon to the denominator.
+bool checkRelativeTolerance(double const reltol,
+                            double const numerator,
+                            double const denominator);
+
 } // namespace NumLib
 
 #endif  // NUMLIB_CONVERGENCECRITERION_H
--- a/NumLib/ODESolver/ConvergenceCriterion.h
+++ b/NumLib/ODESolver/ConvergenceCriterion.h
@@ -53,6 +53,10 @@ public:
     //! Check if the residual satisfies the convergence criterion.
     virtual void checkResidual(GlobalVector const& residual) = 0;
 
+    //! Tell the ConvergenceCriterion that it is called for the first time now
+    //! (while solving a specific nonlinear system).
+    virtual void preFirstIteration() {}
+
     //! Indicate that a new iteration now starts.
     //!
     //! A concrete implementation of ConvergenceCriterion might want to check
--- a/NumLib/ODESolver/ConvergenceCriterion.h
+++ b/NumLib/ODESolver/ConvergenceCriterion.h
@@ -54,6 +54,12 @@ public:
     virtual void checkResidual(GlobalVector const& residual) = 0;
 
     //! Indicate that a new iteration now starts.
+    //!
+    //! A concrete implementation of ConvergenceCriterion might want to check
+    //! both delta x and the residual. I.e., in a new iteration both checks have
+    //! to be done anew. This method will make the ConvergenceCriterion forget
+    //! the result of all checks already done, s.t. all necessary checks will
+    //! have to be repeated in order to satisfy the ConvergenceCriterion.
     virtual void reset() = 0;
 
     //! Tell if the convergence criterion is satisfied.
--- a/NumLib/ODESolver/ConvergenceCriterion.h
+++ b/NumLib/ODESolver/ConvergenceCriterion.h
@@ -1,0 +1,71 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef NUMLIB_CONVERGENCECRITERION_H
+#define NUMLIB_CONVERGENCECRITERION_H
+
+#include <memory>
+#include "NumLib/NumericsConfig.h"
+
+namespace BaseLib {
+class ConfigTree;
+}  // BaseLib
+
+namespace NumLib
+{
+/*! Convergence criterion for nonlinear solvers.
+ *
+ * It is able to check the difference of the solution vector between iterations
+ * and the residual of the nonlinear equation system.
+ */
+class ConvergenceCriterion
+{
+public:
+    //! Tells if the change of the solution between iterations is checked.
+    //!
+    //! \remark
+    //! This method allows to save some computations if no such check will be
+    //! done.
+    virtual bool hasDeltaXCheck() const = 0;
+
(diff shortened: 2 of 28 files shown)


chleh:
I'd like to finish the documentation during code review once everybody agrees about the general structures introduced in this PR. So please at least have a brief look at the three pieces highlighted above.

chleh:
Jenkins, test this please.

NumLib/ODESolver/ConvergenceCriterionPerComponent.h
```diff
@@ -0,0 +1,33 @@
...
+
+#ifndef NUMLIB_CONVERGENCECRITERIONPERCOMPONENT_H
+#define NUMLIB_CONVERGENCECRITERIONPERCOMPONENT_H
+
+#include "ConvergenceCriterion.h"
+
+namespace MeshLib
+{
+class Mesh;
+}  // Mesh
```
endJunction: If commenting, then 'namespace MeshLib', please.

NumLib/ODESolver/ConvergenceCriterionPerComponentDeltaX.h
```diff
@@ -0,0 +1,56 @@
...
+    : public ConvergenceCriterionPerComponent
+{
+public:
+    explicit ConvergenceCriterionPerComponentDeltaX(
+        std::vector<double>&& absolute_tolerances,
+        std::vector<double>&& relative_tolerances,
+        MathLib::VecNormType norm_type);
+
+    bool hasDeltaXCheck() const override { return true; }
+    bool hasResidualCheck() const override { return false; }
```
endJunction: Just to have this point in mind: Having only two types of checks is easy. Adding an n-th one would require modification of n-1 classes, which might become problematic...

endJunction:
Technically OK, didn't look into design too much, tough.

Maybe a 10 minutes overview on Monday would be beneficial.

NumLib/ODESolver/ConvergenceCriterionPerComponentDeltaX.h
chleh: With every check added you'd also have to modify the nonlinear solver class. I think there are only two natural checks: delta x and the residual. The interface allows to combine these two checks in any desirable way.

NumLib/ODESolver/ConvergenceCriterion.cpp
```diff
@@ -0,0 +1,34 @@
...
+    //! \ogs_file_param{process__convergence_criterion__type}
+    auto const type = config.peekConfigParameter<std::string>("type");
+
+    if (type == "DeltaX") {
+        return createConvergenceCriterionDeltaX(config);
+    } else if (type == "PerComponentDeltaX") {
+        return createConvergenceCriterionPerComponentDeltaX(config);
+    }
+
+    OGS_FATAL("There is no convergence criterion of type `%s'.", type.c_str());
```
norihiro-w: ``%s'` -> `'%s'`? i mean using a single quotation for both. If it's your intention, it's ok.

NumLib/ODESolver/ConvergenceCriterion.h
```diff
@@ -0,0 +1,71 @@
...
+}  // BaseLib
+
+namespace NumLib
+{
+/*! Convergence criterion for nonlinear solvers.
+ *
+ * It is able to check the difference of the solution vector between iterations
+ * and the residual of the nonlinear equation system.
+ */
+class ConvergenceCriterion
```
norihiro-w: I would name it `NonlinearConvergenceCriterion` because you explain this is for nonlinear solvers. We need a convergence criterion for coupling iterations.

NumLib/ODESolver/TimeLoopSingleODE.h
```diff
@@ -36,13 +36,18 @@ class TimeLoopSingleODE final
...
+     * nonlinear solver.
      */
-    TimeLoopSingleODE(TDiscODESys& ode_sys,
-                               std::unique_ptr<GlobalLinearSolver>&& linear_solver,
-                               std::unique_ptr<NLSolver>&& nonlinear_solver)
+    TimeLoopSingleODE(
+        TDiscODESys& ode_sys,
+        std::unique_ptr<GlobalLinearSolver>&& linear_solver,
+        std::unique_ptr<NLSolver>&& nonlinear_solver,
+        std::unique_ptr<ConvergenceCriterion>&& convergence_criterion)
```
norihiro-w: why not set a `ConvergenceCriterion` object into a `nonlinear_solver` object beforehand? I don't see a reason to pass these two separately.

NumLib/ODESolver/ConvergenceCriterion.h
```diff
@@ -0,0 +1,71 @@
...
+    //! convergence criterion.
+    //!
+    //! \remark
+    //! The Newton-Raphson solver computes \c minus_delta_x. \c x is needed for
+    //! relative tolerances.
+    virtual void checkDeltaX(GlobalVector const& minus_delta_x,
+                             GlobalVector const& x) = 0;
+
+    //! Check if the residual satisfies the convergence criterion.
+    virtual void checkResidual(GlobalVector const& residual) = 0;
```
norihiro-w: initial or reference norm should be passed to calculate relative norm of residual.

NumLib/ODESolver/ConvergenceCriterion.h
```diff
@@ -0,0 +1,71 @@
...
+
+    //! Indicate that a new iteration now starts.
```
norihiro-w: can you document why this function needs to be called or how this function influences a behavior of this class.

NumLib/ODESolver/ConvergenceCriterionDeltaX.cpp
```diff
@@ -0,0 +1,73 @@
...
+         norm_x, error_dx / norm_x);
+
+    bool satisfied_abs = false;
+    bool satisfied_rel = false;
+
+    if (_abstol) {
+        satisfied_abs = error_dx < *_abstol;
+    }
+    if (_reltol) {
+        satisfied_rel = error_dx < *_reltol * norm_x;
```
norihiro-w: need to consider a case when norm_x is zero

NumLib/ODESolver/ConvergenceCriterionDeltaX.cpp
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    _satisfied = _satisfied && (satisfied_abs || satisfied_rel);
```
norihiro-w: If a user specifies both absolute and relative tolerances, they might expect that both conditions should be satisfied.

NumLib/ODESolver/ConvergenceCriterionPerComponent.h
```diff
@@ -0,0 +1,33 @@
...
+
+namespace NumLib
+{
+class LocalToGlobalIndexMap;
+
+class ConvergenceCriterionPerComponent : public ConvergenceCriterion
+{
+public:
+    virtual void setDOFTable(NumLib::LocalToGlobalIndexMap const& dof_table,
```
norihiro-w: indeed you need to pass only a mesh id, don't you?

norihiro-w: i mean, dof_table needs only a mesh id for query, not a whole mesh object

NumLib/ODESolver/ConvergenceCriterionPerComponentDeltaX.cpp
```diff
@@ -0,0 +1,108 @@
...
+void ConvergenceCriterionPerComponentDeltaX::checkDeltaX(
+    const GlobalVector& minus_delta_x, GlobalVector const& x)
+{
+    if ((!_dof_table) || (!_mesh))
+        OGS_FATAL("D.o.f. table or mesh have not been set.");
+
+    bool satisfied_abs = true;
+    bool satisfied_rel = true;
+
+    for (unsigned global_component = 0; global_component < _abstols.size();
```
norihiro-w: `_abstols` could be empty if a user doesn't specify. isn't it?

NumLib/ODESolver/ConvergenceCriterionPerComponentDeltaX.cpp
```diff
@@ -0,0 +1,108 @@
...
+    _satisfied = _satisfied && (satisfied_abs || satisfied_rel);
+}
+
+void ConvergenceCriterionPerComponentDeltaX::setDOFTable(
+    const LocalToGlobalIndexMap& dof_table, MeshLib::Mesh const& mesh)
+{
+    _dof_table = &dof_table;
+    _mesh = &mesh;
+
+    if (_dof_table->getNumberOfComponents() != _abstols.size())
```
norihiro-w: no need to check `reltols`?

NumLib/ODESolver/NonlinearSolver.cpp
```diff
@@ -69,6 +72,13 @@ bool NonlinearSolver<NonlinearSolverTag::Picard>::solve(
         sys.applyKnownSolutionsPicard(A, rhs, x_new);
         INFO("[time] Applying Dirichlet BCs took %g s.", time_dirichlet.elapsed());
 
+        if (_convergence_criterion->hasResidualCheck()) {
```
norihiro-w: add `!sys.isLinear() && `

NumLib/ODESolver/NonlinearSolver.cpp
```diff
@@ -208,7 +215,8 @@ bool NonlinearSolver<NonlinearSolverTag::Newton>::solve(
         sys.applyKnownSolutionsNewton(J, res, minus_delta_x);
         INFO("[time] Applying Dirichlet BCs took %g s.", time_dirichlet.elapsed());
 
-        auto const error_res = LinAlg::norm2(res);
+        if (_convergence_criterion->hasResidualCheck())
```
norihiro-w: need to check sys.isLinear()

norihiro-w:
and where is a convergence criterion class checking a relative error?

NumLib/ODESolver/ConvergenceCriterion.cpp
chleh: I usually intend it this way.

NumLib/ODESolver/ConvergenceCriterion.h
chleh: OK. But are the coupling criteria fundamentally different?

NumLib/ODESolver/ConvergenceCriterion.h
chleh: OK, indeed there is something missing to the interface here. But instead of passing a reference norm, I'd rather add a `preFirstIteration(residual)` method in which the initial residual can be computed.

NumLib/ODESolver/TimeLoopSingleODE.h
chleh: A nonlinear solver can be used to solve different processes, each of which can have different convergence criteria. At least the convergence criterion has to be accessible from the outside (see the `setDOFTable()`-method in `ConvergenceCriterionPerComponent`).

NumLib/ODESolver/ConvergenceCriterionPerComponent.h
chleh: Yes, I can change it to a mesh ID.

Btw: This method basically is the reason why I pass the convergence criterion to the nonlinear solver separately.

NumLib/ODESolver/ConvergenceCriterionPerComponentDeltaX.cpp
chleh: It is checked in the constructor that both have the same size.

NumLib/ODESolver/ConvergenceCriterionPerComponentDeltaX.cpp
chleh: No, because of the check in the constructor.

NumLib/ODESolver/NonlinearSolver.cpp
chleh: Good observation, thanks.

chleh:
`ConvergenceCriterionDeltaX` checks relative errors if you pass a relative tolerance. Residual checks are not yet implemented as said above. Probably I should do that when I fix the residual check interface.

NumLib/ODESolver/ConvergenceCriterionPerComponent.h
chleh: It would require to change several functions along the way to the DOF-table. I'd like to leave it for later refactoring.

NumLib/ODESolver/ConvergenceCriterion.h
norihiro-w: i guess quite similar, without residual checks

NumLib/ODESolver/TimeLoopSingleODE.h
norihiro-w: why not create a nonlinear solver object for each process? is there any reason to reuse it?

NumLib/ODESolver/ConvergenceCriterionPerComponent.h
norihiro-w: ok

NumLib/ODESolver/ConvergenceCriterionPerComponentDeltaX.cpp
norihiro-w: i see

norihiro-w:
sorry i meant residual check

NumLib/ODESolver/TimeLoopSingleODE.h
norihiro-w: maybe the issue is not scope of this PR. current code is fine in this PR.

norihiro-w:
you can support residual check later but keep in mind tha one can not trust Newton result without checking residuals.

chleh:
Too late... The last two commits add residual checks. :smile:

NumLib/ODESolver/ConvergenceCriterion.h
chleh: Then I would like to reuse the classes introduced here. So I'd like to keep the names.

endJunction:
:fast_forward: from my side.

NumLib/ODESolver/ConvergenceCriterion.h
norihiro-w: then please modify the description at line 22

NumLib/ODESolver/ConvergenceCriterionDeltaX.cpp
norihiro-w: pleasr forget the comment above. I misunderstood it with residual and delta x.

NumLib/ODESolver/ConvergenceCriterionDeltaX.cpp
norihiro-w: no critical, just remark:
as it's done in the residual check, it might be better to use the initial norm for calculating relative error, instead of current norm. Relative errors based on current norm may not show converging behaviour at some iterations, particularly when delta x is close to or larger than x.

NumLib/ODESolver/ConvergenceCriterionPerComponentDeltaX.cpp
```diff
@@ -0,0 +1,111 @@
...
+    auto abstols =
+        config.getConfigParameterOptional<std::vector<double>>("abstols");
+    auto reltols =
+        config.getConfigParameterOptional<std::vector<double>>("reltols");
+    auto const norm_type_str =
+        config.getConfigParameter<std::string>("norm_type");
+
+    if ((!abstols) && (!reltols))
+        OGS_FATAL(
+            "At least one of absolute or relative tolerance has to be "
```
norihiro-w: but the constructor of ConvergenceCriterionPerComponentDeltaX requires both abstols and reltols are provided...

NumLib/ODESolver/ConvergenceCriterion.h
chleh: Modified. I'm not enitrely happy with it, but I hope it has improved.

NumLib/ODESolver/ConvergenceCriterionDeltaX.cpp
chleh: Thanks for the remark. As discussed I'll leave it for later.

chleh:
I just added a separate function for relative tolerance checks, which also works if b in `a/b < tol` is zero.

norihiro-w:
:+1:

TomFischer:
:+1:

PR #1343, opened 2016-08-09
[PL] Add multicomponent boundary conditions.

endJunction:
The main work was done before, but the parser was missing.

The default value for the component is 0 such that there is no need to add <component> for single component variables.

TODO:
 - [x] Documentation of the config param.
 - [x] Changelog

### DIFF
--- a/ProcessLib/ProcessVariable.cpp
+++ b/ProcessLib/ProcessVariable.cpp
@@ -86,8 +86,25 @@ ProcessVariable::ProcessVariable(BaseLib::ConfigTree const& config,
                 "Found geometry type \"%s\"",
                 GeoLib::convertGeoTypeToString(geometry->getGeoType()).c_str());
 
-            // TODO, the 0 is the component_id. Need parser.
-            _bc_configs.emplace_back(std::move(bc_config), *geometry, 0);
+            auto component_id =
+                //! \ogs_file_param{boundary_condition__component}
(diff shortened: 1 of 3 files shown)


chleh:
:fast_forward: 

Regarding the default of zero: Then for multi-component-BCs one implicitly always sets BCs to component no. 0 if one forgets to explicitly specify it.
Could there be a check that for multi-component variables specifying the component is always mandatory, and for single-component variables it's not?

endJunction:
I'm adding an extra check, which in case of missing <component> for multi-component variables will issue the following message:
```
error: ConfigTree: In file `../Tests/Data/Mechanics/Linear/square_1e0.prj' at path <process_variables/process_variable/boundary_conditions/boundary_condition>: Key <component> has not been found at ConfigTree.cpp, line 229
```

ProcessLib/ProcessVariable.cpp
```diff
@@ -89,6 +89,9 @@ ProcessVariable::ProcessVariable(BaseLib::ConfigTree const& config,
             auto const component_id =
                     //! \ogs_file_param{boundary_condition__component}
                     bc_config.getConfigParameter<int>("component", 0);
+            if (_n_components > 1) {
+                bc_config.peekConfigParameter<int>("component");
```
chleh: I thought peek after get doesn't work. At least I didn't want it to work in that order. ^^

norihiro-w:
👍

PR #1345, opened 2016-08-10
[GW] Corrected the sign of the groundwater flow velocity

wenqing:
as titled.

### DIFF
--- a/ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
+++ b/ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
@@ -100,7 +100,7 @@ public:
                                  sm.detJ * wp.getWeight();
 
             // Darcy velocity only computed for output.
-            auto const darcy_velocity = (k * sm.dNdx *
+            auto const darcy_velocity = -(k * sm.dNdx *
                 Eigen::Map<const NodalVectorType>(local_x.data(), ShapeFunction::NPOINTS)
                 ).eval();
 


ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
```diff
@@ -105,7 +105,7 @@ class LocalAssemblerData
                 ).eval();
 
             for (unsigned d=0; d<GlobalDim; ++d) {
-                _darcy_velocities[d][ip] = darcy_velocity[d];
+                _darcy_velocities[d][ip] = -darcy_velocity[d];
```
norihiro-w: why not add the sign at line 103?

chleh:
Btw. There is a second error in the velocity computation: Since the velocity is computed during assembly, i.e. before the equation system is solved, it will always lag one iteration behind in the output (i.e., one timestep since there is only one iteration per timestep for GWFlow.).

ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
wenqing: Done.

wenqing:
@chleh  Yes, you are right. The velocity calculation must be performed after the linear equation being solved. The error will be corrected after liquid pcs is finished.

wenqing:
@chleh correction. After the non-linear solver.

norihiro-w:
:+1:

PR #1347, opened 2016-08-11
Transmit raster data to element properties

TomFischer:
Rework of tool to transmit raster data to a mesh.

### DIFF
--- a/Applications/Utils/SimpleMeshCreation/createMeshElemPropertiesFromASCRaster.cpp
+++ b/Applications/Utils/SimpleMeshCreation/createMeshElemPropertiesFromASCRaster.cpp
@@ -36,6 +36,17 @@
 #include "MeshLib/MeshEditing/Mesh2MeshPropertyInterpolation.h"
 #include "MeshLib/MeshEnums.h"
 
+void scaleMeshPropertyVector(MeshLib::Mesh & mesh,
+                             std::string const& property_name,
+                             double factor)
+{
+    boost::optional<MeshLib::PropertyVector<double> &> pv(
+        mesh.getProperties().getPropertyVector<double>(property_name));
+
+    for (auto & v : *pv)
+        v *= factor;
+}
+
 int main(int argc, char* argv[])
 {
     ApplicationsLib::LogogSetup logo_setup;
@@ -98,6 +109,17 @@ int main(int argc, char* argv[])
                                           "file name");
     cmd.add(mesh_arg);
 
+    std::vector<std::string> allowed_units{ "mm/a", "mm/month", "m/s" };
+    TCLAP::ValuesConstraint<std::string> allowed_units_constraints{
+        allowed_units};
+    TCLAP::ValueArg<std::string> unit_arg("u",
+                                          "input-unit",
+                                          "input unit of the data",
+                                          true,
+                                          "m/s",
+                                          &allowed_units_constraints);
+    cmd.add(unit_arg);
+
     cmd.parse(argc, argv);
 
     // read mesh
@@ -147,6 +169,22 @@ int main(int argc, char* argv[])
     MeshLib::Mesh2MeshPropertyInterpolation mesh_interpolation(
         src_mesh.get(), property_arg.getValue());
     mesh_interpolation.setPropertiesForMesh(dest_mesh.get());
+
+    double scale(1.0);
+    if (unit_arg.getValue() == "m/s")
+    {
+        scale = 1.0;
+    }
+    else if (unit_arg.getValue() == "mm/a")
+    {
+        scale = 1e-3 / (365.25 * 86400);
+    }
+    else if (unit_arg.getValue() == "mm/month")
+    {
+        scale = 1e-3 * (12.0 / (365.25 * 86400));
+    }
+
+    scaleMeshPropertyVector(*dest_mesh, property_arg.getValue(), scale);
 
     if (!out_mesh_arg.getValue().empty())
     {
--- a/Applications/Utils/SimpleMeshCreation/createMeshElemPropertiesFromASCRaster.cpp
+++ b/Applications/Utils/SimpleMeshCreation/createMeshElemPropertiesFromASCRaster.cpp
@@ -11,6 +11,7 @@
 
 #include <algorithm>
 #include <memory>
+#include <cmath>
 #include <numeric>
 
 #include <tclap/CmdLine.h>
@@ -35,74 +36,42 @@
 #include "MeshLib/MeshEditing/Mesh2MeshPropertyInterpolation.h"
 #include "MeshLib/MeshEnums.h"
 
-// From wikipedia:
-// http://en.wikipedia.org/wiki/Algorithms_for_calculating_variance. The
-// original is citing D.E. Knuth. TAOCP, vol 2.
-template <typename InputIterator>
-auto computeMeanAndVariance(InputIterator first, InputIterator last) ->
-    std::pair<typename InputIterator::value_type, typename InputIterator::value_type>
-{
-    using T = typename InputIterator::value_type;
-    std::size_t n = 0;
-    auto mu = T{0};
-    auto M2 = T{0};
-
-    while (first != last)
-    {
-        T const x = *first++;
-        n++;
-        auto delta = x - mu;
-        mu += delta/n;
-        M2 += delta * (x - mu);
-    }
-
-    if (n < 2)
-        return std::make_pair(mu, T{0});
-
-    return std::make_pair(mu, M2/(n - 1));
-}
-
-int main (int argc, char* argv[])
+int main(int argc, char* argv[])
 {
     ApplicationsLib::LogogSetup logo_setup;
 
     TCLAP::CmdLine cmd(
-            "Generates properties for mesh elements of an input mesh deploying a ASC raster file",
-            ' ',
-            "0.1");
+        "Generates properties for mesh elements of an input mesh deploying a "
+        "ASC raster file",
+        ' ',
+        "0.1");
 
-    TCLAP::ValueArg<std::string> out_mesh_arg("o",
-                                              "out-mesh",
-                                              "the mesh is stored to a file of this name",
-                                              false,
-                                              "",
-                                              "filename for mesh output");
-    cmd.add( out_mesh_arg );
+    TCLAP::ValueArg<std::string> out_mesh_arg(
+        "o",
+        "out-mesh",
+        "the mesh is stored to a file of this name",
+        false,
+        "",
+        "filename for mesh output");
+    cmd.add(out_mesh_arg);
 
-    TCLAP::ValueArg<bool> refinement_raster_output_arg("",
-                                                       "output-refined-raster",
-                                                       "write refined raster to a new ASC file",
-                                                       false,
-                                                       false,
-                                                       "0");
-    cmd.add( refinement_raster_output_arg );
+    TCLAP::ValueArg<bool> refinement_raster_output_arg(
+        "",
+        "output-refined-raster",
+        "write refined raster to a new ASC file",
+        false,
+        false,
+        "0");
+    cmd.add(refinement_raster_output_arg);
 
     TCLAP::ValueArg<unsigned> refinement_arg(
-            "r",
-            "refine",
-            "refinement factor that raises the resolution of the raster data",
-            false,
-            1,
-            "factor (default = 1)");
-    cmd.add( refinement_arg );
-
-    TCLAP::ValueArg<std::string> mapping_arg("",
-                                             "mapping-name",
-                                             "file name of mapping",
-                                             true,
-                                             "",
-                                             "file name");
-    cmd.add( mapping_arg );
+        "r",
+        "refine",
+        "refinement factor that raises the resolution of the raster data",
+        false,
+        1,
+        "factor (default = 1)");
+    cmd.add(refinement_arg);
 
     TCLAP::ValueArg<std::string> raster_arg("",
                                             "raster-file",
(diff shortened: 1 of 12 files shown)


MeshLib/Elements/Element.h
```diff
@@ -213,6 +213,8 @@ class Element
 
 }; /* class */
 
+bool isPointInElementXY(MathLib::Point3d const& p, Element const& e);
```
endJunction: Needs docu and a todo for other than triangle element types.

MathLib/GeometricBasics.h
```diff
@@ -154,6 +154,11 @@ bool barycentricPointInTriangle(
     double eps_pnt_out_of_plane = std::numeric_limits<float>::epsilon(),
     double eps_pnt_out_of_tri = std::numeric_limits<float>::epsilon());
 
+bool isPointInTriangleXY(MathLib::Point3d const& p,
```
endJunction: Short description...

Applications/Utils/SimpleMeshCreation/createMeshElemPropertiesFromASCRaster.cpp
```diff
@@ -35,212 +36,158 @@
...
                                             "file name");
-    cmd.add( raster_arg );
+    cmd.add(raster_arg);
+
+    TCLAP::ValueArg<std::string> property_arg(
+        "p",
+        "property-name",
+        "the name of the property the values are stored for",
+        false,
+        "gwn",
```
chleh: Why a default value?

Applications/Utils/SimpleMeshCreation/createMeshElemPropertiesFromASCRaster.cpp
```diff
@@ -35,212 +36,158 @@
...
+        "property name as string");
+    cmd.add(property_arg);
 
     TCLAP::ValueArg<std::string> mesh_arg("m",
                                           "mesh",
                                           "the mesh is read from this file",
                                           true,
                                           "test.msh",
```
chleh: Here is a default value, too. That's not from this PR, but maybe you can correct it?

Applications/Utils/SimpleMeshCreation/createMeshElemPropertiesFromASCRaster.cpp
```diff
@@ -35,212 +36,158 @@
...
+    cmd.add(mesh_arg);
+
+    std::vector<std::string> allowed_units{ "mm/a", "mm/month", "m/s" };
+    TCLAP::ValuesConstraint<std::string> allowed_units_constraints{
+        allowed_units};
+    TCLAP::ValueArg<std::string> unit_arg("u",
+                                          "input-unit",
+                                          "input unit of the data",
+                                          true,
+                                          "m/s",
```
chleh: Why do you cope with units? Why the default? Why the default value? Why the restriction? What if I have raster data with solar intensity (unit W/m^2/s)?

Applications/Utils/SimpleMeshCreation/createMeshElemPropertiesFromASCRaster.cpp
```diff
@@ -35,212 +36,158 @@
...
 
     // read raster and if required manipulate it
     auto raster = std::unique_ptr<GeoLib::Raster>(
-        FileIO::AsciiRasterInterface::getRasterFromASCFile(raster_arg.getValue()));
-    GeoLib::RasterHeader header (raster->getHeader());
-    if (refinement_arg.getValue() > 1) {
+        FileIO::AsciiRasterInterface::getRasterFromASCFile(
+            raster_arg.getValue()));
+    GeoLib::RasterHeader header(raster->getHeader());
+    if (refinement_arg.getValue() > 1)
```
chleh: Why can't I coarsen the data, i.e. refinement < 1?

Applications/Utils/SimpleMeshCreation/createMeshElemPropertiesFromASCRaster.cpp
```diff
@@ -35,212 +36,158 @@
...
-    for (unsigned row(0); row<header.n_rows; row++) {
-        for (unsigned col(0); col<header.n_cols; col++) {
+    for (unsigned row(0); row < header.n_rows; row++)
+    {
+        for (unsigned col(0); col < header.n_cols; col++)
+        {
             src_properties[row * header.n_cols + col] = *raster_it;
             ++raster_it;
         }
     }
```
chleh: Should it be checked that now `raster_it == raster->end()`?

Applications/Utils/SimpleMeshCreation/createMeshElemPropertiesFromASCRaster.cpp
```diff
@@ -35,212 +36,158 @@
...
 
-    {
-        double mu, var;
-        std::tie(mu, var) = computeMeanAndVariance(src_properties.begin(), src_properties.end());
-        INFO("Mean value of source: %f.", mu);
-        INFO("Variance of source: %f.", var);
-    }
+    std::unique_ptr<MeshLib::Mesh> src_mesh(
+        MeshLib::RasterToMesh::convert(*raster,
+                                       MeshLib::MeshElemType::TRIANGLE,
```
chleh: Why triangles? I'd convert pixels to quads.

Applications/Utils/SimpleMeshCreation/createMeshElemPropertiesFromASCRaster.cpp
```diff
@@ -35,212 +36,158 @@
...
-            ERR("Could not open file %s for writing the mapping.", property_fname.c_str());
-            return EXIT_FAILURE;
-        }
-
-        for (std::size_t k(0); k < n_dest_mesh_elements; k++)
-            property_out << k << " " << dest_properties[k] << "\n";
-        property_out.close();
+    else if (unit_arg.getValue() == "mm/a")
+    {
+        scale = 1e-3 / (365.25 * 86400);
```
chleh: I think the scaling from non-SI to SI units should be either done by a user-supplied scaling parameter or in some other tool, e.g., Paraview.

MathLib/GeometricBasics.cpp
```diff
@@ -169,6 +169,31 @@ bool barycentricPointInTriangle(MathLib::Point3d const& p,
...
+    // criterion: p-a = u0 * (b-a) + u1 * (c-a); 0 <= u0, u1 <= 1, u0+u1 <= 1
+    MathLib::DenseMatrix<double> mat(2, 2);
+    mat(0, 0) = b[0] - a[0];
+    mat(0, 1) = c[0] - a[0];
+    mat(1, 0) = b[1] - a[1];
+    mat(1, 1) = c[1] - a[1];
+    double y[2] = {p[0] - a[0], p[1] - a[1]};
+
+    MathLib::GaussAlgorithm<MathLib::DenseMatrix<double>, double*> gauss;
+    gauss.solve(mat, y, true);
```
chleh: There must be a solution without linear solvers...
... https://prlbr.de/2014/liegt-der-punkt-im-dreieck/ (german)
(optional).

MeshLib/Elements/Element.cpp
```diff
@@ -205,4 +206,22 @@ std::ostream& operator<<(std::ostream& os, Element const& e)
...
+bool isPointInElementXY(MathLib::Point3d const& p, Element const& e)
+{
+    for(std::size_t i(0); i<e.getNumberOfBaseNodes(); ++i) {
+        if (MathLib::sqrDist2d(p, *e.getNode(i)) <
+            std::numeric_limits<double>::epsilon())
+        {
+            return true;
+        }
+    }
+    assert(e.getNumberOfNodes() == 3);
```
chleh: The assertion will only be active in debug builds. Maybe better use if(...)?

MeshLib/MeshEditing/Mesh2MeshPropertyInterpolation.h
```diff
@@ -65,7 +61,7 @@ class Mesh2MeshPropertyInterpolation final
     void interpolateElementPropertiesToNodeProperties(std::vector<double> &interpolated_node_properties) const;
 
     Mesh const*const _src_mesh;
-    std::vector<double> const*const _src_properties;
+    std::string const& _property_name;
```
chleh: no reference here, I guess.

MeshLib/MeshEditing/Mesh2MeshPropertyInterpolation.h
```diff
@@ -31,32 +31,28 @@ class Mesh2MeshPropertyInterpolation final
...
-     * @param source_properties a vector containing property information
-     * assigned to the mesh. The number of entries in the property vector
-     * must be at least the number of different properties stored in mesh.
-     * For instance if mesh has \f$n\f$ (pairwise) different property
-     * indices the vector of properties must have \f$\ge n\f$ entries.
+     * @param property_name is the name of a PropertyVector in the \c
+     * source_mesh
      */
-    Mesh2MeshPropertyInterpolation(Mesh const*const source_mesh, std::vector<double> const*const source_properties);
+    Mesh2MeshPropertyInterpolation(Mesh const* const source_mesh,
```
chleh: Shouldn't it be `Mesh const&`? Or are nullptrs allowed?

MeshLib/MeshEditing/Mesh2MeshPropertyInterpolation.cpp
```diff
@@ -45,71 +47,86 @@ bool Mesh2MeshPropertyInterpolation::setPropertiesForMesh(Mesh *dest_mesh, std::
...
+void Mesh2MeshPropertyInterpolation::interpolatePropertiesForMesh(
+    Mesh *dest_mesh) const
 {
+    // check the existence of PropertyVector in the source mesh
+    boost::optional<MeshLib::PropertyVector<double> const&> opt_src_pv(
+        _src_mesh->getProperties().getPropertyVector<double>(_property_name));
+    if (!opt_src_pv) {
+        WARN("Did not find PropertyVector<double> \"%s\" in source mesh.",
+            _property_name.c_str());
+        return;
```
chleh: Maybe there should be a boolean return value telling that the method didn't succeed.

MeshLib/MeshEditing/Mesh2MeshPropertyInterpolation.cpp
```diff
@@ -45,71 +47,86 @@ bool Mesh2MeshPropertyInterpolation::setPropertiesForMesh(Mesh *dest_mesh, std::
...
+        opt_pv = dest_mesh->getProperties().createNewPropertyVector<double>(
+            _property_name, MeshItemType::Cell, 1);
+        if (!opt_pv) {
+            WARN("Could not get or create a PropertyVector of type double"
+                " using the given name \"%s\".", _property_name.c_str());
+            return;
+        }
+    }
+    MeshLib::PropertyVector<double> & dest_properties(opt_pv.get());
+    if (dest_properties.size() != dest_mesh->getNumberOfElements())
```
chleh: If that can only happen upon newly created properties, I'd rather move the resize call upwards to the creation code.

chleh:
Generally OK after questions. :+1:
But I think the unit-related code shouldn't be in a general tool. (Or it should be renamed to something with precipitation).

MeshLib/MeshEditing/Mesh2MeshPropertyInterpolation.h
TomFischer: The const reference  `_property_name` is initialized in the constructor and won't be updated in the life time of the object.

MeshLib/MeshEditing/Mesh2MeshPropertyInterpolation.h
TomFischer: A `nullptr` should not be a valid input. You are right, I'll change this to an refernce.

Applications/Utils/SimpleMeshCreation/createMeshElemPropertiesFromASCRaster.cpp
TomFischer: I'll move the conversion to different units into another tool that can be used separately. To answer your questions:
1. The conversion of values is sometimes necessary and a separate tool can be used in a workflow.
2. I restricted the values because in groundwater flow I had to deal with this unit combinations. I want to make it easy to handle for the user. Of course the the list of possible unit conversions will be more complete in the future.

Applications/Utils/SimpleMeshCreation/createMeshElemPropertiesFromASCRaster.cpp
TomFischer: To coarsen the data means information loss. The additional refinement parameter should give the user the possibility to refine the raster such that each cell of the destination mesh contains at least some (raster) information.

MeshLib/MeshEditing/Mesh2MeshPropertyInterpolation.h
chleh: But the string referenced could be destroyed.

Applications/Utils/SimpleMeshCreation/createMeshElemPropertiesFromASCRaster.cpp
TomFischer: Since `isPointInElementXY` handled only triangle elements. I changed `isPointInElementXY` and now quad elements are handled, too. I'll convert pixels to quads.

MathLib/GeometricBasics.cpp
TomFischer: Is there something wrong with this solution?

MeshLib/MeshEditing/Mesh2MeshPropertyInterpolation.h
TomFischer: But this is true for all parameters given by reference or pointer.

chleh: But a string is more likely to be passed from a local variable.

MathLib/GeometricBasics.cpp
chleh: No, maybe it's a bit slower. Nevermind.

MeshLib/Mesh.h
```diff
@@ -164,6 +164,10 @@ class Mesh : BaseLib::Counter<Mesh>
     Properties _properties;
 }; /* class */
 
+void scaleMeshPropertyVector(MeshLib::Mesh& mesh,
```
endJunction:
```c++
/// Scales the mesh property with name \c property_name by given \c factor. The property must be a "double" property.
```

MeshLib/MeshEditing/Mesh2MeshPropertyInterpolation.cpp
```diff
@@ -23,141 +23,141 @@
...
-        interpolated_node_properties[k] = (*_src_properties)[(*materialIds)[src_nodes[k]->getElement(0)->getID()]];
-        for (std::size_t j(1); j<n_con_elems; j++) {
-            interpolated_node_properties[k] += (*_src_properties)[(*materialIds)[src_nodes[k]->getElement(j)->getID()]];
+    for (std::size_t k(0); k < n_src_nodes; k++)
+    {
+        const std::size_t n_con_elems(src_nodes[k]->getNumberOfElements());
+        interpolated_properties[k] = elem_props[(src_nodes[k]->getElement(0))->getID()];
+        for (std::size_t j(1); j < n_con_elems; j++)
+        {
+            interpolated_properties[k] += elem_props[(src_nodes[k]->getElement(j))->getID()];
```
endJunction: Also possible:
```c++
    std::accumulate(node->getElements().begin(), node->getElements().end(),
                    interpolated_properties[k], [&](double a, Element* e) {
                        return v + elem_props[e->getID()];
                    });
```

PR #1349, opened 2016-08-12
Componentwise norm computation

chleh:
Used by the convergence criteria of #1342.
Predecessor of #1342.

This PR also removes some old norm-related code.

Note: The provided code will not work with PETSc. To be done later on.


NumLib/DOF/DOFTableUtil.cpp
```diff
@@ -30,4 +44,60 @@ std::vector<GlobalIndexType> getIndices(
...
 }
 
+double norm(GlobalVector const& x, unsigned const global_component,
+            MathLib::VecNormType norm_type,
+            LocalToGlobalIndexMap const& dof_table, MeshLib::Mesh const& mesh)
+{
+    // TODO that also includes ghost nodes.
+    std::size_t const n_nodes = mesh.getNumberOfNodes();
+
+    // TODO for PETSc some global accumulation is necessary.
```
wenqing: If x is a PETSc vector, PETSc routine can be called to compute the norms, which collects the distributed data. How about to add members to global vector classes for computing the norms?

NumLib/DOF/DOFTableUtil.cpp
```diff
@@ -30,4 +44,60 @@ std::vector<GlobalIndexType> getIndices(
     return indices;
 }
 
+double norm(GlobalVector const& x, unsigned const global_component,
```
TomFischer: This implementation of a norm function is very special. It mixes the usual norm computation with fetching the values. Is there a special reason to couple these two different tasks?

NumLib/DOF/DOFTableUtil.cpp
```diff
@@ -30,4 +44,60 @@ std::vector<GlobalIndexType> getIndices(
...
+            MathLib::VecNormType norm_type,
+            LocalToGlobalIndexMap const& dof_table, MeshLib::Mesh const& mesh)
+{
+    // TODO that also includes ghost nodes.
+    std::size_t const n_nodes = mesh.getNumberOfNodes();
+
+    // TODO for PETSc some global accumulation is necessary.
+    switch(norm_type)
+    {
+    case MathLib::VecNormType::NORM1:
```
TomFischer: Why not different function for each norm type? What is the advantage to decide the norm type via parameter?

NumLib/DOF/DOFTableUtil.cpp
chleh: For the norms of whole vectors there are already functions in `MathLib/LinAlg/LinAlg.h`, which use the PETSc routines in the respective case.
The norm function introduced here, however, only picks, e.g., every third element of the vector and computes a norm involving only those elements. Therefore I think this function should not be moved to the vector classes.
And I didn't check if PETSc supports such "masked" norm computations out of the box.

NumLib/DOF/DOFTableUtil.cpp
chleh: In the TES case the global vector has components p, T and x. I want to be able to set relative and absolute error tolerances per component.
Another implementation of this functionality would be to copy the wanted components into a temporary vector and compute the norm for that. Then functionality would be more separated but it would require additional storage.

NumLib/DOF/DOFTableUtil.cpp
chleh: Currently different functions are not needed, since this function is only used in the convergence criteria #1342. The advantage of passing the norm type by parameter is that it is configurable at runtime without polymorphism.

TomFischer: Maybe you could implement `norm1`, `norm2` and the infinity norm in separate functions and use `norm` as a wrapper. This way it is possible to "hard code" the norm function if this makes sense. In other situations one can use your flexible variant.

NumLib/DOF/DOFTableUtil.cpp
wenqing: I see.

NumLib/DOF/DOFTableUtil.cpp
chleh: I split off the individual norms.

chleh:
Rebased and squashed as requested.

MathLib/LinAlg/LinAlg.h
```diff
@@ -83,6 +84,21 @@ double norm2(MatrixOrVector const& x);
 template<typename MatrixOrVector>
 double normMax(MatrixOrVector const& x);
 
+template<typename MatrixOrVector>
+double norm(MatrixOrVector const& x, MathLib::VecNormType type)
+{
+    switch (type) {
+    case MathLib::VecNormType::NORM1:
```
endJunction: indentation...

ProcessLib/TES/TESProcess.cpp
```diff
@@ -379,15 +379,15 @@ TESProcess::computeRelativeHumidity(
...
     {
-        auto const p = getNodalValue(x, this->_mesh, dof_table, node_id,
+        auto const p = NumLib::getNodalValue(x, _mesh, dof_table, node_id,
                                      COMPONENT_ID_PRESSURE);
-        auto const T = getNodalValue(x, this->_mesh, dof_table, node_id,
+        auto const T = NumLib::getNodalValue(x, _mesh, dof_table, node_id,
                                      COMPONENT_ID_TEMPERATURE);
-        auto const x_mV = getNodalValue(x, this->_mesh, dof_table, node_id,
+        auto const x_mV = NumLib::getNodalValue(x, _mesh, dof_table, node_id,
                                         COMPONENT_ID_MASS_FRACTION);
```
endJunction: Have a look in TESProcess::computeEquilibriumLoading; `NumLib::` and `this->` are still used there...

ProcessLib/TES/TESProcess.cpp
```diff
@@ -379,15 +379,15 @@ TESProcess::computeRelativeHumidity(
          dof_table_single.dofSizeWithoutGhosts(),
          &dof_table_single.getGhostIndices(), nullptr});
 
-    GlobalIndexType nnodes = this->_mesh.getNumberOfNodes();
+    GlobalIndexType nnodes = _mesh.getNumberOfNodes();
```
endJunction: could be const but seems not to change the binary code in that case...

NumLib/DOF/DOFTableUtil.cpp
norihiro-w: FYI PETSc provides `VecGetSubVector ()` based on Index Set which basically stores mapping information. see http://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Vec/VecGetSubVector.html

endJunction:
Nice!
:fast_forward:

TomFischer:
:+1:

ProcessLib/TES/TESProcess.cpp
chleh: Did your really check the binaries? Pro-tip: Not doing so tremendously decreases the time needed for review :smile:

NumLib/DOF/DOFTableUtil.cpp
chleh: @norihiro-w Thanks for the info. I'd like to make the PETSc information later on. I'm a bit skeptical concerning the index set, because it takes additional storage and one has to build it first of all. (Maybe a manual gather is not more complicated)

PR #1351, opened 2016-08-14
Cleanup

endJunction:
Few things found by clang's address and undefined behaviour sanitizers, and one is pure cosmetical.
Review commit-wise.

@TomFischer one of the tests (ctest Ammer) will be failing now, but I don't know the right solution to this: for discussion.

### DIFF
--- a/GeoLib/AnalyticalGeometry.cpp
+++ b/GeoLib/AnalyticalGeometry.cpp
@@ -465,14 +465,28 @@ std::vector<MathLib::Point3d> lineSegmentIntersect2d(
         return std::vector<MathLib::Point3d>();
     }
 
+    // precondition: points a, b, c are collinear
+    // the function checks if the point c is onto the line segment (a,b)
     auto isCollinearPointOntoLineSegment = [](MathLib::Point3d const& a,
-        MathLib::Point3d const& b, MathLib::Point3d const& c)
+                                              MathLib::Point3d const& b,
+                                              MathLib::Point3d const& c) {
+        if (b[0] - a[0] != 0)
         {
-            double const t((c[0]-a[0])/(b[0]-a[0]));
-            if (0.0 <= t && t <= 1.0)
-                return true;
-            return false;
-        };
+            double const t = (c[0] - a[0]) / (b[0] - a[0]);
+            return 0.0 <= t && t <= 1.0;
+        }
+        else if (b[1] - a[1] != 0)
+        {
+            double const t = (c[1] - a[1]) / (b[1] - a[1]);
+            return 0.0 <= t && t <= 1.0;
+        }
+        else if (b[2] - a[2] != 0)
+        {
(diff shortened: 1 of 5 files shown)


ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
```diff
@@ -100,11 +100,12 @@ class LocalAssemblerData
                                  sm.detJ * wp.getWeight();
 
             // Darcy velocity only computed for output.
-            auto const darcy_velocity = -(k * sm.dNdx *
-                Eigen::Map<const NodalVectorType>(local_x.data(), ShapeFunction::NPOINTS)
-                ).eval();
+            NodalVectorType const darcy_velocity =
```
norihiro-w: is `NodalVectorType` really correct here? `darcy_velocity` should have a length of spatial dimension.

GeoLib/AnalyticalGeometry.cpp
```diff
@@ -466,13 +466,16 @@ std::vector<MathLib::Point3d> lineSegmentIntersect2d(
...
-                return true;
-            return false;
-        };
+                                              MathLib::Point3d const& b,
+                                              MathLib::Point3d const& c) {
+        if (b[0] - a[0] == 0)
+            OGS_FATAL(
+                "isCollinearPointOntoLineSegment is about to divide by "
+                "zero: a[0] = %g, b[0] = %g, and b[0] - a[0] = %g",
+                a[0], b[0], b[0] - a[0]);
```
chleh: Can't the algorithm be fixed by trying any of the other coordinates, i.e. `b[1] - a[1]` or `b[2] - a[2]`?

chleh:
After comments :+1:

ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
endJunction: Thanks! I did miss this one. The compiler missed too, because we don't compile PRs with fixed size matrices...

ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
```diff
@@ -100,11 +101,12 @@ class LocalAssemblerData
                                  sm.detJ * wp.getWeight();
 
             // Darcy velocity only computed for output.
-            auto const darcy_velocity = -(k * sm.dNdx *
```
chleh: The problem why the old code was "insane" according to clang is probably that the minus is not inside the parentheses that are being `eval()`'d.

endJunction: Good idea! ...and one more proof that auto const ... eval is not the best practice.

TomFischer:
:+1:

PR #1354, opened 2016-08-15
Updated vtkdiff

chleh:
* File IO errors detected earlier
* Some more information printed
* Rel/abs tol fixed.


endJunction:
:fast_forward: after the ctests are passing.

chleh:
Tests pass as far as I can see.

TomFischer:
Why do you need to make the relative tolerance weaker than before?

chleh:
Because before vtkdiff did not use the RELTOL value but the ABSTOL value for relative tolerances. Cf. ufz/vtkdiff@38e1bce37466d3ab39b7e58d63e184a9693980c5

TomFischer:
Thanks for clarification. :+1:

PR #1355, opened 2016-08-15
SWMM Interface

rinkk:
Interface to Storm Water Management Model (SWMM) input/output files.
* Conversion of input into Geometry/Mesh/CSV
* Simulation results can be attached to meshes as scalar arrays or output to CSV

The SwmmConverter utility is just a prototype / script example of how to use the interface.  Tom suggested including it as example code. Implementing the full range of available information found in output files would require 4 objects types with a minimum of 37 possible parameters (not counting optional ones) + time step managment. It's relatively easy to write scripts for the interface but a full fledged user interface would require more time and effort than I can currently muster.

The inclusion into CMake works with a new flag, OGS_USE_SWMM. Not sure how testing works with that.

Thanks to Tom + Lars for all the CMake/Git/Submodule help.


Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+    "flow lost to flooding",
+    "flow leaving through outfalls",
+    "volume of stored water",
+    "actual evaporation rate",
+    "potential evaporation rate"
+};
+
+std::array<std::size_t,4> const n_obj_params = { 8, 6, 5, 15 };
+
+SwmmInterface* SwmmInterface::create(std::string const& file_name)
```
endJunction: Returning a unique_ptr would improve memory management...

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+    }
+
+    in.close();
+    return true;
+}
+
+bool SwmmInterface::isSwmmOutputFile(std::string const& out_file_name)
+{
+    std::string path (out_file_name);
+    std::transform(path.begin(), path.end(), path.begin(), tolower);
```
endJunction: std::tolower maybe?

Applications/FileIO/SWMM/SWMMInterface.h
```diff
@@ -0,0 +1,245 @@
...
+    class Point;
+    class Polyline;
+}
+
+namespace MeshLib {
+    class Mesh;
+    class Node;
+    class Element;
+}
+
```
endJunction: Better to put the code into a namespace. (Not sure what is the correct name now, but something with IO...)

Applications/FileIO/SWMM/SWMMInterface.h
```diff
@@ -0,0 +1,245 @@
...
+enum class SwmmObject
+{
+    SUBCATCHMENT = 0,
+    NODE = 1,
+    LINK = 2,
+    SYSTEM = 3
+};
+
+/**
+ * Interface for reading files used within the  * Storm Water Management Model (SWMM) and
```
endJunction: No closing asterisk of SWMM title, or misplaced.

Applications/Utils/SWMMConverter/SWMMConverter.cpp
```diff
@@ -0,0 +1,118 @@
...
+            return -1;
+
+        mesh = swmm->getMesh();
+        MeshLib::IO::VtuInterface vtkIO(mesh, 0, false);
+        vtkIO.writeToFile(mesh_output_arg.getValue());
+    }
+
+    std::cout << "Simulation time steps: " << swmm->getNTimeSteps() << std::endl;
+
+/*
```
endJunction: Please remove unused code.

Applications/FileIO/SWMM/SWMMInterface.h
```diff
@@ -0,0 +1,245 @@
...
+    /// Other non-static methods rely on this being called in the beginning.
+    static SwmmInterface* create(std::string const& file_name);
+
+    /// If a mesh has already been created, this methods allows to add node- or link-arrays as property
+    /// to that mesh. The data vectors can be created using the getArrayAtTimeStep() method.
+    static bool addResultsToMesh(MeshLib::Mesh &mesh, SwmmObject const type,
+        std::string const& vec_name, std::vector<double> const& data);
+
+    /// Returns the mesh generated from SWMM file content.
+    MeshLib::Mesh* getMesh() const { return _mesh.get(); }
```
endJunction: It is not clear who is owning the mesh after `getMesh()` call. If it is the SwmmInterface, then better to return a reference, if not, then a unique_ptr.

Applications/Utils/SWMMConverter/SWMMConverter.cpp
```diff
@@ -0,0 +1,118 @@
...
+        if (!SwmmInterface::SwmmInputToGeometry(swmm_input_arg.getValue(), geo_objects, true))
+            return -1;
+
+        GeoLib::IO::BoostXmlGmlInterface xml(geo_objects);
+        xml.setNameForExport(BaseLib::extractBaseNameWithoutExtension(swmm_input_arg.getValue()));
+        xml.writeToFile(geo_output_arg.getValue());
+        return 0;
+    }
+
+    SwmmInterface* swmm = nullptr;
```
endJunction: swmm is not deleted. Please use unique_ptr to avoid memory leaks.

BaseLib/StringTools.h
```diff
@@ -19,10 +19,19 @@
...
 
 namespace BaseLib {
 
 /**
+ *  Splits a string into a vector of strings. This method only works for string seperation
+ *  recognised by the std::stringstream iterator such as ' ' or '\t'.
+ *  \param str String to be splitted
+ *  \return Vector of strings
+ */
+std::vector<std::string> splitString(std::string const& str);
```
endJunction: Just a note: There is also more general split function in boost.

MeshLib/Elements/LineRule2.h
```diff
@@ -39,6 +39,9 @@ class LineRule2 : public EdgeRule
     /// Constant: The FEM type of the element
     static const CellType cell_type = CellType::LINE2;
 
+    /// Constant: The number of edges
+    static const unsigned n_edges = 1;
```
endJunction: Not sure here, because 1d entities can have only one type of boundary, which is 0d. Can you explain please why it is needed or what is wrong without that edge information?

endJunction:
Some other points: 
 - getline is probably from the std namespace, but not marked as such.
 - it seems to me that the input streams are not checked for validity, especially after reading.

I've not checked the implementation details of the interface.

Applications/FileIO/SWMM/SWMMInterface.h
```diff
@@ -0,0 +1,245 @@
...
+ * For meshes, also output data can be added to nodes and elements.
+ * The interface also provides routines for returning data as vectors as well as convenience
+ * methods for outputting data into CSV files. CSV files will either contain all parameters for one
+ * object at all time steps or all parameters for all objects of a given type at one timestep.
+ */
+class SwmmInterface
+{
+public:
+    /// Basic method to create the interface (containing a mesh) from a SWMM file
+    /// Other non-static methods rely on this being called in the beginning.
```
chleh: Why must objects be constructed by this method? Maybe I'm only confused by the docu.

Applications/FileIO/SWMM/SWMMInterface.h
```diff
@@ -0,0 +1,245 @@
...
+ * Interface for reading files used within the  * Storm Water Management Model (SWMM) and
+ * converting the data therein into corresponding OGS data structures.
+ * SWMM distinguishes four different object types (defined in the SwmmObject enum):
+ *     * Subcatchments
+ *     * Nodes (or junctions)
+ *     * Links (or conduits)
+ *     * System
+ * Note that each object in a SWMM input file has a name.
+ * Each of the four object types has a (different) number of parameters. There can be an arbitrary
+ * number of nodes, links or subcatchments but only ever one system.
```
chleh: Only for my understanding: One system per SwmmInterface instance or one system per running application?

Applications/FileIO/SWMM/SWMMInterface.h
```diff
@@ -0,0 +1,245 @@
...
+ * The interface can convert the SWMM input data into a geometry or a (line-)mesh.
+ * For meshes, also output data can be added to nodes and elements.
+ * The interface also provides routines for returning data as vectors as well as convenience
+ * methods for outputting data into CSV files. CSV files will either contain all parameters for one
+ * object at all time steps or all parameters for all objects of a given type at one timestep.
```
chleh: Looks like this interface is really mighty...

Applications/FileIO/SWMM/SWMMInterface.h
```diff
@@ -0,0 +1,245 @@
...
+ */
+class SwmmInterface
```
chleh: Could be final, I guess.
I think calling a concrete implementation `Interface` is not very fortunate. (minor)

Applications/FileIO/SWMM/SWMMInterface.h
chleh: I'd say it differently: Return a reference if the mesh must not be a nullptr, otherwise the pointer is OK (for me the ownership obviously stays with the SwmmInterface)

Applications/FileIO/SWMM/SWMMInterface.h
```diff
@@ -0,0 +1,245 @@
...
+    std::vector<std::string> getNames(SwmmObject obj_type) const;
+
+    /// Returns the Name for the indexed object of the given type (or an empty string if an error occured).
+    std::string getName(SwmmObject obj_type, std::size_t idx) const;
+
+    /// Returns the number of objects of the given type.
+    std::size_t getNObjects(SwmmObject obj_type) const;
+
+    /// Returns the number of parameters (incl. pollutants) of the given type.
+    std::size_t getNParameters(SwmmObject obj_type) const;
```
chleh: New naming convention is `getNumberOfParameters`. Same for other methods.

Applications/FileIO/SWMM/SWMMInterface.h
```diff
@@ -0,0 +1,245 @@
...
+
+    /// Returns the number of time steps for the simulation results.
+    std::size_t getNTimeSteps() const;
+
+    /// Returns an array for a given variable at all nodes/links from a SWMM output file for a given time step.
+    std::vector<double> getArrayAtTimeStep(SwmmObject obj_type, std::size_t time_step, std::size_t var_idx) const;
```
chleh: Do you really want to return a copy of the array?

Applications/FileIO/SWMM/SWMMInterface.h
```diff
@@ -0,0 +1,245 @@
...
+    /// Returns an array for a given variable for one specific object from a SWMM output file for all time steps.
+    std::vector<double> getArrayForObject(SwmmObject obj_type, std::size_t obj_idx, std::size_t var_idx) const;
+
+    /// Writes the node coordinates into double vectors for writing of CSV files
+    bool getNodeCoordinateVectors(std::vector<double> &x, std::vector<double> &y, std::vector<double> &z) const;
+
+    /// Writes the inlet- and outlet IDs for links into vectors for writing of CSV files
+    bool getLinkPointIds(std::vector<std::size_t> &inlets, std::vector<std::size_t> &outlets) const;
+
+    /// Write a CSV file for all object of the given type at one time step
```
chleh: Is the data for all objects always homogeneous enough that they can be written to a tabular file such as CSV?

Applications/FileIO/SWMM/SWMMInterface.h
```diff
@@ -0,0 +1,245 @@
...
+    bool writeCsvForObject(std::string const& file_name, SwmmObject obj_type, std::size_t obj_idx) const;
+
+    /// Checks if file is a SWMM input file
+    static bool isSwmmInputFile(std::string const& inp_file_name);
+
+    /// Checks if file is a SWMM output file
+    static bool isSwmmOutputFile(std::string const& inp_file_name);
+
+    /// Reading a SWMM input file and conversion into OGS geometry.
+    static bool SwmmInputToGeometry(std::string const& inp_file_name, GeoLib::GEOObjects &geo_objects, bool add_subcatchments);
```
chleh: Long line. Please break.
The method name is "unusual": No verb and starting uppercase.

Applications/FileIO/SWMM/SWMMInterface.h
```diff
@@ -0,0 +1,245 @@
...
+    /// Constructor
+    SwmmInterface(std::string const& swmm_base_name);
+
+    /// Reading a SWMM input file and creating an OGS line mesh. This is automatically called when the object is created
+    bool convertSwmmInputToLineMesh();
+
+    /// Reading points from SWMM input file and converting them into OGS point-type vector.
+    /// This method is shared by geometry- and mesh conversion.
+    template <typename T>
+    static bool readCoordinates(std::ifstream &in, std::vector<T*> &points, std::vector<std::string> &names)
```
chleh: If you only use this method in the cpp file, then you can move the definition of this method to the cpp file, even though it is templated.

Applications/FileIO/SWMM/SWMMInterface.h
```diff
@@ -0,0 +1,245 @@
...
+    {
+        bool finished (false);
```
chleh: Unused variable.

Applications/FileIO/SWMM/SWMMInterface.h
```diff
@@ -0,0 +1,245 @@
...
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ */
+
+#ifndef SWMMINTERFACE_H_
+#define SWMMINTERFACE_H_
+
+#include <iostream>
```
chleh: If it's possible iosfwd would be better in a header.

Applications/FileIO/SWMM/SWMMInterface.h
```diff
@@ -0,0 +1,245 @@
...
+    bool addRainGaugeTimeSeriesLocations(std::ifstream &in);
+
+    /// Matches existing subcatchment names with subsequently read polylines marking the outlines of said subcatchments.
+    bool matchSubcatchmentsWithOutlines(std::vector<GeoLib::Polyline*> const& lines, std::vector<std::string> const& names);
+
+    /// Creates a temporary string vector containing all subcatchment names in correct order
+    std::vector<std::string> getSubcatchmentNameMap() const;
+
+    /// During geometry conversion, this adds elevation values to the existing point vector
+    static bool SwmmInterface::addPointElevation(std::ifstream &in,
```
chleh: This converts points (x, y, 0) -> (x, y, z)? Is that not yet implemented somewhere in OGS?

Applications/FileIO/SWMM/SWMMInterface.h
```diff
@@ -0,0 +1,245 @@
...
+    std::string const _base_name;
+    /// Vector storing the names of all nodes/junctions
+    std::vector<std::string> _id_nodename_map;
+    /// Vector storing the names of all links/conduits
+    std::vector<std::string> _id_linkname_map;
+    /// Vector storing the names of all pollutants
+    std::vector<std::string> _pollutant_names;
+    /// Vector storing information about all subcatchments
+    std::vector<Subcatchment> _subcatchments;
+    /// Seperate node vector containing points for defining subcatchment outlines
```
chleh: Separate

Applications/FileIO/SWMM/SWMMInterface.h
chleh: unneeded `SwmmInterface::`

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+    std::size_t const pos_beg = str.find_first_not_of(' ', 0);
+    if (pos_beg == str.find_first_of(" \n", pos_beg))
+        return true;
+
+    return false;
+}
+
+bool SwmmInterface::isCommentLine(std::string const& str)
+{
+    return (str.compare(str.find_first_not_of(' ', 0),2,";;") == 0);
```
chleh: What if the string is empty or consists only of spaces? What about tabs?

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+}
+
+bool SwmmInterface::isSectionFinished(std::string const& str)
+{
+    if (str.empty())
+        return true;
+
+    std::size_t const pos_beg = str.find_first_not_of(' ', 0);
+    if (pos_beg == str.find_first_of(" \n", pos_beg))
+        return true;
```
chleh: Please add a comment like
`the given string is empty/whitespace only`

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+#include "MeshLib/Mesh.h"
+#include "MeshLib/Node.h"
+#include "MeshLib/Elements/Line.h"
+#include "MeshLib/Properties.h"
+
+#include "Applications/FileIO/CsvInterface.h"
+
+#include "Applications/ApplicationsLib/LogogSetup.h"
+
+const std::array<std::string,9> subcatchment_vars =
```
chleh: Those are all possible variables that are stored in subcatchments/nodes/...? Please add comments.

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+    "direct inflow",
+    "total lateral inflow",
+    "flow lost to flooding",
+    "flow leaving through outfalls",
+    "volume of stored water",
+    "actual evaporation rate",
+    "potential evaporation rate"
+};
+
+std::array<std::size_t,4> const n_obj_params = { 8, 6, 5, 15 };
```
chleh: Please document

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+
+SwmmInterface* SwmmInterface::create(std::string const& file_name)
+{
+    if (file_name.length() < 5)
```
chleh: Isn't that also covered by the isSwmmInputFile and isSwmmOutputFile methods?

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+        return nullptr;
+
+    if (!(SwmmInterface::isSwmmInputFile(file_name) || SwmmInterface::isSwmmOutputFile(file_name)))
+        return nullptr;
+
+    std::string const base_name (file_name.substr(0, file_name.length() - 4));
```
chleh: I think we already have such a method somewhere.

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+    if (swmm->convertSwmmInputToLineMesh())
+        return swmm;
+
+    ERR ("Error creating mesh from SWMM file.");
+    delete swmm;
+    return nullptr;
+}
+
+SwmmInterface::SwmmInterface(std::string const& swmm_base_name)
+: _base_name (swmm_base_name), _mesh(nullptr)
```
chleh: Note: No need to init unique_ptrs, they are nullptrs by default.

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+{
+}
+
+SwmmInterface::~SwmmInterface()
+{
+    for (Subcatchment sc  : _subcatchments)
```
chleh: Subcatchment&, otherwise you create a copy.

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+        delete sc.outline;
+
+    for (GeoLib::Point* pnt : _subcatchment_points)
+        delete pnt;
+}
+
+bool SwmmInterface::isSwmmInputFile(std::string const& inp_file_name)
+{
+    std::string path (inp_file_name);
+    std::transform(path.begin(), path.end(), path.begin(), tolower);
```
chleh: So any of { inp, INP, Inp, iNp, inP, ... } is valid? odd :smile:

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+    std::string line;
+    bool header_found (false);
+    std::size_t pos_beg (0), pos_end (0);
+    while (!header_found)
+    {
+        getline(in, line);
+        pos_beg = line.find_first_not_of(' ', pos_end);
+        pos_end = line.find_first_of(" \n", pos_beg);
+
+        if (line.empty() || pos_beg==pos_end || line.compare(pos_beg,2,";;") == 0)
```
chleh: Use isCommentLine()?

chleh: Maybe add comment like "Skip any number of empty lines at the beginning of the file"

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+            continue;
+
+        if (line == "[TITLE]")
```
chleh: Does line end in "\n"? At least you account for that above in
`pos_end = line.find_first_of(" \n", pos_beg)`.

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+            header_found = true;
+        else
+        {
+            ERR ("SWMMInterface: input file type %s not recognised.",
```
chleh: Wouldn't it be better to say that "Input file XYZ is not a SWMM file?"?

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+                BaseLib::getFileExtension(inp_file_name).c_str());
+            return false;
+        }
+    }
+
+    in.close();
```
chleh: Closed automatically.

Applications/FileIO/SWMM/SWMMInterface.cpp
chleh: BTW: Error messages are inappropriate because you want to check if it is an input file. So better use INFO or DBUG?

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+        {
+            getline(in, line);
+            continue;
+        }
+
+        std::vector<std::string> split_str (BaseLib::splitString(line));
+        if (split_str.size() != 3)
+        {
+            ERR ("Polygon format not recognised.");
+            return false;
```
chleh: Here `p` might leak memory. Better use std::unique_ptr for it.

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+        }
+        std::string const current_name (split_str[0]);
+        auto const it = name_id_map.find(current_name);
+        if (it == name_id_map.end())
+        {
+            ERR ("SwmmInterface::addPointElevation(): Name %s not found coordinates map.", current_name.c_str());
+            return false;
+        }
+        std::size_t const id = it->second;
+        (*points[id])[2] = BaseLib::str2number<double>(split_str[1]);
```
chleh: Btw: `BaseLib::str2number` is bad as far as I can see: It does not check if the conversion succeeded. So `BaseLib::str2number<double>("x")` might silently yield 0. That should be corrected (in a different PR).
For now I think your usage is OK, but we should fix it.

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+        if (split_str.size() < 4)
+        {
+            ERR ("Format not recognised.");
+            return false;
+        }
+        std::string const current_name (split_str[0]);
+        auto const it = name_id_map.find(current_name);
+        if (it == name_id_map.end())
+        {
+            ERR ("SwmmInterface::addPointElevation(): Name %s not found coordinates map.", current_name.c_str());
```
chleh: General thing: Maybe keep a line count and add it to error messages?

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+    {
+        if (isCommentLine(line))
+        {
+            getline(in, line);
+            continue;
+        }
+
+        std::vector<std::string> const split_str (BaseLib::splitString(line));
+        // Conduits = 9, Pumps = 7, Weirs = 8
+        if (split_str.size() < 7)
```
chleh: It would be really nice if you could briefly document what the lines you are reading look like.

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+}
+
+/// Deletes the geometric objects and returns false
+bool geometryCleanup(std::vector<GeoLib::Point*> &points, std::vector<GeoLib::Polyline*> &lines)
+{
+    for (auto line : lines)
+        delete line;
+    for (auto point : points)
+        delete point;
+    return false;
```
chleh: Why false?

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+    {
+        ERR ("Lengt of point vector and point name vector do not match.");
+        return geometryCleanup(*points, *lines);
+    }
+
+    std::map<std::string, std::size_t> *name_id_map (new std::map<std::string, std::size_t>);
+    std::size_t const n_names (pnt_names.size());
+    for (std::size_t i=0; i<n_names; ++i)
+    {
+        if (pnt_names[i] != "")
```
chleh: ! ... .empty()

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+                return geometryCleanup(*points, *lines);
+        }
+    }
+
+    geo_objects.addPointVec(std::move(points), geo_name, name_id_map);
+    if (!lines->empty())
+    {
+        if (lines->size() != line_names.size())
+        {
+            ERR ("Lengt of line vector and line name vector do not match.");
```
chleh: Length

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+        if (split_str.size() < 4)
+        {
+            ERR ("Format not recognised.");
+            return false;
+        }
+        std::string const current_name (split_str[0]);
+        auto const it = name_id_map.find(current_name);
+        if (it == name_id_map.end())
+        {
+            ERR ("SwmmInterface::readNodeData(): Name %s not found coordinates map.", current_name.c_str());
```
chleh: not found **in** coordinates

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+        {
+            ERR ("Conduit format not recognised.");
+            return false;
+        }
+
+        std::string const inlet (split_str[1]);
+        auto const i_it = name_id_map.find(inlet);
+        if (i_it == name_id_map.end())
+        {
+            ERR ("SwmmInterface::readLineElements(): Inlet node %s not found coordinates map.", inlet.c_str());
```
chleh: missing "in"

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+            return false;
+        }
+
+        std::string const outlet (split_str[2]);
+        auto const o_it = name_id_map.find(outlet);
+        if (o_it == name_id_map.end())
+        {
+            ERR ("SwmmInterface::readLineElements(): Outlet node %s not found coordinates map.", outlet.c_str());
```
chleh: missing "in"

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+            return false;
+        }
+
+        Subcatchment sc;
+        sc.name = split_str[0];
+        sc.rain_gauge = std::numeric_limits<std::size_t>::max();
+        std::size_t const n_gauges (_rain_gauges.size());
+        for (std::size_t i=0; i<n_gauges; ++i)
+            if (_rain_gauges[i].first.getName() == split_str[1])
+                sc.rain_gauge = i;
```
chleh: Maybe break if found?

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+
+        if (sc.rain_gauge == std::numeric_limits<std::size_t>::max())
+        {
+            ERR ("Rain gauge for subcatchment \"%s\" not found.", split_str[0].c_str());
+            return false;
+        }
+
+        sc.outlet =  std::numeric_limits<std::size_t>::max();
```
chleh: No need to set it, I guess.

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+    std::string line;
+    while ( getline(in, line) )
+    {
+        if (line == "[COORDINATES]")
+        {
+            INFO ("Reading coordinates...");
+            if (!readCoordinates<MeshLib::Node>(in, nodes, _id_nodename_map))
+                return false;
+        }
+        /* todo: check if needed
```
chleh: I'd prefer all-caps TODO, because my IDE is set up s.t. it recognizes it.

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+        sc.outlet = it->second;
+        sc.area = BaseLib::str2number<double>(split_str[3]);
+        _subcatchments.push_back(sc);
+        getline(in, line);
+    }
+
+    return true;
+}
+
+bool SwmmInterface::convertSwmmInputToLineMesh()
```
chleh: Maybe better call the method readSwmmInput? "Convert" sounds like read, convert, write, forget.

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+        {
+            INFO ("Reading symbols...");
+            std::vector<GeoLib::Point*> points;
+            std::vector<std::string> names;
+            if (!readCoordinates(in,points, names))
+                return false;
+            for (std::size_t i=0; i<points.size(); ++i)
+            {
+                GeoLib::Station stn (points[i], names[i]);
+                _rain_gauges.push_back(std::pair<GeoLib::Station, std::string>(stn, ""));
```
chleh: Minor note: Could probably be `_rain_gauges.emplace_back(stn, "");`

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+            }
+        }
+    }
+
+    if (nodes.empty())
```
chleh: Please comment briefly what happened if nodes are empty.

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+    std::map< std::string, std::size_t> name_id_map;
+    std::size_t const n_nodes (nodes.size());
+    for (std::size_t i=0; i<n_nodes; ++i)
+        name_id_map[_id_nodename_map[i]] = i;
+    in.clear();
+    in.seekg(0, in.beg);
+
+    std::vector< MeshLib::Element* > elements;
+    std::vector<double> max_depth;
+    max_depth.resize(n_nodes);
```
chleh: That's also done by the respective constructor of std::vector.

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+    std::size_t const n_types = 3;
+    std::array< std::size_t, n_types> n_elem_types;
+    while ( getline(in, line) )
+    {
+        if (line == "[RAINGAGES]")
```
chleh: not RAINGAUGES?

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+        {
+            if (!_rain_gauges.empty())
+                addRainGaugeTimeSeriesLocations(in);
```
chleh: Isn't it an error if they are empty?

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+        }
+        if (line == "[SUBCATCHMENTS]")
```
chleh: else if

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+        {
+            INFO ("Reading subcatchment information...");
+            if (!readSubcatchments(in, name_id_map))
+                return false;
+        }
+        if (line == "[SUBAREAS]")
```
chleh: else if? and so on...

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+            INFO ("Reading storages...");
+            if (!readNodeData(in, nodes, name_id_map, max_depth, true))
+                return false;
+        }
+        else if (line == "[CONDUITS]")
+        {
+            INFO ("Reading conduits...");
+            if (!readLineElements(in, elements, nodes, name_id_map))
+                return false;
+            n_elem_types[0] = elements.size();
```
chleh: Can't you use the enum values as index?

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+        if (line == "[Polygons]")
+        {
+            INFO ("Reading subcatchments...");
+            std::vector<GeoLib::Polyline*> lines;
+            std::vector<std::string> line_names;
+            std::vector<std::string> tmp_names; // polygon points are nameless but the method requires a vector
+            if (!readPolygons(in, lines, line_names, _subcatchment_points, tmp_names))
+                return false;
+
+            if (!matchSubcatchmentsWithOutlines(lines, line_names))
```
chleh: Maybe rename the method to `matchSubcatchmentsToOutlines` the old name can be easiyl misread as `matchSubcatchmentsWithoutLines`

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+    {
+        for (MeshLib::Node* node : nodes)
+            delete node;
+        return false;
+    }
+
+    MeshLib::Properties props;
+    boost::optional< MeshLib::PropertyVector<int>& > mat_ids =
+        props.createNewPropertyVector<int>("MaterialIDs", MeshLib::MeshItemType::Cell, 1);
+    mat_ids->resize(elements.size(), 0);
```
chleh: Minor note: Zero is optional

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+
+    if (nodes.size() == max_depth.size())
+    {
+        boost::optional< MeshLib::PropertyVector<double>& > depth =
+            props.createNewPropertyVector<double>("Max Depth", MeshLib::MeshItemType::Node, 1);
+        depth->reserve(max_depth.size());
+        std::copy(max_depth.cbegin(), max_depth.cend(), std::back_inserter(*depth));
+    }
+    else
+        ERR ("Size of max depth array does not fit number of elements. Skipping array.");
```
chleh: Why not return false?

Applications/FileIO/SWMM/SWMMInterface.cpp
chleh: You don't check if you'd overwrite the values in `n_elem_types`.

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+            return _id_linkname_map[idx];
+    case SwmmObject::SUBCATCHMENT:
+        if (idx < _subcatchments.size())
+            return _subcatchments[idx].name;
+    case SwmmObject::SYSTEM:
+        if (idx == 0)
+            return std::string("System");
+    }
+    ERR ("Index out of bounds.");
+    return std::string("");
```
chleh: Does the caller really regularly check if the returned string is empty?

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+}
+
+std::size_t SwmmInterface::getNObjects(SwmmObject obj_type) const
+{
+    std::string const outfile (_base_name + ".out");
+    std::size_t n_time_steps (std::numeric_limits<std::size_t>::max());
+    if (OpenSwmmOutFile(const_cast<char*>(outfile.c_str())) != 0)
+        return 0;
```
chleh: Does that open some file? Is that needed? For me that's rather surprising.

Applications/FileIO/SWMM/SWMMInterface.h
```diff
@@ -0,0 +1,245 @@
...
+
+#include <iostream>
+#include <map>
+#include <memory>
+#include <string>
+#include <vector>
+
+#include <logog/include/logog.hpp>
+
+#include "ThirdParty/SWMMInterface/swmm5_iface.h"
```
chleh: Is it necessary to include this into the header file?

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+
+    switch (obj_type)
+    {
+        case SwmmObject::SUBCATCHMENT:
+            return SWMM_Nsubcatch;
+        case SwmmObject::NODE:
+            return SWMM_Nnodes;
+        case SwmmObject::LINK:
+            return SWMM_Nlinks;
```
chleh: Looks like this is a global variable from SWMM. Can there be ever two instances of SwmmInterface in one application? If not then probably all methods should be static.

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+    }
+    ERR ("Index out of bounds.");
+    return std::string("");
+}
+
+std::size_t SwmmInterface::getNObjects(SwmmObject obj_type) const
+{
+    std::string const outfile (_base_name + ".out");
+    std::size_t n_time_steps (std::numeric_limits<std::size_t>::max());
+    if (OpenSwmmOutFile(const_cast<char*>(outfile.c_str())) != 0)
```
chleh: If you made outfile non-const, you could omit the const cast.

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+    }
+
+    if (swmm_type == SwmmObject::LINK && data.size() != mesh.getNumberOfElements())
+    {
+        ERR ("Number of mesh elements (%d) does not match length of array (%d).", mesh.getNumberOfElements(), data.size());
+        return false;
+    }
+
+    MeshLib::Properties& p = mesh.getProperties();
+    MeshLib::MeshItemType item_type = (swmm_type == SwmmObject::NODE) ? MeshLib::MeshItemType::Node : MeshLib::MeshItemType::Cell;
```
chleh: Very long lines.

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+    {
+        ERR ("Error creating array \"%s\".", vec_name.c_str());
+        return false;
+    }
+    prop->reserve(data.size());
+    std::copy(data.cbegin(), data.cend(), std::back_inserter(*prop));
+    return true;
+}
+
+std::vector<double> SwmmInterface::getArrayAtTimeStep(SwmmObject obj_type, std::size_t time_step, std::size_t var_idx) const
```
chleh: long line

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+    CloseSwmmOutFile();
+    return name;
+}
+
+std::string SwmmInterface::getArrayName(SwmmObject obj_type, std::size_t var_idx, std::size_t n_pollutants) const
+{
+    std::size_t const n_vars (0);
+    if (obj_type == SwmmObject::SUBCATCHMENT)
+    {
+        if (var_idx < n_obj_params[0])
```
chleh: Maybe it would be better to have a variable n_obj_params_subcatchment etc per object type.

Applications/Utils/SWMMConverter/SWMMConverter.cpp
```diff
@@ -0,0 +1,118 @@
...
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ */
+
+#include <tclap/CmdLine.h>
+
+#include "Applications/FileIO/SWMM/SWMMInterface.h"
+#include "ThirdParty/SWMMInterface/swmm5_iface.h"
```
chleh: Is this header really needed?

BaseLib/StringTools.cpp
```diff
@@ -25,6 +25,16 @@
 
 namespace BaseLib
 {
+std::vector<std::string> splitString(std::string const& str)
```
chleh: Nice.

MeshLib/Elements/LineRule2.h
chleh: It seems to be the only occurence in this whole PR, so probably it can be savely removed?

chleh:
So you wrote a complete parser for SWMM input files, but have the SWMM output files handled by a third-party library. Can't that also do the input reading? (Maybe I missed that in your previous github discussions.)

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+    {
+        if (pnt_names[i] != "")
+            name_id_map->insert(std::make_pair(pnt_names[i], i));
+    }
+
+    // rewind stream and read links between junctions
+    in.clear();
+    in.seekg(0, in.beg);
+
+    while ( getline(in, line) )
```
endJunction: This is a good pattern for reading streams (because of the checked stream state). Nice!

endJunction:
I'll set a WIP label for now to indicate that we are waiting for the questions to be answered.

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: As far as I am aware we don't give unique pointers to potentially outside code? This is the same as calling ProjectData::getMesh() ... it also gives a regular pointer and not a unique one.

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: error C2039: 'tolower' : is not a member of 'std'

Applications/Utils/SWMMConverter/SWMMConverter.cpp
rinkk: This shows basically how to use the interface for writing csv and meshes with scalar data. I could uncomment it but given the examples I have that'd mean that the utility would write 1000s of files (> 1 gb) to your hdd without warning.

I could also remove the code but then there's no example how to use the interface.

Applications/FileIO/SWMM/SWMMInterface.h
rinkk: It's difficult to decide this: The mesh is basically defined in the SWMM input file. Everything that we consider scalar information on the mesh is stored in the SWMM output file. The idea is, that you can get just the mesh and still add scalar information afterwards. This'd be important if we extend this to a coupling of OGS / SWMM as is apparently planned. You'd want to get the mesh for OGS to do sth. with it but SWMM will still modify it again at some point.

MeshLib/Elements/LineRule2.h
rinkk: I don't think so, this a general issue I stumbled upon while doing this PR. You can return minimum and maximum edge lengths from any mesh. This is used in various analytical geometry algorithms. However, if the mesh in question is a mesh consisting completely of lines, the getMinEdgeLength() and getMaxEdgeLength() methods will return corrupted values (namely max-limits for min and 0 for max). This line fixes the issue and apparently also causes no problems with any of the benchmarks.

Applications/FileIO/SWMM/SWMMInterface.h
rinkk: So we can check if the interface is in a valid state. If we use a ctor we don't have a return value and wouldn't notice if sth. went wrong during object initalisation. You could then call all kinds of methods on the invalid object.

Applications/FileIO/SWMM/SWMMInterface.h
rinkk: Each SWMM model/simulation has only one System-object. Since one interface can only open one SWMM project, each interface also has one System-object. 
If you create multiple SWMM interfaces in your application, each of those would have its own System object.

Applications/FileIO/SWMM/SWMMInterface.h
rinkk: SWMM is apparently able to do a lot of things at the same time and all of those are stored in the files. That is the reason why there is no user interface (neither gui nor command line switches). You'd have to be very flexible and I currently don't have the time to do that.

Applications/FileIO/SWMM/SWMMInterface.h
rinkk: I can only make methods final but not classes, can't I?

Applications/FileIO/SWMM/SWMMInterface.h
rinkk: Yes, I have to. The vectors do not exist as such. So if the user asks for a certain vector, it is put together on the fly and a copy is returned. If I returned a reference, it'd be instantly gone.

Applications/FileIO/SWMM/SWMMInterface.h
rinkk: Yes. You have to specify an object type and all objects of a given type always have the same number of parameters.

Applications/FileIO/SWMM/SWMMInterface.h
rinkk: No general method for adding elevation exists to my knowledge. Elevation is usually added based on certain given data types such as rasters and meshes. The respective methods need a pnt-vec anyway to correctly interpolate the z-coordinate so it makes sense that no general method exists.

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: This is checked in the method above. This method is called afterwards when we already know there is *something* written in this line.

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: This is precisely what is written in the method-description in the header file.

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: It's not all variables but the ones that we know are *always* defined.

chleh:
@endJunction:
> getline is probably from the std namespace, but not marked as such.
That might be valid because of argument-dependent lookup.

Applications/FileIO/SWMM/SWMMInterface.h
chleh: So is `nullptr` a valid result for `getMesh()`?

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: The input/output methods take a base name and check if the corresponding i/o file for that base name exists. This check takes any swmm project file, i.e. [base name].[extension] which needs to be at least 5 chars because the extension is always 3 chars.

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: Not quite. The file tools method will return the base name of the file, while this returns the complete path without file extension.

Applications/Utils/SWMMConverter/SWMMConverter.cpp
chleh: If that functionality is useful, you could add a commandline flag for it. If it is not useful, a unit test might be appropriate. If you don't agree, the code should be compiled at least, e.g. by wrapping it in `if (false) { ... }` (or something similar that generates no compiler warnings, such as moving it to a separate function); and a comment indicating your intention would also be nice.

MeshLib/Elements/LineRule2.h
chleh: Then it looks like a case for a separate PR plus some unit tests added, doesn't it?

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: Yes, all of these are valid. Old Windows codes don't distinguish between capitalisation in file(names?) or at least extensions.

Applications/FileIO/SWMM/SWMMInterface.h
chleh: Classes can be final, too.

Applications/FileIO/SWMM/SWMMInterface.h
chleh: it was discussed when @norihiro-w made issue #1220 and PR #1243.

Applications/FileIO/SWMM/SWMMInterface.h
chleh: I see. Maybe `std::unique_ptr` would be a better return value.

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: No sure what you mean. I checked for "[whitespace]\n" above. In any case, the given check works, please let me know if you want it changed somehow...

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: It's an error that will cause the termination of the programme. Why would this be a debug message?

Applications/FileIO/SWMM/SWMMInterface.cpp
chleh: In this method you create an object on the heap using `new` and transfer the ownership of that object to the caller (because the caller will be the only one who knows about this memory address), so you do memory management.
On the other hand in `getMesh()` you return a pointer to something that is owned by `ProjectData`, so there is no memory management involved at this point.
The bottom line is: Here unique_ptr is much better.

Applications/FileIO/SWMM/SWMMInterface.cpp
chleh: I see. Seems like the filetools are inconistent. There is dropFileExtension, extractBaseName, extractBaseNameWithoutExtension, which seem to overlap in funcitonality.

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: I think that doesn't work. I add p to the lines vector. When the method is finished, unique pointers will be cleaned and the polygon in the vector will change to nullptr or undefined or something. Also, at this point the worst that could happen is that p is a nullptr. Otherwise it's already been added to the vector.

Applications/FileIO/SWMM/SWMMInterface.cpp
chleh: It's OK, getline docu says '\n' is not included in the read string, as far as I remember.

Applications/FileIO/SWMM/SWMMInterface.cpp
chleh: Missing header?

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: Not straightforward is lines are increamed or reset in various subroutines.

Applications/FileIO/SWMM/SWMMInterface.h
chleh: Thanks for the explanation.

Applications/FileIO/SWMM/SWMMInterface.h
chleh: I see.

Applications/FileIO/SWMM/SWMMInterface.h
chleh: I see.

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: The lines look like that
```
;;               Inlet            Outlet                      Manning    Inlet      Outlet     Init.      Max.      
;;Name           Node             Node             Length     N          Offset     Offset     Flow       Flow      
;;-------------- ---------------- ---------------- ---------- ---------- ---------- ---------- ---------- ----------
Drosselstrecke   343              zuKA             44.48951562 0.013      0          0          0          40        
```
and for various entries I have no idea what they mean.

Applications/FileIO/SWMM/SWMMInterface.cpp
chleh: But it's a general-purpose method. When writing an extension in the future, you might not remember that `str` must not be empty.

Applications/FileIO/SWMM/SWMMInterface.cpp
chleh: The result of this method true/false, telling whether the given fille is a SWMM file. The caller decides to abort the program. So inside this methods it's INFO/DBUG, outside this method it's ERR.

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: Because it equals the return value of the geometry conversion method. It's just a short cut, but if I did this in the geo-conversion I'd have to add "return false;" at least 10 times. Now I just write "return geometryCleanup()". Since the whole method is basically just a macro to save repeating code I figured it wouldn't matter.

Applications/FileIO/SWMM/SWMMInterface.cpp
chleh: You can add `p.release()` to the lines vector. You couldn't add a copy of the unique_ptr anyway, because it's non-copyable.

chleh: Btw: I can imagine that for an invalid input, such as
```
Polygon1 1.0 2.0
Polygon2 1.0 2.0 4.0
```
you will leak memory.

Applications/FileIO/SWMM/SWMMInterface.cpp
chleh: I see.

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: It doesn't hurt either.

Applications/FileIO/SWMM/SWMMInterface.cpp
TomFischer: cctype is the header

Applications/FileIO/SWMM/SWMMInterface.cpp
TomFischer: Could you move this explanation as a comment into the code?

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: What do you mean? If no needs where specified in the input nodes would be empty. It'd probably not a valid SWMM file, but checking doesn't cost anything.

Applications/FileIO/SWMM/SWMMInterface.cpp
chleh: OK, even though I think the code would be easier to read without this additional indirection.

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: no

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: No idea. You could probably run simulations without rain if you'd just add BCs somewhere in the channel network. I'm no expert, though.

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: It's just used in this one place in this one method to distinguish between different input sources. n_elem_types is a std::array so case of index errors you'd get compile errors.

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: Yes, but it doesn't hurt mention it explicitly.

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: Because I'm unsure of the meaning of "max depth". It's not available for all point types and it's input instead of output info. Nothing really points to this being an error, we'd have to consult someone who knows this stuff in detail.

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: Also, you cannot override this, because there's just one conduit section, one pump section, etc. 
The information is probably not even needed later, I'd just like to index the different sources for visualisation purposed and to be on the safe side should it become important later.

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: The empty string is the error value so you'd have to.

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: It opens the SWMM output file and reads the constants from that file.

Applications/FileIO/SWMM/SWMMInterface.h
rinkk: Not anymore. Moved.

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: You could threoretically open two or more SWMM files from your application.

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: I want this to be const.

rinkk:
> So you wrote a complete parser for SWMM input files, but have the SWMM output files handled by a third-party library. Can't that also do the input reading? (Maybe I missed that in your previous github discussions.)

Input are ascii files, output are binary files. The swmm_iface-routines only handle output files.

Applications/FileIO/SWMM/SWMMInterface.cpp
chleh: But the files you are reading here could be hand written, and somebody could write it wrong, e.g. by copy&pasting to many times. I think input routines should be able to detect such wrong data if it's easily possible. And here it seems to be easy.

Applications/FileIO/SWMM/SWMMInterface.cpp
chleh: I see.

Applications/FileIO/SWMM/SWMMInterface.h
rinkk: Not sure what you're aiming at but nullptr is not a valid result. However, should the mesh be nullptr, createSwmmInterface() would already have returned a nullptr and you couldn't call getmesh() in the first place.

Applications/Utils/SWMMConverter/SWMMConverter.cpp
rinkk: Uncommented the code and added comments

Applications/FileIO/SWMM/SWMMInterface.cpp
chleh: OK

Applications/FileIO/SWMM/SWMMInterface.h
rinkk: If I write ```final class SwmmInterface``` I get compile errors.

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: Why does it matter? Even if the string is empty, it's *not* a comment line. Apart from that it's a private method in this interface. Who should call it?

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: Whatever floats your boat. Changed.

Applications/FileIO/SWMM/SWMMInterface.cpp
endJunction: Instead of having `outfile` for one-time usage, you can simplify the whole story to:
```c++
if (OpenSwmmOutFile((_base_name + ".out").c_str()))
{
    ...
}
```

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: I'd like to see write a SWMM input file by hand. The small example I have has 300 nodes and input file is >160k.
Garbage in -> garbage out.

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -0,0 +1,1248 @@
...
+            return std::string("System");
+    }
+    ERR ("Index out of bounds.");
+    return std::string("");
+}
+
+std::size_t SwmmInterface::getNObjects(SwmmObject obj_type) const
+{
+    std::string const outfile (_base_name + ".out");
+    std::size_t n_time_steps (std::numeric_limits<std::size_t>::max());
```
endJunction: n_time_steps is unused, unless I oversee something...

Applications/FileIO/SWMM/SWMMInterface.cpp
chleh: Obviously you never have wrong errors in your input files. lucky you. I find it very annoying to search for errors that could have been easily handled in the input files. The OGS5 input parsers are an example, where file reading has been implemented rather wrong, the *.msh reading code in OGS6 is another example (I already wasted some hours because of that), and this very line here is an example, too.
What's the problem with writing robust, error-reporting input parsing routines?

Applications/FileIO/SWMM/SWMMInterface.h
chleh: But if it's not a valid result anyway, you could without any problem declare the method s.t. it returns a `const&`.

Applications/FileIO/SWMM/SWMMInterface.h
chleh: Like with methods final is added at the end:
`class SwmmInterface final`

Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -497,15 +546,15 @@ bool SwmmInterface::readLineElements(std::ifstream &in, std::vector<MeshLib::Ele
         auto const i_it = name_id_map.find(inlet);
         if (i_it == name_id_map.end())
         {
-            ERR ("SwmmInterface::readLineElements(): Inlet node %s not found coordinates map.", inlet.c_str());
+            ERR ("SwmmInterface::readLineElements(): Inlet node %s not found incoordinates map.", inlet.c_str());
```
endJunction: Is "incoordinates" correct?

Applications/FileIO/SWMM/SWMMInterface.cpp
chleh: Nevermind.

Applications/FileIO/SWMM/SWMMInterface.cpp
chleh: OK, thanks. Not so important.

Applications/FileIO/SWMM/SWMMInterface.h
rinkk: I don't get it. What happens to the mesh reference once the interface goes out of scope?

Applications/FileIO/SWMM/SWMMInterface.cpp
chleh: Nevermind. Just thought that empty node might have some special meaning, such as "invalid file".

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: How exactly would you like me to handle this?

Applications/Utils/SWMMConverter/SWMMConverter.cpp
```diff
@@ -59,26 +58,26 @@ int main(int argc, char *argv[])
...
-        mesh = swmm->getMesh();
-        MeshLib::IO::VtuInterface vtkIO(mesh, 0, false);
+        std::size_t a = swmm->getNumberOfObjects(FileIO::SwmmObject::SUBCATCHMENT);
+        mesh.reset(swmm->getMesh());
+        MeshLib::IO::VtuInterface vtkIO(mesh.get(), 0, false);
         vtkIO.writeToFile(mesh_output_arg.getValue());
     }
 
-    std::cout << "Simulation time steps: " << swmm->getNTimeSteps() << std::endl;
+    std::cout << "Simulation time steps: " << swmm->getNumberOfTimeSteps() << std::endl;
```
endJunction: Writing to cout is not nice. Better use the logog capabilities.

Applications/Utils/SWMMConverter/SWMMConverter.cpp
```diff
@@ -88,31 +87,42 @@ int main(int argc, char *argv[])
     csv.addVectorForWriting("z", z);
     for (std::size_t i=0; i<9; ++i)
     {
-        std::vector<double> data_vec = swmm->getArrayAtTimeStep(SwmmObject::NODE, 10, i);
-        csv.addVectorForWriting(swmm.getArrayName(SwmmObject::NODE, i), data_vec);
+        std::vector<double> data_vec = swmm->getArrayAtTimeStep(FileIO::SwmmObject::NODE, 10, i);
+        csv.addVectorForWriting(swmm->getArrayName(FileIO::SwmmObject::NODE, i), data_vec);
     }
     csv.writeToFile("d:/csvtest.csv");
```
endJunction: It's unlikely that there will be a directory named 'd:'.

Applications/Utils/SWMMConverter/SWMMConverter.cpp
```diff
@@ -88,31 +87,42 @@ int main(int argc, char *argv[])
...
     std::cout << "csv written" << std::endl;
     std::cin.ignore();
```
endJunction: There will be no input in some cases. What are your intentions here?

rinkk: To allow you to stop the programme before it dumps a few hunded or thousand megabytes of data on the hard drive.

Applications/FileIO/SWMM/SWMMInterface.cpp
chleh: You could just check if you already encountered such a section, can't you. If every section must only occur once in the input file, maintaining, e.g., a `std::set<std::string>` of the sections already read would be two extra lines per section plus the declaration:
```
std::set<std::string> already_read;
...
if (!already_read.insert("[CONDUITS]").second)
    OGS_FATAL("Read [CONDUITS] for the second time");
```
or similar.

endJunction:
Repeating: the state of input streams is not checked while reading. This is a major step towards robust code.

Applications/FileIO/SWMM/SWMMInterface.h
chleh: Just do
```
MeshLib::Mesh const& getMesh() const { return *_mesh; }
```
There is no difference in the binary compared to your current solution, but the interface emphasizes that the return value can never be `nullptr` (that's the only semantic difference between returning pointers and references).

rinkk:
> Repeating: the state of input streams is not checked while reading. This is a major step towards robust code.

I added some tests in two places. I don't know if this is what you wanted or if it's at the locations you wanted this at, but there you go. If you had something different in mind, let me know.

Also, in both cases the test is redundant because the steams have been tested before when it was checked if the files are actually the correct type of input file.

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: This requires changes at about 20 locations in that code. I'll do it in a few weeks when I'm back if it's so terribly important.

endJunction:
After each reading of a stream via getline the stream's state should be checked.
This can be done in loops nicely (like you did once)
```c++
while (std::getline(in, line)) { }
```
or using the returned stream of getline with `if (!getline(in, line)) OGS_FATAL...`, or checking the state of `in` directly with `if (!in) OGS_FATAL...`.

Applications/Utils/SWMMConverter/SWMMConverter.cpp
endJunction: I see, this is a nice feature for users.
Can you add a short end-to-end test running as part of the ctests (see for example [tests](https://github.com/ufz/ogs/blob/master/Applications/Utils/Tests.cmake) provided by Tom for the MapGeometryToMeshSurface utility )?

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: Another comment, because I'm actually annoyed by this: Pretty much no other ascii-based interface in OGS checks if code sections appear multiple time. Not Feflow, not Petrel, not even the VTU-Parser tests this and simply crashes. Yet it's important for SWMM files for whatever reason. Even SWMM itself doesn't test this by the way and simply takes whatever it finds ... it'll probably crash during simulation but there're no input file errors.

So basically, this is to say that I've changed my mind and won't add tests for errors that not even the original software tests for.

Applications/Utils/SWMMConverter/SWMMConverter.cpp
rinkk: You've lost me again. What exactly is a nice feature for users?

chleh:
So, how do we want to settle this?

rinkk:
I fixed the getline()-thing. Is there anything else that absolutely needs to be done for this to be merged?

endJunction:
- Use logog instead of iostream for output.
 - Provide a basic ctest (like it is for MapGeometryToMeshSurface  Tom's example).
 - Either remove or split a PR regarding the LineRule2.

rinkk:
Is it a problem that the smallest test example I have is ~370 mb?

rinkk:
replaced iostream and removed linerule

endJunction:
Here are few smaller examples:
http://swmm2000.com/page/swmm-5-water-quality-example
https://www.epa.gov/water-research/storm-water-management-model-swmm#downloads in the application manual.

rinkk:
So it'd be okay to just test input files?

endJunction:
You have some input and the code gives some output. So the output must be checked.

rinkk:
Are you referring to OGS-output generated by the conversion code or to SWMM output generated by the SWMM code?  
I.e. should I just test input files or would I somehow need to generate SWMM output (how?) to test this, too?

rinkk:
The examples from your first link actually cause lots of error messages when loaded into SWMM5. They work with my code but the results look ... weird. The second link only points to the software, code and manuals as far as I can see. 

But if testing input files is enough I could use one of my examples. The input is less than 200k, it's the output that is so large.

rinkk:
Apparently we cannot test the interface because SWMM requires Windows and test functionality is not implemented for Windows.

bilke:
@endJunction That is true: all the functionality with the test wrapper and tester is *nix only yet. I will enable at least compilation on with the interface on Jenkins on Windows after this is merged.

endJunction:
OK, needs discussion about the testing procedures.

@rinkk  Put the tests onto TODO under the condition they will be indeed implemented.

endJunction:
:fast_forward:

Applications/Utils/CMakeLists.txt
```diff
@@ -15,4 +15,6 @@ else()
     add_subdirectory(ModelPreparation/PartitionMesh)
 endif() # OGS_BUILD_GUI
 
-include(Tests.cmake)
```
bilke: @rinkk: I guess this was deleted by accident?

rinkk: Yes, sorry.

PR #1356, opened 2016-08-17
Based Neumann/Robin BCs and ICs on the Parameter class.

chleh:
This PR is for the whole picture. I'll split off a PR for only the parameter class.
See also ufz/ogs-data#18.

Follow-up of #1357.


ProcessLib/Parameter/Parameter.cpp
```diff
@@ -0,0 +1,59 @@
...
+    else if (type == "MeshNode")
+    {
+        INFO("MeshElementParameter: %s", name.c_str());
+        auto param = createMeshNodeParameter(config, *meshes.front());
+        param->name = name;
+        return param;
+    }
+    else
+    {
+        OGS_FATAL("Cannot construct property of given type \'%s\'.",
```
TomFischer: Sould it be *parameter* instead of *property*?

ProcessLib/Parameter/Parameter.h
```diff
@@ -0,0 +1,68 @@
...
+{
+    virtual ~Parameter() = default;
+
+    //! Returns the number of components this Parameter has at every position and
+    //! point in time.
+    virtual unsigned getNumberOfComponents() const = 0;
+
+    //! Returns the parameter value at the given time and position.
+    virtual std::vector<T> const& getTuple(
+        double const t, SpatialPosition const& pos) const = 0;
```
TomFischer: 1. The name suggest to me a return value tuple, not a vector.
2. Why not `getParameterValue`?

Applications/ApplicationsLib/UncoupledProcessesTimeLoop.cpp
```diff
@@ -88,7 +88,7 @@ void UncoupledProcessesTimeLoop::setInitialConditions(
                 ode_sys.getMatrixSpecifications()));
 
         auto& x0 = *_process_solutions[pcs_idx];
-        pcs.setInitialConditions(x0);
+        pcs.setInitialConditions(t0, x0);
```
TomFischer: I do not understand why it is necessary to give the initial condition a time. I thought initial condition means t = 0.

chleh: In general simulations don't need to start at t=0. Furthermore the Parameter class which backs the ICs now requires a time argument.

ProcessLib/Parameter/Parameter.h
chleh: The name getTuple() is inspired by VTKs vtkDataArray class.

ProcessLib/Parameter/Parameter.cpp
chleh: Indeed. I'll correct it in #1357.

chleh:
@TomFischer thanks for your comments. Please keep in mind that this PR is a follow-up of #1357.

ProcessLib/BoundaryCondition/NeumannBoundaryCondition.cpp
```diff
@@ -0,0 +1,40 @@
...
+{
+    DBUG("Constructing Neumann BC from config.");
+    //! \ogs_file_param{boundary_condition__type}
+    config.checkConfigParameter("type", "Neumann");
+
+    //! \ogs_file_param{boundary_condition__Neumann__parameter}
+    auto const param_name = config.getConfigParameter<std::string>("parameter");
+    DBUG("Using parameter %s", param_name.c_str());
+
+    auto& param = findParameter<double>(param_name, parameters, 1);
```
TomFischer: `const`?

ProcessLib/BoundaryCondition/NeumannBoundaryCondition.cpp
```diff
@@ -0,0 +1,40 @@
...
+    NumLib::LocalToGlobalIndexMap const& dof_table, int const variable_id,
+    int const component_id, unsigned const integration_order,
+    unsigned const global_dim,
+    std::vector<std::unique_ptr<ParameterBase>> const& parameters)
+{
+    DBUG("Constructing Neumann BC from config.");
+    //! \ogs_file_param{boundary_condition__type}
+    config.checkConfigParameter("type", "Neumann");
+
+    //! \ogs_file_param{boundary_condition__Neumann__parameter}
```
TomFischer: Good occasion to add docu now.

ProcessLib/BoundaryCondition/NeumannBoundaryConditionLocalAssembler.h
```diff
@@ -26,11 +28,11 @@ class UniformNeumannBoundaryConditionLocalAssembler final
 public:
     /// The neumann_bc_value factor is directly integrated into the local
     /// element matrix.
-    UniformNeumannBoundaryConditionLocalAssembler(
+    NeumannBoundaryConditionLocalAssembler(
         MeshLib::Element const& e,
         std::size_t const local_matrix_size,
         unsigned const integration_order,
-        double neumann_bc_value)
+        Parameter<double> const& neumann_bc_value)
```
TomFischer: Parameter name should be changed.

ProcessLib/BoundaryCondition/NeumannBoundaryConditionLocalAssembler.h
```diff
@@ -42,30 +44,34 @@ class UniformNeumannBoundaryConditionLocalAssembler final
...
+
+        for (std::size_t ip(0); ip < n_integration_points; ip++)
+        {
+            pos.setIntegrationPoint(ip);
             auto const& sm = Base::_shape_matrices[ip];
             auto const& wp = integration_method.getWeightedPoint(ip);
-            _local_rhs.noalias() +=
-                sm.N * _neumann_bc_value * sm.detJ * wp.getWeight();
+            _local_rhs.noalias() += sm.N *
+                                    _neumann_bc_value.getTuple(t, pos).front() *
```
TomFischer: Why do you use only the first (*front*) component here?

ProcessLib/Process.cpp
```diff
@@ -34,7 +35,8 @@ Process::Process(
       _nonlinear_solver(nonlinear_solver),
       _time_discretization(std::move(time_discretization)),
       _convergence_criterion(std::move(convergence_criterion)),
-      _process_variables(std::move(process_variables))
+      _process_variables(std::move(process_variables)),
```
TomFischer: Would be nice if the initializer list order is the same as parameter order. Suggestion: move `parameters` parameter after `convergence_criterion`.

TomFischer:
After questions answered :+1:

ProcessLib/BoundaryCondition/NeumannBoundaryConditionLocalAssembler.h
chleh: Because the tuple only has one component, as defined here: https://github.com/ufz/ogs/pull/1356/files#diff-007117476faf1452c2d31123c4b60cb2R32

ProcessLib/BoundaryCondition/NeumannBoundaryCondition.cpp
chleh: I'd like to do that in a separate docu-PR afterwards.

ProcessLib/BoundaryCondition/NeumannBoundaryConditionLocalAssembler.h
chleh: To neumann_bc_value_parameter?

ProcessLib/Process.cpp
chleh: Changed the order as suggested.

ProcessLib/BoundaryCondition/NeumannBoundaryConditionLocalAssembler.h
endJunction: It's no longer a value (like in uniform Neumann bc). I'd prefer `parameter` here...

ProcessLib/BoundaryCondition/NeumannBoundaryConditionLocalAssembler.h
endJunction: OK, I see that now (after starting to rebase my mechanics branch).
The question came up, because it is not obvious, reading the lines around the access, that it has exactly one component.
I would like to add Parameter::operator(t, pos) returning the only one component (assert it).
Then the name `...value` would make much more sense, when reading `value(t, pos)`...

endJunction:
Independent of discussion points I touched :fast_forward:

endJunction:
I slightly reduced load on Jenkins, but it takes some time to handle all the updates of the PRs'. We have only one docker client running, and that one is not particularly fast.
Now the Linux-PRs are build only if Linux-PRs-dynamic returns success. The latter one does not execute ctests any longer.

ProcessLib/BoundaryCondition/NeumannBoundaryConditionLocalAssembler.h
chleh: I'm against `operator()(...)`, since this method does not have a name at all.
If there should be a method returning the only value, I'd call it `getScalarValue()` or something like that.

endJunction: But then reading `...value.getXXValue(t, x)` is strange. And the _neuman_bc_value is not a value at all, but a parameter or a scalarfield...
For discussion after merge...

ProcessLib/BoundaryCondition/NeumannBoundaryConditionLocalAssembler.h
chleh: renamed to `neumann_bc_parameter`. Plain `parameter` would be rather unspecific, I think.

PR #1357, opened 2016-08-17
General Parameter interface.

chleh:
Predecessor of #1356.

This PR introduces a general Parameter class whose main features are:
* space-time dependency via `double t` and `class SpatialPosition`
* parameters can have multiple components out of the box.
* See [Parameter](https://github.com/chleh/ogs6/blob/param-interface-base/ProcessLib/Parameter/Parameter.h)
and [SpatialPosition](https://github.com/chleh/ogs6/blob/param-interface-base/ProcessLib/Parameter/SpatialPosition.h)

The `SpatialPosition` class represents either a node, an element, an integration point or a set of coordinates.


chleh:
Jenkins, test this please.

ProcessLib/Parameter/MeshElementParameter.cpp
```diff
@@ -0,0 +1,48 @@
...
+    DBUG("Using field_name %s", field_name.c_str());
+
+    if (!mesh.getProperties().hasPropertyVector(field_name)) {
+        OGS_FATAL("The required property %s does not exists in the mesh.",
+                  field_name.c_str());
+    }
+
+    // TODO other data types than only double
+    auto const& property =
+        mesh.getProperties().template getPropertyVector<double>(field_name);
```
TomFischer: Why is *template* necessary?

ProcessLib/Parameter/MeshElementParameter.cpp
```diff
@@ -0,0 +1,48 @@
...
+    if (!property) {
+        OGS_FATAL("The mesh property `%s' is not of the requested type.",
+                  field_name.c_str());
+    }
+
+    if (property->getMeshItemType() != MeshLib::MeshItemType::Cell) {
```
TomFischer: Only for my understanding: Does this mean that parameters are defined on cells/elements only, i.e., parameters can't be defined on nodes for instance?

ProcessLib/Parameter/MeshElementParameter.h
```diff
@@ -0,0 +1,59 @@
...
+
+    std::vector<T> const& getTuple(double const /*t*/,
+                                   SpatialPosition const& pos) const override
+    {
+        auto const e = pos.getElementID();
+        assert(e);
+        auto const num_comp = _property.getNumberOfComponents();
+        for (std::size_t c=0; c<num_comp; ++c) {
+            _cache[c] = _property.getComponent(*e, c);
+        }
```
TomFischer: Suggestion: Line 42 - 45 could be implemented as `PropertyVector::getAllComponents(...)`? If this would be helpful I could implement this.

ProcessLib/Parameter/SpatialPosition.h
```diff
@@ -0,0 +1,93 @@
...
+namespace ProcessLib
+{
+//! Represents a position in space which can be either one of
+//! a node, an element, an integration point or a cartesian coordinates triple.
+//!
+//! The setters of this class make sure that only compatible information can be
+//! stored at the same time; e.g., it is not possible to specify an element ID
+//! and a node ID at the same time (the setAll() method being an exception to
+//! that rule).
+class SpatialPosition
```
TomFischer: This class has similarities with the struct `Location` (`MeshLib/Location.h`). Maybe this classes can be merge together?

ProcessLib/Process.cpp
```diff
@@ -66,18 +66,48 @@ void Process::initialize()
...
-        for (int component_id = 0; component_id < pv.getNumberOfComponents();
-             ++component_id)
+        auto const& ic = pv.getInitialCondition();
+
+        auto const num_comp = pv.getNumberOfComponents();
+
+        for (std::size_t node_id = 0; node_id < n_nodes; ++node_id)
         {
-            setInitialConditions(pv, variable_id, component_id, x);
+            MeshLib::Location const l(_mesh.getID(),
```
TomFischer: Is `l` still used?

ProcessLib/Parameter/SpatialPosition.h
chleh: Interesting observation. `SpatialPosition` is like a combination of a `MeshLib::Location` and a point. Let's talk on monday.

ProcessLib/Parameter/MeshElementParameter.h
chleh: CouLd be helpful in the future, indeed. Maybe also setters.

ProcessLib/Parameter/SpatialPosition.h
TomFischer: Yes, lets talk.

ProcessLib/Parameter/MeshElementParameter.cpp
chleh: Removed. GCC eats it without it.

TomFischer:
Code looks okay to me. After resolving conflicts: :fast_forward:

ProcessLib/Parameter/MeshNodeParameter.h
```diff
@@ -0,0 +1,59 @@
...
+
+    unsigned getNumberOfComponents() const override
+    {
+        return _property.getNumberOfComponents();
+    }
+
+    std::vector<T> const& getTuple(double const /*t*/,
+                                   SpatialPosition const& pos) const override
+    {
+        auto const e = pos.getNodeID();
```
endJunction: e -> node_id, n, ...

ProcessLib/GroundwaterFlow/CreateGroundwaterFlowProcess.cpp
```diff
@@ -37,12 +38,11 @@ std::unique_ptr<Process> createGroundwaterFlowProcess(
...
 
     // Hydraulic conductivity parameter.
-    auto& hydraulic_conductivity = findParameter<double,
-                                                 MeshLib::Element const&>(
+    auto& hydraulic_conductivity = findParameter<double>(
         config,
         //! \ogs_file_param_special{process__GROUNDWATER_FLOW__hydraulic_conductivity}
         "hydraulic_conductivity",
-        parameters);
+        parameters, 1);
```
endJunction: What is the meaning of `1`?

ProcessLib/Parameter/ConstantParameter.cpp
```diff
@@ -0,0 +1,28 @@
...
+#include "ConstantParameter.h"
+#include <logog/include/logog.hpp>
+#include "BaseLib/ConfigTree.h"
+
+namespace ProcessLib
+{
+std::unique_ptr<ParameterBase> createConstantParameter(
+    BaseLib::ConfigTree const& config)
+{
+    //! \ogs_file_param{parameter__type}
```
endJunction: Good time to write the documentation file...

ProcessLib/Parameter/MeshNodeParameter.h
```diff
@@ -0,0 +1,59 @@
...
+    unsigned getNumberOfComponents() const override
+    {
+        return _property.getNumberOfComponents();
+    }
+
+    std::vector<T> const& getTuple(double const /*t*/,
+                                   SpatialPosition const& pos) const override
+    {
+        auto const n = pos.getNodeID();
+        assert(n);
```
endJunction: Just a discussion idea: if getNodeID  is called, than a non-empty optional is expected (asserted). Maybe better to move the assertion into getNodeID and (if required) add an additional getOptionalNodeID() function.

ProcessLib/Parameter/Parameter.cpp
```diff
@@ -0,0 +1,59 @@
...
+    auto name = config.getConfigParameter<std::string>("name");
+    //! \ogs_file_param{parameter__type}
+    auto type = config.peekConfigParameter<std::string>("type");
+
+    // Create parameter based on the provided type.
+    if (type == "Constant")
+    {
+        INFO("ConstantParameter: %s", name.c_str());
+        auto param = createConstantParameter(config);
+        param->name = name;
```
endJunction: It's changed anyway, so I suggest to put the `name` into ctor to have something like this:
```c++
createSomeParameter(name, config);
```
Any reasons not to do so?

ProcessLib/Parameter/Parameter.cpp
```diff
@@ -0,0 +1,59 @@
...
+        return param;
+    }
+    else if (type == "MeshElement")
+    {
+        INFO("MeshElementParameter: %s", name.c_str());
+        auto param = createMeshElementParameter(config, *meshes.front());
+        param->name = name;
+        return param;
```
endJunction: The return value is a unique_ptr; is `return std::move(param)` better? Or should the return value be a `unique_ptr&&`?

ProcessLib/GroundwaterFlow/CreateGroundwaterFlowProcess.cpp
chleh: 1 == The parameter has one component.

ProcessLib/Parameter/Parameter.cpp
chleh: A return-by-value is an r-value automatically, so your second suggestion does not make sense.
The first suggestion looks more interesting, but in the past you (or your clang) told me that `return std::move(...)` defeats compiler optimizations.
Since this PR apparently builds (or built for some older state) on all of our platforms, obviously no compiler tries to copy the `unique_ptr`, so I wouldn't investigate it further now.

endJunction:
:fast_forward: afterwards...

ProcessLib/Parameter/Parameter.cpp
endJunction: Thanks! Got it (again :) .

ProcessLib/Parameter/Parameter.cpp
endJunction: Better to have a map<name, Parameter> in the project. Would simplify the access and the findParameter.
For later.

endJunction:
Jenkins, test this please!

chleh:
I won't squash the fixup commits since I have a file rename commit in between and I'm too lazy to untangle this.

PR #1358, opened 2016-08-17
Mesh element output

TomFischer:
Required for debugging of boundary conditions, especially for the element and node search routines.


MeshGeoToolsLib/convertMeshNodesToGeometry.cpp
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+#include "MeshGeoToolsLib/convertMeshNodesToGeometry.h"
+
+#include "MeshLib/Mesh.h"
+#include "MeshLib/Node.h"
+
+namespace MeshGeoToolsLib
+{
+void convertMeshNodesToGeometry(MeshLib::Mesh const& mesh,
```
rinkk: We already have MeshLib::convertMeshToGeo. It'd probably make sense to either move this functionality there or to move the other functionality here.

MeshLib/Utils/createMeshFromElements.h
```diff
@@ -0,0 +1,52 @@
...
+    {
+        new_elements.emplace_back(e->clone());
+    }
+
+   std::vector<MeshLib::Node*> new_nodes;
+    for (auto * e : new_elements)
+    {
+        for (std::size_t k(0); k < e->getNumberOfNodes(); ++k)
+        {
+            new_nodes.push_back(new MeshLib::Node(*(e->getNodes()[k])));
```
rinkk: If you elements actually form a mesh you will get tons of duplicated nodes here. Not sure if this might be a problem.

MeshLib/Utils/createMeshFromElements.h
```diff
@@ -0,0 +1,52 @@
...
+            e->setNode(k, new_nodes.back());
```
rinkk: What happens to the old node? Isn't this a memory leak?

rinkk: Sorry, it probably isn't a memory leak since the old element vector still exists...

endJunction:
What is to be discussed/done here?

MeshLib/Utils/createMeshFromElements.h
TomFischer: This is true. The nodes will be cleaned up deploying `MeshLib::MeshRevision` object, see lines 46-48.

MeshGeoToolsLib/convertMeshNodesToGeometry.cpp
TomFischer: Yes, you are right. I would move the existing `MeshLib::convertMeshToGeo` here `MeshGeoToolsLib` since `GeoLib` data structures as well as `MeshLib` data structures are needed. But the moving should be a separate PR.

TomFischer:
Was thought as helper/debugger for balance process. Seems this is currently not to be needed anymore.

PR #1359, opened 2016-08-17
Mechanics invariants and B-matrices

endJunction:
This is part of the mechanics PR #1340.

Includes:
 - deformation process specific invariants in the Kelvin mapping.
 - B-matrices (like shape-matrices for mechanics) and Kelvin vector types.

For the tests the error bounds were eyeballed from a gnuplot graphics showing the error vs. input vector norm. Tried to make them as tight as possible. It might happen that they will need to be increased a little bit... we'll see.

Aside from the failing compiler (of the test code) this is ready for review.

Todo: 
 - [x] Cite Nagel et al. Kelvin mapping.
 - [x] rename files/classes. Will do at the very end to reduce rereview:
  - [x] MathLib/mechanics_invariants.h > MaterialLib/SolidModels/KelvinVector.h
  - [x] Split off impl.h or even .cpp for specializations of det/inv.
 - [x] fix MSVC12 compilation.
 - [x] Possibly add test of LinearBMatrix using computeStrain function. Will not do.
 - [x] Test P_dev * P_sph == P_sph * P_dev == 0 6x6; P_dev + P_sph == identity 6x6;

 - [x] Move inverse out of Invariants.
 - [x] Maybe rename BMatrixPolicy::_columns to "number of DOF"? (And _rows to _kelvin_vector_size?)
 - [x] Merge/Unify ShapeMatrixPolicy and the BMatrixPolicy
  -  Now you use the ::detail::EigenMatrixType, why not the respective type from ShapeMatricesType?
  -  Wouldn't it be better to use the typedefs from ShapeMatrixPolicy here instead of always using fixed size matrices?
  - DynamicBMatrixPolicy Can be merged with the above into one type using only the typedefs from ShapeMatrixPolicyType.


Tests/MathLib/mechanics_invariants.cpp
```diff
@@ -48,7 +48,7 @@ Eigen::Matrix<double, KelvinToTensorSize<1>::value,
     kelvinToTensor(KelvinVector<1> const& v)
 {
     Eigen::Matrix<double, 1, 1> m;
-    m << v[0];
+    m.coeffRef(0,0) = v[0];
```
chleh: maybe `m(0,0) = v[0]`, or something like `Eigen::Matrix<double, 1, 1> m = Eigen::Matrix<double, 1, 1>::Constant(v[0])`?

Tests/MathLib/mechanics_invariants.cpp
```diff
@@ -0,0 +1,286 @@
...
+Eigen::Matrix<double, KelvinToTensorSize<Size>::value,
+              KelvinToTensorSize<Size>::value>
+kelvinToTensor(KelvinVector<Size> const& v);
+
+template <>
+Eigen::Matrix<double, KelvinToTensorSize<1>::value,
+              KelvinToTensorSize<1>::value>
+    kelvinToTensor(KelvinVector<1> const& v)
+{
+    Eigen::Matrix<double, 1, 1> m;
```
chleh: In the return type declaration you use `KelvinToTensorSize<1>::value`, here you just say "1".
But it's a test, so it's OK.

MathLib/mechanics_invariants.h
```diff
@@ -0,0 +1,166 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+#ifndef MATHLIB_MECHANICS_INVARIANTS_H
+#define MATHLIB_MECHANICS_INVARIANTS_H
```
chleh: The file name is lowercase, which is rather unusual for OGS.

NumLib/Fem/Deformation/BMatrixPolicy.h
```diff
@@ -0,0 +1,139 @@
...
+    using type =
+        Eigen::Matrix<double, KelvinVectorDimensions<DisplacementDim>::value,
+                      KelvinVectorDimensions<DisplacementDim>::value,
+                      Eigen::RowMajor>;
+};
+
+/// Special case for dimension one because of column major storage requirement
+/// from Eigen3.
+template <>
+struct KelvinMatrixTypeImpl<1>
```
chleh: I guess you need the struct `KelvinMatrixTypeImpl` only because one cannot specialize template typdefs.
If so, maybe that could be documented above.

NumLib/Fem/Deformation/BMatrixPolicy.h
```diff
@@ -0,0 +1,139 @@
...
+    static int const value = 6;
+};
+
+//
+// Kelvin vector and matrix templates for given displacement dimension.
+//
+
+template <int DisplacementDim>
+using KelvinVectorType =
+    Eigen::Matrix<double, KelvinVectorDimensions<DisplacementDim>::value, 1>;
```
chleh: Storage order not explicitly specified (only for the sake of completeness).
Wouldn't it be better to use the typedefs from `ShapeMatrixPolicy` here?

NumLib/Fem/Deformation/BMatrixPolicy.h
```diff
@@ -0,0 +1,139 @@
...
+template <int DisplacementDim>
+using KelvinMatrixType = typename KelvinMatrixTypeImpl<DisplacementDim>::type;
+
+/// An implementation of ShapeMatrixPolicy using fixed size (compile-time) eigen
+/// matrices and vectors.
+template <typename ShapeFunction, unsigned DisplacementDim>
+struct EigenFixedBMatrixPolicy
+{
+    template <int N>
+    using _vectorType = typename ::detail::EigenMatrixType<N, 1>::type;
```
chleh: I think typedefs should always start uppercase. So please put them into the private section if you don't want them to be used outside.

NumLib/Fem/Deformation/BMatrixPolicy.h
```diff
@@ -0,0 +1,139 @@
...
+
+    template <int N, int M>
+    using _matrixType = typename ::detail::EigenMatrixType<N, M>::type;
+
+    // Dimensions of specific b-matrix for n-points and displacement dimension.
+    static int const _columns = ShapeFunction::NPOINTS * DisplacementDim;
```
chleh: If that's private, please make it private.

NumLib/Fem/Deformation/BMatrixPolicy.h
chleh: Similar thing here as above: Now you use the `::detail::EigenMatrixType`, why not the respective type from `ShapeMatricesType`?

NumLib/Fem/Deformation/BMatrixPolicy.h
```diff
@@ -0,0 +1,139 @@
...
+    using StiffnessMatrixType = _matrixType<_columns, _columns>;
+
+    /// Rhs residual
+    using NodalForceVectorType = _vectorType<_columns>;
+
+    /// Sigma
+    using StressVectorType = _vectorType<_rows>;
+
+    /// C
+    using ModulusMatrixType = _matrixType<_rows, _rows>;
```
chleh: rename to Young's modulus? Or is this type something different/more general?

NumLib/Fem/Deformation/BMatrixPolicy.h
chleh: Maybe rename `_columns` to "number of DOF"? (And `_rows` to `_kelvin_vector_size`?)

NumLib/Fem/Deformation/BMatrixPolicy.h
```diff
@@ -0,0 +1,139 @@
...
+
+    using BMatrixType = _matrixType<_rows, _columns>;
+};
+
+/// An implementation of ShapeMatrixPolicy using dynamic size eigen matrices and
+/// vectors.
+template <typename ShapeFunction, unsigned DisplacementDim>
+struct EigenDynamicBMatrixPolicy
```
chleh: Can be merged with the above into one type using only the typedefs from `ShapeMatrixPolicyType`.

NumLib/Fem/Deformation/LinearBMatrix.h
```diff
@@ -0,0 +1,76 @@
...
+
+    b_matrix.setZero();
+
+    switch (DisplacementDim)
+    {
+        case 3:
+        {
+            for (int i = 0; i < NPOINTS; ++i)
+            {
+                b_matrix(2, 2 * NPOINTS + i) = dNdx(2, i);
```
wenqing: Can you documentation the formulas? This B matrix  is not from {\partial d\epsilon}{\partial du}. Besides, std::sqrt(2) can be computed only once.

MathLib/mechanics_invariants.h
```diff
@@ -0,0 +1,166 @@
...
+struct Invariants final
+{
+    // Deviatoric projection matrix.
+    static Eigen::Matrix<double, KelvinVectorSize, KelvinVectorSize> const
+        P_dev;
+    // Spherical projection matrix.
+    static Eigen::Matrix<double, KelvinVectorSize, KelvinVectorSize> const
+        P_sph;
+    /// Kelvin mapping of 2nd order identity
+    static Eigen::Matrix<double, KelvinVectorSize, 1> const ivec;
```
chleh: That's all fixed size matrices. Is it hard to make it switchable (dynamic/fixed)?
That's a minor thing and can be done later on.

Tests/MathLib/mechanics_invariants.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+template <>
+Eigen::Matrix<double, KelvinToTensorSize<6>::value,
+              KelvinToTensorSize<6>::value>
+    kelvinToTensor(KelvinVector<6> const& v)
+{
+    Eigen::Matrix<double, KelvinToTensorSize<6>::value,
+                  KelvinToTensorSize<6>::value>
+        m;
+    m << v[0], v[3] / std::sqrt(2.), v[5] / std::sqrt(2.), v[3] / std::sqrt(2.),
```
wenqing: std::sqrt(2) can be computed only once.

MathLib/mechanics_invariants.h
```diff
@@ -0,0 +1,166 @@
...
+
+    // Determinant of a matrix in Kelvin vector representation.
+    static double determinant(
+        Eigen::Matrix<double, KelvinVectorSize, 1> const& vec);
```
chleh: Introduce a typedef `KelvinVector`?

MathLib/mechanics_invariants.h
```diff
@@ -0,0 +1,166 @@
...
+    // Deviatoric projection matrix.
+    static Eigen::Matrix<double, KelvinVectorSize, KelvinVectorSize> const
+        P_dev;
+    // Spherical projection matrix.
+    static Eigen::Matrix<double, KelvinVectorSize, KelvinVectorSize> const
+        P_sph;
+    /// Kelvin mapping of 2nd order identity
+    static Eigen::Matrix<double, KelvinVectorSize, 1> const ivec;
+
+    // Determinant of a matrix in Kelvin vector representation.
```
chleh: That's no Doxygen comment.

MathLib/mechanics_invariants.h
```diff
@@ -0,0 +1,166 @@
...
+        return 0.5 * deviatoric_stress.transpose() * deviatoric_stress;
+    }
+
+    static double J3(
+        Eigen::Matrix<double, KelvinVectorSize, 1> const& deviatoric_stress)
+    {
+        return determinant(deviatoric_stress);
+    }
+
+    static double I1(Eigen::Matrix<double, KelvinVectorSize, 1> const& stress)
```
chleh: `//! Trace of the nxn tensor corresponding to the given Kelvin vector`.
Btw: Is the Kelvin vector concept restricted to stress or is it valid for any symmetric tensors? If the second fact holds true, I'd rename the argument `stress` to `kelvin_vector` or so; similar for the arguments of the methods above.

MathLib/mechanics_invariants.h
```diff
@@ -0,0 +1,166 @@
...
+    static double effectiveStress(
+        Eigen::Matrix<double, KelvinVectorSize, 1> const& deviatoric_stress)
+    {
+        return std::sqrt(3 * J2(deviatoric_stress));
+    }
+
+    static double J2(
+        Eigen::Matrix<double, KelvinVectorSize, 1> const& deviatoric_stress)
+    {
+        return 0.5 * deviatoric_stress.transpose() * deviatoric_stress;
```
chleh: The deviatoric part of a tensor is traceless (?), so there could be an assertion to check that.

MathLib/mechanics_invariants.h
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    static double I1(Eigen::Matrix<double, KelvinVectorSize, 1> const& stress)
+    {
+        return stress.template block<3, 1>(0, 0).sum();
+    }
+};
+
+template <>
+inline double Invariants<6>::determinant(Eigen::Matrix<double, 6, 1> const& vec)
```
wenqing: The name is confused. Determinant is a quantity of square matrix.

wenqing: It is better to doc. all formulas.

MathLib/mechanics_invariants.h
```diff
@@ -0,0 +1,166 @@
...
+
+    static double J3(
+        Eigen::Matrix<double, KelvinVectorSize, 1> const& deviatoric_stress)
+    {
+        return determinant(deviatoric_stress);
+    }
+
+    static double I1(Eigen::Matrix<double, KelvinVectorSize, 1> const& stress)
+    {
+        return stress.template block<3, 1>(0, 0).sum();
```
chleh: The block always has three rows. That cannot hold true.
Minor: There is also the method `topLeftCorner<R, C>()`.

MathLib/mechanics_invariants.h
```diff
@@ -0,0 +1,166 @@
...
+{
+    return vec(0) * vec(1) * vec(2) + vec(3) * vec(4) * vec(5) / std::sqrt(2.) -
+           vec(3) * vec(3) * vec(2) / 2. - vec(4) * vec(4) * vec(0) / 2. -
+           vec(5) * vec(5) * vec(1) / 2.;
+}
+
+template <>
+inline double Invariants<4>::determinant(Eigen::Matrix<double, 4, 1> const& vec)
+{
+    return vec(0) * vec(1) - vec(3) * vec(3) / 2.;
```
chleh: vec(2) doesn't contribute to the determinant?

MathLib/mechanics_invariants.h
```diff
@@ -0,0 +1,166 @@
...
+}
+
+template <int KelvinVectorSize>
+Eigen::Matrix<double, KelvinVectorSize, KelvinVectorSize> initP_dev()
+{
+    Eigen::Matrix<double, KelvinVectorSize, KelvinVectorSize> P_dev =
+        Eigen::Matrix<double, KelvinVectorSize, KelvinVectorSize>::Identity();
+
+    P_dev.template block<3, 3>(0, 0) -=
+        1. / 3. * Eigen::Matrix<double, 3, 3>::Ones();
```
chleh: Always a 3x3 block?

MathLib/mechanics_invariants.h
```diff
@@ -0,0 +1,166 @@
...
+    inv(1) = vec(0) * vec(2) - vec(5) * vec(5) / 2.;
+    inv(2) = vec(0) * vec(1) - vec(3) * vec(3) / 2.;
+    inv(3) = vec(4) * vec(5) / std::sqrt(2.) - vec(3) * vec(2);
+    inv(4) = vec(3) * vec(5) / std::sqrt(2.) - vec(4) * vec(0);
+    inv(5) = vec(4) * vec(3) / std::sqrt(2.) - vec(1) * vec(5);
+    return inv / determinant(vec);
+}
+
+template <int KelvinVectorSize>
+Eigen::Matrix<double, KelvinVectorSize, KelvinVectorSize> initP_dev()
```
chleh: Maybe put it into a `detail` namespace.
Furthermore: maybe split off a *-impl.h?

MathLib/mechanics_invariants.h
```diff
@@ -0,0 +1,166 @@
...
+const Eigen::Matrix<double, KelvinVectorSize, KelvinVectorSize>
+    Invariants<KelvinVectorSize>::P_dev = initP_dev<KelvinVectorSize>();
+
+template <int KelvinVectorSize>
+Eigen::Matrix<double, KelvinVectorSize, KelvinVectorSize> initP_sph()
+{
+    Eigen::Matrix<double, KelvinVectorSize, KelvinVectorSize> P_sph =
+        Eigen::Matrix<double, KelvinVectorSize, KelvinVectorSize>::Zero();
+
+    P_sph.template block<3, 3>(0, 0).setConstant(1. / 3.);
```
chleh: same 3x3 thing.

MathLib/mechanics_invariants.h
nagelt: it does; first term should be vec(0)*vec(1)*vec(2) I think

MathLib/mechanics_invariants.h
nagelt: +1

MathLib/mechanics_invariants.h
```diff
@@ -0,0 +1,166 @@
...
+template <int KelvinVectorSize>
+struct Invariants final
+{
+    // Deviatoric projection matrix.
+    static Eigen::Matrix<double, KelvinVectorSize, KelvinVectorSize> const
+        P_dev;
+    // Spherical projection matrix.
+    static Eigen::Matrix<double, KelvinVectorSize, KelvinVectorSize> const
+        P_sph;
+    /// Kelvin mapping of 2nd order identity
```
chleh: I'd rewrite the comment to
`/// The nxn identity tensor expressed as a Kelvin vector` or something similar.
Also I'd rename the member to `identity` or `identity_vec`.

MathLib/mechanics_invariants.h
nagelt: These get called quite often. So always checking tracelessness may not be efficient? In general you could pass any vector, make it deviatoric and then return the invariant. But in the interest of efficiency, I think the user "has to know" that deviatoric invariants need to be called with deviatoric arguments and would leave the implementation as is.

MathLib/mechanics_invariants.h
nagelt: It is the determinant of a square matrix. The matrix entries are just represented as a vector. I think the name is good, as it is exactly the quantity the function returns.

MathLib/mechanics_invariants.h
```diff
@@ -0,0 +1,166 @@
...
+ *
+ */
+#ifndef MATHLIB_MECHANICS_INVARIANTS_H
+#define MATHLIB_MECHANICS_INVARIANTS_H
+
+#include <Eigen/Dense>
+
+namespace MathLib
+{
+namespace Mechanics
```
chleh: I think most of the Kelvin business is not restricted to mechanics, but holds for any symmetric 2nd order tensor. So maybe one can omit the Mechanics namespace.
But I'm not an expert, probably @nagelt is...

MathLib/mechanics_invariants.h
nagelt: It holds true for KelvinVectorSizes 4 and 6. 1D is not yet implemented, afaik

MathLib/mechanics_invariants.h
nagelt: Yes, again for KelvinVectorSizes 4 and 6

NumLib/Fem/Deformation/LinearBMatrix.h
```diff
@@ -0,0 +1,76 @@
...
+    static_assert(0 < DisplacementDim && DisplacementDim <= 3,
+                  "LinearBMatrix::computeBMatrix: DisplacementDim must be in "
+                  "range [1,3].");
+
+    b_matrix.setZero();
+
+    switch (DisplacementDim)
+    {
+        case 3:
+        {
```
chleh: Very minor: no need for braces I guess since there are no nested variable declarations. (`int i` is already scoped)

NumLib/Fem/Deformation/LinearBMatrix.h
```diff
@@ -0,0 +1,76 @@
...
+        }
+        // no break for fallthrough.
+        case 1:
+        {
+            for (int i = 0; i < NPOINTS; ++i)
+                b_matrix(0, i) = dNdx(0, i);
+            break;
+        }
+        default:
+            break;
```
chleh: no need to break here, I guess.

NumLib/Fem/Deformation/LinearBMatrix.h
```diff
@@ -0,0 +1,76 @@
...
+        case 2:
+        {
+            for (int i = 0; i < NPOINTS; ++i)
+            {
+                b_matrix(1, NPOINTS + i) = dNdx(1, i);
+                b_matrix(3, i) = dNdx(1, i) / std::sqrt(2);
+                b_matrix(3, NPOINTS + i) = dNdx(0, i) / std::sqrt(2);
+            }
+        }
+        // no break for fallthrough.
```
chleh: please repeat this comment above `case 2`

NumLib/Fem/Deformation/LinearBMatrix.h
```diff
@@ -0,0 +1,76 @@
...
+{
+namespace LinearBMatrix
+{
+/// Fills a B-matrix based on given shape function dN/dx values.
+template <int DisplacementDim, int NPOINTS, typename DNDX_Type,
+          typename BMatrixType>
+void computeBMatrix(DNDX_Type const& dNdx, BMatrixType& b_matrix)
+{
+    static_assert(0 < DisplacementDim && DisplacementDim <= 3,
+                  "LinearBMatrix::computeBMatrix: DisplacementDim must be in "
```
chleh: Very minor: "LinearBMatrix::computeBMatrix" could be omitted. The compiler will tell you anyway.

MathLib/mechanics_invariants.h
```diff
@@ -0,0 +1,166 @@
...
+inline Eigen::Matrix<double, 4, 1> Invariants<4>::inverse(
+    Eigen::Matrix<double, 4, 1> const& vec)
+{
+    Eigen::Matrix<double, 4, 1> inv;
+    inv(0) = vec(1);
+    inv(1) = vec(0);
+    inv(2) = 0;
+    inv(3) = - vec(3);
+    return inv / determinant(vec);
+}
```
nagelt: inv(0) needs to be checked again (as the 6D case, only with 4 and 5 zero)

NumLib/Fem/Deformation/LinearBMatrix.h
```diff
@@ -0,0 +1,76 @@
...
+        }
+        default:
+            break;
+    }
+}
+
+/// Compute strain from given displacment and B-matrix.
+template <typename NodalDisplacementVectorType, typename BMatrixType,
+          typename StrainVectorType>
+void computeStrain(NodalDisplacementVectorType const& local_displacement,
```
chleh: Using this function it would be very easy to test the correctness of the BMatrix implementation, I guess.

Tests/MathLib/AutoCheckTools.h
```diff
@@ -67,6 +68,22 @@ struct IntervalTupleGenerator
     }
 };
 
+/// Generator for MxN fixed size eigen matrices with underlying type T.
+template <typename T, std::size_t M, std::size_t N, typename Gen = generator<T>>
```
chleh: Eigen uses signed types as sizes AFAIK.

MathLib/mechanics_invariants.h
wenqing: Need documentation for the armament, e.g. vector stores all elements of square matrix.

Tests/MathLib/AutoCheckTools.h
```diff
@@ -67,6 +68,22 @@ struct IntervalTupleGenerator
...
+struct randomEigenMatrixGenerator
+{
+    Gen source;
+
+    using result_type = Eigen::Matrix<T, M, N>;
+
+    result_type operator()(std::size_t size = 0)
```
chleh: What is the meaning of `size`? Why is it defaulted?

Tests/MathLib/AutoCheckTools.h
```diff
@@ -67,6 +68,22 @@ struct IntervalTupleGenerator
     }
 };
 
+/// Generator for MxN fixed size eigen matrices with underlying type T.
+template <typename T, std::size_t M, std::size_t N, typename Gen = generator<T>>
+struct randomEigenMatrixGenerator
+{
+    Gen source;
+
+    using result_type = Eigen::Matrix<T, M, N>;
```
chleh: Maybe the storage order should be configurable?

MathLib/mechanics_invariants.h
wenqing: also mention that the square matrix is  symmetric for symmetric strain or stress tensor.

MathLib/mechanics_invariants.h
chleh: Maybe rename it to `MathLib/KelvinVectorBusiness.`? :smile:
The file does not only cover invariants. (It computes the inverse, which is not an invariant I guess)

MathLib/mechanics_invariants.h
```diff
@@ -0,0 +1,166 @@
...
+{
+namespace Mechanics
+{
+/// Invariants used in mechanics, based on Kelvin representation of the vectors
+/// and matrices.
+/// The invariants are computed at process creation time.
+template <int KelvinVectorSize>
+struct Invariants final
+{
+    // Deviatoric projection matrix.
```
chleh: Maybe add a comment about the usage, like
`A_dev = P_dev * A`?

MathLib/mechanics_invariants.h
nagelt: yes, same thing (ok for 4 and 6 for KelvinVectorSize).

Tests/MathLib/mechanics_invariants.cpp
```diff
@@ -0,0 +1,286 @@
...
+template <>
+struct KelvinToTensorSize<6>
+{
+    static const int value = 3;
+};
+
+template <int Size>
+Eigen::Matrix<double, KelvinToTensorSize<Size>::value,
+              KelvinToTensorSize<Size>::value>
+kelvinToTensor(KelvinVector<Size> const& v);
```
chleh: I'm wondering if it makes sense to have these mapping functions in `MathLib`. When there will be anisotropic materials they could become handy. But the move can be done later on.

Tests/MathLib/mechanics_invariants.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+template <>
+Eigen::Matrix<double, KelvinToTensorSize<4>::value,
+              KelvinToTensorSize<4>::value>
+    kelvinToTensor(KelvinVector<4> const& v)
+{
+    Eigen::Matrix<double, KelvinToTensorSize<4>::value,
+                  KelvinToTensorSize<4>::value>
+        m;
+    // The z-component is ignored, otherwise the matrix must be 3x3.
```
chleh: Maybe then the matrix better should be 3x3? I guess it would be nice if `K2T(T2K(A)) == A`

Tests/MathLib/mechanics_invariants.cpp
```diff
@@ -0,0 +1,286 @@
...
+    m << v[0], v[3] / std::sqrt(2.), v[3] / std::sqrt(2.), v[1];
```
chleh: Minor: Should `std::sqrt(2)` be precomputed somewhere? Maybe it's constexpr, but some compilers might not know that...

Tests/MathLib/mechanics_invariants.cpp
```diff
@@ -0,0 +1,286 @@
...
+    Eigen::Matrix<double, 1, 1> v;
+    v << m[0];
+    return v;
+}
+
+template <>
+KelvinVector<4> tensorToKelvin(
+    Eigen::Matrix<double, KelvinToTensorSize<4>::value,
+                  KelvinToTensorSize<4>::value> const& m)
+{
```
chleh: Minor (for later move of the code): Shouldn't there be a check that `m` is symmetric?

Tests/MathLib/mechanics_invariants.cpp
```diff
@@ -0,0 +1,286 @@
...
+    // 0 is for the z-component, otherwise the matrix must be 3x3.
+    v << m(0, 0), m(1, 1), 0, m(0, 1) * std::sqrt(2.);
+    return v;
+}
+
+template <>
+KelvinVector<6> tensorToKelvin(
+    Eigen::Matrix<double, KelvinToTensorSize<6>::value,
+                  KelvinToTensorSize<6>::value> const& m)
+{
```
chleh: Same here: Symmetry check?

Tests/MathLib/mechanics_invariants.cpp
```diff
@@ -0,0 +1,286 @@
...
+//
+
+TEST_F(MathLibMechanicsInvariants1, SelfTestMappingKelvinToTensor)
+{
+    auto f = [](KelvinVector<1> const& v) {
+        return v == tensorToKelvin<1>(kelvinToTensor(v));
+    };
+
+    ac::check<KelvinVector<1>>(
+        f, 1000, ac::make_arbitrary(kelvinVectorGenerator), gtest_reporter);
```
chleh: How long do those tests take? If longer than a few (say three) seconds in a debug build, the number of repetitions should be reduced.

Tests/MathLib/mechanics_invariants.cpp
```diff
@@ -0,0 +1,286 @@
...
+}
+
+TEST_F(MathLibMechanicsInvariants4, Determinant)
+{
+    auto f = [](KelvinVector<4> const& v) {
+        return std::abs(Mechanics::Invariants<4>::determinant(v) -
+                        kelvinToTensor(v).determinant()) <=
+               1e-12 * std::pow(v.norm(), 4);
```
chleh: Are the vector norms bounded? Otherwise I'd rather generate vectors with bounded elements in order to have better feeling of what tolerance is effectively applied.

Tests/MathLib/mechanics_invariants.cpp
```diff
@@ -0,0 +1,286 @@
...
+        gtest_reporter);
+}
+
+TEST_F(MathLibMechanicsInvariants6, Determinant)
+{
+    auto f = [](KelvinVector<6> const& v) {
+        return std::abs(Mechanics::Invariants<6>::determinant(v) -
+                        kelvinToTensor(v).determinant()) <=
+               std::numeric_limits<double>::epsilon() *
+                   std::pow(v.norm(), 3.07);
```
chleh: x^3.07 -- One can really push it too hard. :smile:

Tests/MathLib/mechanics_invariants.cpp
```diff
@@ -0,0 +1,286 @@
...
+    auto f = [](KelvinVector<4> const& v) {
+        return (Mechanics::Invariants<4>::inverse(v) -
+                tensorToKelvin<4>(kelvinToTensor(v).inverse()))
+                   .norm() < 1e-10 * std::pow(v.norm(), 2.7);
+    };
+
+    auto eps = std::numeric_limits<double>::epsilon();
+    ac::check<KelvinVector<4>>(
+        f, 1000,
+        ac::make_arbitrary(ac::map(&setZcomponentZero, kelvinVectorGenerator))
```
chleh: If you set the z component to zero, isn't then the determinant of the Kelvin vector always zero, because then the corresponding 3x3 tensor should be
```
/ a b 0 \
| b c 0 |
\ 0 0 d /
```
with `d == 0`. Did I get something wrong?

Tests/MathLib/mechanics_invariants.cpp
```diff
@@ -0,0 +1,286 @@
...
+            .discard_if([&eps](KelvinVector<4> const& v) {
+                return (std::abs(kelvinToTensor(v).determinant()) <
```
chleh: Very minor: you could check the Kelvin determinant, can't you?

Tests/MathLib/mechanics_invariants.cpp
```diff
@@ -0,0 +1,286 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/LICENSE.txt
+ */
```
chleh: Filename is lowercase.

MathLib/mechanics_invariants.h
```diff
@@ -0,0 +1,166 @@
...
+}
+
+template <>
+inline Eigen::Matrix<double, 4, 1> Invariants<4>::inverse(
+    Eigen::Matrix<double, 4, 1> const& vec)
+{
+    Eigen::Matrix<double, 4, 1> inv;
+    inv(0) = vec(1);
+    inv(1) = vec(0);
+    inv(2) = 0;
```
chleh: The inverse ignores this component. That looks very fishy to me.

chleh:
Generally I'm OK with this PR.
Please don't rebase. I don't want to go over all of this again.

MathLib/mechanics_invariants.h
chleh: > (or would it be possible to only check this in debug mode?)

Yes, that's exactly what `assert(...)` is for. (So one should use whenever it is possible)

MathLib/mechanics_invariants.h
chleh: Then @endJunction should put `static_assert(KelvinVectorSize != 1)` here.

chleh:
Do the projection and identity Kelvin vectors need a test? (e.g., `P_dev * P_sph == P_sph * P_dev == 0; P_dev + P_sph == identity`)
How are products between Kelvin vectors computed?
Will there be other implementations of BMatrices, e.g., for the "usual" non-Kelvin tensor-to-vector mapping? (If so, the BMatrices should be renamed to BMatricesKelvin or similar.)

NumLib/Fem/Deformation/BMatrixPolicy.h
endJunction: @nagelt Please clarify this point.

MathLib/mechanics_invariants.h
endJunction: Yes, it is "hard"; For now all the material models have to be implemented using fixed size matrices.

NumLib/Fem/Deformation/LinearBMatrix.h
endJunction: There is a paper describing all the details of the BMatrices and Kelvin mapping;
@nagelt Would you please provide a citation?

The `std::sqrt(2)` (or similar simple expressions) are usually computed at the compile-time;
Using `static const double sqrt2 = std::sqrt(2)` in this particular function made the binary code larger by 22656 bytes. I didn't (yet) tested its performance. I shall keep it as it is now.

MathLib/mechanics_invariants.h
endJunction: `detail` ns not possible, because specialization must be in the same ns as the declaration.
Moving to impl (or even cpp) after fixing the other things to reduce confusion.

Tests/MathLib/AutoCheckTools.h
endJunction: It is a common pattern in the autocheck generators;
The autocheck is generating samples increasing the 'size' after every sample. So this argument can be used to generate e.g. longer and longer vectors/arrays/strings/magnitudes of numbers/etc. The 'size' passed to the `source` generator is "fixed", such that all the generated values are of the same magnitude.

Tests/MathLib/mechanics_invariants.cpp
endJunction: I'd prefer to keep it here, otherwise somebody could start doing funny things like
T2K(super_func(K2T(v))) potentially violating the requirement of the Kelvin mapping, symmetry or what not.

Tests/MathLib/mechanics_invariants.cpp
endJunction: `K2T(T2K(A)) == A` is tested below in the SelfTest...
Will be fixing this for 2D...

Tests/MathLib/mechanics_invariants.cpp
endJunction: Checked this. Better to keep like it is for sake of performance and clarity.

Tests/MathLib/mechanics_invariants.cpp
endJunction: In debug 30ms for selftest, 30ms for det, and 350ms for inv. for the size 6.
All ' MathLibMechanicsInvariants6.*' tests run in 422ms including testrunner startup/shutdown.

Tests/MathLib/mechanics_invariants.cpp
endJunction: The vector norms are bounded by the 'size'. e.g. for 1000 tests maximum doubles generated by the `source` in the randomEigenMatrixGenerator is +-1000.0. So maximum norm is 2000.
If necessary one can speedup the size grow or fix it...

Tests/MathLib/mechanics_invariants.cpp
endJunction: All correct. Will be fixing it for the 2D case. As of now I have only the abc components, d was ignored.

Tests/MathLib/mechanics_invariants.cpp
endJunction: Eigen is more reliable source. If my KelvinVector::determinant is wrong, the something can be wrong here, etc.

endJunction:
Not yet finished all the comments. Hold on!

Tests/MathLib/mechanics_invariants.cpp
chleh: Sounds good. I was only asking because some of the (I think IO-related) unit tests take "forever" in a debug build, which is pretty annoying.

Tests/MathLib/mechanics_invariants.cpp
chleh: If ||x|| = 2000, then ||x||^4 = 16.0e+12, which means that the tolerance you apply is 16.0, which is not very strict. Or am I getting it wrong?
If the norms were bounded, e.g. 1/2 < ||x|| < 3/2 or ||x|| < 1, the applied tolerance as computed above (`1e-12*||x||^4`) would be bounded, too.

MathLib/mechanics_invariants.h
chleh: I see.

MathLib/mechanics_invariants.h
chleh: I only asked to put `initP_dev` into the detail namespace. That should be possible.
But it's of minor importance.

Tests/MathLib/AutoCheckTools.h
chleh: Thanks for the explanation.

MathLib/mechanics_invariants.h
```diff
@@ -32,89 +32,87 @@ struct Invariants final
...
+        Eigen::Matrix<double, KelvinVectorSize, 1> const& v);
 
     /// Inverse of a matrix in Kelvin vector representation.
     static Eigen::Matrix<double, KelvinVectorSize, 1> inverse(
-        Eigen::Matrix<double, KelvinVectorSize, 1> const& vec);
+        Eigen::Matrix<double, KelvinVectorSize, 1> const& v);
 
     static double effectiveStress(
-        Eigen::Matrix<double, KelvinVectorSize, 1> const& deviatoric_stress)
+        Eigen::Matrix<double, KelvinVectorSize, 1> const& v)
```
chleh: here `deviatoric_stress` would be appropriate because `effectiveStress` is a stress-related method.

MathLib/mechanics_invariants.h
```diff
@@ -32,89 +32,87 @@ struct Invariants final
...
     {
-        return std::sqrt(3 * J2(deviatoric_stress));
+        return std::sqrt(3 * J2(v));
     }
 
-    static double J2(
-        Eigen::Matrix<double, KelvinVectorSize, 1> const& deviatoric_stress)
+    static double J2(Eigen::Matrix<double, KelvinVectorSize, 1> const& v)
```
chleh: Here and below it would be nice if the argument name reflected that a deviatoric tensor is expected.

NumLib/Fem/Deformation/BMatrixPolicy.h
```diff
@@ -61,6 +65,7 @@ struct KelvinMatrixTypeImpl<1>
     using type = Eigen::Matrix<double, 1, 1, Eigen::ColMajor>;
 };
 
+/// Kelvin vector type for given displacement dimension.
```
chleh: I think the documentation is wrong. A KelvinMatrix is a 4th order tensor which is mapped to a square matrix under the Kelvin mapping.

NumLib/Fem/Deformation/LinearBMatrix.h
wenqing: I see. Thanks for explanation.

NumLib/Fem/Deformation/LinearBMatrix.h
endJunction: `error: label at end of compound statement: expected statement` and w/o the `default` label there is warning about missing default case. Reverted to original with break.

chleh: OK.

MathLib/mechanics_invariants.h
```diff
@@ -0,0 +1,176 @@
...
+    /// Determinant of a matrix in Kelvin vector representation.
+    static double determinant(
+        Eigen::Matrix<double, KelvinVectorSize, 1> const& v);
+
+    /// Inverse of a matrix in Kelvin vector representation.
+    static Eigen::Matrix<double, KelvinVectorSize, 1> inverse(
+        Eigen::Matrix<double, KelvinVectorSize, 1> const& v);
+
+    /// The input vector must have trace equal zero.
+    static double effectiveStress(
```
norihiro-w: what is effective stress here?

NumLib/Fem/Deformation/BMatrixPolicy.h
```diff
@@ -0,0 +1,154 @@
...
+
+    // Dimensions of specific b-matrix for n-points and displacement
+    // dimension.
+    static int const _columns = ShapeFunction::NPOINTS * DisplacementDim;
+    static int const _rows = KelvinVectorDimensions<DisplacementDim>::value;
+
+public:
+    using StiffnessMatrixType = MatrixType<_columns, _columns>;
+
+    /// Rhs residual
```
norihiro-w: NumLib should be independent of physics

endJunction:
Not sure if the more general names e.g.: 

ResidualVectorType instead of NodalForceVectorType
LocalVectorType instead of StressVectorType
GlobalSystemMatrixType  instead of StiffnessMatrixType
LocalSystemMatrixType instead of ModulusMatrixType

would be more clear.

norihiro-w:
MathLib and NumLib should be independent of physics. and I don't think these new files need to/can be generalised to put them into the modules.  I suggest creating a new module (e.g. MechanicsLib) and put the new files there.

MathLib/mechanics_invariants.h
```diff
@@ -0,0 +1,189 @@
...
+    /// Kelvin mapping of deviatoric projection tensor. \f$A_{\rm dev} = P_{\rm
+    /// dev}:A\f$ for \f$A\f$ being a second order tensor.
+    static Eigen::Matrix<double, KelvinVectorSize, KelvinVectorSize> const
+        P_dev;
+    /// Kelvin mapping of spherical projection tensor. \f$A_{\rm sph} = P_{\rm
+    /// sph}:A\f$ for \f$A\f$ being a second order tensor.
+    static Eigen::Matrix<double, KelvinVectorSize, KelvinVectorSize> const
+        P_sph;
+    /// Kelvin mapping of 2nd order identity tensor.
+    static Eigen::Matrix<double, KelvinVectorSize, 1> const identity2;
```
norihiro-w: why not use upper case `Identity2`?

MathLib/mechanics_invariants.h
```diff
@@ -0,0 +1,189 @@
...
+
+    /// Determinant of a matrix in Kelvin vector representation.
+    static double determinant(
```
norihiro-w: is this function "invariant"?

MathLib/mechanics_invariants.h
```diff
@@ -0,0 +1,189 @@
...
+        Eigen::Matrix<double, KelvinVectorSize, 1> const& v);
+
+    /// Inverse of a matrix in Kelvin vector representation.
+    static Eigen::Matrix<double, KelvinVectorSize, 1> inverse(
```
norihiro-w: is this function "invariant"?

MathLib/mechanics_invariants.h
```diff
@@ -0,0 +1,189 @@
...
+        Eigen::Matrix<double, KelvinVectorSize, 1> const& v);
+
+    /// Von Mises equivalent stress.
```
norihiro-w: Von -> von

MathLib/mechanics_invariants.h
```diff
@@ -0,0 +1,189 @@
...
+    /// \note The input vector must have trace equal zero.
+    static double equivalentStress(
+        Eigen::Matrix<double, KelvinVectorSize, 1> const& deviatoric_v)
+    {
+        static_assert(KelvinVectorSize == 4 || KelvinVectorSize == 6,
```
norihiro-w: isn't it possible to write this assert only once in some function or a private constructor?

MathLib/mechanics_invariants.h
```diff
@@ -0,0 +1,189 @@
...
+template <>
+inline Eigen::Matrix<double, 4, 1> Invariants<4>::inverse(
+    Eigen::Matrix<double, 4, 1> const& v)
+{
+    Eigen::Matrix<double, 4, 1> inv;
+    inv(0) = v(1) * v(2);
+    inv(1) = v(0) * v(2);
+    inv(2) = v(0) * v(1) - v(3) * v(3) / 2.;
+    inv(3) = - v(3) * v(2);
+    return inv / determinant(v);
```
norihiro-w: i am not sure if ROV works with this expression. `inv /= determinant(v); return inv;` might be better.

MathLib/mechanics_invariants.h
```diff
@@ -0,0 +1,189 @@
...
+                      "KelvinVector invariants, trace: input vector of size "
+                      "different than 4 or 6 is not allowed.");
+        return v.template topLeftCorner<3, 1>().sum();
+    }
+};
+
+template <>
+inline double Invariants<6>::determinant(Eigen::Matrix<double, 6, 1> const& v)
+{
+    return v(0) * v(1) * v(2) + v(3) * v(4) * v(5) / std::sqrt(2.) -
```
norihiro-w: remark for the future optimization work: accessing raw double pointers might be faster than using Eigen `()` operator. need to check the implementation

NumLib/Fem/Deformation/BMatrixPolicy.h
```diff
@@ -0,0 +1,118 @@
...
+/// Special case for dimension one because of column major storage requirement
+/// from Eigen3.
+template <>
+struct KelvinMatrixTypeImpl<1>
+{
+    using type = Eigen::Matrix<double, 1, 1, Eigen::ColMajor>;
+};
+}  // namespace detail
+
+/// Kelvin vector type for given displacement dimension.
```
norihiro-w: vector -> matrix type

NumLib/Fem/Deformation/LinearBMatrix.h
```diff
@@ -0,0 +1,72 @@
...
+                   BMatrixType const& b_matrix, StrainVectorType& eps)
+{
+    eps.noalias() = b_matrix * local_displacement;
```
norihiro-w: do we need a function for this one line?

MathLib/mechanics_invariants.h
endJunction: Yes together with trace and some others. Better explained in the [wiki](https://en.wikipedia.org/wiki/Invariants_of_tensors) for example.

MathLib/mechanics_invariants.h
endJunction: I don't know now. It seems not, but it's just a feeling.

MathLib/mechanics_invariants.h
endJunction: The testrunner binary becomes larger in your case by 66464 bytes after changing both inverse() functions. Maybe I'll look into the assembly, but not now. Keeping it as it is.

MathLib/mechanics_invariants.h
endJunction: Using raw pointers is not an optimization: binary is identical. But readability is horrible.

From Eigen3 [doc](https://eigen.tuxfamily.org/dox-devel/group__QuickRefPage.html):
Coeff access with range checking	`vector(i)     vector.x() vector[i]     vector.y()           vector.z()               vector.w() matrix(i,j)`
Range checking is disabled if NDEBUG or EIGEN_NO_DEBUG is defined

norihiro-w: I was thinking about calculation of index because the vector is a matrix object. probably the index is calculated at compiling time. I don't think readability becomes horrible with raw pointer access (`v(0)` -> `a=v.data(); a[0]`)

MathLib/mechanics_invariants.h
norihiro-w: no idea why it increased...

MathLib/mechanics_invariants.h
```diff
@@ -33,7 +33,7 @@ struct Invariants final
     static Eigen::Matrix<double, KelvinVectorSize, KelvinVectorSize> const
         P_sph;
     /// Kelvin mapping of 2nd order identity tensor.
-    static Eigen::Matrix<double, KelvinVectorSize, 1> const identity2;
+    static Eigen::Matrix<double, KelvinVectorSize, 1> const Identity2;
```
chleh: If the capital `I` shall stress the constness, I'd suggest to have it all-caps. Otherwise I'd prefer to have data members lowercase and only types start with an uppercase letter.

MathLib/mechanics_invariants.h
chleh: The inverse of a matrix obviously changes when transforming coordinate axes. Invariants are (exclusively, I guess) scalars.

NumLib/Fem/Deformation/LinearBMatrix.h
chleh: I'd keep it. It doesn't hurt, I think.

MathLib/mechanics_invariants.h
endJunction: I see. With `a = v.data(); a[5]` readability is better than with naive version `*(v.data() + 5)`.

But the code depends on the storage order then, which is not nice.

MathLib/mechanics_invariants.h
endJunction: Will be moved in MaterialLib/SolidModels/KelvinVector.h as discussed. Generalization iff this would avoid duplication.

Tests/MathLib/AutoCheckTools.h
endJunction: Not now. If needed I'll add that possibility (to check for storage order independence of algorithms, for example).

Tests/MathLib/mechanics_invariants.cpp
endJunction: I'll add sharper bounds.

MathLib/mechanics_invariants.h
endJunction: Moving inverse out of `Invariants`. See :white_circle: todo in the description.

NumLib/Fem/Deformation/LinearBMatrix.h
endJunction: Removing

MathLib/mechanics_invariants.h
endJunction: Changed P_dev/sph to deviatoric_projection and spherical_projection. Going back to identity2. Now all identifier are "style-guided" but J2 and J3, which are difficult to name in short manner and are common used under this name in formulas.

endJunction:
All comments answered. Please have a look; I might have missed 1-2% of the comments.

endJunction:
Jenkins, test this please.

The following ctests [FAILED](https://svn.ufz.de:8443/job/OGS-6/job/Linux-PRs/2665/) for some reason but they don't affect the review for now:
	  4 - MapGeometryToMeshSurface-MapGeometryToMeshSurface_Bode-time-diff (Failed)
	  6 - MapGeometryToMeshSurface-MapGeometryToMeshSurface_Naegelstedt-time-diff (Failed)

ProcessLib/Deformation/BMatrixPolicy.h
```diff
@@ -0,0 +1,121 @@
...
+#include "NumLib/Fem/ShapeMatrixPolicy.h"
+
+namespace ProcessLib
+{
+/// Kelvin vector dimensions for given displacement dimension.
+template <int DisplacementDim>
+struct KelvinVectorDimensions;
+
+template <>
+struct KelvinVectorDimensions<1>
```
chleh: Can that be removed now?

ProcessLib/Deformation/BMatrixPolicy.h
```diff
@@ -0,0 +1,121 @@
...
+    using type =
+        Eigen::Matrix<double, KelvinVectorDimensions<DisplacementDim>::value,
+                      KelvinVectorDimensions<DisplacementDim>::value,
+                      Eigen::RowMajor>;
+};
+
+/// Special case for dimension one because of column major storage requirement
+/// from Eigen3.
+template <>
+struct KelvinMatrixTypeImpl<1>
```
chleh: Still needed?

ProcessLib/Deformation/LinearBMatrix.h
```diff
@@ -0,0 +1,63 @@
...
+namespace ProcessLib
+{
+namespace LinearBMatrix
+{
+/// Fills a B-matrix based on given shape function dN/dx values.
+template <int DisplacementDim, int NPOINTS, typename DNDX_Type,
+          typename BMatrixType>
+void computeBMatrix(DNDX_Type const& dNdx, BMatrixType& b_matrix)
+{
+    static_assert(0 < DisplacementDim && DisplacementDim <= 3,
```
chleh: now 1 < DisplacementDim?

ProcessLib/Deformation/LinearBMatrix.h
```diff
@@ -0,0 +1,63 @@
...
+        // no break for fallthrough.
+        case 2:
+            for (int i = 0; i < NPOINTS; ++i)
+            {
+                b_matrix(1, NPOINTS + i) = dNdx(1, i);
+                b_matrix(3, i) = dNdx(1, i) / std::sqrt(2);
+                b_matrix(3, NPOINTS + i) = dNdx(0, i) / std::sqrt(2);
+            }
+        // no break for fallthrough.
+        case 1:
```
chleh: no case 1?

Tests/MaterialLib/KelvinVector.cpp
```diff
@@ -0,0 +1,230 @@
...
+using KelvinVector = Eigen::Matrix<double, Size, 1, Eigen::ColMajor, Size, 1>;
+
+template <int Size>
+Eigen::Matrix<double, 3, 3> kelvinToTensor(KelvinVector<Size> const& v);
+
+template <>
+Eigen::Matrix<double, 3, 3> kelvinToTensor(KelvinVector<4> const& v)
+{
+    Eigen::Matrix<double, 3, 3> m;
+    // The z-component is ignored, otherwise the matrix must be 3x3.
```
chleh: Is the comment correct?

Tests/MaterialLib/KelvinVector.cpp
```diff
@@ -0,0 +1,230 @@
...
+    return m;
+}
+
+template <int Size>
+KelvinVector<Size> tensorToKelvin(Eigen::Matrix<double, 3, 3> const& m);
+
+template <>
+KelvinVector<4> tensorToKelvin(Eigen::Matrix<double, 3, 3> const& m)
+{
+    EXPECT_NEAR(m(0, 1), m(1, 0), std::numeric_limits<double>::epsilon());
```
chleh: Optional: `m(0,2)==m(1,2)==m(2,0)==m(2,1)==0`

Tests/MaterialLib/KelvinVector.cpp
```diff
@@ -0,0 +1,230 @@
...
+        return error < 1e-7 && 1e-10 * std::pow(v.norm(), 1.2);
+    };
+
+    auto eps = std::numeric_limits<double>::epsilon();
+    ac::check<KelvinVector<4>>(
+        f, 1000,
+        ac::make_arbitrary(kelvinVectorGenerator)
+            .discard_if([&eps](KelvinVector<4> const& v) {
+                return (std::abs(kelvinToTensor(v).determinant()) == 0.0 ||
+                        v(3) == 0.0);  // only invertable matrices
```
chleh: `== 0.0` or better `< eps` ?

Tests/MaterialLib/KelvinVector.cpp
```diff
@@ -0,0 +1,230 @@
...
+        return error < 1e-8 && 1e-10 * std::pow(v.norm(), 1.2);
+    };
+
+    auto eps = std::numeric_limits<double>::epsilon();
+    ac::check<KelvinVector<6>>(
+        f, 1000,
+        ac::make_arbitrary(kelvinVectorGenerator)
+            .discard_if([&eps](KelvinVector<6> const& v) {
+                return (std::abs(kelvinToTensor(v).determinant()) == 0.0 ||
+                        v(3) == 0.0);  // only invertable matrices
```
chleh: Likewise.

Tests/MaterialLib/KelvinVector.cpp
```diff
@@ -0,0 +1,230 @@
...
+    auto const& P_dev = Invariants<size>::deviatoric_projection;
+    auto const& P_sph = Invariants<size>::spherical_projection;
+
+    // Test product P_dev * P_sph is zero.
+    auto const P_dev_P_sph = P_dev * P_sph;
+    EXPECT_LT(P_dev_P_sph.norm(), std::numeric_limits<double>::epsilon());
+    EXPECT_LT(P_dev_P_sph.maxCoeff(), std::numeric_limits<double>::epsilon());
+
+    // Test product P_sph * P_dev is zero.
+    auto const P_sph_P_dev = P_sph * P_dev;
```
chleh: auto and Eigen!

MaterialLib/SolidModels/KelvinVector.h
```diff
@@ -0,0 +1,84 @@
...
+{
+    static_assert(KelvinVectorSize == 4 || KelvinVectorSize == 6,
+                  "KelvinVector invariants for vectors of size different than "
+                  "4 or 6 is not allowed.");
+    /// Kelvin mapping of deviatoric projection tensor. \f$A_{\rm dev} = P_{\rm
+    /// dev}:A\f$ for \f$A\f$ being a second order tensor.
+    static Eigen::Matrix<double, KelvinVectorSize, KelvinVectorSize> const
+        deviatoric_projection;
+    /// Kelvin mapping of spherical projection tensor. \f$A_{\rm sph} = P_{\rm
+    /// sph}:A\f$ for \f$A\f$ being a second order tensor.
```
chleh: bad line break. Do the docs look OK?

chleh:
Only minor additional comments from my side. Basically:
* [ ] Eigen and auto
* [ ] `== 0.0` vs. `< eps`
* [ ] some remaining displacementdim 1 code.

Afterwards :+1:

Tests/MaterialLib/KelvinVector.cpp
endJunction: No, == 0 is more strict.

MaterialLib/SolidModels/KelvinVector.h
endJunction: Yes, looks good!

MaterialLib/SolidModels/KelvinVector-impl.h
```diff
@@ -0,0 +1,106 @@
...
+namespace detail
+{
+template <int KelvinVectorSize>
+Eigen::Matrix<double, KelvinVectorSize, KelvinVectorSize>
+initDeviatoricProjection()
+{
+    Eigen::Matrix<double, KelvinVectorSize, KelvinVectorSize> P_dev =
+        Eigen::Matrix<double, KelvinVectorSize, KelvinVectorSize>::Identity();
+
+    P_dev.template topLeftCorner<3, 3>() -=
```
norihiro-w: unsure: noalias() may influence optimization

MaterialLib/SolidModels/KelvinVector.cpp
```diff
@@ -0,0 +1,57 @@
...
+template <>
+Eigen::Matrix<double, 4, 1, Eigen::ColMajor, 4, 1> inverse(
+    Eigen::Matrix<double, 4, 1, Eigen::ColMajor, 4, 1> const& v)
+{
+    Eigen::Matrix<double, 4, 1, Eigen::ColMajor, 4, 1> inv;
+    inv(0) = v(1) * v(2);
+    inv(1) = v(0) * v(2);
+    inv(2) = v(0) * v(1) - v(3) * v(3) / 2.;
+    inv(3) = -v(3) * v(2);
+    return inv / Invariants<4>::determinant(v);
```
norihiro-w: no need to check if determinat() returns zero

norihiro-w: ?

ProcessLib/Deformation/BMatrixPolicy.h
```diff
@@ -0,0 +1,94 @@
...
+
+#ifndef PROCESSLIB_FEM_BMATRIXPOLICY_H_
+#define PROCESSLIB_FEM_BMATRIXPOLICY_H_
+
+#include "NumLib/Fem/ShapeMatrixPolicy.h"
+
+namespace ProcessLib
+{
+/// Kelvin vector dimensions for given displacement dimension.
+template <int DisplacementDim>
```
norihiro-w: Displacement -> Spatial

norihiro-w: sorry nowi got what it means. please forget the comment above

norihiro-w:
looks good, besides minor comments

endJunction:
Can I rebase and merge, or is there more review to be expected?

The only test failing (again) is
```
[ RUN      ] NumLib.ComponentNormMultiComponent2
/home/travis/build/ufz/ogs/Tests/NumLib/TestComponentNorms.cpp:94: Failure
The difference between total_norm and compwise_total_norm is 8.8817841970012523e-15, which exceeds tolerance, where
total_norm evaluates to 14.747408246182694,
compwise_total_norm evaluates to 14.747408246182685, and
tolerance evaluates to 6.6613381477509392e-15.
[  FAILED  ] NumLib.ComponentNormMultiComponent2 (72 ms)
```

chleh:
I'm fine with this PR. :fast_forward:

endJunction:
Jenkins, test this please!

endJunction:
Jenkins, test this please!

endJunction:
It could happen that the error margins in the unit tests are too tight. In that case I'll increase the tolerance a little.

PR #1361, opened 2016-08-18
add a CLI tool flipNodeCoordinates

norihiro-w:
Adding a CLI tool named `flipNodeCoordinates` to flip node coordinates, e.g. XY to XZ plane.

### DIFF
--- a/Applications/Utils/SimpleMeshCreation/generateStructuredMesh.cpp
+++ b/Applications/Utils/SimpleMeshCreation/generateStructuredMesh.cpp
@@ -46,6 +46,7 @@ unsigned getDimension(MeshLib::MeshElemType eleType)
         case MeshLib::MeshElemType::PYRAMID:
         case MeshLib::MeshElemType::TETRAHEDRON:
             return 3;
+        case MeshLib::MeshElemType::POINT:
         case MeshLib::MeshElemType::INVALID:
             return 0;
     }
--- a/Applications/Utils/MeshEdit/CMakeLists.txt
+++ b/Applications/Utils/MeshEdit/CMakeLists.txt
@@ -77,6 +77,10 @@ add_executable(ExtractSurface ExtractSurface.cpp)
 target_link_libraries(ExtractSurface MeshLib)
 set_target_properties(ExtractSurface PROPERTIES FOLDER Utilities)
 
+add_executable(swapNodeCoordinateAxes swapNodeCoordinateAxes.cpp)
+target_link_libraries(swapNodeCoordinateAxes MeshLib)
+set_target_properties(swapNodeCoordinateAxes PROPERTIES FOLDER Utilities)
+
 ####################
 ### Installation ###
 ####################
@@ -96,5 +100,6 @@ install(TARGETS
     ResetPropertiesInPolygonalRegion
     reviseMesh
     queryMesh
+    swapNodeCoordinateAxes
     RUNTIME DESTINATION bin COMPONENT Utilities
 )
--- a/Applications/Utils/MeshEdit/swapNodeCoordinateAxes.cpp
+++ b/Applications/Utils/MeshEdit/swapNodeCoordinateAxes.cpp
@@ -1,0 +1,111 @@
+/**
+ * @copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/LICENSE.txt
+ *
+ */
+
+#include <array>
+#include <memory>
+#include <string>
+
+#include <tclap/CmdLine.h>
+
+#include "Applications/ApplicationsLib/LogogSetup.h"
+
+#include "BaseLib/BuildInfo.h"
+
+#include "MeshLib/Mesh.h"
+#include "MeshLib/Node.h"
+
+#include "MeshLib/IO/readMeshFromFile.h"
+#include "MeshLib/IO/writeMeshToFile.h"
+
+static void swapNodeCoordinateAxes(MeshLib::Mesh &mesh, std::array<int, 3> const& new_axes_indices)
+{
+    for (MeshLib::Node* node : mesh.getNodes())
+    {
+        for (int i=0; i<3; i++)
+            (*node)[i] = (*node)[new_axes_indices[i]];
+    }
+}
+
+static bool parseNewOrder(std::string const& str_order, std::array<int, 3> &new_axes_indices)
+{
+    if (str_order.length()!=3)
+    {
+        ERR("Invalid argument for the new order. The argument should contain three characters.");
+        return false;
+    }
+
+    for (std::size_t i=0; i<new_axes_indices.size(); i++)
+        new_axes_indices[i] = -1;
+
+    for (int i=0; i<3; i++)
+    {
+        if (str_order[i] == 'x')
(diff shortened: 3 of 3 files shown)


Applications/Utils/MeshEdit/flipNodeCoordinates.cpp
```diff
@@ -0,0 +1,106 @@
...
+int main(int argc, char *argv[])
+{
+    ApplicationsLib::LogogSetup logog_setup;
+
+    TCLAP::CmdLine cmd("Flips node coordinates", ' ', BaseLib::BuildInfo::git_describe);
+    TCLAP::ValueArg<std::string> input_arg("i", "input-mesh-file","input mesh file",true,"","string");
+    cmd.add( input_arg );
+    TCLAP::ValueArg<std::string> output_arg("o", "output-mesh-file","output mesh file",true,"","string");
+    cmd.add( output_arg );
+    TCLAP::ValueArg<std::string> new_order_arg("n", "new-coords-order", "the new order of coordinates (e.g. xzy for flipping XY plane to XZ plane)", true, "", "string");
```
endJunction: I have difficulties to understand the example. Does it mean x->x, y->z, and z->y conversions?

Maybe the "flipping" is wrong. By "Flips node coordinates" do you mean reflection through a plane? But then what is flipping a plane to another plane? Rotation?
...
Can be discussed verbally for efficiency...

Applications/Utils/MeshEdit/flipNodeCoordinates.cpp
```diff
@@ -0,0 +1,106 @@
...
+            new_coord_ids[i] = 1;
+        else if (str_order[i] == 'z')
+            new_coord_ids[i] = 2;
+        else {
+            ERR("Invalid argument for for the new order. The  given argument contains a character other than x, y, z.");
+            return false;
+        }
+    }
+
+    bool marked[3] = {false};
```
endJunction: Not sure, if all elements of `marked` are set `false` or only the first one...

Applications/Utils/MeshEdit/flipNodeCoordinates.cpp
```diff
@@ -0,0 +1,106 @@
...
+#include "MeshLib/IO/writeMeshToFile.h"
+
+void flipNodeCoordinates(MeshLib::Mesh &mesh, std::array<int, 3> const& new_coord_ids)
+{
+    double new_xyz[3] = {};
+    for (MeshLib::Node* node : mesh.getNodes())
+    {
+        for (int i=0; i<3; i++)
+            new_xyz[i] = (*node)[new_coord_ids[i]];
+        node->updateCoordinates(new_xyz[0], new_xyz[1], new_xyz[2]);
```
endJunction: Maybe instead of modifying the original nodes it would be better to create a new vector of `Node*` with same order as original, but modified coordinates and then use Mesh's ctor:
```c++
Mesh{mesh.getName(), new_nodes, mesh.getElements()};
```
This would remove the undesired "friendness" of an exterior function.

Another possibility would be to add a `Mesh::map_nodes(std::function&&)`  function, which would do the mapping internally.

Exposing internals creates unnecessary dependencies.

Applications/Utils/MeshEdit/flipNodeCoordinates.cpp
norihiro-w: - yes. the given order is used to exchange coordinate values in the new mesh.
- probably "flipping" is wrong. I didn't have a better word for it. "Rotation" cannot be used because this tool will not support rotation with arbitrary angles. how about "swap" or "exchange"?

Applications/Utils/MeshEdit/flipNodeCoordinates.cpp
norihiro-w: this sets all entries

Applications/Utils/MeshEdit/flipNodeCoordinates.cpp
norihiro-w: you are right, exterior functions shouldn't be declared as friends. I'll change it to use `Mesh::map_nodes(std::function&&)`.

TomFischer:
Only a remark: Since  coordinate swaping is a special rotation you could also use `GeoLib::rotatePoints` from `AnalyticalGeometry.h`.

Applications/Utils/MeshEdit/flipNodeCoordinates.cpp
```diff
@@ -0,0 +1,106 @@
...
+    }
+}
+
+static bool parseOrderString(std::string const& str_order, std::array<int, 3> &new_coord_ids)
+{
+    if (str_order.length()!=3)
+    {
+        ERR("Invalid argument for for the new order. The argument should contain three characters.");
```
norihiro-w: for for

Applications/Utils/MeshEdit/flipNodeCoordinates.cpp
chleh: Maybe `exchanceCoordinateAxes`?

chleh:
@TomFischer: I think swapping axes is much easier than rotating points, and it guarantees that coordinates are not changed.

chleh:
When swapping XYZ -> XZY this is not a rotation, so one changes clockwise to counter-clockwie orientation. Is that a wanted feature? Should there be a warning for that?

chleh:
In general the PR looks good. After questions :+1:

norihiro-w:
> When swapping XYZ -> XZY this is not a rotation, so one changes clockwise to counter-clockwie orientation. Is that a wanted feature? Should there be a warning for that?

could you exlpain more details? I'm not getting "one changes clockwise to counter-clockwie orientation"

chleh:
The sketched quad would change its orientation upon the transformation XYZ → YXZ. For the quad that might not be a problem, but would it maybe break 3D elements?
```
XYZ, Z=0, nodes clockwise | YXZ, Z=0, nodes counter-clockwise 
                          |
        1 --> 2           |         1 <-- 4
        ^     |           |         |     ^
        |     v           |         v     |
        4 <-- 3           |         2 --> 3
```

norihiro-w:
@chleh you are right. I'll add warning

Applications/Utils/MeshEdit/flipNodeCoordinates.cpp
norihiro-w: renamed to `swapNodeCoordinateAxes()`, because "swap" is easier to type..

Applications/Utils/MeshEdit/flipNodeCoordinates.cpp
norihiro-w: I wrote a `for` loop in the tool source file instead of adding `map_nodes()`.

norihiro-w:
I'll cleanup the commits before merge

Applications/Utils/SimpleMeshCreation/generateStructuredMesh.cpp
```diff
@@ -46,6 +46,7 @@ unsigned getDimension(MeshLib::MeshElemType eleType)
         case MeshLib::MeshElemType::PYRAMID:
         case MeshLib::MeshElemType::TETRAHEDRON:
             return 3;
+        case MeshLib::MeshElemType::POINT:
```
TomFischer: Very good!

TomFischer:
:+1:

chleh:
:+1:

norihiro-w:
jenkins, test this please

PR #1365, opened 2016-08-19
Material properties for fluid and porous media

wenqing:
In this PR, classes for material properties for fluid and porous media are presented. The classes provide generic interfaces for different material property models. Several temperature or pressure dependent fluid density and viscosity models are presented. The implementations are verified by unit tests. The first application of these classes is liquid flow process.

PR #1366, opened 2016-08-19
Material properties for liquid flow process

wenqing:
Based on #1365 , this PR applies the material properties to create the material data for liquid flow process. A unit test about XML structure of the material data is provided (in commit 1df4cdb).

Its application can be found in  https://github.com/wenqing/ogs/blob/liquid_pcs/ProcessLib/LiquidFlow/LiquidFlowLocalAssembler-imp.h#L77  (https://github.com/wenqing/ogs.git, branch liquid_pcs)


MaterialLib/PorousMedium/Permeability/createPermeabilityModel.cpp
```diff
@@ -0,0 +1,64 @@
...
+namespace MaterialLib
+{
+namespace PorousMedium
+{
+CoefMatrix createPermeabilityModel(BaseLib::ConfigTree const* const config)
+{
+    //! \ogs_file_param{material__porous_medium__permeability__values}
+    auto const val_strs = config->getConfigParameter<std::string>("values");
+    std::vector<double> values =
+        BaseLib::convertStringValues2Values<double>(val_strs);
```
chleh: This is not needed. ConfigTree can read vectors of space separated values. (just use `<std::vector<double>>` instead of `<std::string>` above.

MaterialLib/Fluid/Density/createFluidDensityModel.cpp
```diff
@@ -0,0 +1,56 @@
...
+namespace MaterialLib
+{
+namespace Fluid
+{
+FluidProperty* createFluidDensityModel(BaseLib::ConfigTree const* const config)
+{
+    //! \ogs_file_param{material__fluid__density__type}
+    auto const type = config->getConfigParameter<std::string>("type");
+
+    if (type.find("constant") != std::string::npos)
```
chleh: I think `type == "constant"` is better, because `type.find()` also finds substrings of `type`

MaterialLib/Fluid/Density/IdealGasLaw.h
```diff
@@ -0,0 +1,99 @@
...
+
+    /// Get the partial differential of density with the respect to pressure
+    /// \param T  Temperature in K.
+    /// \param pg Gas phase pressure in Pa.
+    double dIdealGasLaw_dp(const double T, const double /* pg */) const
+    {
+        return _molar_mass / (PhysicalConstant::IdealGasConstant * T);
+    }
+
+    typedef double (IdealGasLaw::*ptr2_derivative_f)(const double,
```
chleh: I'd prefer typedefs being camelcase and to start with an uppercase letter.

MaterialLib/Fluid/Density/IdealGasLaw.h
```diff
@@ -0,0 +1,99 @@
...
+                                                     const double) const;
+
+    /// An array of pointer to derivative functions.
+    ptr2_derivative_f _derivative_functions[2];
```
chleh: could be static

PR #1367, opened 2016-08-22
Enabled Eigen preconditioner configuration

chleh:
Until now the setting in the prj file was not strictly parsed and ignored later on.

### DIFF
--- a/MathLib/LinAlg/Eigen/EigenLinearSolver.cpp
+++ b/MathLib/LinAlg/Eigen/EigenLinearSolver.cpp
@@ -102,6 +102,53 @@ private:
     T_SOLVER _solver;
 };
 
+template <template <typename, typename> class Solver, typename Precon>
+std::unique_ptr<EigenLinearSolverBase> createIterativeSolver()
+{
+    using Slv = EigenIterativeLinearSolver<
+        Solver<EigenMatrix::RawMatrixType, Precon>>;
+    return std::unique_ptr<Slv>(new Slv);
+}
+
+template <template <typename, typename> class Solver>
+std::unique_ptr<EigenLinearSolverBase> createIterativeSolver(
+    EigenOption::PreconType precon_type)
+{
+    switch (precon_type) {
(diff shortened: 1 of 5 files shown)


norihiro-w:
looks good

chleh:
Forgot to mention: The prj file for our ctests have to change: ufz/ogs-data#21

MathLib/LinAlg/Eigen/EigenLinearSolver.cpp
```diff
@@ -115,25 +162,16 @@ EigenLinearSolver::EigenLinearSolver(
...
-        break;
-    }
-    case EigenOption::SolverType::CG: {
-        using SolverType = Eigen::ConjugateGradient<Matrix, Eigen::Lower, Eigen::DiagonalPreconditioner<double>>;
-        _solver.reset(new details::EigenIterativeLinearSolver<SolverType>);
-        break;
-    }
-    case EigenOption::SolverType::INVALID:
-        OGS_FATAL("Invalid Eigen linear solver type. Aborting.");
+    switch (_option.solver_type) {
```
endJunction: Switch with two cases is an if-else...
Is the todo comment above still correct? I mean there is (non-obvious) OGS_FATAL in createIterativeSolver...

endJunction:
I'm not into Eigen3 now, but I think for SparseLU there was no preconditioner, only for iterative solvers.
Ignore this comment if it makes no sense.

endJunction:
:fast_forward:

MathLib/LinAlg/Eigen/EigenLinearSolver.cpp
chleh: Fixed the switch.
The todo is still valid: The default solver is set in EigenOption.cpp; pretty hidden, I think.

chleh:
>  I think for SparseLU there was no preconditioner

For SparseLU the preconditioner setting is ignored.

chleh:
Failing tests not related to this PR. Going to merge.

PR #1368, opened 2016-08-23
[MGTL] Deep copy of seg to avoid modifying orig. pnts.

TomFischer:
Fixes a bug for LineSegments that are complete outside of the mesh area.


chleh:
:fast_forward:

norihiro-w:
👍

PR #1369, opened 2016-08-24
[GL] Enlarge delta for grid meta data computation.

TomFischer:
As titled.

### DIFF
--- a/GeoLib/Grid.h
+++ b/GeoLib/Grid.h
@@ -201,6 +201,10 @@ Grid<POINT>::Grid(InputIterator first, InputIterator last,
                                     _max_pnt[1] - _min_pnt[1],
                                     _max_pnt[2] - _min_pnt[2]}};
 
+    // enlarge delta
+    for (auto & d : delta)
+        d = std::nextafter(d, std::numeric_limits<double>::max());
+
     assert(n_pnts > 0);
     initNumberOfSteps(max_num_per_grid_cell, static_cast<std::size_t>(n_pnts), delta);
 


TomFischer:
Jenkins, test this please.

TomFischer:
Jenkins, test this please.

GeoLib/Grid.h
```diff
@@ -201,6 +201,10 @@ Grid<POINT>::Grid(InputIterator first, InputIterator last,
                                     _max_pnt[1] - _min_pnt[1],
                                     _max_pnt[2] - _min_pnt[2]}};
 
+    // enlarge delta
+    for (auto & d : delta)
+        d = std::nextafter(d, std::numeric_limits<double>::max());
```
chleh: `delta` is the space diagonal of the bbox of the grid? Why do you have to extend it such a tiny bit?

TomFischer: Yes, `delta` is the space diagonal of the bounding box and also of the grid. The bounding box itself is extended already such a tiny bit during its construction. But when `delta` is computed from the bounding box the tiny extension of the bounding box could get lost because of numerical precision. So here the `delta` is enlarged again such that all points fits into the grid.

endJunction:
:fast_forward:

GeoLib/Grid.h
chleh: I'm not really happy with this. I think the result of `std::nextafter` can be cancelled out by very many arithmetic operations (e.g. adding or subtracting a large number). I think multiplying with `1+eps` would be more robust, but I don't know if that has any implications for our current code.
So if nobody else objects and the solution works for you: :fast_forward:

TomFischer: > I think the result of std::nextafter can be cancelled out by very many arithmetic operations (e.g. adding or subtracting a large number). 

In general you re right. The bounding box is an half open interval. As you pointed out already out the arithmetic operations computing `delta` cancel out the results of `std::nextafter` in the bounding box. For this reason I reenlarge `delta`.

 > I think multiplying with 1+eps would be more robust ...

Sometimes all points have the same z coordinate, i.e., the `delta[2] == 0`. So the half open interval has to be enlarged. In this case the multiplication with `1+eps` does not lead to the desired solution.

PR #1370, opened 2016-08-24
Started writing some docu.

chleh:
I added some missing comments, did a small fix of a docu script and wrote the documentation for the gml file.

If you have some additions or improvements to the documentation, please directly use Github's online editor.

I know it's only documentation, but since probably nobody will ever go over it again, it probably makes sense to review it anyway.


Documentation/ProjectFile/gml/surfaces/surface/element/i_element.md
```diff
@@ -1 +1,3 @@
-\ogs_missing_documentation
+Defines a triangle by its three vertices.
+The face normal of the triangle is determined such that the vertices are
+located counter-clockwise around it (right-hand rule).
```
chleh: Question to reviewers: Does the right-hand rule thing hold true for all of our code?

Documentation/ProjectFile/gml/t_name.md
```diff
@@ -1 +1,4 @@
-\ogs_missing_documentation
+Name of the geometrical set, i.e., the entirety of all geometries defined in
+this file.
+Each individual geometry additionally has an own name, which has to be unique in
+the geometrical set.
```
TomFischer: - Each polyline or surface has an own name, point can be given without names.
- Good point to mention the uniqueness of the names!

Documentation/ProjectFile/gml/surfaces/surface/element/i_element.md
TomFischer: As far as I aware there isn't any precondition concerning the point ordering of the triangle. I am even unsure if the ordering have to be consistent within a surface. Maybe @rinkk knows it better?!

TomFischer: Here are the results of a short conversion with @rinkk:
- The node ordering of the surface triangles haven't to be consistent within a surface.
- The triangles of a surface haven't to be connected necessarily.

chleh:
I just noticed that I accidentally force-pushed to this branch. @endJunction, @TomFischer please have a short look if I restored your changes correctly.
Directly writing in Github's online editor together with rebasing is obviously not a good idea.

endJunction:
Looks similar to me.

PR #1372, opened 2016-08-24
Unique data type for integer valued properties

TomFischer:
It seems that the handling within the Vtk io routines of the data type `std::size_t` depends on the operating system. For that reason I switched to the more common `int` data type for the `PropertyVector`. This works under arch linux as well as under windows (as reported by Eunseon).

### DIFF
--- a/Applications/Utils/ModelPreparation/ComputeNodeAreasFromSurfaceMesh.cpp
+++ b/Applications/Utils/ModelPreparation/ComputeNodeAreasFromSurfaceMesh.cpp
@@ -83,13 +83,13 @@ int main (int argc, char* argv[])
     // ToDo check if mesh is read correct and if the mesh is a surface mesh
 
     // check if a node property containing the subsurface ids is available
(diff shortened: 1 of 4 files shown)


chleh:
Is it because `std::size_t` has different size on different OSes? If so: `sizeof(int)` also varies; maybe better use a type with a guaranteed size overall platforms?

TomFischer:
> Is it because std::size_t has different size on different OSes?

This isn't the reason. All tools are running on the same OS. The `ExtractSurface` tool wrote a `PropertyVector<std::size_t>` *OriginalSubsurfaceNodeIDs* to the output mesh. Under windows the next tool in the workflow reads this surface mesh but could not read the *OriginalSubsurfaceNodeIDs*. If I take a look into the vtu mesh file generated by `ExtractSurface` I can see the data array for the *OriginalSubsurfaceNodeIDs* (type UInt64). Maybe the vtu io routines under windows interprets UInt64 as an other data type than `std::size_t`.

TomFischer:
Closed since together with @chleh a better solution was developed, see PR #1374 .

PR #1374, opened 2016-08-25
Read and write data arrays of type std::size_t under linux and windows in the vtu io.

TomFischer:
This makes it possible to read and write std::size_t under linux and windows.


chleh:
:fast_forward:

PR #1375, opened 2016-08-26
Docu part 3

chleh:
Documented
* curves
* process variables
* secondary variables

### DIFF
--- a/Documentation/ProjectFile/prj/curves/i_curves.md
+++ b/Documentation/ProjectFile/prj/curves/i_curves.md
@@ -1,4 +1,5 @@
 Contains curve definitions.
 
 Curves are scalar valued functions of a scalar argument.
-They can be used, e.g., to define time-dependent quantities.
+They can be used, e.g., to define time-dependent quantities, or simple
+relationships of scalar quantities as viscosity from temperature.
--- a/Documentation/ProjectFile/prj/curves/i_curves.md
+++ b/Documentation/ProjectFile/prj/curves/i_curves.md
@@ -1,0 +1,4 @@
+Contains curve definitions.
+
+Curves are scalar valued functions of a scalar argument.
+They can be used, e.g., to define time-dependent quantities.
--- a/Documentation/ProjectFile/prj/process_variables/process_variable/t_initial_condition.md
+++ b/Documentation/ProjectFile/prj/process_variables/process_variable/t_initial_condition.md
@@ -1,1 +1,3 @@
-\ogs_missing_documentation
+Initial condition for the process variable.
+
+This setting references a \ref ogs_file_param__parameter "parameter" by name.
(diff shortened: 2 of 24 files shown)


Documentation/ProjectFile/prj/process_variables/process_variable/t_initial_condition.md
```diff
@@ -1 +1,3 @@
-\ogs_missing_documentation
+Initial condition for the process variable.
+
+This setting references a \ref ogs_file_param__parameter "parameter" by name.
```
TomFischer: It is used as an initial solution of the linear solver... non-linear too?

Documentation/ProjectFile/prj/curves/i_curves.md
```diff
@@ -0,0 +1,4 @@
+Contains curve definitions.
+
+Curves are scalar valued functions of a scalar argument.
+They can be used, e.g., to define time-dependent quantities.
```
TomFischer: > e.g., to define time-dependent quantities

, or simple relationships of scalar quantities as viscosity from temperature.

Documentation/ProjectFile/process/secondary_variables/secondary_variable/a_type.md
```diff
@@ -1 +1,6 @@
-\ogs_missing_documentation
+The type of the secondary variable.
+
+Can be <tt>static</tt> for secondary variables that don't need any further
+information from the project file in order to be computer;
```
TomFischer: typo. compute**d** ?

Documentation/ProjectFile/prj/curves/i_curves.md
chleh: I can add it, but after that, IMHO the list shouldn't be extended further.

Documentation/ProjectFile/prj/process_variables/process_variable/t_initial_condition.md
chleh: > initial solution of the linear solver

At least with LIS you can disable that. (I don't know, if you can really disable it in OGS because of our default LIS option); and for the Newton-Raphson method the statement is not true.

For the nonlinear solver IMHO it is natural that the IC is really used as the initial value. Otherwise you would assemble a (possibly) completely meaningless equation system.

TomFischer:
After few fixes:  :+1:

endJunction:
:fast_forward:

Documentation/ProjectFile/prj/process_variables/process_variable/t_initial_condition.md
chleh: I added a few words, but I didn't mention the linear solver. I think that belongs better to the time loop description.

PR #1378, opened 2016-08-28
Multi component ConstantParameter and Parameter access operators.

endJunction:
- Extend the ConstantParameter to read `<values>`. The `<value>` tag remains for scalar parameters.
   This is needed for the [Mechancis PR](https://github.com/ufz/ogs/pull/1340) where the initial conditions are multi-component, constants. Before we had multi-component InitialCondition, which is replaced by Parameter.
 - ~~Also add two Parameter::operator() for scalar and component-wise value access.~~
  Renames `getTuple(t,x)` into call operator. This is for convenience: `rho(t, x)[0]` for scalar parameter rho, or `K(t,x)[i]` for three-dim conductivity/permeability parameter K (accessing the i-th component) is much nicer/less noisier than `rho.getTuple(t,x)[0]` or `K.getTuple(t,x)[i]`.

### DIFF
--- a/ProcessLib/Parameter/ConstantParameter.h
+++ b/ProcessLib/Parameter/ConstantParameter.h
@@ -16,11 +16,20 @@ namespace ProcessLib
 {
 /// Single, constant value parameter.
 template <typename T>
-struct ConstantParameter final : public Parameter<T> {
+struct ConstantParameter final : public Parameter<T>
+{
+    /// Construction with single value.
     ConstantParameter(T const& value) : _value({value}) {}
-
-    // TODO allow for different sizes
-    unsigned getNumberOfComponents() const override { return 1; }
+    /// Construction with a tuple.
+    /// The given tuple must be non-empty.
+    ConstantParameter(std::vector<T> const& value) : _value(value)
+    {
+        assert(!value.empty());
+    }
+    unsigned getNumberOfComponents() const override
+    {
+        return static_cast<unsigned>(_value.size());
+    }
 
     std::vector<T> const& getTuple(
         double const /*t*/, SpatialPosition const& /*pos*/) const override
--- a/ProcessLib/Parameter/ConstantParameter.h
+++ b/ProcessLib/Parameter/ConstantParameter.h
@@ -31,7 +31,7 @@ struct ConstantParameter final : public Parameter<T>
         return static_cast<unsigned>(_value.size());
     }
 
-    std::vector<T> const& getTuple(
+    std::vector<T> const& operator()(
         double const /*t*/, SpatialPosition const& /*pos*/) const override
     {
         return _value;
--- a/ProcessLib/Parameter/Parameter.h
+++ b/ProcessLib/Parameter/Parameter.h
@@ -52,7 +52,7 @@ struct Parameter : public ParameterBase
     virtual unsigned getNumberOfComponents() const = 0;
 
     //! Returns the parameter value at the given time and position.
-    virtual std::vector<T> const& getTuple(
+    virtual std::vector<T> const& operator()(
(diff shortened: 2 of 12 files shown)


TomFischer:
:+1:

ProcessLib/Parameter/Parameter.h
```diff
@@ -54,6 +54,24 @@ struct Parameter : public ParameterBase
     //! Returns the parameter value at the given time and position.
     virtual std::vector<T> const& getTuple(
         double const t, SpatialPosition const& pos) const = 0;
+
+    //! Scalar value parameter access. The number of components of this
+    //! parameter must be 1.
+    double operator()(double const t, SpatialPosition const& x) const
```
chleh: As I said, I am against `operator()`. This looks nice, but completely hides your intentions in code using this method.
Please rather use ordinary methods.

ProcessLib/Parameter/Parameter.h
```diff
@@ -54,6 +54,24 @@ struct Parameter : public ParameterBase
...
+    {
+        assert(getNumberOfComponents() == 1);
+        return getTuple(t, x)[0];
+    }
+
+    //! Componentwise parameter access. The component_id must be smaller than
+    //! the number of components of this parameter.
+    double operator()(double const t,
```
chleh: Minor: If this was virtual, a specific implementation could optimize it.

ProcessLib/Parameter/ConstantParameter.h
```diff
@@ -16,11 +16,20 @@ namespace ProcessLib
...
+    /// Construction with single value.
     ConstantParameter(T const& value) : _value({value}) {}
-
-    // TODO allow for different sizes
-    unsigned getNumberOfComponents() const override { return 1; }
+    /// Construction with a tuple.
+    /// The given tuple must be non-empty.
+    ConstantParameter(std::vector<T> const& value) : _value(value)
+    {
+        assert(!value.empty());
```
chleh: An assertion probably is not enough, since this will come from the prj file.

chleh:
Besides the method-nameing :+1:

ProcessLib/Parameter/Parameter.h
```diff
@@ -54,6 +54,24 @@ struct Parameter : public ParameterBase
     //! Returns the parameter value at the given time and position.
     virtual std::vector<T> const& getTuple(
         double const t, SpatialPosition const& pos) const = 0;
+
+    //! Scalar value parameter access. The number of components of this
+    //! parameter must be 1.
+    double operator()(double const t, SpatialPosition const& x) const
+    {
+        assert(getNumberOfComponents() == 1);
+        return getTuple(t, x)[0];
```
chleh: Btw: `.front()` probably does compile-time index computation. :smile:

endJunction: `front` and `operator[](0)` result in same binary code (at least when optimized).
When not optimized `operator[](int n)` is implemented roughly as `return *(this->data + n)`, whereas `front()` is `return *begin()` = `return *(iterator(this->data))`, so it creates a temporary iterator there... (At least in the gcc-6.1.1 implementation of the STL)

ProcessLib/Parameter/Parameter.h
endJunction: Adding virtual to both methods for consistency.

endJunction:
Renamed getTuple. Keeps the interface slim, while almost as nice as with overloaded call operators.

endJunction:
I had to rebase this, because of the before unknown usage of getTuple()..

chleh:
:fast_forward:
It seems you don't like `.front()` :smile:

PR #1380, opened 2016-08-29
Time dependent Dirichlet and Neumann BCs

chleh:
Time dependence is implemented using curves, i.e. `f(t, x) := c(t) * p(x)`, where `c` is a curve and `p` is a parameter.
The Parameter interface has been extended s.t. parameters now can tell if they are time dependent.


ProcessLib/BoundaryCondition/BoundaryConditionCollection.cpp
```diff
@@ -52,6 +38,14 @@ void BoundaryConditionCollection::addBCsForProcessVariables(
                   std::back_inserter(_boundary_conditions));
     }
 
-    initializeDirichletBCs(_boundary_conditions, _dirichlet_bcs);
+    // for each BC there will be storage for Dirichlet BC, which by default is
+    // empty and has to be filled by the respective BC object if needed.
+    _dirichlet_bcs.resize(_boundary_conditions.size());
```
endJunction: Maybe I misunderstand the comment, but resize sets the _dirichlet_bcs size, so `!_dirichlet_bcs.empty()`

ProcessLib/BoundaryCondition/BoundaryCondition.h
```diff
@@ -32,8 +34,23 @@ class BoundaryCondition
 public:
     //! Applies natural BCs (i.e. non-Dirichlet BCs) to the stiffness matrix
     //! \c K and the vector \c b.
-    virtual void apply(const double t, GlobalVector const& x, GlobalMatrix& K,
-                       GlobalVector& b) = 0;
+    virtual void applyNaturalBC(const double /*t*/, GlobalVector const& /*x*/,
```
norihiro-w: why not NeumannBC here while Dirichlet BC is used below? if it has to be used, Essential BC should be used instead of Dirichlet.

chleh: This could also be a Robin BC, not only Neumann. I'll change the name below.

ProcessLib/BoundaryCondition/BoundaryConditionCollection.cpp
chleh: Each BC can be either Dirichlet (or "essential") or natural, so each of the BCs need some storage for it. As you wrote, this implies that `_dirichlet_bcs` is not empty. However, the memory overhead for that should be rather small, because only some empty `std::vector`s are created.

ProcessLib/BoundaryCondition/BoundaryConditionCollection.h
```diff
@@ -41,8 +44,10 @@ class BoundaryConditionCollection final
         NumLib::LocalToGlobalIndexMap const& dof_table,
         unsigned const integration_order);
 
+    void preTimestep(const double t);
+
 private:
-    std::vector<NumLib::IndexValueVector<GlobalIndexType>> _dirichlet_bcs;
+    mutable std::vector<NumLib::IndexValueVector<GlobalIndexType>> _dirichlet_bcs;
```
chleh: In the long run, probably we should omit `mutable` as much as possible. This can be done by relaxing the constness of several methods. In the current case this might extend to the time integration code. So I'd like to leave it for later refactoring.

endJunction:
I've rebase it and included the mechanics and the thermal conductivity processes.

TomFischer:
:+1:

ProcessLib/BoundaryCondition/DirichletBoundaryCondition.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+    SpatialPosition pos;
+
+    bc_values.ids.clear();
+    bc_values.values.clear();
+
+    // convert mesh node ids to global index for the given component
+    bc_values.ids.reserve(bc_values.ids.size() + _mesh_node_ids.size());
+    bc_values.values.reserve(bc_values.values.size() + _mesh_node_ids.size());
+    for (auto id : _mesh_node_ids)
```
wenqing: ``const auto id``?

PR #1352, opened 2016-08-15
Numerical Jacobian assembly

chleh:
Closes #1338.
Follow-up of #1367.
Maybe this PR needs some discussion...

Some features:
* Basically the global MatrixVectorAssembler now is back again. I obviously removed it too early.
* The individual local assemblers do not own local matrices and vectors anymore.
* Changed the assembly slightly. Before there were the methods `assemble(M, K, b)` and `assembleJacobian(J)`, now there are `assemble(M, K, b)` and `assembleWithJacobian(M, K, b, J)`

Todo (after basic discussions are settled):
- [x] docu
- [x] replace `#pragma once`
- [x] mark TES Newton test as large

PR #1381, opened 2016-08-30
Finished unit tests for time integration

chleh:
Follow-up of #1352.


endJunction:
Now, the numerical Jacobian is merged, this needs a rebase.

Tests/NumLib/TestODEInt.cpp
```diff
@@ -68,189 +87,139 @@ class TestOutput
         const double delta_t = (num_timesteps == 0) ? -1.0
                                                     : ((t_end-t0) / num_timesteps);
 
-        INFO("Running test %s with %u timesteps of size %g s.",
-             _file_name_part.c_str(), num_timesteps, delta_t);
-        init_file(delta_t);
+        INFO("Running test with %u timesteps of size %g s.", num_timesteps,
```
endJunction: Maybe making this a DBUG or even removing is better, because the unit tests are not so noisy then.

endJunction: I even would suggest to disable logog output for these tests (especially the linear solver output); The twelve NumLibODEInt* tests produced 58644 lines of output (instead of expected ~24).

Tests/NumLib/ODEs.h
```diff
@@ -92,7 +92,8 @@ class ODETraits<ODE1>
 
 const double ODETraits<ODE1>::t0 = 0.0;
 
-const double ODETraits<ODE1>::t_end = 30.0;
+const double ODETraits<ODE1>::t_end =
+    2.0 * 3.141592653589793238462643383279502884197169399375105820974944;
```
chleh: Who says there is rounding involved. Cutting simply does the job. :smile:

Tests/NumLib/TestODEInt.cpp
chleh: Regarding your second comment: Making the our linear solver interfaces aware of the log level is a different PR.

endJunction:
I think there was enough time for review. Let's merge it after lunch.

Tests/NumLib/ODEs.h
wenqing: You can set
```
const double pi = 4.0  * std::atan(1.0)
```
somewhere as a global constant.

chleh: This is true analytically, but this approach might suffer from numerical inaccuracies of the atan() function. But in the future we should introduce a global constant for it. @norihiro-w uses `M_PI` in some current PR. But AFAIK that's a non-standard macro.

PR #1386, opened 2016-09-06
[T] Mark two MapGeometryToMeshSurf tests as LARGE.

endJunction:
This are Bode and Naegelstedt tests taking much more
time then other non-LARGE tests.

The tests will be executed every night by Jenkins, not every build (or PR).

### DIFF
--- a/Applications/Utils/Tests.cmake
+++ b/Applications/Utils/Tests.cmake
@@ -10,7 +10,7 @@ AddTest(
 )
 
 AddTest(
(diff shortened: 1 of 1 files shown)


TomFischer:
The test MaterialLibSolidsKelvinVector6.Inverse is falling. I think this has nothing to do with the changes of the PR.

:+1:

endJunction:
I have seen the failing tests. I'll do something about it.

PR #1390, opened 2016-09-07
Add virtual to dtors of polymorphic classes.

endJunction:
...as reported by compiler.
Prevents memory leaks.


norihiro-w:
:+1:

TomFischer:
:+1:

PR #1391, opened 2016-09-08
Burgers material model.

endJunction:
Copy of the OGS5 implementation in general. So while mathematically same, technically different.

There are two larger commented code blocks, testing parameter validity and another with linear solver choices---ignore them for now. They will be either deleted or commented as explicitly needed.

Newton method extracted in #1393, so this is a follow-up

### DIFF
--- a/MaterialLib/SolidModels/CreateLubby2.h
+++ b/MaterialLib/SolidModels/CreateLubby2.h
@@ -1,0 +1,101 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef MATERIALLIB_SOLIDMODELS_CREATELUBBY2_H_
+#define MATERIALLIB_SOLIDMODELS_CREATELUBBY2_H_
+
+#include <logog/include/logog.hpp>
+
+#include "Lubby2.h"
+#include "MechanicsBase.h"
+#include "ProcessLib/Utils/ProcessUtils.h"  // required for findParameter
+
+namespace MaterialLib
+{
+namespace Solids
+{
+template <int DisplacementDim>
+std::unique_ptr<MechanicsBase<DisplacementDim>> createLubby2(
+    std::vector<std::unique_ptr<ProcessLib::ParameterBase>> const& parameters,
+    BaseLib::ConfigTree const& config)
+{
+    config.checkConfigParameter("type", "Lubby2");
+    DBUG("Create Lubby2 material");
+
+    // Kelvin shear modulus.
+    auto& kelvin_shear_modulus = ProcessLib::findParameter<double>(
+        config, "kelvin_shear_modulus", parameters, 1);
+
+    DBUG("Use '%s' as kelvin shear modulus parameter.",
+         kelvin_shear_modulus.name.c_str());
+
+    // Kelvin viscosity.
+    auto& kelvin_viscosity = ProcessLib::findParameter<double>(
+        config, "kelvin_viscosity", parameters, 1);
+
+    DBUG("Use '%s' as kelvin viscosity parameter.",
+         kelvin_viscosity.name.c_str());
+
+    // Maxwell shear modulus.
+    auto& maxwell_shear_modulus = ProcessLib::findParameter<double>(
+        config, "maxwell_shear_modulus", parameters, 1);
+
+    DBUG("Use '%s' as maxwell shear modulus parameter.",
+         maxwell_shear_modulus.name.c_str());
+
+    // Maxwell bulk modulus.
+    auto& maxwell_bulk_modulus = ProcessLib::findParameter<double>(
+        config, "maxwell_bulk_modulus", parameters, 1);
+
+    DBUG("Use '%s' as maxwell bulk modulus parameter.",
+         maxwell_bulk_modulus.name.c_str());
+
+    // Maxwell viscosity.
+    auto& maxwell_viscosity = ProcessLib::findParameter<double>(
+        config, "maxwell_viscosity", parameters, 1);
+
+    DBUG("Use '%s' as maxwell viscosity parameter.",
+         maxwell_viscosity.name.c_str());
+
+    // Dependency parameter for mK.
+    auto& dependency_parameter_mK = ProcessLib::findParameter<double>(
+        config, "dependency_parameter_mk", parameters, 1);
+
+    DBUG("Use '%s' as dependency parameter mK.",
+         dependency_parameter_mK.name.c_str());
+
+    // Dependency parameter for mvK.
+    auto& dependency_parameter_mvK = ProcessLib::findParameter<double>(
+        config, "dependency_parameter_mvk", parameters, 1);
+
+    DBUG("Use '%s' as dependency parameter mvK.",
+         dependency_parameter_mvK.name.c_str());
+
+    // Dependency parameter for mvM.
+    auto& dependency_parameter_mvM = ProcessLib::findParameter<double>(
+        config, "dependency_parameter_mvm", parameters, 1);
+
+    DBUG("Use '%s' as dependency parameter mvM.",
+         dependency_parameter_mvM.name.c_str());
+
+
+    typename Lubby2<DisplacementDim>::MaterialProperties mp{
+        kelvin_shear_modulus,     maxwell_shear_modulus,
+        maxwell_bulk_modulus,     kelvin_viscosity,
+        maxwell_viscosity,        dependency_parameter_mK,
+        dependency_parameter_mvK, dependency_parameter_mvM};
+
+    return std::unique_ptr<MechanicsBase<DisplacementDim>>{
+        new Lubby2<DisplacementDim>{mp}};
+}
+
+}  // namespace Solids
+}  // namespace MaterialLib
+
+#endif  // MATERIALLIB_SOLIDMODELS_CREATELUBBY2_H_
--- a/MaterialLib/SolidModels/Lubby2-impl.h
+++ b/MaterialLib/SolidModels/Lubby2-impl.h
@@ -1,0 +1,269 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef MATERIALLIB_SOLIDMODELS_LUBBY2_IMPL_H_
+#define MATERIALLIB_SOLIDMODELS_LUBBY2_IMPL_H_
+
+#include "NewtonRaphson.h"
+
+namespace MaterialLib
+{
+namespace Solids
+{
+template <int DisplacementDim>
+void Lubby2<DisplacementDim>::computeConstitutiveRelation(
+    double const t,
+    ProcessLib::SpatialPosition const& x,
+    double const dt,
+    KelvinVector const& /*eps_prev*/,
+    KelvinVector const& eps,
+    KelvinVector const& /*sigma_prev*/,
+    KelvinVector& sigma,
+    KelvinMatrix& C,
+    typename MechanicsBase<DisplacementDim>::MaterialStateVariables&
+        material_state_variables)
+{
+    using Invariants = MaterialLib::SolidModels::Invariants<KelvinVectorSize>;
+
+    assert(dynamic_cast<MaterialStateVariables*>(&material_state_variables) !=
+           nullptr);
+    MaterialStateVariables& state =
+        static_cast<MaterialStateVariables&>(material_state_variables);
+    // calculation of deviatoric parts
+    auto const& P_dev = Invariants::deviatoric_projection;
+    KelvinVector const epsd_i = P_dev * eps;
+
+    // initial guess as elastic predictor
+    KelvinVector sigd_j = 2.0 * (epsd_i - state.eps_M_t - state.eps_K_t);
+
+    // Calculate effective stress and update material properties
+    double sig_eff = Invariants::equivalentStress(sigd_j);
+    updateBurgersProperties(t, x, sig_eff * state.GM, state);
+
+    using LocalJacobianMatrix =
+        Eigen::Matrix<double, KelvinVectorSize * 3, KelvinVectorSize * 3,
+                      Eigen::RowMajor>;
+
+    // Linear solver for the newton loop is required after the loop with the
+    // same matrix. This saves one decomposition.
+    Eigen::PartialPivLU<LocalJacobianMatrix> linear_solver(KelvinVectorSize *
+                                                           3);
+
+    // Different solvers are available for the solution of the local system.
+    // TODO Make the following choice of linear solvers available from the
+    // input file configuration:
+    //      K_loc.partialPivLu().solve(-res_loc);
+    //      K_loc.fullPivLu().solve(-res_loc);
+    //      K_loc.householderQr().solve(-res_loc);
+    //      K_loc.colPivHouseholderQr().solve(res_loc);
+    //      K_loc.fullPivHouseholderQr().solve(-res_loc);
+    //      K_loc.llt().solve(-res_loc);
+    //      K_loc.ldlt().solve(-res_loc);
+
+    {  // Local Newton solver
+        using LocalResidualVector =
+            Eigen::Matrix<double, KelvinVectorSize * 3, 1>;
+
+        LocalJacobianMatrix K_loc;
+        auto const update_residual = [&](LocalResidualVector& residual) {
+            calculateResidualBurgers(dt, epsd_i, sigd_j, state.eps_K_j,
+                                     state.eps_K_t, state.eps_M_j,
+                                     state.eps_M_t, residual, state);
+        };
+
+        auto const update_jacobian = [&](LocalJacobianMatrix& jacobian) {
+            calculateJacobianBurgers(
+                t, x, dt, jacobian, sig_eff, sigd_j, state.eps_K_j,
+                state);  // for solution dependent Jacobians
+        };
+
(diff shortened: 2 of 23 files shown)


MaterialLib/SolidModels/CreateViscoElasticBurgers.h
```diff
@@ -0,0 +1,168 @@
...
+#include "ViscoElasticBurgers.h"
+
+namespace Solids
+{
+template <int DisplacementDim>
+std::unique_ptr<MechanicsBase<DisplacementDim>> createViscoElasticBurgers(
+    std::vector<std::unique_ptr<ProcessLib::ParameterBase>> const& parameters,
+    BaseLib::ConfigTree const& config)
+{
+    config.checkConfigParameter("type", "ViscoElasticBurgers");
```
norihiro-w: isn't "Burgers" enough as a type name? I thought Burgers means a visco elastic material.

MaterialLib/SolidModels/CreateViscoElasticBurgers.h
```diff
@@ -0,0 +1,168 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef PHYSICALQUANITIES_SOLIDS_CREATEVISCOELASTICBURGERS_H_
```
norihiro-w: PHYSICALQUANITIES -> MATERIALLIB?

MaterialLib/SolidModels/ViscoElasticBurgers-impl.h
```diff
@@ -0,0 +1,332 @@
...
+    typename ViscoElasticBurgers<DisplacementDim>::KelvinMatrix& C,
+    typename MechanicsBase<DisplacementDim>::MaterialStateVariables&
+        material_state_variables)
+{
+    assert(dynamic_cast<MaterialStateVariables*>(&material_state_variables) !=
+           nullptr);
+    MaterialStateVariables& _state =
+        static_cast<MaterialStateVariables&>(material_state_variables);
+    // calculation of deviatoric parts
+    auto const& P_dev = MaterialLib::SolidModels::Invariants<
```
norihiro-w: suggestion for better readability: `using KelvinInvariants = MaterialLib::SolidModels::Invariants<KelvinVectorSize>;` at the beginning of this function.

MaterialLib/SolidModels/ViscoElasticBurgers-impl.h
```diff
@@ -0,0 +1,332 @@
...
+    typename ViscoElasticBurgers<
+        DisplacementDim>::KelvinVector const& /*sigma_prev*/,
+    typename ViscoElasticBurgers<DisplacementDim>::KelvinVector& sigma,
+    typename ViscoElasticBurgers<DisplacementDim>::KelvinMatrix& C,
+    typename MechanicsBase<DisplacementDim>::MaterialStateVariables&
+        material_state_variables)
+{
+    assert(dynamic_cast<MaterialStateVariables*>(&material_state_variables) !=
+           nullptr);
+    MaterialStateVariables& _state =
```
norihiro-w: having `_` prefix is used for a member variable.

MaterialLib/SolidModels/ViscoElasticBurgers-impl.h
```diff
@@ -0,0 +1,332 @@
...
+    // calculate Kelvin strain residual
+    res.template block<KelvinVectorSize, 1>(KelvinVectorSize, 0).noalias() =
+        1. / dt * (strain_Kel_curr - strain_Kel_t) -
+        1. / (2. * _state.etaK) *
+            (_state.GM * stress_curr - 2. * _state.GK * strain_Kel_curr);
+
+    // calculate Maxwell strain residual
+    res.template block<KelvinVectorSize, 1>(2 * KelvinVectorSize, 0).noalias() =
+        1. / dt * (strain_Max_curr - strain_Max_t) -
+        _state.GM / (2. * _state.etaM) * stress_curr;
```
wenqing: 0.5 * instead of /2.0

MaterialLib/SolidModels/ViscoElasticBurgers-impl.h
```diff
@@ -0,0 +1,332 @@
...
+
+    // build G_13
+    Jac.template block<KelvinVectorSize, KelvinVectorSize>(0,
+                                                           2 * KelvinVectorSize)
+        .diagonal()
+        .setConstant(2);
+
+    // build G_21
+    Jac.template block<KelvinVectorSize, KelvinVectorSize>(KelvinVectorSize, 0)
+        .noalias() = -_state.GM / (2. * _state.etaK) * KelvinMatrix::Identity();
```
wenqing: 0.5 * instead of /( 2.

MaterialLib/SolidModels/ViscoElasticBurgers-impl.h
```diff
@@ -0,0 +1,332 @@
...
+    if (s_eff > 0.)
+    {
+        KelvinVector const eps_K_aid =
+            1. / (_state.etaK * _state.etaK) *
+            (_state.GM * sig_i - 2. * _state.GK * eps_K_i);
+
+        KelvinVector const dG_K =
+            _mp.mK(t, x)[0] * 3. * _state.GK * _state.GM / (2. * s_eff) * sig_i;
```
wenqing: 1.5 * instead of 3. /2.

MaterialLib/SolidModels/ViscoElasticBurgers-impl.h
```diff
@@ -0,0 +1,332 @@
...
+                       _state._eps_M_j, _state._eps_M_t, res_loc, _state);
+    CalJacobianBurgers(t, x, dt, K_loc, sig_eff, sigd_j, _state._eps_K_j,
+                       _state);
+
+    //
+    // Local Newton loop
+    //
+
+    int counter = 0;
+    const int counter_max(20);
```
norihiro-w: should be configurable from input files

MaterialLib/SolidModels/ViscoElasticBurgers-impl.h
```diff
@@ -0,0 +1,332 @@
...
+#ifndef PHYSICALQUANITIES_SOLIDS_VISCOELASTICBURGERS_IMPL_H_
+#define PHYSICALQUANITIES_SOLIDS_VISCOELASTICBURGERS_IMPL_H_
+
+#include <logog/include/logog.hpp>
+
+#include "MaterialLib/SolidModels/KelvinVector.h"
+
+namespace Solids
+{
+template <int DisplacementDim>
```
wenqing: I think it would be nice to document the formulas of the material models.

MaterialLib/SolidModels/ViscoElasticBurgers-impl.h
```diff
@@ -0,0 +1,332 @@
...
+    const int counter_max(20);
+    const double local_tolerance(1.e-20);
+
+    // Linear solver for the newton loop is required after the loop with the
+    // same matrix. This saves one decomposition.
+    Eigen::PartialPivLU<Eigen::Matrix<double, KelvinVectorSize * 3,
+                                      KelvinVectorSize * 3, Eigen::RowMajor>>
+        linear_solver(KelvinVectorSize * 3);
+
+    while (res_loc.squaredNorm() > local_tolerance && counter < counter_max)
```
norihiro-w: suggest using a common Newton Raphson sovler instead of implementing this for each case

MaterialLib/SolidModels/ViscoElasticBurgers-impl.h
```diff
@@ -0,0 +1,332 @@
...
+            KelvinVectorSize>::equivalentStress(sigd_j);
+        UpdateBurgersProperties(t, x, sig_eff * _state.GM, _state);
+
+        // evaluation of new residual
+        CalResidualBurgers(dt, epsd_i, sigd_j, _state._eps_K_j, _state._eps_K_t,
+                           _state._eps_M_j, _state._eps_M_t, res_loc, _state);
+        DBUG("Local Newton: Iteration #%d |dx| = %g, |r| = %g", counter,
+             inc_loc.norm(), res_loc.norm());
+    }
+
```
norihiro-w: what to do if the iteration doesn't converge?

MaterialLib/SolidModels/ViscoElasticBurgers-impl.h
```diff
@@ -0,0 +1,332 @@
...
+    // Hydrostatic part for the stress and the tangent.
+    double const eps_i_trace =
+        MaterialLib::SolidModels::Invariants<KelvinVectorSize>::trace(eps);
+
+    sigma =
```
norihiro-w: noalias()?

MaterialLib/SolidModels/ViscoElasticBurgers-impl.h
```diff
@@ -0,0 +1,332 @@
...
+    double sig_eff = MaterialLib::SolidModels::Invariants<
+        KelvinVectorSize>::equivalentStress(sigd_j);
+    UpdateBurgersProperties(t, x, sig_eff * _state.GM, _state);
+
+    // initial evaluation of residual and Jacobian
+    Eigen::Matrix<double, KelvinVectorSize * 3, 1> res_loc;
+    Eigen::Matrix<double, KelvinVectorSize * 3, KelvinVectorSize * 3,
+                  Eigen::RowMajor>
+        K_loc;
+    CalResidualBurgers(dt, epsd_i, sigd_j, _state._eps_K_j, _state._eps_K_t,
```
norihiro-w: Cal -> Calculate

MaterialLib/SolidModels/ViscoElasticBurgers-impl.h
```diff
@@ -0,0 +1,332 @@
...
+#include "MaterialLib/SolidModels/KelvinVector.h"
+
+namespace Solids
+{
+template <int DisplacementDim>
+void ViscoElasticBurgers<DisplacementDim>::computeConstitutiveRelation(
+    double const t,
+    ProcessLib::SpatialPosition const& x,
+    double const dt,
+    typename ViscoElasticBurgers<
```
norihiro-w: can't you omit `ViscoElasticBurgers<DisplacementDim>::`?

MaterialLib/SolidModels/ViscoElasticBurgers-impl.h
wenqing: same question. suggest to return a flag for the convergence of this local NR (for stress integration). When the function returns false, break the global NR step and quit the program.

MaterialLib/SolidModels/ViscoElasticBurgers.h
```diff
@@ -0,0 +1,200 @@
...
+        {
+            _eps_K_t = _eps_K_j;
+            _eps_M_t = _eps_M_j;
+        }
+
+        using KelvinVector = ProcessLib::KelvinVectorType<DisplacementDim>;
+        using KelvinMatrix = ProcessLib::KelvinMatrixType<DisplacementDim>;
+        /// Deviatoric strain in the viscous kelvin element during the current
+        /// iteration
+        KelvinVector _eps_K_t;
```
norihiro-w: other  member variables like GM don't have prefix `_`. what's a difference?

ProcessLib/SmallDeformation/CreateSmallDeformationProcess.h
```diff
@@ -70,7 +71,12 @@ createSmallDeformationProcess(
         constitutive_relation_config.peekConfigParameter<std::string>("type");
 
     std::unique_ptr<Solids::MechanicsBase<DisplacementDim>> material = nullptr;
-    if (type == "LinearElasticIsotropic")
+    if (type == "ViscoElasticBurgers")
```
norihiro-w: why this comes before elastic? I think it's better if we have some rule about the order, e.g. alphabetic order.

endJunction:
Thanks for review. I'll place a "please review" label after fixing the code.

MaterialLib/SolidModels/ViscoElasticBurgers-impl.h
endJunction: I don't see the point here. Neither the mathematical expression reads better (to me) nor is there any performance difference; Could you explain the suggestion please?

MaterialLib/SolidModels/ViscoElasticBurgers-impl.h
endJunction: There are references to an article and a book on this model. The implementation is not necessarily a good teaching material.

MaterialLib/SolidModels/ViscoElasticBurgers.h
endJunction: Simply wrong. It was a class before, then changed to struct. Fixing.

MaterialLib/SolidModels/ViscoElasticBurgers-impl.h
endJunction: I'm surprised, that even in the case of assignment (not an update like +=) the noalias() made the code shorter.

ProcessLib/SmallDeformation/CreateSmallDeformationProcess.h
endJunction: Maintaining alphabetic order.

MaterialLib/SolidModels/CreateViscoElasticBurgers.h
nagelt: Right. I'd suggest "Lubby2", as we implement the non-linear dependencies according to that model.

endJunction:
All comments addressed, rebased, and using external NewtonRaphson implementation.
I have found another mistake where wrong namespace was used; see [Correct namespace of MaterialLib::Solids](02fca8c) commit.

MaterialLib/SolidModels/CreateLubby2.h
```diff
@@ -0,0 +1,101 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef MATERIALLIB_SOLIDMODELS_CREATEVISCOELASTICBURGERS_H_
```
norihiro-w: CREATELUBBY2

MaterialLib/SolidModels/Lubby2-impl.h
```diff
@@ -0,0 +1,269 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef MATERIALLIB_SOLIDMODELS_VISCOELASTICBURGERS_IMPL_H_
```
norihiro-w: here also with LUBBY2

MaterialLib/SolidModels/Lubby2-impl.h
```diff
@@ -0,0 +1,269 @@
...
+            NewtonRaphson<decltype(linear_solver), LocalJacobianMatrix,
+                          decltype(update_jacobian), LocalResidualVector,
+                          decltype(update_residual), decltype(update_solution)>(
+                linear_solver, update_jacobian, update_residual,
+                update_solution, maximum_iterations, tolerance);
+
+        auto const success_iterations = newton_solver.solve(K_loc);
+
+        if (!success_iterations)
+            OGS_FATAL("The local Newton method did not succeed.");
```
norihiro-w: the failure should be informed to a user so that they can, e.g. decrease a time step size, to get convergence.

MaterialLib/SolidModels/Lubby2.h
```diff
@@ -0,0 +1,205 @@
...
+        KelvinVector const& eps,
+        KelvinVector const& sigma_prev,
+        KelvinVector& sigma,
+        KelvinMatrix& C,
+        typename MechanicsBase<DisplacementDim>::MaterialStateVariables&
+            material_state_variables) override;
+
+private:
+    /// Updates Burgers material parameters in LUBBY2 fashion.
+    void UpdateBurgersProperties(double const t,
```
norihiro-w: start with a lower case in function names

MaterialLib/SolidModels/Lubby2.h
```diff
@@ -0,0 +1,205 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef MATERIALLIB_SOLIDMODELS_VISCOELASTICBURGERS_H_
```
norihiro-w: LUBBY2

MaterialLib/SolidModels/Lubby2-impl.h
endJunction: You mean something along these lines:
```c++
OGS_FATAL("The local Newton method did not converge in given maximum number of"
          "iterations. Decreasing the time step size could improve the"
          "convergence.");
```

norihiro-w: no, I didn't mean changing error message. Adaptive time stepping methods cannot read them.

norihiro-w:
looks ok for me besides some minor comments

MaterialLib/SolidModels/Lubby2-impl.h
endJunction: Second try: Changing OGS_FATAL to ERR and return a false, for example, can be implemented. Would propagate the error to the local assembly, but then it needs some state, read by the time stepper.
Let's have a short chat tomorrow.

endJunction:
ctests will be failing---need a rebase after the time-dependent BCs---but otherwise good.

PR #1392, opened 2016-09-08
Refactored Jenkins pipeline

bilke:
- Added one more Jenkins node *win1* (Windows)
- Windows builds are now better parallelized with [ninja](https://ninja-build.org/)
- Disabled S3 binary uploads for now (will be reimplemented in a different job)
- Win builds use Conan for all required libraries except of Qt ([Qt is in the works too](https://github.com/osechet/conan-qt/issues/7))
- All in all speed up of build time from roughly 1 h before to 10-15 mins now

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,3 +1,5 @@
+#!/usr/bin/env groovy
+
 node('master') {
     checkout scm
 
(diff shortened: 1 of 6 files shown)


endJunction:
Looks good! (and it's actually worth a changelog comment ;)

I'll merge it now, so if something goes horribly wrong (all PRs will be rebuild by pushing to master), it leaves us the opportunity to revert this first thing in the morning...

Good luck!

bilke:
@endJunction Looks good even after all the rebuilds!

Regarding rebuilds:

We can disable this behavior, see this:

PR #1393, opened 2016-09-11
Newton-Raphson for dense systems

endJunction:
A very simple abstraction of the named method required in the current visco-elastic Burgers material model (#1391) and in the following single surface yield function material model (currently in development).

~~I put it into NumLib/NewtonMethods for now, but if this is not general enough (and you don't want extend it), it can be~~ It is placed near the material models implementations, where it is used.

No aspect of the solver can be configured by input files.

### DIFF
--- a/MaterialLib/SolidModels/NewtonRaphson.h
+++ b/MaterialLib/SolidModels/NewtonRaphson.h
@@ -1,0 +1,99 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef MATERIALLIB_SOLIDMODELS_NEWTONRAPHSON_H_
+#define MATERIALLIB_SOLIDMODELS_NEWTONRAPHSON_H_
+
+#include <boost/optional.hpp>
+#include <logog/include/logog.hpp>
+
+#include <Eigen/Dense>
+
+namespace MaterialLib
+{
+namespace Solids
+{
+/// Newton-Raphson solver for system of equations using an Eigen linear solvers
+/// library.
+/// The current implementation does not update the solution itself, but calls a
+/// function for the solution's update with the current increment.
+template <typename LinearSolver, typename JacobianMatrix,
+          typename JacobianMatrixUpdate, typename ResidualVector,
+          typename ResidualUpdate, typename SolutionUpdate>
+class NewtonRaphson final
+{
+public:
+    NewtonRaphson(LinearSolver& linear_solver,
+                  JacobianMatrixUpdate jacobian_update,
+                  ResidualUpdate residual_update,
+                  SolutionUpdate solution_update, int const maximum_iterations,
+                  double const tolerance)
+        : _linear_solver(linear_solver),
+          _jacobian_update(jacobian_update),
+          _residual_update(residual_update),
+          _solution_update(solution_update),
+          _maximum_iterations(maximum_iterations),
+          _tolerance_squared(tolerance * tolerance)
+    {
+    }
+
+    /// Returns true and the iteration number if succeeded, otherwise false and
+    /// an undefined value for the number of iterations.
+    boost::optional<int> solve(JacobianMatrix& jacobian) const
+    {
+        int iteration = 0;
+        ResidualVector increment;
+        ResidualVector residual;
+        do
+        {
+            // The jacobian and the residual are updated simulataniously to keep
+            // consistency. The jacobian is used after the non-linear solver
+            // onward.
+            _jacobian_update(jacobian);
+            _residual_update(residual);
+
+            if (residual.squaredNorm() < _tolerance_squared)
+                break;  // convergence criteria fulfilled.
+
+            increment.noalias() =
+                _linear_solver.compute(jacobian).solve(-residual);
+            // DBUG("Local linear solver accuracy |J dx - r| = %g",
+            //      (jacobian * increment + residual).norm());
+
+            _solution_update(increment);
+
+            // DBUG("Local Newton: Iteration #%d |dx| = %g, |r| = %g",
+            //      iteration, increment.norm(), residual.norm());
+        } while (iteration++ < _maximum_iterations);
+
+        if (iteration > _maximum_iterations)
+        {
+            ERR("The local Newton method did not converge within the given "
+                "number of iterations. Iteration: %d, increment %g, residual: "
+                "%g",
+                iteration - 1, increment.norm(), residual.norm());
+            return {};
+        }
+
+        return iteration;
+    };
+
+private:
+    LinearSolver& _linear_solver;
+    JacobianMatrixUpdate _jacobian_update;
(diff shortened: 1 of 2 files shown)


NumLib/NewtonMethods/NewtonRaphson.h
```diff
@@ -0,0 +1,79 @@
...
+ */
+
+#ifndef NUMLIB_NEWTONMETHODS_NEWTONRAPHSON_H_
+#define NUMLIB_NEWTONMETHODS_NEWTONRAPHSON_H_
+
+#include <logog/include/logog.hpp>
+
+#include <Eigen/Dense>
+
+namespace NumLib
```
norihiro-w: I think you can even put this into MathLib, because it doesn't depend on other NumLib stuff

NumLib/NewtonMethods/NewtonRaphson.h
```diff
@@ -0,0 +1,79 @@
...
+/// function for the solution's update with the current increment.
+template <typename LinearSolver, typename JacobianMatrix,
+          typename JacobianMatrixUpdate, typename ResidualVector,
+          typename ResidualUpdate, typename SolutionUpdate>
+class Newton final
+{
+public:
+    Newton(LinearSolver& linear_solver, JacobianMatrixUpdate jacobian_update,
+           ResidualUpdate residual_update, SolutionUpdate solution_update,
+           int const maximum_iterations, double const tolerance_squared)
```
norihiro-w: why should a user pass a squared tolerance instead of a tolerance?

NumLib/NewtonMethods/NewtonRaphson.h
```diff
@@ -0,0 +1,79 @@
...
+        : _linear_solver(linear_solver),
+          _jacobian_update(jacobian_update),
+          _residual_update(residual_update),
+          _solution_update(solution_update),
+          _maximum_iterations(maximum_iterations),
+          _tolerance_squared(tolerance_squared)
+    {}
+
+    /// Returns number of iterations.
+    int solve(JacobianMatrix& jacobian, ResidualVector& residual) const
```
norihiro-w: a solution vector should also be provided, which a user can set their initial guess and get results.

NumLib/NewtonMethods/NewtonRaphson.h
```diff
@@ -0,0 +1,79 @@
...
+    {
+        int iteration = 0;
+        ResidualVector increment;
+        do
+        {
+            _jacobian_update(jacobian);
```
norihiro-w: evaluate jacobian after checking residual

NumLib/NewtonMethods/NewtonRaphson.h
```diff
@@ -0,0 +1,79 @@
...
+            //DBUG("Local linear solver accuracy |J dx - r| = %g",
+            //     (jacobian * increment + residual).norm());
+
+            _solution_update(increment);
+
+            //DBUG("Local Newton: Iteration #%d |dx| = %g, |r| = %g", iteration,
+            //     increment.norm(), residual.norm());
+        } while (iteration++ < _maximum_iterations);
+
+        return iteration;
```
norihiro-w: i think there is a problem if we return only `iteration`. If it is equal to the given max. iterations, it can mean the iteration is stopped because it reaches the max number (line 62) or is finished because it converged (line 52).

NumLib/NewtonMethods/NewtonRaphson.h
endJunction: This saves one sqrt computation for each iteration.

NumLib/NewtonMethods/NewtonRaphson.h
endJunction: Hmm, this is implicitly handled by the `*_update` functions. The reason behind it is, that the "solution" (at least in Burgers) is spread among several quantities: see MaterialLib/SolidModels/ViscoElasticBurgers-impl.h in the #1391 PR around line 85 (update_solution).
Better to discuss further things...

NumLib/NewtonMethods/NewtonRaphson.h
endJunction: This is indeed, not nice. Will return an indicator for error too.

NumLib/NewtonMethods/NewtonRaphson.h
endJunction: Moving this to MaterialLib/SolidModels because the implementation operates on implicit state which is updated with `update_solution(increment)`, instead of `solution += increment` in the loop body.

MaterialLib/SolidModels/NewtonRaphson.h
```diff
@@ -0,0 +1,97 @@
...
+            // DBUG("Local linear solver accuracy |J dx - r| = %g",
+            //      (jacobian * increment + residual).norm());
+
+            _solution_update(increment);
+
+            // DBUG("Local Newton: Iteration #%d |dx| = %g, |r| = %g",
+            //      iteration, increment.norm(), residual.norm());
+        } while (iteration++ < _maximum_iterations);
+
+        if (iteration == _maximum_iterations)
```
norihiro-w: guess you need > instead of equal

MaterialLib/SolidModels/NewtonRaphson.h
```diff
@@ -0,0 +1,97 @@
...
+            // The jacobian and the residual are updated simulataniously to keep
+            // consistency. The jacobian is used after the non-linear solver
+            // onward.
+            _jacobian_update(jacobian);
+            _residual_update(residual);
+
+            if (residual.squaredNorm() < _tolerance_squared)
+                break;  // convergence criteria fulfilled.
+
+            increment = _linear_solver.compute(jacobian).solve(-residual);
```
norihiro-w: noalias?

MaterialLib/SolidModels/NewtonRaphson.h
```diff
@@ -0,0 +1,97 @@
...
+          _residual_update(residual_update),
+          _solution_update(solution_update),
+          _maximum_iterations(maximum_iterations),
+          _tolerance_squared(tolerance * tolerance)
+    {
+    }
+
+    /// Returns true and the iteration number if succeeded, otherwise false and
+    /// an undefined value.
+    std::pair<bool, int> solve(JacobianMatrix& jacobian,
```
norihiro-w: is a residual vec also used afterwards? if not, you can make it internally.

norihiro-w:
:+1:

Tests/MaterialLib/NewtonRaphson.cpp
```diff
@@ -0,0 +1,55 @@
...
+    auto const newton_solver = MaterialLib::Solids::NewtonRaphson<
+        decltype(linear_solver), LocalJacobianMatrix, decltype(update_jacobian),
+        LocalResidualVector, decltype(update_residual),
+        decltype(update_solution)>(linear_solver, update_jacobian,
+                                   update_residual, update_solution,
+                                   maximum_iterations, tolerance);
+    auto const success_iterations = newton_solver.solve(jacobian);
+
+    EXPECT_TRUE(success_iterations.first);
+    EXPECT_LE(success_iterations.second, 11);
```
TomFischer: conflicts line 25

TomFischer:
After fixing test: :+1:

PR #1394, opened 2016-09-12
Preparations for the BalanceProcess

TomFischer:
The class `MeshSurfaceExtraction` extracts the surface of a subsurface mesh. For the balance process it is also important to know the subsurface element id and the number of the face of the subsurface element that belongs to the surface. 

The PR adds the possibility to write subsurface/bulk element ids and face ids to the mesh.

### DIFF
--- a/MeshLib/Mesh.h
+++ b/MeshLib/Mesh.h
@@ -20,6 +20,7 @@
 #include <vector>
 
 #include "BaseLib/Counter.h"
+#include "BaseLib/Error.h"
 
 #include "MeshEnums.h"
 #include "Properties.h"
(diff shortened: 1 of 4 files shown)


MeshLib/MeshSurfaceExtraction.h
```diff
@@ -56,7 +56,9 @@ class MeshSurfaceExtraction
         const MeshLib::Mesh& mesh,
         const MathLib::Vector3& dir,
         double angle,
-        std::string const& subsfc_node_id_backup_prop_name = "");
+        std::string const& subsfc_node_id_backup_prop_name = "",
```
endJunction: Documentation needs an update...

MeshLib/MeshSurfaceExtraction.h
```diff
@@ -72,6 +74,8 @@ class MeshSurfaceExtraction
     static void get2DSurfaceElements(
         const std::vector<MeshLib::Element*>& all_elements,
         std::vector<MeshLib::Element*>& sfc_elements,
+        std::vector<std::size_t>& element_id_map,
+        std::vector<std::size_t>& face_id_map,
```
endJunction: These maps are mapping from which mesh to which mesh? Needs a documentation or a more descriptive name.

MeshLib/MeshSurfaceExtraction.cpp
```diff
@@ -124,6 +131,31 @@ MeshLib::Mesh* MeshSurfaceExtraction::getMeshSurface(
...
         }
     }
+
+    // transmit the original subsurface element ids as a property
+    if (!subsfc_element_id_prop_name.empty()) {
+        boost::optional<MeshLib::PropertyVector<std::size_t>&> orig_elem_ids(
+            result->getProperties().createNewPropertyVector<std::size_t>(
+                subsfc_element_id_prop_name , MeshLib::MeshItemType::Cell, 1));
+        if (orig_elem_ids) {
+            orig_elem_ids->resize(element_ids_map.size());
```
endJunction: reserve and back_inserter

endJunction:
:fast_forward:

MeshLib/MeshSurfaceExtraction.cpp
```diff
@@ -124,6 +131,31 @@ MeshLib::Mesh* MeshSurfaceExtraction::getMeshSurface(
...
+                      orig_elem_ids->begin());
+        }
+    }
+
+    // transmit the face id of the original subsurface element as a property
+    if (!face_id_prop_name.empty()) {
+        boost::optional<MeshLib::PropertyVector<std::size_t>&> orig_face_ids(
+            result->getProperties().createNewPropertyVector<std::size_t>(
+                face_id_prop_name , MeshLib::MeshItemType::Cell, 1));
+        if (orig_face_ids) {
```
TomFischer: Then the information get lost. I'll document this behaviour.

MeshLib/MeshSurfaceExtraction.h
TomFischer: See commit  	0065099.

MeshLib/MeshSurfaceExtraction.h
TomFischer: Renamed in 0835b06.

MeshLib/MeshSurfaceExtraction.cpp
endJunction: Better to quit. Checking if that thing went wrong after the programm run is much more difficult, than an early exit. (Imagine this runs in a script. What would you need to do to find that error?)

MeshLib/MeshSurfaceExtraction.cpp
```diff
@@ -251,8 +302,10 @@ std::vector<MeshLib::Node*> MeshSurfaceExtraction::getSurfaceNodes(
 {
     INFO ("Extracting surface nodes...");
     std::vector<MeshLib::Element*> sfc_elements;
-    get2DSurfaceElements(mesh.getElements(), sfc_elements, dir, angle,
-                         mesh.getDimension());
+    std::vector<std::size_t> element_ids_map;
+    std::vector<std::size_t> face_id_map;
```
endJunction: Maybe same rename (if not touching too much) would be good here.

MeshLib/MeshSurfaceExtraction.cpp
TomFischer: I'm not sure if to quit here is the right solution. Maybe a user don't need this information and for this reason gives an empty name string?

MeshLib/Mesh.h
```diff
@@ -172,6 +173,23 @@ void scaleMeshPropertyVector(MeshLib::Mesh& mesh,
                              std::string const& property_name,
                              double factor);
 
+template <typename T>
+void addPropertyToMesh(MeshLib::Mesh& mesh, std::string const& name,
```
endJunction: Needs a docu especially about existing properties.

reserve and back_inserter below ;)

norihiro-w:
:+1:

PR #1399, opened 2016-09-13
Changed edge number for line elements

rinkk:
Without this in pure line meshes the min- and max-length for edges will be wrong.

### DIFF
--- a/MeshLib/Elements/LineRule2.h
+++ b/MeshLib/Elements/LineRule2.h
@@ -39,6 +39,9 @@ public:
     /// Constant: The FEM type of the element
     static const CellType cell_type = CellType::LINE2;
 
+    /// Constant: The number of edges
+    static const unsigned n_edges = 1;
+
     /// Constant: The number of neighbors
     static const unsigned n_neighbors = 2;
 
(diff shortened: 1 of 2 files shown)


endJunction:
Same here, please provide a test which shows a defect (or covers new code) and is resolved by this pull request.

For general discussion: I'm not sure about the wording here, but lines do not have edges. If they do, then a hex has 6 edges.

rinkk:
If you have a mesh consisting only of lines, such as this one:

```
#FEM_MSH
$PCS_TYPE
  NotSpecified
$NODES
  2
0 0 0 0
1 1 0 0
$ELEMENTS
  1
0 0 line 0 1
#STOP
```
and call ```Mesh::getMinEdgeLength()```, the result you will get is ```std::numeric_limits<double>::max()```. The same happens for ```Mesh::getMaxEdgeLength()```.
(This is done for every mesh loaded into the Data Explorer but is probably also relevant for a number of analytical geometry algorithms.)

Using the above change both methods will return the correct result.
Should I simply write a test that takes a line mesh and returns the correct values? Would this be sufficient or do you expect something else?

endJunction:
Looks harmless. :fast_forward:

PR #1397, opened 2016-09-13
[FEFLOW] fix a bug that a property vector was not allocated before use

norihiro-w:
follow up of #1400. fix #1395

PR #1400, opened 2016-09-14
activates tests on travis also for GUI builds

norihiro-w:
related to #1397, currently some tests using Qt are not executed on travis, because only GUI builds include Qt and tests are deactivated for those builds. This PR activates the tests.


norihiro-w:
of course, this makes travis jobs longer

norihiro-w:
it seems GUI builds on Jenkins are executing tests. if Jenkins test is enough, I would close this PR.

bilke:
Job run times on Travis vary a lot. Sometime non-gui jobs even take as long as over 15 mins, so feel free to merge this.

endJunction:
So long the jobs are not aborted too often on travis because of the time limit, it's good to include more tests. We'll see.

PR #1407, opened 2016-09-16
Enable swmm interface compilation on Jenkins

bilke:
Windows only.


bilke:
Yes it saves network traffic, reduces failures due to network issues, and should be faster overall (files are compressed on the master then transferred to the slaves and uncompressed there).

PR #1414, opened 2016-09-19
[Jenkins] Explicitly allocate workspace before unstashing.

bilke:
Should fix errors on unstashing, see
https://groups.google.com/forum/#!topicsearchin/jenkinsci-users/pipeline/jenkinsci-users/xj-yAIPjAFI


norihiro-w:
:fast_forward:

PR #1418, opened 2016-09-20
[T] WIP added boost xml test

chleh:
It seems like the boost xml parser does not check for matching opening/closing tags...


endJunction:

PR #1422, opened 2016-09-22
Fix an issue in `BaseLib::copyPathToFileName`

TomFischer:
I had to activate the windows tests only for windows since at one place a path separator is added. In general in my opinion we should switch to something like [boost::filesystem](http://www.boost.org/doc/libs/1_61_0/libs/filesystem/doc/index.htm).

**update**
`BaseLib::copyPathToFileName` takes two arguments: a file name and a path. The method puts them together. If the given `path` argument does not end with a path separator the previous version removed the part of the path from the last path separator till the end, i.e., for example `path/smth` was reduced to the path `path/`. This behaviour is changed with this PR. Now a path separator will be appended, i.e., the path will be modified: `path/smth/`.

### DIFF
--- a/BaseLib/FileTools.cpp
+++ b/BaseLib/FileTools.cpp
@@ -123,6 +123,13 @@ bool hasFileExtension(std::string const& extension, std::string const& filename)
     return boost::iequals(extension, getFileExtension(filename));
 }
 
+static const char pathSeparator =
+#ifdef _WIN32
+                            '\\';
+#else
+                            '/';
+#endif
+
 std::string copyPathToFileName(const std::string &file_name,
                                const std::string &source)
 {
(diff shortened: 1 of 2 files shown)


BaseLib/FileTools.cpp
```diff
@@ -131,6 +138,10 @@ std::string copyPathToFileName(const std::string &file_name,
     if (pos != std::string::npos)
         return file_name;
 
+    if (source.empty())
+        return file_name;
+    if (source.back() != pathSeparator)
+        return BaseLib::extractPath(source+pathSeparator).append(file_name);
```
chleh: Note: Maybe the `joinPaths`  function could be used here.

bilke:
Windows test [fails](https://svn.ufz.de:8443/job/OGS-6/job/ufz/job/PR-1422/1/execution/node/96/log/).

endJunction:
:+1: for boost::filesystem.

PR #1425, opened 2016-09-22
[NodeReordering] check each element

norihiro-w:
Currently `NodeReordering` tool assumes that all elements in the given mesh have wrong node ordering. This PR extends it to checking ordering of each element and modifying only those having wrong ordering.


chleh:
:fast_forward:

PR #1427, opened 2016-09-23
make Line elements return an edge

### LINKED ISSUE #1419: [MeshGeo] Segmentation fault because Line::getEdge(index)  returns nullptr
#1399 has changed the number of edges of line elements from zero to 1. After that, if a mesh includes lines,  functions having an algorithm like below (e.g. in `BoundaryElementsAlongPolyline::BoundaryElementsAlongPolyline()` ) encounters a seg fault, because `Line::getEdge()` still returns `nullptr`.

```
        for (unsigned i=0; i<e->getNumberOfEdges(); i++) {
            auto* edge = e->getEdge(i);
            for (unsigned j=0; j<edge.getNumberOfBaseNodes(); j++) {
                // do something with the edge
            }
        }
```

norihiro-w:
fix #1419 

To keep the same behavior as before, I also changed `BoundaryElementsAlongPolyline` class ignores line elements.


endJunction:
Good to merge.

PR #1428, opened 2016-09-23
[Jenkins] Docs generation

bilke:
The job is always run. Should this be done on ufz master only?

### DIFF
--- a/scripts/jenkins/docs.groovy
+++ b/scripts/jenkins/docs.groovy
@@ -15,7 +15,7 @@ node('docker') {
     }
 
     stage 'Reports (Docs)'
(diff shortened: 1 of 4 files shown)


scripts/jenkins/docs.groovy
```diff
@@ -0,0 +1,31 @@
+node('docker') {
+    def defaultDockerArgs = '-v /home/jenkins/.ccache:/usr/src/.ccache'
```
chleh: Do we need ccache for docs?

bilke: No

PR #1429, opened 2016-09-23
Impl. of BalanceProcess

TomFischer:
This PR implements the basic infrastructure for balancing fluxes. In the moment the balance can only be calculated for a quad mesh that is a boundary of a hexahedra mesh. This will be extended in further PR. At the moment the balance is only implemented into the GroundwaterFlow but is ready to use in any other process, too. This will also changed in further PRs.

**Update**
- [x] After the review I'll rename `BalanceProcess` to `CalculateSurfaceFlux` as suggested by @wenqing and @norihiro-w.
- [x] ChangeLog entry


TomFischer:
Jenkins, test this please.

TomFischer:
Jenkins, test this please.

TomFischer:
Jenkins, test this please.

ProcessLib/BalanceProcess/ParseBalanceProcessData.cpp
```diff
@@ -0,0 +1,36 @@
...
+        //! \ogs_file_param{process__balance}
+        config.getConfigSubtreeOptional("balance");
+    if (!balance_config)
+        return;
+
+    mesh_name =
+        //! \ogs_file_attr{process__balance__mesh}
+        balance_config->getConfigParameter<std::string>("mesh");
+    balance_pv_name =
+        //! \ogs_file_attr{process__balance__balance_pv_name}
```
chleh: ... balance_property_name

ProcessLib/BalanceProcess/MapBulkElementPoint.h
```diff
@@ -0,0 +1,104 @@
...
+#define PROCESSLIB_BALANCEPROCESS_MAPBULKELEMENT_H
+
+#include <array>
+
+#include "MathLib/TemplateWeightedPoint.h"
+
+#include "MeshLib/Elements/FaceRule.h"
+#include "MeshLib/Elements/Elements.h"
+
+inline MathLib::Point3d getBulkElementPoint(MeshLib::Quad const& quad,
```
chleh: Please add some docu.
Maybe the defintions should be moved to a cpp file?

ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
```diff
@@ -117,31 +117,55 @@ class LocalAssemblerData
...
+        using FemType =
+            NumLib::TemplateIsoparametric<ShapeFunction, ShapeMatricesType>;
+
+        FemType fe(*static_cast<const typename ShapeFunction::MeshElement*>(
+            &_element));
+
+        typename ShapeMatricesType::ShapeMatrices shape_matrices(
+            ShapeFunction::DIM, GlobalDim, ShapeFunction::NPOINTS);
+
+        fe.computeShapeFunctions(p.getCoords(), shape_matrices, GlobalDim);
```
chleh: OK, here you need the Point3D. I guess it has to be in natural coordinates. Maybe that should be documented (and reflected in the argument name).

ProcessLib/GroundwaterFlow/GroundwaterFlowProcess.h
```diff
@@ -14,7 +14,46 @@
...
 #include "GroundwaterFlowFEM.h"
 #include "GroundwaterFlowProcessData.h"
+#include "ProcessLib/BalanceProcess/BalanceProcess.h"
 
+// TODO used for output, if output classes are ready this has to be changed
+#include "MeshLib/IO/writeMeshToFile.h"
+
+// TODO Copied from VectorMatrixAssembler. Could be provided by the DOF table.
+inline NumLib::LocalToGlobalIndexMap::RowColumnIndices
+getRowColumnIndices_(std::size_t const id,
```
chleh: I think by now this method is already defined in some `*Util*.{h,cpp}` file. Same for the function below.

ProcessLib/BalanceProcess/BalanceProcess.cpp
```diff
@@ -0,0 +1,68 @@
...
+    std::vector<std::unique_ptr<MeshLib::MeshSubsets>> all_mesh_subsets;
+    std::generate_n(
+        std::back_inserter(all_mesh_subsets),
+        bulk_property_number_of_components,
+        [&]() {
+            return std::unique_ptr<MeshLib::MeshSubsets>{
+                new MeshLib::MeshSubsets{_mesh_subset_all_nodes.get()}};
+        });
+
+    // dof table maps mesh elements to local assemblers
```
chleh: the comment seems wrong, because for each element there is one local assembler, no dof-table needed.

ProcessLib/BalanceProcess/BalanceProcess.cpp
```diff
@@ -0,0 +1,68 @@
...
+
+#include "BalanceProcess.h"
+
+#include <cassert>
+
+#include "ProcessLib/Utils/CreateLocalAssemblers.h"
+
+namespace ProcessLib
+{
+BalanceProcess::BalanceProcess(MeshLib::Mesh& mesh,
```
chleh: is this the bulk or boundary mesh? (It's clear from the doxygen, but maybe change the variable name?)

ProcessLib/BalanceProcess/BalanceProcessLocalAssembler.h
```diff
@@ -0,0 +1,154 @@
...
+            NumLib::TemplateIsoparametric<ShapeFunction, ShapeMatricesType>;
+
+        FemType fe(*static_cast<const typename ShapeFunction::MeshElement*>(
+            &surface_element));
+
+        std::size_t const n_integration_points =
+            _integration_method.getNumberOfPoints();
+
+        std::vector<typename ShapeMatricesType::ShapeMatrices> shape_matrices;
+        shape_matrices.reserve(n_integration_points);
```
chleh: No need to store all shape matrices if you only need detJ.

ProcessLib/BalanceProcess/BalanceProcessLocalAssembler.h
```diff
@@ -0,0 +1,154 @@
...
+        pos.setElementID(element_id);
+
+        auto surface_element_normal =
+            MeshLib::FaceRule::getSurfaceNormal(&_surface_element);
+        surface_element_normal.normalize();
+
+        std::size_t const n_integration_points =
+            _integration_method.getNumberOfPoints();
+        // balance[id_of_element] =
+        //   int_{\Gamma_e} \alpha \cdot grad variable \cdot normal \d \Gamma
```
chleh: ` \cdot grad variable` → ` \cdot flux`

ProcessLib/BalanceProcess/BalanceProcessLocalAssembler.h
```diff
@@ -0,0 +1,154 @@
...
+        for (std::size_t ip(0); ip < n_integration_points; ip++)
+        {
+            pos.setIntegrationPoint(ip);
```
chleh: `pos` is currently not used.

ProcessLib/BalanceProcess/BalanceProcessLocalAssembler.h
```diff
@@ -0,0 +1,154 @@
...
+            auto const bulk_element_point = getBulkElementPoint(
+                bulk_process.getMesh(), _bulk_element_id, _bulk_face_id, wp);
+            auto const bulk_flux =
+                bulk_process.getFlux(_bulk_element_id, bulk_element_point, x);
+
+            for (std::size_t component_id(0);
+                 component_id < balance.getNumberOfComponents();
+                 ++component_id)
+            {
+                // TODO find solution for 2d case
```
chleh: `Eigen::Map<...>(bulk_flux.data(), bulk_flux.size()).dot(Eigen::Map<...>(surface_element_normal.data(), surface_element_normal.size())` or similar.

ProcessLib/BalanceProcess/BalanceProcessLocalAssembler.h
```diff
@@ -0,0 +1,154 @@
...
+    void integrate(std::size_t element_id,
+                   GlobalVector const& x,
+                   MeshLib::PropertyVector<double>& balance,
+                   Process const& bulk_process) override
+    {
+        SpatialPosition pos;
+        pos.setElementID(element_id);
+
+        auto surface_element_normal =
+            MeshLib::FaceRule::getSurfaceNormal(&_surface_element);
```
chleh: Note: A different approach is needed for higher order elements. Maybe add a TODO note?

endJunction: The higher order elements are geometrically linear. Could you explain a little more, what should be different.

ProcessLib/GroundwaterFlow/GroundwaterFlowProcess.h
TomFischer: No, it was in the TES process. I moved the functionality to the dof table.

ProcessLib/BalanceProcess/BalanceProcessLocalAssembler.h
chleh: > The higher order elements are geometrically linear.

AFAIK that is not true in general. The higher-order nodes are not necessarily on the faces, but can deviate from them. I think, @nagelt can explain that better than I.

ProcessLib/BalanceProcess/BalanceProcessLocalAssembler.h
TomFischer: `getSurfaceNormal`  returns a `MathLib::Vector3` object, i.e., the normal of the surface is always a 3d vector. Maybe we have o handle the 2d bulk / 1d boundary case special, i.e., rotate the 3d normal to the 2d plane. Anyway, I will change this and use the `Eigen::Map` solution.

NumLib/DOF/LocalToGlobalIndexMap.cpp
```diff
@@ -191,4 +191,24 @@ std::ostream& operator<<(std::ostream& os, LocalToGlobalIndexMap const& map)
 }
 #endif  // NDEBUG
 
+NumLib::LocalToGlobalIndexMap::RowColumnIndices getRowColumnIndices(
```
chleh: There already is a very similar method (`getIndices()`) in NumLib/DOF/DOFTableUtil.cpp
Your `getVectorValues()` is already provided by `EigenVector::get()` etc.
As far as I can see.

ProcessLib/BalanceProcess/BalanceProcessLocalAssembler.h
TomFischer: In the current implementation only geometricaly linear element are allowed.

TomFischer:
I am reworking the implementation.

NumLib/DOF/LocalToGlobalIndexMap.cpp
TomFischer: Thanks for pointing this out, the changes concerning dof table are moved to a separat PR #1442.

ProcessLib/BalanceProcess/BalanceProcess.cpp
TomFischer: You are right, there is a one to one correspondence between elements and local assemblers. But is the mapping incorrect? I thought that the dof table is necessary for the local assembler creation.

ProcessLib/BalanceProcess/BalanceProcessLocalAssembler.h
TomFischer: Removed.

TomFischer:
I hope I fixed/commented all of @chleh remarks. The review can go on ...

ProcessLib/LocalAssemblerInterface.h
```diff
@@ -53,6 +54,12 @@ class LocalAssemblerInterface
                               NumLib::LocalToGlobalIndexMap const& dof_table,
                               GlobalVector const& x);
 
+    virtual std::vector<double> getFlux(
+        MathLib::Point3d const& p, std::vector<double> const& local_x) const
```
chleh: OK.

ProcessLib/BalanceProcess/BalanceProcess.cpp
chleh: As far as I can see, it's necessary for the assembly, but not for the local assembler creation (it's not passed to the ctor).
The DOF table maps elements to dof's in the global vector (in the simplest case that's essentially the mesh nodes).

ProcessLib/BalanceProcess/BalanceProcessLocalAssembler.h
chleh: I don't want to blame the current implementation, but maybe add a comment somewhere that for higher-order elements further work is needed. (E.g., add a todo note at this line.)

ProcessLib/BalanceProcess/BalanceProcessLocalAssembler.h
chleh: I meant the Eigen::Map stuff as a solution for any dimension.
Even though `getSurfaceNormal()` returns a 3D vector, in 2D the z-component should always be zero, So you don't need to write a fixed `3` in your implementation.
Shouldn't the following work in any dimension (note the `bulk_flux.size()` in both places)?
```{.cpp}
                   Eigen::Map<Eigen::RowVectorXd const>(bulk_flux.data(),
                                                  bulk_flux.size())
                       .dot(Eigen::Map<Eigen::RowVectorXd const>(
                           surface_element_normal.getCoords(), bulk_flux.size())));
```

chleh:
Two new comments (see above)
* getSurfaceNormal
* Eigen::Map<...>

afterwards :+1:

ProcessLib/BalanceProcess/BalanceProcessLocalAssembler.h
TomFischer: The element normal is calculated in the physical space, not in natural coordinates. If the element is not in the x-y-plane then I am not convinced that the z-component of the element normal vanishs.

chleh: > If the element is not in the x-y-plane then I am not convinced that the z-component of the element normal vanishs.

But then you are not in 2D (at least for my understanding). OK, there is the case of having 2D elements in 3D space... Maybe leave it for later.

ProcessLib/BalanceProcess/BalanceProcess.h
```diff
@@ -0,0 +1,46 @@
...
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef PROCESS_LIB_BALANCEPROCESS_H_
+#define PROCESS_LIB_BALANCEPROCESS_H_
+
+#include "NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.h"
```
norihiro-w: it's not used here

ProcessLib/BalanceProcess/BalanceProcess.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+    // needed for creation of local assemblers
+    std::unique_ptr<NumLib::LocalToGlobalIndexMap const> dof_table(
+        new NumLib::LocalToGlobalIndexMap(std::move(all_mesh_subsets),
+                                          NumLib::ComponentOrder::BY_LOCATION));
+
+    boost::optional<MeshLib::PropertyVector<std::size_t> const&>
+        bulk_element_ids(boundary_mesh.getProperties()
+                             .template getPropertyVector<std::size_t>(
+                                 "OriginalSubsurfaceElementIDs"));
```
norihiro-w: i think Subsurface is too specific. why not OriginalElementID?

ProcessLib/BalanceProcess/MapBulkElementPoint.cpp
```diff
@@ -0,0 +1,96 @@
...
+    case 0:
+        return MathLib::Point3d{std::array<double, 3>{{wp[0], 0.0, 0.0}}};
+    case 1:
+        return MathLib::Point3d{std::array<double, 3>{{1.0, wp[0], 0.0}}};
+    case 2:
+        return MathLib::Point3d{std::array<double, 3>{{1.0 - wp[0], 1.0, 0.0}}};
+    case 3:
+        return MathLib::Point3d{std::array<double, 3>{{0.0, 1.0 - wp[0], 0.0}}};
+    default:
+        OGS_FATAL("Invalid face id '%u' for the quad.");
```
norihiro-w: argument is missing

ProcessLib/BalanceProcess/MapBulkElementPoint.h
```diff
@@ -0,0 +1,61 @@
...
+#ifndef PROCESSLIB_BALANCEPROCESS_MAPBULKELEMENT_H
+#define PROCESSLIB_BALANCEPROCESS_MAPBULKELEMENT_H
+
+#include <array>
+
+#include "MathLib/TemplateWeightedPoint.h"
+
+#include "MeshLib/Elements/FaceRule.h"
+#include "MeshLib/Elements/Elements.h"
+
```
norihiro-w: i would add namespace ProcessLib

ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
```diff
@@ -117,31 +117,58 @@ class LocalAssemblerData
...
+            &_element));
+
+        typename ShapeMatricesType::ShapeMatrices shape_matrices(
+            ShapeFunction::DIM, GlobalDim, ShapeFunction::NPOINTS);
+
+        fe.computeShapeFunctions(p_local_coords.getCoords(), shape_matrices,
+                                 GlobalDim);
+        std::vector<double> gradient;
+        gradient.resize(3);
+        Eigen::Map<Eigen::RowVectorXd>(gradient.data(), gradient.size()) =
```
norihiro-w: why is Conductivity not included?

ProcessLib/GroundwaterFlow/GroundwaterFlowProcess.h
```diff
@@ -34,14 +38,55 @@ class GroundwaterFlowProcess final : public Process
...
+        return _local_assemblers[element_id]->getFlux(p, local_x);
+    }
+
+    void postTimestepConcreteProcess(GlobalVector const& x) override
+    {
+        if (_balance_mesh) // computing the balance is optional
+        {
+            std::vector<double> init_values(
+                _balance_mesh->getNumberOfElements(), 0.0);
+            MeshLib::addPropertyToMesh(*_balance_mesh, _balance_pv_name,
```
norihiro-w: does it add a new property every time step?

ProcessLib/Process.h
```diff
@@ -175,13 +182,13 @@ class Process
 
     VectorMatrixAssembler _global_assembler;
 
+    /// Variables used by this process.
+    std::vector<std::reference_wrapper<ProcessVariable>> _process_variables;
```
norihiro-w: isn't it possible to use getProcessVariables()?

ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
TomFischer: Until know for the simple tests conductivity was always 1. I'll add conductivity to the formula.

ProcessLib/GroundwaterFlow/GroundwaterFlowProcess.h
TomFischer: Here I assume that `postTimeStepConcreteProcess()` is called only once. If it is called for every time step  a new name for PropertyVector has to be generated.

ProcessLib/Process.h
TomFischer: Yes, it is possible. I revert my change and use `getProcessVariables()`.

ProcessLib/BalanceProcess/BalanceProcess.cpp
TomFischer: If I change the name here, I have to change also the name in the tool ExtractSurface. If you really want change the name I would do this in a separate PR.

ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
```diff
@@ -135,13 +136,21 @@ class LocalAssemblerData
...
-        return gradient;
+        // fetch hydraulic conductivity
+        SpatialPosition pos;
+        pos.setElementID(_element.getID());
+        // TODO remove follwing line if time dependency is implemented
+        double const t = 0.0;
+        auto const k = _process_data.hydraulic_conductivity(t, pos)[0];
+
+        Eigen::Map<Eigen::RowVectorXd>(flux.data(), flux.size()) =
+            k * shape_matrices.dNdx *
```
norihiro-w: we need a negative sign here, q = - K * grad(head)

ProcessLib/BalanceProcess/BalanceProcess.cpp
norihiro-w: I think it's better to make the name general, but it's not urgent issue. One can change it later when they need it.

ProcessLib/GroundwaterFlow/GroundwaterFlowProcess.h
norihiro-w: is it because GroundwaterFlowProcess solves only one time step? I though `postTimeStepConcreteProcess()` is called for every time step...

chleh: Obviously it has to be configurable at which timesteps the surface fluxes will be computed.
I suggest to implement it in a follow-up PR, and to leave the current implementation as it is.

TomFischer: Yes, here I assumed that GroundwaterFlowProcess solves only one time step.

ProcessLib/BalanceProcess/BalanceProcess.cpp
TomFischer: Opened #1449.

norihiro-w:
:+1: from my side

TomFischer:
Thanks for the review. Will start rebase and renaming ...

PR #1426, opened 2016-09-23
Add GroupBasedParameter class

norihiro-w:
As far as I saw the code, there is no parameter class for material ID dependent values. This PR tries to implement it. Because material IDs are stored in a property vector, I generalized the class to accept any property vector.

Update: The parameter name is changed to GroupBasedParameter after talking with @endJunction

PR #1430, opened 2016-09-23
Another implementation of material ID dependent parameter

norihiro-w:
As suggested by @TomFischer in #1426. This PR uses/extends existing `PropertyVector<T*>` and `MeshElementParameter`. I had to modify `PropertyVector<T*>` to support multi-component values. 

Question is now which one, #1426 or this one, should we take?

### DIFF
--- a/MeshLib/PropertyVector.h
+++ b/MeshLib/PropertyVector.h
@@ -138,7 +138,7 @@ public:
     ~PropertyVector()
     {
         for (auto v : _values)
-            delete v;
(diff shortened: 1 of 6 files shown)


chleh:
I don't like `PropertyVector<T*>` (its implementation is not very intuitive; I don't mean this PR!), so I prefer your old PR.

norihiro-w:
@chleh I take your suggestion because nobody else has preference.

PR #1431, opened 2016-09-23
fix CMake error with VTK shared libs

### LINKED ISSUE #1408: CMake fails if BUILD_SHARED_LIBS=ON and VTK_DIR is not in PATH
Precisely, `get_prerequisites()` in  `InstallDependencies.cmake` fails when it tries to locate linked VTK libraries.

norihiro-w:
fix #1408


bilke:
Looks harmless.

PR #1432, opened 2016-09-23
Abort in OGS_FATAL()

norihiro-w:
This PR adds a new CMake option `OGS_FATAL_ABORT` which enables calling `abort()` in `OGS_FATAL()` instead of throwing exceptions.  The changes is helpful in debugging because it stops at at abort() and can backtrace from there.

### DIFF
--- a/BaseLib/Error.h
+++ b/BaseLib/Error.h
@@ -11,29 +11,30 @@
 #define BASELIB_ERROR_H
 
 #ifdef OGS_FATAL_ABORT
+
 #include <cstdlib>
-#endif
-#include <stdexcept>
-
-#ifdef OGS_FATAL_ABORT
 #include <logog/include/logog.hpp>
-#endif
-
 #include "StringTools.h"
-#include "FileTools.h"
 
-#define OGS_STR(x) #x
-#define OGS_STRINGIFY(x) OGS_STR(x)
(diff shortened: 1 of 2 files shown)


BaseLib/Error.h
```diff
@@ -10,15 +10,30 @@
 #ifndef BASELIB_ERROR_H
 #define BASELIB_ERROR_H
 
+#ifdef OGS_FATAL_ABORT
+#include <cstdlib>
+#endif
 #include <stdexcept>
```
endJunction: Maybe the exceptions are no longer needed in the case of OGS_FATAL_ABORT...

TomFischer:
Why is it not possible to set a breakpoint to the line where the exception is thrown? Then backtracing is possible, isn't it?

norihiro-w:
@TomFischer of course it's possible, but is inconvenient and inefficient for me. i'm using QtCreator and would like to see backtrack immediately whenever an error occur during debugging, without restarting with a breakpoint, or without modifying gdb config.

TomFischer:
Thanks for the explanation.

BaseLib/Error.h
norihiro-w: corrected

chleh:
@norihiro-w Just a minor notice: GDB with QtCreator can stop at any throw. You only need to add one special breakpoint for that, and if you don't remove that break point there is no need to restart in order to debug.

PR #1433, opened 2016-09-23
Fix LocalToGlobalIndexMap with mutliple variables and with multiple componets

norihiro-w:
This PR fixes a problem that `LocalToGlobalIndexMap` cannot correctly handle multiple variables & multiple variable components cases (e.g. pressure and displacement vector in HM). Important changes to the interface is that you need to pass a vector of the number of variable components to `LocalToGlobalIndexMap` constructor, if you have any variables having more than one components.

### DIFF
--- a/NumLib/DOF/LocalToGlobalIndexMap.cpp
+++ b/NumLib/DOF/LocalToGlobalIndexMap.cpp
@@ -48,35 +48,54 @@ LocalToGlobalIndexMap::findGlobalIndices(
 LocalToGlobalIndexMap::LocalToGlobalIndexMap(
     std::vector<std::unique_ptr<MeshLib::MeshSubsets>>&& mesh_subsets,
     NumLib::ComponentOrder const order)
+    : LocalToGlobalIndexMap(std::move(mesh_subsets), std::vector<unsigned>(mesh_subsets.size(), 1), order)
+{
+}
+
+
+LocalToGlobalIndexMap::LocalToGlobalIndexMap(
+    std::vector<std::unique_ptr<MeshLib::MeshSubsets>>&& mesh_subsets,
+    std::vector<unsigned> const& vec_var_n_components,
+    NumLib::ComponentOrder const order)
     : _mesh_subsets(std::move(mesh_subsets)),
       _mesh_component_map(_mesh_subsets, order)
 {
+    // construct a mapping table to get a global component ID from a variableID & a variable comp ID
+    _map_varCompID_to_globalCompID.resize(vec_var_n_components.size());
+    {
+        unsigned global_component_ID = 0;
+        for (unsigned i=0; i<vec_var_n_components.size(); i++)
+            for (unsigned j=0; j<vec_var_n_components[i]; j++)
+                _map_varCompID_to_globalCompID[i].push_back(global_component_ID++);
+    }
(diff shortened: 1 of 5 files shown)


norihiro-w:
I will rework on this after #1437

norihiro-w:
reworked after #1437

norihiro-w:
@endJunction For better performance (though I haven't tested), I changed it back to use the original offset vector ` _variable_component_offsets`.

NumLib/DOF/LocalToGlobalIndexMap.cpp
```diff
@@ -12,6 +12,28 @@
 namespace NumLib
 {
 
+namespace
+{
+
+// Make the cumulative sum of an array
```
chleh: Minor: Maybe add a comment that the cumulative sums start with zero.
Also minor IMHO: Could be done with std::partial_sum

NumLib/DOF/LocalToGlobalIndexMap.h
```diff
@@ -73,6 +89,14 @@ class LocalToGlobalIndexMap final
 
     std::size_t size() const;
 
+    std::size_t getNumberOfVariables() const { return (_variable_component_offsets.size() - 1); }
+
+    std::size_t getNumberOfVariableComponents(int variable_id) const
+    {
+        assert(static_cast<unsigned>(variable_id) < getNumberOfVariables());
```
chleh: Minor: I think std::vector already asserts this in debug mode in the line below.

Tests/NumLib/LocalToGlobalIndexMap.cpp
```diff
@@ -107,12 +107,41 @@ TEST_F(NumLibLocalToGlobalIndexMapTest, DISABLED_SubsetByComponent)
...
     ASSERT_EQ(selected_nodes.size(), dof_map_subset->dofSizeWithGhosts());
 }
+
+#ifndef USE_PETSC
+TEST_F(NumLibLocalToGlobalIndexMapTest, MultipleVariablesMultipleComponents)
+#else
+TEST_F(NumLibLocalToGlobalIndexMapTest, DISABLED_MultipleVariablesMultipleComponents)
+#endif
+{
+    // test 2 variables (1st variable with 1 component, 2nd variable with 2 components)
```
chleh: Didn't that already work with the unfixed DOF table? (I think @endJunction said so, if I remember correctly)
1st var 2 comp, 2nd var 1 comp didn't work, did it?

NumLib/DOF/LocalToGlobalIndexMap.h
norihiro-w: it depends on a compiler

NumLib/DOF/LocalToGlobalIndexMap.cpp
norihiro-w: thanks for the hint! I updated

Tests/NumLib/LocalToGlobalIndexMap.cpp
norihiro-w: you are right, Dima told me it already worked if 1st var has 1 comp. I added another test with 1st var with two components.

PR #1434, opened 2016-09-23
Material model for fracture deformation

norihiro-w:
This PR adds two models for fracture deformation, linear elastic and Mohr-Coulomb.

Remarks
- Mohr-Coulomb model is limited to 2D case
- Fixed-size Eigen matrix/vectors are not considered. let me know if it's really important...

### DIFF
--- a/MaterialLib/FractureModels/CreateLinearElasticIsotropic.cpp
+++ b/MaterialLib/FractureModels/CreateLinearElasticIsotropic.cpp
@@ -25,17 +25,11 @@ createLinearElasticIsotropic(
     config.checkConfigParameter("type", "LinearElasticIsotropic");
     DBUG("Create LinearElasticIsotropic material");
 
-    // Youngs modulus
     auto& Kn = ProcessLib::findParameter<double>(
-        config, "kn", parameters, 1);
+        config, "normal_stiffness", parameters, 1);
 
-    DBUG("Use '%s' as Kn parameter.", Kn.name.c_str());
-
-    // Poissons ratio
     auto& Ks = ProcessLib::findParameter<double>(
-        config, "ks", parameters, 1);
-
-    DBUG("Use '%s' as Ks parameter.", Ks.name.c_str());
+        config, "shear_stiffness", parameters, 1);
 
     typename LinearElasticIsotropic<DisplacementDim>::MaterialProperties mp{
         Kn, Ks};
--- a/MaterialLib/FractureModels/CreateLinearElasticIsotropic.cpp
+++ b/MaterialLib/FractureModels/CreateLinearElasticIsotropic.cpp
@@ -1,0 +1,61 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#include "CreateLinearElasticIsotropic.h"
+
+#include "LinearElasticIsotropic.h"
+
+namespace MaterialLib
+{
+namespace Fracture
+{
+
+template <int DisplacementDim>
+std::unique_ptr<FractureModelBase<DisplacementDim>>
+createLinearElasticIsotropic(
+    std::vector<std::unique_ptr<ProcessLib::ParameterBase>> const& parameters,
+    BaseLib::ConfigTree const& config)
+{
+    config.checkConfigParameter("type", "LinearElasticIsotropic");
+    DBUG("Create LinearElasticIsotropic material");
+
+    // Youngs modulus
+    auto& Kn = ProcessLib::findParameter<double>(
+        config, "kn", parameters, 1);
+
+    DBUG("Use '%s' as Kn parameter.", Kn.name.c_str());
+
+    // Poissons ratio
+    auto& Ks = ProcessLib::findParameter<double>(
+        config, "ks", parameters, 1);
+
+    DBUG("Use '%s' as Ks parameter.", Ks.name.c_str());
+
+    typename LinearElasticIsotropic<DisplacementDim>::MaterialProperties mp{
+        Kn, Ks};
+
+    return std::unique_ptr<LinearElasticIsotropic<DisplacementDim>>{
+        new LinearElasticIsotropic<DisplacementDim>{mp}};
+}
+
+
+template
+std::unique_ptr<FractureModelBase<2>>
+createLinearElasticIsotropic(
+    std::vector<std::unique_ptr<ProcessLib::ParameterBase>> const& parameters,
+    BaseLib::ConfigTree const& config);
+
+template
+std::unique_ptr<FractureModelBase<3>>
+createLinearElasticIsotropic(
+    std::vector<std::unique_ptr<ProcessLib::ParameterBase>> const& parameters,
+    BaseLib::ConfigTree const& config);
+
+}  // namespace Fracture
+}  // namespace MaterialLib
--- a/MaterialLib/FractureModels/CreateLinearElasticIsotropic.cpp
+++ b/MaterialLib/FractureModels/CreateLinearElasticIsotropic.cpp
@@ -9,6 +9,7 @@
 
 #include "CreateLinearElasticIsotropic.h"
 
+#include "ProcessLib/Utils/ProcessUtils.h"  // required for findParameter
 #include "LinearElasticIsotropic.h"
 
 namespace MaterialLib
--- a/MaterialLib/FractureModels/FractureModelBase.h
+++ b/MaterialLib/FractureModels/FractureModelBase.h
@@ -1,0 +1,40 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#pragma once
+
+#include <Eigen/Eigen>
+
+#include "ProcessLib/Parameter/Parameter.h"
+
+namespace MaterialLib
+{
+namespace Fracture
+{
+
+template <int DisplacementDim>
+class FractureModelBase
+{
+public:
+    virtual ~FractureModelBase() {}
+
+    virtual void computeConstitutiveRelation(
+            double const t,
+            ProcessLib::SpatialPosition const& x,
+            Eigen::Ref<Eigen::VectorXd const> w_prev,
+            Eigen::Ref<Eigen::VectorXd const> w,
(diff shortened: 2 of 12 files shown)


MaterialLib/FractureModels/CreateLinearElasticIsotropic.cpp
```diff
@@ -0,0 +1,61 @@
...
+template <int DisplacementDim>
+std::unique_ptr<FractureModelBase<DisplacementDim>>
+createLinearElasticIsotropic(
+    std::vector<std::unique_ptr<ProcessLib::ParameterBase>> const& parameters,
+    BaseLib::ConfigTree const& config)
+{
+    config.checkConfigParameter("type", "LinearElasticIsotropic");
+    DBUG("Create LinearElasticIsotropic material");
+
+    // Youngs modulus
```
wenqing: Young's

MaterialLib/FractureModels/CreateLinearElasticIsotropic.cpp
```diff
@@ -0,0 +1,61 @@
...
+    auto& Kn = ProcessLib::findParameter<double>(
```
wenqing: const?

MaterialLib/FractureModels/CreateLinearElasticIsotropic.cpp
```diff
@@ -0,0 +1,61 @@
...
+        config, "kn", parameters, 1);
+
+    DBUG("Use '%s' as Kn parameter.", Kn.name.c_str());
+
+    // Poissons ratio
```
wenqing: Poisson's

MaterialLib/FractureModels/CreateLinearElasticIsotropic.cpp
norihiro-w: removed

MaterialLib/FractureModels/CreateLinearElasticIsotropic.cpp
norihiro-w: not necessary. the variable is only passed to MaterialProperties constructor.

MaterialLib/FractureModels/CreateLinearElasticIsotropic.cpp
norihiro-w: removed

endJunction:
@nagelt Could you have a look onto the material models please?

bilke:
I guess we are hitting [this Qt bug](https://bugreports.qt.io/browse/QTBUG-22829). I also guess the workaround would be to add the proposed `#ifndef Q_MOC_RUN` as a wrapper around the Boost include in `MathTools.h`.

norihiro-w:
@bilke thanks! The workaround solved the error.

nagelt:
Looks good to me. The code can be followed easily.

Fixed size matrices would be sigma/w of (displacement_dim x 1) and the various K/C matrices (displacement_dim x displacement_dim), right? So it's one implementation for MC (2) and two for linear elastic (2,3). @endJunction coming back to Nori's question: How much do we need fixed size when you look at our standard M implementations?

Aside from that I have no further comments on the content.

MaterialLib/FractureModels/MohrCoulomb.cpp
```diff
@@ -0,0 +1,115 @@
...
+        OGS_FATAL("MohrCoulomb fracture model does not support 3D case.");
+        return;
+    }
+    MaterialPropertyValues const mat(_mp, t, x);
+    Eigen::VectorXd const dw = w - w_prev;
+
+    Eigen::MatrixXd Ke(2,2);
+    Ke.setZero();
+    Ke(0,0) = mat.Ks;
+    Ke(1,1) = mat.Kn;
```
chleh: Optimization idea: Maybe there is something possible with Eigen::MatrixXd::Diagonal(...) or similar.

MaterialLib/FractureModels/MohrCoulomb.cpp
```diff
@@ -0,0 +1,115 @@
...
+    // if opening
+    if (sigma[1] > 0)
+    {
+        Kep.setZero();
+        sigma.setZero();
+        return;
+    }
+
+    // check shear yield function (Fs)
+    double const Fs = std::abs(sigma[0]) + sigma[1] * tan(mat.phi) - mat.c;
```
chleh: std::tan, also below.

MaterialLib/FractureModels/FractureModelBase.h
```diff
@@ -0,0 +1,58 @@
...
+     * @param w_prev      fracture displacement at previous time step
+     * @param w           fracture displacement at current time step
+     * @param sigma_prev  stress at previous time step
+     * @param sigma       stress at current time step
+     * @param C           tangent matrix for stress and fracture displacements
+     */
+    virtual void computeConstitutiveRelation(
+            double const t,
+            ProcessLib::SpatialPosition const& x,
+            Eigen::Ref<Eigen::VectorXd const> w_prev,
```
norihiro-w: I'm using `Ref` because it can simply take both fixed size and dynamic size Eigen objects. If the performance is important (I doubt because the number of fracture elements are not many), I can change it as Dima did in `Solids::MechanicsBase`.

MaterialLib/FractureModels/LinearElasticIsotropic.cpp
```diff
@@ -0,0 +1,48 @@
...
+        Eigen::Ref<Eigen::VectorXd const> w_prev,
+        Eigen::Ref<Eigen::VectorXd const> w,
+        Eigen::Ref<Eigen::VectorXd const> sigma_prev,
+        Eigen::Ref<Eigen::VectorXd> sigma,
+        Eigen::Ref<Eigen::MatrixXd> C)
+{
+    const int index_ns = DisplacementDim - 1;
+    C.setZero();
+    for (int i=0; i<index_ns; i++)
+        C(i,i) = _mp.shear_stiffness(t, x)[0];
```
norihiro-w: i would postpone optimization until we figure out its necessity.

MaterialLib/FractureModels/MohrCoulomb.h
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    void computeConstitutiveRelation(
+            double const t,
+            ProcessLib::SpatialPosition const& x,
+            Eigen::Ref<Eigen::VectorXd const> w_prev,
+            Eigen::Ref<Eigen::VectorXd const> w,
+            Eigen::Ref<Eigen::VectorXd const> sigma_prev,
+            Eigen::Ref<Eigen::VectorXd> sigma,
+            Eigen::Ref<Eigen::MatrixXd> Kep) override;
```
norihiro-w: "e" means elastic, "p" means plastic. Kep means it includes both elastic and plastic parts.

MaterialLib/FractureModels/MohrCoulomb.cpp
```diff
@@ -0,0 +1,115 @@
...
+void MohrCoulomb<DisplacementDim>::computeConstitutiveRelation(
+        double const t,
+        ProcessLib::SpatialPosition const& x,
+        Eigen::Ref<Eigen::VectorXd const> w_prev,
+        Eigen::Ref<Eigen::VectorXd const> w,
+        Eigen::Ref<Eigen::VectorXd const> sigma_prev,
+        Eigen::Ref<Eigen::VectorXd> sigma,
+        Eigen::Ref<Eigen::MatrixXd> Kep)
+{
+    if (DisplacementDim == 3)
```
norihiro-w: i am not sure if it works. static_assert means we don't compile fracture stuff at all for 3D, but other fracture model such as elastic one can work with 3D.

MaterialLib/FractureModels/MohrCoulomb.cpp
norihiro-w: postponed

MaterialLib/FractureModels/MohrCoulomb.h
chleh: Thanks for the explanation.

MaterialLib/FractureModels/MohrCoulomb.cpp
chleh: I see.

MaterialLib/FractureModels/FractureModelBase.h
chleh: Sounds like a cool idea. However, I am not entirely sure if it reduces complexity (fixed & dynamic at the same time) or increases it (now we have to deal with Eigen::Ref, too).
Maybe at some point in the future there should be a discussion about the matrix types used in material models.

MaterialLib/FractureModels/MohrCoulomb.h
```diff
@@ -0,0 +1,85 @@
...
+    struct MaterialProperties
+    {
+        using P = ProcessLib::Parameter<double>;
+        using X = ProcessLib::SpatialPosition;
+
+        MaterialProperties(
+                P const& normal_stiffness, P const& shear_stiffness,
+                P const& friction_angle, P const& dilatancy_angle,
+                P const& cohesion)
+            : normal_stiffness(normal_stiffness), shear_stiffness(shear_stiffness),
```
endJunction: Small improvement:
A warning is issued "declaration shadows a field of MaterialLib::Fracture::MohrCoulomb::MaterialProperties" for all constructor's arguments. I usually postfix the arguments with underscore:
```c++
MaterialProperties(P const& normal_stiffness_) : normal_stiffness(normal_stiffness_) {}
```

MathLib/MathTools.h
```diff
@@ -111,6 +115,16 @@ double sqrDist(const double* p0, const double* p1)
...
 double getAngle (const double p0[3], const double p1[3], const double p2[3]);
 
+/// sign function
+template <typename T> int sgn(T val) {
+    return (T(0) < val) - (val < T(0));
+}
+
+/// converts the given degrees to radians
+inline double to_radians(double degrees) {
+    return degrees*boost::math::constants::pi<double>()/180.;
```
endJunction: FYI: There also is a direct pi/180 constant in boost called `radians`

MathLib/MathTools.h
```diff
@@ -111,6 +115,16 @@ double sqrDist(const double* p0, const double* p1)
  */
 double getAngle (const double p0[3], const double p1[3], const double p2[3]);
 
+/// sign function
+template <typename T> int sgn(T val) {
+    return (T(0) < val) - (val < T(0));
+}
```
endJunction: The sign function is already implemented in [boost](http://www.boost.org/doc/libs/1_50_0/libs/math/doc/sf_and_dist/html/math_toolkit/utils/sign_functions.html): `#include <boost/math/special_functions/sign.hpp>`
I'd prefer an existing function

MathLib/MathTools.h
norihiro-w: thanks for the info. but i don't see a reason to change current code.

PR #1436, opened 2016-09-23
[Jenkins] Exclude CLI from gcc Data Exlorer build.

### LINKED ISSUE #1416: Still issues with Jenkins
All PRs are currently failing: https://github.com/ufz/ogs/pulls

I am working on it.. hopefully I can fix it tomorrow.

bilke:
Possibly fixing sporadic ldd errors. Closes #1416.


endJunction:
Let's try this one, which I also like because it separates binaries with different requirements.
We can always go back to the full builds---this is rather a simple change.

PR #1437, opened 2016-09-23
Use mesh subsets instead of mesh nodes

endJunction:
When creating dof table, setting initial conditions, creating boundary conditions, and in output
currently all nodes of the mesh are involved. If the underlying mesh subsets do not consist of all nodes (e.g. only base nodes) then the mentioned algorithms do return wrong results.

Maybe there are more places with similar patterns, please tell me then.

### DIFF
--- a/MeshLib/MeshSubset.h
+++ b/MeshLib/MeshSubset.h
@@ -125,6 +125,12 @@ public:
         return _msh.getElements().cend();
     }
 
+    std::vector<Node*> const& getNodes() const
+    {
+        assert(_nodes);
+        return *_nodes;
+    }
+
     /// Constructs a new mesh subset which is a set intersection of the current
     /// nodes and the provided vector of nodes.
     /// An empty mesh subset may be returned, not a nullptr, in case of empty
--- a/NumLib/DOF/LocalToGlobalIndexMap.cpp
+++ b/NumLib/DOF/LocalToGlobalIndexMap.cpp
@@ -16,6 +16,7 @@ template <typename ElementIterator>
 void
 LocalToGlobalIndexMap::findGlobalIndices(
     ElementIterator first, ElementIterator last,
+    std::vector<MeshLib::Node*> const& nodes,
     std::size_t const mesh_id,
     const unsigned comp_id, const unsigned comp_id_write)
 {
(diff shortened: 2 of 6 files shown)


MeshLib/MeshSubset.h
```diff
@@ -125,6 +125,12 @@ class MeshSubset
         return _msh.getElements().cend();
     }
 
+    std::vector<Node*> getNodes() const
```
norihiro-w: why isn't it possible to return pointer or reference here?

NumLib/DOF/LocalToGlobalIndexMap.cpp
```diff
@@ -26,16 +27,16 @@ LocalToGlobalIndexMap::findGlobalIndices(
...
-        std::size_t const nnodes = (*e)->getNumberOfNodes();
-
         LineIndex indices;
-        indices.reserve(nnodes);
 
-        for (unsigned n = 0; n < nnodes; n++)
+        for (auto* n = (*e)->getNodes(); n < (*e)->getNodes()+(*e)->getNumberOfNodes(); ++n)
         {
+            // Check if the element's node is in the given list of nodes.
+            if (std::find(std::begin(nodes), std::end(nodes), *n) == std::end(nodes))
```
norihiro-w: TBD. My LIE implementation assumes `npos` is stored for those nodes.

chleh:
Last Friday you showed some HM-process results in Paraview. There the pressure values had a peculiar pattern. Will that pattern be observed also with the code from this PR?

endJunction:
@chleh For a given problem not using the full mesh (like omitting second order nodes) there is no sensible output on the whole mesh. What is coming to my mind is:
 - initialization with NaN of the output data
 - creation of a submesh corresponding to a mesh-subset
 - other ideas are welcome...

MeshLib/MeshSubset.h
endJunction: This is of course a mistake. I'll update this.

NumLib/DOF/LocalToGlobalIndexMap.cpp
norihiro-w: we agreed to keep this function as suggested here and created another function for handling my special case. I'll update my PR after this PR is merged.

ProcessLib/ProcessOutput.cpp
```diff
@@ -67,13 +67,15 @@ void doProcessOutput(
...
     int global_component_offset = 0;
     int global_component_offset_next = 0;
 
     // primary variables
-    for (ProcessVariable& pv : process_variables)
+    for (int variable_id = 0;
+         variable_id < static_cast<int>(process_variables.size());
+         ++variable_id)
     {
+        ProcessVariable& pv = process_variables[variable_id];
```
TomFischer: Could `pv` be const?

TomFischer:
:+1:

PR #1440, opened 2016-09-26
Supporting heterogeneously distributed DoF

norihiro-w:
follow up of #1433 

Untill now, `LocalToGlobalIndexMap` assumes every DoF is distributed over all mesh elements, and creates `_rows` object storing all the DoF data for every elements. However in my application, some DoF are defined only on lower-dimension mesh element nodes (e.g. lines in 2D case), and `_rows` object stores unnecessary empty DoF data for some elements. This PR tries to avoid creating the unnecessary DoF data with the following changes
- add a constructor which can take a vector of active elements for each DoF. "Active elements" means elements which contains DoF in at least one of its nodes and should be involved in local assembly.
- add another version of `findGlobalIndices()` (`findGlobalIndicesWithElementID()`), which creates `_rows` object from an element ID. The original version assumes the ID is equal to an index in a given vector, which is probably necessary for creating boundary conditions.

### DIFF
--- a/NumLib/DOF/LocalToGlobalIndexMap.cpp
+++ b/NumLib/DOF/LocalToGlobalIndexMap.cpp
@@ -30,6 +30,46 @@ std::vector<T> to_cumulative(std::vector<T> const& vec)
 } // no named namespace
 
 template <typename ElementIterator>
+void
+LocalToGlobalIndexMap::findGlobalIndicesWithElementID(
+    ElementIterator first, ElementIterator last,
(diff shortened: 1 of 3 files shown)


norihiro-w:
i will rework on this after #1437

norihiro-w:
The elements of MeshSubset mean they are included in DoF. If active elements, which are not Dof but used for local assembly, are to be included there, we need to add another variables and functions.

endJunction:

PR #1441, opened 2016-09-26
Fix PropertyVector<T*> for multi-component case

norihiro-w:
Currently `PropertyVector<T*> ` works only with a single component.

### DIFF
--- a/MeshLib/PropertyVector.h
+++ b/MeshLib/PropertyVector.h
@@ -197,7 +197,8 @@ public:
         );
         // copy pointers to property values
         for (std::size_t j(0); j<_values.size(); j++) {
-            t->initPropertyValue(j, *(_values[j]));
+            std::vector<T> values(_values[j], _values[j] + _n_components);
+            t->initPropertyValue(j, values);
         }
         return t;
     }
--- a/MeshLib/PropertyVector.h
+++ b/MeshLib/PropertyVector.h
@@ -157,7 +157,8 @@ public:
     void initPropertyValue(std::size_t group_id, T const& value)
     {
         if (_n_components != 1)
-            OGS_FATAL("Single-component version of initPropertyValue() is called for a multi-components PropertyVector<T*>");
+            OGS_FATAL("Single-component version of initPropertyValue() is called "
+                      "for a multi-components PropertyVector<T*>");
         T* p = new T[1];
         p[0] = value;
         _values[group_id] = p;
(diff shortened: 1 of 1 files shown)


MeshLib/PropertyVector.h
```diff
@@ -185,6 +197,20 @@ friend class Properties;
         return t;
     }
 
+    //! Returns the value for the given component stored in the given tuple.
+    T const& getComponent(std::size_t tuple_index,
+                                      std::size_t component) const
```
TomFischer: At least on github the indentation seems to be not correct.

MeshLib/PropertyVector.h
```diff
@@ -155,13 +156,24 @@ friend class Properties;
...
     {
-        _values[group_id] = new T(value);
+        T* p = new T[1];
+        p[0] = value;
+        _values[group_id] = p;
+    }
+
+    void initPropertyValue(std::size_t group_id, std::vector<T> const& values)
+    {
+        T* p = new T[values.size()];
```
TomFischer: Maybe a check `values.size() == _n_components` could be inserted?

MeshLib/PropertyVector.h
```diff
@@ -155,13 +156,24 @@ friend class Properties;
 
     void initPropertyValue(std::size_t group_id, T const& value)
     {
-        _values[group_id] = new T(value);
+        T* p = new T[1];
```
TomFischer: Maybe a check `_n_components == 1` could be inserted?

TomFischer:
If tests are ok: :+1:

MeshLib/PropertyVector.h
```diff
@@ -155,13 +156,31 @@ friend class Properties;
 
     void initPropertyValue(std::size_t group_id, T const& value)
     {
-        _values[group_id] = new T(value);
+        if (_n_components != 1)
+            OGS_FATAL("Single-component version of initPropertyValue() is called "
+                      "for a multi-components PropertyVector<T*>");
+        T* p = new T[1];
+        p[0] = value;
```
chleh: Note: This will not work if `T` is an abstract class or `T` is not copyable. However, it did not work either before.

norihiro-w:
@chleh As far as I know, `PropertyVector<T*>` is used only in tests at the moment. I fully agree with your concerns. It should be discussed at some point.

PR #1442, opened 2016-09-26
Move dof table functionality

TomFischer:
As titled.

### DIFF
--- a/NumLib/DOF/DOFTableUtil.cpp
+++ b/NumLib/DOF/DOFTableUtil.cpp
@@ -44,6 +44,26 @@ std::vector<GlobalIndexType> getIndices(
     return indices;
 }
 
+NumLib::LocalToGlobalIndexMap::RowColumnIndices getRowColumnIndices(
+    std::size_t const id,
+    NumLib::LocalToGlobalIndexMap const& dof_table,
+    std::vector<GlobalIndexType>& indices)
+{
+    assert(dof_table.size() > id);
+    indices.clear();
+
+    // Local matrices and vectors will always be ordered by component,
+    // no matter what the order of the global matrix is.
+    for (unsigned c = 0; c < dof_table.getNumberOfComponents(); ++c)
+    {
+        auto const& idcs = dof_table(id, c).rows;
+        indices.reserve(indices.size() + idcs.size());
+        indices.insert(indices.end(), idcs.begin(), idcs.end());
+    }
+
+    return NumLib::LocalToGlobalIndexMap::RowColumnIndices(indices, indices);
+}
+
 double norm1(GlobalVector const& x, unsigned const global_component,
              LocalToGlobalIndexMap const& dof_table, MeshLib::Mesh const& mesh)
 {
--- a/NumLib/DOF/DOFTableUtil.h
+++ b/NumLib/DOF/DOFTableUtil.h
@@ -28,6 +28,13 @@ std::vector<GlobalIndexType> getIndices(
     std::size_t const mesh_item_id,
     NumLib::LocalToGlobalIndexMap const& dof_table);
 
+//! Returns row/column indices for the item identified by \c id from the
+//! given \c dof_table.
+LocalToGlobalIndexMap::RowColumnIndices getRowColumnIndices(
+    std::size_t const id,
+    NumLib::LocalToGlobalIndexMap const& dof_table,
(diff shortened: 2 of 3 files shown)


chleh:
Please have a look at [DOFTableUtil](/ufz/ogs/tree/master/NumLib/DOF/DOFTableUtil.cpp). There is already similar code.

TomFischer:
> Please have a look at DOFTableUtil. There is already similar code.

Thanks for the hint, I'll check this tomorrow.

NumLib/DOF/DOFTableUtil.cpp
```diff
@@ -44,6 +44,42 @@ std::vector<GlobalIndexType> getIndices(
     return indices;
 }
 
+NumLib::LocalToGlobalIndexMap::RowColumnIndices getRowColumnIndices(
```
norihiro-w: It looks for me the function is similar to `getIndices()`. Can't you reuse it?

NumLib/DOF/DOFTableUtil.cpp
```diff
@@ -44,6 +44,42 @@ std::vector<GlobalIndexType> getIndices(
...
+    {
+        auto const& idcs = dof_table(id, c).rows;
+        indices.reserve(indices.size() + idcs.size());
+        indices.insert(indices.end(), idcs.begin(), idcs.end());
+    }
+
+    return NumLib::LocalToGlobalIndexMap::RowColumnIndices(indices, indices);
+}
+
+void getVectorValues(
```
norihiro-w: `GlobalVector` such  as `EigenVector` has `get(std::vector<IndexType> const& indices)`. maybe you can use it.

NumLib/DOF/DOFTableUtil.cpp
TomFischer: Indeed these functions are very similar. The important difference is that the `getRowColumnIndices` does not create the result vector `indices` again and again for every call of the function. Anyway, in this PR I only want to move existing code such that I can use it a second time at another place.

NumLib/DOF/DOFTableUtil.cpp
TomFischer: Yes, you are right. The function `get()` is already used internally in `getVectorValues()`, see line 78. The function is a wrapper for the `GlobalVector::get`  that ensures that the vector `local_x` has the appropriate size.

norihiro-w: is `get()` at line 78 `get(int)` or `get(std::vector<int>&)`?

NumLib/DOF/DOFTableUtil.cpp
norihiro-w: i see. thanks for the explanation

NumLib/DOF/DOFTableUtil.cpp
TomFischer: If I am not wrong it is `IDX_TYPE` which is probably `int`.

norihiro-w: what I actually wanted to say is that you can write it as below. maybe you already know it and intentionally you are not using it. 

```
local_x = x.get(r_c_indices.rows);
```

TomFischer: Thanks for the hint, I did not know this. I needed only the functionality in my own balance implementation. So I moved the code to a general accessible location.

TomFischer:
@chleh Are the changes in TES okay?

chleh:
Changes in TES are OK. Code looks good. So you might merge this PR.

I only want to make a short remark:
The functionality added is, as you said, that `getRowColumnIndices()` doesn't create the indices vector. However `x.get(r_c_indices.rows)` will always create a vector.
So if this optimization is really important, then it might be important in both places (note: IMHO it's currently not so important). Furthermore, if it's important, it might be important for `getIndices()`, too. And when changing `getIndices()`, one would have to check what that means for the usability of this function, i.e., whether the user has to write more code.
So this PR duplicates some functionality; I am not really happy with it. But in order to not block you: Merge and forget :smile:

TomFischer:
I am unsure. Should I let `getVectorValues()` in the TES process?

chleh:
For TES you can keep it as you have it in this PR.

PR #1443, opened 2016-09-26
Solve Axially Symmetric Problems

### LINKED ISSUE #1310: Extend shape functions for axial-symmetric problems.
This feature will be required sooner or later for axial-symmetric problems.

I start collecting the necessary changes here:
 - Needs modification of the gradients
 - Needs modification of the Neumann assembler to include 2*pi*r multiplication.

chleh:
Closes #1310.

The coordinates are mapped as follows:
Cartesian x to cylindrical r, Cartesian y to cylindrical z, i.e., axial symmetry is about the Cartesian y axis.

The solution in this PR is rather ad-hoc, so maybe some discussions are needed.

Major changes:
* Mesh now has flag `is_axially_symmetric`
* Constructors of all processes and local assemblers now have an argument telling if the problem is axially symmetric.
* The ShapeMatrices structure is extended by an `integralMeasure` member, which is 1 in the Cartesian case and 2 * pi * r in the axial case.

The added test results for GWFlow, T and TES are compared to numerical data from a wedge-shaped domain. I've added the corresponding wedge-files to the data repo and commented test cases Tests.cmake.

### DIFF
--- a/Applications/CLI/Tests.cmake
+++ b/Applications/CLI/Tests.cmake
@@ -416,6 +416,110 @@ if(NOT OGS_USE_MPI)
         cube_1e3_expected_pcs_0_ts_101_t_1.000000.vtu cube_1e3_pcs_0_ts_101_t_1.000000.vtu displacement displacement
     )
 
+    # SQUARE 1x1 GROUNDWATER FLOW TEST -- AXIALLY SYMMETRIC
+    # test results are compared to 3D simulation on a wedge-shaped domain
+    AddTest(
+        NAME GroundWaterFlowProcess_square_1x1_1e2_axi
+        PATH Elliptic/square_1x1_GroundWaterFlow
+        EXECUTABLE ogs
+        EXECUTABLE_ARGS square_1e2_axi.prj
+        WRAPPER time
+        TESTER vtkdiff
+        ABSTOL 1.6e-5 RELTOL 1e-5
+        DIFF_DATA
+        wedge-1e2-ang-0.02-surface.vtu square_1e2_axi_pcs_0_ts_1_t_1.000000.vtu temperature temperature
+    )
+    # # WEDGE 1x1 GROUNDWATER FLOW TEST -- computes reference results for the above test
+    # AddTest(
+    #     NAME GroundWaterFlowProcess_wedge_1e2_ang_0.02
+    #     PATH Elliptic/square_1x1_GroundWaterFlow
+    #     EXECUTABLE ogs
+    #     EXECUTABLE_ARGS wedge_1e2_axi_ang_0.02.prj
+    # )
+
+    # SQUARE 1x1 GROUNDWATER FLOW TEST -- AXIALLY SYMMETRIC
+    # test results are compared to 3D simulation on a wedge-shaped domain
+    AddTest(
+        NAME GroundWaterFlowProcess_square_1x1_1e4_axi_ang_0.02
+        PATH Elliptic/square_1x1_GroundWaterFlow
+        EXECUTABLE ogs
+        EXECUTABLE_ARGS square_1e4_axi.prj
+        WRAPPER time
+        TESTER vtkdiff
+        ABSTOL 1.6e-5 RELTOL 1e-5
+        DIFF_DATA
+        wedge-1e4-ang-0.02-surface.vtu square_1e4_axi_pcs_0_ts_1_t_1.000000.vtu temperature temperature
+    )
+    # # WEDGE 1x1 GROUNDWATER FLOW TEST -- computes reference results for the above test
+    # AddTest(
+    #     NAME GroundWaterFlowProcess_wedge_1e4_ang_0.02
+    #     PATH Elliptic/square_1x1_GroundWaterFlow
+    #     EXECUTABLE ogs
+    #     EXECUTABLE_ARGS wedge_1e4_axi_ang_0.02.prj
+    # )
+
+    # SMALL DEFORMATION TEST -- AXIALLY SYMMETRIC
+    AddTest(
+        NAME SmallDeformation_ring_plane_strain_axi
+        PATH Mechanics/Linear
+        EXECUTABLE ogs
+        EXECUTABLE_ARGS ring_plane_strain.prj
+        WRAPPER time
+        TESTER vtkdiff
+        ABSTOL 6e-4 RELTOL 1e-4
+        DIFF_DATA
+        ring_plane_strain_1e4_solution.vtu ring_plane_strain_pcs_0_ts_1_t_1.000000.vtu u displacement
+        ring_plane_strain_1e4_solution.vtu ring_plane_strain_pcs_0_ts_1_t_1.000000.vtu sigma_rr sigma_rr
+        ring_plane_strain_1e4_solution.vtu ring_plane_strain_pcs_0_ts_1_t_1.000000.vtu sigma_ff sigma_ff
+    )
+
+    # SQUARE 1x1 HEAT CONDUCTION TEST -- AXIALLY SYMMETRIC
+    # test results are compared to 3D simulation on a wedge-shaped domain
+    AddTest(
+         NAME 2D_HeatConduction_axi
+         PATH Parabolic/T/2D_axially_symmetric
+         EXECUTABLE ogs
+         EXECUTABLE_ARGS square_1e2_axi.prj
+         WRAPPER time
+         TESTER vtkdiff
+         ABSTOL 1.7e-5 RELTOL 1e-5
+         DIFF_DATA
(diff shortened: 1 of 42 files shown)


NumLib/Fem/CoordinatesMapping/ShapeMatrices-impl.h
```diff
@@ -64,6 +64,7 @@ inline void setZero(ShapeMatrices<T_N, T_DNDR, T_J, T_DNDX> &shape, ShapeDataFie
     setZero(shape, ShapeDataFieldType<ShapeMatrixType::DNDR>());
     setMatrixZero(shape.J);
     shape.detJ = .0;
+    shape.integralMeasure = 0.0;
```
wenqing: Why not 1.0?

MeshLib/Mesh.h
```diff
@@ -125,6 +125,16 @@ class Mesh : BaseLib::Counter<Mesh>
     MeshLib::Properties & getProperties() { return _properties; }
     MeshLib::Properties const& getProperties() const { return _properties; }
 
+    bool isAxiallySymmetric() const { return _is_axially_symmetric; }
+    void setAxiallySymmetric(bool is_axial_symmetric) {
```
wenqing: It could also to 1D mesh (excluding that for the deformation problem)

NumLib/Fem/FiniteElement/TemplateIsoparametric.h
```diff
@@ -101,13 +103,40 @@ class TemplateIsoparametric
...
+    double interpolateZerothCoordinate(
+        typename ShapeMatrices::ShapeType const& N) const
+    {
+        auto* nodes = _ele->getNodes();
+        typename ShapeMatrices::ShapeType rs(N.size());
+        for (int i=0; i<rs.size(); ++i) {
+            rs[i] = (*nodes[i])[0];
+        }
+        auto const r = N.dot(rs);
+        return r;
```
wenqing: need
```
if (r == 0.)
  r = 0.5/pi
```
for line integration at r = 0, which is used to apply a distributed source term along the axis.

In addition the case of integration point at edges (only exist in triangle) must be avoid, which are secured with the present  integration point definitions in GaussLegendreTri.cpp.

NumLib/Fem/FiniteElement/TemplateIsoparametric.h
```diff
@@ -101,13 +103,40 @@ class TemplateIsoparametric
...
+                                ShapeMatrices& shape) const
+    {
+        if (!is_axially_symmetric) {
+            shape.integralMeasure = 1.0;
+            return;
+        }
+
+        auto const r = interpolateZerothCoordinate(shape.N);
+        shape.integralMeasure =
+            2.0 * 3.14159265358979323846264338327950288419 * r;
```
wenqing: use boost constant of pi like Nori used in his PR, or set it somewhere as a global constant?

NumLib/Fem/CoordinatesMapping/ShapeMatrices-impl.h
chleh: Because it's the setZero() function.

MeshLib/Mesh.h
chleh: I excluded it, because no interesting 1D axially symmetric problem came to my mind. Do you want 1D, too?

NumLib/Fem/FiniteElement/TemplateIsoparametric.h
chleh: I don't fully understand your comment. Let's discuss tomorrow.

MeshLib/Mesh.h
wenqing: It may be needed for some benchmarking, and it can be considered in future if it is really needed.

chleh: Then I'd like to postpone the 1D case.

NumLib/Fem/FiniteElement/TemplateIsoparametric.h
wenqing: For the general case of Neumann BC, 2 pi* int q(l) r \phi dl is the integration formulas. However, if the Neumann BC is applied to the axis, where r = 0,  the correct formulas is
int q(z) \phi  dz. To avoid changing Neumann BC integration function, I suggested to added this if-condition.

wenqing: Maybe the good place to check that is in the Neumann BC calculation.

chleh: Resolved in discussion.
I leave the NeumannBC as it is.
Regarding the zero radius for integration points on edges I'll add a comment in `computeIntegralMeasure()`.

NumLib/Fem/FiniteElement/TemplateIsoparametric.h
```diff
@@ -133,6 +133,9 @@ class TemplateIsoparametric
             return;
         }
 
+        // Note: If an integration point is located at the rotation axis, r will
+        // be zero which might lead to problems with the assembled equation
+        // system.
```
wenqing: `` If an integration point is located at the rotation axis, r will be zero``-->

```
For triangle element, if an integration point is located at edge of the unit triangle,  r is possible to be zero
```

Applications/CLI/Tests.cmake
```diff
@@ -393,6 +393,113 @@ if(NOT OGS_USE_MPI)
...
+    # SQUARE 1x1 HEAT CONDUCTION TEST -- AXIALLY SYMMETRIC
+    # test results are compared to 3D simulation on a wedge-shaped domain
+    AddTest(
+         NAME 2D_HeatConduction_axi
+         PATH Parabolic/T/2D_axially_symmetric
+         EXECUTABLE ogs
+         EXECUTABLE_ARGS square_1e2_axi.prj
+         WRAPPER time
+         TESTER vtkdiff
+         ABSTOL 1.7e-5 RELTOL 1e-5
```
endJunction: TBD: Why are the results so different?

ProcessLib/SmallDeformation/CreateLocalAssemblers.h
```diff
@@ -30,7 +30,6 @@ template <unsigned GlobalDim, int DisplacementDim,
 void createLocalAssemblers(
     NumLib::LocalToGlobalIndexMap const& dof_table,
     std::vector<MeshLib::Element*> const& mesh_elements,
-    unsigned const integration_order,
```
endJunction: The integration_order is removed because it's not used in createlocalassemblers, but it is still in the concrete local assemblers, isn't it?

endJunction:
In ogs-data there are some strange files with '.msh' extension ;)
Can we have vtu please!

Applications/CLI/Tests.cmake
```diff
@@ -393,6 +393,113 @@ if(NOT OGS_USE_MPI)
...
+    AddTest(
+        NAME GroundWaterFlowProcess_square_1x1_1e2_axi
+        PATH Elliptic/square_1x1_GroundWaterFlow
+        EXECUTABLE ogs
+        EXECUTABLE_ARGS square_1e2_axi.prj
+        WRAPPER time
+        TESTER vtkdiff
+        #ABSTOL 1.6e-3 RELTOL 1e-3
+        #DIFF_DATA
+        #wedge-1e2-ang-0.2-surface.vtu square_1e2_axi_pcs_0_ts_1_t_1.000000.vtu temperature temperature
```
chleh: Insufficient cleanup...

Applications/CLI/Tests.cmake
chleh: The reference results are from a simulation on a wedge-shaped domain having a wedge angle of 0.02 (rad). I assume that the difference is mostly caused by the size of the angle: Larger angle results in larger deviations.

ProcessLib/SmallDeformation/CreateLocalAssemblers.h
chleh: The integration order is now contained in `extra_ctor_args`. That way one does not need to declare additional arguments in utility code, like `createLocalAssemblers()`, anymore.

chleh:
Actually I'm a bit too lazy to convert them and erase them from the git history only for the reason of style...

endJunction:
I checked with the updated tests' input (msh->vtu) and `ctest -R axi` has 100% passed.

NumLib/Fem/FiniteElement/TemplateIsoparametric.h
chleh: Added your comment for further clarification.

endJunction:
Also don't forget the changelog (depending on the version of the ufz/master)

PR #1444, opened 2016-09-26
Fix a bug in the initialization of the secondary variable

Yonghui56:
as titled.
 in ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h 
Replaced the size of the secondary variable vector as the number of integration points

### DIFF
--- a/ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
+++ b/ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
@@ -74,7 +74,9 @@ public:
           _integration_method(integration_order),
           _shape_matrices(initShapeMatrices<ShapeFunction, ShapeMatricesType,
                                             IntegrationMethod, GlobalDim>(
-              element, is_axially_symmetric, _integration_method))
+              element, is_axially_symmetric, _integration_method)),
+          _darcy_velocities(GlobalDim,
+			  std::vector<double>(_integration_method.getNumberOfPoints()))
     {
     }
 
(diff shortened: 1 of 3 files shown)


chleh:
:+1:.
The same bug also occurs in HeatConduction and TES. SmallDeformation seems to be fine...

ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
```diff
@@ -156,7 +156,7 @@ class LocalAssemblerData
 
     std::vector<std::vector<double>> _darcy_velocities
         = std::vector<std::vector<double>>(
-            GlobalDim, std::vector<double>(ShapeFunction::NPOINTS));
+            GlobalDim, std::vector<double>(_integration_method.getNumberOfPoints()));
```
chleh: Regarding the failing clang compiler (#1448): Maybe this code should be moved to the initializer list of the constructor.

endJunction: @chleh It is finely working with clang-3.5; see #1448 (I tested with this PR merged together and the issues is indeed resolved).

chleh: I am only saying this because all the other members are initialized in the ctor. So the failing clang-3.4 here correlates with bad style. And maybe improving style here (by moving the initialization code) would make clang-3.4 succeed.
Of course, updating clang is also a valid solution.

endJunction:
add to whitelist

chleh:
@Yonghui56, now you pushed far too much to this PR. Can you clean up, please.
Your fix is good, but who will fix the other processes?

Yonghui56:
@chleh  I fix the Heat Conduction process as well. TES process seems to be fine.

endJunction:
@chleh We were not able to find the relevant code in the TES process, please give a more precise pointer where to find secondary variable initialization.

chleh:
@Yonghui56 I sent a PR to your branch regarding the TES process.

PR #1423, opened 2016-09-22
Fix a bug accessing data in PropertyVector.

TomFischer:

PR #1445, opened 2016-09-26
fix a bug in obtaining the value for the given component stored in the given tuple

Yonghui56:
as titiled.
in MeshLib/PropertyVector.h


bilke:
add to whitelist

chleh:
Thanks for your PR, but I think this bug has already been fixed in #1423.

PR #1446, opened 2016-09-27
[Jenkins] Delete build dir at the job end.

bilke:


chleh:
Will it also delete the build dir if some test fails in between? I think that would be nice.
If everything runs :+1:

TomFischer:
Looks good :

bilke:
No, errors are in general not yet handled. I will do so in another PR.

PR #1452, opened 2016-09-28
[SD-LIE] Small deformation process with lower-dimensional interface elements

norihiro-w:
followup of #1455, #1456 and #1457

This PR adds the new process type "SMALL_DEFORMATION_WITH_LIE", which includes lower-dimensional interface elements for modeling fracture deformations. To make review easier, I would like to keep this first PR as simple as possible and focus on deformation with a single fracture cutting a domain. This is indeed enough for the DECOVALEX benchmark. Complexity can be increased later on step by step.

PR #1453, opened 2016-09-28
BoundaryConditionBuilder

norihiro-w:
In #1452, I need [a custom local assembler for Neumann boundary conditions](https://github.com/norihiro-w/ogs/blob/1b30d78beddf459bed84be64f74062c6f159b9d4/ProcessLib/SmallDeformationWithLIE/BoundaryCondition/NeumannBoundaryConditionLocalAssembler.h) .  In order to enable it, I came to a solution introducing `BoundaryConditionBuilder` class. The class can be derived and Primary variable class owns a pointer to the base class. Thus I can set a custom builder for some variables. Let me know if you have a better solution for it.

### DIFF
--- a/ProcessLib/BoundaryCondition/BoundaryCondition.cpp
+++ b/ProcessLib/BoundaryCondition/BoundaryCondition.cpp
@@ -33,13 +33,12 @@ static std::vector<MeshLib::Element*> getClonedElements(
 
 namespace ProcessLib
 {
-std::unique_ptr<BoundaryCondition> createBoundaryCondition(
+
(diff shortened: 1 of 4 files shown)


chleh:
Just wondering: Can't you simply give the LIE BCs different names, e.g. "NeumannLIE" instead of "Neumann"? (that might not be a better solution, though)

norihiro-w:
@chleh I need to pass process data to the local assembler. How can I do it in that case?

chleh:
I don't have a spontaneous idea for that.

chleh:
:+1:

PR #1455, opened 2016-09-29
[SD-LIE] Common tools

norihiro-w:
As #1452 is quite big (>3k lines), I would like to ask code review piece by piece. This first piece is focusing on common tools for the process, located under ProcessLib/SmallDeformationWithLIE/Common. 

The directory includes the followings
- Fracture property class and its configuration
- H-matrix tools similar to B-matrix (H-matrix is based on shape vectors)
- Levelset function
- Mesh tools for getting information about fracture and rock matrix elements
- Other tools for computing a normal vector and a rotation matrix, and for computing physical coordinates from integration point coordinates

TODO
- [ ] replace `#pragma once`

PR #1456, opened 2016-09-29
[SD-LIE] Boundary condition tools

norihiro-w:
part of #1452, follow up of #1453 and #1455 

This PR adds source files under ProcessLib/SmallDeformationWithLIE/BoundaryCondition, which is aimed at providing a custom NeumannBC assembler based on `BoundaryConditionBuilder` in #1453 for displacement jump variables. Boundary integration for Neumann BCs of the variables needs to involve levelset functions.


endJunction:

endJunction:
Code is clean, as I've seen it. I'm just wondering if this large code duplication can be resolved different way. I didn't dig up into the details to make own suggestions

ProcessLib/SmallDeformationWithLIE/BoundaryCondition/GenericNaturalBoundaryCondition.h
```diff
@@ -0,0 +1,83 @@
...
+        typename std::enable_if<
+            std::is_same<typename std::decay<BoundaryConditionData>::type,
+                         typename std::decay<Data>::type>::value,
+            bool>::type is_axially_symmetric,
+        unsigned const integration_order,
+        NumLib::LocalToGlobalIndexMap const& dof_table_bulk,
+        int const variable_id, int const component_id,
+        unsigned const global_dim,
+        std::vector<MeshLib::Element*>&& elements, Data&& data,
+        FractureProperty const& fracture_prop);
```
chleh: Is it possible to put `fracture_prop` into `data`? Then you probably wouldn't need to copy the `GenericNaturalBoundaryCondition`.
Another way of reducing code duplication could be to define the constructor of the "original" `GenericNaturalBoundaryCondition` as follows:
```
template <typename... Data>
GenericNaturalBoundaryCondition(...)
```

ProcessLib/SmallDeformationWithLIE/BoundaryCondition/BoundaryConditionBuilder.cpp
```diff
@@ -0,0 +1,110 @@
...
+    MeshGeoToolsLib::MeshNodeSearcher& mesh_node_searcher =
+        MeshGeoToolsLib::MeshNodeSearcher::getMeshNodeSearcher(mesh);
+
+    MeshGeoToolsLib::BoundaryElementsSearcher boundary_element_searcher(
+        mesh, mesh_node_searcher);
+
+    //! \ogs_file_param{boundary_condition__type}
+    auto const type = config.config.peekConfigParameter<std::string>("type");
+
+    if (type == "Dirichlet")
```
chleh: I think this part could be shared with the normal BC creation. `getClonedElements()` above probably, too.

ProcessLib/SmallDeformationWithLIE/BoundaryCondition/GenericNaturalBoundaryCondition.h
norihiro-w: - no, `data` is actually used to pass `parameters`. 
- `template <typename... Data>` cannot simply solve the issue, because this GenericNaturalBoundaryCondition also needs to pass `variable_id` (which is not in the original one) to local assemblers.

norihiro-w:
I added more virtual functions the original `BoundaryConditionBuilder` class to reduce duplicated codes in the derived.

ProcessLib/SmallDeformationWithLIE/BoundaryCondition/GenericNaturalBoundaryCondition.h
chleh: I see, but actually it looks like the variable_id is not used inside SmallDeformationWithLIE/BoundaryCondition/NeumannBoundaryConditionLocalAssembler.h. So can it be removed from the local assemblers? Or did I overlook something?

norihiro-w: sorry i didn't write comments there, but the variable id is necessary for multiple fractures.

chleh: OK, thanks for the explanation. And thanks for the code reduction you already did.

chleh:
:+1:

norihiro-w:
@endJunction do you have further comments?

endJunction:
:fast_forward:

PR #1457, opened 2016-09-29
[SD-LIE] Local assemblers - part 2

norihiro-w:
part of #1452, follow up of #1475, followed by #1476. ~~Relevant commits are d472a6a and 	3cdb30d~~

This PR includes 
- SmallDeformationLocalAssemblerInterface
- SmallDeformationLocalAssembler for matrix elements, matrix elements having fracture nodes, fracture elements

There are mainly two points to be remarked
- There are three local assemblers for 1) matrix elements (quad or triangle elements) including no fracture nodes,  2) matrix elements including fracture  nodes, and 3) fracture elements (line elements). Case 1) should be identical to the assembler of the standard SD process. Case 2 involves enrichment variables (i.e. displacement jumps) in addition to displacement variables. Case 3 deals only with the enrichment variables.
- In case 2 and 3, the size of local matrix is not equal to the number of element nodes multiplied by N, because additional DoFs are assigned to only fracture nodes. Therefore, I introduced a wrapper function to `assembleWithJacobian()` in `SmallDeformationLocalAssemblerInterface` such that actual local assembly routines do not need to know which nodes have the additional DoF or not.

TODO
- [ ] replace pragma once

### DIFF
--- a/ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerFracture-impl.h
+++ b/ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerFracture-impl.h
@@ -105,6 +105,8 @@ assembleWithJacobian(
     FractureProperty const& frac_prop = *_process_data._fracture_property;
     auto const& R = frac_prop.R;
 
+    // the index of a normal (normal to a fracture plane) component
+    // in a displacement vector
     int const index_normal = DisplacementDim - 1;
 
     unsigned const n_integration_points =
--- a/ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerFracture-impl.h
+++ b/ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerFracture-impl.h
@@ -1,0 +1,174 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef PROCESSLIB_SMALLDEFORMATION_WITH_LIE_SMALLDEFORMATIONLOCALASSEMBLER_FRACTURE_IMPL_H_
+#define PROCESSLIB_SMALLDEFORMATION_WITH_LIE_SMALLDEFORMATIONLOCALASSEMBLER_FRACTURE_IMPL_H_
+
+#include "SmallDeformationLocalAssemblerFracture.h"
+
+#include <Eigen/Eigen>
+
+#include "MathLib/LinAlg/Eigen/EigenMapTools.h"
+
+#include "ProcessLib/Utils/InitShapeMatrices.h"
+
+#include "ProcessLib/SmallDeformationWithLIE/Common/LevelSetFunction.h"
+
+
+namespace ProcessLib
+{
+namespace SmallDeformationWithLIE
+{
+template <typename ShapeFunction, typename IntegrationMethod,
+          int DisplacementDim>
+SmallDeformationLocalAssemblerFracture<ShapeFunction, IntegrationMethod,
+                               DisplacementDim>::
+    SmallDeformationLocalAssemblerFracture(
+        MeshLib::Element const& e,
+        std::size_t const /*local_matrix_size*/,
+        std::vector<unsigned> const& dofIndex_to_localIndex,
+        bool is_axially_symmetric,
+        unsigned const integration_order,
+        SmallDeformationProcessData<DisplacementDim>& process_data)
+    : SmallDeformationLocalAssemblerInterface(
+          ShapeFunction::NPOINTS * DisplacementDim, // no intersection
+          dofIndex_to_localIndex),
+      _process_data(process_data),
+      _integration_method(integration_order),
+      _shape_matrices(
+          initShapeMatrices<ShapeFunction, ShapeMatricesType,
+                            IntegrationMethod, DisplacementDim>(
+              e, is_axially_symmetric, _integration_method)),
+      _element(e)
+{
+    assert(_element.getDimension() == DisplacementDim-1);
+
+    unsigned const n_integration_points =
+        _integration_method.getNumberOfPoints();
+
+    _ip_data.reserve(n_integration_points);
+    _secondary_data.N.resize(n_integration_points);
+
+    auto const* frac_prop = _process_data._fracture_property.get();
+
+    SpatialPosition x_position;
+    x_position.setElementID(_element.getID());
+    for (unsigned ip = 0; ip < n_integration_points; ip++)
+    {
+        x_position.setIntegrationPoint(ip);
+
+        _ip_data.emplace_back(*_process_data._fracture_model);
+        auto const& sm = _shape_matrices[ip];
+        auto& ip_data = _ip_data[ip];
+        ip_data._detJ = sm.detJ;
+        ip_data._integralMeasure = sm.integralMeasure;
+        ip_data._h_matrices.resize(
+            DisplacementDim,
+            ShapeFunction::NPOINTS * DisplacementDim);
+
+        computeHMatrix<
+            DisplacementDim, ShapeFunction::NPOINTS,
+            typename ShapeMatricesType::NodalRowVectorType, HMatrixType>(
+            sm.N, ip_data._h_matrices);
+
+        ip_data._w.resize(DisplacementDim);
+        ip_data._w_prev.resize(DisplacementDim);
+        ip_data._sigma.resize(DisplacementDim);
+        ip_data._sigma_prev.resize(DisplacementDim);
+        ip_data._C.resize(DisplacementDim, DisplacementDim);
+        ip_data._aperture0 = (*frac_prop->aperture0)(0, x_position)[0];
+        ip_data._aperture_prev = ip_data._aperture0;
+
+        _secondary_data.N[ip] = sm.N;
+    }
+}
+
+
+template <typename ShapeFunction, typename IntegrationMethod,
+          int DisplacementDim>
+void SmallDeformationLocalAssemblerFracture<ShapeFunction, IntegrationMethod,
+                                    DisplacementDim>::
+assembleWithJacobian(
+    double const t,
+    Eigen::VectorXd const& local_u,
+    Eigen::VectorXd& local_b,
+    Eigen::MatrixXd& local_J)
+{
+    auto const& nodal_jump = local_u;
+
+    FractureProperty const& frac_prop = *_process_data._fracture_property;
+    auto const& R = frac_prop.R;
+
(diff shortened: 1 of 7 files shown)


endJunction:

norihiro-w:
the PR is ready for review

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerFracture-impl.h
```diff
@@ -0,0 +1,174 @@
...
+{
+    double ele_b = 0;
+    unsigned const n_integration_points =
+        _integration_method.getNumberOfPoints();
+    for (unsigned ip = 0; ip < n_integration_points; ip++)
+    {
+        ele_b += _ip_data[ip]._aperture;
+    }
+    ele_b /= n_integration_points;
+    (*_process_data._mesh_prop_b)[_element.getID()] = ele_b;
```
chleh: Do you use this for auxiliary output?
For now it's OK. But our output functionality clearly has to be improved in this regard (Actually, I wanted to do this already several months ago...)

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerInterface.h
```diff
@@ -0,0 +1,115 @@
...
+    virtual void assembleWithJacobian(
+        double const t,
+        Eigen::VectorXd const& local_u,
+        Eigen::VectorXd& local_b,
+        Eigen::MatrixXd& local_J) {
+        (void)t;
+        (void)local_u;
+        (void)local_b;
+        (void)local_J;
+        OGS_FATAL("SmallDeformationLocalAssemblerInterface::assembleWithJacobian() is not implemented");
```
chleh: Is this default implementation needed? Or has it any benefit? If not I'd suggest to remove it since then the error message is already printed at compile time.
Furthermore I'd prefer to name this second `assembleWithJacobian` method slightly differently, e.g. `assembleWithJacobianConcrete`.

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerMatrix-impl.h
```diff
@@ -0,0 +1,196 @@
...
+
+
+template <typename ShapeFunction, typename IntegrationMethod,
+          int DisplacementDim>
+void SmallDeformationLocalAssemblerMatrix<ShapeFunction, IntegrationMethod,
+                                    DisplacementDim>::
+postTimestepConcrete(std::vector<double> const& /*local_x*/)
+{
+    const int n = 3;
+    std::valarray<double> ele_stress(0.0, n);
```
chleh: Minor: I'd prefer Eigen types here instead of `valarray`, because the former are used everywhere anyways.

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerMatrix-impl.h
```diff
@@ -0,0 +1,196 @@
...
+        auto& eps = _ip_data[ip]._eps;
+        auto& sigma = _ip_data[ip]._sigma;
+        auto& C = _ip_data[ip]._C;
+        auto& material_state_variables =
+            *_ip_data[ip]._material_state_variables;
+
+        eps.noalias() =
+            B *
+            Eigen::Map<typename BMatricesType::NodalForceVectorType const>(
+                local_x.data(), ShapeFunction::NPOINTS * DisplacementDim);
```
chleh: This could be done with some function from `EigenMapTools.h`.

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerMatrixNearFracture-impl.h
```diff
@@ -0,0 +1,228 @@
...
+    Eigen::MatrixXd& local_J)
+{
+    assert (_element.getDimension() == DisplacementDim);
+
+    auto const n_dof_per_var = ShapeFunction::NPOINTS * DisplacementDim;
+
+    auto localRhs_ru = local_b.segment(0, n_dof_per_var);
+    auto localRhs_du = local_b.segment(n_dof_per_var, n_dof_per_var);
+
+    auto localA_uu = local_J.block(0, 0, n_dof_per_var, n_dof_per_var);
```
chleh: Eigen provides template versions of block: `block<N, M>(...)`. Since the size is known at compile time, you could use them. Same for `segment` above
Furthermore there are variants like upperLeftPart (or similar) which might in some cases make the code a bit easier to read.

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerFracture-impl.h
norihiro-w: yes, only for output

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerMatrix-impl.h
norihiro-w: I would keep current. it doesn't harm

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerMatrixNearFracture-impl.h
norihiro-w: I keep it as it is now. this part will be changed later when multiple fractures are considered.

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerInterface.h
norihiro-w: the reason for this is, the second version is not used for the usual matrix elements. Thus I cannot make it pure virtual.

norihiro-w:
sorry I misunderstood the label.

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerMatrix-impl.h
wenqing: There is a function called MathLib::toMatrix.

Since the reverts #1479  looks like removed stuff, I leave my comment here.

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerMatrix-impl.h
```diff
@@ -0,0 +1,196 @@
...
+    std::valarray<double> ele_strain(0.0, n);
+    unsigned const n_integration_points =
+        _integration_method.getNumberOfPoints();
+    for (unsigned ip = 0; ip < n_integration_points; ip++)
+    {
+        auto& ip_data = _ip_data[ip];
+
+        ele_stress[0] += ip_data._sigma[0];
+        ele_stress[1] += ip_data._sigma[1];
+        ele_stress[2] += ip_data._sigma[3];
```
wenqing: Is this correct?
 ele_stress[2] += ip_data._sigma[3];

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerMatrixNearFracture-impl.h
```diff
@@ -0,0 +1,228 @@
...
+        auto& C = ip_data._C;
+        auto& material_state_variables = *ip_data._material_state_variables;
+
+        eps.noalias() = B * nodal_u;
+
+        if (!ip_data._solid_material.computeConstitutiveRelation(
+                t, x_position, _process_data.dt, eps_prev, eps, sigma_prev,
+                sigma, C, material_state_variables))
+            OGS_FATAL("Computation of local constitutive relation failed.");
+
```
wenqing: It might be good to introduce a constant for detJ * wp.getWeight() * integralMeasure.

Besides,  transpose() seems perform a real transpose. It is true, to place the transpose B matrix with  transposeInPlace() might be good.

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerMatrixNearFracture-impl.h
```diff
@@ -0,0 +1,228 @@
...
+    std::valarray<double> ele_strain(0.0, n);
+    unsigned const n_integration_points =
+        _integration_method.getNumberOfPoints();
+    for (unsigned ip = 0; ip < n_integration_points; ip++)
+    {
+        auto& ip_data = _ip_data[ip];
+
+        ele_stress[0] += ip_data._sigma[0];
+        ele_stress[1] += ip_data._sigma[1];
+        ele_stress[2] += ip_data._sigma[3];
```
wenqing: Same question here and hereafter:  ele_stress[2] += ip_data._sigma[3];

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerMatrix-impl.h
norihiro-w: yes, I want to output XY component.

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerMatrixNearFracture-impl.h
norihiro-w: I take it as an optimization, which should be done later if necessary.

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerMatrixNearFracture-impl.h
norihiro-w: same as above

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerMatrixNearFracture-impl.h
chleh: @wenqing: probably ` transposeInPlace()` alters the `B` matrix, what is not what you want here, because you need the original `B` afterwards.

wenqing: @chleh You are right, in place transpose changes the memory of the matrix itself. What I mean is to assign the transpose of B to another matrix if the transpose is really performed by member transpose().

PR #1458, opened 2016-09-29
[CMake] Moved packaging executable dependencies to package target.

bilke:
Beforehand this logic was executed in a forced additional CMake run
everytime after the specific executable was built. Now this is done
at the moment this information is needed (at building the package or
install target).


norihiro-w:
:+1:

PR #1459, opened 2016-09-30
[Mesh] searche boundary nodes in NodeSearch

norihiro-w:
As titled.


chleh:
Isn't such a feature already implemented somewhere? AFAIK we need to search boundary nodes on a geometry in order to set up BCs.
What's the difference to this PR?

norihiro-w:
there are some implementations similar to this, but are not what I need. What eventually I need is to check if a node is on domain boundary, not on a particular geometry.

endJunction:

TomFischer:
:+1:

PR #1461, opened 2016-10-01
Calculate surface flux; small fixes

endJunction:
Fixing the boost::optional<T const&> usage, which causes compilation error when using boost-1.61.
Second one is small safety net feature.


chleh:
:fast_forward: 

Btw: I think `boost::optional<MeshLib::PropertyVector<std::size_t> const&>` is not sensible. `MeshLib::PropertyVector<std::size_t> const*` expresses the same. But that's a different issue.

endJunction:
@chleh I agree, `boost::optional<T const&>` does not make much sense. Maybe that's one of the reasons why it is no longer available in boost-1.61.
Actually it is similar to `std::vector<T const>` vs. `std::vector<T> const`.....

PR #1462, opened 2016-10-01
Optimization loop over mesh subsets nodes

endJunction:
Follow up of https://github.com/ufz/ogs/pull/1461 (otherwise it does not compile for me) but the changes are independent.

The loops over all nodes of mesh subsets are rewritten in the findGlobalIndices and createBoundaryCondition. The new version uses a sorted vector of nodes (or ids) for faster lookup on large meshes.

### DIFF
--- a/ProcessLib/BoundaryCondition/BoundaryCondition.cpp
+++ b/ProcessLib/BoundaryCondition/BoundaryCondition.cpp
@@ -57,23 +57,29 @@ std::unique_ptr<BoundaryCondition> BoundaryConditionBuilder::createBoundaryCondi
             mesh_node_searcher.getMeshNodeIDs(config.geometry);
         // Filter out ids, which are not part of mesh subsets corresponding to
         // the variable_id and component_id.
+
+        // Sorted ids of all mesh_subsets.
+        std::vector<std::size_t> sorted_nodes_ids;
+
         auto const& mesh_subsets =
             dof_table.getMeshSubsets(variable_id, config.component_id);
-        auto ids_new_end_iterator = std::end(ids);
         for (auto const& mesh_subset : mesh_subsets)
         {
             auto const& nodes = mesh_subset->getNodes();
-            ids_new_end_iterator = std::remove_if(std::begin(ids), ids_new_end_iterator,
-                    [&nodes](std::size_t const node_id)
-                    {
-                        return std::end(nodes) == std::find_if(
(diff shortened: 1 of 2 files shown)


ProcessLib/BoundaryCondition/BoundaryCondition.cpp
```diff
@@ -57,23 +57,29 @@ std::unique_ptr<BoundaryCondition> BoundaryConditionBuilder::createBoundaryCondi
...
+        std::sort(std::begin(sorted_nodes_ids), std::end(sorted_nodes_ids));
+
+        auto ids_new_end_iterator = std::end(ids);
+        ids_new_end_iterator = std::remove_if(
+            std::begin(ids), ids_new_end_iterator,
+            [&sorted_nodes_ids](std::size_t const node_id) {
+                return !std::binary_search(std::begin(sorted_nodes_ids),
+                                           std::end(sorted_nodes_ids), node_id);
+            });
         ids.erase(ids_new_end_iterator, std::end(ids));
```
chleh: Since you are optimizing: Wouldn't `ids.resize(...)` be more efficient? I.e., it has constant complexity.

ProcessLib/BoundaryCondition/BoundaryCondition.cpp
```diff
@@ -57,23 +57,29 @@ std::unique_ptr<BoundaryCondition> BoundaryConditionBuilder::createBoundaryCondi
...
+                           [](MeshLib::Node* const n) { return n->getID(); });
         }
+        std::sort(std::begin(sorted_nodes_ids), std::end(sorted_nodes_ids));
+
+        auto ids_new_end_iterator = std::end(ids);
+        ids_new_end_iterator = std::remove_if(
+            std::begin(ids), ids_new_end_iterator,
+            [&sorted_nodes_ids](std::size_t const node_id) {
+                return !std::binary_search(std::begin(sorted_nodes_ids),
+                                           std::end(sorted_nodes_ids), node_id);
```
chleh: I think if you also sorted `ids` before the removal, you could do a better than a simple binary search. (I.e., do a binary search where the lower boundary increases monotonically.)

ProcessLib/BoundaryCondition/BoundaryCondition.cpp
endJunction: According to cppreference both, resize and erase, are linear in their complexity.

PR #1464, opened 2016-10-05
Integration order in project files.

endJunction:
Until now the integration order was hardcoded into default value of 2 in Process class.
I introduced a new tag `<integration_order>` for each `<process>` which is parsed and propagated to the `Process` ctor.

### DIFF
--- a/Applications/ApplicationsLib/ProjectData.cpp
+++ b/Applications/ApplicationsLib/ProjectData.cpp
@@ -265,6 +265,10 @@ void ProjectData::parseProcesses(BaseLib::ConfigTree const& processes_config,
         //! \ogs_file_param{process__name}
         auto const name = process_config.getConfigParameter<std::string>("name");
 
+        auto const integration_order =
+            //! \ogs_file_param{process__integration_order}
+            process_config.getConfigParameter<int>("integration_order");
+
         std::unique_ptr<ProcessLib::Process> process;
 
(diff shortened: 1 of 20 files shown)


Applications/ApplicationsLib/ProjectData.cpp
```diff
@@ -266,6 +266,10 @@ void ProjectData::parseProcesses(BaseLib::ConfigTree const& processes_config,
         //! \ogs_file_param{process__type}
         auto const name = process_config.getConfigParameter<std::string>("name");
 
+        //! \ogs_file_param{process__integration_order}
```
chleh: This comment has to be in the line before `getConfigParameter`. I.e., please move it down one line.

endJunction:
Updated the surface flux too; the integration order is same as in the caller process (GWF in this case).

chleh:
:+1:

wenqing:
:+1:

endJunction:

endJunction:
Jenkins, test this please!

wenqing:
@endJunction  In the data repository, branch IntegrationOrder is not merged to master.
Another issue: The order can only be 2. If it is 1, the least square extrapolation does not work.

endJunction:
@wenqing Yes, thanks forgot to update the Tests/Data.

PR #1466, opened 2016-10-06
Added VTU support to moveMeshNodes utility

rinkk:


endJunction:
Please rebase to the 6.0.7 release tag before merging.

Applications/Utils/MeshEdit/moveMeshNodes.cpp
```diff
@@ -201,8 +209,7 @@ int main (int argc, char* argv[])
     }
     /**** add other keywords here ****/
 
-    MeshLib::IO::writeMeshToFile(
-        *mesh, msh_name.substr(0, msh_name.length() - 4) + "_new.msh");
+    MeshLib::IO::VtuInterface vtu (mesh.get(), 0, false);
+    vtu.writeToFile(msh_name.substr(0, msh_name.length() - 4) + "_new.vtu");
```
endJunction: Just confirming: the 'msh' output is no longer needed? Otherwise there is writeMeshToFile() which is writing either 'vtu' or 'msh' depending on the extension, and is more general.

Applications/Utils/MeshEdit/moveMeshNodes.cpp
```diff
@@ -151,10 +155,14 @@ int main (int argc, char* argv[])
         }
         const std::string value (argv[3]);
         double max_dist(pow(strtod(argv[4],0), 2));
-        //const std::string value("D:\\Rappbodevorsperre_elevation440m.msh");
-        //double max_dist (25.0);    // squared maximum distance at which reference points are used
         double offset (0.0); // additional offset for elevation (should be 0)
-        MeshLib::Mesh* ground_truth (MeshLib::IO::readMeshFromFile(value));
+        std::unique_ptr<MeshLib::Mesh> ground_truth (MeshLib::IO::readMeshFromFile(value));
+        if (mesh == nullptr)
```
endJunction: Why checking mesh pointer here again? Should be ground_truth probably.

TomFischer:
At some places the tool returns in error case 1 and at some places -1. Is there a reason for this? If not, I would suggest to return `EXIT_FAILURE` on the error case and if there is not any error `EXIT_SUCCESS`.

TomFischer:
:+1:

endJunction:
Looks good!

rinkk:
If I see this correctly, travis fails because the "MaterialLibSolidsKelvinVector6" test fails?
What's the procedure here? I don't think this has anything to do with this PR?!

PR #1468, opened 2016-10-06
Liquid flow process

wenqing:
Added classes for modelling liquid flow process based on the previous implementations about fluid and porous medium properties. The implementation is verified with the following
~~three~~ four tests from OGS5:
- LineDirichletNeumannBC  ~~sat1D~~
- PressureBCatCornerOfAnisotropicSquare ~~sat2D~~
- GravityDriven
- AxiSymTheis (with axisymmetry)

In the current ogs6/master, the calculations of velocity and flux were implemented in wrong places for groundwater, heat conduction. A PR about the correction of such calculations for all non-deformation processes comes immediately once this one is merged.


ProcessLib/LiquidFlow/LiquidFlowLocalAssembler-imp.h
```diff
@@ -0,0 +1,136 @@
...
+            const double K = perm(0, 0) / mu;
+            const double fac = K * integration_factor;
+            local_K.noalias() += fac * sm.dNdx.transpose() * sm.dNdx;
+            if (_compute_gravitational_term)
+            {
+                local_b.noalias() -=
+                    fac * sm.dNdx.transpose().col(GlobalDim - 1) * rho_g;
+            }
+
+            // Compute the velocity
```
wenqing: The velocity calculation will be moved to a new member function in the  next PR in order to perform it after the linearization.

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler-imp.h
```diff
@@ -0,0 +1,136 @@
...
+
+    auto local_M = MathLib::createZeroedMatrix<NodalMatrixType>(
+        local_M_data, local_matrix_size, local_matrix_size);
+    auto local_K = MathLib::createZeroedMatrix<NodalMatrixType>(
+        local_K_data, local_matrix_size, local_matrix_size);
+    auto local_b = MathLib::createZeroedVector<NodalVectorType>(
+        local_b_data, local_matrix_size);
+
+    // For velocity calculation:
+    // TODO: move velocity calculation to a member function, and call the
```
wenqing: Done in my branch for the next PR.

Applications/CLI/Tests.cmake
```diff
@@ -520,6 +520,52 @@ if(NOT OGS_USE_MPI)
     #     EXECUTABLE_ARGS tes-inert-wedge.prj
     # )
 
+    # Liquid flow
+    AddTest(
+        NAME LiquidFlowProcess_sat1D
```
endJunction: Is it possible to name the tests, s.t. its general setup is clear from the name.
Otherwise I've to look into the prj files every time to find out what that could be.
Something like LiquidFlow_1D_line100_saturationZone_quadraticPressureLeft?

ProcessLib/LiquidFlow/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+
```
endJunction: Why do we need these empty files?
`APPEND_SOURCE_FILES(SOURCES LiquidFlow)` in ProcessLib/CMakeLists.txt is sufficient. Are there other reasons?

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler-imp.h
```diff
@@ -0,0 +1,136 @@
...
+
+        // Compute density:
+        const double rho_g =
+            _material_properties.getLiquidDensity(p, _temperature) *
+            MaterialLib::PhysicalConstant::g;
+        // Compute viscosity:
+        const double mu = _material_properties.getViscosity(p, _temperature);
+
+        // Assemble Laplacian, K, and RHS by the gravitational term
+        if (perm.size() == 1)  // Save the computing time for isotropic permeability.
```
endJunction: I'd prefer no code duplication here unless this saves more than 1% runtime.
Did you measure how much computation time is saved, I'm just curious about the trade off btw. a branch and additional multiplications/additions.

ProcessLib/LiquidFlow/LiquidFlowMaterialProperties.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+        //! \ogs_file_param{prj__material_property__porous_medium__porous_medium__storage}
+        auto const& stora_conf = conf.getConfigSubtree("storage");
+        auto beta = MaterialLib::PorousMedium::createStorageModel(stora_conf);
+        storage.emplace_back(std::move(beta));
+    }
+}
+
+double LiquidFlowMaterialProperties::getMassCoefficient(
+    const double var4porosity, const double var4storage,
```
endJunction: The 4 means fourth position or some other index?

ProcessLib/LiquidFlow/LiquidFlowMaterialProperties.h
```diff
@@ -0,0 +1,96 @@
...
+
+namespace ProcessLib
+{
+namespace LiquidFlow
+{
+struct LiquidFlowMaterialProperties
+{
+    typedef MaterialLib::Fluid::FluidProperty::ArrayType ArrayType;
+
+    explicit LiquidFlowMaterialProperties(BaseLib::ConfigTree const& config);
```
endJunction: Usually a create function is used for better testability of the class and to separate parsing from actual construction.

ProcessLib/LiquidFlow/LiquidFlowMaterialProperties.h
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * \brief Compute the coefficient of the mass term by
+     *      \f[
+     *           n \frac{partial \rho_l}{\partial p} + \beta_s
+     *      \f]
+     *     where \f$n\f$ is the porosity, \f$rho_l\f$ is the liquid density,
+     *     \f$bata_s\f$ is the storage.
+     * \param var4porosity       The first variable for porosity model, and it could be
+     *                           saturation, and invariant of stress or strain.
```
endJunction: I'm not very knowledgeable about material models, but from technical point of view it is difficult to find out if a value came from saturation, some invariant of stress, or some invariant of strain. Better to use some enum or comparable techniques to avoid ambiguity.

ProcessLib/LiquidFlow/LiquidFlowMaterialProperties.h
```diff
@@ -0,0 +1,96 @@
...
+     * \param var4storage        Variable for storage model.
+     * \param p                  Pressure value
+     * \param T                  Temperature value
+     * \param material_group_id  Material ID of the element
+     */
+    double getMassCoefficient(const double p, const double T,
+                             const double var4porosity, const double var4storage,
```
endJunction: Instead of interesting shortcuts, simply call the variables `porosity_variable` and  `storage_variable`.

ProcessLib/LiquidFlow/LiquidFlowProcess.h
```diff
@@ -0,0 +1,100 @@
...
+    void preIterationConcreteProcess(const unsigned iter,
+                                     GlobalVector const& x) override;
+
+    bool isLinear() const override { return true; }
+private:
+    void initializeConcreteProcess(
+        NumLib::LocalToGlobalIndexMap const& dof_table,
+        MeshLib::Mesh const& mesh, unsigned const integration_order) override;
+
+    void assembleConcreteProcess(const double t, GlobalVector const& p,
```
endJunction: Sometimes the argument of GlobalVector type is called p, sometimes x, and there is some assignment of p = x in postTimeStep. Better to choose one name and use it consistently.
I'd prefer the original x, because it can have more then pressure variable and is consistent with the caller.

ProcessLib/LiquidFlow/createLiquidFlowProcess.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+    NumLib::NamedFunctionCaller named_function_caller({"LiquidFlow_pressure"});
+
+    ProcessLib::parseSecondaryVariables(config, secondary_variables,
+                                        named_function_caller);
+
+    auto const gravitational_term =
+    //! \ogs_file_param{process__LIQUID_FLOW__gravitational_term}
+    config.getConfigParameter<std::string>("gravitational_term");
+    const bool has_gravitational_term = (gravitational_term == "enabled") ? true : false;
```
endJunction: Simply:
```c++
const bool has_gravitational_term = (gravitational_term == "enabled")
```

ProcessLib/LiquidFlow/createLiquidFlowProcess.h
```diff
@@ -0,0 +1,32 @@
...
+
+namespace ProcessLib
+{
+namespace LiquidFlow
+{
+std::unique_ptr<Process> createLiquidFlowProcess(
+    MeshLib::Mesh& mesh,
+    std::unique_ptr<ProcessLib::AbstractJacobianAssembler>&& jacobian_assembler,
+    std::vector<ProcessVariable> const& variables,
+    std::vector<std::unique_ptr<ParameterBase>> const& /*parameters*/,
```
endJunction: Why is only parameters commented?

Tests/ProcessLib/LiquidFlow/TestLiquidFlowMaterialProperties.cpp
```diff
@@ -0,0 +1,124 @@
...
+using namespace MaterialLib::Fluid;
+using ArrayType = MaterialLib::Fluid::FluidProperty::ArrayType;
+using Matrix = Eigen::MatrixXd;
+
+TEST(ProcessLibLiquidFlow, checkLiquidFlowMaterialProperties)
+{
+    const char xml[] =
+        "<material_property>"
+        "    <fluid>"
+        "        <fluid>"
```
endJunction: Why is fluid repeated here?

wenqing:
@endJunction Thanks for the comments, and I will revise the source code late on. BTW, just captured the error in MaterialLibSolidsKelvinVector4.Inverse, and you can find it by clicking above building "Details".
```
[       OK ] MaterialLibSolidsKelvinVector4.SelfTestMappingKelvinToTensor (19 ms)

[ RUN      ] MaterialLibSolidsKelvinVector4.DeviatoricSphericalProjections

[       OK ] MaterialLibSolidsKelvinVector4.DeviatoricSphericalProjections (0 ms)

[ RUN      ] MaterialLibSolidsKelvinVector4.Inverse

/home/travis/build/ufz/ogs/ThirdParty/autocheck/include/autocheck/reporter.hpp:91: Failure

Value of: AUTOCHECK_SUCCESS

  Actual: false

Expected: true

Falsifiable, after 194 tests:

(  20.715

 88.8961

-17.4266

-60.6874)

[  FAILED  ] MaterialLibSolidsKelvinVector4.Inverse (5 ms)

[ RUN      ] MaterialLibSolidsKelvinVector4.Determinant
```

endJunction:
Thanks for the inverse test failing pointer; I didn't see any for some time now. I'll relax the tolerance a little more.

Applications/CLI/Tests.cmake
wenqing: Changed the name of two tests.

ProcessLib/LiquidFlow/CMakeLists.txt
wenqing: Removed.

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler-imp.h
wenqing: I think for 3D elements, such as hexehadra, it could save the computing time for a homogeneous medium. Image to compare the commutating times of 
- Matrix with a size of (20, 3) multiplied with  a matrix with a size of (3, 3)
- Matrix with a size of (20, 3) multiplied with  a scalar number.

To make this function concise, the computation of velocity, which will be in the next PR, is removed.

ProcessLib/LiquidFlow/LiquidFlowMaterialProperties.cpp
wenqing: Replaced it with a more meaningful name.

ProcessLib/LiquidFlow/LiquidFlowMaterialProperties.h
wenqing: I've tried. To do that, not only this struct needs a copy constructor, but also classes of fluid properties and porous medium properties need their copy constructors in order to  move the unique_ptr type instances of these classes between the creator and receiver. Besides, the copy of eigen matrix of permeability is a concrete copy. I also noticed that TESProcess  has parsring inside its ctor too. Therefore, I would like to keep the parsing in the ctor.

ProcessLib/LiquidFlow/LiquidFlowMaterialProperties.h
wenqing: The name is replaced with a meaningful one. This augment just passes a double type value, and hereby  it is unnecessary to distinguish what is the value for. With the value by this argument, the function just calls the base of polymorphic classes  to compute a porosity.

The comment is improved according the above explanation.

ProcessLib/LiquidFlow/LiquidFlowMaterialProperties.h
wenqing: Changed.

ProcessLib/LiquidFlow/LiquidFlowProcess.h
wenqing: Unified.

ProcessLib/LiquidFlow/createLiquidFlowProcess.h
wenqing: Removed the comment.

Tests/ProcessLib/LiquidFlow/TestLiquidFlowMaterialProperties.cpp
wenqing: Removed.

wenqing:
@endJunction ctests excluding those for this PR failed because that branch IntegrationOrder of the data repository is not merged to the master.

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler-imp.h
```diff
@@ -0,0 +1,114 @@
...
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ * \file   LiquidFlowLocalAssembler.cpp
+ *
+ * Created on August 19, 2016, 2:28 PM
+ */
+
+#ifndef OGS_LIQUIDFLOWLOCALASSEMBLER_IMP_H
+#define OGS_LIQUIDFLOWLOCALASSEMBLER_IMP_H
```
endJunction: Until now everybody named the implementation files ending with '-impl.h'; don't break that please.

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler-imp.h
```diff
@@ -0,0 +1,114 @@
...
+
+    unsigned const n_integration_points =
+        _integration_method.getNumberOfPoints();
+
+    const unsigned mat_id = 0;  // TODO for heterogeneous medium
+    const Eigen::MatrixXd& perm =
+        _material_properties.intrinsic_permeabiliy[mat_id];
+
+    // Note: For Inclined 1D in 2D/3D or 2D element in 3D, the first item in
+    //  the assert must be changed to perm.rows() == _element->getDimension()
```
endJunction: Can you do this please.

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler-imp.h
endJunction: I cannot imagine any runtime numbers. But the code duplication is error-prone and more difficult to maintain. If there is no obvious difference in the runtime I recommend to use code w/o duplication.

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler.h
```diff
@@ -0,0 +1,138 @@
...
+
+    using NodalMatrixType = typename LocalAssemblerTraits::LocalMatrix;
+    using NodalVectorType = typename LocalAssemblerTraits::LocalVector;
+    using GlobalDimVectorType = typename ShapeMatricesType::GlobalDimVectorType;
+
+public:
+    LiquidFlowLocalAssembler(
+        MeshLib::Element const& element,
+        std::size_t const /*local_matrix_size*/,
+        bool is_axially_symmetric,
```
endJunction: Can be const, as the other arguments are.

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler.h
```diff
@@ -0,0 +1,138 @@
...
+    std::vector<ShapeMatrices> _shape_matrices;
+
+    std::vector<std::vector<double>> _darcy_velocities =
+        std::vector<std::vector<double>>(
+            GlobalDim,
+            std::vector<double>(_integration_method.getNumberOfPoints()));
+
+    const bool _compute_gravitational_term;
+    LiquidFlowMaterialProperties const& _material_properties;
+    double _temperature;  ///< Temperature.
```
endJunction: no need to comment the obvious.

Tests/ProcessLib/LiquidFlow/TestLiquidFlowMaterialProperties.cpp
```diff
@@ -0,0 +1,122 @@
...
+        "           <beta> 2.0e-4 </beta> "
+        "           <rho0>999.8</rho0>"
+        "        </density>"
+        "        <viscosity>"
+        "            <type>Vogels</type>"
+        "            <liquid_type>Water </liquid_type>"
+        "        </viscosity>"
+        "    </fluid>"
+        "    <porous_medium>"
+        "        <porous_medium>"
```
endJunction: why the duplication?

ProcessLib/LiquidFlow/LiquidFlowMaterialProperties.h
```diff
@@ -0,0 +1,98 @@
...
+    double getMassCoefficient(const double p, const double T,
+                              const double porosity_variable,
+                              const double storage_variable,
+                              const unsigned material_group_id = 0) const;
+
+    double getLiquidDensity(const double p, const double T) const
+    {
+        ArrayType vars;
+        vars[0] = T;
+        vars[1] = p;
```
endJunction: Just an understanding question. How do you know that T has index 0 and p has index 1 and not vice versa, like the arguments are?
There must be a better way to designate such implicit knowledge (maybe using enums).

ProcessLib/LiquidFlow/LiquidFlowMaterialProperties.h
```diff
@@ -0,0 +1,98 @@
...
+
+    double getViscosity(const double p, const double T) const
+    {
+        ArrayType vars;
+        vars[0] = T;
+        vars[1] = p;
+        return viscosity->getValue(vars);
+    }
+
+    std::unique_ptr<MaterialLib::Fluid::FluidProperty> density_l;
```
endJunction: what is the '_l' mean?

ProcessLib/LiquidFlow/LiquidFlowMaterialProperties.h
```diff
@@ -0,0 +1,98 @@
...
+    std::unique_ptr<MaterialLib::Fluid::FluidProperty> viscosity;
+
+    /// Porous medium properties of different material zones.
+    /// The vector is left empty if the property data are given in vtu file,
+    /// e.g for heterogeneous medium.
+    std::vector<Eigen::MatrixXd> intrinsic_permeabiliy;
```
endJunction: typo in permeability.

ProcessLib/LiquidFlow/LiquidFlowProcess.cpp
```diff
@@ -0,0 +1,124 @@
...
+}
+
+void LiquidFlowProcess::preTimestepConcreteProcess(GlobalVector const& x,
+                                                   const double /*t*/,
+                                                   const double /*delta_t*/)
+{
+    DBUG("New time step");
+
+    _p_previous_timestep =
+        MathLib::MatrixVectorTraits<GlobalVector>::newInstance(x);
```
endJunction: p = x?

ProcessLib/LiquidFlow/LiquidFlowProcess.h
```diff
@@ -0,0 +1,102 @@
...
+        const double dxdot_dx, const double dx_dx, GlobalMatrix& M,
+        GlobalMatrix& K, GlobalVector& b, GlobalMatrix& Jac) override;
+
+    const bool _compute_gravitational_term;
+    LiquidFlowMaterialProperties _material_properties;
+
+    std::vector<std::unique_ptr<LiquidFlowLocalAssemblerInterface>>
+        _local_assemblers;
+
+    std::unique_ptr<GlobalVector> _p_previous_timestep;
```
endJunction: Where is this used?

Tests/ProcessLib/LiquidFlow/TestLiquidFlowMaterialProperties.cpp
```diff
@@ -0,0 +1,122 @@
...
+        "            <permeability>"
+        "                <values>2.e-10 0. 0. 0. 3.e-10 0. 0. 0. 4.0e-10</values>"
+        "            </permeability>"
+        "            <porosity>"
+        "                <type>Constant</type>"
+        "                <value> 0.2 </value> "
+        "            </porosity>"
+        "            <storage>"
+        "                <type>Constant</type>"
+        "                <value> 1.e-4 </value>"
```
endJunction: Maybe I'm on a wrong train, but shouldn't be the properties used for this instead of reimplementation of Constant, MeshElement/NodeProperties, etc?

ProcessLib/CMakeLists.txt
```diff
@@ -14,6 +14,8 @@ APPEND_SOURCE_FILES(SOURCES Parameter)
 add_subdirectory(GroundwaterFlow)
 APPEND_SOURCE_FILES(SOURCES GroundwaterFlow)
 
+append_source_files(SOURCES LiquidFlow)
```
TomFischer: Why not `APPEND_SOURCE_FILES` to  stay consistent within the file?

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler-imp.h
wenqing: Do you mean Inclined 1D in 2D/3D or 2D element in 3D? I can do it, but it should be in done the general way in the integration of MathLib.

wenqing: I can get time for that after finish TH and a project.

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler-imp.h
wenqing: Changed.

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler.h
wenqing: Changed.

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler.h
wenqing: removed.

Tests/ProcessLib/LiquidFlow/TestLiquidFlowMaterialProperties.cpp
wenqing: It is for patch-wise heterogeneous porous medium properties. For example,  a layered 3D domain or a domain with different material sub domains.

ProcessLib/LiquidFlow/LiquidFlowMaterialProperties.h
wenqing: Renamed it with liquid_density. The struct may get solid density in future development.

ProcessLib/LiquidFlow/LiquidFlowMaterialProperties.h
wenqing: Corrected.

ProcessLib/LiquidFlow/LiquidFlowProcess.cpp
```diff
@@ -0,0 +1,124 @@
...
+}
+
+void LiquidFlowProcess::preIterationConcreteProcess(const unsigned /*iter*/,
+                                                    GlobalVector const& /*x*/)
+{
+    // TODO
```
wenqing: Removed this function for current step.

ProcessLib/LiquidFlow/LiquidFlowProcess.h
wenqing: Removed.

Tests/ProcessLib/LiquidFlow/TestLiquidFlowMaterialProperties.cpp
wenqing: As mentioned above, currently I only consider the patch-wise heterogeneous porous medium properties, which occur in most applications of subsurface modelling. If element-wise heterogeneous porous  medium property is used, we can easily added by using  MeshElementProperties.

wenqing: This Contant is just a type of storage/porosity models, which may be defined by some empirical formulas.

ProcessLib/LiquidFlow/LiquidFlowProcess.cpp
```diff
@@ -0,0 +1,124 @@
...
+{
+LiquidFlowProcess::LiquidFlowProcess(
+    MeshLib::Mesh& mesh,
+    std::unique_ptr<AbstractJacobianAssembler>&& jacobian_assembler,
+    std::vector<std::unique_ptr<ParameterBase>> const& parameters,
+    unsigned const integration_order,
+    std::vector<std::reference_wrapper<ProcessVariable>>&& process_variables,
+    SecondaryVariableCollection&& secondary_variables,
+    NumLib::NamedFunctionCaller&& named_function_caller,
+    const bool compute_gravitational_term,
```
wenqing: Changed the order (also in LiquidFlowLocalAssembler). See the amended commit: 5421473.

Tests/ProcessLib/LiquidFlow/TestLiquidFlowMaterialProperties.cpp
endJunction: I suggest to use the properties that have been already merged to the master.

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler.h
```diff
@@ -0,0 +1,138 @@
...
+    using GlobalDimVectorType = typename ShapeMatricesType::GlobalDimVectorType;
+
+public:
+    LiquidFlowLocalAssembler(
+        MeshLib::Element const& element,
+        std::size_t const /*local_matrix_size*/,
+        bool const is_axially_symmetric,
+        unsigned const integration_order,
+        bool const compute_gravitational_term,
+        LiquidFlowMaterialProperties const& matetial_propertries)
```
TomFischer: spelling `matetial_properties`

ProcessLib/LiquidFlow/LiquidFlowMaterialProperties.h
```diff
@@ -0,0 +1,86 @@
...
+     */
+    double getMassCoefficient(const double p, const double T,
+                              const double porosity_variable,
+                              const double storage_variable,
+                              const unsigned material_group_id = 0) const;
+
+    double getLiquidDensity(const double p, const double T) const;
+
+    double getViscosity(const double p, const double T) const;
+
```
TomFischer: All the attributes of the struct are public. Is this intended?

wenqing: Yes. I can also change it to class, make the data private, and have LiquidFlowLocalAssembler as it friend.

wenqing: and enable it for a patch-wise heterogeneous domain with different material models.

wenqing:
@endJunction  I put the reorderVector to another PR (#1481), please review that PR first. In addition to the idea of the general models for porous medium properties,  the statistical data of  porous medium properties can be treated as mesh properties as the previous implementations for other processes. For this, a draft is provided in commit f6a025e, which will be dropped from this PR before merging and will be polished late on.

wenqing:
Commit: [Base] Use swap instead of the assign operator in reoerderVector is in PR #1481

TomFischer:
@wenqing wrote:
> In the current ogs6/master, the calculations of velocity and flux were implemented in wrong places for groundwater, heat conduction. A PR about the correction of such calculations for all non-deformation processes comes immediately once this one is merged.

Please contact me before moving the flux calculation since I am working on this topic, too.

wenqing:
@TomFischer  As for the velocity calculation, I have already had a work branch to be merged:
https://github.com/wenqing/ogs/tree/velocity

wenqing:
@TomFischer Sorry, to be sent as a PR.

ProcessLib/LiquidFlow/createLiquidFlowProcess.cpp
```diff
@@ -0,0 +1,99 @@
...
+    auto const g =
+        //! \ogs_file_attr{process__LIQUID_FLOW__gravitational_term__g}
+        gravity_param.getConfigAttributeOptional<double>("g");
+
+    int gravity_axis_id = -1;
+    if (*axis == "x")
+        gravity_axis_id = 0;
+    else if (*axis == "y")
+        gravity_axis_id = 1;
+    else if (*axis == "z")
```
endJunction: Specifying axis like this has few problems:
 - If there are other characters (' ' or '  ' or other invisible) output is not parsed correctly
 - It is hard to teach what the difference between "X" 'x' 'х' and 'Х' (the last two are from different alphabet)
 - If the mesh is in xy plane specifying gravity in z direction is undefined behavior, a segv at best.

A more general way of specifying body forces is preferred. That way when explicitly using one, two, or three values for a vector of same dimension as the input mesh (required), it is much more difficult to create wrong (non-intended) input. Have a look into the implementation of Richards flow process, please.

endJunction:
I have run the tests compiled with clang in debug mode using address and undefined behaviour sanitizers. The address sanitizer reported one issue
```
ERROR: AddressSanitizer: heap-use-after-free
```
see full output [here](https://gist.github.com/endJunction/603a7c673517f8f2eb2f73edf635428d)
Could you try to reproduce this on your plattform please.

During test I also have seen the line ctest failing sporadically but I cannot give any more specific details.

endJunction:
Could you also rename createLiquidFlow to CreateLiquidFlow which is the common way for naming files?

ProcessLib/LiquidFlow/createLiquidFlowProcess.cpp
```diff
@@ -0,0 +1,99 @@
...
+        gravity_axis_id = 2;
+
+    //! \ogs_file_param{process__LIQUID_FLOW__material_property}
+    auto const& mat_config = config.getConfigSubtree("material_property");
+
+    auto const& mat_ids =
+        mesh.getProperties().getPropertyVector<int>("MaterialIDs");
```
Yonghui56: what would happen if a mesh has the field name MaterialIDs while all the ids are exactly the same, in such case it is still homogeneous but in the code it will go to heterogeneous case, is it correct?

wenqing: If the mesh dos not has MaterialDs properpty, the material property is considered as  homogeneous. 

If the mesh has MaterialDs properpty with an unique ID, the code will first check whether the ID is smaller than the size of material model vector, and then fetch the corresponding model by the ID.

The current implementation of the porous media properties can be for the case of patch wise heterogeneous media with different material models. 

For the case of element wise heterogeneity with statistical data or the case of the same model with different parameters,  Parameter data as that in current GW and Richards will be added late on after this PR.

ProcessLib/LiquidFlow/CreateLiquidFlowProcess.cpp
```diff
@@ -48,22 +50,19 @@ std::unique_ptr<Process> createLiquidFlowProcess(
...
-    if (*axis == "x")
-        gravity_axis_id = 0;
-    else if (*axis == "y")
-        gravity_axis_id = 1;
-    else if (*axis == "z")
-        gravity_axis_id = 2;
+    // Get the gravity vector for the Darcy velocity
+    //! \ogs_file_param{process__LIQUID_FLOW__darcy_gravity}
+    auto const& darcy_g_config = config.getConfigSubtree("darcy_gravity");
+    int gravity_axis_id =
```
endJunction: const

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler-impl.h
```diff
@@ -121,7 +121,11 @@ void LiquidFlowLocalAssembler<ShapeFunction, IntegrationMethod, GlobalDim>::
 
     if (gravitational_axis_id >= 0)
     {
-        local_b.noalias() -=
+        // Note: Since perm, K, is a scalar number in this function,
+        // (gradN)*K*V becomes K*(grad N)*V. With the gravity vector of V,
+        // the simplification of (grad N)*V is the gravitational_axis_id th
+        // column of the transpose of (grad N) multiplied with rho_g.
+        local_b.noalias() +=
             fac * sm.dNdx.transpose().col(gravitational_axis_id) * rho_g;
```
endJunction: transpose().col() == row() ?

wenqing: It is for (grad(N))^T * V. If use sm.dNdx.row(), then (grad(N))^T * V should be written as
```
sm.dNdx.row(gravitational_axis_id).transpose() * rho_g
```
and it is less understandable.

PS: There is a ctest GravityDriven for this implementation.

wenqing:
@endJunction I have  to rebase to resolve the conflict with the ufz/master. My changes (added const to a variable) according to your last comment are in the last commit:  180cff8.

In the last commit, I also changed the sign of, or the range of g,  which is now changed to g>=0. This reflects its physical meaning.

wenqing:
@endJunction Tests with 1 million hex mesh.
Machine: envinf
Exe: Release

CPU time (s)

|   |  Scalar K |Tensor K | 
|---|---|---|
| Assembly  | 7.46527   | 32.6332  |

TomFischer:
> Besides, treated isotropic permeability as tensor causes a little raising in the number of solver iterations, which increases the total computation time.

Although it is not clear to me why the number of iterations in the linear solver increases at all and furthermore why the time for the solver using the matrix multiplication is twice as long as in the scalar multiplication case I think the PR can be merged. 

@wenqing Can you please open an issue containing the open questions and more information about your run time tests.

wenqing:
@TomFischer 
Conducted the tests (required around 7GB RAM each) again since the previous comparison is unfair. When I conducted the previous tests on envinf, there are many jobs are running. It might be that the test with tenor K raced to get compute resources such that it took longer time than the other. In the previous tests, even the time of Applying Dirichlet BCs of the test with tensor K is longer than that of  the other. Hereafter, the comparisons of computing times of the previous tests and the present tests are provided.

Previous Tests:

|   |  Scalar K |Tensor K | 
|---|---|---|
| Applying Dirichlet BCs  | 1.69089   | 2.63142  |  
| Assembly  | 7.46527   | 32.6332  |  
| Solver  | 49.98   | 94.0126   |  
| Output took |20.6294|47.8303|
| Total | 142.723  | 311.183|

Present Tests:

|   |  Scalar K |Tensor K | 
|---|---|---|
| Applying Dirichlet BCs  |  0.854581   |  0.910834   |  
| Assembly  | 4.71904   |  7.93674  |  
| Solver  | 43.6308    | 39.0444   |  
| Output took |19.114|44.7218|
| Total | 121.372   |149.402|

There is a mistake in the previous information about the Eigen linear solve: The case with scalar K took less iteration.  Actually that case took a little more iterations. Here is the information
  
|   |  Scalar K |Tensor K | 
|---|---|---|
| Linear solver iterations | 775    | 764   | 
Such difference in the numbers of  linear solver iteration is reasonable due to the round off error in taking scalar as tensor. Therefore, I will not create an issue about this difference.

wenqing:
Rebased.

endJunction:
Thanks for repeating the test. The new results look more trustful. When you have time please add that test to the ctests "LARGE", so other can experiment with it.

@wenqing Could you please check where the ufz/ogs/Tests/Data points to. It seems to me that in the last commit on ufz/ogs/master (https://github.com/ufz/ogs/commit/dd97ac9921976bfd6c2ca2405f09ceb8ec1fa296) you have set the submodule wrong.

wenqing:
@endJunction The pointer to Data is corrected. I will commit the large tests this afternoon.

wenqing:
@endJunction push -f is not allowed, I will add a new commit for this.

wenqing:
Done with git amend

PR #1469, opened 2016-10-06
[Jenkins] Run doxygen job in Ubuntu 16.04 container.

bilke:
16.04 has a more up-to-date doxygen package and so this
fixes some non-html-ized characters in the navigation.


endJunction:
Which gcc is used in there? Can we update travis to same version?

endJunction:
Issue is indeed fixed by upgrading doxygen. Thanks!

bilke:
Travis Trusty uses `4.8.4`, in Docker we use `4.9.x`. We can consolidate this if you like. Open another PR if yes.

PR #1470, opened 2016-10-06
[CTest] Move SmallDeformation_ring_plane_strain_axi to Large tests.

bilke:
Takes around 15-20 minutes.


endJunction:
@chleh I have tried BiCGSTAB with ILUT solver for the named case and it is something like 3-5 times faster, but still it takes almost 400s on a super fast 6800K (we are talking about debug build). I'm for moving this test into the LARGE category.

norihiro-w:
maybe you should try a direct solver. it is common for mechanical problems

chleh:
Sorry that I forget to move it to LARGE.
@norihiro-w: But LIS does not have direct solvers and currently we are not able to exclude a test in the LIS case.

PR #1472, opened 2016-10-09
Change create/getPropertyVector to return a pointer.

endJunction:
Current implementation returns an optional reference to a PropertyVector. This is semantically equivalent to a pointer. (Thanks to Christoph pointing this out). See the first commit's message for some details.

The main change is contained in the first commit, other are just updates to the new type.
Along the changed lines you will find changes:
 - `(*something).memberfunction()` to `something->memberfunction()`
 - removal of `T& obj_T = an_optional_to_T.get()` and direct usage of `*` and `->` operators.
 - removal of unecessary fill_n preceded by resize; std::vector::resize takes a second argument for the value. (Especially if the value is the type's default value.)
 - one untangling of nested if else with returns into a flat if error return ...
 - remove boost/optional.hpp include, fwd declare PropertyVector.
 - clang-format of all changed lines (automatically!).

It was a mistake from my side to suggest the usage of `boost::optional` for references. I hope this PR clarifies the semantic/usage of get and create property vectors.

### DIFF
--- a/Applications/DataExplorer/DataView/CreateStructuredGridDialog.cpp
+++ b/Applications/DataExplorer/DataView/CreateStructuredGridDialog.cpp
@@ -256,10 +256,10 @@ void CreateStructuredGridDialog::accept()
         return;
     }
 
-    boost::optional<MeshLib::PropertyVector<int>&> mat_ids (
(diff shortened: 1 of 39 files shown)


endJunction:
Jenkins, test this please!

norihiro-w:
looks good

TomFischer:
:+1:

PR #1474, opened 2016-10-11
Boundary element search along polylines for Line3 elements.

endJunction:
In the current implementation the Neumann boundary conditions applied on second order meshes (e.g. quad8) is not working correctly.

To fix it the BoundaryElementsAlongPolylines searcher is extended for Line3 element type.
Also a MeshLib::Element::construct function is required to construct new elements polymorphically.

### DIFF
--- a/MeshGeoToolsLib/BoundaryElementsAlongPolyline.cpp
+++ b/MeshGeoToolsLib/BoundaryElementsAlongPolyline.cpp
@@ -9,6 +9,7 @@
 #include "BoundaryElementsAlongPolyline.h"
 
 #include <algorithm>
+#include <typeinfo>
 
 #include "GeoLib/Polyline.h"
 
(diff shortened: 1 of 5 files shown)


TomFischer:
If tests are green: :+1:

MeshGeoToolsLib/BoundaryElementsAlongPolyline.cpp
```diff
@@ -88,18 +89,38 @@ bool BoundaryElementsAlongPolyline::includesAllEdgeNodeIDs(const std::vector<std
...
-        new_edge = new MeshLib::Line(new_nodes);
+    // The first node of the edge should be always closer to the beginning of
+    // the polyline than other nodes.
+    if (edge_node_distances_along_ply.front() >
+            edge_node_distances_along_ply.back() ||
+        (ply.isClosed() &&
+         edge_node_distances_along_ply.back() == node_ids_on_poly.size() - 1))
+    {  // Create a new element with reversed local node index
+        auto new_nodes = new MeshLib::Node*[edge.getNumberOfNodes()];
+        if (auto const* e = dynamic_cast<MeshLib::Line const*>(&edge))
```
norihiro-w: isn't it possible to check it by the number of nodes (2:linear, 3:quadratic) of the edge object?

endJunction: Can be done, although I find this to be more explicit.
If some one also votes for a change, I will change it.

MeshLib/Elements/Element.h
```diff
@@ -173,6 +173,14 @@ class Element
     virtual Element* clone() const = 0;
 
     /**
+     * Constructs a new object polymorphically. This is similar to clone, but
+     * accepts new nodes and id.
+     * \pre The length of the \c nodes vector is equal to the derived element's
+     * total number of nodes.
+     */
+    virtual Element* construct(Node** nodes, std::size_t id) const = 0;
```
norihiro-w: it doesn't construct itself. `clone()` also calls the constructor of the same type, isn't it? since the argument is different, you can still use the same function name.

norihiro-w:
:+1: if tests are working

endJunction:
Hmm, the rename `construct` to `clone` caused an ambiguity where the function is referred to without arguments... If it is to complicate (or too ugly) to fix it, I'll revert the rename.

norihiro-w:
i see

PR #1480, opened 2016-10-13
CTests on Windows

bilke:
- `time`-wrapper is now optional and was removed from all tests
- All tests using the `vtkdiff`-tester run on Windows

### DIFF
--- a/scripts/jenkins/msvc.groovy
+++ b/scripts/jenkins/msvc.groovy
@@ -26,6 +26,7 @@ node('win && conan') {
 
         stage 'Test (Win)'
         build.win 'build', 'tests'
+        build.win 'build', 'ctest'
 
         stage 'Data Explorer (Win)'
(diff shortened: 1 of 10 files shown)


scripts/jenkins/msvc.groovy
```diff
@@ -26,6 +26,7 @@ node('win && conan') {
 
         stage 'Test (Win)'
         build.win 'build', 'tests'
+        build.win 'build', 'ctest'
```
endJunction: This is release mode I hope; it could take too much time in debug even for non-LARGE ctests.

bilke: Yes [it is release](https://github.com/ufz/ogs/blob/master/scripts/jenkins/lib/build.groovy#L24).

bilke:
Resolved conan error by recreating libgeotifff packages.

PR #1481, opened 2016-10-14
Reorder vector with a set of given indices

wenqing:
As tilted. Although BaseLib::quicksort can be used to reorder a vector, it changes the given index vector as well.

The application of this function is in PR #1468, the which the function re-orders several vectors with one index vector.

### DIFF
--- a/BaseLib/reorderVector.h
+++ b/BaseLib/reorderVector.h
@@ -17,23 +17,19 @@
 namespace BaseLib
 {
 /**
- * Reorder a vector by a given index vector.
- *  From
- *  <a href="reorderV">http://stackoverflow.com/questions/838384/reorder-vector-using-a-vector-of-indices</a>
+ *  Reorder a vector by a given index vector.
  *
- *  Note: The simple version on that page is taken, which is good enough in performance
- *        for medium size vectors.
+ *  Note: It is good enough in performance for medium size vectors.
  */
 template <typename ValueType, typename IndexType>
 void reorderVector(std::vector<ValueType>& v,
                    std::vector<IndexType> const& order)
 {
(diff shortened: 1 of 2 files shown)


chleh:
I think the algorithm you implemented is rather hard to verify (especially since there are no comments about how it works)
If you'd create a temporary vector, the implementation would be `O(n)` and straightforward.
Furthermore, if changing the permutation vector is not an issue, you could use `BaseLib::quicksort()`.
(Conversely, `BaseLib::quicksort()` could probably be improved by using the vector reordering method from this PR.
The Code is OK. If you don't want to change it, I'm fine with it.
The Test case is only a bit basic. A better test case would be to create a random permutation (`std::iota`, `std::random_shuffle`) and reorder the vector according to that.

chleh:
:+1:

wenqing:
@chleh  
The previous code,  the origin of which is from stackoverflow, is discarded. I reimplemented the function with a buffer vector according to you comment. Besides, the unit test gets a random generated vector and a randomly shuffled permutation to check the function.
  
The permutation vector is assumed to to untouched such that it can be used to reorder more than one vectors.

chleh:
@wenqing Thank you for changing the code.

wenqing:
#include <numeric> for std:: iota is added for some compilers by amending the commit.

wenqing:
@chleh Thank you for your comments, which make this implementation better.

wenqing:
There is a problem for vector<unique_ptr>, where the type of the elements dose have a copy ctor. Working on it.

wenqing:
The problem is solved by using swap instead of assign operator.

norihiro-w:
:+1:

endJunction:
:fast_forward:

PR #1482, opened 2016-10-14
[PL] TES: Remove source tree modifier.

endJunction:
This part was modifying the source tree and prohibited
parallel builds with different settings of shape matrices
(fixed or dynamic).


chleh:
Are there effects on the build time? In particular for the following scenario:
```
make ogs # Debug build
touch ProcessLib/TES/TESLocalAssemblerInner-impl.h # or something that makes ccache recompile
time make ogs # that's the interesting figure
```
Because this construction was introduced for this very case (i.e. frequent changes to the local assembler)
If only negligible effect: :fast_forward:

endJunction:
Measurement with ccache, make -j2 ogs, debug, shared libraries, dynamic shape matrices, clang on this branch (merged together with EigenAlignment, otherwise it does not compile for me):
State: everything build, now rename a template parameter in TESLocalAssemblerInner-impl.h:
make -j2 ogs  42.13s user 2.92s system 145% cpu 30.924 total
another rename, (just to be sure):
make -j2 ogs  43.13s user 2.82s system 146% cpu 31.400 total

On EigenAlignment branch (w/o this PR):
make -j2 ogs  13.65s user 1.08s system 99% cpu 14.832 total
make -j2 ogs  14.05s user 1.16s system 99% cpu 15.326 total


So this particular thing is 15s faster.

On the other hand parallel builds are not compiling at all. I'd prefer correctness for speed.

It is possible to use that trick, but it must not modify the source tree; the resulting file must be placed in the binary tree. Then it would be consistent across different builds.

chleh:
> the resulting file must be placed in the binary tree.

That's hardly possible.

Thanks for the measurement. I'd actually prefer the 15s time saving. Waiting 15s longer is pretty painful when you're rebuilding during a debug session. But TES is basically finished, so it's not too important. :+1:

wenqing:
There is a compilation warning in the TES of the current master. Since it is related TES, I put this issue here:

```
e:\jenkins\workspace\ogs-6\ufz\pr-1481\ogs\processlib\tes\TESLocalAssemblerInner-impl.h(235) : warning C4244: 'argument' : conversion from 'const __int64' to 'const unsigned int', possible loss of data
        e:\jenkins\workspace\ogs-6\ufz\pr-1481\ogs\processlib\tes\TESLocalAssemblerInner-impl.h(227) : while compiling class template member function 'void ProcessLib::TES::TESLocalAssemblerInner<ProcessLib::detail::LocalAssemblerTraitsFixed<EigenDynamicShapeMatrixPolicy<ShapeFunction,0>,0,0,0>>::assembleIntegrationPoint(unsigned int,const std::vector<double,std::allocator<_Ty>> &,const NumLib::ShapeMatrices<EigenDynamicShapeMatrixPolicy<ShapeFunction,GlobalDim>::NodalRowVectorType,EigenDynamicShapeMatrixPolicy<ShapeFunction,GlobalDim>::DimNodalMatrixType,EigenDynamicShapeMatrixPolicy<ShapeFunction,GlobalDim>::DimMatrixType,EigenDynamicShapeMatrixPolicy<ShapeFunction,GlobalDim>::GlobalDimNodalMatrixType> &,const double,Eigen::Map<Derived,0,Eigen::Stride<0,0>> &,Eigen::Map<Derived,0,Eigen::Stride<0,0>> &,Eigen::Map<Eigen::Matrix<double,-1,1,0,-1,1>,0,Eigen::Stride<0,0>> &)'
        with
        [
            ShapeFunction=void
,            _Ty=double
,            Derived=Eigen::Matrix<double,-1,-1,1,-1,-1>
        ]
        e:\jenkins\workspace\ogs-6\ufz\pr-1481\ogs\processlib\tes\TESLocalAssemblerInner-fwd.h(22) : see reference to class template instantiation 'ProcessLib::TES::TESLocalAssemblerInner<ProcessLib::detail::LocalAssemblerTraitsFixed<EigenDynamicShapeMatrixPolicy<ShapeFunction,0>,0,0,0>>' being compiled
        with
        [
            ShapeFunction=void
        ]
```

TomFischer:
Looks good :+1:

PR #1483, opened 2016-10-14
Eigen aligned storage

endJunction:
Some part of my OS changed in the course of last month and the default heap allocation is no longer sufficiently aligned for the Eigen matrices. This is well known to the Eigen people and is discussed in their [documention](http://eigen.tuxfamily.org/dox/group__DenseMatrixManipulation__Alignement.html).

Until now we have tried not to see this special requirement but now a point where the code is no longer running correctly on modern linux' program set is reached.

There might be more places where this changes are needed and so far I only changed the minimum to get all the tests running.

At one place (MechanicsBase) I had to add a copy-ctor required by the operator new (though I don't understand all details why).

### DIFF
--- a/ProcessLib/SmallDeformation/SmallDeformationFEM.h
+++ b/ProcessLib/SmallDeformation/SmallDeformationFEM.h
@@ -41,9 +41,22 @@ struct IntegrationPointData final
     {
     }
 
-#if defined(_MSC_VER) && _MSC_VER < 1900
-    // The default generated move-ctor is correctly generated for other
-    // compilers.
+    explicit IntegrationPointData(IntegrationPointData const& other)
+        : _b_matrices(other._b_matrices),
+          _sigma(other._sigma),
+          _sigma_prev(other._sigma_prev),
+          _eps(other._eps),
+          _eps_prev(other._eps_prev),
+          _solid_material(other._solid_material),
+          _material_state_variables(other._material_state_variables->clone()),
+          _C(other._C),
+          _detJ(other._detJ),
+          _integralMeasure(other._integralMeasure)
+    {
+    }
+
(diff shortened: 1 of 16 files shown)


ProcessLib/Utils/InitShapeMatrices.h
```diff
@@ -10,6 +10,7 @@
 #define PROCESSLIB_UTILS_INIT_SHAPE_MATRICES_H_
 
 #include <vector>
+#include <Eigen/StdVector>
```
chleh: That might be obsolete now, cf. http://eigen.tuxfamily.org/bz/show_bug.cgi?id=829
Maybe this include even causes some problems. Could you check that, please?

ProcessLib/SmallDeformation/SmallDeformationFEM.h
```diff
@@ -39,9 +41,22 @@ struct IntegrationPointData final
     {
     }
 
-#if defined(_MSC_VER) && _MSC_VER < 1900
-    // The default generated move-ctor is correctly generated for other
-    // compilers.
+    explicit IntegrationPointData(IntegrationPointData const& other)
```
chleh: As suggested [here](https://eigen.tuxfamily.org/dox/group__TopicStructHavingEigenMembers.html#othersolutions), IMHO a private structure containing only Eigen members would be a solution superior to the `clone()` method.
But maybe even removing the `<Eigen/StdVector>` include makes `clone()` unnecessary.

endJunction:
@chleh Thanks for the pointer to the std=c++11/Eigen discussion. I'll check this.

endJunction:
I'm not sure why, but the alignment on my system is correct again. I'll keep the branch around for a while in case the problem pops up again.

PR #1484, opened 2016-10-15
Accelerate extrapolation

chleh:
I recognized that the shape matrices N are the same for a given element with a given set of integration points.

Therefore the extrapolation matrices only need to be computed once for each (element/integration point) combination, which speeds extrapolation up by roughly a factor of 10 (the TestExtrapolation unit test with 4th integration order, release build, fixed-size shape matrices, 20^3 hex elements; time needed for extrapolation before this PR: 0.075s, afterwards: 0.0072s).

The central assumption of this PR is that the set of shape matrices is uniquely defined by the number of nodes and the number of integration points in an element.
That should work with our current element types. I'd leave any better identification method for later refactoring.

Actually, the precomputation could even happen offline, s.t. linear solvers are not needed for extrapolation anymore.

### DIFF
--- a/Tests/NumLib/TestExtrapolation.cpp
+++ b/Tests/NumLib/TestExtrapolation.cpp
@@ -196,8 +196,8 @@ void extrapolate(TestProcess const& pcs, IntegrationPointValuesMethod method,
 {
     namespace LinAlg = MathLib::LinAlg;
 
-    auto const tolerance_dx  = 20.0 * std::numeric_limits<double>::epsilon();
-    auto const tolerance_res =  5.0 * std::numeric_limits<double>::epsilon();
+    auto const tolerance_dx  = 30.0 * std::numeric_limits<double>::epsilon();
+    auto const tolerance_res = 15.0 * std::numeric_limits<double>::epsilon();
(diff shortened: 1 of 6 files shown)


wenqing:
:+1:

Tests/NumLib/TestExtrapolation.cpp
```diff
@@ -195,8 +196,8 @@ void extrapolate(TestProcess const& pcs, IntegrationPointValuesMethod method,
 {
     namespace LinAlg = MathLib::LinAlg;
 
-    auto const tolerance_dx  = 20.0 * std::numeric_limits<double>::epsilon();
-    auto const tolerance_res =  5.0 * std::numeric_limits<double>::epsilon();
+    auto const tolerance_dx  = 30.0 * std::numeric_limits<double>::epsilon();
+    auto const tolerance_res = 15.0 * std::numeric_limits<double>::epsilon();
```
endJunction: Why the relaxed tolerances? (Just curious why this changed, whether it is 5*1e-16 or 15*1e-16 does not really matter).

PR #1487, opened 2016-10-17
[SD-LIE] Multiple fractures

norihiro-w:
This PR introduces a support of multiple non-intersecting fractures in SD-LIE process. Major changes are
- extend `getFractureMatrixDataInMesh()` to looking for multiple fracture groups having distinct material IDs
- extend the local assembler for matrix near fractures. Now the displacement function is given as `u(x) =  N(x) {u} + sum_i [phi_i(x) N(x) {g_i}]` where `phi_i(x)` is the levelest function, `g_i` is  the nodal enrichment values for `i` th fracture.
- extend the post processing tool

### DIFF
--- a/ProcessLib/SmallDeformationWithLIE/Common/MeshUtils.cpp
+++ b/ProcessLib/SmallDeformationWithLIE/Common/MeshUtils.cpp
@@ -62,73 +62,106 @@ private:
 void getFractureMatrixDataInMesh(
         MeshLib::Mesh const& mesh,
         std::vector<MeshLib::Element*>& vec_matrix_elements,
-        std::vector<MeshLib::Element*>& vec_fracture_elements,
-        std::vector<MeshLib::Element*>& vec_fracture_matrix_elements,
-        std::vector<MeshLib::Node*>& vec_fracture_nodes
+        std::vector<int>& vec_fracture_mat_IDs,
+        std::vector<std::vector<MeshLib::Element*>>& vec_fracture_elements,
+        std::vector<std::vector<MeshLib::Element*>>& vec_fracture_matrix_elements,
+        std::vector<std::vector<MeshLib::Node*>>& vec_fracture_nodes
         )
 {
     IsCrackTip isCrackTip(mesh);
 
     // get vectors of matrix elements and fracture elements
     vec_matrix_elements.reserve(mesh.getNumberOfElements());
+    std::vector<MeshLib::Element*> all_fracture_elements;
     for (MeshLib::Element* e : mesh.getElements())
     {
         if (e->getDimension() == mesh.getDimension())
             vec_matrix_elements.push_back(e);
         else
-            vec_fracture_elements.push_back(e);
+            all_fracture_elements.push_back(e);
     }
     DBUG("-> found total %d matrix elements and %d fracture elements",
-         vec_matrix_elements.size(), vec_fracture_elements.size());
+         vec_matrix_elements.size(), all_fracture_elements.size());
+
+    // get fracture material IDs
+    auto opt_material_ids(mesh.getProperties().getPropertyVector<int>("MaterialIDs"));
+    if (!opt_material_ids)
+        OGS_FATAL("MaterialIDs propery vector not found in a mesh");
+    for (MeshLib::Element* e : all_fracture_elements)
+        vec_fracture_mat_IDs.push_back((*opt_material_ids)[e->getID()]);
+    std::sort(vec_fracture_mat_IDs.begin(), vec_fracture_mat_IDs.end());
+    vec_fracture_mat_IDs.erase(
+        std::unique(vec_fracture_mat_IDs.begin(), vec_fracture_mat_IDs.end()),
+        vec_fracture_mat_IDs.end());
+    DBUG("-> found %d fracture material groups", vec_fracture_mat_IDs.size());
+
+    // create a vector of fracture elements for each group
+    vec_fracture_elements.resize(vec_fracture_mat_IDs.size());
+    for (unsigned frac_id=0; frac_id<vec_fracture_mat_IDs.size(); frac_id++)
+    {
+        const auto frac_mat_id = vec_fracture_mat_IDs[frac_id];
+        std::vector<MeshLib::Element*> &vec_elements = vec_fracture_elements[frac_id];
+        for (MeshLib::Element* e : all_fracture_elements)
+        {
+            if ((*opt_material_ids)[e->getID()] == frac_mat_id)
+                vec_elements.push_back(e);
+        }
+        DBUG("-> found %d elements on the fracture %d", vec_elements.size(), frac_id);
+    }
 
     // get a vector of fracture nodes
-    for (MeshLib::Element* e : vec_fracture_elements)
+    vec_fracture_nodes.resize(vec_fracture_mat_IDs.size());
+    for (unsigned frac_id=0; frac_id<vec_fracture_mat_IDs.size(); frac_id++)
     {
-        for (unsigned i=0; i<e->getNumberOfNodes(); i++)
+        std::vector<MeshLib::Node*> &vec_nodes = vec_fracture_nodes[frac_id];
+        for (MeshLib::Element* e : vec_fracture_elements[frac_id])
         {
-            if (isCrackTip(*e->getNode(i)))
-                continue;
-            vec_fracture_nodes.push_back(const_cast<MeshLib::Node*>(e->getNode(i)));
+            for (unsigned i=0; i<e->getNumberOfNodes(); i++)
+            {
+                if (isCrackTip(*e->getNode(i)))
+                    continue;
+                vec_nodes.push_back(const_cast<MeshLib::Node*>(e->getNode(i)));
+            }
         }
+        std::sort(vec_nodes.begin(), vec_nodes.end(),
+            [](MeshLib::Node* node1, MeshLib::Node* node2) { return (node1->getID() < node2->getID()); }
+            );
+        vec_nodes.erase(std::unique(vec_nodes.begin(), vec_nodes.end()), vec_nodes.end());
+        DBUG("-> found %d nodes on the fracture %d", vec_nodes.size(), frac_id);
     }
-    std::sort(vec_fracture_nodes.begin(), vec_fracture_nodes.end(),
-        [](MeshLib::Node* node1, MeshLib::Node* node2) { return (node1->getID() < node2->getID()); }
-        );
-    vec_fracture_nodes.erase(
-                std::unique(vec_fracture_nodes.begin(), vec_fracture_nodes.end()),
-                vec_fracture_nodes.end());
-    DBUG("-> found %d nodes on the fracture", vec_fracture_nodes.size());
 
     // create a vector fracture elements and connected matrix elements,
     // which are passed to a DoF table
-    // first, collect matrix elements
-    for (MeshLib::Element *e : vec_fracture_elements)
+    for (auto fracture_elements : vec_fracture_elements)
     {
-        for (unsigned i=0; i<e->getNumberOfBaseNodes(); i++)
+        std::vector<MeshLib::Element*> vec_ele;
+        // first, collect matrix elements
+        for (MeshLib::Element*e : fracture_elements)
         {
-            MeshLib::Node const* node = e->getNode(i);
-            if (isCrackTip(*node))
-                continue;
-            for (unsigned j=0; j<node->getNumberOfElements(); j++)
+            for (unsigned i=0; i<e->getNumberOfBaseNodes(); i++)
             {
-                // only matrix elements
-                if (node->getElement(j)->getDimension() == mesh.getDimension()-1)
+                MeshLib::Node const* node = e->getNode(i);
+                if (isCrackTip(*node))
                     continue;
-                vec_fracture_matrix_elements.push_back(const_cast<MeshLib::Element*>(node->getElement(j)));
+                for (unsigned j=0; j<node->getNumberOfElements(); j++)
+                {
+                    // only matrix elements
+                    if (node->getElement(j)->getDimension() < mesh.getDimension())
+                        continue;
+                    vec_ele.push_back(const_cast<MeshLib::Element*>(node->getElement(j)));
+                }
             }
         }
-    }
-    std::sort(vec_fracture_matrix_elements.begin(), vec_fracture_matrix_elements.end(),
-        [](MeshLib::Element* p1, MeshLib::Element* p2) { return (p1->getID() < p2->getID()); }
-        );
-    vec_fracture_matrix_elements.erase(
-                std::unique(vec_fracture_matrix_elements.begin(), vec_fracture_matrix_elements.end()),
-                vec_fracture_matrix_elements.end());
+        std::sort(vec_ele.begin(), vec_ele.end(),
+            [](MeshLib::Element* p1, MeshLib::Element* p2) { return (p1->getID() < p2->getID()); }
+            );
+        vec_ele.erase(std::unique(vec_ele.begin(), vec_ele.end()), vec_ele.end());
 
-    // second, append fracture elements
-    vec_fracture_matrix_elements.insert(
-                vec_fracture_matrix_elements.end(),
-                vec_fracture_elements.begin(), vec_fracture_elements.end());
+        // second, append fracture elements
+        vec_ele.insert(vec_ele.end(), fracture_elements.begin(), fracture_elements.end());
+
+        vec_fracture_matrix_elements.push_back(vec_ele);
+    }
 }
 
 }  // namespace SmallDeformationWithLIE
--- a/ProcessLib/SmallDeformationWithLIE/Common/MeshUtils.cpp
+++ b/ProcessLib/SmallDeformationWithLIE/Common/MeshUtils.cpp
@@ -8,6 +8,7 @@
 
 #include "MeshUtils.h"
 
+#include "BaseLib/makeVectorUnique.h"
 #include "MeshLib/MeshSearch/NodeSearch.h"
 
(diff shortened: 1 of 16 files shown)


norihiro-w:
@endJunction I got MSVC error C2797 ('SmallDeformationProcessData::_fracture_property': list initialization inside member initializer list or non-static data member initializer is not implemented
) for the following code.  Do you have any idea how to achieve move of a vector of unique pointers?

```
struct SmallDeformationProcessData
{
    SmallDeformationProcessData(SmallDeformationProcessData&& other)
        : _fracture_property{std::move(other._fracture_property)} // error C2797
    {
    }

    std::vector<std::unique_ptr<FractureProperty>> _fracture_property;
}
```

chleh:
@norihiro-w, maybe you have to use round brackets, not curly braces:
`: _fracture_property(std::move(other._fracture_property))`
instead of
`: _fracture_property{std::move(other._fracture_property)}`.

norihiro-w:
@chleh it worked! thanks!

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerMatrixNearFracture-impl.h
```diff
@@ -115,21 +118,60 @@ assembleWithJacobian(
...
+    //--------------------------------------------------------------------------------------
+    // prepare sub vectors, matrices for regular displacement (u) and displacement jumps (g)
+    //
+    // example with two fractures:
+    //     |b(u)|
+    // b = |b(g1)|
+    //     |b(g2)|
+    //
+    //     |J(u,u)  J(u,g1)  J(u,g2) |
+    // J = |J(g1,u) J(g1,g1) J(g2,g2)|
```
norihiro-w: g1,g2

ProcessLib/SmallDeformationWithLIE/Common/MeshUtils.cpp
```diff
@@ -62,73 +63,106 @@ class IsCrackTip
...
-         vec_matrix_elements.size(), vec_fracture_elements.size());
+         vec_matrix_elements.size(), all_fracture_elements.size());
+
+    // get fracture material IDs
+    auto opt_material_ids(mesh.getProperties().getPropertyVector<int>("MaterialIDs"));
+    if (!opt_material_ids)
+        OGS_FATAL("MaterialIDs propery vector not found in a mesh");
+    for (MeshLib::Element* e : all_fracture_elements)
+        vec_fracture_mat_IDs.push_back((*opt_material_ids)[e->getID()]);
+    BaseLib::makeVectorUnique(vec_fracture_mat_IDs);
```
chleh: Minor: acutally it looks like `std::set` would be a better type for `vec_fracture_mat_IDs`.

ProcessLib/SmallDeformationWithLIE/Common/MeshUtils.cpp
```diff
@@ -62,73 +63,106 @@ class IsCrackTip
...
+        for (MeshLib::Element* e : vec_fracture_elements[frac_id])
         {
-            if (isCrackTip(*e->getNode(i)))
-                continue;
-            vec_fracture_nodes.push_back(const_cast<MeshLib::Node*>(e->getNode(i)));
+            for (unsigned i=0; i<e->getNumberOfNodes(); i++)
+            {
+                if (isCrackTip(*e->getNode(i)))
+                    continue;
+                vec_nodes.push_back(const_cast<MeshLib::Node*>(e->getNode(i)));
```
chleh: Just wondering: Can't the nodes be `const`? (Then we probably need `std::vector<std::vector<MeshLib::Node const*>>& vec_fracture_nodes`, too)

ProcessLib/SmallDeformationWithLIE/Common/MeshUtils.cpp
```diff
@@ -62,73 +63,106 @@ class IsCrackTip
...
-        for (unsigned i=0; i<e->getNumberOfBaseNodes(); i++)
+        std::vector<MeshLib::Element*> vec_ele;
+        // first, collect matrix elements
+        for (MeshLib::Element*e : fracture_elements)
         {
-            MeshLib::Node const* node = e->getNode(i);
-            if (isCrackTip(*node))
-                continue;
-            for (unsigned j=0; j<node->getNumberOfElements(); j++)
+            for (unsigned i=0; i<e->getNumberOfBaseNodes(); i++)
```
chleh: Maybe add a comment that it is sufficient to iterate over base nodes, because they are already connected to all neighbours?

ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
```diff
@@ -52,43 +52,56 @@ PostProcessTool::PostProcessTool(
...
     {
-        auto duplicated_node = new MeshLib::Node(org_node->getCoords(), new_nodes.size());
-        new_nodes.push_back(duplicated_node);
-        _map_dup_newNodeIDs[org_node->getID()] = duplicated_node->getID();
+        for (auto const* org_node : vec_fracture_nodes)
+        {
+            auto duplicated_node = new MeshLib::Node(org_node->getCoords(), new_nodes.size());
+            new_nodes.push_back(duplicated_node);
+            assert(_map_dup_newNodeIDs.count(org_node->getID())==0);
+            _map_dup_newNodeIDs[org_node->getID()] = duplicated_node->getID();
```
chleh: Minor: I'd do the check like this (if `_map_dup_newNodeIDs` is a map):
```{.cpp}
if (! _map_dup_newNodeIDs.emplace(org_node->getID(), duplicated_node->getID()).second)
    OGS_FATAL("error ...");
```

ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
```diff
@@ -52,43 +52,56 @@ PostProcessTool::PostProcessTool(
...
-        auto const eid = org_e->getID();
-
-        // keep original if the element has levelset=0
-        if ((*prop_levelset)[eid] == 0)
-            continue;
-
-        // replace fracture nodes with duplicated ones
-        MeshLib::Element* e = new_eles[eid];
-        for (unsigned i=0; i<e->getNumberOfNodes(); i++)
+        auto const& vec_fracutre_matrix_elements = vec_vec_fracutre_matrix_elements[fracture_id];
```
chleh: typo `fracutre`

ProcessLib/SmallDeformationWithLIE/CreateSmallDeformationProcess.cpp
```diff
@@ -136,16 +137,26 @@ createSmallDeformationProcess(
...
-    auto fracture_properties_config = config.getConfigSubtree("fracture_properties");
-    auto &para_b0 = ProcessLib::findParameter<double>(fracture_properties_config, "initial_aperture", parameters, 1);
-    std::unique_ptr<FractureProperty> frac_prop(new FractureProperty());
-    frac_prop->mat_id = fracture_properties_config.getConfigParameter<int>("material_id");
-    frac_prop->aperture0 = &para_b0;
+    for (auto fracture_properties_config : config.getConfigSubtreeList("fracture_properties"))
+    {
+        auto& para_b0 = ProcessLib::findParameter<double>(
+            fracture_properties_config, "initial_aperture", parameters, 1);
+        auto frac_prop(new FractureProperty());
```
chleh: Using std::unique_ptr directly would be better.

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerMatrixNearFracture-impl.h
```diff
@@ -115,21 +118,60 @@ assembleWithJacobian(
...
+    //--------------------------------------------------------------------------------------
+    auto local_b_u = local_b.segment<N_DOF_PER_VAR>(0);
+    std::vector<BlockVectorType> vec_local_b_g;
+    for (unsigned i=0; i<n_fractures; i++)
+        vec_local_b_g.push_back(local_b.segment<N_DOF_PER_VAR>(N_DOF_PER_VAR*(i+1)));
+
+    auto local_J_uu = local_J.block<N_DOF_PER_VAR, N_DOF_PER_VAR>(0, 0);
+    std::vector<BlockMatrixType> vec_local_J_ug;
+    std::vector<BlockMatrixType> vec_local_J_gu;
+    std::vector<std::vector<BlockMatrixType>> vec_local_J_gg(n_fractures);
```
chleh: Maybe reserve sizes for the vectors?

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerMatrixNearFracture-impl.h
```diff
@@ -163,30 +209,36 @@ assembleWithJacobian(
...
-        // r_ru = B^T Stress = B^T C B (u + phi*[u])
-        localRhs_ru.noalias() -= B.transpose() * sigma * detJ * wp.getWeight() * integralMeasure;
-
-        // r_[u] = (phi*B)^T Stress = (phi*B)^T C B (u + phi*[u])
-        localRhs_du.noalias() -= levelsets * B.transpose() * sigma * detJ * wp.getWeight() * integralMeasure;
+        // r_u = B^T * Sigma = B^T * C * B * (u+phi*[u])
+        // r_[u] = (phi*B)^T * Sigma = (phi*B)^T * C * B * (u+phi*[u])
+        local_b_u.noalias() -= B.transpose() * sigma * ip_factor;
+        for (unsigned i=0; i<n_fractures; i++)
+            vec_local_b_g[i].noalias() -= levelsets[i] * B.transpose() * sigma * ip_factor;
```
chleh: Is it really better to store the Eigen::Blocks in a std:vector instead of building them on demand? E.g., `vec_local_b_g` seems to be used only here.

ProcessLib/SmallDeformationWithLIE/SmallDeformationProcess.cpp
```diff
@@ -57,25 +57,57 @@ SmallDeformationProcess<DisplacementDim>::SmallDeformationProcess(
...
-                        *_process_data._fracture_property.get());
+    if (_vec_fracture_mat_IDs.size() != _process_data._vec_fracture_property.size())
+        OGS_FATAL("The number of the given fracture properties (%d) are not consistent"
+                  " with the number of fracture groups in a mesh (%d).",
+                  _process_data._vec_fracture_property.size(),
+                  _vec_fracture_mat_IDs.size());
+
+    // create a map from a material ID to a fracture ID
+    auto max_frac_mat_id = std::max_element(_vec_fracture_mat_IDs.begin(),
+                                            _vec_fracture_mat_IDs.end());
```
chleh: Currently fractures are identified with some material IDs. Does that also work, if somebody already has multiple materials in the matrix (i.e., non-fracture part)?

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerMatrixNearFracture-impl.h
```diff
@@ -163,30 +209,36 @@ assembleWithJacobian(
...
-        // J_[u][u] += (levelset B)^T * C * (levelset B)
-        localA_gg.noalias() += levelsets * B.transpose() * C * levelsets * B * detJ * wp.getWeight() * integralMeasure;
+        local_J_uu.noalias() += B.transpose() * C * B * ip_factor;
+
+        for (unsigned i=0; i<n_fractures; i++)
+        {
+            // J_u[u] += B^T * C * (levelset * B)
+            vec_local_J_ug[i].noalias() += B.transpose() * C *(levelsets[i] * B) * ip_factor;
+
+            // J_[u]u += (levelset * B)^T * C * B
```
wenqing: According to the following line
``(levelset * B)^T`` should be ``(levelset * B^T)``

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerMatrixNearFracture-impl.h
```diff
@@ -163,30 +209,36 @@ assembleWithJacobian(
...
-
-        // J_u[u] += B^T * C * B * levelset
-        localA_ug.noalias() += B.transpose() * C * levelsets * B * detJ * wp.getWeight() * integralMeasure;
-
-        // J_[u]u += (levelset B)^T * C * B
-        localA_gu.noalias() += levelsets * B.transpose() * C * B * detJ * wp.getWeight() * integralMeasure;
-
-        // J_[u][u] += (levelset B)^T * C * (levelset B)
-        localA_gg.noalias() += levelsets * B.transpose() * C * levelsets * B * detJ * wp.getWeight() * integralMeasure;
+        local_J_uu.noalias() += B.transpose() * C * B * ip_factor;
```
wenqing: ``B^T * C`` is performed several times in the following two loops. If  n_fractures is a large number, a computational optimization might be needed here after.

ProcessLib/SmallDeformationWithLIE/Common/MeshUtils.cpp
norihiro-w: could be, but I would not include optimisation in this PR.

ProcessLib/SmallDeformationWithLIE/Common/MeshUtils.cpp
norihiro-w: no, because the vector will be passed to MeshSubset which accepts `vector<Node*> const*`.

ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
norihiro-w: again...  sorry

ProcessLib/SmallDeformationWithLIE/CreateSmallDeformationProcess.cpp
norihiro-w: i don't see a benefit. keep it as it is

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerMatrixNearFracture-impl.h
norihiro-w: optimization will not be included in this PR.

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerMatrixNearFracture-impl.h
norihiro-w: only here, but is called for each integration point.

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerMatrixNearFracture-impl.h
norihiro-w: yes you are right. but I will not include the optimisation at this stage. Because I'll not maintain the LIE stuff and someone has to take over this, I think readability is more important.

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerMatrixNearFracture-impl.h
norihiro-w: Correct expression is `(levelset * B)^T`. Because the levelset I use is a scalar, we can code it as below.

ProcessLib/SmallDeformationWithLIE/SmallDeformationProcess.cpp
norihiro-w: yes, it works, because the code accesses `_process_data._map_materialID_to_fractureID` only with material IDs of fractures.

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerMatrixNearFracture-impl.h
chleh: Since this code is executed in every local assemble, I think it would be very good to do the reservation.
It's three lines of code. Please do it. We do it in many other places, too. It stresses good style.

ProcessLib/SmallDeformationWithLIE/SmallDeformationProcess.cpp
chleh: Thanks.

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerMatrixNearFracture-impl.h
norihiro-w: as i said. i'll not include optimisation. if optimisation is necessary, one should first profile the code. if this part significantly slows down local assembly, one can fix it easily.

ProcessLib/SmallDeformationWithLIE/LocalAssembler/SmallDeformationLocalAssemblerMatrixNearFracture-impl.h
wenqing: I see.

PR #1488, opened 2016-10-17
[SD-LIE] supporting fracture tips

norihiro-w:
follow up of #1452

Current implementation supports a fracture cutting through an entire domain. This PR will extend it to enabling to have fracture tips inside of a domain. What the implementation does is simply, not to create DoFs for those tip nodes, meaning displacement jumps are zero at those nodes.

### DIFF
--- a/ProcessLib/SmallDeformationWithLIE/Common/MeshUtils.cpp
+++ b/ProcessLib/SmallDeformationWithLIE/Common/MeshUtils.cpp
@@ -18,6 +18,10 @@ namespace SmallDeformationWithLIE
 namespace
 {
 
+// A class to check whether a node is located on a crack tip with
+// the following conditions:
+// - the number of connected fracture elements is one
+// - the node is not located on a domain boundary
 class IsCrackTip
 {
 public:
--- a/ProcessLib/SmallDeformationWithLIE/Common/MeshUtils.cpp
+++ b/ProcessLib/SmallDeformationWithLIE/Common/MeshUtils.cpp
@@ -8,11 +8,52 @@
 
 #include "MeshUtils.h"
 
+#include "MeshLib/MeshSearch/NodeSearch.h"
 
 namespace ProcessLib
 {
 namespace SmallDeformationWithLIE
 {
+
+namespace
+{
+
+class IsCrackTip
+{
+public:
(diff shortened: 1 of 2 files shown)


norihiro-w:
@bilke Jenkins is failing with the following error. can you please check it? 
`org.jenkinsci.plugins.workflow.cps.steps.ParallelStepException: Parallel step gcc failed`

ProcessLib/SmallDeformationWithLIE/Common/MeshUtils.cpp
```diff
@@ -8,12 +8,53 @@
...
 
 namespace ProcessLib
 {
 namespace SmallDeformationWithLIE
 {
 
+namespace
+{
+
+class IsCrackTip
```
chleh: A bit of documentation would be nice.

ProcessLib/SmallDeformationWithLIE/Common/MeshUtils.cpp
```diff
@@ -8,12 +8,53 @@
...
+public:
+    explicit IsCrackTip(MeshLib::Mesh const& mesh)
+        : _fracture_element_dim(mesh.getDimension()-1)
+    {
+        _is_internal_node.resize(mesh.getNumberOfNodes(), true);
+
+        MeshLib::NodeSearch nodeSearch(mesh);
+        nodeSearch.searchBoundaryNodes();
+        for (auto i : nodeSearch.getSearchedNodeIDs())
+            _is_internal_node[i] = false;
```
chleh: Is it only a short-cut for later computation to store whether a node is internal or is it truly essential to know that?

norihiro-w: it's used at line 32 to check whether a node is internal or not. one can also use `std::find()` to a result of  `nodeSearch.getSearchedNodeIDs()`. I thought this implementation is more efficient.

ProcessLib/SmallDeformationWithLIE/Common/MeshUtils.cpp
norihiro-w: done

PR #1490, opened 2016-10-19
Does that restore the lost pull requests?

chleh:
I created this branch from c50dad011bcb21ab7d32742925b591952973a9b1, which was still accessible on github.

I suggest that force-pushing to ogs/master be disabled. It can be handy from time to time, but that can be explicitly enabled in those rare cases, and disabled again afterwards.


norihiro-w:
:+1: thanks!

TomFischer:
:+1:

PR #1491, opened 2016-10-19
[SD-LIE] post-processing tool

norihiro-w:
This PR adds a post-processing tool for results obtained with the LIE approach. The tool reads a PVD file and process each VTU file with the following jobs:
- create a new mesh which includes duplicated fracture nodes. This is needed to compute discontinuities in displacement field and visualize it on ParaView
- compute total displacements based on regular displacements, displacement jumps, and levelset functions

### DIFF
--- a/ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
+++ b/ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
@@ -77,14 +77,17 @@ PostProcessTool::PostProcessTool(
         MeshLib::Element* e = new_eles[eid];
         for (unsigned i=0; i<e->getNumberOfNodes(); i++)
         {
+            // only fracture nodes
             auto itr = _map_dup_newNodeIDs.find(e->getNodeIndex(i));
             if (itr == _map_dup_newNodeIDs.end())
                 continue;
 
+            // check if a node belongs to the particular fracture group
             auto itr2 = std::find_if(vec_fracture_nodes.begin(), vec_fracture_nodes.end(),
                                      [&](MeshLib::Node const*node) { return node->getID()==e->getNodeIndex(i);});
             if (itr2 == vec_fracture_nodes.end())
                 continue;
+
             e->setNode(i, new_nodes[itr->second]);
         }
     }
--- a/ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
+++ b/ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
@@ -1,0 +1,205 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ */
+
+#include "PostUtils.h"
+
+#include <map>
+#include <vector>
+
+#include "MeshLib/Elements/Element.h"
+#include "MeshLib/Node.h"
+#include "MeshLib/MeshEditing/DuplicateMeshComponents.h"
+
+namespace ProcessLib
+{
+namespace SmallDeformationWithLIE
+{
+
+namespace
+{
+
+template <typename T>
+inline void sort_unique(std::vector<T> &vec)
+{
+    std::sort(vec.begin(), vec.end());
+    vec.erase(std::unique(vec.begin(), vec.end()), vec.end());
+}
+
+} // no named namespace
+
+PostProcessTool::PostProcessTool(
+    MeshLib::Mesh const& org_mesh,
+    std::vector<MeshLib::Node*> const& vec_fracture_nodes,
+    std::vector<MeshLib::Element*> const& vec_fracutre_matrix_elements)
+    :_org_mesh(org_mesh)
+{
+    if (!org_mesh.getProperties().hasPropertyVector("displacement")
+        || !org_mesh.getProperties().hasPropertyVector("displacement_jump1")
+        || !org_mesh.getProperties().hasPropertyVector("levelset1")
+        )
+    {
+        OGS_FATAL("The given mesh does not have relevant properties");
+    }
+
+    // clone nodes and elements
+    std::vector<MeshLib::Node*> new_nodes(MeshLib::copyNodeVector(org_mesh.getNodes()));
+    std::vector<MeshLib::Element*> new_eles(
+        MeshLib::copyElementVector(org_mesh.getElements(), new_nodes));
+
+    // duplicate fracture nodes
+    for (auto const* org_node : vec_fracture_nodes)
+    {
+        auto duplicated_node = new MeshLib::Node(org_node->getCoords(), new_nodes.size());
+        new_nodes.push_back(duplicated_node);
+        _map_dup_newNodeIDs[org_node->getID()] = duplicated_node->getID();
+    }
+
+    // split elements using the new duplicated nodes
+    auto prop_levelset = org_mesh.getProperties().getPropertyVector<double>("levelset1");
+    for (auto const* org_e : vec_fracutre_matrix_elements)
+    {
+        // only matrix elements
+        if (org_e->getDimension() != org_mesh.getDimension())
+            continue;
+
+        auto const eid = org_e->getID();
+
+        // keep original if the element has levelset=0
+        if ((*prop_levelset)[eid] == 0)
+            continue;
+
+        // replace fracture nodes with duplicated ones
+        MeshLib::Element* e = new_eles[eid];
+        for (unsigned i=0; i<e->getNumberOfNodes(); i++)
+        {
+            auto itr = _map_dup_newNodeIDs.find(e->getNodeIndex(i));
+            if (itr == _map_dup_newNodeIDs.end())
+                continue;
+
+            auto itr2 = std::find_if(vec_fracture_nodes.begin(), vec_fracture_nodes.end(),
+                                     [&](MeshLib::Node const*node) { return node->getID()==e->getNodeIndex(i);});
+            if (itr2 == vec_fracture_nodes.end())
+                continue;
+            e->setNode(i, new_nodes[itr->second]);
+        }
+    }
+
+    // new mesh
+    _output_mesh.reset(new MeshLib::Mesh(org_mesh.getName(), new_nodes, new_eles));
+    createProperties<int>();
+    createProperties<double>();
+    copyProperties<int>();
+    copyProperties<double>();
+    calculateTotalDisplacement();
+}
+
+template <typename T>
+void PostProcessTool::createProperties()
+{
+    MeshLib::Properties const& src_properteis = _org_mesh.getProperties();
+    for (auto name : src_properteis.getPropertyVectorNames())
+    {
+        auto const* src_prop = src_properteis.getPropertyVector<T>(name);
+        if (!src_prop)
+            continue;
+        auto const n_src_comp = src_prop->getNumberOfComponents();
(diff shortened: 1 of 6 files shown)


endJunction:
Code looks fine, though I didn't check the details of manipulations there.
Would it be possible to have a test?
Some sort of input vtu and then this tool outputs the result which is compared (just vtkdiff on the fields will do) to expected output.

norihiro-w:
I added a test

ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+PostProcessTool::PostProcessTool(
+    MeshLib::Mesh const& org_mesh,
+    std::vector<MeshLib::Node*> const& vec_fracture_nodes,
+    std::vector<MeshLib::Element*> const& vec_fracutre_matrix_elements)
+    :_org_mesh(org_mesh)
+{
+    if (!org_mesh.getProperties().hasPropertyVector("displacement")
+        || !org_mesh.getProperties().hasPropertyVector("displacement_jump1")
+        || !org_mesh.getProperties().hasPropertyVector("levelset1")
```
chleh: Is the number '1' related to the *first* fracture?

ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
```diff
@@ -0,0 +1,205 @@
...
+    for (auto const* org_e : vec_fracutre_matrix_elements)
+    {
+        // only matrix elements
+        if (org_e->getDimension() != org_mesh.getDimension())
+            continue;
+
+        auto const eid = org_e->getID();
+
+        // keep original if the element has levelset=0
+        if ((*prop_levelset)[eid] == 0)
```
chleh: You are comparing double values here, aren't you? Is it safe to use strict equality?

ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
```diff
@@ -0,0 +1,205 @@
...
+        // replace fracture nodes with duplicated ones
+        MeshLib::Element* e = new_eles[eid];
+        for (unsigned i=0; i<e->getNumberOfNodes(); i++)
+        {
+            auto itr = _map_dup_newNodeIDs.find(e->getNodeIndex(i));
+            if (itr == _map_dup_newNodeIDs.end())
+                continue;
+
+            auto itr2 = std::find_if(vec_fracture_nodes.begin(), vec_fracture_nodes.end(),
+                                     [&](MeshLib::Node const*node) { return node->getID()==e->getNodeIndex(i);});
```
chleh: That looks a bit expensive, but maybe it's not so important.

ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
```diff
@@ -0,0 +1,205 @@
...
+            if (itr2 == vec_fracture_nodes.end())
```
chleh: Here you know, that node `i` is a fracture node, but you don't find it in `vec_fracture_nodes`. What does that mean in practice? Should there be a warning/error message? Probably there should be a comment in the source code.

ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
```diff
@@ -0,0 +1,205 @@
...
+            e->setNode(i, new_nodes[itr->second]);
+        }
+    }
+
+    // new mesh
+    _output_mesh.reset(new MeshLib::Mesh(org_mesh.getName(), new_nodes, new_eles));
+    createProperties<int>();
+    createProperties<double>();
+    copyProperties<int>();
+    copyProperties<double>();
```
chleh: Don't we in principle support other properties, too; e.g., char, unsigned, std::size_t ...?

ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
```diff
@@ -0,0 +1,205 @@
...
+    calculateTotalDisplacement();
+}
+
+template <typename T>
+void PostProcessTool::createProperties()
+{
+    MeshLib::Properties const& src_properteis = _org_mesh.getProperties();
+    for (auto name : src_properteis.getPropertyVectorNames())
+    {
+        auto const* src_prop = src_properteis.getPropertyVector<T>(name);
```
chleh: typo properties

ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
```diff
@@ -0,0 +1,205 @@
...
+        if (!src_prop)
+            continue;
+        auto const n_src_comp = src_prop->getNumberOfComponents();
+        // convert 2D vector to 3D. Otherwise Paraview Calculator filter does not recognize
+        // it as a vector
```
chleh: Should we really workaround Paraview here?

ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
```diff
@@ -0,0 +1,205 @@
...
+            new_prop->resize(src_prop->size());
+        }
+    }
+}
+
+template <typename T>
+void PostProcessTool::copyProperties()
+{
+    MeshLib::Properties const& src_properteis = _org_mesh.getProperties();
+    for (auto name : src_properteis.getPropertyVectorNames())
```
chleh: typo

ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+template <typename T>
+void PostProcessTool::createProperties()
+{
+    MeshLib::Properties const& src_properteis = _org_mesh.getProperties();
+    for (auto name : src_properteis.getPropertyVectorNames())
+    {
+        auto const* src_prop = src_properteis.getPropertyVector<T>(name);
+        if (!src_prop)
+            continue;
```
chleh: That can never happen, can it? Make fatal? Add comment `// should never happen`?

ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
```diff
@@ -0,0 +1,205 @@
...
+void PostProcessTool::copyProperties()
+{
+    MeshLib::Properties const& src_properteis = _org_mesh.getProperties();
+    for (auto name : src_properteis.getPropertyVectorNames())
+    {
+        auto const* src_prop = src_properteis.getPropertyVector<T>(name);
+        if (!src_prop)
+            continue;
+        auto* dest_prop = _output_mesh->getProperties().getPropertyVector<T>(name);
+        assert(dest_prop);
```
chleh: Better do a check also in release mode?

ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
```diff
@@ -0,0 +1,205 @@
...
+        auto const n_dest_comp = (n_src_comp==2) ? 3 : n_src_comp;
+
+        if (src_prop->getMeshItemType() == MeshLib::MeshItemType::Node)
+        {
+            auto new_prop =
+                _output_mesh->getProperties().createNewPropertyVector<T>(
+                    name, MeshLib::MeshItemType::Node, n_dest_comp);
+            new_prop->resize(_output_mesh->getNumberOfNodes() * n_dest_comp);
+        }
+        else if (src_prop->getMeshItemType() == MeshLib::MeshItemType::Cell)
```
chleh: If we later on add some field properties to VTK files, we will have to extend this code. We should keep that in mind.

ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
```diff
@@ -0,0 +1,205 @@
...
+{
+    // nodal value of levelset
+    std::vector<double> nodal_levelset(_output_mesh->getNodes().size(), 0.0);
+    auto const& ele_levelset =  *_output_mesh->getProperties().getPropertyVector<double>("levelset1");
+    for (MeshLib::Element const* e : _output_mesh->getElements())
+    {
+        if (e->getDimension() != _output_mesh->getDimension())
+            continue;
+        const double e_levelset = ele_levelset[e->getID()];
+        if (e_levelset == 0)
```
chleh: Is this double comparison safe?

ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+} // no named namespace
+
+PostProcessTool::PostProcessTool(
+    MeshLib::Mesh const& org_mesh,
+    std::vector<MeshLib::Node*> const& vec_fracture_nodes,
+    std::vector<MeshLib::Element*> const& vec_fracutre_matrix_elements)
+    :_org_mesh(org_mesh)
+{
+    if (!org_mesh.getProperties().hasPropertyVector("displacement")
```
chleh: Maybe all those property names shouldn't be hard-coded?

ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
norihiro-w: yes, it's corresponding to displacement_jump1

ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
norihiro-w: it's safe because the levelset I use takes 0 or 1.

ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
norihiro-w: i don't think it's important. the names will hardly be changed and does appear many places

ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
norihiro-w: not important

ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
norihiro-w: i will change it to assert

ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
norihiro-w: i can add a check here

ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
norihiro-w: no, debugging assert is enough, since creation of the properties is also done within this class.

ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
norihiro-w: yes

ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
norihiro-w: actually this is for the multiple fracture case, because I need to check if this node belongs a particular fracture group. will add a comment here

ProcessLib/SmallDeformationWithLIE/Common/PostUtils.cpp
chleh: I see.

PR #1495, opened 2016-10-21
[Mesh] check IDs of nonlinear nodes

norihiro-w:
Current implementation of `Mesh` class assumes nodes are stored in the order: 1. base nodes, 2. nonlinear nodes. But the assumption has not been checked by ogs and it's possible a user provides a mesh with wrong node ordering.

### DIFF
--- a/MeshLib/Mesh.cpp
+++ b/MeshLib/Mesh.cpp
@@ -42,6 +42,8 @@ Mesh::Mesh(const std::string &name,
     assert(n_base_nodes <= nodes.size());
     this->resetNodeIDs();
     this->resetElementIDs();
+    if (isNonlinear())
+        this->checkNonlinearNodeIDs();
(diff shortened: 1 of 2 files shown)


norihiro-w:
travis error is `dpkg-deb: error: `conan.deb' is not a debian format archive` and is not related to this PR

chleh:
:+1:
Doesn't the mentioned assumption seriously degrade the structure of any linear equation system assembled on nonlinear meshes? Why is this assumption necessary (only out of curiosity)?

norihiro-w:
because I noticed mesh files in Dima's HM examples are not following the assumption. The assumption came from ogs5 and can be changed if needed. But till then, because some algorithms and the node partitioned mesh class are based on this, we need the check.

chleh:
Thank you for the explanation.

PR #1502, opened 2016-10-27
Assert unique parameter and process variable names

chleh:
... since it is rather likely that users mess up their input files.
- ProjectData now stores parameters and process variables in a std::map
- ParameterBase::name is const now

I think it's best to review commit-wise since there are some commits with search & replace changes.
Note, however, that the "dependent changes for SD with LIE" are a tiny bit more than search & replace.

### DIFF
--- a/ProcessLib/Parameter/CurveScaledParameter.h
+++ b/ProcessLib/Parameter/CurveScaledParameter.h
@@ -19,10 +19,10 @@ namespace ProcessLib
 {
 template <typename T>
 struct CurveScaledParameter final : public Parameter<T> {
-    CurveScaledParameter(MathLib::PiecewiseLinearInterpolation const& curve,
+    CurveScaledParameter(std::string const& name_,
+                         MathLib::PiecewiseLinearInterpolation const& curve,
                          std::string const& parameter_name)
-        : _curve(curve),
-          _parameter_name(parameter_name)
+        : Parameter<T>(name_), _curve(curve), _parameter_name(parameter_name)
     {
     }
 
@@ -63,6 +63,7 @@ private:
 };
 
 std::unique_ptr<ParameterBase> createCurveScaledParameter(
+    std::string const& name,
     BaseLib::ConfigTree const& config,
     std::map<std::string,
              std::unique_ptr<MathLib::PiecewiseLinearInterpolation>> const&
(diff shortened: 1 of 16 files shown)


wenqing:
@chleh  ufz/master is just updated with the correct pointer to Data submodule, please rebase.

norihiro-w:
Why do classes other than ProjectData need `std::map`? If the container type is needed only for checking unique parameter names, it should be used only in ProjectData and not influence the rest of the codes.

ProcessLib/Parameter/CurveScaledParameter.h
```diff
@@ -19,19 +19,17 @@ namespace ProcessLib
 {
 template <typename T>
 struct CurveScaledParameter final : public Parameter<T> {
-    CurveScaledParameter(MathLib::PiecewiseLinearInterpolation const& curve,
+    CurveScaledParameter(std::string const& name_,
+                         MathLib::PiecewiseLinearInterpolation const& curve,
                          std::string const& parameter_name)
-        : _curve(curve),
-          _parameter_name(parameter_name)
+        : Parameter<T>(name_), _curve(curve), _parameter_name(parameter_name)
```
endJunction: I'm confused by Parameter name_ and parameter_name.
Renaming one of them would be IMO good idea.

chleh:
@norihiro-w The container type is already exposed to all of OGS (`std::vector<std::unique_ptr<ProcessLib::ParameterBase>>`, including the fact that we manage memory with `std::unique_ptr`). I changed the container type, so all the dependent code changes, too.
Since `std::map` stresses that the key (i.e. the name in this case) is unique, I think it's the better container type.
However, if you insist that everything has to be `std::vector`, I can take the change back.

norihiro-w:
i still don't get why other classes need a dictionary type of container. however i don't insist it if you and others think the changes makes sense.

wenqing:
Same with Nori, I do not get the idea. The uniqueness of Paramter can be checked in  ProjectData.

chleh:
std::map does not only check at read time, but asserts the uniqueness for the whole lifetime.

wenqing:
@chleh I see. Go forward if you think it is Ok.

TomFischer:
:fast_forward:

ProcessLib/Parameter/CurveScaledParameter.h
chleh: Chagned

chleh:
I set the container type back to std::vector.

PR #1503, opened 2016-10-28
Add the shape function order to ProcessVariable class

norihiro-w:
Until now, the shape function order is decided by the order of the given mesh. This PR enables users to specify the shape function order in a config file, which can be different from the order of the given mesh, though appropriate mesh (e.g. 2nd order mesh for use of linear or quadratic order shape functions) should be provided. 

This PR is extracted & modified after Dima's HM branch and is needed for implementing HM and "HM with LIE" processes, which use linear order for pressure and quadratic order for displacements. Major changes are
- add the shape function order to ProcessVariable class
- create local assemblers with the order provided in ProcessVariable
- support the order in BC handling
- pass the shape function order to local assembler creations in every process classes

Remarks
- Currently linear and quadratic orders are supported
- Config file needs <order> tag for each process variable, e.g. `<order>1</order>`
- I assume existing processes use the same order for all its variables. 
- The order in CalculateSurfaceFlux is fixed to 1. Correct me if it's wrong.

### DIFF
--- a/ProcessLib/ProcessVariable.cpp
+++ b/ProcessLib/ProcessVariable.cpp
@@ -37,6 +37,9 @@ ProcessVariable::ProcessVariable(
 {
     DBUG("Constructing process variable %s", _name.c_str());
 
+    if (_shapefunction_order < 1 || 2 < _shapefunction_order)
+        OGS_FATAL("The given shape function order %d is not supported", _shapefunction_order);
+
     // Boundary conditions
     //! \ogs_file_param{prj__process_variables__process_variable__boundary_conditions}
     if (auto bcs_config = config.getConfigSubtreeOptional("boundary_conditions"))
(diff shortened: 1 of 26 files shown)


norihiro-w:
I'll add the order=1 in the project files.

ProcessLib/ProcessVariable.cpp
```diff
@@ -34,6 +34,9 @@ ProcessVariable::ProcessVariable(
       _bc_builder(new BoundaryConditionBuilder())
 {
     DBUG("Constructing process variable %s", _name.c_str());
+    //! \ogs_file_param{prj__process_variables__process_variable__order}
+    auto opt_shapefunction_order = config.getConfigParameterOptional<int>("order");
+    _shapefunction_order = (opt_shapefunction_order ? opt_shapefunction_order.get() : 1);
```
endJunction: When the order is given in the prj files, then this can be removed.

wenqing: Need an assert somewhere such that this input order is not larger than the order of element of the mesh.

norihiro-w: it's not possible because Mesh/Element class doesn't know which order it is. In current implementation, LocalDataInitializer is checking if a proper element is found for the given shape function order.

PR #1504, opened 2016-10-28
[Test] Added two large size tests for liquid flow process

wenqing:
as Titled.

Note: The tests use a mesh with one million hexahedra. Each  test requires at least 8.2 GB RAM. If you run them on eve via job system, you have to specify h_vmem=9G. Otherwise, you will get  an error message about std::bad_alloc.

~~Question: Can travis CI handle such tests with high RAM usage?~~


endJunction:
The large tests running in about 1 min on my computer, which is not so long.

endJunction:
@wenqing FYI: I've removed analytical_vz data array from hex.vtu which was not used and equal 0.

wenqing:
@endJunction That's Ok to remove vz, which was prepared for my branch velocity.

PR #1506, opened 2016-11-01
Test Line3 coordinates mapping

endJunction:
The tests of coordinates mapping for quadratic order elements/shape functions are not all implemented.

This one is mainly a copy of the TestLine2 with an additional node (the origin 0,0,0) for the middle node of the line3 element.
In principle only the expected values `nat_exp_N`, `nat_exp_dNdr`, and `ir_exp_dNdx` have to be checked.


norihiro-w:
:+1: if the test is working

PR #1507, opened 2016-11-01
[PL] DirichletBC: Comment out the already computed

endJunction:
This optimization is not working correctly with time-dependent boundary
conditions and result in a "step" of the values for the second time step.

image

This is required for upcoming HM Process PR. The shown plots are from the HM test case with and without the optimization. The boundary condition is a constant parameter scaled with curve.


norihiro-w:
+1

PR #1509, opened 2016-11-02
support scaling and GMRES in Eigen linear solvers

norihiro-w:
This PR introduces
- add `OGS_USE_EIGEN_UNSUPPORTED` CMake option (default is ON) to support unsupported modules in Eigen. If this option is ON, the followings become available:
- support scaling of a linear system before solve. `<scaling>1</scaling>` should be specified under `<eigen>` tag in a config file.
- support GMRES solver type

### DIFF
--- a/MathLib/LinAlg/Eigen/EigenLinearSolver.cpp
+++ b/MathLib/LinAlg/Eigen/EigenLinearSolver.cpp
@@ -154,11 +154,15 @@ std::unique_ptr<EigenLinearSolverBase> createIterativeSolver(
         case EigenOption::SolverType::CG: {
             return createIterativeSolver<EigenCGSolver>(precon_type);
         }
-#ifdef USE_EIGEN_UNSUPPORTED
         case EigenOption::SolverType::GMRES: {
+#ifdef USE_EIGEN_UNSUPPORTED
             return createIterativeSolver<Eigen::GMRES>(precon_type);
-        }
+#else
+            OGS_FATAL(
+                "The code is not compiled with the Eigen unsupported modules. "
+                "Linear solver type GMRES is not available.");
 #endif
+        }
         default:
             OGS_FATAL("Invalid Eigen iterative linear solver type. Aborting.");
     }
@@ -186,9 +190,7 @@ EigenLinearSolver::EigenLinearSolver(
         }
         case EigenOption::SolverType::BiCGSTAB:
         case EigenOption::SolverType::CG:
-#ifdef USE_EIGEN_UNSUPPORTED
         case EigenOption::SolverType::GMRES:
-#endif
             _solver = details::createIterativeSolver(_option.solver_type,
                                                      _option.precon_type);
             return;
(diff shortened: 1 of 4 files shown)


MathLib/LinAlg/Eigen/EigenLinearSolver.cpp
```diff
@@ -12,6 +12,8 @@
 #include <logog/include/logog.hpp>
 
 #ifdef USE_EIGEN_UNSUPPORTED
+#include <iostream> // to fix compiling errors in Eigen
```
endJunction: Please add a version of Eigen where this is needed in the comment. Could help later to eliminate unnecessary includes.

endJunction:
Can you drop a pointer to the "scaling" documentation, my google does not produce adequate results.

norihiro-w:
see 
- "Note" in https://eigen.tuxfamily.org/dox-devel/classEigen_1_1SparseLU.html
- "Detailed Description" in https://eigen.tuxfamily.org/dox/unsupported/namespaceEigen.html
- References : D. Ruiz and B. Ucar, A Symmetry Preserving Algorithm for Matrix Scaling, INRIA Research report RR-7552

MathLib/LinAlg/Eigen/EigenLinearSolver.cpp
norihiro-w: The error happens in unsupported/Eigen/src/IterativeSolvers/ConstrainedConjGrad.h, which is included in unsupported/Eigen/IterativeSolvers.h. I checked the history of the file and the header file is not included since 2009. maybe I should change it to include only GMRES.

norihiro-w:
@bilke It seems Eigen package in conan does not include "unsupported" directory. Could you please add the directory to the package?

endJunction:
Thanks for the pointers.

MathLib/LinAlg/Eigen/EigenLinearSolver.cpp
```diff
@@ -209,15 +223,36 @@ void EigenLinearSolver::setOption(BaseLib::ConfigTree const& option)
             ptSolver->getConfigParameterOptional<int>("max_iteration_step")) {
         _option.max_iterations = *max_iteration_step;
     }
+#ifdef USE_EIGEN_UNSUPPORTED
+    if (auto scaling =
+            //! \ogs_file_param{linear_solver__eigen__scaling}
+            ptSolver->getConfigParameterOptional<int>("scaling")) {
```
chleh: Why don't you use bool? Please use bool.

MathLib/LinAlg/Eigen/EigenOption.cpp
```diff
@@ -29,6 +32,10 @@ EigenOption::SolverType EigenOption::getSolverType(const std::string &solver_nam
         return SolverType::BiCGSTAB;
     if (solver_name == "SparseLU")
         return SolverType::SparseLU;
+#ifdef USE_EIGEN_UNSUPPORTED
+    if (solver_name == "GMRES")
+        return SolverType::GMRES;
```
chleh: Just curious: Is there a specific reason why you don't use LIS' Gmres?

norihiro-w: no, there is no reason and I prefer LIS. However currently those who download executables from the OGS website can use only Eigen solvers and they should be able to choose GMRES for unsymmetric matrices. BiCGSTab cannot be used for all the problems as it can break down.

TomFischer:
Looks good. When Eigen package is 'fixed': :fast_forward:

MathLib/LinAlg/Eigen/EigenOption.cpp
chleh: OK, thanks.

bilke:
@norihiro-w Eigen unsupported-directory [should be in the package now](https://github.com/bilke/conan-eigen3/commit/051d9927cf96c76fdd491c7a7b2cdff0f7866ee8).

norihiro-w:
@bilke thanks! how can I restart the Jenkins job?

bilke:
@norihiro-w I restarted it (currently there is no possibility for you to restart, I will add that later on) and it ran successfully.

norihiro-w:
thanks!

Issue #1175, opened 2016-04-29
Pardiso linear solver


chleh:
I have a branch that adds support for Pardiso. However, major cleanup is required.

https://github.com/chleh/ogs6/tree/pardiso

endJunction:
Done in https://github.com/ufz/ogs/pull/1510

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/1175).

PR #1510, opened 2016-11-02
support Pardiso in Eigen

norihiro-w:
Credit goes to @chleh, because it's based on his branch (see #1175) and enables to use Pardiso solver with Eigen interface. To use it, one needs to set CMake as `-DOGS_USE_EIGEN=ON -DOGS_USE_MKL=ON -DMKL_DIR=(a path to MKL root directory)`. In a config file, you can set `PardisoLU` as a solver type.
 
@chleh Let me know if I miss something from your branch.

### DIFF
--- a/scripts/cmake/cmake/FindMKL.cmake
+++ b/scripts/cmake/cmake/FindMKL.cmake
@@ -1,0 +1,69 @@
+# Find Intel Math Karnel Library (MKL)
+#
+# Options
+# - MKL_DIR      MKL root directory
+# - MKL_OPENMP   use OpenMP threading
+#
+# Results
+# - MKL_INCLUDE_DIR
+# - MKL_LIBRARIES
+
+# Lookg for MKL root dir
+#SET(MKL_DIR $ENV{MKL_DIR} CACHE PATH "MKL root directory")
+if (NOT MKL_DIR)
+    find_path(MKL_DIR
+        include/mkl.h
+        PATHS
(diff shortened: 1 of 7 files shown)


scripts/cmake/cmake/FindMKL.cmake
```diff
@@ -0,0 +1,69 @@
+# Find Intel Math Karnel Library (MKL)
```
endJunction: Either  a source of this file, or our copyright message is missing. (Though the other cmakes don't have it too, but let's start here.)

chleh:
:+1: Looks good. Thanks for the credits, but actually it looks like you completely did it on your own (although I don't remember my code anymore).

When testing Eigen+MKL I think there were severe issues if I used Eigen's BiCGStab multithreaded with OpenMP (i.e. it crashed). Have you tried it? Did you see crashes?

norihiro-w:
@chleh No, I haven't tried the multi-threaded BiCGStab.

norihiro-w:
@bilke Jenkins job "default" is pending for a long time. what's going on here?

bilke:
The default job are the Linux-PR-Dynamic jobs on the [UFZ internal Jenkins](https://svn.ufz.de:8443/) which is currently down. I do not know why. The jobs run on envinf1 which was done the last days too.

norihiro-w:
@bilke i see. then I wait until the internal Jenkins is back

PR #1511, opened 2016-11-02
Small fixes

norihiro-w:
please see each commit

### DIFF
--- a/NumLib/DOF/DOFTableUtil.cpp
+++ b/NumLib/DOF/DOFTableUtil.cpp
@@ -20,15 +20,20 @@ double norm(GlobalVector const& x, unsigned const global_component,
              LocalToGlobalIndexMap const& dof_table, MeshLib::Mesh const& mesh, CalculateNorm calculate_norm)
 {
     // TODO that also includes ghost nodes.
-    MeshLib::MeshSubset const& ms =
-        dof_table.getMeshSubsets(global_component).getMeshSubset(mesh.getID());
     double res = 0.0;
-    for (MeshLib::Node const* node : ms.getNodes())
+    MeshLib::MeshSubsets const& mss = dof_table.getMeshSubsets(global_component);
+    for (unsigned i=0; i<mss.size(); i++)
     {
-        auto const value =
-            getNodalValue(x, mesh, dof_table, node->getID(), global_component);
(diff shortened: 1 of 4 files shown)


NumLib/DOF/DOFTableUtil.cpp
```diff
@@ -22,6 +22,8 @@ double getNodalValue(GlobalVector const& x, MeshLib::Mesh const& mesh,
 
     auto const index = dof_table.getLocalIndex(
         l, global_component_id, x.getRangeBegin(), x.getRangeEnd());
+    if (index == NumLib::MeshComponentMap::nop)
+        return 0.0;
```
endJunction: I'd think it is better to return something more harmful like NaN. But it would better to prevent such case at all with assert, which is probably more difficult.
Let's discuss this tomorrow.

norihiro-w: no, NaN should not be returned because this function is called from norm calculations in the same file. A problem I have is that the norm functions have a for loop for all mesh nodes and they call this function for nodes which are actually included in a DoF table, e.g. quadratic order nodes for pressure in HM. Because the norm calculation is sum up or taking the maximum, returning zero can still lead to correct norm.

norihiro-w:
After discussed with Dima, we found a better solution for getNodeValue(). I'm working on it

NumLib/DOF/DOFTableUtil.cpp
```diff
@@ -22,8 +47,7 @@ double getNodalValue(GlobalVector const& x, MeshLib::Mesh const& mesh,
 
     auto const index = dof_table.getLocalIndex(
         l, global_component_id, x.getRangeBegin(), x.getRangeEnd());
-    if (index == NumLib::MeshComponentMap::nop)
-        return 0.0;
+    assert (index != NumLib::MeshComponentMap::nop);
```
chleh: Is this assertion necessary only to check that the code of OGS is consistent or is it there to check if user-provided input data is correct? In the latter case the check should also be there in release mode.
Apart from that this commit looks good.

chleh:
@norihiro-w Please have a look at my last comment. It's flagged outdated, I don't know why.

PR #1512, opened 2016-11-02
[Nonlinear] print norms of delta x in ConvergenceCriterionPerComponentResidual

norihiro-w:
With this PR, I want to print norms of delta x in Newton-Raphson iterations although only residual is used for convergence checks. This helps to understand why the iteration is not converging, e.g. the criteria for residual is too strict.

### DIFF
--- a/NumLib/ODESolver/ConvergenceCriterionPerComponentResidual.cpp
+++ b/NumLib/ODESolver/ConvergenceCriterionPerComponentResidual.cpp
@@ -35,6 +35,30 @@ ConvergenceCriterionPerComponentResidual::
         OGS_FATAL("The given tolerances vector is empty.");
 }
 
+
+void ConvergenceCriterionPerComponentResidual::checkDeltaX(
+    const GlobalVector& minus_delta_x, GlobalVector const& x)
+{
+    if ((!_dof_table) || (!_mesh))
+        OGS_FATAL("D.o.f. table or mesh have not been set.");
+
+    for (unsigned global_component = 0; global_component < _abstols.size();
+         ++global_component)
+    {
+        // TODO short cut if tol <= 0.0
+        auto error_dx = norm(minus_delta_x, global_component, _norm_type,
+                             *_dof_table, *_mesh);
+        auto norm_x =
+            norm(x, global_component, _norm_type, *_dof_table, *_mesh);
+
+        INFO(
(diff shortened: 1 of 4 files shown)


NumLib/ODESolver/ConvergenceCriterionPerComponentResidual.cpp
```diff
@@ -35,6 +35,30 @@ ConvergenceCriterionPerComponentResidual::
...
+         ++global_component)
+    {
+        // TODO short cut if tol <= 0.0
+        auto error_dx = norm(minus_delta_x, global_component, _norm_type,
+                             *_dof_table, *_mesh);
+        auto norm_x =
+            norm(x, global_component, _norm_type, *_dof_table, *_mesh);
+
+        INFO(
+            "Convergence criterion, component %u: |dx|=%.4e, |x|=%.4e, "
```
TomFischer: I am unsure if the format specifiers are correct? I would expect `%f` or `%g` for the output of floating point values.

norihiro-w: yes and it's same as other outputs in convergence criteria classes. `%g` is possible but the output format can change depending on the given value. `%f` should not be used, because it is fixed digits output and we can have various magnitudes of values, e.g. from 1e-18 to 1e+6

NumLib/ODESolver/ConvergenceCriterionPerComponentResidual.h
```diff
@@ -33,11 +33,11 @@ class ConvergenceCriterionPerComponentResidual
         std::vector<double>&& relative_tolerances,
         MathLib::VecNormType norm_type);
 
-    bool hasDeltaXCheck() const override { return false; }
+    bool hasDeltaXCheck() const override { return true; }
```
chleh: I think there should be at least documentation that there will only be diagnostic output and no actual check.

norihiro-w:
the same output is added to other residual checks.

NumLib/ODESolver/ConvergenceCriterionPerComponentResidual.cpp
TomFischer: Thanks for your answer. I saw my mistake.

chleh:
Thanks :+1:

PR #1513, opened 2016-11-02
[Nonlinear] skip linear solve if residual is satisfied

norihiro-w:
This PR tries to introduce the following changes:
- allows convergence in the first iteration when the residual is checked (e.g. solution is same as previous time step, if BC is inactive, or if the system reaches steady-state)
- skip solving a linear system if the residual error is below the given tolerance
- store previous solution instead of new solution for better efficiency 

Let me know if I miss something here

### DIFF
--- a/NumLib/ODESolver/NonlinearSolver.cpp
+++ b/NumLib/ODESolver/NonlinearSolver.cpp
@@ -218,6 +218,7 @@ bool NonlinearSolver<NonlinearSolverTag::Newton>::solve(
         if (_convergence_criterion->hasResidualCheck())
             _convergence_criterion->checkResidual(res);
 
+        GlobalVector* x_old = nullptr;
         if (_convergence_criterion->hasResidualCheck() && _convergence_criterion->isSatisfied())
         {
             minus_delta_x.setZero();
@@ -236,20 +237,20 @@ bool NonlinearSolver<NonlinearSolverTag::Newton>::solve(
                 error_norms_met = false;
                 break;
             }
+
+            // copy previous solution to x_old
+            x_old =
+                &NumLib::GlobalVectorProvider::provider.getVector(
+                    x, _x_old_id);
+            LinAlg::copy(x, *x_old);
+            // update solution
+            LinAlg::axpy(x, -_alpha, minus_delta_x);
         }
 
-        // TODO could be solved in a better way
-        // cf.
-        // http://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Vec/VecWAXPY.html
-        auto& x_new =
-            NumLib::GlobalVectorProvider::provider.getVector(
-                x, _x_new_id);
-        LinAlg::axpy(x_new, -_alpha, minus_delta_x);
-
         if (postIterationCallback)
-            postIterationCallback(iteration, x_new);
+            postIterationCallback(iteration, x);
 
-        switch(sys.postIteration(x_new))
+        switch(sys.postIteration(x))
         {
             case IterationResult::SUCCESS:
                 break;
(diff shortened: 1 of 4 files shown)


NumLib/ODESolver/NonlinearSolver.cpp
```diff
@@ -215,63 +215,65 @@ bool NonlinearSolver<NonlinearSolverTag::Newton>::solve(
         sys.applyKnownSolutionsNewton(J, res, minus_delta_x);
         INFO("[time] Applying Dirichlet BCs took %g s.", time_dirichlet.elapsed());
 
-        if (!sys.isLinear() && _convergence_criterion->hasResidualCheck())
```
chleh: Why do you have to check if the system is not non-linear?

NumLib/ODESolver/NonlinearSolver.cpp
```diff
@@ -215,63 +215,65 @@ bool NonlinearSolver<NonlinearSolverTag::Newton>::solve(
...
             _convergence_criterion->checkResidual(res);
 
-        BaseLib::RunTime time_linear_solver;
-        time_linear_solver.start();
-        bool iteration_succeeded = _linear_solver.solve(J, res, minus_delta_x);
-        INFO("[time] Linear solver took %g s.", time_linear_solver.elapsed());
-
-        if (!iteration_succeeded)
+        GlobalVector* x_old = nullptr;
+        if (_convergence_criterion->hasResidualCheck() && _convergence_criterion->isSatisfied())
```
chleh: Why does this branch only apply if you check the residual?

NumLib/ODESolver/NonlinearSolver.cpp
norihiro-w: without checking the residual, how can the iteration stop? besides, if the residual check is satisfied, there is no need to solve a linear system even if it's a linear problem. checking residual is cheaper than linear solve.

NumLib/ODESolver/NonlinearSolver.cpp
norihiro-w: it's because, if residual is enough small with current solution, we don't need to solve a linear system. If you use a convergence criteria based on delta x, convergence cannot be checked until you get next solution, i.e. solve a linear system.

norihiro-w:
@chleh The behavior is changed only when you use residual for a convergence criteria, except for c373045. If delta x is used, the behavior is still same as before. All the tests are still working.

NumLib/ODESolver/NonlinearSolver.cpp
chleh: If the Process is not non-linear the nonlinear solver will automatically stop after one iteration. But it's true, without the residual check the shortcut that you introduced wouldn't be possible.

NumLib/ODESolver/NonlinearSolver.cpp
chleh: Actually it would be nice if one could just write `if (_convergence_criterion->isSatisfied())`. But there's an interplay between the convergence criterion classes and the nonlinear solver. So this is probably hard to obtain.

chleh:
But anyway, now the behaviour of the Newton and the Picard solver is different in some details. I think the logic of the `solve()` method is somewhat complicated. Therefore the behaviour should be entirely the same (i.e. the code should be as equal as possible), because otherwise it would be more difficult, e.g., to introduce new convergence criteria (e.g. one checking both delta x and the residual).
Could you please introduce the same changes into the Picard solver, s.t. the code matches more closely (before this PR the two methods were ~90% the same, now you reordered the Newton)?

norihiro-w:
I close this PR at the moment because I need to some work on this but its priority is not high.

PR #1514, opened 2016-11-02
[CLI] add --unbuffered-std-out option to deactivate buffer for cout

norihiro-w:
as titled. This is helpful for those who saves/forwards stadard output to a file and wants to trace real-time output from ogs.


chleh:
Looks good. But wouldn't it be better to automatically flush stdout once after every iteration/timestep?

norihiro-w:
@chleh automatic flushing should be avoided in some simulations particularly on cluster computers, because it can cause high load on the system File I/O.

TomFischer:
> automatic flushing should be avoided in some simulations particularly on cluster computers, because it can cause high load on the system File I/O. 

Maybe this can be added as documentation into the command line option?

PR #1515, opened 2016-11-02
[Mesh] fix incorrect use of getNumberOfBaseNodes()

norihiro-w:
I've just realized some Mesh routines ignore nonlinear nodes, e.g. in constructing a list of connected elements to a node. This PR tries to fix them by using `getNumberOfNodes()` instead of `getNumberOfBaseNodes()`, except for `Mesh::setNodesConnectedByEdges()` where special care is required for nonlinear case. I haven't looked into classes used only in utility.


norihiro-w:
the bug also causes incorrect sparsity patterns for quadratic order variables

PR #1516, opened 2016-11-03
[Mesh] fix the nonlinear node ID check

norihiro-w:
fix #1495. The check doesn't work if the number of base nodes is provided to a constructor, which is the case in VtuInterface.

### DIFF
--- a/MeshLib/Mesh.cpp
+++ b/MeshLib/Mesh.cpp
@@ -42,7 +42,7 @@ Mesh::Mesh(const std::string &name,
     assert(n_base_nodes <= nodes.size());
     this->resetNodeIDs();
     this->resetElementIDs();
-    if (isNonlinear())
+    if ((n_base_nodes==0 && hasNonlinearElement()) || isNonlinear())
         this->checkNonlinearNodeIDs();
     this->setDimension();
     this->setElementsConnectedToNodes();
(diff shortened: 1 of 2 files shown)


MeshLib/Mesh.cpp
```diff
@@ -278,6 +278,17 @@ void Mesh::checkNonlinearNodeIDs() const
...
+bool Mesh::hasNonlinearElement() const
+{
+    for (auto* const e : _elements)
+    {
+        if (e->getNumberOfNodes() == e->getNumberOfBaseNodes())
+            continue;
+        return true;
+    }
+    return false;
+}
```
endJunction: My personal preference would be:
```c++
return std::any_of(std::begin(_elements), std::end(_elements),
    [](Element const* const e) {
        return e->getNumberOfNodes() != e->getNumberOfBaseNodes();
    };
```

norihiro-w:
I changed it to use ERR instead of OGS_FATAL in the check. Otherwise we cannot create and correct a Mesh object.

endJunction:
:fast_forward: when tests pass.

MeshLib/Mesh.cpp
```diff
@@ -259,25 +268,30 @@ void Mesh::checkNonlinearNodeIDs() const
...
-                    "than the number of base nodes %d. Please renumber node IDs in the mesh.",
-                    e->getNodeIndex(i), getNumberOfBaseNodes());
+            if (e->getNodeIndex(i) >= getNumberOfBaseNodes())
+                continue;
+
+            DBUG(
+                "Node %d is a non-linear node, but the ID is smaller "
+                "than the number of base nodes %d.",
+                e->getNodeIndex(i), getNumberOfBaseNodes());
+            ERR("Found a nonlinear node whose ID is smaller than base node IDs."
```
TomFischer: Could you make the error message as specific as the debug message, i.e., add the valuable node id and the number of base nodes? Maybe its possible to merge both messages?

PR #1520, opened 2016-11-03
Reduce number of instantiated local asms

endJunction:
As of now the local assemblers were instantiated for example for Hex8 and GlobalDim=1, 2, 3, but the dimensions 1 and 2 do not make any sense. These are excluded from the instantiation.
For Line3 still all GlobalDims are used.

See for example output of `nm -C bin/ogs | grep Hex8 | grep GroundwaterFlow::LocalAssemblerData | grep assemble`

There exist a computer s.t. the compilation took before and after (in Release, with fixed size matrices, clang-4.0 and LTO):
CCACHE_DISABLE=1 make -j5  1135.72s user 31.24s system 400% cpu 4:51.06 total
CCACHE_DISABLE=1 make -j5  1047.01s user 28.57s system 413% cpu 4:20.00 total
so around 8% less cpu time.

### DIFF
--- a/ProcessLib/LIE/SmallDeformation/LocalAssembler/LocalDataInitializer.h
+++ b/ProcessLib/LIE/SmallDeformation/LocalAssembler/LocalDataInitializer.h
@@ -14,6 +14,7 @@
 #include <memory>
 #include <typeindex>
 #include <typeinfo>
+#include <type_traits>
 #include <unordered_map>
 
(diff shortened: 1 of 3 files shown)


norihiro-w:
:+1:

endJunction:
This unfortunately triggers an internal compiler error in MSVC 18.0.40629.0.

PR #1522, opened 2016-11-04
[CMake] set the default OGS_LOG_LEVEL to debug also in release build

norihiro-w:
This PR suggests  
- to change the default compilation logog level in release builds from info to debug
- to change the default runtime logog level in release builds from all to info 

IMO the default setting should be configured for normal users who download the OGS executable from the website. With current setting, users cannot see debug messages even if they set the runtime logog level with `-l debug`. This also helps us/developers to support users because we can ask them to run ogs with debug level to get detailed logs and we don't need to run it by ourselves (of course, sometimes it's not enough).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -86,6 +86,15 @@ option(EIGEN_NO_DEBUG "Disables Eigen's assertions" OFF)
 
 # Logging
 option(OGS_DISABLE_LOGGING "Disables all logog messages." OFF)
+set(OGS_LOG_LEVEL "LOGOG_LEVEL_DEBUG" CACHE STRING "Set logging level included in compilation.")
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -166,6 +175,13 @@ if(OGS_BUILD_TESTS)
     set(Data_BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR}/Tests/Data CACHE INTERNAL "")
 endif()
 
+# Logging level
+if(OGS_DISABLE_LOGGING)
```
chleh: Is this option still needed? It feels duplicate now. Maybe remove it?

norihiro-w: I can remove it if nobody needs it.

bilke: Remove it

PR #1524, opened 2016-11-07
restructure ProcessLib/SmallDeformationWithLIE

norihiro-w:
This PR creates `ProcessLib/LIE` directory and move `ProcessLib/SmallDeformationWithLIE` under the new directory. Some common functions are extracted from the SD dir and moved directory under the `LIE` dir so that HydroMechanics process (which I'm implementing under `LIE/HydroMechanics`) can share the functions. 

I set  "please review" label but don't think it's worth to spend time on reviewing this PR, because there is no logic changes and the changes are moving files, renaming namespaces and  macros for the include guards. If the tests are ok, this can be merged.


chleh:
:fast_forward:

PR #1528, opened 2016-11-08
[Jenkins/Benchmarks] save the standard output into a file

norihiro-w:
I would like to see log messages for failed benchmarks without manually running them again.


chleh:
Doesn't ctest already automatically save stdout/stderr to `$BUILDDIR/Testing/Temporary/LastTest.log`?

endJunction:
@norihiro-w  Could you please comment Christoph's question?

norihiro-w:
Christoph's comment is correct, though my preference is saving the output into a separate file for each benchmark, because it's easier to find outputs from a particular benchmark and do further works, e.g. compare the logs between different versions. 

But I don't mind to keep it in my branch and close this PR if it's only me who would like to have the separate output.

norihiro-w:
It seems one cannot use `tail -f` with `$BUILDDIR/Testing/Temporary/LastTest.log`. At least on my environment, the output will be saved into the file after the job is completed. I don't know how to change the buffered behavior in ctest. 

Alternative solution is `ctest --verbose &> test.log` which includes additional messages from ctest

chleh:
Thank you for the explanation. If you find the introduced feature handy, I won't block it. Especially comparing different versions will be more convenient after this PR, indeed.
My only concern was that we shouldn't reimplement features that are already provided by the tools we use. But as said, this PR adds some convenience. So :+1:

PR #1529, opened 2016-11-09
[Alg] Added a  class for piece wise linear curve

wenqing:
The class is derived form PiecewiseLinearInterpolation, and it has members
1.  getVariable(y) to get a   variable (x) by value (y). 
2.  added a member to check the monotonicity of the pair data for getVariable.

A function is also added to create a curve.

The  class is for  curve type capillary pressure saturation models and other curve type material models.


MathLib/InterpolationAlgorithms/PiecewiseLinearInterpolation.cpp
```diff
@@ -14,10 +14,12 @@
 #include <cmath>
 #include <limits>
 #include <utility>
+#include <algorithm>
 
-#include "BaseLib/quicksort.h"
 #include "PiecewiseLinearInterpolation.h"
```
norihiro-w: please move this to the top of the includes

MathLib/InterpolationAlgorithms/PiecewiseLinearInterpolation.cpp
```diff
@@ -14,10 +14,12 @@
 #include <cmath>
 #include <limits>
 #include <utility>
+#include <algorithm>
```
norihiro-w: please move this before `<cmath>`.  The includes should be sorted by the alphabetical order.

MathLib/InterpolationAlgorithms/PiecewiseLinearInterpolation.h
```diff
@@ -60,7 +60,26 @@ class PiecewiseLinearInterpolation final
...
+     * x_j\f$ and
+     * \f$x_{\max} = \max_{1 \le j \le n} x_j\f$. Points outside of this
+     * interval are
+     * set to f(x_{\min}) or f(x_{\max}).
+     * @return The interpolated value, \f$x\f$.
+     *
+     * \attention Check the monotonicity of the data beforehand if this function
+     *            is used.
+     */
+    double getInverseValue(double const value) const;
```
norihiro-w: - instead of `variable`, `supporting point` is used in other documentation in this class. please make it consistent. 
- I guess this function returns `variable` (or `supporting point`), not the value. Current function name is confusing and should be renamed, e.g. `getVariable()` or `getPoint()`? 
- can you add a description why the monotonicity is required? and also what will happen if the function is called for non-monotonic case?

MathLib/InterpolationAlgorithms/PiecewiseLinearInterpolation.h
```diff
@@ -78,9 +97,16 @@ class PiecewiseLinearInterpolation final
     double getSupportMax() const;
     double getSupportMin() const;
 
+    bool isMonotonic() const;
```
norihiro-w: need documentation

MathLib/InterpolationAlgorithms/PiecewiseLinearInterpolation.cpp
```diff
@@ -35,9 +37,27 @@ PiecewiseLinearInterpolation::PiecewiseLinearInterpolation(
...
 }
 
-double PiecewiseLinearInterpolation::getValue(double pnt_to_interpolate) const
+double PiecewiseLinearInterpolation::interpolate(
+    std::vector<double> const& supp_pnts,
+    std::vector<double> const& values_at_supp_pnts,
+    std::size_t const interval_idx,
+    double const pnt_to_interpolate) const
+{
+    assert(std::fabs(supp_pnts[interval_idx + 1] - supp_pnts[interval_idx]) >
```
norihiro-w: this depends on users input. i think this should be checked once in the constructor with OGS_FATAL.

MathLib/InterpolationAlgorithms/PiecewiseLinearInterpolation.cpp
```diff
@@ -48,17 +68,59 @@ double PiecewiseLinearInterpolation::getValue(double pnt_to_interpolate) const
...
-                   (_supp_pnts[interval_idx + 1] - _supp_pnts[interval_idx]));
+    return interpolate(_supp_pnts, _values_at_supp_pnts, interval_idx,
+                       pnt_to_interpolate);
+}
 
-    return m * (pnt_to_interpolate - _supp_pnts[interval_idx]) +
-           _values_at_supp_pnts[interval_idx];
+double PiecewiseLinearInterpolation::getInverseValue(double const value) const
+{
+    std::size_t interval_idx = 0;
```
norihiro-w: why is the monotonic not checked?

Tests/MathLib/TestPiecewiseLinearInterpolation.cpp
```diff
@@ -155,3 +155,30 @@ TEST(MathLibInterpolationAlgorithms, PiecewiseLinearInterpolationDerivative)
     ASSERT_NEAR(0, interpolation.getDerivative(1001),
                 std::numeric_limits<double>::epsilon());
 }
+
+TEST(MathLibInterpolationAlgorithms, PiecewiseLinearInterpolationGetInverseValue)
```
norihiro-w: suggest having a test for monotonically decrease case

MathLib/InterpolationAlgorithms/PiecewiseLinearInterpolation.cpp
wenqing: It is checked once after an instance of this class is created.

wenqing:
@TomFischer I also think the name of the class does not fit my purpose. According to your request, I want to create a class CurveFunction in MathLib.

MathLib/InterpolationAlgorithms/PiecewiseLinearInterpolation.cpp
wenqing: Changed.

MathLib/InterpolationAlgorithms/PiecewiseLinearInterpolation.cpp
wenqing: This is to move it to the top.

MathLib/InterpolationAlgorithms/PiecewiseLinearInterpolation.h
wenqing: Changed the name to getVariable, which is now in class PiecewiseLinearCurve.

MathLib/InterpolationAlgorithms/PiecewiseLinearInterpolation.h
wenqing: Added.

MathLib/InterpolationAlgorithms/PiecewiseLinearInterpolation.cpp
wenqing: Added it to  PiecewiseLinearInterpolation::PiecewiseLinearInterpolation.

MathLib/InterpolationAlgorithms/PiecewiseLinearInterpolation.cpp
wenqing: On demand, the monotonic is only checked once immediately after an instance the class is created.

Tests/MathLib/TestPiecewiseLinearInterpolation.cpp
wenqing: Added.

wenqing:
@TomFischer @norihiro-w  According to Tom's request, I derived a class, PiecewiseLinearCurve, from PiecewiseLinearInterpolation for all the implementation in this PR. For this purpose, there is minor changes in PiecewiseLinearInterpolation.

The new class PiecewiseLinearCurve computes a curve y(x)
to get y(x), x(y), or dy/dx.

The monotonicity is on demand checked if computing x(y) will be conducted. 

A parsing function for creating an unique_ptr type of this class is created too. The XML syntax of the input looks like
```
       <curve>
          <type>PiecewiseLinear</type>
          <point><data> 0.2 20.</data></point>
          <point><data> 0.4 10.</data></point>
          <point><data> 0.5 5.</data></point>
          <point><data> 0.6 3.</data></point>
          <point><data> 0.7 2.</data></point>
        </curve>
```
 The reason why I need this sort of parsing other than that is in PorjectData is given in PR #1531.

TomFischer:
Sorry @wenqing : I did not have time to review. Will do this tomorrow.

MathLib/Curve/PiecewiseLinearCurve.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+#include <algorithm>
+#include <cmath>
+#include <limits>
+
+namespace MathLib
+{
+bool PiecewiseLinearCurve::isMonotonic() const
+{
+    const double tangent0 = getDerivative(_supp_pnts[0]);
```
TomFischer: Should this be named `slope` or `gradient`?

Tests/MathLib/TestPiecewiseLinearCurve.cpp
```diff
@@ -0,0 +1,125 @@
...
+{
+    const char xml[] =
+        "<curve>"
+        "   <type>PiecewiseLinear</type>"
+        "   <point><data> 0.2 20.</data></point> "
+        "   <point><data> 0.4 10.</data></point> "
+        "   <point><data> 0.5 5.</data></point> "
+        "   <point><data> 0.6 3.</data></point> "
+        "   <point><data> 0.7 2.</data></point> "
+        "</curve>";
```
TomFischer: This is not the curve format that is already implemented. Do you plan to change this?

wenqing: After discussed with Diam, we decided to change it to the one that already existed.

MathLib/Curve/PiecewiseLinearCurve.cpp
wenqing: Will changed tangent0  to gradient0

wenqing:
The XML syntax is changed to the existing one. However, I think, the column input is a little human readable.

MathLib/Curve/PiecewiseLinearCurve.h
```diff
@@ -0,0 +1,44 @@
...
+ */
+
+#ifndef OGS_PIECE_WISE_LINEAR_CURVE_H
+#define OGS_PIECE_WISE_LINEAR_CURVE_H
+
+#include "MathLib/InterpolationAlgorithms/PiecewiseLinearInterpolation.h"
+
+namespace MathLib
+{
+class PiecewiseLinearInterpolation;
```
norihiro-w: i think you can remove this

MathLib/Curve/PiecewiseLinearCurve.cpp
```diff
@@ -0,0 +1,87 @@
...
+        return std::none_of(_supp_pnts.begin(), _supp_pnts.end(),
+                            [&](const double p) {
+                                return this->getDerivative(p) * gradient0 <= 0.;
+                            });
+    }
+}
+
+double PiecewiseLinearCurve::getVariable(const double y) const
+{
+    std::size_t interval_idx = 0;
```
norihiro-w: I suggest calling `isMonotonic() ` at the beginning of this function, to help a user. of this class to detect mistakes. of course, it's enough to call it once and store the result in a member variable.

MathLib/Curve/PiecewiseLinearCurve.cpp
```diff
@@ -0,0 +1,87 @@
...
+            // search interval in the reverse direction for the point inside
+            auto const& it(std::lower_bound(_values_at_supp_pnts.rbegin(),
+                                            _values_at_supp_pnts.rend(), y));
+            interval_idx = _values_at_supp_pnts.size() -
+                           (std::distance(_values_at_supp_pnts.rbegin(), it)) -
+                           1;
+        }
+    }
+
+    // compute linear interpolation polynom: y = m * (x - support[i]) + value[i]
```
TomFischer: The comment has to be changed, too.

MathLib/InterpolationAlgorithms/PiecewiseLinearInterpolation.cpp
```diff
@@ -33,6 +35,17 @@ PiecewiseLinearInterpolation::PiecewiseLinearInterpolation(
...
             _values_at_supp_pnts);
     }
+
+    for (std::size_t i=0; i<_supp_pnts.size()-1; i++)
+    {
+        if (std::fabs(_supp_pnts[i+1] - _supp_pnts[i])
+                                   < std::numeric_limits<double>::min())
+        {
+            OGS_FATAL("Variable %d and variable %d are the same. "
+                      "Piece wise linear interpolation is not possible\n",
```
TomFischer: Spelling: Piecewise.

MathLib/Curve/PiecewiseLinearCurve.h
wenqing: Removed.

MathLib/Curve/PiecewiseLinearCurve.cpp
wenqing: It is expensive to call this function here.   isMonotonic() can be called when after a monotonic curve is created

MathLib/InterpolationAlgorithms/PiecewiseLinearInterpolation.cpp
```diff
@@ -33,6 +35,17 @@ PiecewiseLinearInterpolation::PiecewiseLinearInterpolation(
             _supp_pnts, static_cast<std::size_t>(0), _supp_pnts.size(),
             _values_at_supp_pnts);
     }
+
+    for (std::size_t i=0; i<_supp_pnts.size()-1; i++)
+    {
+        if (std::fabs(_supp_pnts[i+1] - _supp_pnts[i])
+                                   < std::numeric_limits<double>::min())
```
wenqing: Changed.

MathLib/Curve/PiecewiseLinearCurve.h
```diff
@@ -0,0 +1,44 @@
...
+class PiecewiseLinearCurve final : public PiecewiseLinearInterpolation
+{
+public:
+    /**
+     * @ x x coordinates of curve points
+     * @ y y coordinates of curve points
+     */
+    PiecewiseLinearCurve(std::vector<double>&& x, std::vector<double>&& y)
+        : PiecewiseLinearInterpolation(std::move(x), std::move(y), false)
+    {
```
wenqing: Because the monotonicity is not required for all curves.

MathLib/Curve/PiecewiseLinearCurve.h
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    ~PiecewiseLinearCurve() = default;
+
+    /// Check monotonicity of this curve.
+    bool isMonotonic() const;
+
+    /// Get variable by a given value \c y.
+    //  monotonicity must be check for the curve for using this function
+    double getVariable(const double y) const;
```
wenqing: For some cases, e.g, capillary pressure curve, both normal and revere  interpolation are needed for a curve. Therefore  ''exchange supporting points and supporting values within the constructor of the existing PiecewiseInterpolation class?'' is not feasible.

MathLib/InterpolationAlgorithms/PiecewiseLinearInterpolation.cpp
wenqing: Corrected.

MathLib/InterpolationAlgorithms/PiecewiseLinearInterpolation.cpp
```diff
@@ -33,6 +35,17 @@ PiecewiseLinearInterpolation::PiecewiseLinearInterpolation(
             _supp_pnts, static_cast<std::size_t>(0), _supp_pnts.size(),
             _values_at_supp_pnts);
     }
+
+    for (std::size_t i=0; i<_supp_pnts.size()-1; i++)
```
wenqing: Thanks. Changed.

MathLib/Curve/PiecewiseLinearCurve.h
TomFischer: Is the solution of `getVariable()` unique when the piecewise linear function is not monotonic? How is it ensured to find the correct interval in non-monotonic case in `getVariable()`?

MathLib/Curve/PiecewiseLinearCurve.cpp
norihiro-w: yes it's expensive to call it every time. but you can compute it only once, e.g. in the constructor, and store the result into a member variable like `bool _is_monotonic`. then in this function, you can check at the beginning without cost as 

```
if (!_is_monotonic)
    OGS_FATAL("The given curve is not monotonic. getVariable() cannot be used.");
```

MathLib/Curve/PiecewiseLinearCurve.h
```diff
@@ -0,0 +1,49 @@
...
+     *    \f$y(x_{i+1})>y(x_{i})\f$ if \f$x_{i+1}>x_{i}\f$
+     *  or for any \f$i\f$
+     *    \f$y(x_{i+1})<y(x_{i})\f$ if \f$x_{i+1}>x_{i}\f$
+     *  \return True if the curve exhibits strong monotonic.
+     */
+    bool isStrongMonotonic() const;
+
+    /// Get variable by a given value \c y.
+    //  monotonicity must be check for the curve for using this function
+    double getVariable(const double y) const;
```
norihiro-w: please document what will happen in this function if the curve is not monotonic.

MathLib/Curve/PiecewiseLinearCurve.cpp
wenqing: Thank. I revised accordingly. However, I got another idea: create two classes as
PiecewiseLinearCurve
PiecewiseLinearMonotonicCurve

With that, the monotonicity can only assert in ctor. To do this, I have to wait until  PR #1561 is merged.

@TomFischer  How do you think about this idea?

MathLib/Curve/PiecewiseLinearCurve.h
wenqing: see my last comment.  How do you think about the idea to create two classes
PiecewiseLinearCurve
PiecewiseLinearMonotonicCurve

With that
getDerivative will be moved to PiecewiseLinearCurve, and PiecewiseLinearMonotonicCurve is derived from PiecewiseLinearCurve.

MathLib/Curve/PiecewiseLinearCurve.h
```diff
@@ -24,13 +24,22 @@ class PiecewiseLinearCurve final : public PiecewiseLinearInterpolation
      * @ x x coordinates of curve points
      * @ y y coordinates of curve points
      */
-    PiecewiseLinearCurve(std::vector<double>&& x, std::vector<double>&& y)
-        : PiecewiseLinearInterpolation(std::move(x), std::move(y), false)
+    PiecewiseLinearCurve(std::vector<double>&& x, std::vector<double>&& y,
+                         const bool check_monotonicity)
+        : PiecewiseLinearInterpolation(std::move(x), std::move(y), false),
+          _is_monotonic(check_monotonicity ? isStrongMonotonic() : false)
```
endJunction: could also be
```
_is_monotonic(check_monotonicity && isStrongMonotonic())
```

MathLib/Curve/PiecewiseLinearCurve.h
```diff
@@ -0,0 +1,54 @@
...
+{
+class PiecewiseLinearCurve final : public PiecewiseLinearInterpolation
+{
+public:
+    /**
+     * @ x x coordinates of curve points
+     * @ y y coordinates of curve points
+     */
+    PiecewiseLinearCurve(std::vector<double>&& x, std::vector<double>&& y,
+                         const bool check_monotonicity)
```
TomFischer: Please remove `check_monocity` from the argument list. You have to check the monotonicity in order to be able to use `getVariable()`.

wenqing:
Class name is changed to PiecewiseLinearMonotonicCurve. While, createPiecewiseLinearCurve is changed as a general parser for curves.

endJunction:
@wenqing  There are 3 new doxygen warnings after the merge.
Would you please fix them:
https://jenkins.opengeosys.org/job/OGS-6/job/ufz/job/master/82/warnings15Result/new/package.565819657/

PR #1517, opened 2016-11-03
Capillary pressure models

wenqing:
This PR presents:

- Base class for capillary pressure model
- class for van Genuchten capillaryPressure model
- class for BrookCoreyCapillaryPressure model
- Creator for capillary pressure models
- unit tests for these classes

PR #1531, opened 2016-11-09
Relative permeability model

wenqing:
This PR presents:
- base class for relative permeability models
- class for WettingPhaseVanGenuchten model
- class for NonWettingPhaseVanGenuchten model
- class for WettingPhaseBrookCoreyOilGas model
- class for NonWettingPhaseBrookCoreyOilGas model
- class for curve type model
- unit tests for these classes.

### DIFF
--- a/MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/CreateRelativePermeabilityModel.cpp
+++ b/MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/CreateRelativePermeabilityModel.cpp
@@ -142,8 +142,8 @@ std::unique_ptr<RelativePermeability> createRelativePermeabilityModel(
     else
     {
         OGS_FATAL(
-            "The capillary pressure mode %s is unavailable.\n"
-            "The available types are:"
+            "The capillary pressure model %s is unavailable.\n"
+            "The available models are:"
             "\n\tWettingPhaseVanGenuchten,"
             "\n\tNonWettingPhaseVanGenuchten,"
             "\n\tWettingPhaseBrookCoreyOilGas,"
--- a/MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/CreateRelativePermeabilityModel.cpp
+++ b/MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/CreateRelativePermeabilityModel.cpp
@@ -34,97 +34,167 @@ namespace MaterialLib
 {
 namespace PorousMedium
 {
-enum class BrookCoreyOrVanGenuchtenModelType
+/**
+    \param config ConfigTree object which contains the input data
+                  including <type>WettingPhaseVanGenuchten</type>
+                  and it has a tag of <relative_permeability>
+*/
+std::unique_ptr<RelativePermeability> createWettingPhaseVanGenuchten(
+    BaseLib::ConfigTree const& config)
 {
-    WettingPhaseVanGenuchten,
-    NonWettingPhaseVanGenuchten,
-    WettingPhaseBrookCoreyOilGas,
-    NonWettingPhaseBrookCoreyOilGas
-};
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__type}
+    config.checkConfigParameter("type", "WettingPhaseVanGenuchten");
+
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__WettingPhaseVanGenuchten__sr}
+    const double Sr = config.getConfigParameter<double>("sr");
+
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__WettingPhaseVanGenuchten__smax}
+    const double Smax = config.getConfigParameter<double>("smax");
+
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__WettingPhaseVanGenuchten__m}
+    const double m = config.getConfigParameter<double>("m");
+    if (m > 1.0)  // m <= 1
+    {
+        OGS_FATAL(
+            "The exponent parameter of WettingPhaseVanGenuchten relative\n"
+            " permeability model, m, must be in an interval of [0, 1]");
+    }
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__WettingPhaseVanGenuchten__m}
+    const double krel_max = config.getConfigParameter<double>("krel_min");
+
+    return std::unique_ptr<RelativePermeability>(
+        new WettingPhaseVanGenuchten(Sr, Smax, m, krel_max));
+}
 
 /**
- *     \param config ConfigTree object which contains the input data
- *                   including <type>WettingPhaseBrookCoreyOilGas</type>
- *                          or <type>NonWettingPhaseBrookCoreyOilGas</type>
- *                          or <type>WettingPhaseVanGenuchten</type>
- *                          or <type>NonWettingPhaseVanGenuchten</type>
- *                   and it has a tag of <relative_permeability>
+    \param config ConfigTree object which contains the input data
+                  including <type>NonWettingPhaseVanGenuchten</type>
+                  and it has a tag of <relative_permeability>
 */
-static std::unique_ptr<RelativePermeability> createBrookCoreyOrVanGenuchten(
-    BaseLib::ConfigTree const& config,
-    const BrookCoreyOrVanGenuchtenModelType mode_type)
+std::unique_ptr<RelativePermeability> createNonWettingPhaseVanGenuchten(
+    BaseLib::ConfigTree const& config)
 {
-    std::array<double, 4> parameters = {
-        {//! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__type__sr}
-         config.getConfigParameter<double>("sr"),
-         //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__type__smax}
-         config.getConfigParameter<double>("smax"),
-         //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__type__m}
-         config.getConfigParameter<double>("m"),
-         //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__type__krel_min}
-         config.getConfigParameter<double>("krel_min")}};
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__type}
+    config.checkConfigParameter("type", "NonWettingPhaseVanGenuchten");
 
-    switch (mode_type)
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__NonWettingPhaseVanGenuchten__sr}
+    const double Sr = config.getConfigParameter<double>("sr");
+
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__NonWettingPhaseVanGenuchten__smax}
+    const double Smax = config.getConfigParameter<double>("smax");
+
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__NonWettingPhaseVanGenuchten__m}
+    const double m = config.getConfigParameter<double>("m");
+    if (m > 1.0)  // m <= 1
     {
-        case BrookCoreyOrVanGenuchtenModelType::WettingPhaseVanGenuchten:
-            assert(parameters[2] <= 1.0);  // m <= 1
-            return std::unique_ptr<RelativePermeability>(
-                new WettingPhaseVanGenuchten(parameters));
-        case BrookCoreyOrVanGenuchtenModelType::NonWettingPhaseVanGenuchten:
-            assert(parameters[2] <= 1.0);  // m <= 1
-            return std::unique_ptr<RelativePermeability>(
-                new NonWettingPhaseVanGenuchten(parameters));
-        case BrookCoreyOrVanGenuchtenModelType::WettingPhaseBrookCoreyOilGas:
-            assert(parameters[2] >= 1.0);  // m >= 1
-            return std::unique_ptr<RelativePermeability>(
-                new WettingPhaseBrookCoreyOilGas(parameters));
-        case BrookCoreyOrVanGenuchtenModelType::NonWettingPhaseBrookCoreyOilGas:
-            assert(parameters[2] >= 1.0);  // m >= 1
-            return std::unique_ptr<RelativePermeability>(
-                new NonWettingPhaseBrookCoreyOilGas(parameters));
+        OGS_FATAL(
+            "The exponent parameter of NonWettingPhaseVanGenuchten relative\n"
+            " permeability model, m, must be in an interval of [0, 1]");
     }
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__NonWettingPhaseVanGenuchten__m}
+    const double krel_max = config.getConfigParameter<double>("krel_min");
 
-    return nullptr;
+    return std::unique_ptr<RelativePermeability>(
+        new NonWettingPhaseVanGenuchten(Sr, Smax, m, krel_max));
+}
+
+/**
+    \param config ConfigTree object which contains the input data
+                  including <type>WettingPhaseBrookCoreyOilGas</type>
+                  and it has a tag of <relative_permeability>
+*/
+std::unique_ptr<RelativePermeability> createWettingPhaseBrookCoreyOilGas(
+    BaseLib::ConfigTree const& config)
+{
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__type}
+    config.checkConfigParameter("type", "WettingPhaseBrookCoreyOilGas");
+
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__WettingPhaseBrookCoreyOilGas__sr}
+    const double Sr = config.getConfigParameter<double>("sr");
+
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__WettingPhaseBrookCoreyOilGas__smax}
+    const double Smax = config.getConfigParameter<double>("smax");
+
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__WettingPhaseBrookCoreyOilGas__m}
+    const double m = config.getConfigParameter<double>("m");
+    if (m < 1.0)  // m >= 1
+    {
+        OGS_FATAL(
+            "The exponent parameter of WettingPhaseBrookCoreyOilGas\n"
+            "relative permeability model, m, must not be smaller than 1");
+    }
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__WettingPhaseBrookCoreyOilGas__m}
+    const double krel_max = config.getConfigParameter<double>("krel_min");
+
+    return std::unique_ptr<RelativePermeability>(
+        new WettingPhaseBrookCoreyOilGas(Sr, Smax, m, krel_max));
+}
+
+/**
+    \param config ConfigTree object which contains the input data
+                  including <type>NonWettingPhaseBrookCoreyOilGas</type>
+                  and it has a tag of <relative_permeability>
+*/
+std::unique_ptr<RelativePermeability> createNonWettingPhaseBrookCoreyOilGas(
+    BaseLib::ConfigTree const& config)
+{
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__type}
+    config.checkConfigParameter("type", "NonWettingPhaseBrookCoreyOilGas");
+
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__NonWettingPhaseBrookCoreyOilGas__sr}
+    const double Sr = config.getConfigParameter<double>("sr");
+
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__NonWettingPhaseBrookCoreyOilGas__smax}
+    const double Smax = config.getConfigParameter<double>("smax");
+
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__NonWettingPhaseBrookCoreyOilGas__m}
+    const double m = config.getConfigParameter<double>("m");
+    if (m < 1.0)  // m >= 1
+    {
+        OGS_FATAL(
+            "The exponent parameter of NonWettingPhaseBrookCoreyOilGas\n"
+            "relative permeability model, m, must not be smaller than 1");
+    }
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__NonWettingPhaseBrookCoreyOilGas__m}
+    const double krel_max = config.getConfigParameter<double>("krel_min");
+
+    return std::unique_ptr<RelativePermeability>(
+        new NonWettingPhaseBrookCoreyOilGas(Sr, Smax, m, krel_max));
 }
 
 std::unique_ptr<RelativePermeability> createRelativePermeabilityModel(
     BaseLib::ConfigTree const& config)
 {
     //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__type}
-    auto const type = config.getConfigParameter<std::string>("type");
+    auto const type = config.peekConfigParameter<std::string>("type");
 
     if (type == "WettingPhaseVanGenuchten")
     {
-        return createBrookCoreyOrVanGenuchten(
-            config,
-            BrookCoreyOrVanGenuchtenModelType::WettingPhaseVanGenuchten);
+        return createWettingPhaseVanGenuchten(config);
     }
     else if (type == "NonWettingPhaseVanGenuchten")
     {
-        return createBrookCoreyOrVanGenuchten(
-            config,
-            BrookCoreyOrVanGenuchtenModelType::NonWettingPhaseVanGenuchten);
+        return createNonWettingPhaseVanGenuchten(config);
     }
     else if (type == "WettingPhaseBrookCoreyOilGas")
     {
-        return createBrookCoreyOrVanGenuchten(
-            config,
-            BrookCoreyOrVanGenuchtenModelType::WettingPhaseBrookCoreyOilGas);
+        return createWettingPhaseBrookCoreyOilGas(config);
     }
     else if (type == "NonWettingPhaseBrookCoreyOilGas")
     {
-        return createBrookCoreyOrVanGenuchten(
-            config,
-            BrookCoreyOrVanGenuchtenModelType::NonWettingPhaseBrookCoreyOilGas);
+        return createNonWettingPhaseBrookCoreyOilGas(config);
     }
     else if (type == "Curve")
     {
+        config.checkConfigParameter("type", "Curve");
+
         std::vector<double> variables, values;
-        //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__type__curve}
+        //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__Curve__curve}
         auto const& curve_config = config.getConfigSubtree("curve");
-        //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__type__curve__data}
-        for (auto const& data_string :
-             curve_config.getConfigParameterList<std::string>("data"))
+        for (
+            auto const& data_string :
+            //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__Curve_curve__data}
+            curve_config.getConfigParameterList<std::string>("data"))
         {
             std::stringstream ss(data_string);
             double var, val;
@@ -142,12 +212,13 @@ std::unique_ptr<RelativePermeability> createRelativePermeabilityModel(
     else
     {
         OGS_FATAL(
-            "The capillary pressure model %s is unavailable.\n"
+            "The relative permeability model %s is unavailable.\n"
             "The available models are:"
             "\n\tWettingPhaseVanGenuchten,"
             "\n\tNonWettingPhaseVanGenuchten,"
             "\n\tWettingPhaseBrookCoreyOilGas,"
-            "\n\tNonWettingPhaseBrookCoreyOilGas.\n",
+            "\n\tNonWettingPhaseBrookCoreyOilGas,",
+            "\n\tCurve.\n",
             type.data());
     }
 }
--- a/MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/CreateRelativePermeabilityModel.cpp
+++ b/MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/CreateRelativePermeabilityModel.cpp
@@ -1,0 +1,156 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ * \file   CreateRelativePermeabilityModel.cpp
+ *
+ * Created on November 2, 2016, 11:43 AM
+ */
+
+#include "CreateRelativePermeabilityModel.h"
+
+#include <array>
+#include <memory>
+#include <sstream>
+#include <string>
+#include <vector>
+
+#include "BaseLib/ConfigTree.h"
+#include "BaseLib/Error.h"
+
+#include "MathLib/InterpolationAlgorithms/PiecewiseLinearInterpolation.h"
+
+#include "RelativePermeability.h"
+#include "ReletivePermeabilityCurve.h"
+#include "WettingPhaseVanGenuchten.h"
+#include "NonWettingPhaseVanGenuchten.h"
+#include "WettingPhaseBrookCoreyOilGas.h"
+#include "NonWettingPhaseBrookCoreyOilGas.h"
+
+namespace MaterialLib
+{
+namespace PorousMedium
+{
+enum class BrookCoreyOrVanGenuchtenModelType
+{
+    WettingPhaseVanGenuchten,
+    NonWettingPhaseVanGenuchten,
+    WettingPhaseBrookCoreyOilGas,
+    NonWettingPhaseBrookCoreyOilGas
+};
+
+/**
+ *     \param config ConfigTree object which contains the input data
+ *                   including <type>WettingPhaseBrookCoreyOilGas</type>
+ *                          or <type>NonWettingPhaseBrookCoreyOilGas</type>
+ *                          or <type>WettingPhaseVanGenuchten</type>
+ *                          or <type>NonWettingPhaseVanGenuchten</type>
+ *                   and it has a tag of <relative_permeability>
+*/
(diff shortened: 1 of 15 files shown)


MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/CreateRelativePermeabilityModel.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__WettingPhaseVanGenuchten__m}
+    const double m = config.getConfigParameter<double>("m");
+    if (m > 1.0)  // m <= 1
+    {
+        OGS_FATAL(
+            "The exponent parameter of WettingPhaseVanGenuchten relative\n"
+            " permeability model, m, must be in an interval of [0, 1]");
+    }
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__WettingPhaseVanGenuchten__m}
```
chleh: Typo: `__m` at the end?

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/CreateRelativePermeabilityModel.cpp
```diff
@@ -0,0 +1,227 @@
...
+    const double krel_max = config.getConfigParameter<double>("krel_min");
```
chleh: Typo: krel_max vs. min?

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/CreateRelativePermeabilityModel.cpp
```diff
@@ -0,0 +1,227 @@
...
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__NonWettingPhaseVanGenuchten__m}
+    const double m = config.getConfigParameter<double>("m");
+    if (m > 1.0)  // m <= 1
+    {
+        OGS_FATAL(
+            "The exponent parameter of NonWettingPhaseVanGenuchten relative\n"
+            " permeability model, m, must be in an interval of [0, 1]");
+    }
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__NonWettingPhaseVanGenuchten__m}
+    const double krel_max = config.getConfigParameter<double>("krel_min");
```
chleh: Typo? max vs min? And `__m` above.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/CreateRelativePermeabilityModel.cpp
```diff
@@ -0,0 +1,227 @@
...
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__WettingPhaseBrookCoreyOilGas__m}
+    const double m = config.getConfigParameter<double>("m");
+    if (m < 1.0)  // m >= 1
+    {
+        OGS_FATAL(
+            "The exponent parameter of WettingPhaseBrookCoreyOilGas\n"
+            "relative permeability model, m, must not be smaller than 1");
+    }
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__WettingPhaseBrookCoreyOilGas__m}
+    const double krel_max = config.getConfigParameter<double>("krel_min");
```
chleh: Looks like a recurrent pattern :-)

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/CreateRelativePermeabilityModel.cpp
```diff
@@ -0,0 +1,227 @@
...
+
+    return std::unique_ptr<RelativePermeability>(
+        new WettingPhaseBrookCoreyOilGas(Sr, Smax, m, krel_max));
+}
+
+/**
+    \param config ConfigTree object which contains the input data
+                  including <type>NonWettingPhaseBrookCoreyOilGas</type>
+                  and it has a tag of <relative_permeability>
```
chleh: Do you mean with that that the XML looks like that:
```
<relative_permeability>
  <type>NonWettingPhaseBrookCoreyOilGas</type>
  ...
<relative_permeability>
```
?

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/CreateRelativePermeabilityModel.cpp
```diff
@@ -0,0 +1,227 @@
...
+        //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__Curve__curve}
+        auto const& curve_config = config.getConfigSubtree("curve");
+        for (
+            auto const& data_string :
+            //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__Curve_curve__data}
+            curve_config.getConfigParameterList<std::string>("data"))
+        {
+            std::stringstream ss(data_string);
+            double var, val;
+            ss >> var >> val;
```
chleh: The error-checking is much better if you use `curve_config.getConfigParameterList<std::vector<double>>("data")` above.
Is it much better to have var and value alternating in the data? Until now curves were configured with separate variables and values tags.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/NonWettingPhaseBrookCoreyOilGas.cpp
```diff
@@ -0,0 +1,35 @@
...
+namespace PorousMedium
+{
+double NonWettingPhaseBrookCoreyOilGas::getValue(
+    const double saturation_w) const
+{
+    const double S = MathLib::limitValueInInterval(
+        saturation_w, _Sr + _minor_offset, _Smax - _minor_offset);
+    const double Se = (S - _Sr) / (_Smax - _Sr);
+    const double krel =
+        std::pow(1.0 - Se, 2) * (1.0 - std::pow(Se, 1.0 + 2.0 / _mm));
```
chleh: Writing the square out should be faster than std::pow(..., 2).

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/NonWettingPhaseBrookCoreyOilGas.cpp
```diff
@@ -0,0 +1,35 @@
...
+    return krel < _krel_min ? _krel_min : krel;
```
chleh: Minor: could be std::max

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/NonWettingPhaseBrookCoreyOilGas.h
```diff
@@ -0,0 +1,73 @@
...
+
+    /// Get model name.
+    std::string getName() const override
+    {
+        return "Non-wetting phase Brook-Corey relative permeability model.";
+    }
+
+    /// Get relative permeability value.
+    /// \param saturation_w Non-wetting phase saturation
+    double getValue(const double saturation_w) const override;
```
chleh: What about derivatives? Are they not needed?

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/NonWettingPhaseVanGenuchten.cpp
```diff
@@ -0,0 +1,33 @@
...
+namespace MaterialLib
+{
+namespace PorousMedium
+{
+double NonWettingPhaseVanGenuchten::getValue(const double saturation_w) const
+{
+    const double S = MathLib::limitValueInInterval(
+        saturation_w, _Sr + _minor_offset, _Smax - _minor_offset);
+    const double Se = (S - _Sr) / (_Smax - _Sr);
+    const double krel = std::pow(1.0 - Se, 1.0 / 3.0) *
```
chleh: Minor: there is also `std::cbrt()`, which is probably a bit faster.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/NonWettingPhaseVanGenuchten.h
```diff
@@ -0,0 +1,74 @@
...
+
+    /// Get model name.
+    std::string getName() const override
+    {
+        return "Non-wetting phase van Genuchten relative permeability model.";
+    }
+
+    /// Get relative permeability value.
+    /// \param saturation_w Non-wetting phase saturation
+    double getValue(const double saturation_w) const override;
```
chleh: Likewise: Are derivatives needed?

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/ReletivePermeabilityCurve.h
```diff
@@ -0,0 +1,60 @@
...
+#include "RelativePermeability.h"
+
+#include "MathLib/MathTools.h"
+#include "MathLib/InterpolationAlgorithms/PiecewiseLinearInterpolation.h"
+
+namespace MaterialLib
+{
+namespace PorousMedium
+{
+class ReletivePermeabilityCurve final : public RelativePermeability
```
chleh: typo: Relative

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/WettingPhaseVanGenuchten.cpp
```diff
@@ -0,0 +1,35 @@
...
+namespace PorousMedium
+{
+double WettingPhaseVanGenuchten::getValue(const double saturation) const
+{
+    const double S = MathLib::limitValueInInterval(
+        saturation, _Sr + _minor_offset, _Smax - _minor_offset);
+    const double Se = (S - _Sr) / (_Smax - _Sr);
+    const double krel =
+        std::sqrt(Se) *
+        std::pow(1.0 - std::pow(1.0 - std::pow(Se, 1.0 / _mm), _mm), 2);
```
chleh: Minor: `std::pow` for squaring is rather slow.

Tests/MaterialLib/TestRelativePermeabilityModel.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+std::unique_ptr<RelativePermeability> createRelativePermeabilityModel(
+    const char xml[])
+{
+    auto const ptree = readXml(xml);
+    BaseLib::ConfigTree conf(ptree, "", BaseLib::ConfigTree::onerror,
+                             BaseLib::ConfigTree::onwarning);
+    auto const& sub_config = conf.getConfigSubtree("relative_permeability");
+    //! \ogs_file_attr{prj__material_property__porous_medium__porous_medium__id}
+    auto const id = sub_config.getConfigAttributeOptional<int>("id");
```
chleh: If you don't need an attribute, you can use `sub_config.ignoreAttribute("id")`. I think such a method exists.
Btw: What is the id for? You don't read it in the rest of this PR, so at least you could remove it from the test.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/CreateRelativePermeabilityModel.cpp
wenqing: Corrected to __krel_min

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/CreateRelativePermeabilityModel.cpp
wenqing: Corrected to krel_min

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/CreateRelativePermeabilityModel.cpp
wenqing: Corrected.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/CreateRelativePermeabilityModel.cpp
wenqing: corrected.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/CreateRelativePermeabilityModel.cpp
wenqing: getConfigParameterList<std::vector<double>> does not work. Changed syntax to
```
             <curve>
                   <point><data> 0. 0.9   </data></point>
                   <point><data> 0.4 0.5  </data></point>
                   <point><data> 0.9 0.01 </data></point>
              </curve>
```
and read it by
```
        for (  auto const& point_config :  curve_config.getConfigSubtreeList("point"))
        {
            const auto& point =
                               point_config.getConfigParameter<std::vector<double>>("data");
            assert(point.size() == 2);
           .... 
       }
```

I think, reading curve data in pair is easy for manually checking.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/NonWettingPhaseBrookCoreyOilGas.cpp
wenqing: Changed.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/NonWettingPhaseBrookCoreyOilGas.cpp
wenqing: Changed.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/NonWettingPhaseVanGenuchten.cpp
wenqing: Changed.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/NonWettingPhaseVanGenuchten.h
wenqing: Added

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/WettingPhaseVanGenuchten.cpp
wenqing: Changed.

Tests/MaterialLib/TestRelativePermeabilityModel.cpp
wenqing: id is reserved for material IDs of elements. We can not input the model parameters as Parameter data because that we may use different models for different material zones in a modelling.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/CreateRelativePermeabilityModel.cpp
chleh: Thank you for the change. I misunderstood your previous code. Sadly, currently we cannot get around `<point><data> 0. 0.9   </data></point>`, because, there is no `std::vector<T>` overload for `ConfigTree::getValue<T>()`.

chleh:
Thanks for the update.

wenqing:
@chleh  For the curve, when I started to implement the curve type material model, I found  that the curve should be placed to where it used or belong to. Parsing all curves and saved them as a member of ProjectData is not the best choice. Material models can have curves, BCs can have curves. These curves should be encapsulated to the classes that use them.

Besides, input the variables and values in separated vector is not good to perform a manual input.

Therefore, I did not add the parsing of the curve for relative permeability to  ProjectData class.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/CreateRelativePermeabilityModel.cpp
```diff
@@ -0,0 +1,229 @@
...
+        std::vector<double> variables, values;
+        //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__Curve__curve}
+        auto const& curve_config = config.getConfigSubtree("curve");
+        for (
+            auto const& point_config :
+            //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__Curve_curve__point}
+            curve_config.getConfigSubtreeList("point"))
+        {
+            const auto& point =
+                //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__Curve_curve__points__data}
```
endJunction: I guess the ending should be '__point__data', one 's' too much.

Tests/MaterialLib/TestRelativePermeabilityModel.cpp
```diff
@@ -0,0 +1,201 @@
...
+
+std::unique_ptr<RelativePermeability> createRelativePermeabilityModel(
+    const char xml[])
+{
+    auto const ptree = readXml(xml);
+    BaseLib::ConfigTree conf(ptree, "", BaseLib::ConfigTree::onerror,
+                             BaseLib::ConfigTree::onwarning);
+    auto const& sub_config = conf.getConfigSubtree("relative_permeability");
+    sub_config.ignoreConfigAttribute("id");
+    return MaterialLib::PorousMedium::createRelativePermeabilityModel(
```
endJunction: If you are 'using namespace MaterialLib::PorousMedium' than this can be shorter. Choose one variant.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/NonWettingPhaseVanGenuchten.h
```diff
@@ -0,0 +1,78 @@
...
+#define OGS_NON_WETTING_PHASE_VAN_GENUCHTEN_H
+
+#include "RelativePermeability.h"
+
+namespace MaterialLib
+{
+namespace PorousMedium
+{
+/**
+ *   \brief BrookCorey oil-gas model: non-wetting phase
```
endJunction: This one seems to be of some van Genuchten type.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/NonWettingPhaseVanGenuchten.cpp
```diff
@@ -0,0 +1,46 @@
...
+{
+    const double S = MathLib::limitValueInInterval(
+        saturation_w, _Sr + _minor_offset, _Smax - _minor_offset);
+    const double Se = (S - _Sr) / (_Smax - _Sr);
+    const double cbrt1_Se = std::cbrt(1.0 - Se);
+    const double temp_val = 1.0 - std::pow(Se, 1.0 / _mm);
+    return (-std::pow(temp_val, 2. * _mm) / (3. * cbrt1_Se * cbrt1_Se) -
+            2. * cbrt1_Se * std::pow(temp_val, 2. * _mm - 1.) *
+                std::pow(Se, (1. - _mm) / _mm)) /
+           (_Smax - _Sr);
```
endJunction: Unfortunately (and I made that mistake recently too) identifiers starting with an underscore followed by a capital case letter are reserved for the compilers.

Tests/MaterialLib/TestRelativePermeabilityModel.cpp
```diff
@@ -0,0 +1,201 @@
...
+TEST(MaterialPorousMedium, checkReletivePermeabilityCurve)
+{
+    const char xml[] =
+        "<relative_permeability id=\"0\">"
+        "   <type>Curve</type>"
+        "       <curve>"
+        "           <point><data> 0. 0.9   </data></point>"
+        "           <point><data> 0.4 0.5  </data></point>"
+        "           <point><data> 0.9 0.01 </data></point>"
+        "       </curve>"
```
wenqing: See my message to Christoph:
```
 For the curve, when I started to implement the curve type material model,
 I found that the curve should be placed to where it used or belong to. 
 Parsing all curves and saved them as a member of ProjectData
 is not the best choice. Material models can have curves, BCs can 
have curves. 
These curves should be encapsulated to the classes that use them.

Besides, input the variables and values in separated vector is not 
good to perform a manual input.

Therefore, I did not add the parsing of the curve for relative
 permeability to ProjectData class.
```

If you and others think it is not good to have this XML syntax for a curve. I can use the two array type input by extracting  createPiecewiseLinearInterpolation(BaseLib::ConfigTree const& config) from Project Data and saving it to a single file in MathLib/InterpolationAlgorithms.

Tests/MaterialLib/TestRelativePermeabilityModel.cpp
```diff
@@ -0,0 +1,201 @@
...
+    for (std::size_t i = 0; i < S.size(); i++)
+    {
+        ASSERT_NEAR(krel[i], perm_model->getValue(S[i]), 1.e-9);
+
+        // i ==  S.size() -1, S(=0.85) is limited to 0.8, and the numerical
+        // derivative of krel is unavailable.
+        if (i == S.size() - 1)
+            continue;
+        // Compare the derivative with numerical one.
+        ASSERT_NEAR((perm_model->getValue(S[i] + purterbation) -
```
wenqing: Corrected.

Tests/MaterialLib/TestRelativePermeabilityModel.cpp
wenqing: Removed.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/NonWettingPhaseVanGenuchten.h
wenqing: Corrected.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/CreateRelativePermeabilityModel.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__WettingPhaseVanGenuchten__sr}
+    const double Sr = config.getConfigParameter<double>("sr");
+
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__WettingPhaseVanGenuchten__smax}
+    const double Smax = config.getConfigParameter<double>("smax");
+
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__WettingPhaseVanGenuchten__m}
+    const double m = config.getConfigParameter<double>("m");
+    if (m > 1.0)  // m <= 1
```
TomFischer: If `m` is negative is not checked.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/CreateRelativePermeabilityModel.cpp
```diff
@@ -0,0 +1,229 @@
...
+    return std::unique_ptr<RelativePermeability>(
+        new WettingPhaseVanGenuchten(Sr, Smax, m, krel_min));
+}
+
+/**
+    \param config ConfigTree object which contains the input data
+                  including <type>NonWettingPhaseVanGenuchten</type>
+                  and it has a tag of <relative_permeability>
+*/
+std::unique_ptr<RelativePermeability> createNonWettingPhaseVanGenuchten(
```
TomFischer: `createNonWettingPhaseVanGenuchten` and `createWettingPhaseVanGenuchten` seems to be almost identical. Is it possible to refactor common things?

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/WettingPhaseBrookCoreyOilGas.h
```diff
@@ -0,0 +1,72 @@
...
+    /// Get the derivative of relative permeability with respect to saturation.
+    /// \param saturation Wetting phase saturation
+    double getdValue(const double saturation) const override;
+
+private:
+    const double _saturation_r;    ///< Residual saturation.
+    const double _saturation_max;  ///< Maximum saturation.
+    const double _mm;              ///< Exponent (>=1.0), n=1/(1-mm).
+    const double _krel_min;        ///< Minimum relative permeability
+};
```
TomFischer: The above attributes are use in other classes, too (for instance in `NonWettingPhaseBrookCoreyOilGas`). Maybe it makes sense that these classes have a common base class?

TomFischer: Additional, it would be possible to implement `limitValueInInterval(saturation)` as a method in this base class. The base class method could use the needed attributes (`_saturation_r`, `_minor_offset`, and `_saturation_max`) and the code would be more readable in my opinion.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/CreateRelativePermeabilityModel.cpp
wenqing: At the beginning, I had only one local function with an option for creating different models. Because of Dima's comment in PR #1517, which has a similar function, I split the single function to four functions. Another consideration is keyword documentation. Although parameter names are the same in the four models, they should be  documented for each model individually.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/WettingPhaseBrookCoreyOilGas.h
wenqing: @TomFischer limitValueInInterval is also used in capillary pressure classes. Therefore, it was  added to MathTool

wenqing: Moved two common members, _saturation_r and saturation_max to the base class.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/CreateRelativePermeabilityModel.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__NonWettingPhaseBrookCoreyOilGas__sr}
+    const double Sr = config.getConfigParameter<double>("sr");
+
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__NonWettingPhaseBrookCoreyOilGas__smax}
+    const double Smax = config.getConfigParameter<double>("smax");
+
+    //! \ogs_file_param{material_property__porous_medium__porous_medium__relative_permeability__NonWettingPhaseBrookCoreyOilGas__m}
+    const double m = config.getConfigParameter<double>("m");
+    if (m < 1.0)  // m >= 1
```
TomFischer: In this modell `m < 0` is allowed?

Tests/MaterialLib/TestRelativePermeabilityModel.cpp
TomFischer: I also would prefer one implementation for reading curves.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/CreateRelativePermeabilityModel.cpp
wenqing: No. m<0 is included in m<1.0.

Tests/MaterialLib/TestRelativePermeabilityModel.cpp
wenqing: @TomFischer I prepared one in PR  #1529.
@endJunction @TomFischer. If all of you prefer two array to read a curve. I can change it in #1529.

For this PR. I think, I can drop the curve type and move it a new PR after the curve input is unified.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/CreateRelativePermeabilityModel.cpp
TomFischer: In other models you check if `m` is  in `[0,1]`. Here you do not do this. So is it allowed to give negative values for `m` in this model?

wenqing: m is a double paramerer. If a negative value is given for it, the value is smaller than 1.0 too, and the program gives a fatal error as well. For Brook-Corey models, m>1.

wenqing: For van Genuchten model m in[0,1].

TomFischer: Now I understand. Thanks!

TomFischer:
When rebase is fine, tests are ok: :+1:

wenqing:
Thanks for your comments. If there is not further comment, I am going to merge it.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/CreateRelativePermeabilityModel.cpp
```diff
@@ -0,0 +1,237 @@
...
+
+    //! \ogs_file_param{material__porous_medium__relative_permeability__WettingPhaseVanGenuchten__sr}
+    const double Sr = config.getConfigParameter<double>("sr");
+
+    //! \ogs_file_param{material__porous_medium__relative_permeability__WettingPhaseVanGenuchten__smax}
+    const double Smax = config.getConfigParameter<double>("smax");
+
+    //! \ogs_file_param{material__porous_medium__relative_permeability__WettingPhaseVanGenuchten__m}
+    const double m = config.getConfigParameter<double>("m");
+    if (m < 0. || m > 1.0)  // m <= 1
```
endJunction: the comment does not match the expression. I'd remove the comments; the reason is already given in the error message.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/NonWettingPhaseVanGenuchten.h
```diff
@@ -0,0 +1,78 @@
...
+     * @param Snmax     Maximum saturation  of the non-wetting phase,
+     *                  \f$ S^n_{\mbox{max}} \f$
+     * @param m         Exponent, \f$ m \in [0,1]\f$
+     * @param krel_min  Minimum relative permeability,
+     *                  \f$ k_{rel}^{\mbox{min}}\f$
+     */
+    NonWettingPhaseVanGenuchten(const double Snr, const double Snmax,
+                                const double m, const double krel_min)
+        : RelativePermeability(1. - Snmax, 1. - Snr),
+          _mm(m),
```
endJunction: just curious, why not `_m`? It would be consistent with the parameter name.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/NonWettingPhaseVanGenuchten.h
```diff
@@ -0,0 +1,78 @@
...
+    /// Get relative permeability value.
+    /// \param saturation_w Wetting phase saturation
+    double getValue(const double saturation_w) const override;
+
+    /// Get the derivative of relative permeability with respect to saturation.
+    /// \param saturation_w Wetting phase saturation
+    double getdValue(const double saturation_w) const override;
+
+private:
+    const double _mm;        ///< Exponent (<=1.0), n=1/(1-mm).
```
endJunction: In the corresponding create function you check if this value is in [0,1] interval; the comment here is different. I can not say which one is correct, but if the checks and comments are kept together it is easier to spot mistakes.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/RelativePermeability.h
```diff
@@ -0,0 +1,61 @@
...
+    /// Get relative permeability value.
+    /// \param saturation Wetting phase saturation
+    virtual double getValue(const double saturation) const = 0;
+
+    /// Get the derivative of relative permeability with respect to saturation.
+    /// \param saturation Wetting phase saturation
+    virtual double getdValue(const double saturation) const = 0;
+
+protected:
+    /** A small number for an offset to set the bound of S, the saturation, such
```
endJunction: Sometimes doxygen comments are in /** */ style, sometimes in ///.  Choose one, stick to it.

Tests/MaterialLib/TestRelativePermeabilityModel.cpp
```diff
@@ -0,0 +1,199 @@
...
+    auto const& sub_config = conf.getConfigSubtree("relative_permeability");
+    sub_config.ignoreConfigAttribute("id");
+    return createRelativePermeabilityModel(sub_config);
+}
+
+TEST(MaterialPorousMedium, checkWettingPhaseVanGenuchten)
+{
+    const char xml[] =
+        // Should be <relative_permeability id="0">
+        "<relative_permeability id=\"0\">"
```
endJunction: The comment's statement is true. Probably the comment is no longer needed.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/NonWettingPhaseVanGenuchten.h
wenqing: Changed to _m.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/NonWettingPhaseVanGenuchten.h
wenqing: Changed the comment to 
///< Exponent m, m in [0, 1], n=1/(1-m).

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/RelativePermeability.h
wenqing: I use /// for comments be less than two lines, and /** */ for those more than two lines.

Tests/MaterialLib/TestRelativePermeabilityModel.cpp
wenqing: Removed.

MaterialLib/PorousMedium/UnsaturatedProperty/RelativePermeability/RelativePermeability.h
endJunction: :) the rule is fine, but this particular one has only two lines ;)
But keep it like it is, I don't mind it too much.

PR #1534, opened 2016-11-10
TH monolithic process

TomFischer:
Tianyuan and I implemented the TH process. The implementation has at the moment one test for constant viscosity case. Next step will be including the laws from material lib and at least one benchmarks for variable viscosity will be prepared.


TomFischer:
Jenkins, test this please.

TomFischer:
Jenkins, test this please.

ProcessLib/DensityDrivenFlow/CreateDensityDrivenFlowProcess.cpp
```diff
@@ -0,0 +1,175 @@
...
+    // Parameter for the thermal conductivity of the fluid.
+    auto& thermal_conductivity_fluid = findParameter<double>(
+        config,
+        //! \ogs_file_param_special{process__DensityDrivenFlow__thermal_conductivity_fluid}
+        "thermal_conductivity_fluid", parameters, 1);
+    DBUG("Use \'%s\' as thermal_conductivity_fluid parameter.",
+         thermal_conductivity_fluid.name.c_str());
+
+    // Parameter for the thermal expansion coefficient.
+    auto& thermal_expansion_coefficient = findParameter<double>(
```
norihiro-w: of solid or fluid?

ProcessLib/DensityDrivenFlow/DensityDrivenFlowFEM.h
```diff
@@ -0,0 +1,250 @@
...
+            double p_int_pt = 0.0;
+            // Order matters: First T, then P!
+            NumLib::shapeFunctionInterpolate(local_x, sm.N, T_int_pt, p_int_pt);
+
+            double const delta_T(T_int_pt - temperature0);
+            // TODO include this via material lib
+            double density_water_T = density_fluid * (1 - beta * delta_T);
+            // TODO include viscosity computations via material lib
+            // double const viscosity = viscosity0 * std::exp(- delta_T/75.0);
+            double hydraulic_conductivity = intrinsic_permeability / viscosity0;
```
norihiro-w: this is not hydraulic conductivity. `K=rho*g*k/mu`

ProcessLib/DensityDrivenFlow/CreateDensityDrivenFlowProcess.cpp
```diff
@@ -0,0 +1,175 @@
...
+{
+    //! \ogs_file_param{process__type}
+    config.checkConfigParameter("type", "DensityDrivenFlow");
+
+    DBUG("Create DensityDrivenFlowProcess.");
+
+    // Process variable.
+    auto process_variables = findProcessVariables(
+        variables, config,
+        {"temperature_variable", "pressure_variable"});  // configure two Pcs
```
norihiro-w: minor: can be shortly "temperature" and "pressure"

ProcessLib/DensityDrivenFlow/CreateDensityDrivenFlowProcess.cpp
```diff
@@ -0,0 +1,175 @@
...
+    // Parameter for the intrinsic permeability.
+    auto& intrinsic_permeability = findParameter<double>(
+        config,
+        //! \ogs_file_param_special{process__DensityDrivenFlow__intrinsic_permeability}
+        "intrinsic_permeability", parameters, 1);
+    DBUG("Use \'%s\' as intrinsic_permeability parameter.",
+         intrinsic_permeability.name.c_str());
+
+    // Parameter for the storage.
+    auto& storage = findParameter<double>(
```
norihiro-w: minor: HM process uses "sepcific_storage" for this

ProcessLib/DensityDrivenFlow/CreateDensityDrivenFlowProcess.cpp
```diff
@@ -0,0 +1,175 @@
...
+        config,
+        //! \ogs_file_param_special{process__DensityDrivenFlow__storage}
+        "storage", parameters, 1);
+    DBUG("Use \'%s\' as storage parameter.", storage.name.c_str());
+
+    // Parameter for the reference_temperature.
+    auto& reference_temperature = findParameter<double>(
```
norihiro-w: minor: the name can be more specific, since it will be used in the fluid density model.

ProcessLib/DensityDrivenFlow/CreateDensityDrivenFlowProcess.h
```diff
@@ -0,0 +1,31 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#ifndef PROCESS_LIB_CREATE_DensityDrivenFlowPROCESS_H_
```
norihiro-w: why not all upper case characters?

Applications/ApplicationsLib/ProjectData.cpp
```diff
@@ -338,6 +339,13 @@ void ProjectData::parseProcesses(BaseLib::ConfigTree const& processes_config,
                         "dimension");
             }
         }
+        else if (type == "DensityDrivenFlow")
```
norihiro-w: It might be better to make the process name more specific. In general, density driven flow occurs due to solute concentration gradient or temperature gradient. So far this process targets only the temperature gradient.

norihiro-w:
No critical points in the code. Some parameter/variable names should be corrected because they are unspecific or wrong. Minor comments can be ignored if you don't like.

norihiro-w:
BTW current implementation doesn't get benefit of monolithic coupling because there is no coupling term in the equations, i.e. same as partitioned coupling. But I guess you will introduce some coupling terms later on.

norihiro-w:
last thing and this is also minor comment: I think you can remove all `hasGravity` checks. You cann't make density driven flow without gravity effects.

ProcessLib/DensityDrivenFlow/DensityDrivenFlowFEM.h
```diff
@@ -0,0 +1,250 @@
...
+            Mtt.noalias() +=
+                integral_term * sm.N.transpose() * heat_capacity * sm.N;
+            Mpp.noalias() += integral_term * sm.N.transpose() * storage * sm.N;
+            if (_process_data.has_gravity)
+                Bp += hydraulic_conductivity * integral_term *
+                      sm.dNdx.transpose() * b * density_water_T;
+            /* with Oberbeck-Boussing assumption density difference only exists
+             * in buoyancy effects */
+
+            // velocity computed for output.
```
wenqing: The velocity for output have been calculated again in computeSecondaryVariableConcrete. Otherwise, the output velocity is from the previous nonlinear iteration or from the previous time step. Therefore following loop has be move to computeSecondaryVariableConcrete too.

ProcessLib/DensityDrivenFlow/DensityDrivenFlowFEM.h
```diff
@@ -0,0 +1,250 @@
...
+                (sm.dNdx.transpose() * thermal_conductivity * sm.dNdx +
+                 sm.N.transpose() * velocity.transpose() * sm.dNdx *
+                     density_fluid * specific_heat_capacity_fluid);
+            Kpp.noalias() +=
+                integral_term * sm.dNdx.transpose() *
+                /* Real_hydraulic_conductivity */ hydraulic_conductivity *
+                sm.dNdx;
+            Mtt.noalias() +=
+                integral_term * sm.N.transpose() * heat_capacity * sm.N;
+            Mpp.noalias() += integral_term * sm.N.transpose() * storage * sm.N;
```
wenqing: Fluid thermal expansion term (related to drho_l/dT) in fluid flow equation is missing. Of course, it can be added easily later.

ProcessLib/DensityDrivenFlow/DensityDrivenFlowFEM.h
```diff
@@ -0,0 +1,250 @@
...
+
+        for (std::size_t ip(0); ip < n_integration_points; ip++)
+        {
+            pos.setIntegrationPoint(ip);
+            auto const temperature0 =
+                _process_data.reference_temperature(t, pos)[0];
+            auto const beta =
+                _process_data.thermal_expansion_coefficient(t, pos)[0];
+
+            auto const body_force = _process_data.specific_body_force(t, pos);
```
wenqing: This can be moved outside the integration loop. Also, it is not body force, it is for the gravitational term of the Darcy velocity and it is homogeneous constant.  I suggest to introduce two constants as that in LiquidFlow
https://github.com/ufz/ogs/blob/master/ProcessLib/LiquidFlow/LiquidFlowLocalAssembler.h#L188.

The input syntax for the two constants are 
https://github.com/ufz/ogs-data/blob/9f817e0221d4b19de2e053898c4cec5c50c0d98c/Parabolic/LiquidFlow/GravityDriven/gravity_driven.prj#L12

Then the computation for this term in the current implementation is simplified as
```
           if (_process_data.has_gravity)
+                velocity[_gravitational_axis_id] += hydraulic_conductivity * density_water_T * _gravitational_acceleration ;
```
Also the anisotropicity of the permeability can be handled easily with the two constants.

ProcessLib/DensityDrivenFlow/CreateDensityDrivenFlowProcess.cpp
```diff
@@ -0,0 +1,175 @@
...
+    // Parameter for the specific heat capacity of the fluid.
+    auto& specific_heat_capacity_fluid = findParameter<double>(
+        config,
+        //! \ogs_file_param_special{process__DensityDrivenFlow__specific_heat_capacity_fluid}
+        "specific_heat_capacity_fluid", parameters, 1);
+    DBUG("Use \'%s\' as specific_heat_capacity_fluid parameter.",
+         specific_heat_capacity_fluid.name.c_str());
+
+    // Parameter for the thermal conductivity of the solid.
+    auto& thermal_conductivity_solid = findParameter<double>(
```
wenqing: Add a comment that only isotropic thermal conductivity is considered.

ProcessLib/DensityDrivenFlow/CreateDensityDrivenFlowProcess.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+    // Porosity parameter.
+    auto& porosity = findParameter<double>(
+        config,
+        //! \ogs_file_param_special{process__DensityDrivenFlow__porosity}
+        "porosity", parameters, 1);
+    DBUG("Use \'%s\' as porosity parameter.", porosity.name.c_str());
+
+    // Parameter for the intrinsic permeability.
+    auto& intrinsic_permeability = findParameter<double>(
```
wenqing: Add a comment that only isotropic permeability is considered.

ProcessLib/DensityDrivenFlow/DensityDrivenFlowFEM.h
TomFischer: > Also, it is not body force, it is for the gravitational term of the Darcy velocity and it is homogeneous constant.

A body force is a force that acts throughout a volume. In my opinion gravity is a body force. I am unsure what do you mean by *homogeneous constant*. Here the `body_force` it is not a scalar, it is a vector.

ProcessLib/DensityDrivenFlow/CreateDensityDrivenFlowProcess.h
TomFischer: If an appropriate process name is choosen I'll will update the guards using upper case characters.

TomFischer:
@norihiro-w 
> BTW current implementation doesn't get benefit of monolithic coupling because there is no coupling term in the equations, i.e. same as partitioned coupling. But I guess you will introduce some coupling terms later on.

Yes, I know that the coupling terms are missing. As far as I understand, Tianyuan started with the Overbeck-Boussinesq approximation where it is assumed that the density difference is only in buoyancy effect.

Applications/ApplicationsLib/ProjectData.cpp
TomFischer: Just discussed with Fabien: Thermohaline would be fine if we had mass transport included. At this stage we should first decide how to handle the processes. Should we implement one process with the possibility to switch between thermal convection and thermaohaline convection or should there be two implementations?

ProcessLib/DensityDrivenFlow/DensityDrivenFlowFEM.h
wenqing: I mean a constant vector. I think it is the gravity vector, since the density is not included in the input. If it is called   body_force, one may make a mistake by inputting (0, 0, rho*g) for it. 

Since rho can vary, that type of input for LiuidFlowProcess just means 'gravity vector component  in i direction'.

Applications/ApplicationsLib/ProjectData.cpp
fabienma: I think Thermohaline will eventually be implemented. Let us decide together with @waltherm

waltherm: Hi all, I just discussed with @TomFischer and think, that we should call its something like HT, HC, or HTC. ´´´DensityDrivenFlow´´´ is not complete in its description, as we may have additional variables to change, like viscosity. Also, the processes may simply be used with constant variables for "simple" solute/heat transport.

fabienma: yes, this settles any confusion. Will it be then HTC with the option to switch from HTC to HC / HT or we keep 3 separates implementations?

waltherm: @TomFischer said, that theoretically, you could run an HTC implementation as a purely HT setup, but this would not be as high-performing as an HT implementation. So he suggested to have several implementation. This would also fit into our development schedule, ie to build upon successive work any next step of complexity (eg HT -> HC -> HTC -> H_unsatTC).

endJunction:
I think the CMakeLists.txt in ProcessLib/DensityDrivenFlow is not needed.

ProcessLib/DensityDrivenFlow/CreateDensityDrivenFlowProcess.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+    // Parameter for the thermal expansion coefficient.
+    auto& thermal_expansion_coefficient_fluid = findParameter<double>(
+        config,
+        //! \ogs_file_param_special{process__DensityDrivenFlow__thermal_expansion_coefficient_fluid}
+        "thermal_expansion_coefficient_fluid", parameters, 1);
+    DBUG("Use \'%s\' as thermal expansion coefficient of the fluid parameter.",
+         thermal_expansion_coefficient_fluid.name.c_str());
+
+    // Specific body force parameter.
```
endJunction: There is a more correct version of reading this; see (HM process creation)[https://github.com/ufz/ogs/blob/master/ProcessLib/HydroMechanics/CreateHydroMechanicsProcess.cpp#L163]
Was my mistake to make body forces a parameter in the first place.

ProcessLib/DensityDrivenFlow/DensityDrivenFlowFEM.h
```diff
@@ -0,0 +1,251 @@
...
+          unsigned GlobalDim>
+class LocalAssemblerData : public DensityDrivenFlowLocalAssemblerInterface
+{
+    using ShapeMatricesType = ShapeMatrixPolicyType<ShapeFunction, GlobalDim>;
+    using ShapeMatrices = typename ShapeMatricesType::ShapeMatrices;
+
+    using LocalAssemblerTraits =
+        ProcessLib::LocalAssemblerTraits<ShapeMatricesType,
+                                         ShapeFunction::NPOINTS, NUM_NODAL_DOF,
+                                         /* number of pcs vars */ GlobalDim>;
```
endJunction: The comment might be superfluous

ProcessLib/DensityDrivenFlow/DensityDrivenFlowFEM.h
```diff
@@ -0,0 +1,251 @@
...
+                _process_data.specific_heat_capacity_fluid(t, pos)[0];
+            double const heat_capacity =
+                density_solid * specific_heat_capacity_solid * (1 - porosity) +
+                density_fluid * specific_heat_capacity_fluid * porosity;
+
+            auto const thermal_conductivity_solid =
+                _process_data.thermal_conductivity_solid(t, pos)[0];
+            auto const thermal_conductivity_fluid =
+                _process_data.thermal_conductivity_fluid(t, pos)[0];
+            double thermal_conductivity =
```
endJunction: const?

ProcessLib/DensityDrivenFlow/DensityDrivenFlowFEM.h
```diff
@@ -0,0 +1,251 @@
...
+            // TODO include this via material lib
+            double density_water_T = density_fluid * (1 - beta * delta_T);
+            // TODO include viscosity computations via material lib
+            // double const viscosity = viscosity0 * std::exp(- delta_T/75.0);
+            double const perm_visc =
+                intrinsic_permeability / viscosity0;
+
+            auto velocity = (- perm_visc *
+                             (sm.dNdx * p_nodal_values - density_water_T * b))
+                                .eval();
```
endJunction: Could you please give the concrete type for velocity (also const) and drop the eval?

ProcessLib/DensityDrivenFlow/DensityDrivenFlowFEM.h
```diff
@@ -0,0 +1,251 @@
...
+            auto const integral_term = sm.detJ * wp.getWeight();
+            // matrix assembly
+            Ktt.noalias() +=
+                integral_term *
+                (sm.dNdx.transpose() * thermal_conductivity * sm.dNdx +
+                 sm.N.transpose() * velocity.transpose() * sm.dNdx *
+                     density_fluid * specific_heat_capacity_fluid);
+            Kpp.noalias() +=
+                integral_term * sm.dNdx.transpose() *
+                /* Real_hydraulic_conductivity */ perm_visc *
```
endJunction: The comment is probably no longer needed.

ProcessLib/DensityDrivenFlow/DensityDrivenFlowFEM.h
```diff
@@ -0,0 +1,251 @@
...
+            // TODO include viscosity computations via material lib
+            // double const viscosity = viscosity0 * std::exp(- delta_T/75.0);
+            double const perm_visc =
+                intrinsic_permeability / viscosity0;
+
+            auto velocity = (- perm_visc *
+                             (sm.dNdx * p_nodal_values - density_water_T * b))
+                                .eval();
+
+            auto const integral_term = sm.detJ * wp.getWeight();
```
endJunction: integral measure for axialsymmetric case is missing; See [GWF FEM](https://github.com/ufz/ogs/blob/master/ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h#L107)

ProcessLib/DensityDrivenFlow/DensityDrivenFlowProcess.cpp
```diff
@@ -0,0 +1,122 @@
...
+    MeshLib::Mesh const& mesh,
+    unsigned const integration_order)
+{
+    ProcessLib::ProcessVariable const& pv = getProcessVariables()[0];
+    ProcessLib::createLocalAssemblers<LocalAssemblerData>(
+        mesh.getDimension(), mesh.getElements(), dof_table,
+        pv.getShapeFunctionOrder(), _local_assemblers,
+        mesh.isAxiallySymmetric(), integration_order, _process_data);
+
+    /* _secondary_variables.addSecondaryVariable(
```
endJunction: Please explain in a comment why we need this to be commented out. Potentially add a todo if this is needed in later improvements.

ProcessLib/DensityDrivenFlow/CreateDensityDrivenFlowProcess.cpp
TomFischer: In the creation of the HM process there is an dimension information via template parameter used to create the `Eigen::Matrix<double, DisplacementDim, 1> specific_body_force;`. To implement the creation of `specific_body_force` analogous here would require an compile time (since template argument) mechanism to find out the dimension. This is possible but would need larger changes. If this is desired, I would change this in a subsequent PR.

norihiro-w: I don't mind if you change it in another PR. FYI, an easy solution without using template parameter could be to declare `specific_body_force` as `Eigen::Vector3d` and access it with `specific_body_force.head(mesh.getDimension())` or `specific_body_force.head<GlobalDim>()`.

TomFischer:
Thanks for the comments and discussions. I rebased, cleaned up the history and renamed the process to HT.

norihiro-w:
👍  if tests are working

TomFischer:
Forgot to update the data repo. Jenkins test this please.

PR #1535, opened 2016-11-10
Hydromechanics with the LIE approach

norihiro-w:
Update: this PR is divided into #1537, #1538, #1539, #1540, #1541. Changes in those PRs will be merged into this branch and Jenkins will check if the benchmark still works.

This PR adds the new process HYDRO_MECHANICS_WITH_LIE which solves HM problems with fractures. At the moment, only a single fracture is supported.


chleh:
Can this PR be closed now, since it is divided into several parts?
If it is closed, it won't cause any work for Jenkins if some other PRs are merged (because Jenkins then rebuilds all open PRs AFAIK).

norihiro-w:
replaced by #1541

PR #1536, opened 2016-11-11
Tiny change

chleh:
fixes an uninitialized warning

A similar change would be sensible for SmallDeformationProcessData. Furthermore there the naming of members is inconsistent: _material, t, dt.

### DIFF
--- a/ProcessLib/SmallDeformation/SmallDeformationProcessData.h
+++ b/ProcessLib/SmallDeformation/SmallDeformationProcessData.h
@@ -45,8 +45,8 @@ struct SmallDeformationProcessData
 
     std::unique_ptr<MaterialLib::Solids::MechanicsBase<DisplacementDim>>
         _material;
-    double dt;
-    double t;
+    double dt = 0;
+    double t = 0;
 };
 
 }  // namespace SmallDeformation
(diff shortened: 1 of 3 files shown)


norihiro-w:
:fast_forward:

ProcessLib/SmallDeformation/SmallDeformationProcessData.h
```diff
@@ -25,12 +25,12 @@ struct SmallDeformationProcessData
     SmallDeformationProcessData(
         std::unique_ptr<MaterialLib::Solids::MechanicsBase<DisplacementDim>>&&
             material)
-        : _material{std::move(material)}
+        : material{std::move(material)}
```
TomFischer: This seems to be not a problem, although I would expect that there is a problem with the same names. Just for curiosity: Can someone explain why this is not a problem?

TomFischer:
:+1:

ProcessLib/SmallDeformation/SmallDeformationProcessData.h
chleh: See http://stackoverflow.com/a/6185043

PR #1537, opened 2016-11-11
LIE/HM part 1: prerequisite

norihiro-w:
a part of #1535

This PR includes changes required for implementing HM process with the LIE approach. Please check each commit

PR #1538, opened 2016-11-11
LIE/HM part 2: Data structure

norihiro-w:
a part of #1535, depends on #1537 

This PR focuses on data structures used in LIE/HM, i.e. process data and integration point data


norihiro-w:
@wenqing can you please review this?

ProcessLib/LIE/HydroMechanics/LocalAssembler/IntegrationPointDataFracture.h
```diff
@@ -0,0 +1,79 @@
...
+          aperture(std::move(other.aperture)),
+          aperture0(std::move(other.aperture0)),
+          fracture_material(other.fracture_material),
+          C(std::move(other.C)),
+          integration_weight(std::move(other.integration_weight)),
+          darcy_velocity(std::move(other.darcy_velocity))
+    {
+    }
+
+    typename HMatricesType::HMatrixType H_u;
```
chleh: Note: Default-constructing these matrices might fail miserably if they are of dynamic size. Currently that probably doesn't matter, because displacement-related stuff is always fixed size.

norihiro-w: yes, it's fixed size matrix. but isn't Jenkins testing both Eigen dynamic size matrix and fixed size matrix cases?

chleh: I see, I was confused.
If you resize the matrices in your assembler, or assign to them as a whole, not component-wise, s.t. they have a proper size in the end, there will be no problem.

PR #1546, opened 2016-11-15
[CMake] fix the message to status in FindMKL

norihiro-w:
this avoid an unnecessary dialog in ccmake.


bilke:
Please change status to uppercase: `STATUS`. See https://cmake.org/cmake/help/v3.0/command/message.html.

Then ⏩

PR #1547, opened 2016-11-15
[App/CLI] set command line arguments from the environment variable

norihiro-w:
This PR suggests configuration of the command line arguments via the environmental variable `OGS6_CLI_DEFAULT_ARGS`.  I found it useful to specify some lengthy options which are not project dependent.

My use-case is 

```bash
export OGS6_CLI_DEFAULT_ARGS="-l debug --unbuffered-std-out"
ogs test1.prj &> test1.log
ogs test2.prj &> test2.log
```

### DIFF
--- a/Applications/CLI/ogs.cpp
+++ b/Applications/CLI/ogs.cpp
@@ -11,6 +11,9 @@
  */
 
 #include <chrono>
+#include <cstdlib>
+#include <string>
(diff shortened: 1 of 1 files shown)


norihiro-w:
indeed there is other way to achieve my purpose. if someone finds this useful, I can reopen it.

PR #1548, opened 2016-11-15
Fix computing sparsity pattern

norihiro-w:
`computeSparsityPatternNonPETSc()` is computing a wrong sparsity pattern if a process uses multiple variables with linear and quadratic order nodes, e.g. Hydromechanics.

### DIFF
--- a/NumLib/DOF/ComputeSparsityPattern.cpp
+++ b/NumLib/DOF/ComputeSparsityPattern.cpp
@@ -53,19 +53,11 @@ GlobalSparsityPattern computeSparsityPatternNonPETSc(
     // Map adjacent mesh nodes to "adjacent global indices".
     for (std::size_t n = 0; n < mesh.getNumberOfNodes(); ++n)
     {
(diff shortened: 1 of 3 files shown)


NumLib/DOF/ComputeSparsityPattern.cpp
```diff
@@ -53,19 +53,11 @@ GlobalSparsityPattern computeSparsityPatternNonPETSc(
...
-            auto const num_components = row_ids.size();
-            for (auto r : row_ids)
-            {
-                // Each component leads to an entry in the row.
-                // For the sparsity pattern only the number of entries are
-                // needed.
-                sparsity_pattern[r] += num_components;
-            }
-        }
+        unsigned n_connected_dof = 0;
```
wenqing: a question: do we only consider the case of  NumLib::ComponentOrder::BY_COMPONENT for HM?

wenqing: Thanks for the explanation.

PR #1549, opened 2016-11-15
speedup LocalToGlobalIndexMap construction

norihiro-w:
This PR suggests speeding up `LocalToGlobalIndexMap` construction by using `unordered_map` in `LocalToGlobalIndexMap::findGlobalIndices()`.


norihiro-w:
i tested only in debug mode. the bunary search is really slow in debug mode. maybe difference is not significant in release build

endJunction:
Indeed, that's a big difference in debug.

PR #1551, opened 2016-11-17
[Jenkins] generating Windows executable with static linking

norihiro-w:
WIP


norihiro-w:
@bilke I thought current Windows executables generated by Jenkins may not run on some computers because dynamic linking is specified. Is this correct or are the executables actually working on any computers? Unfortunately I don't have a test environment where MSVC is not installed. I tried to run on the UFZ server "raywin1" but got a strange error `Dieses Programm wurde durch eine Gruppenrichtlinie geblockt. `.

norihiro-w:
BTW Qt libraries are missing in the zip.

bilke:
@norihiro-w In general generated Windows packages should work on other Windows PCs. The Visual Studio Runtime is linked dynamically but its DLLs are included in the package `msvcp120.dll`, `msvcr120.dll`, ...
Qt is also linked dynamically because of licensing requirements but indeed the DLLs are missing.. I will check that..
All other libs are linked statically.

PR #1553, opened 2016-11-17
[CMake] Fixes packaging of DLLs (again).

bilke:
`make package` should package DLLs but this did not work: it is necessary to run `make package` at least twice. As an alternative shell scripts generating a complete package are provided. They can be found in the build directory and are invoked as follows:

Win:  `CALL package.cmd`
*nix: `bash package.sh`

TODO after merge:

- [ ] Update Jenkins pipeline to use the scripts


norihiro-w:
looks good

Issue #1084, opened 2016-03-07
Minimum GCC 4.9.0


endJunction:
Some major changes to the STL (regarding the c++11 standard) were implemented in the GCC version 4.9.0 and are not included in the currently required version 4.8.1.

The changes include updates of the STL containers' interfaces, _e.g._ the `const_iterator` for the `erase` function, but many more others. Unfortunately not reflected in their changelog, but see for example this [commit message](https://github.com/gcc-mirror/gcc/commit/6b0e18ca48bb4b4c01e7b5be2b98849943fdcf91) in github/gcc-mirror.

The changes do reflect the requirements of the c++11 standard; for example the `std::list::erase(iterator)` is **replaced** by `std::list::erase(const_iterator)`.

TODO:
- [x] Update travis.
- [ ] Update modules on envinf
- [ ] VTK (vtk-7.1.0 does not work with gcc-6.2.0/5.4.0 on envinf; linking errors referring to glibc-3.4.15.)
- [ ] Maybe PETSc, but I didn't check.

Not sure if all this recompilations are necessary on conan, but if yes this is the initial list:
- [ ] VTK
- [ ] Qt
- [ ] PETSc
- [ ] LIS
- [ ] Boost
- [ ] ...

norihiro-w:
Could you give us more details? What changes in the new version do we need to use?

chleh:
I'm skeptical, too. Don't we have more serious problems than the perfect const correctness in #1075?

TomFischer:
@chleh You are right, there are many places under construction in OGS-6 development. But why should we not use newer library versions to work on these open tasks that make the code better and programming more comfortable?

Is it the time effort that you afraid of or do you have other concerns? Arch linux can install a very new gcc version. I guess ubuntu users with installed new LTS ubuntu version (2016-04) should be able to install a newer gcc. On envinf1 we can load an even newer gcc module. Maybe only on envinf Qt, VTK and PETSc have to be recompiled once and maybe the WKDV can do this for us?

We have a similar situation with the DataExplorer. The DE version is based on an old version of Qt (4.8) while version 5.5.1 was released Oct 2015. When we decide to set the minimum gcc version to 4.9 (**or newer**) we should also discuss to switch to a newer Qt version. So the update work has to be done only once.

chleh:
It's simply the time effort. I agree, that @endJunction's intentions are good, but I'd give this particular issue a rather low priority. But that's only my personal opinion.

[...]

PR #1556, opened 2016-11-18
Ehlers Single surface yield function material model.

endJunction:
Initial copy from OGS-5 by Xing-Yuan Miao @xingyuanmiao.

### DIFF
--- a/Applications/CLI/Tests.cmake
+++ b/Applications/CLI/Tests.cmake
@@ -486,6 +486,47 @@ if(NOT OGS_USE_MPI)
         cube_1e3_expected_pcs_0_ts_101_t_1.000000.vtu cube_1e3_pcs_0_ts_101_t_1.000000.vtu displacement displacement
     )
 
+    # Mechanics; Small deformations, Ehlers (SDE)
+    AddTest(
+        NAME Mechanics_SDE_cube_1e0
+        PATH Mechanics/Ehlers
+        EXECUTABLE ogs
+        EXECUTABLE_ARGS cube_1e0.prj
+        TESTER vtkdiff
+        ABSTOL 1e-15 RELTOL 1e-13
+        DIFF_DATA
+        expected_cube_1e0_pcs_0_ts_101_t_2.550000.vtu cube_1e0_pcs_0_ts_101_t_2.550000.vtu displacement displacement
+        expected_cube_1e0_pcs_0_ts_101_t_2.550000.vtu cube_1e0_pcs_0_ts_101_t_2.550000.vtu sigma_xx sigma_xx
+        expected_cube_1e0_pcs_0_ts_203_t_5.100000.vtu cube_1e0_pcs_0_ts_203_t_5.100000.vtu displacement displacement
+        expected_cube_1e0_pcs_0_ts_203_t_5.100000.vtu cube_1e0_pcs_0_ts_203_t_5.100000.vtu sigma_xx sigma_xx
+    )
+    AddTest(
+        NAME Mechanics_SDE_cube_1e1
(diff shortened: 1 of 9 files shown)


endJunction:
@chleh Could you please have a look into travis' failing gcc jobs? Seems to me the gcc-4.8.5 is too old, because other gcc builds do well, though those are gcc-4.9.4. What do you think?

MaterialLib/SolidModels/Ehlers-impl.h
```diff
@@ -0,0 +1,666 @@
...
+ * Copyright (c) 2012-2016, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+/**
+ * Common convenitions for naming:
+ * x_D              - deviatoric part of x
```
wenqing: x --> tensor x

MaterialLib/SolidModels/Ehlers-impl.h
```diff
@@ -0,0 +1,666 @@
...
+/// Special product of v with itself: v \odot v
+/// v is given in Kelvin mapping.
+template <int DisplacementDim>
+ProcessLib::KelvinMatrixType<DisplacementDim> s_odot_s(
+    ProcessLib::KelvinVectorType<DisplacementDim> const& v)
+{
+    ProcessLib::KelvinMatrixType<DisplacementDim> result;
+
+    result(0, 0) = v(0) * v(0);
+    result(0, 1) = result(1, 0) = v(3) * v(3) / 2.;
```
wenqing: /2.0 --> * 0.5

endJunction: This makes no difference in the generated assembly.

MaterialLib/SolidModels/Ehlers.h
```diff
@@ -0,0 +1,213 @@
...
+
+        using KelvinVector = ProcessLib::KelvinVectorType<DisplacementDim>;
+
+        KelvinVector eps_p_D =
+            KelvinVector::Zero();  ///< deviatoric plastic strain
+        double eps_p_V = 0;    ///< volumetric strain
+        double eps_p_eff = 0;  ///< effective plastic strain
+
+        // Initial values from previous timestep
+        KelvinVector eps_p_D_prev = KelvinVector::Zero();  ///< \copydoc eps_p_D
```
endJunction: Must be a bug in the compiler, because your suggestion works.
I think this is another reason to go for #1084 ;)

Applications/CLI/Tests.cmake
```diff
@@ -486,6 +486,47 @@ if(NOT OGS_USE_MPI)
         cube_1e3_expected_pcs_0_ts_101_t_1.000000.vtu cube_1e3_pcs_0_ts_101_t_1.000000.vtu displacement displacement
     )
 
+    # Mechanics; Small deformations, Ehlers (SDE)
+    AddTest(
+        NAME Mechanics_SDE_cube_1e0
```
wenqing: Is there a test for Drucker-Prager model, which can be a simplification of Ehlers and which could be a good example for verification of the implementation.

nagelt: @wenqing good idea, @xingyuanmiao will add a DP model test in the next days. Please consider this as an additional treat, but not a reason to stall the PR. The full model is verified by the existing, more complex benchmark.

endJunction:
If no further comments, I'll merge this tonight.

PR #1561, opened 2016-11-21
Richards flow optimizations

endJunction:
A forgotten branch with small improvements.

### DIFF
--- a/MathLib/InterpolationAlgorithms/PiecewiseLinearInterpolation.cpp
+++ b/MathLib/InterpolationAlgorithms/PiecewiseLinearInterpolation.cpp
@@ -52,13 +52,18 @@ double PiecewiseLinearInterpolation::getValue(double pnt_to_interpolate) const
                                     pnt_to_interpolate));
     std::size_t const interval_idx = std::distance(_supp_pnts.begin(), it) - 1;
 
-    // compute linear interpolation polynom: y = m * (x - support[i]) + value[i]
-    const double m((_values_at_supp_pnts[interval_idx + 1] -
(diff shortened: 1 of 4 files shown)


MathLib/InterpolationAlgorithms/PiecewiseLinearInterpolation.cpp
```diff
@@ -52,13 +52,18 @@ double PiecewiseLinearInterpolation::getValue(double pnt_to_interpolate) const
...
-                    _values_at_supp_pnts[interval_idx]) /
-                   (_supp_pnts[interval_idx + 1] - _supp_pnts[interval_idx]));
+    const double m = (f_r - f) / (x_r - x);
 
-    return m * (pnt_to_interpolate - _supp_pnts[interval_idx]) +
-           _values_at_supp_pnts[interval_idx];
+    return m * (pnt_to_interpolate - x) + f;
 }
 
 double PiecewiseLinearInterpolation::getDerivative(
```
TomFischer: The name of the method is still very misleading. In the future we should move this functionality out of this class. But this is not your mistake and not the point in your changes.

wenqing: If agree, I would like to move this function to the derived class in PR #1529, and rename it to getGradient once this PR is merged..

TomFischer: It would be good if the implementation would be (re)moved from the class since it neither computes the derivative nor it computes the gradient of a piecewise linear function.

PR #1567, opened 2016-11-22
[Mat] FractureModel. Add explicit tpl declaration.

endJunction:
Avoids a warning by clang-3.9, details in the commit message.
I'm not sure if this is the correct way to fix it.


chleh:
:fast_forward:

PR #1568, opened 2016-11-22
When XML parse error occures, print good message.

endJunction:
the mismatched tags are easier to find this way ;)

### DIFF
--- a/scripts/doc/linked-xml-file.py
+++ b/scripts/doc/linked-xml-file.py
@@ -216,7 +216,12 @@ for (dirpath, dirnames, filenames) in os.walk(datadir, topdown=False):
 <tt>
 """ % (pagename, fn))
 
-                xmlroot = ET.parse(filepath).getroot()
+                try:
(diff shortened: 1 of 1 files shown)


scripts/doc/linked-xml-file.py
```diff
@@ -216,8 +216,13 @@ def dict_of_set_add(dos, key, value):
 <tt>
 """ % (pagename, fn))
 
-                xmlroot = ET.parse(filepath).getroot()
-                print_tags(xmlroot, 0, "prj", fh, None, 0, relfilepath)
+                try:
+                    xmlroot = ET.parse(filepath).getroot()
+                    print_tags(xmlroot, 0, "prj", fh, None, 0, relfilepath)
```
chleh: Minor: Could be after the `except` block.
Optional: Is it possible to also print the line number? (Or is it already printed by `print(err)`?)
:fast_forward:

endJunction: Moved after except block. Line is printed by `print(err)`

PR #1575, opened 2016-11-24
[DOCFL] Removed doxygen warnings in the documentation of fluid classes

wenqing:
as titled.


endJunction:
I'm not sure if it is intended but the verbatim environment creates a new line for every tag.
You might also consider only escaping the < and > like:
```c++
/// the \<type\> of ...
```
or use backticks for [code spans](https://www.stack.nl/~dimitri/doxygen/manual/markdown.html#md_codespan)
```c++
/// the `<type>` of ...
```

wenqing:
Replaced \verbatim with backticks

PR #1580, opened 2016-11-25
Doxygen Mathjax fix and warnings build status

bilke:
- Fixed Mathjax (formulas)
- Changed Jenkins to mark build as unstable (yellow) when there are new Doxygen warnings

### DIFF
--- a/Documentation/Doxyfile.in
+++ b/Documentation/Doxyfile.in
@@ -1487,7 +1487,7 @@ MATHJAX_FORMAT         = HTML-CSS
 # The default value is: http://cdn.mathjax.org/mathjax/latest.
 # This tag requires that the tag USE_MATHJAX is set to YES.
 
-MATHJAX_RELPATH        = http://cdn.mathjax.org/mathjax/latest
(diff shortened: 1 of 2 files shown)


chleh:
:fast_forward: 

But what will be the policy for new doxygen warnings? Will they have to be fixed immediately in the respective PR? Or later on in some Documentation-Specific PR? Furthermore, Doxygen warnings cover only part of the documentation. E.g., they don't cover the keyword docu. What will be the procedure for that in the future? I assume this PR was @endJunction's idea... :smile:

endJunction:
At some point (next week) there will be only a dozen doxygen warnings at all, then the "new warning" will be armed properly, i.e. the PR will not pass (or at least report something).
The keywords will be also covered because the prj file references will not be resolved if there is no prj....keyword.md.  In the following stages the keyword.md files will be filled with documentation.

bilke:
I noticed one problem:

A PR creates a new Jenkins job (`PR-xxxx`) but a new job has no previous builds and therefore the warnings parser cannot find out if there were new warnings introduced.. So this will only popup once the PR is merged into master..

PR #1593, opened 2016-11-30
Remove documentation pages not referenced in code.

endJunction:
These are files which are not referenced from the source code via any of the `\ogs_file_param{}` commands. Usually leftovers after renames and moves. Though it can indicate a wrong .md file name.


chleh:
:fast_forward:

PR #1596, opened 2016-12-01
Added fastpow that was implemented by Tom for OGS5

wenqing:
As titled.
This function,  fastpow, is needed by fluid property models from IAPWS.

### DIFF
--- a/MathLib/MathTools.h
+++ b/MathLib/MathTools.h
@@ -106,6 +106,29 @@ double sqrDist(const double* p0, const double* p1)
 }
 
 /**
+ * simple power function that takes as a second argument an integer instead of a
+ * float
+ * @param base basis of the expression
+ * @param exp exponent of the expression
+ * @return base^exp
+ */
+template <typename T>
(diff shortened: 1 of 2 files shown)


wenqing:
@TomFischer Could you please take a look?

MathLib/MathTools.h
```diff
@@ -120,6 +120,25 @@ inline double to_radians(double degrees) {
...
+ * @return base^exp
+ */
+template <typename T>
+inline T fastpow(T base, size_t exp)
+{
+	T result(base);
+	if (exp == 0)
+		result = static_cast<T>(1);
+	for (size_t k(1); k < exp; k++)
+		result *= base;
```
chleh: I think this function is not fast. Exponentiation by squaring is faster. Btw. there is a similar function inside `boost::math`

wenqing:
Ok. Will use boost::math::pow. If you think the new update with  squaring is Ok, I can also that.

PR #1601, opened 2016-12-05
Corrected the documentations for the fluid viscosity

wenqing:
As titled.

### DIFF
--- a/Documentation/ProjectFile/material/fluid/density/t_type.md
+++ b/Documentation/ProjectFile/material/fluid/density/t_type.md
@@ -1,1 +1,2 @@
-Type of density model
+Type of density model. It can be Constant, LiquidDensity, TemperatureDependent,
+or IdealGasLaw.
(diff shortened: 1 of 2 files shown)


MaterialLib/Fluid/Viscosity/createViscosityModel.cpp
```diff
@@ -32,7 +32,7 @@ namespace Fluid
 static std::unique_ptr<FluidProperty> createLinearPressureDependentViscosity(
```
wenqing: The first character of this file name and that in createDensityModel.cpp will be changed to its capital one, "C".

MaterialLib/Fluid/Viscosity/createViscosityModel.cpp
```diff
@@ -32,7 +32,7 @@ namespace Fluid
...
     BaseLib::ConfigTree const& config)
 {
-    //! \ogs_file_param{material__fluid__viscosity__type}
+    //! \ogs_file_param{material__fluid__viscosity__LinearPressure}
```
endJunction: In ogs_file_param{xxx} the xxx refers to the first argument to the checkConfigParameter() call. So in this case the original version with '__type' is correct. Only the subsequent references in ogs_file_param include the specific case , e.g. LinearPressure.

endJunction:
Also have a look on the generated doxygen documentation. The issues shown below should not be added.
image

wenqing:
@endJunction  After our discussion, I modified this PR to documentation the unreferenced input parameters of viscosity types and density types.

PR #1604, opened 2016-12-08
[HT] Add thermodispersion.

TomFischer:


ProcessLib/HT/CreateHTProcess.cpp
```diff
@@ -115,6 +115,25 @@ std::unique_ptr<Process> createHTProcess(
     DBUG("Use \'%s\' as specific_heat_capacity_fluid parameter.",
          specific_heat_capacity_fluid.name.c_str());
 
+
+    // Parameter for the thermal conductivity of the solid (only one scalar per
+    // element, i.e., the isotropic case is handled at the moment)
+    auto& thermal_dispersivity_longitudinal = findParameter<double>(
```
wenqing: If thermal_conductivity is input as matrix,  thermal dispersivity data can be input as  thermal_conductivity directly. Or you keep the  thermal dispersivity input but add the values to thermal_conductivity matrix data directly after parsing the input.

TomFischer: @wenqing 
Maybe I misunderstood you. As far as I understand the thermo dispersion tensor consists of a conductive and a dispersive part. The dispersive part depends on the darcy velocitiy and for this reason can not computed directly after parsing the input. The two new parameters introduced in the PR are used for the computation of the dispersive part.

wenqing: I see that is a dispersion conductivity. Then the thermal conductivity is anisotropic. That means that  the conductivity parameter for the Laplacian has to be set as matrix.

ProcessLib/HT/CreateHTProcess.cpp
```diff
@@ -115,6 +115,25 @@ std::unique_ptr<Process> createHTProcess(
...
+        config,
+        //! \ogs_file_param_special{process__HT__thermal_dispersivity_longitudinal}
```
endJunction: must be 'prj__processes__process__HT_...'

ProcessLib/HT/HTFEM.h
```diff
@@ -110,6 +110,10 @@ class LocalAssemblerData : public HTLocalAssemblerInterface
             Eigen::Map<const Eigen::VectorXd>(&local_x[num_nodes], num_nodes);
 
         auto const & b = _process_data.specific_body_force.head(GlobalDim);
+        Eigen::MatrixXd unit_mat(GlobalDim, GlobalDim);
+        unit_mat.setZero();
+        for (std::size_t k(0); k < GlobalDim; ++k)
+            unit_mat(k,k) = 1.0;
```
endJunction: Actually eigen does provide the identity; have a look on the http://eigen.tuxfamily.org/dox/group__QuickRefPage.html page and search for  Identity there.

ProcessLib/HT/CreateHTProcess.cpp
wenqing: Solved after discussion.

PR #1605, opened 2016-12-08
IAPWS viscosity model of water

wenqing:
as titled.  

Next:
- IAPWS density model of water
- class FluidProperties, which consists models of individual fluid properties and which only takes  primary variable as arguments to get property values.

### DIFF
--- a/MaterialLib/Fluid/FluidProperty.h
+++ b/MaterialLib/Fluid/FluidProperty.h
@@ -25,6 +25,7 @@ enum class PropertyVariableType
 {
     T = 0,                   ///< temperature.
     p = 1,                   ///< pressure.
+    rho = p,                 ///< density. For some models, rho substitutes p
     number_of_variables = 2  ///< Number of property variables.
 };
 
--- a/MaterialLib/Fluid/Viscosity/WaterViscosityIAPWS.cpp
+++ b/MaterialLib/Fluid/Viscosity/WaterViscosityIAPWS.cpp
@@ -18,15 +18,24 @@ namespace MaterialLib
 {
 namespace Fluid
 {
-const double WaterViscosityIAPWS::_hi[4] = {1.67752, 2.20462, 0.6366564,
-                                            -0.241605};
-const double WaterViscosityIAPWS::_hij[6][7] = {
+static const double Hi[4] = {1.67752, 2.20462, 0.6366564, -0.241605};
+static const double Hij[6][7] = {
     {0.520094, 0.222531, -0.281378, 0.161913, -0.0325372, 0, 0},
     {0.0850895, 0.999115, -0.906851, 0.257399, 0, 0, 0},
     {-1.08374, 1.88797, -0.772479, 0, 0, 0, 0},
     {-0.289555, 1.26613, -0.489837, 0, 0.0698452, 0, -0.00435673},
     {0, 0, -0.25704, 0, 0, 0.00872102, 0},
     {0, 0.120573, 0, 0, 0, 0, -0.000593264}};
+
+static double computeBarMu0Factor(const double barT);
+
+static double series_factorT[6];
+static double series_factorRho[7];
+static void computeSeriesFactorForMu1(const double barT, const double bar_rho);
+static double computeBarMu1Factor();
+
+static double computedBarMu_dbarT(const double barT, double bar_rho);
+static double computedBarMu_dbarRho(const double barT, double bar_rho);
 
 double WaterViscosityIAPWS::getValue(const ArrayType& var_vals) const
 {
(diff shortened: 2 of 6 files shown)


MaterialLib/Fluid/FluidProperty.h
```diff
@@ -25,6 +25,7 @@ enum class PropertyVariableType
 {
     T = 0,                   ///< temperature.
     p = 1,                   ///< pressure.
+    rho = p,                 ///< density. For some models, rho substitutes p
     number_of_variables = 2  ///< Number of property variables.
```
TomFischer: Do we need to increment the number of variables, also?

wenqing: No. The variables for models are (T, p, C..) or (T, rho, C), or other (T, other fluid property). 

Models with different variable will be handled inside class FluidProperties (in the coming PR), in which  variables for external access are primary variables (T, p, c, ...) .

wenqing: When C is considered, the number of variables will be increased.

Tests/MaterialLib/TestFluidViscosity.cpp
```diff
@@ -124,3 +126,48 @@ TEST(Material, checkVogelViscosity)
...
+    const auto mu_w = createTestViscosityModel(xml_w);
+    const double T[] = {298.15, 298.15, 373.15,  433.15,  433.15, 873.15,
+                        873.15, 873.15, 1173.15, 1173.15, 1173.15};
+    const double rho[] = {998, 1200, 1000, 1, 1000, 1, 100, 600, 1, 100, 400};
+
+    const double expected_mumu[] = {
+        889.735100, 1437.649467, 307.883622, 14.538324, 217.685358, 32.619287,
+        35.802262,  77.430195,   44.217245,  47.640433, 64.154608};
+
+    const double purterbation = 1.e-9;
```
TomFischer: Spelling 'pertubation'

Tests/MaterialLib/TestFluidViscosity.cpp
```diff
@@ -124,3 +126,48 @@ TEST(Material, checkVogelViscosity)
...
+    ArrayType vars;
+    for (int i = 0; i < 11; i++)
+    {
+        // Test mu
+        vars[static_cast<unsigned>(PropertyVariableType::T)] = T[i];
+        vars[static_cast<unsigned>(PropertyVariableType::rho)] = rho[i];
+        const double mu = mu_w->getValue(vars);
+        ASSERT_NEAR(expected_mumu[i] * 1.e-6, mu, 1.e-9);
```
TomFischer: You could also multiply all values of the array `expected_mumu` by '1e-6' during its initialization in lines 141-143.

Tests/MaterialLib/TestFluidViscosity.cpp
TomFischer: Could you explain why `pertubation` has exectly this value of '1e-9'?

wenqing: It is used to calculate the numerical differentiation of mu with respect to  rho or T. Compared to the scale of rho and T, 1e-9 is small enough.

Tests/MaterialLib/TestFluidViscosity.cpp
wenqing: Just added comment about the data, which are for the given test in http://www.iapws.org/relguide/visc.pdf. Therefore, I would like to keep the original data.

MaterialLib/Fluid/Viscosity/WaterViscosityIAPWS.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+#include <cmath>
+
+namespace MaterialLib
+{
+namespace Fluid
+{
+const double WaterViscosityIAPWS::_hi[4] = {1.67752, 2.20462, 0.6366564,
+                                            -0.241605};
+const double WaterViscosityIAPWS::_hij[6][7] = {
```
endJunction: It seems the both arrays are only needed in the cpp file. If so, there is no need to add them to the class definition, just make them static in the cpp.

wenqing: According the concept of OOP, the constants should be the members of the class because they are the data of the this model. However, since this class is not inheritable, moving the constants to the cpp does not break the data encapsulation and it makes the head file concise as well. Therefore I changed the implematation according to your suggestion. Thanks.

MaterialLib/Fluid/Viscosity/WaterViscosityIAPWS.cpp
```diff
@@ -0,0 +1,178 @@
...
+    {0.0850895, 0.999115, -0.906851, 0.257399, 0, 0, 0},
+    {-1.08374, 1.88797, -0.772479, 0, 0, 0, 0},
+    {-0.289555, 1.26613, -0.489837, 0, 0.0698452, 0, -0.00435673},
+    {0, 0, -0.25704, 0, 0, 0.00872102, 0},
+    {0, 0.120573, 0, 0, 0, 0, -0.000593264}};
+
+static double computeBarMu0Factor(const double barT);
+
+static double series_factorT[6];
+static double series_factorRho[7];
```
wenqing: removed these two static buffer arrays.

endJunction:
:fast_forward: 
@TomFischer If no further comments, set it to "rebase and merge."

PR #1610, opened 2016-12-12
[CMake] Added optional parameter REQUIREMENTS to AddTest().

bilke:
Handles optional simple boolean expression which has to be true to
enable the test, e.g.

```cmake
AddTest(
  ..
  REQUIREMENTS OGS_USE_PETSC AND (OGS_USE_EIGEN OR OGS_USE_LIS)
)
```


chleh:
:+1: nice.
What will be the first test that needs such a distinction?

endJunction:
First test case: The current SD/LIE and HM/LIE benchmarks run only with Eigen; there is no LIS solver good enough for this problems.

PR #1611, opened 2016-12-12
Move local NewtonRaphson from the Material\Solid to NumLib

Yonghui56:
As titled, the local NewtonRaphson is moved from MaterialLib/SolidModels to NumLib in order for further reusing. 
This class provides a basic NewtonRaphson routine to handle the nonlinear constitutive relationships existing in the local problem. After being moved to NumLib , it can be reused by other processes, e.g the two-phase multicomponent process which I am working on now.
(p.s. Some of the changes are caused by running clang-format)

### DIFF
--- a/NumLib/NewtonRaphson.h
+++ b/NumLib/NewtonRaphson.h
@@ -7,17 +7,15 @@
  *
  */
 
-#ifndef MATERIALLIB_SOLIDMODELS_NEWTONRAPHSON_H_
-#define MATERIALLIB_SOLIDMODELS_NEWTONRAPHSON_H_
+#ifndef NUMERICAL_NEWTONRAPHSON_H_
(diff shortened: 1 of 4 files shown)


endJunction:
@Yonghui56 Please tell how it will be reused and from which process

NumLib/NewtonRaphson.h
```diff
@@ -7,17 +7,15 @@
  *
  */
 
-#ifndef MATERIALLIB_SOLIDMODELS_NEWTONRAPHSON_H_
-#define MATERIALLIB_SOLIDMODELS_NEWTONRAPHSON_H_
+#ifndef NUMERICAL_NEWTONRAPHSON_H_
```
endJunction: Please use NUMLIB_NEWTONRAPHSON_H_

PR #1616, opened 2016-12-15
fluid properties for specific heat capacity, thermal conductivity, density, and viscosity models. 

wenqing:
This PR presents
-  fluid heat capacity and thermal conductivity model creators together with constant models.
-  a base class, FluidProperties,  to encapsulate fluid models of density, viscosity, specific heat capacity and thermal conductivity. With the class, the variables for the models are the primary variables of PDE (T, p, ...) only. 
-  class PrimaryVariableDependentFluidProperties, which is derived from FluidProperties and which is for  models with the primary variables as variables.
-  class PrimaryVariableDependentFluidProperties, which is also derived from FluidProperties but  for  models with density as one variable. It is for the models by  the IAPWS.
- a creator for FluidProperties.
- corresponding unit tests.


MaterialLib/Fluid/FluidProperties/CreateFluidProperties.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+#include <string>
+
+#include "BaseLib/ConfigTree.h"
+
+#include "FluidProperties.h"
+#include "PrimaryVariableDependentFluidProperties.h"
+#include "FluidPropertiesWithDensityDependentModels.h"
+
+#include "MaterialLib/Fluid/FluidPropertyHeaders.h"
```
endJunction: The "prefixed" includes are usually placed before the local includes.

MaterialLib/Fluid/FluidProperties/CreateFluidProperties.cpp
```diff
@@ -0,0 +1,100 @@
...
+    auto const& rho_conf = config.getConfigSubtree("density");
+    auto liquid_density = MaterialLib::Fluid::createFluidDensityModel(rho_conf);
+
+    //! \ogs_file_param{material__fluid__viscosity}
+    auto const& mu_conf = config.getConfigSubtree("viscosity");
+    auto viscosity = MaterialLib::Fluid::createViscosityModel(mu_conf);
+    const bool is_mu_density_dependent =
+        (viscosity->getName().find("density dependent") != std::string::npos)
+            ? true
+            : false;
```
endJunction: The result of a comparison is already a boolean; no need for ?: operator

MaterialLib/Fluid/FluidProperties/FluidProperties.h
```diff
@@ -0,0 +1,99 @@
...
+        std::unique_ptr<MaterialLib::Fluid::FluidProperty>&& heat_capacity,
+        std::unique_ptr<MaterialLib::Fluid::FluidProperty>&&
+            thermal_conductivity)
+        : _property_models{{std::move(density), std::move(viscosity),
+                            std::move(heat_capacity),
+                            std::move(thermal_conductivity)}}
+    {
+    }
+
+    virtual ~FluidProperties(){};
```
endJunction: `virtual ~FluidProperties() = default;` is also possible.

MaterialLib/Fluid/FluidProperties/FluidPropertiesWithDensityDependentModels.h
```diff
@@ -0,0 +1,87 @@
...
+     *                         concentration, which is defined in enum class
+     *                         PropertyVariableType.
+     *  \param variable_type   Variable type
+     */
+    double getdValue(const FluidPropertyType property_type,
+                     const ArrayType& variable_values,
+                     const PropertyVariableType variable_type) const override;
+
+private:
+    mutable double _density_value = 0.;
```
endJunction: `mutable` doesn't play well in multithreaded environments. Is it possible to avoid it?

MaterialLib/Fluid/FluidProperties/CreateFluidProperties.cpp
wenqing: Changed.

MaterialLib/Fluid/FluidProperties/FluidProperties.h
wenqing: Changed accordingly.

MaterialLib/Fluid/FluidProperties/FluidPropertiesWithDensityDependentModels.cpp
```diff
@@ -0,0 +1,142 @@
...
+    const double density_value, const FluidPropertyType property_type,
+    const ArrayType& variable_values) const
+{
+    const auto& fluid_density_model =
+        _property_models[static_cast<unsigned>(FluidPropertyType::Density)];
+    const double drho_dT = fluid_density_model->getdValue(
+        variable_values, PropertyVariableType::T);
+    ArrayType var_vals = variable_values;
+    var_vals[static_cast<unsigned>(PropertyVariableType::rho)] = density_value;
+    // return d()/dT + d ()/drho * drho/dT
```
wenqing: Moved it to the right place.

MaterialLib/Fluid/FluidProperties/FluidPropertiesWithDensityDependentModels.h
wenqing: changed the variable type to std::atomic<double>. In order not  or break the unique interface, ``_density_value`` is used as a buffer to store the calculated density value, which can be an argument  of  other fluid property model.

endJunction:
@wenqing I'm not sure if `mutable std::atomic<T>` is good design in this case. I have a slightly different solution [here](https://github.com/endJunction/ogs/commit/b890983b07e7aceb421aefaa491d123292d17c4a). Let's discuss some options later.

MaterialLib/Fluid/FluidProperties/FluidPropertiesWithDensityDependentModels.cpp
```diff
@@ -34,7 +34,7 @@ FluidPropertiesWithDensityDependentModels::
     : FluidProperties(std::move(density), std::move(viscosity),
                       std::move(heat_capacity),
                       std::move(thermal_conductivity)),
-      _is_density_depedent{{true, is_viscosity_density_dependent,
+      _is_density_depedent{{false, is_viscosity_density_dependent,
```
TomFischer: spelling: `_is_density_dependent`

endJunction:
This needs a rebase due to changes in the test system.

MaterialLib/Fluid/FluidProperties/FluidPropertiesWithDensityDependentModels.cpp
wenqing: Corrected by amending a commit.

wenqing:
@bilke Does jenkins have any problem?

endJunction:
Jenkins seems to be fine. You'll need to fix the doxygen warnings.

wenqing:
@endJunction Thanks. Nice feature of doxygen warnings.

MaterialLib/Fluid/FluidProperties/FluidPropertiesWithDensityDependentModels.h
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * \file   FluidPropertiesWithDensityDependentModels.h
+ *
+ * Created on November 29, 2016, 4:18 PM
+ */
+
+#ifndef OGS_FLUID_PROPERTIES_WITH_DENSITY_DEPENDENT_MODELS_H
+#define OGS_FLUID_PROPERTIES_WITH_DENSITY_DEPENDENT_MODELS_H
+
+#include <atomic>
```
endJunction: This is no longer needed, isn't it?

endJunction:
Please update the copyright years to 2012-2017.
I'll review this later today for other issues.

MaterialLib/Fluid/FluidProperties/FluidPropertiesWithDensityDependentModels.h
wenqing: removed.

MaterialLib/Fluid/FluidProperties/FluidProperties.h
```diff
@@ -0,0 +1,111 @@
...
+public:
+    typedef std::array<double, PropertyVariableNumber> ArrayType;
+
+    FluidProperties(
+        std::unique_ptr<MaterialLib::Fluid::FluidProperty>&& density,
+        std::unique_ptr<MaterialLib::Fluid::FluidProperty>&& viscosity,
+        std::unique_ptr<MaterialLib::Fluid::FluidProperty>&& heat_capacity,
+        std::unique_ptr<MaterialLib::Fluid::FluidProperty>&&
+            thermal_conductivity)
+    /* Will use this if MS visual studio compiler supports
```
endJunction: I'd at least state the version of the compiler and the problems associated with this code. If possible, also version which is working.
Better in my opinion would be to make an ifdef around the workaround with compiler versions:
```
#if defined(_MSC_VER) && _MSC_VER < 1700
...
```

MaterialLib/Fluid/FluidProperties/FluidPropertiesWithDensityDependentModels.cpp
```diff
@@ -0,0 +1,127 @@
...
+            return compute_df_drho_drho_dp(property_type, variable_values);
+        }
+    }
+    else
+    {
+        return _property_models[static_cast<unsigned>(property_type)]
+            ->getdValue(variable_values, variable_type);
+    }
+
+    return 0.;
```
endJunction: This return is only to silence a compiler warning? If not, please remove the unreachable code.

MaterialLib/Fluid/PropertyVariableType.h
```diff
@@ -0,0 +1,34 @@
...
+namespace MaterialLib
+{
+namespace Fluid
+{
+/// Variable that determine the property.
+enum class PropertyVariableType
+{
+    T = 0,                   ///< temperature.
+    p = 1,                   ///< pressure.
+    rho = p,                 ///< density. For some models, rho substitutes p
```
endJunction: It seems to me this is excluding material models where both rho and p are used, isn't it?
Anyway relying on actual enumeration numbers is not a good idea. To be discussed.

endJunction:
If you like you can switch to `#pragma once` in the new headers before merging. See #1627.

MaterialLib/Fluid/FluidProperties/FluidProperties.h
wenqing: The latest MSVC compiler still does not support list initialization inside member initializer list or non-static data member initializer. Therefore I added ifdef MSVC for that.

MaterialLib/Fluid/FluidProperties/FluidPropertiesWithDensityDependentModels.cpp
wenqing: removed.

MaterialLib/Fluid/PropertyVariableType.h
wenqing: Yes, cases of rho and p are excluded.
In EOS,  p can be calculated via rho, T, and  vice versa. As far as I knew, there is not any model that has rho and p as variables.

MaterialLib/Fluid/FluidProperties/FluidProperties.h
```diff
@@ -48,11 +48,16 @@ class FluidProperties
         std::unique_ptr<MaterialLib::Fluid::FluidProperty>&& heat_capacity,
         std::unique_ptr<MaterialLib::Fluid::FluidProperty>&&
             thermal_conductivity)
-    /* Will use this if MS visual studio compiler supports
-    : _property_models{{std::move(density), std::move(viscosity),
-                        std::move(heat_capacity),
-                        std::move(thermal_conductivity)}}
-    */
+#ifndef MSVC
+        // Up to the latest Visual Studio compiler, list initialization inside
```
endJunction: Still, could you add the exact version for which you tested this, please?

wenqing: Added Visual Studio 2015. I did not test with the lasted msvc because I do not have it. However, I did it according to the the document on MSDN
 https://msdn.microsoft.com/en-us/library/dn793970.aspx

MaterialLib/Fluid/FluidProperties/CreateFluidProperties.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+    bool is_KT_density_dependent = false;
+    std::unique_ptr<MaterialLib::Fluid::FluidProperty> thermal_conductivity =
+        nullptr;
+    auto const& thermal_conductivity_opt_conf =
+        //! \ogs_file_param{material__fluid__thermal_conductivity}
+        config.getConfigSubtreeOptional("thermal_conductivity");
+    if (thermal_conductivity_opt_conf)
+    {
+        auto& thermal_conductivity_conf = *thermal_conductivity_opt_conf;
```
TomFischer: Why not `auto const &`?

Tests/MaterialLib/TestFluidSpecificHeatCapacityModel.cpp
```diff
@@ -0,0 +1,51 @@
...
+using ArrayType = MaterialLib::Fluid::FluidProperty::ArrayType;
+
+std::unique_ptr<FluidProperty> createSpecificFluidHeatCapacityModel(
+    const char xml[])
+{
+    auto const ptree = readXml(xml);
+    BaseLib::ConfigTree conf(ptree, "", BaseLib::ConfigTree::onerror,
+                             BaseLib::ConfigTree::onwarning);
+    auto const& sub_config = conf.getConfigSubtree("specific_heat_capacity");
+    return MaterialLib::Fluid::createSpecificFluidHeatCapacityModel(sub_config);
```
TomFischer: Namespaces `MaterialLib::Fluid` in front are not needed since the namespaces are made available in line 26.

wenqing: removed by amending a commit. Also removed it in TestFluidSpecificThermalConductivity.cpp

MaterialLib/Fluid/FluidProperties/CreateFluidProperties.cpp
wenqing: added by amending a commit.

MaterialLib/Fluid/FluidProperties/FluidPropertiesWithDensityDependentModels.cpp
wenqing: Added  return 0.; again due to gnu compiler warning.

PR #1617, opened 2016-12-15
[FL] Fixed two compilation warnings in WaterViscosityIAPWS.cpp

wenqing:
as titled.


wenqing:
Only Configure (Win) got problem but it is not related to this PR.

PR #1619, opened 2016-12-15
Update CTests

endJunction:
- Using REQUIREMENTS keyword of the AddTest cmake macro to distinguish between test working (or not working) with lis/petsc.
 - Move individual process' tests into corresponding folders.
 - Small fixes.


Applications/CLI/Tests.cmake
```diff
@@ -2,7 +2,6 @@
 add_test(NAME ogs_no_args COMMAND ogs)
 set_tests_properties(ogs_no_args PROPERTIES WILL_FAIL TRUE)
 
-if(NOT OGS_USE_MPI)
     # CUBE 1x1x1 GROUNDWATER FLOW TESTS
     foreach(mesh_size 1e0 1e1 1e2 1e3)
```
bilke: After removing the `if`-clause you should also adapt the indentation of the former content of the clauses.

endJunction: That's right. I have split the whole content in the last commit.

endJunction:
I'd like to merge it today. Any further comments?

PR #1623, opened 2017-01-02
Cleanup output messages

TomFischer:
The PR unifiies output message format at some locations, inserts spaces to make the output better parseable for evaluation scripts and removes redundant linear solver time output messages.


MeshLib/IO/MPI_IO/NodePartitionedMeshReader.cpp
```diff
@@ -79,18 +79,18 @@ MeshLib::NodePartitionedMesh* NodePartitionedMeshReader::read(
...
     else
     {
-        INFO("-->Reading binary mesh file ...");
+        INFO("Reading binary mesh file ...");
 
         mesh = readBinary(file_name_base);
     }
 
-    INFO("\t\n>>Total elapsed time in reading mesh:%f s\n", timer.elapsed());
+    INFO("Total elapsed time in reading mesh: %f s.", timer.elapsed());
```
chleh: Other outputs of time measurements are formatted as `INFO("[time] ...`. I think this occurence here should be formatted the same way. Afterwards 👍

PR #1624, opened 2017-01-04
Raise the VTK version number from 6.3.0 to 7.0.0.

TomFischer:


endJunction:
Why?

TomFischer:
The DKRZ cluster MISTRAL has not installed a vtk lib. So the OGS build system downloads vtk and tries to compile vtk 6.3.0 with gcc version 6.2 (from the MISTRAL module system). But the compilation of vtk fails. I hope this error will vanish with the newer vtk version.

bilke:
I just found out that VTK 7.0 does not compile with gcc 6.2! VTK 7.1 [added support](https://gitlab.kitware.com/vtk/vtk/commit/5c5930d3f63cd5afaaf3bed3b866c19e7eb2a4b7) for it but with 7.1 not all ctests run yet... 🔥 dependency hell 🔥

TomFischer:
Thanks @bilke for testing.

PR #1626, opened 2017-01-09
Update copyright years to 2012-2017.

endJunction:
Automatic substitution...


chleh:
1277 lines changed but only 1274 files. So there must be some files with duplicate years. Is that correct? I don't want to scroll through all the files.
If OK ⏩

endJunction:
These are coming from the output messages in 
```
> ag 'Copyright \(c\) 2012-' -c | grep -v ':1'                            
Applications/DataExplorer/DataView/LicenseDialog.cpp:2
Applications/Utils/SimpleMeshCreation/generateStructuredMesh.cpp:2
Applications/CLI/ogs.cpp:2
```

chleh:
OK, thanks.

bilke:
⏩

PR #1631, opened 2017-01-12
Apply FluidProperties class to LiquidFlow

wenqing:
- Apply FluidProperties class to LiquidFlow
- Added a creator for LiquidFlowMaterialProperties

Note: This is the 2nd PR for the staggered scheme implementation.

### DIFF
--- a/ProcessLib/LiquidFlow/CreateLiquidFlowMaterialProperties.cpp
+++ b/ProcessLib/LiquidFlow/CreateLiquidFlowMaterialProperties.cpp
@@ -1,0 +1,94 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2017, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ * \file   CreateLiquidFlowMaterialProperties.cpp
+ *
+ *   Created on December 14, 2016, 1:20 PM
+ */
+
+#include "CreateLiquidFlowMaterialProperties.h"
+
+#include "BaseLib/ConfigTree.h"
+#include "BaseLib/reorderVector.h"
+
+#include "MeshLib/PropertyVector.h"
+
+#include "MaterialLib/Fluid/FluidProperty.h"
+#include "MaterialLib/PorousMedium/Porosity/Porosity.h"
+#include "MaterialLib/PorousMedium/Storage/Storage.h"
+#include "MaterialLib/Fluid/FluidProperties/CreateFluidProperties.h"
+
+#include "MaterialLib/Fluid/FluidPropertyHeaders.h"
+#include "MaterialLib/PorousMedium/PorousPropertyHeaders.h"
+
+#include "LiquidFlowMaterialProperties.h"
+
+namespace ProcessLib
+{
+namespace LiquidFlow
+{
+class LiquidFlowMaterialProperties;
+
+std::unique_ptr<LiquidFlowMaterialProperties>
+createLiquidFlowMaterialProperties(
+    BaseLib::ConfigTree const& config, bool const has_material_ids,
+    MeshLib::PropertyVector<int> const& material_ids)
+{
+    DBUG("Reading material properties of liquid flow process.");
+
+    //! \ogs_file_param{prj__processes__process__LIQUID_FLOW__material_property__fluid}
+    auto const& fluid_config = config.getConfigSubtree("fluid");
+    auto fluid_properties =
+        MaterialLib::Fluid::createFluidProperties(fluid_config);
+
+    // Get porous properties
+    std::vector<Eigen::MatrixXd> intrinsic_permeability_models;
+    std::vector<std::unique_ptr<MaterialLib::PorousMedium::Porosity>>
+        porosity_models;
+    std::vector<std::unique_ptr<MaterialLib::PorousMedium::Storage>>
+        storage_models;
+
+    std::vector<int> mat_ids;
+    //! \ogs_file_param{prj__processes__process__LIQUID_FLOW__material_property__porous_medium}
+    auto const& poro_config = config.getConfigSubtree("porous_medium");
+    //! \ogs_file_param{prj__processes__process__LIQUID_FLOW__material_property__porous_medium__porous_medium}
+    for (auto const& conf : poro_config.getConfigSubtreeList("porous_medium"))
+    {
+        //! \ogs_file_attr{prj__processes__process__LIQUID_FLOW__material_property__porous_medium__porous_medium__id}
+        auto const id = conf.getConfigAttributeOptional<int>("id");
+        mat_ids.push_back(*id);
+
+        //! \ogs_file_param{prj__processes__process__LIQUID_FLOW__material_property__porous_medium__porous_medium__permeability}
+        auto const& perm_conf = conf.getConfigSubtree("permeability");
+        intrinsic_permeability_models.emplace_back(
+            MaterialLib::PorousMedium::createPermeabilityModel(perm_conf));
+
+        //! \ogs_file_param{prj__processes__process__LIQUID_FLOW__material_property__porous_medium__porous_medium__porosity}
+        auto const& poro_conf = conf.getConfigSubtree("porosity");
+        auto n = MaterialLib::PorousMedium::createPorosityModel(poro_conf);
+        porosity_models.emplace_back(std::move(n));
+
+        //! \ogs_file_param{prj__processes__process__LIQUID_FLOW__material_property__porous_medium__porous_medium__storage}
+        auto const& stora_conf = conf.getConfigSubtree("storage");
+        auto beta = MaterialLib::PorousMedium::createStorageModel(stora_conf);
+        storage_models.emplace_back(std::move(beta));
+    }
+
+    BaseLib::reorderVector(intrinsic_permeability_models, mat_ids);
+    BaseLib::reorderVector(porosity_models, mat_ids);
+    BaseLib::reorderVector(storage_models, mat_ids);
+
+    return std::unique_ptr<LiquidFlowMaterialProperties>(
+        new LiquidFlowMaterialProperties(
+            std::move(fluid_properties),
+            std::move(intrinsic_permeability_models),
+            std::move(porosity_models), std::move(storage_models),
+            has_material_ids, material_ids));
+}
+
+}  // end of namespace
+}  // end of namespace
--- a/ProcessLib/LiquidFlow/CreateLiquidFlowMaterialProperties.cpp
+++ b/ProcessLib/LiquidFlow/CreateLiquidFlowMaterialProperties.cpp
@@ -59,8 +59,8 @@ createLiquidFlowMaterialProperties(
     for (auto const& conf : poro_config.getConfigSubtreeList("porous_medium"))
     {
         //! \ogs_file_attr{prj__processes__process__LIQUID_FLOW__material_property__porous_medium__porous_medium__id}
-        auto const id = conf.getConfigAttributeOptional<int>("id");
-        mat_ids.push_back(*id);
+        auto const id = conf.getConfigAttribute<int>("id");
+        mat_ids.push_back(id);
 
         //! \ogs_file_param{prj__processes__process__LIQUID_FLOW__material_property__porous_medium__porous_medium__permeability}
         auto const& perm_conf = conf.getConfigSubtree("permeability");
--- a/ProcessLib/LiquidFlow/CreateLiquidFlowMaterialProperties.cpp
+++ b/ProcessLib/LiquidFlow/CreateLiquidFlowMaterialProperties.cpp
@@ -54,27 +54,35 @@ createLiquidFlowMaterialProperties(
 
     std::vector<int> mat_ids;
     //! \ogs_file_param{prj__processes__process__LIQUID_FLOW__material_property__porous_medium}
-    auto const& poro_config = config.getConfigSubtree("porous_medium");
+    auto const& porous_medium_configs =
+        config.getConfigSubtree("porous_medium");
     //! \ogs_file_param{prj__processes__process__LIQUID_FLOW__material_property__porous_medium__porous_medium}
-    for (auto const& conf : poro_config.getConfigSubtreeList("porous_medium"))
+    for (auto const& porous_medium_config :
+         porous_medium_configs.getConfigSubtreeList("porous_medium"))
     {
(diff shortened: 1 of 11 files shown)


ProcessLib/LiquidFlow/CreateLiquidFlowMaterialProperties.cpp
```diff
@@ -0,0 +1,94 @@
...
+        auto n = MaterialLib::PorousMedium::createPorosityModel(poro_conf);
+        porosity_models.emplace_back(std::move(n));
+
+        //! \ogs_file_param{prj__processes__process__LIQUID_FLOW__material_property__porous_medium__porous_medium__storage}
+        auto const& stora_conf = conf.getConfigSubtree("storage");
+        auto beta = MaterialLib::PorousMedium::createStorageModel(stora_conf);
+        storage_models.emplace_back(std::move(beta));
+    }
+
+    BaseLib::reorderVector(intrinsic_permeability_models, mat_ids);
```
TomFischer: I don't understand why the `intrinsic_permeability` (and `porosity_models` as well as `storage_models`) have to be reordered. Please, can you explain it with a small example what is done here?

ProcessLib/LiquidFlow/CreateLiquidFlowMaterialProperties.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+    std::vector<int> mat_ids;
+    //! \ogs_file_param{prj__processes__process__LIQUID_FLOW__material_property__porous_medium}
+    auto const& poro_config = config.getConfigSubtree("porous_medium");
+    //! \ogs_file_param{prj__processes__process__LIQUID_FLOW__material_property__porous_medium__porous_medium}
+    for (auto const& conf : poro_config.getConfigSubtreeList("porous_medium"))
+    {
+        //! \ogs_file_attr{prj__processes__process__LIQUID_FLOW__material_property__porous_medium__porous_medium__id}
+        auto const id = conf.getConfigAttributeOptional<int>("id");
+        mat_ids.push_back(*id);
```
TomFischer: The `id` seems to be an optional parameter. What is the purpose of this parameter and is it okay to insert `id` into the `mat_ids` when `id` is optional?

ProcessLib/LiquidFlow/CreateLiquidFlowMaterialProperties.cpp
```diff
@@ -0,0 +1,94 @@
...
+#include "LiquidFlowMaterialProperties.h"
+
+namespace ProcessLib
+{
+namespace LiquidFlow
+{
+class LiquidFlowMaterialProperties;
+
+std::unique_ptr<LiquidFlowMaterialProperties>
+createLiquidFlowMaterialProperties(
```
TomFischer: The documentations for the parameters read in this function is missing. Probably the documentation will make the review a bit easier.

ProcessLib/LiquidFlow/LiquidFlowMaterialProperties.cpp
```diff
@@ -107,7 +61,8 @@ double LiquidFlowMaterialProperties::getViscosity(const double p,
     ArrayType vars;
     vars[static_cast<int>(MaterialLib::Fluid::PropertyVariableType::T)] = T;
     vars[static_cast<int>(MaterialLib::Fluid::PropertyVariableType::p)] = p;
-    return _viscosity->getValue(vars);
+    return _fluid_properties->getValue(
+        MaterialLib::Fluid::FluidPropertyType::Vicosity, vars);
```
TomFischer: Spelling: `Vicosity` should be `Viscosity`.

ProcessLib/LiquidFlow/LiquidFlowMaterialProperties.h
```diff
@@ -44,9 +53,23 @@ class LiquidFlowMaterialProperties
     typedef MaterialLib::Fluid::FluidProperty::ArrayType ArrayType;
 
     LiquidFlowMaterialProperties(
```
TomFischer: Is it correct to say: `LiquidFlowMaterialProperties` is the collection of material properties special to the `LiquidFlow` process? Maybe you can add some documentation?

ProcessLib/LiquidFlow/LiquidFlowMaterialProperties.h
```diff
@@ -79,10 +102,12 @@ class LiquidFlowMaterialProperties
 
     double getViscosity(const double p, const double T) const;
```
TomFischer: I thought that this class only collects porous media material properties and `LiquidFlowMaterialProperties` is responsible for viscosity and density. So probably my understanding is not correct. So my maby stupid question: Why is the viscosity in this class?

ProcessLib/LiquidFlow/CreateLiquidFlowMaterialProperties.cpp
wenqing: Removed the ``optional``. This parameter is the material ID, which is read from vtu mesh file.

ProcessLib/LiquidFlow/LiquidFlowProcess.h
```diff
@@ -12,12 +12,17 @@
 
 #pragma once
 
+#include <memory>
+#include <memory>
```
TomFischer: You want to be really sure to `#include <memory>`? 😃

ProcessLib/LiquidFlow/LiquidFlowProcess.h
```diff
@@ -72,6 +77,14 @@ class LiquidFlowProcess final : public Process
                                           GlobalVector const& x) override;
 
     bool isLinear() const override { return true; }
+
+    std::vector<MaterialLib::Fluid::FluidProperties*> getFluidProperties() const
+    {
+        std::vector<MaterialLib::Fluid::FluidProperties*> material_properties =
+            {_material_properties->getFluidProperties()};
```
TomFischer: Verz minor: Why are the curly braces needed here?

ProcessLib/LiquidFlow/LiquidFlowProcess.h
```diff
@@ -88,7 +101,7 @@ class LiquidFlowProcess final : public Process
 
     const int _gravitational_axis_id;
     const double _gravitational_acceleration;
-    LiquidFlowMaterialProperties _material_properties;
+    const std::unique_ptr<LiquidFlowMaterialProperties> _material_properties;
```
TomFischer: Why not a `const reference`? If it would be a `const&` maybe the implementation of `getFluidProperties()` above can return a `const& FluidProperties` and put it outside of the class into a `std::vector`?

ProcessLib/LiquidFlow/LiquidFlowProcess.h
TomFischer: I could not find the place where `getFluidProperties()` is used? Why is it necessary to return a `std::vector<MaterialLib::Fluid::FluidProperties*>` and not only something like `MaterialLib::Fluid::FluidProperties const&`?

ProcessLib/LiquidFlow/CreateLiquidFlowMaterialProperties.cpp
wenqing: Here is an example:
```
     <porous_medium>
          <porous_medium  id="0">
                <permeability>
                     ...
                 </permeability>
                 <porosity>
                    ...
                 </porosity>
                 <storage>
                   ...
                  </storage>
            </porous_medium>
          <porous_medium  id="1">
                <permeability>
                     ...
                 </permeability>
                 <porosity>
                    ...
                 </porosity>
                 <storage>
                   ...
                  </storage>
            </porous_medium>
            ...
        </porous_medium>
```
After parsing, we can not assure that parameters under ``id="N"`` are pushed to the ``N`` th element of the corresponding vector. Therefore a sorting is performed according to the IDs that are read from the tag attribution.

ProcessLib/LiquidFlow/CreateLiquidFlowMaterialProperties.cpp
wenqing: The parameters in this function are documented (see Documentation/prj/processes/process/LIQUID_FLOW). The parameters of material models are given in the creating functions of the models.

Actually, the body of this function is exactly moved from the constructor of class LiquidFlowMaterialProperties.

ProcessLib/LiquidFlow/LiquidFlowMaterialProperties.cpp
wenqing: Crrected.

ProcessLib/LiquidFlow/CreateLiquidFlowMaterialProperties.cpp
TomFischer: Thank you for the explanation. Please add this as an explanation/comment  in the code.

ProcessLib/LiquidFlow/LiquidFlowMaterialProperties.h
wenqing: It is somehow a forward call to compute the viscosity. However, the arguments are p and T directly instead of an array of variable for this specific flow process.

Besides, I added a sentence of documentation about this class as
```
/** This class has a collection of material properties of liquid flow process
 *  and it provides the computation of the coefficients of the PDE of liquid
 *  flow process.
 */
```

ProcessLib/LiquidFlow/LiquidFlowProcess.h
wenqing: :smile: Be afraid of memory shortage.:smile:. Removed one.

ProcessLib/LiquidFlow/LiquidFlowProcess.h
wenqing: With that curly braces, I want to return a vector with one entry.

Initially, I want to have this function for all flow processes for coupling computation. Now I changed the mind, and made this function be unique only for LiquidFlow and return only a pointer to  class FluidProperties.

ProcessLib/LiquidFlow/LiquidFlowProcess.h
wenqing: As explained above,  getFluidProperties() is now changed to return a pointer. Therefore, I prefer to keep this declaration as a concrete instance. While the local assembler of LiquidFlow takes  a reference to this unique_ptr member.

ProcessLib/LiquidFlow/LiquidFlowMaterialProperties.h
wenqing: Added a documentation about the class.

ProcessLib/LiquidFlow/CreateLiquidFlowMaterialProperties.cpp
TomFischer: Thanks for explanation.

ProcessLib/LiquidFlow/LiquidFlowProcess.h
TomFischer: Okay, but why is it necessary to return the pointer in a vector?

wenqing: Originally it is returned in a vector in case to have a same return turn type with multiphase flow process, which can have more than one fluids. Now function is changed to
```
    MaterialLib::Fluid::FluidProperties* getFluidProperties() const
    {
        return _material_properties->getFluidProperties();
    }
```
for the LiquidFlow only.

TomFischer: Thanks, now I got it.

wenqing:
@endJunction  Two commits according to your comments and my improvement.
- Commit  fcdeac7: Use full variable names in createLiquidFlowMaterialProperties.
- Commit 	1656b03: Changed the type of a member from unique_ptr reference to class reference
                               and removed a member, _current_material_id for thread safe.

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler.h
```diff
@@ -61,21 +62,23 @@ class LiquidFlowLocalAssembler : public LiquidFlowLocalAssemblerInterface
...
-                             double const gravitational_acceleration,
-                             LiquidFlowMaterialProperties& material_propertries)
+    LiquidFlowLocalAssembler(
+        MeshLib::Element const& element,
+        std::size_t const /*local_matrix_size*/,
+        bool const is_axially_symmetric,
+        unsigned const integration_order,
+        int const gravitational_axis_id,
+        double const gravitational_acceleration,
+        std::unique_ptr<LiquidFlowMaterialProperties> const&
```
endJunction: Could be simply `LiquidFlowMaterialProperties const&`...

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler.h
```diff
@@ -61,21 +62,23 @@ class LiquidFlowLocalAssembler : public LiquidFlowLocalAssemblerInterface
...
+            material_propertries)
         : _element(element),
           _integration_method(integration_order),
           _shape_matrices(initShapeMatrices<ShapeFunction, ShapeMatricesType,
                                             IntegrationMethod, GlobalDim>(
               element, is_axially_symmetric, _integration_method)),
           _gravitational_axis_id(gravitational_axis_id),
           _gravitational_acceleration(gravitational_acceleration),
-          _material_properties(material_propertries)
+          _material_properties(*(material_propertries.get()))
```
endJunction: I just mentioned this somewhere else: `get` is not necessary because `material_properties` is a pointer, so simply dereferencing it is good.

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler.h
wenqing: Changed. Accordingly, ``_material_properties`` in line 65 of LiquidFlowProcess.cpp is changed to ``*_material_properties``.

 See: https://github.com/ufz/ogs/pull/1631/commits/76dd96bcdf43d60b79d3f313f612d6a8c6f52f15#diff-97a96abf0d3aa134d63d918100ee67b7R65

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler.h
wenqing: Changed back to ``_material_properties(material_propertries)`` because the argument type ``std::unique_ptr<LiquidFlowMaterialProperties> const&`` has been changed to ``LiquidFlowMaterialProperties const&`` according to you last comment.

PS: I made these two changes by amending the last commit.

endJunction:
:fast_forward:

PR #1673, opened 2017-01-20
[CMake] OGS_EIGEN_DYNAMIC_SHAPE_MATRICES defaults to OFF on Release.

bilke:
The performance differences between static and dynamic matrices are not obvious and some accidently use dynamic matrices when doing performance measurements (using release compilation modes).


endJunction:
When tests are green, this one is good to be merged.

PR #1678, opened 2017-01-24
Fixing Lubby2 initialization of local variables.

endJunction:
See second commit's message for some details.

Small restructuring: Moving the nested class MaterialProperties outside (Lubby2MaterialProperties) and introducing a new class to save temporary variables.


xingyuanmiao:
Very clear.

nagelt:
Clearer differentiation between state and parameters. Good work. Minor comments above.

PR #1684, opened 2017-01-26
Fixing eigen alignment

endJunction:
TODO: 
 - [x] Fix compilation problems.

Starting with Eigen-3.3 the alignment issue came up again and is solved here.
Without the proper alignment in Release mode there are segv's and undefined memory usage problems.

I tried to minimize set of changes, so not every structure got a special new-operator (we might have to extend this if errors happen).

I have changed some of the tests (and found some errors there) to run successfully on my computer with Eigen-3.2.10 and Eigen-3.3.2 and LIS-1.7.24.
~~Lars will setup a test in Jenkins using eigen-3.3 in nearest future.~~ See https://github.com/ufz/ogs/pull/1687 for the eigen-3.3 changes.

One problem I could not solve is the failing LIS-BiCG solver for the `ogs-Mechanics_SDE_cube_1e0` benchmark. I changed the solver to CG, which is fine in this case because the matrices are symmetric.

### DIFF
--- a/ProcessLib/BoundaryCondition/NeumannBoundaryConditionLocalAssembler.h
+++ b/ProcessLib/BoundaryCondition/NeumannBoundaryConditionLocalAssembler.h
@@ -69,6 +69,9 @@ public:
 private:
     Parameter<double> const& _neumann_bc_parameter;
     typename Base::NodalVectorType _local_rhs;
+
+public:
+    EIGEN_MAKE_ALIGNED_OPERATOR_NEW;
 };
 
 }   // namespace ProcessLib
--- a/ProcessLib/BoundaryCondition/GenericNaturalBoundaryConditionLocalAssembler.h
+++ b/ProcessLib/BoundaryCondition/GenericNaturalBoundaryConditionLocalAssembler.h
@@ -48,7 +48,10 @@ public:
 
 protected:
     IntegrationMethod const _integration_method;
-    std::vector<typename ShapeMatricesType::ShapeMatrices> const _shape_matrices;
+    std::vector<typename ShapeMatricesType::ShapeMatrices,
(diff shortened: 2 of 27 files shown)


endJunction:
@bilke Can we set the minimum version of Eigen to 3.2.9? I have a commit ["Set minimum version Eigen-3.2.9."](https://github.com/ufz/ogs/pull/1684/commits/90c78eff5c5a9217fd5e506d7cd548586de57da6) as preparation, but the CI environments needs an update.

ProcessLib/BoundaryCondition/NeumannBoundaryConditionLocalAssembler.h
```diff
@@ -69,6 +69,9 @@ class NeumannBoundaryConditionLocalAssembler final
 private:
     Parameter<double> const& _neumann_bc_parameter;
     typename Base::NodalVectorType _local_rhs;
+
+public:
+    EIGEN_MAKE_ALIGNED_OPERATOR_NEW;
```
wenqing: Where is it defined?

endJunction: See this link https://eigen.tuxfamily.org/dox-devel/group__TopicStructHavingEigenMembers.html for details and it is defined in Core/util/Memory.h in Eigen.

bilke:
👍

ProcessLib/BoundaryCondition/NeumannBoundaryConditionLocalAssembler.h
wenqing: I see. May be it is good to have a comment here.

PR #1701, opened 2017-02-01
[Bugfix] Fixed error in mesh node to raster interpolation

rinkk:


rinkk:
I don't think it does. I had an external request yesterday, calling this to my attention.

PR #1703, opened 2017-02-02
Add FieldData to mesh properties

endJunction:
This is a preparation to output of secondary variables, especially of integration-point-data and of internal material model states.

In general any data (available to the local assemblers) can be written as a MeshLib::Property on the new `MeshItemType` `IntegrationPoint`. (Maybe IntegrationPoint is not a good name though.)
The mesh properties stored as `IntegrationPoint` type are written to vtk's FieldData.

Tests for creating/reading the mesh property and a vtk import/export *i.e.* roundtrip are implemented.


bilke:
👍

chleh:
:+1:

PR #1705, opened 2017-02-07
Fix error in TwoPhaseFlowWithPrho process

Yonghui56:
This pr includes:
Move IntegrationPointData struct into namespace.
Fix a wrong debug message.


ProcessLib/TwoPhaseFlowWithPrho/TwoPhaseFlowWithPrhoLocalAssembler.h
```diff
@@ -21,39 +21,40 @@
...
+
 namespace ProcessLib
 {
 namespace TwoPhaseFlowWithPrho
 {
+    template <typename NodalMatrixType>
+    struct IntegrationPointData final
+    {
+        explicit IntegrationPointData(
+            ProcessLib::TwoPhaseFlowWithPrho::
```
TomFischer: Is the qualification `ProcessLib::TwoPhaseFlowWithPrho::` necessary? If not, I would suggest to remove it since this makes the code more readable in my point of view.

PR #1707, opened 2017-02-09
[MaL] EigenMatrix: set/add values even if 0.

endJunction:
Under certain circumstances the sparsity pattern will be destroyed and
adding/setting values would then result in an expensive reallocation.

Assume that in the first global iteration only one half of the values is
set to non-zero. Before the linear solver is called the sparse matrix is
compressed thus reducing the allocated space in each row. In the
subsequent iterations, which might add or set the other half of the
matrix values reallocations would be necessary.

~~I'll update the PR with explicit performance measurements later today. For a specific test (phase-field) the speedup is around 6.5 times (for every time-step). Of course there is some overhead for other processes because of additional `coeffRef() = value` calls; I'll quantify that later today.~~
See detailed benchmark details below in the comment.

### DIFF
--- a/MathLib/LinAlg/Eigen/EigenMatrix.h
+++ b/MathLib/LinAlg/Eigen/EigenMatrix.h
@@ -71,7 +71,7 @@ public:
     int setValue(IndexType row, IndexType col, double val)
     {
         assert(row < (IndexType) getNumberOfRows() && col < (IndexType) getNumberOfColumns());
-        if (val != 0.0) _mat.coeffRef(row, col) = val;
+        _mat.coeffRef(row, col) = val;
         return 0;
     }
 
@@ -79,7 +79,7 @@ public:
     /// inserted.
     int add(IndexType row, IndexType col, double val)
     {
-        if (val != 0.0) _mat.coeffRef(row, col) += val;
+        _mat.coeffRef(row, col) += val;
         return 0;
     }
 


endJunction:
Benchmark and (number runs) | This PR's version | original with `if (0) do not set` | t_old / t_new
------------ | ------------- | ------------ | ------------
PhaseField/square_open_line.prj' (30) | 1.564375289 s +-  0.78% | 11.233134249 s +-  0.34% | 7.1805
Mechanics/Linear/ring_plane_strain.prj' (30) | 1.538794099 s +-  1.06% |1.524961765 s +-  0.77% | 0.9910
HydroMechanics/Linear/square_1e2.prj' (30) | 4.938046186 s +-  0.28% | 5.139376288 s +-  0.24% |1.041
Parabolic/TES/2D/tes-inert-wedge.prj' (3) | 81.258264770 s +-  1.51% | 90.015583056 s +-  0.72% | 1.108
Elliptic/GWF/cube_1e6_neumann.prj' (3) | 43.305907215 s +-  0.14% | 43.576099636 s +-  0.23% |1.006

clang-3.9 -O3 -march=native , Fixed size matrices, static linking, Eigen-3.3.2 (with the changes from https://github.com/ufz/ogs/pull/1684)

bilke:
👍

PR #1716, opened 2017-02-14
[Bug] Set return 0. for not drho/dT in LinearTemperatureDependentDensity

wenqing:
There is a bug in ``LinearTemperatureDependentDensity->getdValue(..)``, which does not consider the cases of the derivative of the density with a non-temperature variable.  This PR fixes this bug.


wenqing:
@TomFischer  If you use this model for computation, please update.

endJunction:
@TomFischer Merge, when ok.

Issue #1726, opened 2017-02-21
[USER] untraceable error message


waltherm:
I do not know how to trace an error message I get from a failing ogs run. I searched in doxygen.opengeosys.org and docs.opengeosys.org without a hit getting returned. Similar setups in Tests/Data are not being tested on Jenkins and do not have this property in a project or vtu file. 

Can anyone tell me how to trace this error?

The error message is:
```error: A property with the specified name "MaterialIDs" is not available.```

The full output is:
```info: This is OpenGeoSys-6 version 6.0.8-332-gb152302.
info: OGS started on 2017-02-21 10:37:33+0100.
info: ConstantParameter: kappa
info: ConstantParameter: rho_solid
info: ConstantParameter: lambda_solid
info: ConstantParameter: lambda_fluid
info: ConstantParameter: c_solid
info: ConstantParameter: c_fluid
info: ConstantParameter: alpha_l
info: ConstantParameter: alpha_t
info: ConstantParameter: c0
info: ConstantParameter: conc
info: ConstantParameter: p0
info: ConstantParameter: pDirichletIN
info: ConstantParameter: pDirichletOUT
info: No boundary conditions found.
error: A property with the specified name "MaterialIDs" is not available.
error: ConfigTree: In file `dispersion.prj' at path <processes/process>: Key <fluid_reference_density> has not been found at ConfigTree.cpp, line 229
info: OGS terminated on 2017-02-21 10:37:33+010
info: [time] Execution took 0.0744271 s.
```

bilke:
Marc is talking about the benchmark [LARGE_2D_ThermalConvection_constviscosity](https://github.com/ufz/ogs/blob/master/ProcessLib/HT/CMakeLists.txt#L2). Please also note that this benchmark is never run by Jenkins because

- it is a large benchmark
- large benchmark have the option `OGS_USE_LIS=ON`
- lis is an exclusion in this benchmark: `REQUIREMENTS NOT (OGS_USE_LIS OR OGS_USE_MPI)`

How to proceed?

- is the `NOT OGS_USE_LIS` requirement necessary?
- should there be a new Jenkins job testing this configuration

TomFischer:
The project file of the benchmark [LARGE_2D_ThermalConvection_constviscosity](https://github.com/ufz/ogs/blob/master/ProcessLib/HT/CMakeLists.txt#L2) specifies 'SparseLU' as the solver for the system of linear equations. So 'LIS' is not necessary for solving the benchmark.

[...]

PR #1734, opened 2017-03-01
Change Properties

TomFischer:
PR contains
- The method `Properties::getPropertyVector` is more strict and doesn't return a `nullptr`, it crashes instead.
- The new method `Properties::existsPropertyVector` checks if a `PropertyVector` with a given name **and** a given data type is available.
- Occurrence of `getPropertyVector` are substituted by a check with `existsPropertyVector` and a fetch with `getPropertyVector`.
- Some checks aren't necessary anymore since they are done within the `getPropertyVector` implementation.

This fixes some details of issue #1726.

### DIFF
--- a/Applications/Utils/MeshEdit/queryMesh.cpp
+++ b/Applications/Utils/MeshEdit/queryMesh.cpp
@@ -48,8 +48,6 @@ int main(int argc, char *argv[])
     if (!mesh)
         return EXIT_FAILURE;
 
-    auto materialIds = mesh->getProperties().getPropertyVector<int>("MaterialIDs");
-
     std::vector<std::size_t> selected_node_ids;
     if  (showNodeWithMaxEle_arg.getValue())
     {
@@ -63,6 +61,10 @@ int main(int argc, char *argv[])
     }
     selected_node_ids.insert(selected_node_ids.end(), nodeId_arg.getValue().begin(), nodeId_arg.getValue().end());
 
+    MeshLib::PropertyVector<int> const*const materialIds =
+        mesh->getProperties().existsPropertyVector<int>("MaterialIDs")
+            ? mesh->getProperties().getPropertyVector<int>("MaterialIDs")
+            : nullptr;
     for (auto ele_id : eleId_arg.getValue())
     {
         std::stringstream out;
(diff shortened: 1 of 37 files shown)


MeshLib/Properties-impl.h
```diff
@@ -75,18 +75,42 @@ PropertyVector<T>* Properties::createNewPropertyVector(
...
+    if (it == _properties.end())
+    {
+        return false;
+    }
+    // Check that the PropertyVector has the correct data type.
+    if (!dynamic_cast<PropertyVector<T> const*>(it->second))
+    {
+        return false;
+    }
+    return true;
```
endJunction: Optional:
```c++
return dynamic_cast<PV<T> const*>(it->second) != nullptr;
```

Applications/Utils/MeshEdit/queryMesh.cpp
```diff
@@ -63,6 +61,12 @@ int main(int argc, char *argv[])
     }
     selected_node_ids.insert(selected_node_ids.end(), nodeId_arg.getValue().begin(), nodeId_arg.getValue().end());
 
+    MeshLib::PropertyVector<int>* materialIds(nullptr);
```
wenqing: How about
```
MeshLib::PropertyVector<int> const* materialIds = mesh->getProperties().existsPropertyVector<int>("MaterialIDs") ?
mesh->getProperties().getPropertyVector<int>("MaterialIDs") :  nullptr;
```

wenqing: For some other similar places.

endJunction: I was close to suggest it too, especially because of `const` but the readability degrades IMO.

TomFischer: Maybe a proper formating will make it readable. I will check this.

MeshLib/Properties-impl.h
TomFischer: Will change this.

bilke:
@TomFischer Jenkins is green.

Applications/Utils/MeshEdit/queryMesh.cpp
TomFischer: Please take a look at commit 8fa8d03c6761363a159e6cbfaf46d1a3b0031561 and judge yourself about the readability. I like Wenqings suggestion more than my previous implementation.

endJunction:
:fast_forward:

PR #1739, opened 2017-03-07
Cleanup: Clang format, fix spelling, remove debug code, ...

TomFischer:
For details see commit messages.

### DIFF
--- a/Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
+++ b/Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
@@ -518,29 +518,6 @@ void NodeWiseMeshPartitioner::writeASCII(const std::string& file_name_base)
     writeNodesASCII(file_name_base);
 }
 
-void NodeWiseMeshPartitioner::resetGlobalNodeIndices()
-{
-    for (std::size_t i = 0; i < _mesh->_nodes.size(); i++)
-    {
-        _mesh->_nodes[i]->setID(_nodes_global_ids[i]);
-    }
-    // sort
-    std::sort(_mesh->_nodes.begin(), _mesh->_nodes.end(),
-              [](const MeshLib::Node* a, const MeshLib::Node* b) {
-                  return a->getID() < b->getID();
-              });
-}
-
-void NodeWiseMeshPartitioner::writeGlobalMeshVTU(
-    const std::string& file_name_base)
-{
-    resetGlobalNodeIndices();
-    MeshLib::IO::VtuInterface writer(_mesh.get());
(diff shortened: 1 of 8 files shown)


Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
```diff
@@ -518,29 +519,6 @@ void NodeWiseMeshPartitioner::writeASCII(const std::string& file_name_base)
     writeNodesASCII(file_name_base);
 }
 
-void NodeWiseMeshPartitioner::resetGlobalNodeIndices()
-{
-    for (std::size_t i = 0; i < _mesh->_nodes.size(); i++)
-    {
-        _mesh->_nodes[i]->setID(_nodes_global_ids[i]);
```
wenqing: If you removed these two members, please move Node::setID() to the private section as it was originally there before.

Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
```diff
@@ -518,29 +519,6 @@ void NodeWiseMeshPartitioner::writeASCII(const std::string& file_name_base)
...
-    }
-    // sort
-    std::sort(_mesh->_nodes.begin(), _mesh->_nodes.end(),
-              [](const MeshLib::Node* a, const MeshLib::Node* b) {
-                  return a->getID() < b->getID();
-              });
-}
-
-void NodeWiseMeshPartitioner::writeGlobalMeshVTU(
```
wenqing: Sometimes it is useful to have a global mesh for checking the correction of the node indexing after partitioning. It is Ok to remove it, the global mesh will be generated in a post processing.

Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
TomFischer: Commit e38cb1b makes `Node::setID()` and all the `protected` attributes and method now `privat`.

Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
TomFischer: In the future we'll need a post processing tool that converts the binary format to the vtu or pvtu. I think such a tool would be the proper place for the removed code.

wenqing: Good.

wenqing:
:fast_forward:

PR #1740, opened 2017-03-09
[PL] Add output of darcy velocity to RichardsFlow process.

TomFischer:


ProcessLib/RichardsFlow/RichardsFlowFEM.h
```diff
@@ -96,6 +108,11 @@ class LocalAssemblerData : public RichardsFlowLocalAssemblerInterface
             _integration_method.getNumberOfPoints();
         SpatialPosition pos;
         pos.setElementID(_element.getID());
+
```
wenqing: With this implementation, the velocity is always from the previous time step.

For the correct calculation, you need to
1)  put the calculation of the velocity by overloading this member of LocalAssember
https://github.com/ufz/ogs/blob/master/ProcessLib/LocalAssemblerInterface.h#L69
2)  overload the Process:computeSecondaryVariableConcrete

TomFischer: Thanks for the hint. I will take a look.

TomFischer: @wenqing I changed the code according to your suggestions. Would be nice if you could have a look again.

ProcessLib/RichardsFlow/RichardsFlowFEM.h
```diff
@@ -159,6 +172,57 @@ class LocalAssemblerData : public RichardsFlowLocalAssemblerInterface
...
+            double P_int_pt = 0.0;
+            NumLib::shapeFunctionInterpolate(local_x, sm.N, P_int_pt);
+            double const Pc = -P_int_pt;
+            double const Sw = interpolated_Pc.getValue(Pc);
+            double const k_rel = interpolated_Kr.getValue(Sw);
+            double const K_mat_coeff = K * k_rel / mu;
+
+            auto const num_nodes = ShapeFunction::NPOINTS;
+            auto p_nodal_values =
```
wenqing: Remove line 202 and use ShapeFunction::NPOINTS directly as
```
const auto  p_nodal_values =  Eigen::Map<const NodalVectorType>(&local_x[0], ShapeFunction::NPOINTS);
```
Besides, move it to a place just ahead of  the  loop of `` for (unsigned ip = 0; ...)``

ProcessLib/RichardsFlow/RichardsFlowFEM.h
```diff
@@ -159,6 +172,57 @@ class LocalAssemblerData : public RichardsFlowLocalAssemblerInterface
...
+                -K_mat_coeff * sm.dNdx * p_nodal_values;
+            if (_process_data.has_gravity)
+            {
+                auto const rho_w = _process_data.water_density(t, pos)[0];
+                auto const body_force =
+                    _process_data.specific_body_force(t, pos);
+                assert(body_force.size() == GlobalDim);
+                auto const b = MathLib::toVector<GlobalDimVectorType>(
+                    body_force, GlobalDim);
+                velocity += K_mat_coeff * rho_w * b;
```
wenqing: If added a positive, the body force element data must be input as negative as:

(0, -1), (0, 0,-1) or (0, -1, 0), etc.

ProcessLib/RichardsFlow/RichardsFlowFEM.h
```diff
@@ -159,6 +172,57 @@ class LocalAssemblerData : public RichardsFlowLocalAssemblerInterface
...
+                Eigen::Map<const NodalVectorType>(&local_x[0], num_nodes);
+
+            GlobalDimVectorType velocity =
+                -K_mat_coeff * sm.dNdx * p_nodal_values;
+            if (_process_data.has_gravity)
+            {
+                auto const rho_w = _process_data.water_density(t, pos)[0];
+                auto const body_force =
+                    _process_data.specific_body_force(t, pos);
+                assert(body_force.size() == GlobalDim);
```
wenqing: lines 213-215 can be moved out side of the integration loop.

ProcessLib/RichardsFlow/RichardsFlowProcess.cpp
```diff
@@ -82,5 +105,15 @@ void RichardsFlowProcess::assembleWithJacobianConcreteProcess(
         dx_dx, M, K, b, Jac, coupling_term);
 }
 
+void RichardsFlowProcess::computeSecondaryVariableConcrete(
+    double const t, GlobalVector const& x,
+    StaggeredCouplingTerm const& coupling_term)
+{
+    DBUG("Compute the darcy velocity for RichardsFlowProcess");
```
wenqing: darcy  --> Darcy

ProcessLib/RichardsFlow/RichardsFlowFEM.h
wenqing: A comment for this is needed.

ProcessLib/RichardsFlow/RichardsFlowFEM.h
TomFischer: Is it allowed to access `_process_data.specific_body_force` if the process hasn't gravity?

wenqing: Better not allow it. I see that the problem is in that assert. 
Since the vector is a constant vector, its initialization and  assertion can be performed only once.
 
There is already a solution for that:
1. In RichardsFlowProcessData, replace ``Parameter<double> const& specific_body_force;`` with `` Eigen::VectorXd const _specific_body_force; ``
  (ref. https://github.com/ufz/ogs/blob/master/ProcessLib/TwoPhaseFlowWithPP/TwoPhaseFlowWithPPProcessData.h#L69)
2. make assert of specific_body_force only once in the parsing function of  the process
(ref. https://github.com/ufz/ogs/blob/master/ProcessLib/TwoPhaseFlowWithPP/CreateTwoPhaseFlowWithPPProcess.cpp#L63)

With that modifications, you can use  _process_data.specific_body_force here directly without calling assert and MathLib::toVector.

Due to these modifications, you have to modify the corresponding part in assemble(...) as well:
see: https://github.com/TomFischer/ogs/blob/79f7e5dd3eb8e200aa785680056c22704bb256f6/ProcessLib/RichardsFlow/RichardsFlowFEM.h#L154

 If you like, such modifications could be put to another PR.

TomFischer: Since the `RichardsFlowProcess` doesn't include the new material properties implementation until now I would shift this to a PR that uses the new material classes.

PR #1745, opened 2017-03-10
[HPC] Partition PropertyVectors and read partitioned PropertyVectors

TomFischer:
The PR adds the partitioning computed by metis also to the `PropertVector`s in the mesh and stores the meta data as well as the property values in separate binary files. The current PR contains the partitioning only for `PropertyVector`'s assigned to mesh nodes. The PR also includes the reading of the `PropertyVector`'s into a parallel running ogs.

### DIFF
--- a/Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
+++ b/Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
@@ -326,6 +326,12 @@ void NodeWiseMeshPartitioner::writePropertiesBinary(
         }
         MeshLib::IO::writePropertyVectorMetaDataBinary(out, pvmd);
     }
+    for (const auto& partition : _partitions)
+    {
+        MeshLib::IO::PropertyVectorPartitionMetaData pvpmd;
+        pvpmd.number_of_tuples = partition.number_of_non_ghost_nodes;
+        MeshLib::IO::writePropertyVectorPartitionMetaData(out, pvpmd);
+    }
     out.close();
 }
 
@@ -349,6 +355,17 @@ void NodeWiseMeshPartitioner::readPropertiesConfigDataBinary(
             INFO("readPropertiesConfigMetaDataBinary:");
             MeshLib::IO::writePropertyVectorMetaData(*pvmd);
         }
+    }
+    auto pos = is.tellg();
+    for (std::size_t i(0); i < _npartitions; ++i) {
+        auto offset =
+            pos + static_cast<std::streampos>(
+                      i * sizeof(MeshLib::IO::PropertyVectorPartitionMetaData));
+        is.seekg(offset);
+        unsigned long number_of_tuples = 0;
+        is.read(reinterpret_cast<char*>(&number_of_tuples),
+                sizeof(unsigned long));
+        INFO("%u tuples in partition %u.", number_of_tuples, i);
     }
 }
 
--- a/Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
+++ b/Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
@@ -336,6 +336,10 @@ NodeWiseMeshPartitioner::getNumberOfIntegerVariablesOfElements(
 void NodeWiseMeshPartitioner::writePropertiesBinary(
     const std::string& file_name_base) const
 {
+    auto const& properties(_partitioned_properties);
+    auto const& property_names(properties.getPropertyVectorNames());
+    if (property_names.empty())
+        return;
     const std::string fname_cfg = file_name_base +
                                   "_partitioned_properties_cfg" +
                                   std::to_string(_npartitions) + ".bin";
@@ -346,8 +350,6 @@ void NodeWiseMeshPartitioner::writePropertiesBinary(
                                   std::to_string(_npartitions) + ".bin";
     std::ofstream out_val(fname_val.c_str(), std::ios::binary | std::ios::out);
 
-    auto const& properties(_partitioned_properties);
-    auto const& property_names(properties.getPropertyVectorNames());
     std::size_t number_of_properties(property_names.size());
     out.write(reinterpret_cast<char*>(&number_of_properties),
               sizeof(number_of_properties));
--- a/Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
+++ b/Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
@@ -336,11 +336,14 @@ void NodeWiseMeshPartitioner::writePropertiesBinary(
         }
         MeshLib::IO::writePropertyVectorMetaDataBinary(out, pvmd);
     }
+    unsigned long offset = 0;
     for (const auto& partition : _partitions)
     {
         MeshLib::IO::PropertyVectorPartitionMetaData pvpmd;
+        pvpmd.offset = offset;
         pvpmd.number_of_tuples = partition.number_of_non_ghost_nodes;
         MeshLib::IO::writePropertyVectorPartitionMetaData(out, pvpmd);
+        offset += pvpmd.number_of_tuples;
     }
     out.close();
     out_val.close();
--- a/Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
+++ b/Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
@@ -17,6 +17,7 @@
 #include <limits>
 #include <iomanip>
 #include <cstdio>  // for binary output
+#include <numeric>
 
 #include <logog/include/logog.hpp>
 
@@ -203,6 +204,65 @@ void NodeWiseMeshPartitioner::processPartition(std::size_t const part_id,
                                extra_nodes.end());
 }
 
+void NodeWiseMeshPartitioner::processProperties()
+{
+    std::size_t const total_number_of_tuples =
+        std::accumulate(std::begin(_partitions), std::end(_partitions), 0,
+                        [](std::size_t const sum, Partition const& p) {
+                            return sum + p.nodes.size();
+                        });
+
+    INFO("*** total number of tuples after partitioning: %d ***",
+         total_number_of_tuples);
+    // 1 create new PV
+    // 2 resize the PV with total_number_of_tuples
+    // 3 copy the values according to the partition info
+    auto const& original_properties(_mesh->getProperties());
+    auto property_names = original_properties.getPropertyVectorNames();
+    for (auto const& name : property_names)
+    {
+        if (original_properties.existsPropertyVector<double>(name))
+        {
+            auto const& pv(original_properties.getPropertyVector<double>(name));
+            auto partitioned_pv =
+                _partitioned_properties.createNewPropertyVector<double>(
+                    name, pv->getMeshItemType(), pv->getNumberOfComponents());
+            partitioned_pv->resize(total_number_of_tuples *
+                                   pv->getNumberOfComponents());
+            std::size_t cnt(0);
+            for (auto p : _partitions)
+            {
+                for (std::size_t i = 0; i < p.nodes.size(); ++i)
+                {
+                    const auto global_id = p.nodes[i]->getID();
+                    (*partitioned_pv)[cnt+i] = (*pv)[global_id];
+                }
+                cnt += p.nodes.size();
+            }
+        }
+        if (original_properties.existsPropertyVector<int>(name))
+        {
+            auto const& pv(original_properties.getPropertyVector<int>(name));
+            auto partitioned_pv =
+                _partitioned_properties.createNewPropertyVector<int>(
+                    name, pv->getMeshItemType(), pv->getNumberOfComponents());
+            partitioned_pv->resize(total_number_of_tuples *
+                                   pv->getNumberOfComponents());
+            std::size_t cnt(0);
+            for (auto p : _partitions)
+            {
+                for (std::size_t i = 0; i < p.nodes.size(); ++i)
+                {
+                    const auto global_id = p.nodes[i]->getID();
+                    (*partitioned_pv)[cnt+i] = (*pv)[global_id];
+                }
+                cnt += p.nodes.size();
+            }
+
+        }
+    }
+}
+
 void NodeWiseMeshPartitioner::partitionByMETIS(
     const bool is_mixed_high_order_linear_elems)
 {
@@ -213,6 +273,8 @@ void NodeWiseMeshPartitioner::partitionByMETIS(
     }
 
     renumberNodeIndices(is_mixed_high_order_linear_elems);
+
+    processProperties();
 }
 
 void NodeWiseMeshPartitioner::renumberNodeIndices(
@@ -304,7 +366,7 @@ void NodeWiseMeshPartitioner::writePropertiesBinary(
                                   std::to_string(_npartitions) + ".bin";
     std::ofstream out_val(fname_val.c_str(), std::ios::binary | std::ios::out);
 
-    auto const& properties(_mesh->getProperties());
+    auto const& properties(_partitioned_properties);
     auto const& property_names(properties.getPropertyVectorNames());
     std::size_t number_of_properties(property_names.size());
     out.write(reinterpret_cast<char*>(&number_of_properties),
@@ -315,9 +377,9 @@ void NodeWiseMeshPartitioner::writePropertiesBinary(
         pvmd.property_name = name;
         pvmd.is_int_type = false;
         {
-            auto *pv = properties.getPropertyVector<double>(name);
-            if (pv)
+            if (properties.existsPropertyVector<double>(name))
             {
+                auto* pv = properties.getPropertyVector<double>(name);
                 pvmd.is_int_type = false;
                 pvmd.is_data_type_signed = false;
                 pvmd.data_type_size_in_bytes = sizeof(double);
@@ -327,9 +389,9 @@ void NodeWiseMeshPartitioner::writePropertiesBinary(
             }
         }
         {
-            auto *pv = properties.getPropertyVector<float>(name);
-            if (pv)
+            if (properties.existsPropertyVector<float>(name))
             {
+                auto* pv = properties.getPropertyVector<float>(name);
                 pvmd.is_int_type = false;
                 pvmd.is_data_type_signed = false;
                 pvmd.data_type_size_in_bytes = sizeof(float);
@@ -339,9 +401,9 @@ void NodeWiseMeshPartitioner::writePropertiesBinary(
             }
         }
         {
-            auto* pv = properties.getPropertyVector<int>(name);
-            if (pv)
+            if (properties.existsPropertyVector<int>(name))
             {
+                auto* pv = properties.getPropertyVector<int>(name);
                 pvmd.is_int_type = true;
                 pvmd.is_data_type_signed = true;
                 pvmd.data_type_size_in_bytes = sizeof(int);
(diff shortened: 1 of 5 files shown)


endJunction:
For me the same ctests are failing as in [Jenkins bulid](https://jenkins.opengeosys.org/job/ufz/job/ogs/view/change-requests/job/PR-1745/2/execution/node/159/log/) especially the nodePartitionedMeshReader test...

Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
```diff
@@ -84,106 +85,195 @@ void NodeWiseMeshPartitioner::readMetisData(const std::string& file_name_base)
...
+        }
+    }
+    partition.number_of_non_ghost_base_nodes = partition.nodes.size();
+    partition.number_of_non_ghost_nodes =
+        partition.number_of_non_ghost_base_nodes + extra_nodes.size();
+}
 
-        INFO("Processing partition: %d", part_id);
+void NodeWiseMeshPartitioner::findElementsInPartition(
+    std::size_t part_id, const bool is_mixed_high_order_linear_elems)
```
endJunction: could be const as the other argument is...

Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
```diff
@@ -84,106 +85,195 @@ void NodeWiseMeshPartitioner::readMetisData(const std::string& file_name_base)
...
+{
+    auto& partition = _partitions[part_id];
+    std::vector<MeshLib::Element*> const& elements = _mesh->getElements();
+    for (std::size_t elem_id = 0; elem_id < elements.size(); elem_id++)
+    {
+        const auto* elem = elements[elem_id];
+        if (_elements_status[elem_id])
```
endJunction: Maybe not from this PR, but if _element_status would be an enum, the if-condition would be easier to understand.

Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
```diff
@@ -84,106 +85,195 @@ void NodeWiseMeshPartitioner::readMetisData(const std::string& file_name_base)
...
+            const unsigned node_id = ghost_elem->getNodeIndex(i);
+            if (nodes_reserved[node_id])
                 continue;
 
-            std::size_t non_ghost_node_number = 0;
-            for (unsigned i = 0; i < elem->getNumberOfNodes(); i++)
+            if (_nodes_partition_ids[node_id] != part_id)
             {
-                if (_nodes_partition_ids[elem->getNodeIndex(i)] == part_id)
+                if (is_mixed_high_order_linear_elems)
```
endJunction: This is repeated (or very similar to above). Is it possible to extract a method?

Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
```diff
@@ -84,106 +85,195 @@ void NodeWiseMeshPartitioner::readMetisData(const std::string& file_name_base)
...
-                partition.regular_elements.push_back(elem);
-                _elements_status[elem_id] = true;
-            }
-            else
+void NodeWiseMeshPartitioner::processProperties()
+{
+    std::size_t total_number_of_tuples = 0;
+    for (auto const& p : _partitions)
+    {
+        total_number_of_tuples += p.nodes.size();
```
endJunction: Alternative (though a little longer):
```c++
std::size_t const n_tuples = std::accumulate(
    std::begin(_partitions), std::end(_partitions), 0, 
    [](std::size_t const sum, PartitionType const& p) { return sum + p.size(); });
```

Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
```diff
@@ -84,106 +85,195 @@ void NodeWiseMeshPartitioner::readMetisData(const std::string& file_name_base)
...
+    {
+        if (original_properties.existsPropertyVector<double>(name))
+        {
+            auto const& pv(original_properties.getPropertyVector<double>(name));
+            auto partitioned_pv =
+                _partitioned_properties.createNewPropertyVector<double>(
+                    name, pv->getMeshItemType(), pv->getNumberOfComponents());
+            partitioned_pv->resize(total_number_of_tuples *
+                                   pv->getNumberOfComponents());
+            std::size_t cnt(0);
```
endJunction: Can you add a short comment about what is counted, please.

Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
```diff
@@ -84,106 +85,195 @@ void NodeWiseMeshPartitioner::readMetisData(const std::string& file_name_base)
...
         {
-            for (unsigned i = 0; i < ghost_elem->getNumberOfNodes(); i++)
+            auto const& pv(original_properties.getPropertyVector<int>(name));
+            auto partitioned_pv =
+                _partitioned_properties.createNewPropertyVector<int>(
+                    name, pv->getMeshItemType(), pv->getNumberOfComponents());
+            partitioned_pv->resize(total_number_of_tuples *
+                                   pv->getNumberOfComponents());
+            std::size_t cnt(0);
+            for (auto p : _partitions)
```
endJunction: This counting seems to be the same as above...

Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
```diff
@@ -262,6 +352,133 @@ NodeWiseMeshPartitioner::getNumberOfIntegerVariablesOfElements(
...
+    std::size_t number_of_properties(property_names.size());
+    out.write(reinterpret_cast<char*>(&number_of_properties),
+              sizeof(number_of_properties));
+    for (auto const& name : property_names)
+    {
+        MeshLib::IO::PropertyVectorMetaData pvmd;
+        pvmd.property_name = name;
+        pvmd.is_int_type = false;
+        {
+            if (properties.existsPropertyVector<double>(name))
```
endJunction: This and the following ifs are good candidates for a template over the scalar type...

Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
```diff
@@ -262,6 +352,133 @@ NodeWiseMeshPartitioner::getNumberOfIntegerVariablesOfElements(
...
+    }
+    std::size_t number_of_properties = 0;
+    is.read(reinterpret_cast<char*>(&number_of_properties), sizeof(std::size_t));
+    for (std::size_t i(0); i < number_of_properties; ++i)
+    {
+        boost::optional<MeshLib::IO::PropertyVectorMetaData> pvmd(
+            MeshLib::IO::readPropertyVectorMetaData(is));
+        if (pvmd) {
+            INFO("readPropertiesConfigMetaDataBinary:");
+            MeshLib::IO::writePropertyVectorMetaData(std::cout, *pvmd);
```
endJunction: Not sure about the `std::cout` here. Let's discuss later.

MeshLib/IO/MPI_IO/PropertyVectorMetaData.h
```diff
@@ -0,0 +1,138 @@
...
+                                        PropertyVectorMetaData const& pvmd)
+{
+    os << "size of name: " << pvmd.property_name.length() << std::endl;
+    os << "name: '" << pvmd.property_name << "'" << std::endl;
+    os << "is_int_data_type: " << pvmd.is_int_type << std::endl;
+    os << "is_data_type_signed: " << pvmd.is_data_type_signed << std::endl;
+    os << "data_type_size_in_bytes: " << pvmd.data_type_size_in_bytes
+       << std::endl;
+    os << "number of components: " << pvmd.number_of_components << std::endl;
+    os << "number of tuples: " << pvmd.number_of_tuples << std::endl;
```
endJunction: Better to use "\n" and if flushing is needed keep the last `std::endl`.

Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
TomFischer: I'm unsure but I think the vector `_element_status` contains a flag for every element. If the flag is set to true the element is a regular element (an element without ghost nodes), else it is set to false. Maybe a renaming of the vector raises the readability, too? Maybe @wenqing can confirm this?!

Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
TomFischer: It looks similar, but the innermost things are different. One implementation fills and uses a vector `nodes_reserved`, the other doesn't. One implementation checks if the current node belongs to the partition, the other checks if the current node does not belong to the current partition. I don't see an easy way to extract a method.

Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
TomFischer: Changed to logog.

MeshLib/IO/MPI_IO/PropertyVectorMetaData.h
TomFischer: The use of logog will fix this issue.

Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
TomFischer: The alternative has the advantage that `total_number_of_tuples` can be `const`.

Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
TomFischer: Renamed to `position_offset`. Hope this makes it clearer.

Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
```diff
@@ -262,6 +333,133 @@ NodeWiseMeshPartitioner::getNumberOfIntegerVariablesOfElements(
...
+    out_val.close();
+}
+
+void NodeWiseMeshPartitioner::readPropertiesConfigDataBinary(
+    const std::string& file_name_base) const
+{
+    const std::string fname = file_name_base + "_partitioned_properties_cfg"
+                              + std::to_string(_npartitions) + ".bin";
+    std::ifstream is(fname.c_str(), std::ios::binary | std::ios::in);
+    if (!is)
```
wenqing: Since the parallel reading is implemented by your own instead of using MPI_File_read, you need more MPI barriers to synchronise the computations among compute code.  Besides, if conditions, which may be in different status on different compute cores,  have be to check by MPI communication. Therefore,  I am not sure about the efficiency of the current implementation.

You can keep the current implementation by adding
``MPI_Barrier(_mpi_comm )`` before this if-condition and before the end of this function.

Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
```diff
@@ -262,6 +333,133 @@ NodeWiseMeshPartitioner::getNumberOfIntegerVariablesOfElements(
...
+    {
+        ERR("Could not open file '%s' in binary mode.", fname.c_str());
+    }
+    std::size_t number_of_properties = 0;
+    is.read(reinterpret_cast<char*>(&number_of_properties), sizeof(std::size_t));
+    for (std::size_t i(0); i < number_of_properties; ++i)
+    {
+        boost::optional<MeshLib::IO::PropertyVectorMetaData> pvmd(
+            MeshLib::IO::readPropertyVectorMetaData(is));
+        if (pvmd) {
```
wenqing: Need to check all cores via the MPI communication.

Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
```diff
@@ -262,6 +333,133 @@ NodeWiseMeshPartitioner::getNumberOfIntegerVariablesOfElements(
...
+    }
+    auto pos = is.tellg();
+    for (std::size_t i(0); i < _npartitions; ++i) {
+        auto offset =
+            pos + static_cast<std::streampos>(
+                      i * sizeof(MeshLib::IO::PropertyVectorPartitionMetaData));
+        is.seekg(offset);
+        boost::optional<MeshLib::IO::PropertyVectorPartitionMetaData> pvpmd(
+            MeshLib::IO::readPropertyVectorPartitionMetaData(is));
+        if (pvpmd)
```
wenqing: The same here.

MeshLib/IO/MPI_IO/NodePartitionedMeshReader.cpp
```diff
@@ -184,8 +184,121 @@ MeshLib::NodePartitionedMesh* NodePartitionedMeshReader::readBinary(
...
+    MeshLib::Properties p(readPropertiesBinary(file_name_base));
+
+    return newMesh(BaseLib::extractBaseName(file_name_base), mesh_nodes,
+                   glb_node_ids, mesh_elems, p);
+}
+
+MeshLib::Properties NodePartitionedMeshReader::readPropertiesBinary(
+    const std::string& file_name_base) const
+{
+    const std::string fname_cfg = file_name_base +
```
wenqing: Make the reading optional.

MeshLib/IO/MPI_IO/NodePartitionedMeshReader.cpp
```diff
@@ -184,8 +184,121 @@ MeshLib::NodePartitionedMesh* NodePartitionedMeshReader::readBinary(
...
+                                  "_partitioned_properties_cfg" +
+                                  std::to_string(_mpi_comm_size) + ".bin";
+    std::ifstream is(fname_cfg.c_str(), std::ios::binary | std::ios::in);
+    if (!is)
+    {
+        WARN("Could not open file '%s' in binary mode.", fname_cfg.c_str());
+        return MeshLib::Properties();
+    }
+    std::size_t number_of_properties = 0;
```
wenqing: call MPI_barrier.

MeshLib/IO/MPI_IO/NodePartitionedMeshReader.cpp
```diff
@@ -184,8 +184,121 @@ MeshLib::NodePartitionedMesh* NodePartitionedMeshReader::readBinary(
...
+    is.read(reinterpret_cast<char*>(&number_of_properties), sizeof(std::size_t));
+    std::vector<boost::optional<MeshLib::IO::PropertyVectorMetaData>> vec_pvmd(
+        number_of_properties);
+    for (std::size_t i(0); i < number_of_properties; ++i)
+    {
+        vec_pvmd[i] = MeshLib::IO::readPropertyVectorMetaData(is);
+        if (!vec_pvmd[i])
```
wenqing: check if-condition via MPI communication.

MeshLib/IO/MPI_IO/NodePartitionedMeshReader.cpp
```diff
@@ -184,8 +184,121 @@ MeshLib::NodePartitionedMesh* NodePartitionedMeshReader::readBinary(
...
+    }
+    auto pos = is.tellg();
+    auto offset =
+        pos +
+        static_cast<long>(_mpi_rank *
+                          sizeof(MeshLib::IO::PropertyVectorPartitionMetaData));
+    is.seekg(offset);
+    boost::optional<MeshLib::IO::PropertyVectorPartitionMetaData> pvpmd(
+        MeshLib::IO::readPropertyVectorPartitionMetaData(is));
+    if (!pvpmd)
```
wenqing: Check if condition via MPI communication.

MeshLib/IO/MPI_IO/NodePartitionedMeshReader.h
```diff
@@ -158,18 +165,49 @@ class NodePartitionedMeshReader
...
+                                         pvmd.number_of_components);
+        pv->resize(pvpmd.number_of_tuples * pvmd.number_of_components);
+        // jump to the place for reading the specific part of the
+        // PropertyVector
+        is.seekg(global_offset + pvpmd.offset * sizeof(T));
+        // read the values
+        unsigned long const number_of_bytes = pvmd.data_type_size_in_bytes *
+                                              pvpmd.number_of_tuples *
+                                              pvmd.number_of_components;
+        if (!is.read(reinterpret_cast<char*>(pv->data()), number_of_bytes))
```
wenqing: Check if-condition via MPI communication.

wenqing:
After discussed with Tom, we deleted some of my comments. Except the optional I/O, the suggestion from my side is to add  MPI_barrier(_mpi_comm) before return of
``MeshLib::Properties NodePartitionedMeshReader::readPropertiesBinary(...)``

wenqing:
@TomFischer Since my suggestion of if-condition MPI checking is not necessary, MPI_barrier is useless here.

wenqing:
@TomFischer Then my suggestion is only about the optional I/O.

MeshLib/IO/MPI_IO/NodePartitionedMeshReader.cpp
wenqing: Suggest to replace it with
```
 boost::optional<MeshLib::IO::PropertyVectorPartitionMetaData> pvpmd_i(
        MeshLib::IO::readPropertyVectorPartitionMetaData(is));
MPI_C_BOOL pvpmd_read_ok_i = static_cast<MPI_C_BOOL>(pvpmd_i);
MPI_C_BOOL pvpmd_read_ok; 
MPI_Allreduce( &pvpmd_read_ok_i, &pvpmd_read_ok, 1, MPI_C_BOOL, MPI_LOR, MPI_COMM_WORLD);
if (!pvpmd_read_ok)
```
Otherwise, one rank fails and quit while some other rank may be good and still run.

wenqing: I have tested with the following code. When you run it, you can see that program on core 0 stops, while program on other core is still running. Therefore, it is better to use my suggestion.
```
#include <mpi.h>
#include <iostream>
#include <cstdlib>

int main(int argc, char** argv) {
    MPI_Init(NULL, NULL);

    int world_size;
    MPI_Comm_size(MPI_COMM_WORLD, &world_size);

    int world_rank;
    MPI_Comm_rank(MPI_COMM_WORLD, &world_rank);

    char processor_name[MPI_MAX_PROCESSOR_NAME];
    int name_len;
    MPI_Get_processor_name(processor_name, &name_len);

    for (int i=0; i<100000; i++)
    {
       if (world_rank == 0) // abort
           abort();
       std::cout<< world_rank << " out of processors " <<  world_size << " is still running"<< std::endl;
    }

    MPI_Finalize();
}
```

wenqing: Maybe you can leave like it is, and we modify the OGS_FATAL for MPI.

wenqing:
:fast_forward:

Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
```diff
@@ -262,6 +329,57 @@ NodeWiseMeshPartitioner::getNumberOfIntegerVariablesOfElements(
...
+            OGS_FATAL(
+                "writePropertiesBinary: Could not write PropertyVector '%s'.",
+                name.c_str());
+    }
+    unsigned long offset = 0;
+    for (const auto& partition : _partitions)
+    {
+        MeshLib::IO::PropertyVectorPartitionMetaData pvpmd;
+        pvpmd.offset = offset;
+        pvpmd.number_of_tuples = partition.nodes.size();
```
wenqing: This causes problem for element wise properties, e.g MaterialID.

TomFischer: As mentioned in the introduction to the PR this works only for `PropertyVector`s assigned to mesh nodes.

PR #1748, opened 2017-03-14
[MatL] Correct s_odot_s() product for 2D cases.

endJunction:
Small bug fix for a problem showing up since aligned Eigen matrices are used. The previous implementation would write behind the allocated space in 2D cases.

The 3D case code is just moved. The 2D case code is new.

### DIFF
--- a/MaterialLib/SolidModels/Ehlers.cpp
+++ b/MaterialLib/SolidModels/Ehlers.cpp
@@ -19,7 +19,7 @@ template class SolidEhlers<2>;
 template class SolidEhlers<3>;
 
 template <>
-ProcessLib::KelvinMatrixType<3> s_odot_s<3>(
+ProcessLib::KelvinMatrixType<3> sOdotS<3>(
     ProcessLib::KelvinVectorType<3> const& v)
 {
     ProcessLib::KelvinMatrixType<3> result;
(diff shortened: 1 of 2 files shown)


MaterialLib/SolidModels/Ehlers.cpp
```diff
@@ -18,6 +18,67 @@ namespace Ehlers
...
+        v(0) * v(4) / std::sqrt(2.) + v(3) * v(5) / 2.;
+
+    result(4, 4) = v(1) * v(2) + v(4) * v(4) / 2.;
+    result(4, 5) = result(5, 4) =
+        v(3) * v(2) / std::sqrt(2.) + v(5) * v(4) / 2.;
+
+    result(5, 5) = v(0) * v(2) + v(5) * v(5) / 2.;
+    return result;
+}
+
```
nagelt: @xingyuanmiao `d(A^-1)/dA = -A^-1 \odot A^-1` .... where has the minus gone? is it used outside directly in the Ehlers function?

xingyuanmiao: Yes, the minus sign is used directly in the equation of calculating the second derivative of theta, and it is the only function that needs involve sOdotS.

endJunction: @nagelt @xingyuanmiao I've extended the comment. If there is need to improve the code, let me know.

PR #1750, opened 2017-03-15
Fix Hex20 shape functions; Add partition of unity tests.

endJunction:
Same as in the OGS-5 code in [FEM/mathlib.cpp](https://github.com/ufz/ogs5/blob/master/FEM/mathlib.cpp#L4788) ( and in all previous versions of that code looked up to revision 6 in the svn, which is an import from 4.2.17.) Maybe the node ordering was different (for example (1, 1, 1) is the first point not the eighth).

The assumption that the current implementation is wrong is based on simple test for the first node of the hexahedron with coordinates (-1, -1, -1), where the shape function's value must be 1...


Tests/AutoCheckTools.h
```diff
@@ -13,6 +13,22 @@
 
 #include "MathLib/Point3d.h"
 
+
+namespace std{
```
TomFischer: I thought the namespace `std` is exclusively for standard template library. Why is it neccessary to put the operator in the namespace?

Tests/NumLib/TestShapeFunctions.cpp
```diff
@@ -11,16 +11,142 @@
...
+#include <numeric>
 #include <valarray>
-#include <algorithm>
 
+#include "MeshLib/Elements/Elements.h"
+#include "NumLib/Fem/ShapeFunction/ShapeHex20.h"
+#include "NumLib/Fem/ShapeFunction/ShapeHex8.h"
+#include "NumLib/Fem/ShapeFunction/ShapeLine2.h"
+#include "NumLib/Fem/ShapeFunction/ShapeLine3.h"
+#include "NumLib/Fem/ShapeFunction/ShapePrism15.h"
```
TomFischer: I don't see all types included here in the `ShapeFunctionTestTypes` in line 101 ff. Why are there more included types (`ShapePrism*`, `ShapePyra*`, `ShapeQuad9`)?

TomFischer:
Thanks for fixing the bug and adding tests for a lot more Shape functions. 🥇

Tests/AutoCheckTools.h
endJunction: There is no output operator for std::array in the STL. The operator is required for output of the arrays in case of a failing test.
One could move the operator to `autocheck` namespace but it must be visible _before_ the `autochcek.hpp` include...
Having it in the `std` namespace allows for an argument dependent lookup.

wenqing:
Looks good. However, some mesh tools may have different node numbering order, like GINA, with which hex is numbering from top. Therefore I think it is better to use the absolute value of the determinate of the coordinate mapping Jacobian (shapemat.detJ = std::fabs(shapemat.J.determinant();)
https://github.com/ufz/ogs/blob/master/NumLib/Fem/CoordinatesMapping/NaturalCoordinatesMapping.cpp#L126

endJunction:
I think it is dangerous to allow negative det(J); better to force consistent meshes.
The current mesh input format is that of VTK; other tools have to write correct vtu files...

wenqing:
vtu can also accept the the hex mesh that hex nodes are numbered from top, positive  det(J) does mean the numbering is Ok. det(J) = 0 definitely tells us the element has overlapped nodes. Anyway, we can stick to the node number order given by vtk format.

PR #1754, opened 2017-03-22
HM process; secondary variables extrapolation

endJunction:
Copied together from SmallDeformation and LiquidFlow processes.

The output of Darcy velocity needs validation.
@grubbymoon Would you please do this?

### DIFF
--- a/ProcessLib/HydroMechanics/HydroMechanicsFEM.h
+++ b/ProcessLib/HydroMechanics/HydroMechanicsFEM.h
@@ -61,7 +61,7 @@ struct IntegrationPointData final
     typename ShapeMatrixTypeDisplacement::template MatrixType<
         DisplacementDim, NPoints * DisplacementDim>
         N_u;
-    //typename ShapeMatrixTypeDisplacement::NodalRowVectorType N_u;
+    // typename ShapeMatrixTypeDisplacement::NodalRowVectorType N_u;
     typename BMatricesType::BMatrixType b_matrices;
     typename BMatricesType::KelvinVectorType sigma_eff, sigma_eff_prev;
     typename BMatricesType::KelvinVectorType eps, eps_prev;
@@ -207,8 +207,8 @@ public:
             auto& ip_data = _ip_data[ip];
             auto const& sm = shape_matrices_u[ip];
             _ip_data[ip].integration_weight =
-                _integration_method.getWeightedPoint(ip).getWeight() * sm.integralMeasure *
-                shape_matrices_u[ip].detJ;
+                _integration_method.getWeightedPoint(ip).getWeight() *
+                sm.integralMeasure * shape_matrices_u[ip].detJ;
             ip_data.b_matrices.resize(
                 kelvin_vector_size,
                 ShapeFunctionDisplacement::NPOINTS * DisplacementDim);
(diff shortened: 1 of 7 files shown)


endJunction:
@grubbymoon Please reply before weekend if the output of Darcy's velocity is correct.

ProcessLib/HydroMechanics/HydroMechanicsFEM.h
```diff
@@ -354,6 +416,178 @@ class HydroMechanicsLocalAssembler : public ProcessLib::LocalAssemblerInterface
...
+            double const K_over_mu =
+                _process_data.intrinsic_permeability(t, x_position)[0] /
+                _process_data.fluid_viscosity(t, x_position)[0];
+
+            auto const rho_fr = _process_data.fluid_density(t, x_position)[0];
+            auto const& b = _process_data.specific_body_force;
+
+            // Compute the velocity
+            auto const& dNdx_p = _ip_data[ip].dNdx_p;
+            GlobalDimVectorType const darcy_velocity = -K_over_mu * dNdx_p * p
```
TomFischer: formatting

grubbymoon:
I have tested the velocity output with different transient pressure boundary and the result is correct !

ProcessLib/HydroMechanics/HydroMechanicsFEM.h
```diff
@@ -99,9 +99,68 @@ struct IntegrationPointData final
     EIGEN_MAKE_ALIGNED_OPERATOR_NEW;
 };
 
+/// Used by for extrapolation of the integration point values. It is ordered
+/// (and stored) by integration points.
+template <typename ShapeMatrixType>
+struct SecondaryData
+{
+    std::vector<ShapeMatrixType, Eigen::aligned_allocator<ShapeMatrixType>> N;
```
endJunction: @TomFischer shape functions for the displacement variable are not directly stored in the integration point data; Would like to keep SecondaryData separate.

chleh: IMHO there should be some kind of indication that N is for the displacement and not for the pressure variable. E.g., rename to `N_u`

ProcessLib/HydroMechanics/HydroMechanicsFEM.h
```diff
@@ -354,6 +416,181 @@ class HydroMechanicsLocalAssembler : public ProcessLib::LocalAssemblerInterface
         }
     }
 
+    void postTimestepConcrete(std::vector<double> const& local_x) override
+    {
+        double const& t = _process_data.t;
+
+        auto p =
+            Eigen::Map<typename ShapeMatricesTypePressure::template VectorType<
```
chleh: Note: There are EigenMapTools in MathLib/LinAlg. Could maybe be more compact.

endJunction: The `MathLib::toVector(vector, size)` is not applicable here, because the data is located at the offset `pressure_index` and is not the whole vector...
One could add the additional parameters and it would look like this (on the caller side):
```c++
auto p =
            MathLib::toVector<typename ShapeMatricesTypePressure::template VectorType<
                pressure_size> const>(local_x, pressure_index, pressure_size);
```
which is only one character shorter then the current version....

chleh: I see.

PR #1755, opened 2017-03-28
Add output compression to project configuration.

endJunction:
The default value is to compress the vtu output, but for a lot of output in production code the compression takes around 1/6th of each timestep (with output). The optional `<compress_output>` option allows switching off the compression.

Some results for the GWF/cube_1e6.prj:
File size increases a lot, almost five-fold:
-rw-r--r-- 1 naumov users   179 597 443   Mar 27 23:37 cube_1e6_pcs_0_ts_1_t_1.000000.vtu
-rw-r--r-- 1 naumov users   36 323 990   Mar 28 00:45 cube_1e6_pcs_0_ts_1_t_1.000000.vtu

But the time goes down
from 28.086064819 seconds ( +-  0.88% )
to 23.330982410 seconds ( +-  1.07% ),
which is 83% of the original runtime ;) (It would be more efficient for a transient simulation even...)

### DIFF
--- a/MeshLib/IO/VtkIO/VtuInterface-impl.h
+++ b/MeshLib/IO/VtkIO/VtuInterface-impl.h
@@ -48,8 +48,10 @@ bool VtuInterface::writeVTU(std::string const &file_name, const int num_partitio
 
     vtuWriter->SetInputConnection(vtkSource->GetOutputPort());
 
-    if(_use_compressor && num_partitions == 0)
+    if(_use_compressor)
         vtuWriter->SetCompressorTypeToZLib();
+    else
+        vtuWriter->SetCompressorTypeToNone();
 
     vtuWriter->SetDataMode(_data_mode);
     if (_data_mode == vtkXMLWriter::Appended)
--- a/ProcessLib/ProcessOutput.h
+++ b/ProcessLib/ProcessOutput.h
@@ -30,14 +30,14 @@ struct ProcessOutput final
 
 
 //! Writes output to the given \c file_name using the VTU file format.
-void doProcessOutput(
-        std::string const& file_name,
-        GlobalVector const& x,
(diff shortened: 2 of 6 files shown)


MeshLib/IO/VtkIO/VtuInterface-impl.h
```diff
@@ -48,8 +48,10 @@ bool VtuInterface::writeVTU(std::string const &file_name, const int num_partitio
 
     vtuWriter->SetInputConnection(vtkSource->GetOutputPort());
 
-    if(_use_compressor && num_partitions == 0)
```
endJunction: @wenqing I could not figure out if removing this condition is OK or not. For the serial case `num_partitions` is 1 and now it is working as advertised. I'm not sure in which case the number of partitions would be zero.
So this might need an additional test case....

ProcessLib/ProcessOutput.h
```diff
@@ -30,14 +30,14 @@ struct ProcessOutput final
...
-        std::string const& file_name,
-        GlobalVector const& x,
-        MeshLib::Mesh& mesh,
-        NumLib::LocalToGlobalIndexMap const& dof_table,
-        std::vector<std::reference_wrapper<ProcessVariable>> const&
-        process_variables,
-        SecondaryVariableCollection secondary_variables,
-        ProcessOutput const& process_output);
+void doProcessOutput(std::string const& file_name,
+                     bool const& compress_output,
```
TomFischer: Is there a special reason to give `compress_output` as a const reference?

bilke:
👍

ProcessLib/ProcessOutput.h
endJunction: Good catch!
Interestingly the compiler produces slightly larger code for the `const &` argument...

PR #1757, opened 2017-04-03
Nodal force output

endJunction:
Output of so called nodal forces common for all mechanics processes.

TODO:

- [x] Find and replace other places of getOrCreateMeshProperty() #1766 
- [x] Test linear cube example output
- [x] ~~Use in HM~~ (Removed for now, because of lack of a good test case. @nagelt Ideas?)

### DIFF
--- a/ProcessLib/SmallDeformation/SmallDeformationFEM.h
+++ b/ProcessLib/SmallDeformation/SmallDeformationFEM.h
@@ -89,6 +89,7 @@ struct SecondaryData
 
 struct SmallDeformationLocalAssemblerInterface
     : public ProcessLib::LocalAssemblerInterface,
+      public ProcessLib::SmallDeformation::NodalForceCalculationInterface,
       public NumLib::ExtrapolatableElement
 {
     virtual std::vector<double> const& getIntPtSigmaXX(
@@ -126,9 +127,6 @@ struct SmallDeformationLocalAssemblerInterface
 
     virtual std::vector<double> const& getIntPtEpsilonYZ(
         std::vector<double>& cache) const = 0;
-
-    virtual std::vector<double> const& getNodalValues(
-        std::vector<double>& nodal_values) const = 0;
 };
 
 template <typename ShapeFunction, typename IntegrationMethod,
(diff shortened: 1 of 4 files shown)


nagelt:
@endJunction Idea for HM Benchmark:
Use a confined compression test with an axisymmetric setup (to also test his feature). The axial forces based on total stress should show the relaxation due to flow-dependent viscoelasticity corresponding to the applied load. The radial forces should equal the integration of the radial stress solution over the circumference. For both we have analytical solutions.

ProcessLib/SmallDeformation/SmallDeformationFEM.h
```diff
@@ -126,6 +126,9 @@ struct SmallDeformationLocalAssemblerInterface
 
     virtual std::vector<double> const& getIntPtEpsilonYZ(
         std::vector<double>& cache) const = 0;
+
+    virtual std::vector<double> const& getNodalValues(
```
chleh: Maybe rename to getNodalForces?

ProcessLib/SmallDeformationCommon/Common.h
```diff
@@ -0,0 +1,51 @@
...
+template <typename LocalAssemblerInterface>
+void writeNodalForces(
+    MeshLib::PropertyVector<double>& nodal_forces,
+    std::vector<std::unique_ptr<LocalAssemblerInterface>> const&
+        local_assemblers,
+    NumLib::LocalToGlobalIndexMap const& local_to_global_index_map)
+{
+    DBUG("Compute nodal forces for small deformation process.");
+
+    std::fill(std::begin(nodal_forces), std::end(nodal_forces), 0);
```
chleh: Maybe add comment: Zero out because nodal forces are summed up from all adjacent elements.

ProcessLib/SmallDeformationCommon/Common.h
```diff
@@ -14,12 +14,51 @@
...
 #include "NumLib/DOF/DOFTableUtil.h"
 
 namespace ProcessLib
 {
 namespace SmallDeformation
 {
+struct NodalForceCalculationInterface
+{
+    virtual std::vector<double> const& getNodalForces(
+        std::vector<double>& nodal_values) const = 0;
```
chleh: virtual dtor missing?

endJunction: Thanks. That's good catch!

PR #1770, opened 2017-04-13
Face rule normals and vectors

endJunction:
Used in the LIE process and needed for its 3D extension.

Optionally we could move some dealing with normals _etc._ code from LIE/Common.h to MeshLib.

### DIFF
--- a/MeshLib/Elements/FaceRule.cpp
+++ b/MeshLib/Elements/FaceRule.cpp
@@ -20,14 +20,24 @@ bool FaceRule::testElementNodeOrder(const Element* e)
 {
     MathLib::Vector3 up_vec (0,0,1);
     return (MathLib::scalarProduct(getSurfaceNormal(e), up_vec) < 0) ? true : false;
+
+MathLib::Vector3 FaceRule::getFirstSurfaceVector(Element const* const e)
+{
+    Node* const* const _nodes = e->getNodes();
+    return {*_nodes[1], *_nodes[0]};
+}
+
+MathLib::Vector3 FaceRule::getSecondSurfaceVector(Element const* const e)
+{
+    Node* const* const _nodes = e->getNodes();
+    return {*_nodes[1], *_nodes[2]};
 }
 
 MathLib::Vector3 FaceRule::getSurfaceNormal(const Element* e)
 {
-    Node * const * _nodes = e->getNodes();
-    const MathLib::Vector3 u (*_nodes[1], *_nodes[0]);
-    const MathLib::Vector3 v (*_nodes[1], *_nodes[2]);
-    return MathLib::crossProduct(u,v);
+    const MathLib::Vector3 u = getFirstSurfaceVector(e);
+    const MathLib::Vector3 v = getSecondSurfaceVector(e);
+    return MathLib::crossProduct(u, v);
 }
 
 } /* namespace */
--- a/MeshLib/Elements/FaceRule.cpp
+++ b/MeshLib/Elements/FaceRule.cpp
@@ -15,11 +15,10 @@
 
 namespace MeshLib
 {
-
 bool FaceRule::testElementNodeOrder(const Element* e)
 {
-    MathLib::Vector3 up_vec (0,0,1);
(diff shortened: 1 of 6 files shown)


ProcessLib/LIE/Common/Utils.h
```diff
@@ -18,11 +18,28 @@ namespace ProcessLib
...
-void computeNormalVector(MeshLib::Element const& e, Eigen::Vector3d & normal_vector);
-
-/// compute a rotation matrix from global to local coordinates based on the given normal vector
-void computeRotationMatrix(Eigen::Vector3d const& normal_vector, int dim, Eigen::MatrixXd &matR);
+/// Compute a normal vector of the given element
+///
+/// Computed normal vector is oriented in the left direction of the given line
+/// element such that computeRotationMatrix2D() returns the indentity matrix for
+/// line elements parallel to a vector (1,0,0)
+void computeNormalVector(MeshLib::Element const& e, unsigned const global_dim,
```
endJunction: I tried to move it to MeshLib, but the code (due to global dimension checks) is too specific for the current applications and needs more changes before it can go into MeshLib.

MeshLib/Elements/FaceRule.cpp
```diff
@@ -15,19 +15,29 @@
...
+    return MathLib::scalarProduct(getSurfaceNormal(e), z_unit_vector) < 0;
+}
+
+MathLib::Vector3 FaceRule::getFirstSurfaceVector(Element const* const e)
+{
+    Node* const* const _nodes = e->getNodes();
+    return {*_nodes[1], *_nodes[0]};
+}
+
+MathLib::Vector3 FaceRule::getSecondSurfaceVector(Element const* const e)
```
endJunction: I think we have to live with the old decision for a while now...

MeshLib/Elements/FaceRule.cpp
```diff
@@ -15,19 +15,29 @@
 
 namespace MeshLib
 {
-
 bool FaceRule::testElementNodeOrder(const Element* e)
 {
-    MathLib::Vector3 up_vec (0,0,1);
-    return (MathLib::scalarProduct(getSurfaceNormal(e), up_vec) < 0) ? true : false;
+    MathLib::Vector3 const z_unit_vector(0, 0, 1);
+    return MathLib::scalarProduct(getSurfaceNormal(e), z_unit_vector) < 0;
```
chleh: Wouldn't it be more efficient to just check the sign of the third component of `getSurfaceNormal(e)`?

ProcessLib/LIE/Common/Utils.cpp
```diff
@@ -15,57 +15,56 @@
...
-void computeRotationMatrix(Eigen::Vector3d const& normal_vector, int dim, Eigen::MatrixXd &matR)
-{
-    if (dim==2)
-        computeRotationMatrix2D(normal_vector, matR);
-    else
-        OGS_FATAL("2D elements are not supported in computeNormalVector()");
+        R.resize(3, 3);
+        R << u[0] , u[1] , u[2]
+           , v[0] , v[1] , v[2]
+           , n[0] , n[1] , n[2];
```
chleh: Is there a  unit test for this case?

MeshLib/Elements/FaceRule.cpp
endJunction: The MeshLib.a is by 80 bytes smaller (so presumably faster). Didn't see that because the change was automatic with clang-tidy. And if I use clang-tidy I switch off my brain...

Actually this whole thing can be replaced with `return true` without any effects on the tests. Furthermore the promise of the `testElementOrder` is not fulfilled for any element with more than three nodes.

ProcessLib/LIE/Common/Utils.cpp
endJunction: No. The ctest for LIE is covering this part (since it is specific to LIE)

endJunction: Pushing two very simple tests right now documenting the current behaviour. (Keep in mind the OGS' normals are oriented in wrong direction...)

chleh:
Thanks for adding the test. :+1:

PR #1771, opened 2017-04-13
Remove some Qt4 related code.

endJunction:
If you are aware of more qt4 workarounds, this would be a good point to finish them off.


bilke:
👍

PR #1773, opened 2017-04-18
Pre-compiled headers with cotire

bilke:
Simple PCH setup with [cotire](https://github.com/sakra/cotire) (**co**mpile **ti**me **re**ducer).

All (non-Qt) OGS libraries and the `ogs` and `testrunner` executables have PCH enabled which reduces build time by around 17 % for me:

```
With    PCH : make -j 4  1547,02s user 77,22s system 331% cpu 8:10,35 total
Without PCH : make -j 4  1860,68s user 89,95s system 359% cpu 9:03,25 total

```

For Qt targets I have not got it to work yet..
For other executables like the utilities it is not useful to use PCH:

```
-- CXX target generateMatPropsFromMatID cotired without precompiled header. Too few applicable sources.
-- CXX target GMSH2OGS cotired without precompiled header. Too few applicable sources.
...
```

See the [cotire manual](https://github.com/sakra/cotire/blob/master/MANUAL.md) for advanced usage.

cotire also feautures single compilation units (unity builds) but they do not work for all ogs targets out of the box. You can check by adding `_unity` to the target, e.g.:

```
make MeshLib_unity # works
make MaterialLib_unity # does not work
```

## ccache

See the [docs](https://docs.opengeosys.org/docs/devguide/advanced/using-ccache) for required ccache configuration.


endJunction:
Sounds interesting (and actually very promising). But. It seems that ccache is no longer working (though it is called); Try this:
```sh
make # to get everything clean and compiled and cached
touch ProcessLib/ProcessVariable.h # used in all processes
time make
```
I tried it multiple times in different orders, separate building dirs and source dirs, but w/o cotire it needs 51s, w/ cotire 4m30s because it recompiles the all the processes...
The only difference in the compilation (`make VERBOSE=1`) is that cotire adds
```
-fext-numeric-literals  -Winvalid-pch -include /home/naumov/w/ogs/x/ProcessLib/cotire/ProcessLib_CXX_prefix.hxx
```

bilke:
@endJunction 

- I added the ccache env var to the envinf1/eve module. Otherwise it cannot be set automatically.
- I added the required config to the [docs](https://docs.opengeosys.org/docs/devguide/advanced/using-ccache).
- Qt works

endJunction:
Also works with clang.

bilke:
Everything is fine except Travis + Qt (which is installed via systems package manager):

https://travis-ci.org/ufz/ogs/jobs/223698678

endJunction:
Maybe we should include cotire as a submodule?

bilke:
cotire is just one file. I would directly include it and update manually when necessary.

PR #1777, opened 2017-04-23
Multicomponent DOF table upgrade.

endJunction:
Required in ["pressure" boundary condition](https://github.com/endJunction/ogs/tree/PressureBC)

### DIFF
--- a/NumLib/DOF/LocalToGlobalIndexMap.cpp
+++ b/NumLib/DOF/LocalToGlobalIndexMap.cpp
@@ -268,7 +268,7 @@ LocalToGlobalIndexMap* LocalToGlobalIndexMap::deriveBoundaryConstrainedMap(
         getGlobalComponent(variable_id, component_id);
 
     auto mesh_component_map =
-        _mesh_component_map.getSubset(global_component_id, *mesh_subsets);
+        _mesh_component_map.getSubset({global_component_id}, *mesh_subsets);
 
     std::vector<std::unique_ptr<MeshLib::MeshSubsets>> all_mesh_subsets;
     all_mesh_subsets.emplace_back(std::move(mesh_subsets));
--- a/NumLib/DOF/LocalToGlobalIndexMap.cpp
+++ b/NumLib/DOF/LocalToGlobalIndexMap.cpp
@@ -221,6 +221,41 @@ LocalToGlobalIndexMap::LocalToGlobalIndexMap(
     }
 }
 
+LocalToGlobalIndexMap::LocalToGlobalIndexMap(
+    std::vector<std::unique_ptr<MeshLib::MeshSubsets>>&& mesh_subsets,
+    std::vector<std::size_t> const& global_component_ids,
+    std::vector<MeshLib::Element*> const& elements,
+    NumLib::MeshComponentMap&& mesh_component_map)
+    : _mesh_subsets(std::move(mesh_subsets)),
+      _mesh_component_map(std::move(mesh_component_map)),
+      _variable_component_offsets(
+          to_cumulative(std::vector<unsigned>(1, 1)))  // Single variable only.
+{
+    // Each subset in the mesh_subsets represents a single component.
+    if (_mesh_subsets.size() != global_component_ids.size())
+        OGS_FATAL(
+            "Number of mesh subsets is not equal to number of components. "
+            "There are %d mesh subsets and %d components.",
+            mesh_subsets.size(), global_component_ids.size());
+
+    for (std::size_t i = 0; i < global_component_ids.size(); ++i)
+    {
+        auto const& mss = *_mesh_subsets[i];
+
+        // For all MeshSubset in mesh_subsets and each element of that
+        // MeshSubset
+        // save a line of global indices.
+        for (MeshLib::MeshSubset const* const ms : mss)
+        {
+            std::size_t const mesh_id = ms->getMeshID();
+
+            findGlobalIndices(elements.cbegin(), elements.cend(),
+                              ms->getNodes(), mesh_id, global_component_ids[i],
+                              i);
+        }
+    }
+}
+
 LocalToGlobalIndexMap* LocalToGlobalIndexMap::deriveBoundaryConstrainedMap(
     int const variable_id,
     int const component_id,
@@ -239,6 +274,40 @@ LocalToGlobalIndexMap* LocalToGlobalIndexMap::deriveBoundaryConstrainedMap(
     all_mesh_subsets.emplace_back(std::move(mesh_subsets));
     return new LocalToGlobalIndexMap(std::move(all_mesh_subsets),
                                      global_component_id, elements,
+                                     std::move(mesh_component_map));
+}
+
+LocalToGlobalIndexMap* LocalToGlobalIndexMap::deriveBoundaryConstrainedMap(
+    int const variable_id,
+    std::vector<int> const& component_ids,
+    std::unique_ptr<MeshLib::MeshSubsets>&& mesh_subsets,
+    std::vector<MeshLib::Element*> const& elements) const
+{
+    DBUG("Construct reduced local to global index map.");
+
+    if (component_ids.empty())
+        OGS_FATAL("Expected non-empty vector of component ids.");
+
+    // Create a subset of the current mesh component map.
+    std::vector<std::size_t> global_component_ids;
+
+    for (auto component_id : component_ids)
+        global_component_ids.push_back(
+            getGlobalComponent(variable_id, component_id));
+
+    auto mesh_component_map =
+        _mesh_component_map.getSubset(global_component_ids, *mesh_subsets);
+
+    // Create copies of the mesh_subsets for each of the global components.
+    // The last component is moved after the for-loop.
+    std::vector<std::unique_ptr<MeshLib::MeshSubsets>> all_mesh_subsets;
+    for (std::size_t i = 0; i < global_component_ids.size() - 1; ++i)
+        all_mesh_subsets.emplace_back(
+            new MeshLib::MeshSubsets{*mesh_subsets});
+    all_mesh_subsets.emplace_back(std::move(mesh_subsets));
+
+    return new LocalToGlobalIndexMap(std::move(all_mesh_subsets),
+                                     global_component_ids, elements,
                                      std::move(mesh_component_map));
 }
 
--- a/NumLib/DOF/LocalToGlobalIndexMap.cpp
+++ b/NumLib/DOF/LocalToGlobalIndexMap.cpp
@@ -194,33 +194,6 @@ LocalToGlobalIndexMap::LocalToGlobalIndexMap(
     }
 }
 
-
-LocalToGlobalIndexMap::LocalToGlobalIndexMap(
-    std::vector<std::unique_ptr<MeshLib::MeshSubsets>>&& mesh_subsets,
-    int const component_id,
-    std::vector<MeshLib::Element*> const& elements,
-    NumLib::MeshComponentMap&& mesh_component_map)
-    : _mesh_subsets(std::move(mesh_subsets)),
-      _mesh_component_map(std::move(mesh_component_map)),
-      _variable_component_offsets(to_cumulative(std::vector<unsigned>(1,1))) // Single variable only.
-{
-    // There is only one mesh_subsets in the vector _mesh_subsets.
-    assert(_mesh_subsets.size() == 1);
-    auto const mss = *_mesh_subsets.front();
-
-    // For all MeshSubset in mesh_subsets and each element of that MeshSubset
-    // save a line of global indices.
-    for (MeshLib::MeshSubset const* const ms : mss)
-    {
-        std::size_t const mesh_id = ms->getMeshID();
-
-        findGlobalIndices(elements.cbegin(), elements.cend(), ms->getNodes(), mesh_id,
-                          component_id, 0);  // There is only one component to
-                                             // write out, therefore the zero
-                                             // parameter.
-    }
-}
-
 LocalToGlobalIndexMap::LocalToGlobalIndexMap(
     std::vector<std::unique_ptr<MeshLib::MeshSubsets>>&& mesh_subsets,
     std::vector<std::size_t> const& global_component_ids,
--- a/NumLib/DOF/LocalToGlobalIndexMap.cpp
+++ b/NumLib/DOF/LocalToGlobalIndexMap.cpp
@@ -258,27 +258,6 @@ LocalToGlobalIndexMap::LocalToGlobalIndexMap(
 
 LocalToGlobalIndexMap* LocalToGlobalIndexMap::deriveBoundaryConstrainedMap(
     int const variable_id,
-    int const component_id,
-    std::unique_ptr<MeshLib::MeshSubsets>&& mesh_subsets,
-    std::vector<MeshLib::Element*> const& elements) const
-{
-    DBUG("Construct reduced local to global index map.");
-    // Create a subset of the current mesh component map.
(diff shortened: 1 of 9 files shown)


NumLib/DOF/MeshComponentMap.h
```diff
@@ -45,9 +45,24 @@ class MeshComponentMap final
     /// The order (BY_LOCATION/BY_COMPONENT) of components is the same as of the
     /// current map.
     ///
+    /// This is single-component version.
+    ///
     /// \param component_id  The global components id.
     /// \param components components that should remain in the created subset
-    MeshComponentMap getSubset(std::size_t const component_id,
+    MeshComponentMap getSubset(std::size_t const component_ids,
```
chleh: misnomer: `component_ids` → `component_id`
Btw: Is that removed in a later commit in favour of the method below?

NumLib/DOF/MeshComponentMap.h
```diff
@@ -45,9 +45,24 @@ class MeshComponentMap final
     /// The order (BY_LOCATION/BY_COMPONENT) of components is the same as of the
     /// current map.
     ///
+    /// This is single-component version.
```
chleh: minor: "This is the..."

NumLib/DOF/MeshComponentMap.h
```diff
@@ -45,9 +45,24 @@ class MeshComponentMap final
...
     /// \param components components that should remain in the created subset
-    MeshComponentMap getSubset(std::size_t const component_id,
+    MeshComponentMap getSubset(std::size_t const component_ids,
+                               MeshLib::MeshSubsets const& components) const;
+
+    /// Creates a multi-component subset of the current mesh component map.
+    /// The order (BY_LOCATION/BY_COMPONENT) of components is the same as of the
+    /// current map.
+    ///
+    /// This is multi-component version.
```
chleh: "This is the"

NumLib/DOF/MeshComponentMap.h
```diff
@@ -45,9 +45,24 @@ class MeshComponentMap final
...
+    ///
+    /// \note For each component the same mesh subset will be used.
```
chleh: `MeshSubset` or `MeshSubsets`? Btw. I don't remember the difference between them any more, nor how we actually use them :-)

NumLib/DOF/MeshComponentMap.cpp
```diff
@@ -176,6 +176,36 @@ MeshComponentMap MeshComponentMap::getSubset(
     return MeshComponentMap(subset_dict, 1);
 }
 
+MeshComponentMap MeshComponentMap::getSubset(
+    std::vector<std::size_t> const& component_ids,
+    MeshLib::MeshSubsets const& mesh_subsets) const
+{
+    assert(component_ids.size() <= _num_components);
```
chleh: For future general improvement: Actually we could always turn on assertions seldomly called code, i.e. also in release mode.

NumLib/DOF/MeshComponentMap.cpp
```diff
@@ -176,6 +176,36 @@ MeshComponentMap MeshComponentMap::getSubset(
...
+                            component_id));
+        for (std::size_t j = 0; j < mesh_subset->getNumberOfElements(); j++)
+            for (std::size_t component_id : component_ids)
+                subset_dict.insert(
+                    getLine(Location(mesh_id, MeshLib::MeshItemType::Cell,
+                                     mesh_subset->getElementID(j)),
+                            component_id));
+    }
+
+    return MeshComponentMap(subset_dict, 1);
```
chleh: What's the `1` for? Is it `ComponentOrder`? Why always 1?

Tests/NumLib/TestMeshComponentMap.cpp
```diff
@@ -215,3 +215,82 @@ TEST_F(NumLibMeshComponentMapTest, DISABLED_SubsetOfNodesByLocation)
...
+    // Select some nodes from the full mesh.
+    std::array<std::size_t, 3> const ids = {{ 0, 5, 9 }};
+    std::vector<MeshLib::Node*> some_nodes;
+    for (std::size_t id : ids)
+        some_nodes.push_back(const_cast<MeshLib::Node*>(mesh->getNode(id)));
+
+    MeshLib::MeshSubset some_nodes_mesh_subset(*mesh, &some_nodes);
+
+    auto selected_component = std::unique_ptr<MeshLib::MeshSubsets>{
+        new MeshLib::MeshSubsets{&some_nodes_mesh_subset}};
```
chleh: Minor: actually there is probably no need for a unique_ptr. A normal object would suffice.

NumLib/DOF/LocalToGlobalIndexMap.h
```diff
@@ -174,6 +189,17 @@ class LocalToGlobalIndexMap final
...
         NumLib::MeshComponentMap&& mesh_component_map);
 
+    /// Private constructor used by internally created local-to-global index
+    /// maps. The mesh_component_map is passed as argument instead of being
+    /// created by the constructor.
+    /// \attention The passed mesh_component_map is in undefined state after
+    /// this construtor.
+    explicit LocalToGlobalIndexMap(
+        std::vector<std::unique_ptr<MeshLib::MeshSubsets>>&& mesh_subsets,
+        std::vector<std::size_t> const global_component_ids,
```
chleh: In the ctor above the component id is `int`.
The vector is passed by value. Is that intended? If so, then IMO there should be a comment that it is done so deliberately.

NumLib/DOF/LocalToGlobalIndexMap.cpp
```diff
@@ -221,6 +221,40 @@ LocalToGlobalIndexMap::LocalToGlobalIndexMap(
...
+    // Each subset in the mesh_subsets represents a single component.
+    if (_mesh_subsets.size() != global_component_ids.size())
+        OGS_FATAL(
+            "Number of mesh subsets is not equal to number of components. "
+            "There are %d mesh subsets and %d components.",
+            mesh_subsets.size(), global_component_ids.size());
+
+    for (std::size_t i = 0; i < global_component_ids.size(); ++i)
+    {
+        auto const mss = *_mesh_subsets[i];
```
chleh: Is that a copy? If so, is that intended?

NumLib/DOF/LocalToGlobalIndexMap.cpp
```diff
@@ -242,6 +276,34 @@ LocalToGlobalIndexMap* LocalToGlobalIndexMap::deriveBoundaryConstrainedMap(
...
+            getGlobalComponent(variable_id, component_id));
+
+    auto mesh_component_map =
+        _mesh_component_map.getSubset(global_component_ids, *mesh_subsets);
+
+    // Create copies of the mesh_subsets for each of the global components.
+    std::vector<std::unique_ptr<MeshLib::MeshSubsets>> all_mesh_subsets;
+    for (std::size_t i = 0; i < global_component_ids.size(); ++i)
+        all_mesh_subsets.emplace_back(
+            new MeshLib::MeshSubsets{*mesh_subsets});
```
chleh: Above you `std::move(mesh_subsets)`, here you don't. So IMO passing `mesh_subsets` as &&-reference is not needed here.

NumLib/DOF/LocalToGlobalIndexMap.h
```diff
@@ -174,6 +189,17 @@ class LocalToGlobalIndexMap final
         std::vector<MeshLib::Element*> const& elements,
         NumLib::MeshComponentMap&& mesh_component_map);
 
+    /// Private constructor used by internally created local-to-global index
+    /// maps. The mesh_component_map is passed as argument instead of being
+    /// created by the constructor.
+    /// \attention The passed mesh_component_map is in undefined state after
+    /// this construtor.
+    explicit LocalToGlobalIndexMap(
+        std::vector<std::unique_ptr<MeshLib::MeshSubsets>>&& mesh_subsets,
```
chleh: Btw, just out of curiosity: Why is `MeshLib::MeshSubsets` used everywhere in combination with `std::unique_ptr`? Is it an abstract type? Otherwiese one could store and pass it a usual object or reference.

Tests/NumLib/LocalToGlobalIndexMapMultiComponent.cpp
```diff
@@ -102,11 +102,45 @@ class NumLibLocalToGlobalIndexMapMultiDOFTest : public ::testing::Test
...
+    {
+        assert(selected_components.size() <= num_components);
+
+        std::vector<std::unique_ptr<MeshLib::MeshSubsets>> components;
+        for (unsigned i = 0; i < num_components; ++i)
+        {
+            components.emplace_back(
+                new MeL::MeshSubsets{mesh_items_all_nodes.get()});
+        }
+        std::vector<unsigned> vec_var_n_components(1, num_components);
```
chleh: Inconsistent naming `n_components` 😄

Tests/NumLib/LocalToGlobalIndexMapMultiComponent.cpp
```diff
@@ -200,8 +234,68 @@ void NumLibLocalToGlobalIndexMapMultiDOFTest::test(
...
+            auto const& global_idcs = (*dof_map_boundary)(e, c).rows;
+
+            ASSERT_EQ(
+                2,
+                global_idcs.size());  // boundary of quad is line with two nodes
+
+            for (unsigned n = 0; n < 2;
+                 ++n)  // boundary of quad is line with two nodes
+            {
+                auto const node = e + n;
```
chleh: Why does this formula hold? Is it because of the special mesh that is used in this test?

NumLib/DOF/MeshComponentMap.h
endJunction: MeshSubset is a part of a single Mesh. The collection of mesh subsets is stored in MeshSubsets.

NumLib/DOF/MeshComponentMap.cpp
endJunction: assert is controlled by NDEBUG macro. Better to use if (!something) OGS_FATAL("something") for such cases.

NumLib/DOF/LocalToGlobalIndexMap.h
endJunction: int for local ids and std::size_t for global.
const& is correct. Now I understand your point about pass-by-value a little better; Those who pass-by-value must std::move that thing, otherwise it's a potential problem.

Tests/NumLib/LocalToGlobalIndexMapMultiComponent.cpp
endJunction: Not a nice name, but consistently in the NumLib/DOF. The docu saying:
```c++
    /// \param vec_var_n_components  a vector of the number of variable components.
```

Tests/NumLib/LocalToGlobalIndexMapMultiComponent.cpp
endJunction: I already forgot it. The multi-component test-code is copy of the single-component with an additional loop over the components.
Let's try to figure out: e is the number of a boundary element (which is a line, so two nodes) from 0 to something. e+n must be the node ids along the x-axis of the mesh; I changed the boundary to be in y direction, and the test failed, so the assumption seems reasonable.

endJunction:
Thank you! Two or three bugs spotted!
Because the code is so buggy, someone please look thoroughly one more time.

NumLib/DOF/MeshComponentMap.h
chleh: OK, but actually I wanted to say that the docu should use the exact class name, either `MeshSubset` or `MeshSubsets`, not "mesh subset".

Tests/NumLib/LocalToGlobalIndexMapMultiComponent.cpp
chleh: So, shouldn't we your answer as a comment?

NumLib/DOF/MeshComponentMap.cpp
chleh: That's what I wanted to say. Or we could even introduce something like `OGS_ASSERT(cond, msg)`.

chleh:
I want to complain: You didn't answer to some of my comments :smile:
Unfortunately not all bugs marked as "outdated" by github were actually outdated IMO (or I just don't see your answers). Some of them are (please find them above, they are not too important.)

* What's the `1` for? Is it `ComponentOrder`? Why always 1?
* Above you std::move(mesh_subsets), here you don't. So IMO passing mesh_subsets as &&-reference is not needed here.
* Btw, just out of curiosity: Why is MeshLib::MeshSubsets used everywhere in combination with std::unique_ptr? Is it an abstract type? Otherwiese one could store and pass it a usual object or reference.

Regarding the bugs you spotted: For me the code in general looks perfectly fine. I am not very familiar with the DOF table (anymore), so maybe it's better to discuss directly, the bugs you found. Unfortunately, since you rebased during review, it's hard to figure out what you did in order to fix the "spotted bugs".

NumLib/DOF/LocalToGlobalIndexMap.cpp
```diff
@@ -194,51 +194,66 @@ LocalToGlobalIndexMap::LocalToGlobalIndexMap(
...
     auto mesh_component_map =
-        _mesh_component_map.getSubset(global_component_id, *mesh_subsets);
+        _mesh_component_map.getSubset(global_component_ids, *mesh_subsets);
 
+    // Create copies of the mesh_subsets for each of the global components.
     std::vector<std::unique_ptr<MeshLib::MeshSubsets>> all_mesh_subsets;
-    all_mesh_subsets.emplace_back(std::move(mesh_subsets));
+    for (std::size_t i = 0; i < global_component_ids.size(); ++i)
+        all_mesh_subsets.emplace_back(
+            new MeshLib::MeshSubsets{*mesh_subsets});
```
TomFischer: - Why a copy? In the original code it was a `std::move`?
- Could the for loop also be written as ```for (auto global_component_id : global_component_ids)```?

Tests/NumLib/LocalToGlobalIndexMap.cpp
```diff
@@ -107,8 +107,8 @@ TEST_F(NumLibLocalToGlobalIndexMapTest, DISABLED_SubsetByComponent)
         new MeshLib::MeshSubsets{selected_subset.get()}};
 
     auto dof_map_subset = std::unique_ptr<NumLib::LocalToGlobalIndexMap>{
-        dof_map->deriveBoundaryConstrainedMap(1,  // variable id
-                                              0,  // component id
+        dof_map->deriveBoundaryConstrainedMap(1,    // variable id
+                                              {0},  // component id
```
TomFischer: Should the comment be updated: either 'component ids' or 'vector of component ids'?

Tests/NumLib/LocalToGlobalIndexMapMultiComponent.cpp
wenqing: Yes, add the answer as a comment.

endJunction: :white_check_mark:

NumLib/DOF/LocalToGlobalIndexMap.cpp
endJunction: Now the last one is moved, others are copied.

endJunction:
@chleh [Removal of unique_ptr's](https://github.com/endJunction/ogs/tree/DofTableSimplified) will come in a separate PR

chleh:
:+1:

Issue #1761, opened 2017-04-06
petsc, output files


JobstM:
I’m looking at a transient gw flow problem, solved by petsc. Partmesh is used to decompose the mesh. The OGS run is fine and the results look reasonable.
In order to evaluate the results and doing post-processing (including merging and converting to Tecplot), we have some problems with the *.pvtu, *.pvd and *.vtu files:

   1) There are a lot of files generated. Is it possible to define a sub-dir, in which all the files will be written?
    2) PVD: it contains links to the *.vtu files, which do not exist. Shouldn’t it link to the *.pvtu files?
    3) PVTU: it contains links to one and only one *vtu file (part_0). Shouldn’t it link to all *.vtu files of the corresponding time step?
    4) VTU: “ghost” elements and nodes are included. Is this really needed? However, for post-processing it would be fine, to get information about the overlap. We would like to get information about the global node/element number of each output node/element.
 
Proposal: 1) no output of ghost elements; 2) a new output variable in the *vtu file for each node, similar to the coordinates, which contains the corresponding global node number

endJunction:
1. You can specify the output directory via `-o` flag _e.g._
```sh
bin/ogs -o results test.prj
```

wenqing:
Answers:
1) There are a lot of files generated. Is it possible to define a sub-dir, in which all the files will be written?
Make a directory first, e.g. ``/home/xxx/example/results`` 
use command option ``-o`` to specify the output directory, e.g,
```
mpirun -np 5 ogs foo.prj -o /home/xxx/example/results
```
2) PVD: it contains links to the *.vtu files, which do not exist. Shouldn’t it link to the *.pvtu files?
Currently not.
3) PVTU: it contains links to one and only one *vtu file (part_0). Shouldn’t it link to all *.vtu files of the corresponding time step?
A new bug due to file name change.
4) VTU: “ghost” elements and nodes are included. Is this really needed? However, for post-processing it would be fine, to get information about the overlap. We would like to get information about the global node/element number of each output node/element.
Yes it it needed to have ghost nodes and partitions, otherwise, the geometry info of a partition is incomplete.
 
Proposal: 1) no output of ghost elements; 2) a new output variable in the *vtu file for each node, similar to the coordinates, which contains the corresponding global node number
 A binary global output is planed to offer such features.

JobstM:
Thanks for your answers.
Which files should be listed in the PVD in a parallel run? Why do not list the pvtu files here?

[...]

PR #1780, opened 2017-04-25
Fix PVTU output

TomFischer:
With this changes the pvtu files in the parallel benchmarks are written correctly. This fixes one issue of the issues collection https://github.com/ufz/ogs/issues/1761.

### DIFF
--- a/Applications/CLI/ogs.cpp
+++ b/Applications/CLI/ogs.cpp
@@ -13,6 +13,10 @@
 #include <chrono>
 #include <tclap/CmdLine.h>
 
+#ifdef USE_PETSC
+#include <vtkMPIController.h>
+#endif
+
 // BaseLib
 #include "BaseLib/BuildInfo.h"
 #include "BaseLib/ConfigTreeUtil.h"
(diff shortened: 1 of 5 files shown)


Applications/CLI/ogs.cpp
```diff
@@ -109,7 +113,11 @@ int main(int argc, char *argv[])
         {
             ApplicationsLib::LinearSolverLibrarySetup
                 linear_solver_library_setup(argc, argv);
-
+#if defined(USE_PETSC)
+            vtkMPIController* controller = vtkMPIController::New();
```
wenqing: Use vtk smart pointer as ``vtkSmartPointer<vtkMPIController> controller = vtkSmartPointer<vtkMPIController>::New();`` with including  <vtkSmartPointer.h>

Applications/CLI/ogs.cpp
```diff
@@ -152,6 +160,9 @@ int main(int argc, char *argv[])
             if (isInsituConfigured)
                 InSituLib::Finalize();
  #endif
+#if defined(USE_PETSC)
+            controller->Delete() ;
```
wenqing: Replace it with ``controller->Finalize(1);``

Applications/CLI/ogs.cpp
wenqing: I have tried to put this vtkMPIController stuff as a member of `` linear_solver_library_setup``,  it did not work. The place here for vtkMPIController stuff  makes correct pvtu  output.

PR #1783, opened 2017-04-25
[Travis] gcc 4.9

bilke:
Closes https://github.com/ufz/ogs/issues/1084


bilke:
Yes docker and travis are on gcc 4.9.

PR #1786, opened 2017-05-02
Fix PETSc norm computation

TomFischer:
Working on a test ...

### DIFF
--- a/NumLib/DOF/DOFTableUtil.cpp
+++ b/NumLib/DOF/DOFTableUtil.cpp
@@ -14,12 +14,11 @@ namespace NumLib
 {
 namespace
 {
-
 template <class CalculateNorm>
 double norm(GlobalVector const& x, unsigned const global_component,
-             LocalToGlobalIndexMap const& dof_table, MeshLib::Mesh const& mesh, CalculateNorm calculate_norm)
+            LocalToGlobalIndexMap const& dof_table, MeshLib::Mesh const& mesh,
+            CalculateNorm calculate_norm)
 {
-    // TODO that also includes ghost nodes.
     double res = 0.0;
     MeshLib::MeshSubsets const& mss = dof_table.getMeshSubsets(global_component);
     for (unsigned i=0; i<mss.size(); i++)
@@ -29,8 +28,8 @@ double norm(GlobalVector const& x, unsigned const global_component,
             continue;
         for (MeshLib::Node const* node : ms.getNodes())
         {
-            auto const value =
-                getNodalValue(x, mesh, dof_table, node->getID(), global_component);
+            auto const value = getNonGhostNodalValue(
+                x, mesh, dof_table, node->getID(), global_component);
 
             res = calculate_norm(res, value);
         }
@@ -46,6 +45,23 @@ double norm(GlobalVector const& x, unsigned const global_component,
 }
 
 } // anonymous namespace
+
+double getNonGhostNodalValue(GlobalVector const& x, MeshLib::Mesh const& mesh,
+                             NumLib::LocalToGlobalIndexMap const& dof_table,
+                             std::size_t const node_id,
+                             std::size_t const global_component_id)
+{
+    MeshLib::Location const l{mesh.getID(), MeshLib::MeshItemType::Node,
+                              node_id};
+
+    auto const index = dof_table.getGlobalIndex(l, global_component_id);
+    assert (index != NumLib::MeshComponentMap::nop);
+
+    if (index < 0)  // ghost node value
+        return 0.0;
+
+    return x.get(index);
+}
 
 double getNodalValue(GlobalVector const& x, MeshLib::Mesh const& mesh,
                      NumLib::LocalToGlobalIndexMap const& dof_table,
--- a/NumLib/DOF/DOFTableUtil.cpp
+++ b/NumLib/DOF/DOFTableUtil.cpp
@@ -36,7 +36,12 @@ double norm(GlobalVector const& x, unsigned const global_component,
         }
     }
 
-    // TODO for PETSc some global accumulation is necessary.
+#ifdef USE_PETSC
+    double global_result = 0.0;
+    MPI_Allreduce(&res, &global_result, 1, MPI_DOUBLE, MPI_SUM,
+                  PETSC_COMM_WORLD);
+    res = global_result;
+#endif
     return res;
 }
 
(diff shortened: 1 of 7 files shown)


NumLib/DOF/DOFTableUtil.cpp
```diff
@@ -36,7 +36,12 @@ double norm(GlobalVector const& x, unsigned const global_component,
         }
     }
 
-    // TODO for PETSc some global accumulation is necessary.
+#ifdef USE_PETSC
```
wenqing: This could be an inaccurate temporary solution  because the values of the entries of the global x are fetched individually and the values of the ghost nodes are not excluded.

One better solution under PETSc is that to replace the above norm computation code with
1. call  VecGetSubVector(Vec X,IS is,Vec *Y) to get a sub vector for a component, where IS can be set by using DOF table. Or call VecGetValues(Vec x,PetscInt ni,const PetscInt ix[],PetscScalar y[]) to get an array of a component.
2. compute the norm directly.

chleh: Just want to mention:
1. The local norms cannot simply be summed if one uses l-infinity norm
2. ghost nodes could be excluded in the loop above, maybe. Is there a performance impact if one only fetches data local to some rank?

TomFischer:
- Commit 052a38b addresses the issue of ghost nodes. Maybe some parallel tests are failing because of this change.
- Commit d2a8a0e corrects the norm calculation in the parallel case, i.e., the local results are exchanged and updated accordingly.

NumLib/DOF/DOFTableUtil.cpp
```diff
@@ -100,24 +86,62 @@ double norm(GlobalVector const& x, unsigned const global_component,
...
-        return norm(x, global_component, dof_table, mesh,
-                    [](double res, double value) {
-                        return std::max(res, std::abs(value));
-                    });
-    default:
-        OGS_FATAL("An invalid norm type has been passed.");
+        MeshLib::MeshSubset const& ms = mss.getMeshSubset(i);
+        if (ms.getMeshID() != mesh.getID())
+            continue;
+        switch (norm_type)
```
endJunction: you could use template specialization instead of run-time switch...

chleh: IMHO you could have kept most of the old code.
Isn't it just necessary to add the `#ifdef PETSC` part (and remove the return statements from the original switch-case?

chleh:
With your last two commits, the norm computation for me looks correct now. Is it hard to add a test for it?

NumLib/DOF/DOFTableUtil.cpp
```diff
@@ -100,24 +86,62 @@ double norm(GlobalVector const& x, unsigned const global_component,
...
+        {
```
wenqing: Although these computation  by fetching vector entries could be slow under PETSc, the norm is now correct. 

I will think about a version with  PETSc  routines later on.

NumLib/DOF/DOFTableUtil.cpp
TomFischer: In the previous code, in the PETSc part you need to know the norm type in order to call `MPI_Allreduce` with the correct 5th argument, i.e., `MPI_SUM` or `MPI_MAX`.

Tests/CMakeLists.txt
```diff
@@ -93,7 +93,7 @@ add_custom_target(tests-cleanup ${CMAKE_COMMAND} -E remove testrunner.xml)
 set_target_properties(tests-cleanup PROPERTIES FOLDER Testing)
 
 if(OGS_USE_PETSC)
-    set(TEST_FILTER_MPI --gtest_filter=-MPITest_Math.*)
+    set(TEST_FILTER_MPI --gtest_filter=-MPITest_Math.*:MPITest_NumLib.*)
```
endJunction: Would setting the filter to `-MPITest*` still work?
It would make two disjoint sets for `tests` and `tests_mpi` targets.

TomFischer:
@endJunction I can not reproduce the SEGV. But something is strange in the backtrace. I do not understand why `do_test` from 'TestComponentNormsUsingPETSc.cpp:105' calls `DOFTableData::DOFTableData` from 'TestComponentNorms.cpp:40' although there is a `DOFTableData' definition inside 'TestComponentNormsUsingPETSc.cpp'?

Tests/CMakeLists.txt
TomFischer: Changed to `-MPITest*`.

Tests/VectorUtils.h
```diff
@@ -24,4 +24,21 @@ void fillVectorRandomly(Vector& x)
     for (Index i = 0; i < size; ++i) {
         MathLib::setVector(x, i, rnd(random_number_generator));
     }
+#ifdef USE_PETSC
+    x.finalizeAssembly();
+#endif
+}
+
+template <typename Vector>
```
wenqing: For a global vector x filled with a constant a, you can use 
```
 MathLib::LinAlg::set(x, a);
```

Besides, this function is not used in this PR.

NumLib/DOF/DOFTableUtil.cpp
```diff
@@ -14,33 +14,101 @@ namespace NumLib
...
-             LocalToGlobalIndexMap const& dof_table, MeshLib::Mesh const& mesh, CalculateNorm calculate_norm)
+            LocalToGlobalIndexMap const& dof_table, MeshLib::Mesh const& mesh,
+            CalculateNorm calculate_norm)
 {
-    // TODO that also includes ghost nodes.
+#ifdef USE_PETSC
+    x.setLocalAccessibleVector();
+#endif
+
     double res = 0.0;
```
wenqing: It might be good to use a name with meaning, e.g., norm_result.

Tests/VectorUtils.h
TomFischer: You are right, I changed the implementation of the test and forgot to remove it. I will do it now.

NumLib/DOF/DOFTableUtil.cpp
TomFischer: I think it is clear that the result of the `norm` function is the 'norm'. I would prefer to leave it as it is.

wenqing: Ok.

TomFischer:
Thanks to the reviewers!

PR #1787, opened 2017-05-02
[MeL] Another Properties::getPropertyVectorNames().

TomFischer:
With the additional parameter 't' the returned names can be filtered by mesh item type of the `PropertyVector`.


chleh:
Looks innocent. ⏩ 
Maybe in the long run, just using `std::map<std::string, PropertyVectorBase*> _properties;` directly would be better.

PR #1788, opened 2017-05-02
Set initial guess for PETSc solver

TomFischer:


chleh:
⏩ if it runs.

wenqing:
⏩ if it runs.

MathLib/LinAlg/PETSc/PETScLinearSolver.cpp
```diff
@@ -63,7 +63,8 @@ PETScLinearSolver::PETScLinearSolver(const std::string /*prefix*/,
         KSPSetOptionsPrefix(_solver, prefix.c_str());
     }
 
-    KSPSetFromOptions(_solver); // set running time option
+    KSPSetFromOptions(_solver); // set run-time options
+    KSPSetInitialGuessNonzero(_solver, PETSC_TRUE);
```
TomFischer: Keep only this one, not before solve. Move before setting of the other options. (endJ)

endJunction:
Just a memo: The tests do not pass, because the [petsc convergence criterion](http://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/KSP/KSPConvergedDefault.html#KSPConvergedDefault) is dependent on the initial guess; pass if for residual r
      |r| < max(r_tol * |r_0|, abs_tol).
There is also a divergence criterion (d_tol);
I could not find the default values for r_tol, a_tol, d_tol...

TomFischer:
Following ctests are failing:
- 5 - ogs-ParallelFEM_GroundWaterFlow3D_DirichletBC-mpirun-vtkdiff
- 7 - ogs-ParallelFEM_GroundWaterFlow3D_NeumannBC-mpirun-vtkdiff

However, the comparison of the computed solution with the reference solution shows only small differences:

PR #1802, opened 2017-05-10
MSVC 2015 builds with Appveyor

bilke:
As MSVC 2015 is not tested on our own hardware anymore this PR fills that gap. A new check is added on every PR. Also this PR enables shared linking on Windows with VS 2017.

This builds the CLI, runs unit and benchmark tests with standard config. Please note that this already takes 50 minutes. I can enable building the GUI but then it will build even longer..

*Edit:* Now the GUI is build too and it takes about 27 mins. Do not know why it took so long before.

Issues:

- [ ] Conan packages are downloaded every time even as they are cached

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -44,7 +44,7 @@ before_build:
 build_script:
   - mkdir build & cd build
   - conan install ..
-  - cmake .. -G "Ninja" -DCMAKE_BUILD_TYPE=%configuration%
+  - cmake .. -G "Ninja" -DCMAKE_BUILD_TYPE=%configuration% -DBUILD_SHARED_LIBS=ON -DOGS_EIGEN_DYNAMIC_SHAPE_MATRICES=ON
   - cmake --build . --config %configuration%
 
 test_script:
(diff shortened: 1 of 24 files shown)


endJunction:
Very nice. Can you split the times for build, unit-tests, and ctests? I could not find the log for the build.

bilke:
The build was not started as there was a previous build running. See [the logs](https://ci.appveyor.com/project/LarsBilke/ogs/build/6.0.8.1) now. There is no timing information but the build itself takes almost the whole time (ProcessLib I look at you..).

endJunction:
Thinking aloud about compilation time: Not sure how that thing is build, but using shared libraries and dynamic eigen matrices, will speed up things.
If appveyor has persistent filesystem, ccache will help.

bilke:
ccache is not available on MSVC. I will try shared libs and dynamic eigen matrices..

endJunction:
It seems to me there is a mismatch for logog library between static and dynamic.

endJunction:
Another speedup recommendation: https://blogs.msdn.microsoft.com/vcblog/2016/10/26/recommendations-to-speed-c-builds-in-visual-studio/

bilke:
:rage4: `LINK : fatal error LNK1189: library limit of 65535 objects exceeded` :rage4:

Issue #1799, opened 2017-05-09
Crank-Nicolson: _delta_t is not initialized in first time step.


endJunction:
This leads to spurious errors in the TWOPHASE_FLOW_THERMAL process (see https://github.com/ufz/ogs/pull/1763)---the first process using it.

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/1799).

PR #1806, opened 2017-05-16
[PL] Pass initial delta_t to time discretization.

endJunction:
This fixes #https://github.com/ufz/ogs/issues/1799.

Kudos to @chleh!


chleh:
But I probably also programmed the bug.  ⏩

PR #1809, opened 2017-05-17
Search length configuration

TomFischer:
For setting boundary conditions it is necessary to search the corresponding mesh nodes to a given geometry. Now it is possible to configure the search length algorithm for each boundary condition seperately.


MeshGeoToolsLib/CreateSearchLength.h
```diff
@@ -0,0 +1,50 @@
...
+    {
+        //! \ogs_file_param{prj__search_length_algorithm__value}
+        double const length = config->getConfigParameter<double>("value");
+        return MeshGeoToolsLib::SearchLength(length);
+    }
+    if (type == "heuristic")
+    {
+        return MeshGeoToolsLib::HeuristicSearchLength(mesh);
+    }
+    else
```
endJunction: else not needed, because of returns in the if...

ProcessLib/BoundaryCondition/BoundaryCondition.cpp
```diff
@@ -10,53 +10,42 @@
...
-std::unique_ptr<BoundaryCondition> BoundaryConditionBuilder::createBoundaryCondition(
+std::unique_ptr<BoundaryCondition>
+BoundaryConditionBuilder::createBoundaryCondition(
     const BoundaryConditionConfig& config,
     const NumLib::LocalToGlobalIndexMap& dof_table, const MeshLib::Mesh& mesh,
     const int variable_id, const unsigned integration_order,
     const unsigned shapefunction_order,
     const std::vector<std::unique_ptr<ProcessLib::ParameterBase>>& parameters)
 {
-    MeshGeoToolsLib::SearchLength search_length_algorithm;
```
chleh: Is there a special reason for this code move?

MeshGeoToolsLib/CreateSearchLength.h
```diff
@@ -0,0 +1,50 @@
...
+#include "MeshGeoToolsLib/HeuristicSearchLength.h"
+#include "MeshGeoToolsLib/SearchLength.h"
+
+namespace MeshGeoToolsLib
+{
+MeshGeoToolsLib::SearchLength createSearchLengthAlgorithm(
+    BaseLib::ConfigTree const& external_config, MeshLib::Mesh const& mesh)
+{
+    boost::optional<BaseLib::ConfigTree> config =
+        //! \ogs_file_param{prj__search_length_algorithm}
```
chleh: Is that (`prj__search_length_algorithm`) the right place for documenting this?

MeshGeoToolsLib/CreateSearchLength.h
```diff
@@ -0,0 +1,50 @@
...
+
+    if (!config)
+        return MeshGeoToolsLib::SearchLength();
+
+    //! \ogs_file_param{prj__search_length_algorithm__type}
+    std::string const type = config->getConfigParameter<std::string>("type");
+
+    if (type == "fixed")
+    {
+        //! \ogs_file_param{prj__search_length_algorithm__value}
```
chleh: I guess the documentation will show that `prj__search_length_algorithm__value` is always mandatory, which is only true for the `fixed` case.

MeshGeoToolsLib/MeshNodeSearcher.h
```diff
@@ -142,9 +148,9 @@ class MeshNodeSearcher
     MeshGeoToolsLib::SearchLength _search_length_algorithm;
     bool _search_all_nodes;
     // with newer compiler we can omit to use a pointer here
-    std::vector<MeshNodesOnPoint*> _mesh_nodes_on_points;
-    std::vector<MeshNodesAlongPolyline*> _mesh_nodes_along_polylines;
-    std::vector<MeshNodesAlongSurface*> _mesh_nodes_along_surfaces;
+    mutable std::vector<MeshNodesOnPoint*> _mesh_nodes_on_points;
+    mutable std::vector<MeshNodesAlongPolyline*> _mesh_nodes_along_polylines;
+    mutable std::vector<MeshNodesAlongSurface*> _mesh_nodes_along_surfaces;
```
TomFischer: `_mesh_nodes_on_points`, `_mesh_nodes_along_polylines` and `_mesh_nodes_along_surfaces` are used as caches. These caches are the only internals that are modified by some methods. Until now, this fact has prevented that some of the methods are `const`.

MeshGeoToolsLib/MeshNodeSearcher.h
```diff
@@ -59,10 +59,9 @@ class MeshNodeSearcher
      * @param search_all_nodes switch between searching all mesh nodes and
      * searching the base nodes.
      */
-    MeshNodeSearcher(
-        MeshLib::Mesh const& mesh,
-        MeshGeoToolsLib::SearchLength const& search_length_algorithm,
-        bool search_all_nodes = true);
+    MeshNodeSearcher(MeshLib::Mesh const& mesh,
+                     MeshGeoToolsLib::SearchLength&& search_length_algorithm,
+                     bool search_all_nodes = true);
```
TomFischer: Yes, I will remove the default value.

MeshGeoToolsLib/MeshNodeSearcher.cpp
```diff
@@ -154,10 +154,26 @@ MeshNodeSearcher& MeshNodeSearcher::getMeshNodeSearcher(
     if (_mesh_node_searchers.size() < mesh_id+1)
         _mesh_node_searchers.resize(mesh_id+1);
 
-    if (!_mesh_node_searchers[mesh_id])
+    if (_mesh_node_searchers[mesh_id])
+    {
+        // recreate searcher if search length algorithm does not fit
```
TomFischer: Maybe my comment in the code is misleading. The existing searcher is only returned when it is from the same type (fixed or heuristic) and in the fixed case has the same internal search length. Else it is recreated. I will modify the comment ...

ProcessLib/BoundaryCondition/BoundaryCondition.cpp
TomFischer: The `SearchLength` object can only be created after parsing the boundary condition. So the creation is moved there.

MeshGeoToolsLib/MeshNodeSearcher.h
chleh: OK, so it's mainly for stylistic reasons. Thanks

ProcessLib/BoundaryCondition/BoundaryCondition.cpp
chleh: Ah, OK.

MeshGeoToolsLib/CreateSearchLength.h
```diff
@@ -11,22 +11,25 @@
...
+
 #include "BaseLib/ConfigTree.h"
 
 #include "MeshGeoToolsLib/HeuristicSearchLength.h"
 #include "MeshGeoToolsLib/SearchLength.h"
 
 namespace MeshGeoToolsLib
 {
-MeshGeoToolsLib::SearchLength createSearchLengthAlgorithm(
+std::unique_ptr<MeshGeoToolsLib::SearchLength> createSearchLengthAlgorithm(
```
chleh: I guess this change now is only to fit better into our current style of writing create...() functions?

endJunction: Could be, but the actual reason is that HeuristicSearchLength (which is derived from SearchLength) could be returned here and this would be an implicit cast. Pointer or reference type is needed to keep the polymorphic behaviour.

endJunction:
@TomFischer Any ideas why the tetgen tests are failing?

bilke:
envinf1 was cleaned up a little bit (`/data/ogs` had a disk usage of 2 TB). But there is problem with docker which fails all builds:

```
 Failed to kill container '604368d2d8021333bec35823f056ea289498df08d4f4196cd1836e50b9d7b31a'.
```

@endJunction Can you please check your machine (maybe a restart fixes this)?

bilke:
Jenkins is green.

MeshGeoToolsLib/CreateSearchLength.h
chleh: OK, I see. Looks like another of these C++ quirks: At least to me it's not obvious why C++ supports such constructions: virtual methods which are dispatched dynamically only if pointers or references are used.

Maybe then `HeuristicSearchLength` shouldn't be derived from `SearchLength`, but both should have a common abstratct base class.

PR #1810, opened 2017-05-17
[Jenkins] envinf1 environment updated with gcc 6.2.

bilke:


chleh:
Thanks for your effort updating the test environment.

But I have some complaint: In #1084 we discussed the compiler update issue. There @endJunction suggested "to set the minimum to ubuntu's 16.04 gcc-5.3 (and clang-3.8).". And I said, that we should test the minimum requirement (and not the rather arbitrary choice 6.2). Of course there might be other machines testing the minimum, but I think that the minimum should be most thoroughly tested.

Saying that, I don't want to blame this PR, but I blame our decision-making process: in #1084 something was discussed, no final conclusion was fixed, and now the work of somebody establishes a new status quo. Maybe we can improve upon that in the future.

Slightly Different topic: In #1084 @endJunction stated that the ogs executables don't work correctly when compiled with gcc-4.8. Subsequently in #1783 the gcc version on travis was raised. But the cmake config still says that gcc-4.8 is the minimum requirement. Should there be an update for our cmake scripts? See https://github.com/ufz/ogs/blob/master/scripts/cmake/CompilerSetup.cmake#L55

endJunction:
-  The minimum is tested (gcc-4.9 in docker and travis, and VisualStudio15 in appveyor).
 - In the scripts the gcc-4.8 is still there because of envinf testing node (which will use gcc-6.2 from now on) => @bilke please update the script.
 - The decision making strategy is not clear, that's true and mostly my fault.

So it is still open if the minimum should be gcc-5.3, but because of necessity (some issues with gcc-4.8) now gcc-4.9 is tested (and with the envinf node update becomes the required minimum).
The choice of gcc-6.2 is because any update requires recompilation of vtk and openmpi. This will be not necessary for the next year or so, and we can build the test environment around gcc-6.2.

I hope this clarifies some of the hidden decisions.

bilke:
Dimas statements are correct. I picked 6.2 for envinf1 as it is the latest and greatest available compiler there that maybe produces faster code which is all you want when running time-consuming simulations on the cluster. So 4.9 is the new minimum. I will update the CMake check.

chleh:
Thank you for clarifying this. So when the tests succeed I'm fine with this PR.
Maybe we can have some regular short meetings again in the future, where we can briefly discuss all upcoming changes.

bilke:
@endJunction Can you please have a look why the following benchmark fails:

```
The following tests FAILED:
16:04:45 	  5 - ogs-ParallelFEM_GroundWaterFlow3D_DirichletBC-mpirun-vtkdiff (Failed)
```

wenqing:
:+1: 
It is Ok to increase the vtkdiff tolerance for the two tests.

endJunction:
Eventually I have updated the results for the new petsc and set tighter linear solver tolerance yielding almost analytical solution.

PR #1811, opened 2017-05-17
Use c++14.

endJunction:
With recent compiler improvements I hope to be able to use the three-year-old standard.

### DIFF
--- a/NumLib/ODESolver/TimeDiscretizationBuilder.cpp
+++ b/NumLib/ODESolver/TimeDiscretizationBuilder.cpp
@@ -19,28 +19,28 @@ std::unique_ptr<TimeDiscretization> createTimeDiscretization(
     //! \ogs_file_param{prj__time_loop__processes__process__time_discretization__type}
     auto const type = config.getConfigParameter<std::string>("type");
 
+    //! \ogs_file_param_special{prj__time_loop__processes__process__time_discretization__BackwardEuler}
     if (type == "BackwardEuler")
     {
-        //! \ogs_file_param_special{prj__time_loop__processes__process__time_discretization__BackwardEuler}
         return std::make_unique<BackwardEuler>();
     }
+    //! \ogs_file_param_special{prj__time_loop__processes__process__time_discretization__ForwardEuler}
     else if (type == "ForwardEuler")
     {
-        //! \ogs_file_param_special{prj__time_loop__processes__process__time_discretization__ForwardEuler}
         return std::make_unique<ForwardEuler>();
     }
+    //! \ogs_file_param_special{prj__time_loop__processes__process__time_discretization__CrankNicolson}
     else if (type == "CrankNicolson")
     {
         //! \ogs_file_param{prj__time_loop__processes__process__time_discretization__CrankNicolson__theta}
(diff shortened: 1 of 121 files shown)


endJunction:
So, this works on VisualStudio15 too. Of course not all features can be used, but this is still good news.
constexpr, make_unique_ptr, templated lambdas maybe...

chleh:
Could you also add a commit here that uses C++14 features, e.g. using `make_unique` everywhere? Otherwise it's easy for a compiler to claim that it supports C++14. :smile:

endJunction:
@chleh The unique pointer is not available in gcc-4.8.1, so this is based on the new envinf environment.

chleh:
Very nice!

But should we really change `... .reset(new ...)` to `... = make_unique<...>(...)`?

endJunction:
clang-tidy's automatic changes.

NumLib/ODESolver/TimeDiscretizationBuilder.cpp
```diff
@@ -16,36 +16,32 @@ namespace NumLib
...
-        return T(new ConcreteTD);
+        return std::make_unique<ForwardEuler>();
     }
     else if (type == "CrankNicolson")
     {
         //! \ogs_file_param{prj__time_loop__processes__process__time_discretization__CrankNicolson__theta}
         auto const theta = config.getConfigParameter<double>("theta");
-        using ConcreteTD = CrankNicolson;
-        return T(new ConcreteTD(theta));
+        //! \ogs_file_param_special{prj__time_loop__processes__process__time_discretization__CrankNicolson}
```
chleh: Already documented two lines above as far as i can see.

NumLib/ODESolver/TimeDiscretizationBuilder.cpp
```diff
@@ -16,36 +16,32 @@ namespace NumLib
...
+        return std::make_unique<CrankNicolson>(theta);
     }
     else if (type == "BackwardDifferentiationFormula")
     {
         //! \ogs_file_param{prj__time_loop__processes__process__time_discretization__BackwardDifferentiationFormula__order}
         auto const order = config.getConfigParameter<unsigned>("order");
-        using ConcreteTD = BackwardDifferentiationFormula;
-        return T(new ConcreteTD(order));
+        //! \ogs_file_param_special{prj__time_loop__processes__process__time_discretization__BackwardDifferentiationFormula}
```
chleh: Likewise. Note: Documentation has to be put above `config.getConfig...(...);` lines. It has to be there and only there. (If I'm not wrong.)

wenqing:
:+1: .
It's OK to increase the vtkdiff tolerance for ParallelFEM_GroundWaterFlow3D_DirichletBC.

NumLib/ODESolver/TimeDiscretizationBuilder.cpp
chleh: Clarifeid in discussion I was wrong. But this probably solves #1816.

chleh:
:+1:

endJunction:
:fireworks: :confetti_ball: :tada:

PR #1813, opened 2017-05-19
Mechanical phasefield process

xingyuanmiao:
The mechanical phase-field process was implemented. Currently, the implementation has a 3D test for crack beam under tension. Next step will be adding two longer tests: tension of single-edge-notch test; shear of single-edge-notch test .


ProcessLib/PhaseField/PhaseFieldFEM.h
```diff
@@ -0,0 +1,609 @@
...
+            Kud.noalias() += B.transpose() * 2 * d_ip * sigma_tensile * N * w;
+
+            double const d_dot_ip = N.dot(d_dot);
+
+            if (history_variable_prev < strain_energy_tensile)
+            {
+                history_variable = strain_energy_tensile;
+                Kdu.noalias() = Kud.transpose();
+            }
+            else
```
nagelt: If you have drastic convergence issues you can decouple the system (and restore convexity) by updating H only once a global increment has converged. Then K_du is always zero. But I guess this would be better done via staggered coupling and controlling the coupling iterations, as in that case monolithic coupling only serves to increase the equation system.
So: Nothing to do here.
@endJunction FYI as an update on today's discussion. Btw, how far is staggered scheme/coupling manager? Processes of immediate interest for sequential coupling are T-(H)M, Phase field.

ProcessLib/PhaseField/PhaseFieldFEM.h
```diff
@@ -0,0 +1,609 @@
...
+
+    std::vector<double> const& getIntPtEpsilon(
+        std::vector<double>& cache, std::size_t const component) const
+    {
+        cache.clear();
+        cache.reserve(_ip_data.size());
+
+        for (auto const& ip_data : _ip_data)
+        {
+            cache.push_back(ip_data.eps[component]);
```
nagelt: Kelvin mapping missing (see stress routine).

web/content/docs/benchmarks/phase-field/phasefield.md
```diff
@@ -0,0 +1,47 @@
...
+
+We solve a homogeneous beam model under a given displacement loading. The length of the beam is 2\,mm. Detailed model description can refer [this PDF](../Miao_Biot2017Fullpaper.pdf).
+## Results and evaluation
+
+Result showing crack phase-field and displacement field distributions through the length of the beam:
+
+{{< img src="../beam.png" >}}
+{{< img src="../beam_d.png" >}}
+{{< img src="../beam_u.png" >}}
+
```
nagelt: again, the scaling of the images makes it difficult to see anything ...

ProcessLib/PhaseField/PhaseFieldFEM.h
```diff
@@ -0,0 +1,609 @@
...
+                             double const /*t*/,
+                             double const /*delta_t*/) override
+    {
+        unsigned const n_integration_points =
+            _integration_method.getNumberOfPoints();
+
+        for (unsigned ip = 0; ip < n_integration_points; ip++)
+        {
+            _ip_data[ip].pushBackState();
+            if (_ip_data[ip].history_variable_prev <
```
nagelt: The if is not needed. Push back probably always has to happen. The if is done above and determines which value is in history_variable (either history_variable_prev or the new tensile SED).

endJunction:
@xingyuanmiao You need to add aligned memory allocation like I have shown you in https://github.com/endJunction/ogs/commit/562f0ae1aae7076c141dbb1d5641f5706972f084
These are used in all current processes (and required), otherwise the program crashes; Cannot do performance analysis on this until this is fixed.

nagelt:
@xingyuanmiao Has this alignment issue been fixed?
@endJunction Do we have a performance test?

endJunction:
Closed as discussed in the developer meeting.

PR #1819, opened 2017-05-29
Don't use else after return

endJunction:
clang-tidy's automatic changes. Same code with less code ;)

PR #1821, opened 2017-05-31
Do not use else after return. II

endJunction:
Just discovered that applying clang-tidy repeatedly is required to get all of the changes.

This is follow up of https://github.com/ufz/ogs/pull/1819


chleh:
:fas

PR #1827, opened 2017-06-02
Code coverage Jenkins job

bilke:
- Added Jenkins job testing code coverage on testrunner
- Removed cmake scripts in `scripts/cmake/cmake` and updated the repo at https://github.com/bilke/cmake-modules
- Pointed submodule `ThirdParty/cmake-modules` to https://github.com/bilke/cmake-modules

PR #1828, opened 2017-06-07
[CMake] Run git submodule sync in CMake

bilke:
This fixes submodule checkout errors when the submodules URL changed like this:

```
-- Updating submodule ThirdParty/cmake-modules
error: no such remote ref af334f92548b7349a1bb038d56ce91fc7b3d3d00
Fetched in submodule path 'ThirdParty/cmake-modules', but it did not contain af334f92548b7349a1bb038d56ce91fc7b3d3d00. Direct fetching of that commit failed.
```

Fixed also a CMake error when CVODE_KLU variable was not defined:

```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
CVODE_KLU
    linked by target "MathLib" in directory /home/naumov/w/ogs/s/MathLib

-- Configuring incomplete, errors occurred!
```

See #1827.

### DIFF
--- a/scripts/cmake/SubmoduleSetup.cmake
+++ b/scripts/cmake/SubmoduleSetup.cmake
@@ -25,6 +25,13 @@ if(OGS_BUILD_SWMM)
     list(APPEND REQUIRED_SUBMODULES ThirdParty/SwmmInterface)
 endif()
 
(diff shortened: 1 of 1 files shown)


grubbymoon:
Thanks !

PR #1834, opened 2017-06-12
Removed some compilation warnings

wenqing:
as titled.

### DIFF
--- a/Tests/NumLib/TestODEInt.cpp
+++ b/Tests/NumLib/TestODEInt.cpp
@@ -254,7 +254,8 @@ public:
         for (std::size_t i = 0; i < sol_picard.ts.size(); ++i)
         {
             ASSERT_EQ(sol_picard.ts[i], sol_newton.ts[i]);
-            for (std::size_t comp = 0; comp < sol_picard.solutions[i].size();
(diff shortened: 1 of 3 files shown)


Tests/NumLib/TestODEInt.cpp
```diff
@@ -254,7 +254,7 @@ class NumLibODEIntTyped : public ::testing::Test
         for (std::size_t i = 0; i < sol_picard.ts.size(); ++i)
         {
             ASSERT_EQ(sol_picard.ts[i], sol_newton.ts[i]);
-            for (std::size_t comp = 0; comp < sol_picard.solutions[i].size();
+            for (int comp = 0; comp < static_cast<int>(sol_picard.solutions[i].size());
```
wenqing: With PETScVector,  the type of sol_picard.solutions[i].size() is PetscInt, while the type is std::size_t for other vector types.

endJunction: Wouldn't be `GlobalIndexType` better?

wenqing: With GlobalIndexType,  static_cast is still needed.

Issue #1765, opened 2017-04-10
petsc, deformation process


JobstM:
I tried ogs6-petsc with a mechanical problem (sd). With single process, the results look fine. Using partitioning (with partmesh -q), I get the errors:
error: Newton: Could not solve the given nonlinear system within 5 iterations
error: The nonlinear solver failed in time step #1 at t = 0.25 s for process #0.
What can I do?

endJunction:
Could you try to increase the allowed number of iterations for the Newton? Have a look at our [documentation](
https://doxygen.opengeosys.org/ogs_file_param__prj__nonlinear_solvers__nonlinear_solver__max_iter.html) too.

The tests are (usually/sometimes) set to smallest possible number of iterations, so we can detect, and have also to check, if any deviations were introduced by changes.

JobstM:
I used the example disc_with_hole. I tried with 50 non-linear iterations - no change.
As far as I understood, the process should be linear. When ogs6-petsc runs on one core, I see no message from the non-linear solver and the results look fine.

wenqing:
Sounds there is a bug. I will debug the code soon for the reason.
PS: The  option under test,  ``-q``,  is for HM with the case of linear element for H and quadratic element for M. For pure M problem, this option is not required.

JobstM:
I see, thanks.
Does this mean, pure M problems are calculated on linear shape functions on default?

endJunction:
This is not entirely correct; the shapefunction order depends on the mesh (can not be higher than the mesh element order) and the project file configuration for the process variable: see [documentation](https://doxygen.opengeosys.org/ogs_file_param__prj__process_variables__process_variable__order.html).

There are tests, where the linear elements are sufficient, _e.g._ the `disc_with_hole.prj`.

wenqing:
It works for deformation process now.

JobstM:
Great!

We will test it soon!


Von: wenqing [mailto:notifications@github.com]
Gesendet: Thursday, June 29, 2017 10:52 AM
An: ufz/ogs <ogs@noreply.github.com>
Cc: Maßmann, Jobst <Jobst.Massmann@bgr.de>; Author <author@noreply.github.com>
Betreff: Re: [ufz/ogs] petsc, deformation process (#1765)


It works for deformation process now.

wenqing:
But not for HydroMechanicsProcess and ThermoMechanicsProcess using both of  linear and quadratic elements.

chleh:
[...]

PR #1797, opened 2017-05-05
Fixed a bug in PETScVector

wenqing:
The bug is in PETScVector::get, which calls VecGetValues. VecGetValues  can only get values on the same processor. To fix the bug, the current implementation introduces ~a scattered global array to store the values of the vector~ an array,  ``_entry_array``,  to store the entries of the vector fetched across partitions. 

``_entry_array`` contains
 either
  *the entries of the global vector owned by the current rank if the vector is created with given ghost IDs*
or
*all entries of the global vector if the vector is created without ghost IDs*

``_entry_array`` only gets memory if entries of the vector are going to be fetched. For the vector created with given ghost IDs, a lookup table (by std::map<PetscInt, PetscInt>) is created to map the global indices of the ghost entries to local ones if ``_entry_array`` gets memory. By this way, the memory usage is saved.
~TODO: Using PetscScalar* PETScVector::getLocalVector()  (call VecGhostGetLocalForm) instead of a scattered global array. To do this, local node indices of partitions have to be added to DOF table.~
Therefore the local indices of the ghost entries are computed on the fly.

Note: In the current implementation of PDE solving part, all global vectors are now created with given ghost IDs. Therefore in the computation of physical process modelling, ``_entry_array``  is local within partition.

PR #1838, opened 2017-06-14
Enabled parallel computing with PETSc in deformation analysis

wenqing:
as titled.

File SmallDeformationCommon/Common.h is removed, and its associated part is removed as well. With the functions in that file, the nodal force vector computation is repeated.

If the removed part is needed, commit 
 	[pDM] Corrected getNodalForces for parallel computing 			4289841
, which corrects the parallel nodal force computation,  will be selected.

A test with linear element is added to the Ctest list.

testparalleldm


wenqing:
To resolve  issue #1765.
Future work: secondary variable output under MPI.

wenqing:
This PR follows up PR #1797.

endJunction:
Extending of mechanics processes for parallel runs is great. We could use some speedup along this lines.
Removing nodal force computation is not an option at all. There will be material forces added to the same interface.

wenqing:
Added Common.h again. The file has been changed for parallel computing, however the functions in it are not used in this PR after commit  4289841.

ProcessLib/SmallDeformation/SmallDeformationFEM.h
```diff
@@ -88,7 +88,6 @@ struct SecondaryData
 
 struct SmallDeformationLocalAssemblerInterface
     : public ProcessLib::LocalAssemblerInterface,
-      public ProcessLib::SmallDeformation::NodalForceCalculationInterface,
```
endJunction: Could you please explain why the interface is removed?

endJunction:
I'm not sure about the changes to the computation of the nodal forces. When source terms would be added to the process (e.g. gravity) would this require changes to the nodal forces computation?

Independent of this, could you please drop the commit which is removing the "Common.h" file?

The independent calculation of the nodal forces allows an integration into other (small deformation) processes like HM. Now this become more difficult I think.

Let's discuss this before more time is invested in the changes.

endJunction:
I have compared the provided `pvtu` result with the result of the serial run with vtkdiff, and it shows a rather large deviation:
```
Computed difference between data arrays:
abs l1 norm      = 6.393239226763097e+01
abs l2-norm^2    = 2.619448994702831e+00
abs l2-norm      = 1.618471190569307e+00
abs maximum norm = 1.105040679911180e-01
```

wenqing:
@endJunction  From the point view of computational mechanics, which originated the FEM,  the discretized weak form terms of all source terms and also  coupled terms in the momentum balance equation are added to nodal force to have momentum balance numerically. Therefore using the RHS of the system of equation of deformation process as the nodal force is corrected except for the case of M being coupled to other processes and the monolithic scheme being applied.  Under such consideration, I think, we only need to computing the extra contribution to the nodal force from the coupled terms in the process classes of HM, TM or THM with monolithic scheme.

wenqing:
or using the current nodal force computation in Common.h but add computation of total stress in HydroMechanicsFEM  and other similar assemblers.

ProcessLib/SmallDeformation/SmallDeformationProcess.h
```diff
@@ -201,12 +202,12 @@ class SmallDeformationProcess final : public Process
...
-    void postTimestepConcreteProcess(GlobalVector const&) override
+    void postTimestepConcreteProcess(GlobalVector const& /*x*/) override
     {
         DBUG("PostTimestep SmallDeformationProcess.");
-
-        ProcessLib::SmallDeformation::writeNodalForces(
-            *_nodal_forces, _local_assemblers, *_local_to_global_index_map);
+        std::transform(_nodal_forces->begin(), _nodal_forces->end(),
+            _nodal_forces->begin(),
+            std::bind1st(std::multiplies<double>(), -1.0));
```
wenqing:
```c++
transform(n.b, n.e, o.b, [](double v) { return -v;});
```

ProcessLib/SmallDeformation/SmallDeformationProcess.h
```diff
@@ -186,6 +185,10 @@ class SmallDeformationProcess final : public Process
             _global_assembler, &VectorMatrixAssembler::assembleWithJacobian,
             _local_assemblers, *_local_to_global_index_map, t, x, xdot,
             dxdot_dx, dx_dx, M, K, b, Jac, coupling_term);
+
+        b.copyValues(*_nodal_forces);
+        std::transform(_nodal_forces->begin(), _nodal_forces->end(),
+            _nodal_forces->begin(), [](double val) { return -val;});
```
wenqing: Setting negative values of ``_nodal_forces`` is moved here.

ProcessLib/SmallDeformation/SmallDeformationProcess.h
```diff
@@ -201,14 +204,6 @@ class SmallDeformationProcess final : public Process
             _local_assemblers, *_local_to_global_index_map, x, t, dt);
     }
 
-    void postTimestepConcreteProcess(GlobalVector const&) override
```
wenqing: Removed this overloaded member.

ProcessLib/SmallDeformation/SmallDeformationProcess.h
wenqing: Changed by amending the second commit.

wenqing:
Data pointer has been updated.

PR #1852, opened 2017-06-23
Remove MSVC ifdefs for version less than 19.00 = VisualStudio2015

endJunction:
Just looking if the `#ifdefs` for old VisualStudio can be removed.


endJunction:
So, this seems to work fine with VisualStudio 2015. Less code, less workarounds!

PR #1853, opened 2017-06-23
Fix two parallel tests

wenqing:
One test needs to change file names.
The failing of another test is caused by allowing setting nonzero  initial guess for ksp solve.


NumLib/ODESolver/NonlinearSolver.cpp
```diff
@@ -208,6 +207,8 @@ bool NonlinearSolver<NonlinearSolverTag::Newton>::solve(
         sys.getJacobian(J);
         INFO("[time] Assembly took %g s.", time_assembly.elapsed());
 
+        minus_delta_x.setZero();
```
wenqing: set dx = 0 for each linearization step.

endJunction: :+1:

wenqing:
All tests run. This PR is going to be merged soon.

PR #1860, opened 2017-06-30
[Mat/Fracture] support 3d in MC model

norihiro-w:
as titled, though LIE isn't supporting 3d yet.

### DIFF
--- a/MaterialLib/FractureModels/MohrCoulomb.cpp
+++ b/MaterialLib/FractureModels/MohrCoulomb.cpp
@@ -68,9 +68,8 @@ void MohrCoulomb<DisplacementDim>::computeConstitutiveRelation(
     Ke(index_ns, index_ns) = mat.Kn;
 
     sigma.noalias() = sigma_prev + Ke * dw;
-    double const tau = (DisplacementDim==2) ? sigma[0] : std::sqrt(sigma[0]*sigma[0] + sigma[1]*sigma[1]);
-    double const sigma_n = sigma[index_ns];
 
+    double const sigma_n = sigma[index_ns];
     // if opening
     if (sigma_n > 0)
     {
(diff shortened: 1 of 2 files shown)


MaterialLib/FractureModels/MohrCoulomb.cpp
```diff
@@ -82,22 +81,34 @@ void MohrCoulomb<DisplacementDim>::computeConstitutiveRelation(
...
+
+    if (DisplacementDim == 2)
+    {
+        dFs_dS[0] = boost::math::sign(tau);
+        dQs_dS[0] = boost::math::sign(tau);
+    }
+    else
+    {
+        for (int i=0; i<index_ns-1; i++)
+            dFs_dS[i] = boost::math::sign(tau) / tau * sigma[i];
```
endJunction: Here `sigma[i] / std::abs(tau)` produces less assembly, and I believe it is equivalent. In case tau = 0 the expression is not defined anyway...

MaterialLib/FractureModels/MohrCoulomb.cpp
```diff
@@ -82,22 +81,34 @@ void MohrCoulomb<DisplacementDim>::computeConstitutiveRelation(
...
+    if (DisplacementDim == 2)
+    {
+        dFs_dS[0] = boost::math::sign(tau);
+        dQs_dS[0] = boost::math::sign(tau);
+    }
+    else
+    {
+        for (int i=0; i<index_ns; i++)
+            dFs_dS[i] = sigma[i] / std::abs(tau);
+
```
nagelt: tau seems to be non-negative in the 3D case (sqrt); no abs needed? Or rather this: If I see right, tau could in general be given the absolute value in line 71 (it is for 3D). This would correspond better to its meaning as a "net" shear stress and would make this if/else statement here (line 97ff) unnecessary. The statements in the 3D loop would then also work for the 2D case. Also the abs in line 84 would no longer be needed.

norihiro-w: improved

bilke:
Merging now. Jenkins is red for unknown reasons, everything seems to be fine.

PR #1862, opened 2017-06-30
Added two more tests for the staggered scheme with adaptive time stepping

wenqing:
as tiled.

One test has temperature dependent liquid density and constant viscosity, another test gets both liquid density and viscosity being temperature dependent.

Difference from what in HT: Temperature dependent fluid density always gives thermal expansion coefficient in LiquidProcess. On the contrary,  thermal expansion coefficient  is not available in HTProcess.

**Temperature perturbation result of that with temperature dependent liquid density and constant viscosity:**

t_pert


wenqing:
will be merged to PR #1867 .

PR #1864, opened 2017-06-30
Force the program abort when nonlinear solver fails during FixedTimeStepping

wenqing:
as titled.

### DIFF
--- a/ProcessLib/UncoupledProcessesTimeLoop.cpp
+++ b/ProcessLib/UncoupledProcessesTimeLoop.cpp
@@ -9,6 +9,7 @@
 
 #include "UncoupledProcessesTimeLoop.h"
 
+#include "BaseLib/Error.h"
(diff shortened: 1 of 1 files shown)


ProcessLib/UncoupledProcessesTimeLoop.cpp
```diff
@@ -840,6 +846,12 @@ bool UncoupledProcessesTimeLoop::solveUncoupledEquationSystems(
             _output->doOutputAlways(pcs, spd->process_output, timestep_id, t,
                                     x);
 
+            auto& timestepper = spd->timestepper;
+            if (!timestepper->isSolutionErrorComputationNeeded())
```
wenqing: There are only two time stepping algorithms available. If more  time stepping algorithms added, this condition will be changed.

PR #1866, opened 2017-07-04
Patches for adaptive time stepping

wenqing:
This PR contains updates or bugfixing for the adaptive time stepping implementation:

- stop program when a rejected time step gets an unchanged  new step size if the previous step is also rejected.
- reset the initial guess of global x vector to its old value if a time step is rejected. Actually only the Picard  nonlinear solver needs  this computation.
- only display the accepted time step number as step number.

Besides, a test for the case of Richards flow with adaptive time stepping is also included in this PR.

### DIFF
--- a/NumLib/ODESolver/TimeDiscretization.h
+++ b/NumLib/ODESolver/TimeDiscretization.h
@@ -142,6 +142,14 @@ public:
     virtual void pushState(const double t, GlobalVector const& x,
                            InternalMatrixStorage const& strg) = 0;
 
+    /*! Indicate that the current timestep is rejected and that the computation
+     *  of the present time step will be restarted with a new time step size.
+     *
+     * \param x    The solution at the current time step, which is going to be
+     *             reset to its previous value.
+     */
+    virtual void popState(GlobalVector& x) = 0;
+
     /*! Indicate that the computation of a new timestep is being started now.
      *
      * \warning Currently changing timestep sizes are not supported. Thus,
@@ -262,6 +270,11 @@ public:
         MathLib::LinAlg::copy(x, _x_old);
     }
 
+    void popState(GlobalVector& x) override
+    {
+        MathLib::LinAlg::copy(_x_old, x);
+    }
+
     void nextTimestep(const double t, const double delta_t) override
     {
         _t = t;
@@ -312,6 +325,11 @@ public:
                    InternalMatrixStorage const&) override
     {
         MathLib::LinAlg::copy(x, _x_old);
+    }
+
+    void popState(GlobalVector& x) override
+    {
+        MathLib::LinAlg::copy(_x_old, x);
     }
 
     void nextTimestep(const double t, const double delta_t) override
@@ -392,6 +410,11 @@ public:
         strg.pushMatrices();
     }
 
+    void popState(GlobalVector& x) override
+    {
+        MathLib::LinAlg::copy(_x_old, x);
+    }
+
     void nextTimestep(const double t, const double delta_t) override
     {
         _t = t;
@@ -462,6 +485,11 @@ public:
 
     void pushState(const double, GlobalVector const& x,
                    InternalMatrixStorage const&) override;
+
+    void popState(GlobalVector& x) override
+    {
+        MathLib::LinAlg::copy(*_xs_old[_xs_old.size()-1], x);
+    }
 
     void nextTimestep(const double t, const double delta_t) override
     {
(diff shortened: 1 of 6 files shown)


wenqing:
@waltherm The updates are here, and I am going to test your problem with Tom's branch.

NumLib/ODESolver/TimeDiscretization.h
```diff
@@ -142,6 +142,17 @@ class TimeDiscretization
     virtual void pushState(const double t, GlobalVector const& x,
                            InternalMatrixStorage const& strg) = 0;
 
+    /*! Indicate that the current timestep is rejected and that the computation
+     *  of the present time step will be restarted with a new time step size.
+     *
+     * \param t    The current time step.
```
endJunction: t seems to be unused in all functions...

wenqing: Oh it is! Will be removed.

wenqing: removed `t` by amending the commit of ``[dt] Set current x to old x when a step is ...``.

waltherm:
@wenqing thanks, let me know, if I can help and test something...

NumLib/ODESolver/TimeDiscretization.h
```diff
@@ -142,6 +142,14 @@ class TimeDiscretization
     virtual void pushState(const double t, GlobalVector const& x,
                            InternalMatrixStorage const& strg) = 0;
 
+    /*! Indicate that the current timestep is rejected and that the computation
+     *  of the present time step will be restarted with a new time step size.
```
endJunction: Sounds a little misleading to me: The function does not indicate anything. Maybe:
> Restores previous time step result into the given vector.
> Used for repeating of the time step with new time step size when the current timestep is rejected.

endJunction:
@waltherm Please have a look if this resolves the time stepping problems for your test cases. Then it can be merged.

waltherm:
@endJunction will do and report.

wenqing:
@waltherm  By took Tom's branch and rebased on this PR, I just tested your example of RichardsComponentTransportProcess with the following time stepping input
```
                 <time_stepping>
                     <type>EvolutionaryPIDcontroller</type>
                     <t_initial> 0.0 </t_initial>
                     <t_end> 20 </t_end>
                     <dt_guess> 0.1 </dt_guess>
                     <dt_min> 0.01 </dt_min>
                     <dt_max> 5 </dt_max>
                     <rel_dt_min> 0.01 </rel_dt_min>
                     <rel_dt_max> 5 </rel_dt_max>
                     <tol> 0.2 </tol>
                 </time_stepping>
```
You can also use PerComponentDeltaX for convergence criterion.

waltherm:
Thanks, @wenqing for the patches. Here are some comments.

Issues **solved**:
- ogs stops now, if the time step is repeated, but the minimum time step size was used.
- Warning message for time step repetition now correctly refers to current time step number.

Issues **open**:
- There is no additional information why a time step was rejected, although the iteration converged.
- Repetitions of a time step are not marked with a repetition index (not really important).

Wording **suggestion**:
- "warning: Time step 1 is rejected. The computation is back to the previous time." 
-> "warning: Time step 1 was rejected and will be repeated with a reduced time step size."
- "error: 	This time step is rejected and the new computed step size is the same as that was just used. Suggest to adjust the parameters of the time stepper or try other time stepper." 
-> "error: Minimum time step size was rejected. Simulation is halted."

**General**: 
- I tried to run a simulation with RichardsComponentTransport and varied the maximum time step size. The time step algorithm will increase the time step to really large values, although the simulation shows clear temporal and spatial oscillations of the primary variable. Practically, this makes automatic time stepping not very useful. How can we solve this? Here are figures for the oscillations: 
spatial_osc
temp_osc

NumLib/ODESolver/TimeDiscretization.h
wenqing: Changed.

wenqing:
@waltherm Added a commit about what ``Issues open `` and  ``Wording suggestion``in you comment.
The increase ratio of step size can be changed by setting the tag of ``<rel_dt_max>  </rel_dt_max>``.

waltherm:
Just a minor issue: output for repeatedly rejected time steps is now always: "warning: Time step 1 was rejected 1 times and it will be repeated with a reduced step size.", it never increases to "2 times or more".

PR #1868, opened 2017-07-06
 Removed Biot's constant from strain coupling term

wenqing:
As titled.

Note: this does not solve the problem at edge node.

### DIFF
--- a/ProcessLib/HydroMechanics/HydroMechanicsFEM.h
+++ b/ProcessLib/HydroMechanics/HydroMechanicsFEM.h
@@ -298,6 +298,8 @@ public:
         SpatialPosition x_position;
         x_position.setElementID(_element.getID());
 
+        auto const b_alpha = _process_data.biot_coefficient(t, x_position)[0];
+
         unsigned const n_integration_points =
             _integration_method.getNumberOfPoints();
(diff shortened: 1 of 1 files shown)


ProcessLib/HydroMechanics/HydroMechanicsFEM.h
```diff
@@ -298,6 +298,8 @@ class HydroMechanicsLocalAssembler
         SpatialPosition x_position;
         x_position.setElementID(_element.getID());
 
+        auto const b_alpha = _process_data.biot_coefficient(t, x_position)[0];
```
chleh: In principle, `biot_coefficient` could be different for every integration point, which now cannot be modelled by this code anymore.
But probably you had some discussion before, why the old implementation caused some errors.
IMHO there should be a comment here, that `biot_coefficient` cannot be per integration point (Sure, currently we don't have per-integration-point properties anyway, but maybe in the future we might have.). A check somewhere in the code would be even better, but probably cumbersome and badly maintainable.

nagelt: The Biot-coefficient should remain, as any other material parameter, locally assignable ..

PR #1871, opened 2017-07-11
Fixing HM specific error.

endJunction:
Not finished for the review, but the main part is changing the enumeration order of the Quad4 and Quad9.

TODO:

- [x] Update HM and LIE/HM test results
- [ ] Extend HM confined compression benchmark description with spatial comparison of displacement.

It works now :fireworks:


chleh:
:smile: :gun:
TODO: Implement further shape function unit tests.

NumLib/Fem/ShapeFunction/ShapeQuad4-impl.h
```diff
@@ -16,23 +16,24 @@ namespace NumLib
...
 {
-    N[0] = (1.0 + r[0]) * (1.0 + r[1]) / 4;
-    N[1] = (1.0 - r[0]) * (1.0 + r[1]) / 4;
-    N[2] = (1.0 - r[0]) * (1.0 - r[1]) / 4;
-    N[3] = (1.0 + r[0]) * (1.0 - r[1]) / 4;
+    N[0] = (1.0 - r[0]) * (1.0 - r[1]) / 4;
+    N[1] = (1.0 + r[0]) * (1.0 - r[1]) / 4;
+    N[2] = (1.0 + r[0]) * (1.0 + r[1]) / 4;
+    N[3] = (1.0 - r[0]) * (1.0 + r[1]) / 4;
 }
```
wenqing: I think if the ordering orientation of the reference element is the same as the of the real element, this ordering is not the problem because x_i = sum_i^n (N_ x_i)  is always kept when the ordering orientations of reference  and real element are the same. It is the same for the quad9.

wenqing: In other words, it does not matter which vertex of an element is numbered as the first node but the node numbering order  is the matter.

endJunction: It does matter, because now the solution is correct. The ordering of N_u and N_p shape functions must correspond, otherwise products like dN_u / dx * N_p are wrong.

NumLib/Fem/ShapeFunction/ShapeQuad9-impl.h
```diff
@@ -9,44 +9,42 @@
...
 }
 
 template <class T_X, class T_N>
-void ShapeQuad9::computeGradShapeFunction(const T_X &r, T_N &dNdr)
+void ShapeQuad9::computeGradShapeFunction(const T_X& r, T_N& dNdr)
 {
-    dNdr[8] = -2.0 * r[0] * (1.0 - r[1] * r[1]);
-    dNdr[7] = +0.5 * (1.0 - r[1] * r[1]) - 0.5 * dNdr[8];
-    dNdr[6] = -1.0 * r[0] * (1.0 - r[1]) - 0.5 * dNdr[8];
-    dNdr[5] = -dNdr[7];
```
wenqing: Finally, I found that dNdr[5] differs from ogs 5 and it is wrong. It is quit similar to  dNdr[7]
```
 -0.5 * (1.0 - r[1] * r[1]) - 0.5 * dNdr[8];
```
but it is not.

NumLib/Fem/ShapeFunction/ShapeQuad9-impl.h
```diff
@@ -9,44 +9,42 @@
...
+    dNdr[5] = (1 - r[1] * r[1]) * (r[0] + 0.5);
+    dNdr[6] = -r[0] * r[1] * (1 + r[1]);
+    dNdr[7] = (1 - r[1] * r[1]) * (r[0] - 0.5);
+    dNdr[8] = 2 * r[0] * (r[1] * r[1] - 1);
 
-    dNdr[17] = -2.0 * r[1] * (1.0 - r[0] * r[0]);
-    dNdr[16] = -1.0 * r[1] * (1.0 + r[0]) - 0.5 * dNdr[17];
-    dNdr[15] = -0.5 * (1.0 - r[0] * r[0]) - 0.5 * dNdr[17];
-    dNdr[14] = -1.0 * r[1] * (1.0 - r[0]) - 0.5 * dNdr[17];
-    dNdr[13] = -dNdr[15];
```
wenqing: Here the same.

NumLib/Fem/ShapeFunction/ShapeQuad4-impl.h
wenqing: Yes,   N_u and N_p shape functions must correspond. It means that for a geometrical element type, the node ordering rule for all finite element types must be the same.
In summary,  the problem is caused by
1. quad4 and quad8 has  the different node ordering.
2. quad4 and quad9 has the same node order but quad9 got wrong modification in grad N after copying. Got correct results by removing the modification in grad N of quad9 without change the order of N.

I am going to prepare a test to check grad N and whether the same order rule are applied for both linear and high order   elements within a geometrical element type.

NumLib/Fem/ShapeFunction/ShapeQuad9-impl.h
wenqing: The change could due to my comment for simplification in the corresponding PR.

endJunction: The derivative is correct: http://www.wolframalpha.com/input/?i=D%5Br+*+(r+%2B+1)+*+(1+-+s+*+s)+%2F+2,+r%5D

wenqing: I means dNdr[5] = -dNdr[7] is wrong,  `-0.5 * (1.0 - r[1] * r[1]) - 0.5 * dNdr[8]` is correct as that shown in your link. Here is the results after this correction.
quad9

NumLib/Fem/ShapeFunction/ShapeQuad4-impl.h
norihiro-w: Wenqing is right. the node ordering of quad8 is wrong. and my apologies to the mistakes in quad9.

wenqing: @norihiro-w  I gave the comment for the simplification the shape function and I should say apology :smiley: .

endJunction:
Few of the ctest still needs an update:
- [ ] HydroMechanics_HML_square_1e2_unconfined_compression_early
- [ ] HydroMechanics_HML_square_1e2_unconfined_compression_late
- [ ] LIE_HM_single_fracture
- [ ] LIE_HM_TaskB

endJunction:
The LIE/HM single fracture test need a more careful examination, especially a comparison with semianalytical solution, though the solution looks better and in the main nodes it is closely following the previous reference.

norihiro-w:
Current LIE/HM cannot provide the exactly same result as the semi-analytical solution which neglects deformation in x-direction.  The simulation result should be slightly underestimate of the semianalytical. I think, to get exact, we need an anisotropic model in elasticity.

norihiro-w:
@endJunction @wenqing  what is a reason for still keeping the changes in the node ordering in quad4 and quad9? Following the wenqing comments, isn't it sufficient to correct the few quad9 shape functions and the node-ordering of quad8? These shape functions were intended to be kept same as ogs5 so that it can help debugging because one can make exact comparison of local matrices even for each gauss points.

endJunction:
@norihiro-w Quad8 changed now (instead of Q4 and Q9) and for the Q9 the gradient is fixed. Not sure if Quad8 implementation in OGS-5 is the same.

Ready for final review.

norihiro-w:
@endJunction thanks for the update. The node ordering of Quad8 in ogs5 is starting from lower-left corner, which is inconsistent with Quad4 and Quad9 (thanks to this PR, I realized it). This hasn't become a problem because in ogs5 Quad8 is used only for surface integration (i.e. ST) in 3D problems (because of Hex20) and there is no chance to combine it with Quad4. 

quick question: is Quad9 tested somewhere in Tests? Currently 'CreateQuadraticMesh' tools generates only Quad8.

endJunction:
In this PR there is an HM test using quad9 mesh.

norihiro-w:
thanks

norihiro-w:
looks good for me

PR #1874, opened 2017-07-19
Enable vectorial output for secondary variables.

### LINKED ISSUE #1552: Extrapolation of vectorial quantities
I'm currently working on it. Just FYI in order to avoid duplicate work. I won't make a PR immediately, though.

This will also make it unnecessary to save, e.g. the Darcy velocity inside the local assemblers.

chleh:
This PR adds the possibility to extrapolate vectorial integration point quantities to mesh nodes (extrapolation proceeds component-wise).

Furthermore, now more data is passed to the `getIntPt...()` methods:
```cpp
std::vector<double> const& getIntPtDarcyVelocity(
    const double t,
    GlobalVector const& current_solution,
    std::vector<double>& cache) const override
{
...
}
```
Namely time, current solution and the processes DOF table. That allows for on-demand computation of secondary variables. In particulary, e.g., the Darcy velocity does not have to be saved inside the local assemblers anymore, just in order to write it to VTU files in the end, but can be computed from the pressure solution.

Furthermore the function pointers computing integration point values now have become full `std::function`s (see [here](https://github.com/ufz/ogs/compare/master...chleh:vector-extrapolation?expand=1#diff-86bffb51b4ee67d7d0e5a81132e8d0b7R92)), which makes them much more flexible.
This feature is a necessary prerequisite for being able to output secondary (internal) variables, which are defined by individual material models. Being able to output those secondary variables is a long-awaited feature. Soon there will be a follow-up PR implementing this.

Due to this PR the amount of C++ code needed to define a secondary variable will decrease a bit. Additionally, in the prj files less outputs have to be specified.

Limitations
---------------

2D vector fields are written as two-component vectors to the vtu file. Unfortunately paraview does not treat 2D fields as vector fields (AFAIK). Suggestions for improving this are welcome. Though, those suggestions will not be included into this PR.

TODO
------------

Many processes have not been refactored to produce vectorial output. I'll leave that for later. I'd be happy if someone could help with that, e.g., updating test reference results. See also https://github.com/chleh/ogs6/tree/vector-extrapolation-PART2.

 Closes #1552.

This is now a follow-up of #1879.

### DIFF
--- a/NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.cpp
+++ b/NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.cpp
@@ -121,8 +121,9 @@ void LocalLinearLeastSquaresExtrapolator::extrapolateElement(
             _integration_point_values_cache);
 
     auto const& N_0 = extrapolatables.getShapeMatrix(element_index, 0);
-    const unsigned num_nodes = N_0.cols();
-    const unsigned num_values = integration_point_values.size();
+    auto const num_nodes = static_cast<unsigned>(N_0.cols());
+    auto const num_values =
+        static_cast<unsigned>(integration_point_values.size());
 
     if (num_values % num_components != 0)
         OGS_FATAL(
@@ -249,7 +250,7 @@ void LocalLinearLeastSquaresExtrapolator::calculateResidualElement(
         element_index, t, current_solution, dof_table,
         _integration_point_values_cache);
 
-    const unsigned num_values = int_pt_vals.size();
+    auto const num_values = static_cast<unsigned>(int_pt_vals.size());
     if (num_values % num_components != 0)
         OGS_FATAL(
             "The number of computed integration point values is not divisable "
@@ -262,7 +263,7 @@ void LocalLinearLeastSquaresExtrapolator::calculateResidualElement(
 
     const auto& global_indices =
         _dof_table_single_component(element_index, 0).rows;
-    const unsigned num_nodes = global_indices.size();
+    const auto num_nodes = static_cast<unsigned>(global_indices.size());
 
     auto const& interpolation_matrix =
         _qr_decomposition_cache.find({num_nodes, num_int_pts})->second.A;
@@ -285,7 +286,8 @@ void LocalLinearLeastSquaresExtrapolator::calculateResidualElement(
                                  int_pt_vals_mat.row(comp).transpose())
                                     .squaredNorm();
 
-        auto const eidx = num_components * element_index + comp;
+        auto const eidx =
+            static_cast<GlobalIndexType>(num_components * element_index + comp);
         _residuals->set(eidx, std::sqrt(residual / num_int_pts));
     }
 }
--- a/NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.cpp
+++ b/NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.cpp
@@ -288,7 +288,9 @@ void LocalLinearLeastSquaresExtrapolator::calculateResidualElement(
 
         auto const eidx =
             static_cast<GlobalIndexType>(num_components * element_index + comp);
-        _residuals->set(eidx, std::sqrt(residual / num_int_pts));
+        // The residual is set to the root mean square value.
+        auto const root_mean_square = std::sqrt(residual / num_int_pts);
+        _residuals->set(eidx, root_mean_square);
     }
 }
 
--- a/NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.cpp
+++ b/NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.cpp
@@ -60,7 +60,7 @@ void LocalLinearLeastSquaresExtrapolator::extrapolate(
     // compute the average afterwards
     auto counts =
         MathLib::MatrixVectorTraits<GlobalVector>::newInstance(*_nodal_values);
-    counts->setZero();  // TODO BLAS?
+    counts->setZero();
 
     auto const size = extrapolatables.size();
     for (std::size_t i = 0; i < size; ++i)
@@ -154,7 +154,8 @@ void LocalLinearLeastSquaresExtrapolator::extrapolateElement(
         interpolation_matrix.resize(num_int_pts, num_nodes);
 
         interpolation_matrix.row(0) = N_0;
-        for (unsigned int_pt = 1; int_pt < num_int_pts; ++int_pt) {
+        for (unsigned int_pt = 1; int_pt < num_int_pts; ++int_pt)
+        {
             auto const& shp_mat =
                 extrapolatables.getShapeMatrix(element_index, int_pt);
             assert(shp_mat.cols() == num_nodes);
@@ -165,7 +166,8 @@ void LocalLinearLeastSquaresExtrapolator::extrapolateElement(
 
         // JacobiSVD is extremely reliable, but fast only for small matrices.
         // But we usually have small matrices and we don't compute very often.
-        // Cf. http://eigen.tuxfamily.org/dox/group__TopicLinearAlgebraDecompositions.html
+        // Cf.
+        // http://eigen.tuxfamily.org/dox/group__TopicLinearAlgebraDecompositions.html
         //
         // Decomposes interpolation_matrix = U S V^T.
         Eigen::JacobiSVD<Eigen::MatrixXd> svd(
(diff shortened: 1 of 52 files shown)


chleh:
From https://jenkins.opengeosys.org/job/ufz/job/ogs/job/PR-1874/3/console:
```
16:51:18 fatal: no tag exactly matches 'a6e342efedc9938fa2b1a5e135493a3a12ffa919'
```

Does that always happen if a branch has to be merged into master and cannot be fast-forwarded? Is so, it would be very undesirable behaviour IMHO. Cf.
```
16:51:14 Merging remotes/origin/master commit 5d8ca6b6b2ae3189c0ac62e40f10c8947ebfbf91 into PR head commit a6e342efedc9938fa2b1a5e135493a3a12ffa919
```

ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
```diff
@@ -197,34 +149,40 @@ class LocalAssemblerData : public GroundwaterFlowLocalAssemblerInterface
...
+        const double t,
+        GlobalVector const& current_solution,
+        NumLib::LocalToGlobalIndexMap const& dof_table,
+        std::vector<double>& cache) const override
     {
-        assert(!_darcy_velocities.empty());
-        return _darcy_velocities[0];
-    }
+        // auto const num_nodes = ShapeFunction_::NPOINTS;
+        auto const num_intpts = _shape_matrices.size();
```
TomFischer: Better to use common pattern
```c++
unsigned const n_integration_points =
_integration_method.getNumberOfPoints();
```

ProcessLib/HT/HTFEM.h
```diff
@@ -272,15 +258,36 @@ class LocalAssemblerData : public HTLocalAssemblerInterface
...
+        return Eigen::Map<const Eigen::RowVectorXd>(N.data(), N.size());
+    }
+
+    std::vector<double> const& getIntPtDarcyVelocity(
+        const double t,
+        GlobalVector const& current_solution,
+        NumLib::LocalToGlobalIndexMap const& dof_table,
+        std::vector<double>& cache) const override
     {
+        auto const num_intpts = _ip_data.size();
```
TomFischer: Same as in groundwater flow.

ProcessLib/HT/HTFEM.h
```diff
@@ -289,8 +296,7 @@ class LocalAssemblerData : public HTLocalAssemblerInterface
         auto const p_nodal_values = Eigen::Map<const NodalVectorType>(
             &local_x[ShapeFunction::NPOINTS], ShapeFunction::NPOINTS);
 
-        for (unsigned ip = 0; ip < n_integration_points; ++ip)
-        {
+        for (unsigned ip = 0; ip < n_integration_points; ++ip) {
```
TomFischer: Formatting.

ProcessLib/HT/HTFEM.h
```diff
@@ -303,64 +309,26 @@ class LocalAssemblerData : public HTLocalAssemblerInterface
...
             auto const mu = _process_data.fluid_properties->getValue(
                 MaterialLib::Fluid::FluidPropertyType::Viscosity, vars);
             GlobalDimMatrixType const K_over_mu = K / mu;
 
-            GlobalDimVectorType velocity = -K_over_mu * dNdx * p_nodal_values;
-            if (_process_data.has_gravity)
-            {
+            cache_vec.col(ip).noalias() = -K_over_mu * dNdx * p_nodal_values;
+
+            if (_process_data.has_gravity) {
```
TomFischer: Formatting.

NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.cpp
```diff
@@ -146,51 +184,100 @@ void LocalLinearLeastSquaresExtrapolator::extrapolateElement(
...
+            MathLib::toVector(integration_point_values);
+
+        // Apply the pre-computed pseudo-inverse.
+        Eigen::VectorXd const nodal_values =
+            cached_data.A_pinv * integration_point_values_vec;
 
-    // TODO does that give rise to PETSc problems?
-    _nodal_values.add(global_indices, nodal_values);
-    counts.add(global_indices, std::vector<double>(global_indices.size(), 1.0));
+        // TODO does that give rise to PETSc problems?
```
chleh: Fixed in later commit.

NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.cpp
```diff
@@ -137,60 +178,111 @@ void LocalLinearLeastSquaresExtrapolator::extrapolateElement(
...
+        }
+
+        double const residual = (interpolation_matrix * nodal_vals_element -
+                                 int_pt_vals_mat.row(comp).transpose())
+                                    .squaredNorm();
 
-    _residuals.set(element_index, std::sqrt(residual / num_int_pts));
+        auto const eidx =
+            static_cast<GlobalIndexType>(num_components * element_index + comp);
+        _residuals->set(eidx, std::sqrt(residual / num_int_pts));
```
TomFischer: Why not `1/n * sqrt(residual)`?

ProcessLib/HT/HTFEM.h
```diff
@@ -273,30 +273,28 @@ class LocalAssemblerData : public HTLocalAssemblerInterface
         NumLib::LocalToGlobalIndexMap const& dof_table,
         std::vector<double>& cache) const override
     {
-        auto const num_intpts = _ip_data.size();
+        auto const n_integration_points =
```
TomFischer: Same is of course in most other processes....

NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.cpp
```diff
@@ -137,60 +178,111 @@ void LocalLinearLeastSquaresExtrapolator::extrapolateElement(
...
+        // filter nodal values of the current element
+        for (unsigned i = 0; i < num_nodes; ++i) {
+            // TODO PETSc negative indices?
+            auto const idx = num_components * global_indices[i] + comp;
+            nodal_vals_element[i] = (*_nodal_values)[idx];
+        }
+
+        double const residual = (interpolation_matrix * nodal_vals_element -
+                                 int_pt_vals_mat.row(comp).transpose())
+                                    .squaredNorm();
```
TomFischer: This residual computation and the interpolation procedure is documented somewhere, isn't it?

ProcessLib/Process.cpp
```diff
@@ -264,7 +267,7 @@ void Process::preIteration(const unsigned iter, const GlobalVector &x)
 {
     // In every new iteration cached values of secondary variables are expired.
     for (auto& cached_var : _cached_secondary_variables) {
-        cached_var->expire();
+        cached_var->updateCurrentSolution(x, *_local_to_global_index_map);
```
TomFischer: Wouldn't it be sufficient to update it in postTimestep?

ProcessLib/ProcessOutput.cpp
```diff
@@ -119,25 +120,70 @@ void doProcessOutput(std::string const& file_name,
...
+    {
+        switch (type) {
+        case MeshLib::MeshItemType::Cell: return mesh.getNumberOfElements();
+        case MeshLib::MeshItemType::Node: return mesh.getNumberOfNodes();
+        default: break; // avoid compiler warning
+        }
+        return 0;
+    };
+
+    auto get_or_create_mesh_property = [&mesh, &count_mesh_items](
```
TomFischer: ```MeshLib::getOrCreateMeshProperty```

NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.cpp
chleh: There is some short documentation [here](https://github.com/ufz/ogs/blob/master/NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.h#L20). Please check if it's sufficient. IMHO everyone interested in the details should read the source code.

ProcessLib/TES/TESLocalAssembler-impl.h
```diff
@@ -227,40 +238,55 @@ TESLocalAssembler<ShapeFunction_, IntegrationMethod_, GlobalDim>::
...
-template <typename ShapeFunction_, typename IntegrationMethod_,
-          unsigned GlobalDim>
-std::vector<double> const& TESLocalAssembler<
-    ShapeFunction_, IntegrationMethod_,
-    GlobalDim>::getIntPtDarcyVelocityY(std::vector<double>& /*cache*/) const
-{
-    assert(_d.getData().velocity.size() > 1);
-    return _d.getData().velocity[1];
-}
+    auto const indices = NumLib::getIndices(_element.getID(), dof_table);
```
TomFischer: You only need the id of the element, isn't it?

NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.cpp
chleh: From the following line of thought: `residual` is made by summing squared errors. If now e.g. summing two squared errors `e*e` the following picture can be drawn:
```
res = std::sqrt((e*e + e*e) / 2);
_residuals->set(eidx, res);
```.
So `res = sqrt((e*e + e*e) / 2) = sqrt(e*e) = |e|`, which is always the same, no matter how many of the `e*e` terms are added,
in contrast when pulling the `1/n` out of the sqrt, one gets `res = 1/n * sqrt(n * e*e) = |e| / sqrt(n)`, which decreases for increasing `n`.
So I found it more natural to take the choice as in the current implementation.
Note: This PR does not change this formula.

ProcessLib/Process.cpp
chleh: `CachedSecondaryVariable`s are used to produce output. Since output can be produced after every nonlinear iteration (e.g., for debugging), the update has to be done in every iteration.
I chose to do the update in the `preIteration()` method.
But probably I could remove the update from the timestep method. TODO.

ProcessLib/TES/TESLocalAssembler-impl.h
chleh: Yes, probably. But in other processes the pattern is the same, cf. HT. As far as I can see.

ProcessLib/ProcessOutput.cpp
chleh: Thanks for the hint.

NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.cpp
chleh: I now added the term "root mean square" explicitly.

PR #1875, opened 2017-07-19
Internal variables output

chleh:
... for SmallDeformation process with Ehlers/Damage model.

follow-up of #1874.

This PR implements nodal output of material-model-defined internal variables at the example of the SD process.
I do not plan to implement more such output functionality now. Instead this PR can serve as a starting point for others.

TODO:
* [ ] remove component-wise output of stress and strain from small deformation process
* [ ] update SmallDeformation ctests to use vectorial stress/strain and to test that damage and plastic strain are computed correctly.

It would be nice if somebody could help me with the TODOs above, since SmallDeformation is not my main business.

### DIFF
--- a/MaterialLib/SolidModels/Ehlers-impl.h
+++ b/MaterialLib/SolidModels/Ehlers-impl.h
@@ -749,21 +749,84 @@ SolidEhlers<DisplacementDim>::getInternalVariables() const
     return {
         {"damage.kappa_d", 1,
          [](typename MechanicsBase<
-             DisplacementDim>::MaterialStateVariables const& state) -> double {
+                DisplacementDim>::MaterialStateVariables const& state,
+            std::vector<double>& cache) -> std::vector<double> const& {
              assert(dynamic_cast<StateVariables<DisplacementDim> const*>(
                         &state) != nullptr);
              auto const& ehlers_state =
                  static_cast<StateVariables<DisplacementDim> const&>(state);
-             return ehlers_state.damage.kappa_d();
+
+             cache.resize(1);
+             cache.front() = ehlers_state.damage.kappa_d();
+             return cache;
          }},
         {"damage.value", 1,
          [](typename MechanicsBase<
-             DisplacementDim>::MaterialStateVariables const& state) -> double {
+                DisplacementDim>::MaterialStateVariables const& state,
+            std::vector<double>& cache) -> std::vector<double> const& {
              assert(dynamic_cast<StateVariables<DisplacementDim> const*>(
                         &state) != nullptr);
              auto const& ehlers_state =
                  static_cast<StateVariables<DisplacementDim> const&>(state);
-             return ehlers_state.damage.value();
+
+             cache.resize(1);
+             cache.front() = ehlers_state.damage.value();
+             return cache;
+         }},
+        {"eps_p.D", KelvinVector::RowsAtCompileTime,
+         [](typename MechanicsBase<
+                DisplacementDim>::MaterialStateVariables const& state,
+            std::vector<double>& cache) -> std::vector<double> const& {
+             assert(dynamic_cast<StateVariables<DisplacementDim> const*>(
+                        &state) != nullptr);
+             auto const& ehlers_state =
+                 static_cast<StateVariables<DisplacementDim> const&>(state);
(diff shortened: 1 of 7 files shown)


endJunction:
Maybe we could have a discussion where you would explain main features to me and Tom tomorrow?

MaterialLib/SolidModels/MechanicsBase.h
```diff
@@ -104,6 +104,30 @@ struct MechanicsBase
                     KelvinVector const& sigma_prev,
                     MaterialStateVariables const& material_state_variables) = 0;
 
+    /// Helper type for providing access to internal variables.
+    struct InternalVariable
+    {
+        using Getter = std::function<std::vector<double> const&(
```
endJunction: vector and functional includes missing....

chleh: added. thanks.

MaterialLib/SolidModels/Ehlers-impl.h
```diff
@@ -740,6 +740,94 @@ SolidEhlers<DisplacementDim>::integrateStress(
...
+             assert(dynamic_cast<StateVariables<DisplacementDim> const*>(
+                        &state) != nullptr);
+             auto const& ehlers_state =
+                 static_cast<StateVariables<DisplacementDim> const&>(state);
+             auto const& D = ehlers_state.eps_p.D;
+
+             cache.resize(KelvinVector::RowsAtCompileTime);
+
+             // TODO make a general implementation for converting KelvinVectors
+             // back to symmetric rank-2 tensors.
```
endJunction: Have a look into the KelvinVector [tests](https://github.com/ufz/ogs/blob/master/Tests/MaterialLib/KelvinVector.cpp#L21-L71)

chleh: The difference is that here I am converting, e.g. a 6-component Kelvin vector to a 6-component vector interpreted a symm. tensor by Paraview, and not converting to a 3x3 matrix.

endJunction: Oh, I see. Nevermind.

MaterialLib/SolidModels/Ehlers-impl.h
```diff
@@ -740,6 +740,94 @@ SolidEhlers<DisplacementDim>::integrateStress(
...
+         [](typename MechanicsBase<
+                DisplacementDim>::MaterialStateVariables const& state,
+            std::vector<double>& cache) -> std::vector<double> const& {
+             assert(dynamic_cast<StateVariables<DisplacementDim> const*>(
+                        &state) != nullptr);
+             auto const& ehlers_state =
+                 static_cast<StateVariables<DisplacementDim> const&>(state);
+             auto const& D = ehlers_state.eps_p.D;
+
+             cache.resize(KelvinVector::RowsAtCompileTime);
```
wenqing: Just curious, is it efficient to call vector::resize() at integration point level?

chleh: AFAIK resize will allocate new memory only if the new size exceeds the old vector capacity. So there won't be a new allocation for every integration point.
But for every element there will be a new `cache` instance (if I remember correctly), and subsequently there will be (at least) one heap allocation per element.
So this approach is not very efficient, but the problem does not in this line, rather it lies in the fact that there is a new cache instance for each element (the advantage of this approach is that there is no "global" cache, which makes the code a bit simpler).
Additionally this is used for output only, so it should not be too expensive.

wenqing: I see. Thanks for explanation.

ProcessLib/SmallDeformation/SmallDeformationFEM.h
```diff
@@ -352,6 +432,18 @@ class SmallDeformationLocalAssembler
         return getIntPtEpsilon(cache, 5);
     }
 
+    unsigned getNumberOfIntegrationPoints() const override
+    {
+        return static_cast<unsigned>(_ip_data.size());
```
endJunction: Could be more direct information from the _integration_method.

chleh: Changed.

PR #1877, opened 2017-07-20
Restructure HydroMechanics

endJunction:
Continuation of #1855 for (re)compilation speed improvements.


endJunction:
Only unrelated doxygen warnings, otherwise all green.

ProcessLib/HydroMechanics/LocalDataInitializer.h
```diff
@@ -115,11 +115,11 @@ namespace HydroMechanics
 ///
 /// \attention This is modified version of the ProcessLib::LocalDataInitializer
 /// class which does not include line elements, allows only shapefunction of
-/// order 2, and provides additional template argument DisplacementDim.
+/// order 2.
```
chleh: If I remember correctly, the additional template argument was one reason why you can't use the general LocalDataInitializer class. Another reason was, that HM cannot be instantiated for 1D and 0D elements.
Maybe now it's easier to make HM use the general LocalDataInitializer.
I suggest that we put that with a rather low priority somewhere on some agenda.

endJunction: Yes, now it is easier to create a unified LocalDataInitializer.

ProcessLib/HydroMechanics/HydroMechanicsProcess-impl.h
```diff
@@ -0,0 +1,240 @@
...
+    std::vector<MeshLib::MeshSubsets> all_mesh_subsets_single_component;
+    all_mesh_subsets_single_component.emplace_back(
+        _mesh_subset_all_nodes.get());
+    _local_to_global_index_map_single_component =
+        std::make_unique<NumLib::LocalToGlobalIndexMap>(
+            std::move(all_mesh_subsets_single_component),
+            // by location order is needed for output
+            NumLib::ComponentOrder::BY_LOCATION);
+
+    Base::_secondary_variables.addSecondaryVariable(
```
wenqing: A comment not on this PR: It could be good that all mechanical related processes share this piece of code.

endJunction: Would be interesting to try out polymorphism for this:
```
MechanicsExtrapolatableProcess <- HM_Process
                             | <- SD_Process
```

And then the same for the local assemblers...

PR #1847, opened 2017-06-20
Small fixes

endJunction:
Several bug fixes and small improvements. See each commit message.

PR #1887, opened 2017-07-27
Floating point exceptions for debugging.

endJunction:
NaN were introduced https://github.com/ufz/ogs/pull/1847/commits/0498c773278ec81c49b02261f8bb6f00a90aa73b here and are used by default in debug compilations.

### DIFF
--- a/Applications/CLI/ogs.cpp
+++ b/Applications/CLI/ogs.cpp
@@ -13,9 +13,9 @@
 #include <chrono>
 #include <tclap/CmdLine.h>
 
-#ifndef NDEBUG
+#ifndef _WIN32
 #include <cfenv>
-#endif  // NDEBUG
+#endif  // _WIN32
 
 #ifdef USE_PETSC
 #include <vtkMPIController.h>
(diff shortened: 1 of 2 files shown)


Applications/CLI/ogs.cpp
```diff
@@ -97,6 +101,11 @@ int main(int argc, char *argv[])
     INFO("This is OpenGeoSys-6 version %s.",
          BaseLib::BuildInfo::git_describe.c_str());
 
+// Enable floating point exceptions for debugging.
+#ifndef NDEBUG
+    feenableexcept(FE_DIVBYZERO | FE_INVALID | FE_OVERFLOW);
```
chleh: I'd add a command line switch that enables/disables this feature, because even when debugging one might not always want the progrm to crash, e.g. when dividing by zero.

chleh:
Nice feature, could be very helpful in finding some errors, especially if data is initialized with NaN by default.

chleh:
Maybe add some TODO, that windows should be supported in the future, too.

Applications/CLI/ogs.cpp
```diff
@@ -85,6 +89,11 @@ int main(int argc, char *argv[])
         "use unbuffered standard output");
     cmd.add(unbuffered_cout_arg);
 
+    TCLAP::SwitchArg enable_fpe_arg("",
```
endJunction: It's documented no-op.

chleh: Thanks.

PR #1888, opened 2017-07-27
Writing SWMM data categories

rinkk:
This makes the SWMM converter slightly more usable for non-programmers. By default *all* node and link data is written into vtu-files (one for every timestep) and *all* subcatchment and system data is written into csv-files (one for every subcatchment/system)

For selective conversion of the data (only certain time steps, specific nodes/links/etc., assessment of pollutants, ...) it'd still be necessary to either adjust the source code or write a comprehensive user interface that accounts for such things.

### DIFF
--- a/Applications/FileIO/SWMM/SWMMInterface.cpp
+++ b/Applications/FileIO/SWMM/SWMMInterface.cpp
@@ -917,11 +917,11 @@ std::size_t SwmmInterface::getNumberOfParameters(SwmmObject obj_type) const
     switch (obj_type)
     {
         case SwmmObject::SUBCATCHMENT:
-            return (n_obj_params[0] - 1 + SWMM_Npolluts);
+            return (n_obj_params[0] + SWMM_Npolluts);
         case SwmmObject::NODE:
-            return (n_obj_params[1] - 1 + SWMM_Npolluts);
+            return (n_obj_params[1] + SWMM_Npolluts);
         case SwmmObject::LINK:
-            return (n_obj_params[2] - 1 + SWMM_Npolluts);
+            return (n_obj_params[2] + SWMM_Npolluts);
         case SwmmObject::SYSTEM:
             return n_obj_params[3];
         default:
@@ -944,7 +944,7 @@ std::size_t SwmmInterface::getNumberOfTimeSteps() const
 bool SwmmInterface::addResultsToMesh(MeshLib::Mesh &mesh, SwmmObject const swmm_type,
     std::string const& vec_name, std::vector<double> const& data)
 {
-    if (!(swmm_type == SwmmObject::NODE) || (swmm_type == SwmmObject::LINK))
+    if (!(swmm_type == SwmmObject::NODE || swmm_type == SwmmObject::LINK))
     {
         ERR ("Information of this object type cannot be added to mesh.");
         return false;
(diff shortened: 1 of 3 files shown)


Applications/FileIO/SWMM/SWMMInterface.cpp
```diff
@@ -1151,14 +1148,14 @@ std::string SwmmInterface::getArrayName(SwmmObject obj_type, std::size_t var_idx
         if (var_idx < n_obj_params[1])
             return node_vars[var_idx];
         if (var_idx < n_obj_params[1]+n_pollutants)
-            return _pollutant_names[var_idx-n_obj_params[1]];
+            return std::string("Node_" + _pollutant_names[var_idx-n_obj_params[1]]);
```
endJunction:
```c++
return "Node_" + _pullutant_names;
```
Should do the same, if I'm not mistaken...

Applications/Utils/SWMMConverter/SWMMConverter.cpp
```diff
@@ -23,109 +24,180 @@
...
+        if (!swmm.addResultsToMesh(mesh, type, vec_name, data_vec))
+            return -3;
+    }
+    return 0;
+}
+
+int writeMeshOutput(std::string const& input_file, std::string const& output_file,
+    bool const node_args, bool const link_args)
+{
+    std::unique_ptr<FileIO::SwmmInterface> swmm = nullptr;
```
endJunction: `swmm` should be initialized:
```c++
auto swmm = std::make_unique(FileIO::SwmmInterface::create(input_file));
```

Applications/Utils/SWMMConverter/SWMMConverter.cpp
```diff
@@ -23,109 +24,180 @@
...
+        return -1;
+
+    GeoLib::IO::BoostXmlGmlInterface xml(geo_objects);
+    xml.setNameForExport(BaseLib::extractBaseNameWithoutExtension(input_file));
+    xml.writeToFile(output_file);
+    return 0;
+}
+
+int addObjectsToMesh(FileIO::SwmmInterface &swmm, MeshLib::Mesh &mesh,
+    FileIO::SwmmObject const type, std::size_t const timestep)
```
rinkk: It's an enum :)

Applications/Utils/SWMMConverter/SWMMConverter.cpp
```diff
@@ -23,109 +24,180 @@
...
+
+int writeMeshOutput(std::string const& input_file, std::string const& output_file,
+    bool const node_args, bool const link_args)
+{
+    std::unique_ptr<FileIO::SwmmInterface> swmm = nullptr;
+    swmm = FileIO::SwmmInterface::create(input_file);
+    if (swmm == nullptr)
+    return -1;
+
+    MeshLib::Mesh mesh = swmm->getMesh();
```
endJunction: Is the mesh copy intentional?

endJunction:
- Formatting could be improved. (hint: [clang-format](https://llvm.org/builds/) ^^)
 - It would be nice if you would get rid of the merge commits by rebasing on the current master.

rinkk:
How do I see the problems with the Jenkins build?

Applications/FileIO/SWMM/SWMMInterface.cpp
endJunction: ... and could save one copy or move ctor call.

endJunction:
The problems with Jenkins are Jenkins' problems, don't worry for now. The code compiles just fine. There are no tests for the tool, isn't it?

Formatting still needs improvement.

rinkk:
You could simply let me know what exactly needs improvement in your opinion so I don't have to keep guessing what you might be referring to.

Concerning tests: The smallest SWMM example I have is

rinkk:
You could simply let me know what exactly needs improvement in your opinion so I don't have to keep guessing what you might be referring to.

Concerning tests: The smallest SWMM example I have is a ~275 mb file. When creating the interface I was told this was to large for a testfile.

Applications/FileIO/SWMM/SWMMInterface.cpp
rinkk: Does this *really* make a difference? I would assume that a string is constructed implicitely either way because the method cannot return multiple string fragments. If this can be avoided somehow I'm sure the compiler will do so in the optimization process.

And even if none of this happens it'll simply make no noticable difference in runtime. This is about returning 10-20 names in a process that reads/writes hundreds of megabytes.

It's just that the explicit statement of a string construction makes this more readable in my opinion.

PR #1893, opened 2017-08-02
Fixed a bug in time stepper when non-linear solver diverges

wenqing:
The  step size estimation after a diverged non-linear solver is now integrated into the time stepping function. This resolves 
1) the problem that time stepping goes endless rejected steps after non-linear solver fails,
2) the rejected time step after a diverged non-linear solver restarts with the solution of the previous non-linear step but not with the solution of the previous time step.

With the changes, the time step size after a diverged nonlinear solver is computed in time stepper with the in-converged solution of the non-linear solver.

### DIFF
--- a/ProcessLib/UncoupledProcessesTimeLoop.cpp
+++ b/ProcessLib/UncoupledProcessesTimeLoop.cpp
@@ -126,6 +126,7 @@ struct SingleProcessData
     //! other members of this struct to their concrety types.
     NumLib::NonlinearSolverTag const nonlinear_solver_tag;
     NumLib::NonlinearSolverBase& nonlinear_solver;
+    bool nonlinear_solver_converged;
     std::unique_ptr<NumLib::ConvergenceCriterion> conv_crit;
 
     std::unique_ptr<NumLib::TimeDiscretization> time_disc;
@@ -152,6 +153,7 @@ SingleProcessData::SingleProcessData(
     : timestepper(std::move(timestepper_)),
       nonlinear_solver_tag(NLTag),
       nonlinear_solver(nonlinear_solver),
+      nonlinear_solver_converged(true),
       conv_crit(std::move(conv_crit_)),
       time_disc(std::move(time_disc_)),
       process(process_),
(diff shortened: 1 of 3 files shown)


ProcessLib/UncoupledProcessesTimeLoop.cpp
```diff
@@ -126,6 +126,7 @@ struct SingleProcessData
     //! other members of this struct to their concrety types.
     NumLib::NonlinearSolverTag const nonlinear_solver_tag;
     NumLib::NonlinearSolverBase& nonlinear_solver;
+    bool is_nonlinear_solver_converged;
```
endJunction: tiny: `nonlinear_solver_converged` would also be a good name w/o "time" indication.

wenqing: Changed to ``nonlinear_solver_converged`` by amending the first commit.

ProcessLib/UncoupledProcessesTimeLoop.cpp
```diff
@@ -658,15 +668,16 @@ double UncoupledProcessesTimeLoop::computeTimeStepping(
         }
         else
         {
-            if (std::abs(dt -prev_dt) < std::numeric_limits<double>::min()
-                && _last_step_rejected)
+            if (std::abs(dt - prev_dt) < std::numeric_limits<double>::min() &&
```
chleh: I think this will only be true if `dt == prev_dt`, because AFAIK std::numeric_limits<double>::min() is 1e-300 or so.

ProcessLib/UncoupledProcessesTimeLoop.cpp
```diff
@@ -658,15 +668,16 @@ double UncoupledProcessesTimeLoop::computeTimeStepping(
...
-                          "\tSuggest to adjust the parameters of the time"
-                          " stepper or try other time stepper.\n"
-                          "\tThe program stops");
+                OGS_FATAL(
+                    "\tThis time step is rejected and the new computed"
+                    " step size is the same as\n"
+                    "\tthat was just used.\n"
+                    "\tSuggest to adjust the parameters of the time"
+                    " stepper or try other time stepper.\n"
+                    "\tThe program stops");
```
chleh: I know it's not introduced in this PR, but OGS_FATAL here might prevent final cleanup, such as writing the last (failed) timestep result.

ProcessLib/UncoupledProcessesTimeLoop.cpp
wenqing: Yes, the condition is true only if ``dt == prev_dt`` and it is for the case that the time step was rejected with the same time step size.

ProcessLib/UncoupledProcessesTimeLoop.cpp
wenqing: Good point.

This piece of source code have moved from this function to loop(...), and it is documented now (see  commit  2191287).

wenqing: Also, a final clean up is enabled if this condition is satisfied.

chleh:
Thanks for adding the additional output.

PR #1896, opened 2017-08-03
Pressure boundary conditions

fparisio:
Normal mechanical stress at external boundaries. Tests include thick walled cylinder and sphere in bi-dimensional, tri-dimensional conditions and with elastic and plastic material models.


fparisio:
@endJunction please add prj docu

ProcessLib/BoundaryCondition/BoundaryCondition.cpp
```diff
@@ -49,6 +51,15 @@ BoundaryConditionBuilder::createBoundaryCondition(
                     integration_order, shapefunction_order, parameters);
     }
 
+    //
+    // Special boundary conditions
+    //
+    if (type == "Pressure")
```
wenqing: ``Pressure`` is a general name, and this sort of BC is only for normal stress or normal traction BC. Therefore, I think it would be good to use ``NormalTraction`` or ``NormalStress`` for this specific BC type.

ProcessLib/BoundaryCondition/PressureBoundaryConditionLocalAssembler.h
```diff
@@ -0,0 +1,174 @@
...
+
+        auto const shape_matrices_u =
+            initShapeMatrices<ShapeFunctionDisplacement,
+                              ShapeMatricesTypeDisplacement, IntegrationMethod,
+                              GlobalDim>(e, is_axially_symmetric,
+                                         _integration_method);
+
+        GlobalDimVectorType element_normal(GlobalDim);
+
+        // TODO Extend to rotated 2d meshes and line elements.
```
wenqing: Extend to axisymmetry? That case could be included in  ``shape_matrices_u[ip].integralMeasure``.

ProcessLib/BoundaryCondition/PressureBoundaryConditionLocalAssembler.h
```diff
@@ -0,0 +1,174 @@
...
+                _integration_method.getWeightedPoint(ip).getWeight() *
+                shape_matrices_u[ip].integralMeasure *
+                shape_matrices_u[ip].detJ;
+
+            _ip_data.emplace_back(N_u.transpose() * element_normal,
+                                  integration_weight);
+        }
+    }
+
+    void assemble(std::size_t const id,
```
wenqing: What is this member for? I can only see the calculation of \sum {w * Nu * p} without detJ.

endJunction: Not sure about the member question. The detJ is included in the w factor, see https://github.com/fparisio/ogs/blob/pressure_bc/ProcessLib/BoundaryCondition/PressureBoundaryConditionLocalAssembler.h#L114-L117

wenqing: I see that PressureBoundaryConditionLocalAssembler calculates Nu*n and w, and assemble() computes the traction BC along with time.

xingyuanmiao:
@endJunction @fparisio Deviation between the numerical and analytical results is found in plane strain elastic cylinder, axisymmetric elastic sphere (where notable difference is found in comparison of displacement u_r), 3d elastic sphere, ..., any idea about the difference?

Also, typos are found in document.

fparisio:
@xingyuanmiao We discussed today with @chleh, @nagelt and @endJunction and probably the cause is related to the fact that pressure is normal to the linearized boundary, not to the "real" curved one. Partly the difference is due to interpolation (a small deviation, as confirmed by axisymmetric pipe test with a linear boundary) and partly to the non-linear geometry. Next fix of pressure_bc with quadratic elements should fix the issue.

I will check for typos.

ProcessLib/BoundaryCondition/PressureBoundaryConditionLocalAssembler.h
endJunction: The rotation does not refer to the axisymmetry but to 2D-meshes in non-xy-plane.

xingyuanmiao:
Thank you! Waiting for the update. :)

renchao-lu:
@fparisio I have read your documentation and have had the following major comments so far
1) There shows great discrepancy at r=1.0 mm in the elastic cylinder case under plane strain conditions (see Figure 1) while under axisymmetric conditions the numerical simulation reproduces the analytical solution (see Figure 2). If the discrepancy arises from the issue of nodal interpolation, why isn't the discrepancy shown in the latter?
2) Could you please use another legend in Figure 6? Using a solid line with empty circles as a legend may make readers confused that the discrepancy shown in the elastic and plastic cases has been resolved, as the solid line and the empty circle, representing the numerical and analytical results respectively, are used in the preceding result comparison.

endJunction:
Set this to WIP because there is an update needed after merging of the non-uniform BCs.

fparisio:
@renchao-lu Thank you for the comments. Question 1 was already answered and is related to curved boundary. In the documentation, a final note was also added. I understand question 2 but I do not agree, the legend is clear and symbols are just helpful to identify at which nodes quantities are computed. Also, the caption and the reference text clearly explain the figure, so that the attentive reader could not misunderstand it.

ProcessLib/BoundaryCondition/BoundaryCondition.cpp
wenqing: Still, I think Pressure may not be a good BC type name. There might be fluid pressure on boundaries in  coupled THM processes, which is Dirichlet type. It could be MechanicalPressure.

nagelt:
coarse_mesh
fine_mesh

As the issue was raised repeatedly please find attached some results from another FE simulator which reproduces the behaviour Francesco observed (sigma_zz increasing on the inside despite the constant analytical solution for a plane-strain pipe). I used other material parameters, so don't pay attention to the absolute numbers. As correctly stated in the above discussion, this is due to the linear approximation of the curved boundary. Also note that these are linear elements and stresses are proportional to the gradient of the primary variable. If we refine the mesh (second image) the effect decreases as we approximate gradients better as well as the curved boundary.

renchao-lu:
@fparisio @xingyuanmiao @nagelt @endJunction @chleh @wenqing It can be seen from the tests with both coarse and fine mesh grids that a limited error is distributed along the radial direction. The relative error is less than 1%. Using fine mesh grids for sure improves the simulation result. However, in the present plane strain case (implemented by ogs6), there shows a relative error which is clearly visible to the naked eye (grossly estimated greater than 10%) and merely appears at the inner boundary node rather than at all the nodes along the radial direction.  Actually this is the point which I cannot understand. I have doubts if such a behaviour can be also attributed to the linear approximation of the curved boundary? Maybe I am too stupid.

nagelt:
@renchao-lu nobody here is stupid. It's a valid point. We have another test coming up! @chleh I have another question regarding the extrapolation residual. Let's discuss once you're here.

PR #1897, opened 2017-08-03
Fix eigen align

fparisio:


chleh:
I assume that all types derived from `MechanicsBase<DisplacementDim>::MaterialStateVariables` and containing some Eigen vector/matrix member should have an `EIGEN_MAKE_ALIGNED_OPERATOR_NEW;`

Issue #1881, opened 2017-07-24
Ehlers Damage test failing sporadically


chleh:
Symptom: `ogs-Mechanics_PlasticModel_SDED_Ehlers_Damage_UniaxialTension` sometimes falies (roughly 50% of the time)
```
[8] lehmannc@envinf141l:build-release-eigenlis$ OMP_NUM_THREADS=1 ctest -R ogs-Mechanics_PlasticModel_SDED_Ehlers_Damage_UniaxialTension
Test project /home/lehmannc/prog/ogs6/github-chleh-PRs/build-release-eigenlis
    Start 210: ogs-Mechanics_PlasticModel_SDED_Ehlers_Damage_UniaxialTension
1/2 Test #210: ogs-Mechanics_PlasticModel_SDED_Ehlers_Damage_UniaxialTension ...........***Failed    9.26 sec
    Start 211: ogs-Mechanics_PlasticModel_SDED_Ehlers_Damage_UniaxialTension-vtkdiff
2/2 Test #211: ogs-Mechanics_PlasticModel_SDED_Ehlers_Damage_UniaxialTension-vtkdiff ...***Failed    0.05 sec

0% tests passed, 2 tests failed out of 2

Total Test time (real) =   9.33 sec

The following tests FAILED:
	210 - ogs-Mechanics_PlasticModel_SDED_Ehlers_Damage_UniaxialTension (Failed)
	211 - ogs-Mechanics_PlasticModel_SDED_Ehlers_Damage_UniaxialTension-vtkdiff (Failed)
Errors while running CTest
[8] lehmannc@envinf141l:build-release-eigenlis$ OMP_NUM_THREADS=1 ctest -R ogs-Mechanics_PlasticModel_SDED_Ehlers_Damage_UniaxialTension
Test project /home/lehmannc/prog/ogs6/github-chleh-PRs/build-release-eigenlis
    Start 210: ogs-Mechanics_PlasticModel_SDED_Ehlers_Damage_UniaxialTension
1/2 Test #210: ogs-Mechanics_PlasticModel_SDED_Ehlers_Damage_UniaxialTension ...........***Failed    9.42 sec
    Start 211: ogs-Mechanics_PlasticModel_SDED_Ehlers_Damage_UniaxialTension-vtkdiff
2/2 Test #211: ogs-Mechanics_PlasticModel_SDED_Ehlers_Damage_UniaxialTension-vtkdiff ...***Failed    0.05 sec

[...]

Issue #1323, opened 2016-07-20
Eigen alignment again


chleh:
Following error message on an Ubuntu 16.04 installation:

```
[----------] 3 tests from NumLibFemIsoTest/9, where TypeParam = (anonymous namespace)::TestCase<FeTestData::TestFeHEX8, EigenFixedShapeMatrixPolicy>
[ RUN      ] NumLibFemIsoTest/9.CheckMassMatrix
testrunner: /usr/include/eigen3/Eigen/src/Core/DenseStorage.h:128: Eigen::internal::plain_array<T, Size, MatrixOrArrayOptions, 32>::plain_array() [with T = double; int Size = 64; int MatrixOrArrayOptions = 1]: Assertion `(reinterpret_cast<size_t>(eigen_unaligned_array_assert_workaround_gcc47(array)) & (31)) == 0 && "this assertion is explained here: " "http://eigen.tuxfamily.org/dox-devel/group__TopicUnalignedArrayAssert.html" " **** READ THIS WEB PAGE !!! ****"' failed.
```

```
$ g++ --version
g++ (Ubuntu 5.4.0-6ubuntu1~16.04.1) 5.4.0 20160609

$ uname -a
Linux envinf141l 4.4.0-31-generic #50-Ubuntu SMP Wed Jul 13 00:07:12 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux

$ file bin/testrunner 
bin/testrunner: ELF 64-bit LSB executable, x86-64, version 1 (GNU/Linux), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.32, BuildID[sha1]=ef502296a868b9a41f09493c45b4383e73de2626, not stripped
```

Cmake build type: Debug.

For me it looks like the alignment issue has returned.

chleh:
Cf. #1221.

chleh:
Stale.

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/1323).

PR #1898, opened 2017-08-06
Eigen alignment

endJunction:
Adding Eigen's operator new to classes using any (fixed size) Eigen::Matrix as members.

Continuation of #1897.
Relates to  #1881 #1323
clang's matcher:
```
recordDecl(
    hasDescendant(fieldDecl(hasType(cxxRecordDecl(hasName("Eigen::Matrix"))))),
    unless(hasDescendant(typedefDecl(hasName("eigen_aligned_operator_new_marker_type")))),
    unless(isExpansionInSystemHeader()))
```
This can be refined not to match dynamic matrices...

### DIFF
--- a/MaterialLib/SolidModels/Ehlers-impl.h
+++ b/MaterialLib/SolidModels/Ehlers-impl.h
@@ -135,6 +135,8 @@ struct PhysicalStressWithInvariants final
     double I_1;
     double J_2;
     double J_3;
+
+    EIGEN_MAKE_ALIGNED_OPERATOR_NEW;
 };
 
 /// Holds powers of 1 + gamma_p*theta to base 0, m_p, and m_p-1.
--- a/MaterialLib/SolidModels/Lubby2.h
+++ b/MaterialLib/SolidModels/Lubby2.h
@@ -149,6 +149,8 @@ public:
         /// iteration
         KelvinVector eps_M_t;
         KelvinVector eps_M_j;
+
+        EIGEN_MAKE_ALIGNED_OPERATOR_NEW;
     };
 
     std::unique_ptr<
--- a/NumLib/Fem/CoordinatesMapping/ShapeMatrices.h
+++ b/NumLib/Fem/CoordinatesMapping/ShapeMatrices.h
@@ -14,6 +14,8 @@
 
 #include <ostream>
 
+#include <Eigen/Core>
+
 namespace NumLib
 {
 
@@ -103,6 +105,7 @@ struct ShapeMatrices
      */
     void write (std::ostream& out) const;
 
+    EIGEN_MAKE_ALIGNED_OPERATOR_NEW;
 }; // ShapeMatrices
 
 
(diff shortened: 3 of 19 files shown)


endJunction:
Just for future reference an email by Christoph pointing to the relevant documentation parts:
>Noch ein paar Infos aus der Eigen Doku:

== 1. ==

http://eigen.tuxfamily.org/dox-devel/group__TopicStructHavingEigenMembers.html

> What about dynamic-size matrices and vectors?

> Dynamic-size matrices and vectors, such as Eigen::VectorXd, allocate
> dynamically their own array of coefficients, so they take care of
> requiring absolute alignment automatically. So they don't cause this
> issue. The issue discussed here is only with fixed-size vectorizable
> matrices and vectors.

Also sollte falsches Alignment für dynamische Matrizen nicht unser
Fehler sein. Die Geschichten mit den STL Containern gelten anscheinend
auch nur für "fixed-size vectorizable
types" (http://eigen.tuxfamily.org/dox-devel/group__TopicStlContainers.html).


== 2. ==

Tun wir irgendwo das:
http://eigen.tuxfamily.org/dox-devel/group__TopicPassingByValue.html

Das könnte evtl. auch Fehler hervorrufen.


== 3. ==

http://eigen.tuxfamily.org/dox-devel/group__TopicUnalignedArrayAssert.html#title6

> How can I check my code is safe regarding alignment issues?
> [...]
> For instance, on most 64 bits systems buffer are aligned on 16 bytes
> boundary and so, if you do not enable AVX instruction set, then your
> code will run fine. On the other hand, the same code may assert if
> moving to a more exotic platform, or enabling AVX instructions that
> required 32 bytes alignment by default.


== 4. ==

Einige Präprozessormakros:
https://eigen.tuxfamily.org/dox-devel/TopicPreprocessorDirectives.html#title3

EIGEN_MAX_ALIGN_BYTES
EIGEN_MAX_STATIC_ALIGN_BYTES
EIGEN_DONT_VECTORIZE
EIGEN_UNALIGNED_VECTORIZE


== 5. == 

Eigen 3.3 Changelog
http://eigen.tuxfamily.org/index.php?title=3.3#Vectorization

> Eigen 3.3 adds support for AVX (x86_64), FMA (x86_64)   [...]   To
> enable AVX or FMA, you need to compile your code with these
> instruction sets enabled on the compiler side, for instance using the
> -mavx and -mfma options with gcc, clang or icc   [...]   By default,
> Eigen's dense objects are thus automatically aligned on 32 bytes when
> AVX is enabled.   [...]   When AVX is enabled, Eigen automatically
> falls back to half register sizes for fixed-size types which are not
> a multiple of the full AVX register size (256bits)

Wie du schon sagtest verbieten wir standardmäßig nur AVX2.

chleh:
@endJunction derived the following from the email cited above:
An HM-test with linear elasticity failed for dynamic Eigen matrices. Anyway, the B-matrices are fixed-size always. Therefore they might have caused the observed problems (i.e. failing test). In order to fix that, there must be `EIGEN_MAKE_ALIGNED_OPERATOR_NEW;` in every class that has Eigen members.

PR #1905, opened 2017-08-09
Enable vector extrapolation within the ComponentTransport process

TomFischer:


TomFischer:
> How did you convert the files for tests?

I wrote a small python script. I'll make a PR to the https://github.com/ufz/ogs-utils

PR #1907, opened 2017-08-09
[CMake] Remove external projects

bilke:
This PR removes the possibility to automatically build required third-party packages as this is already integrated by Conan.

Now you have two choices to handle third-party libs:

1. Installing manually / via system packages
2. Installing with Conan (by enabling the CMake option `OGS_USE_CONAN=ON`) which is fully integrated into CMake, [see docs](https://docs.opengeosys.org/docs/devguide/advanced/conan-package-manager)

Conan installs prebuilt binaries. As not all compiler / os combinations are available the default behavior is to build missing binaries locally.
---

## Points to discuss

 ~~1. Should header-only libs (i.e. Eigen) be integrated with `git submodule`? This allows for easily changing libraries version (with e.g. `cd ThirdParty/Eigen; git checkout 3.3`) and we have less Conan packages to manage.~~ After discussion: No for the moment.

## Conan migration guide

DELETED, not required anymore

## TODOs

- [x] Update docs.opengeosys.org -> https://docs.opengeosys.org/docs/devguide/advanced/conan-package-manager


scripts/cmake/CompilerSetup.cmake
```diff
@@ -148,3 +160,4 @@ endif()
 # preceding cxx flags definition in order to override earlier flags, e.g. for
 # optimization.
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OGS_CXX_FLAGS}")
+add_compile_options(${OGS_CXX_FLAGS})
```
endJunction: Just curious, why is this `add_compile_options` needed beside the setting of the `CMAKE_CXX_FLAGS`?

scripts/jenkins/gcc.groovy
```diff
@@ -76,11 +76,30 @@ image.inside(defaultDockerArgs) {
...
+    step([$class: 'WarningsPublisher', canResolveRelativePaths: false,
+        messagesPattern: """
+            .*DOT_GRAPH_MAX_NODES.
+            .*potential recursive class relation.*""",
+        parserConfigurations: [[parserName: 'Doxygen', pattern:
+        'build/DoxygenWarnings.log']], unstableTotalAll: '0'])
+
+    if (helper.isOriginMaster(this)) {
+        sshagent(credentials: ['www-data_jenkins']) {
+            sh 'rsync -a --delete --stats -e "ssh -o StrictHostKeyChecking=no"' +
```
chleh: Is it wise to put `StrictHostKeyChecking=no` here? In this case it might not be a problem, because we just copy some public data there, probably using public key authentication. But in general host keys should be checked. If editing the usual known_hosts file is not an option, there also the possibility to specify a `-oUserKnownHostsFile` on the command line.

scripts/cmake/CompilerSetup.cmake
```diff
@@ -49,30 +49,38 @@ endif()
...
+            using ${CMAKE_CXX_COMPILER_VERSION}.")
     endif()
     if(NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
         message(STATUS "Set GCC release flags")
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O3 -DNDEBUG")
     else()
         # Enable assertions in STL in debug mode.
-        if (NOT STL_NO_DEBUG)
-            set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -D_GLIBCXX_DEBUG -D_GLIBCXX_DEBUG_ASSERT -D_GLIBCXX_DEBUG_PEDASSERT -D_GLIBCXX_DEBUG_VERIFY")
+        if(NOT STL_NO_DEBUG)
```
endJunction: This is an extra option introduced in #807.

endJunction:
In the ctests the  `Adaptive_dt_ThermalConvectionFlow2D` benchmark is failing for some reasons but there were no changes to the actual FEM code.
Could someone have a look on these? ( @wenqing or @TomFischer )?

scripts/cmake/CompilerSetup.cmake
chleh: Ah, I see. Thanks for telling me. But actually we could ask ourselves if that option shouldn't be better removed from our cmake config, because (i) it's rather special, i.e., it is not an OGS feature in any way (ii) every pro-user can set it in their custom cmake config.
But I agree, `STL_NO_DEBUG` is a useful option. If we wanted to remove it, we should put it into some "How to debug" document.

wenqing:
I am going to take a look at the ctest.

wenqing:
@bilke @endJunction 
I ran the tests on Jenkins at my account on envinf1. All the tests are passed.
The failed test  in PR #1907 is
```
 [gcc] 170/199 Test #145: ogs-Adaptive_dt_ThermalConvectionFlow2D_Constant_Viscosity-time-vtkdiff ...................***Failed 
```
Since the directory of  /data/ogs/jenkins/workspace/7-ELU3QZ43AJ/build/ is gone, I cannot find any clue to the failure. Could you please restart tests on the jenkins?

scripts/cmake/CompilerSetup.cmake
```diff
@@ -49,30 +49,38 @@ endif()
...
+            set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} \
+                -D_GLIBCXX_DEBUG -D_GLIBCXX_DEBUG_ASSERT \
+                -D_GLIBCXX_DEBUG_PEDASSERT -D_GLIBCXX_DEBUG_VERIFY")
         endif()
     endif()
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CPU_FLAGS} -Wno-deprecated -Wall -Wextra -fext-numeric-literals")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CPU_FLAGS} -Wno-deprecated -Wall \
+        -Wextra -fext-numeric-literals")
```
bilke: Somehow this was necessary but I don't remember the details..

scripts/cmake/CompilerSetup.cmake
```diff
@@ -91,20 +99,22 @@ endif() # COMPILER_IS_INTEL
...
-        message(STATUS "When using profiling you should set CMAKE_BUILD_TYPE to Release.")
+        message(STATUS "When using profiling you should set CMAKE_BUILD_TYPE \
+            to Release.")
     endif()
     set(PROFILE_FLAGS "-pg -fno-omit-frame-pointer -O2 -DNDEBUG")
     # clang compiler does not know the following flags
     if(NOT COMPILER_IS_CLANG)
-        set(PROFILE_FLAGS "${PROFILE_FLAGS} -fno-inline-functions-called-once -fno-optimize-sibling-calls")
+        set(PROFILE_FLAGS "${PROFILE_FLAGS} -fno-inline-functions-called-once \
+            -fno-optimize-sibling-calls")
```
bilke: We had a Jenkinks job in the past with this enabled...

scripts/jenkins/gcc.groovy
bilke: No. Initially this comes from the times as we had no ssh access to the Jenkins server ... Will remove it.

scripts/cmake/CompilerSetup.cmake
```diff
@@ -49,30 +49,38 @@ endif()
...
+            ${CMAKE_CXX_COMPILER_VERSION}")
     endif()
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CPU_FLAGS} -Wall -Wno-c++98-compat-pedantic")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CPU_FLAGS} -Wall \
+        -Wno-c++98-compat-pedantic")
     if(CMAKE_BUILD_TYPE STREQUAL "Debug")
         # Enable assertions in STL in debug mode.
-        if (NOT STL_NO_DEBUG)
-            set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -D_GLIBCXX_DEBUG -D_GLIBCXX_DEBUG_ASSERT -D_GLIBCXX_DEBUG_PEDASSERT -D_GLIBCXX_DEBUG_VERIFY")
+        if(NOT STL_NO_DEBUG)
```
bilke: Was introduced here: https://github.com/ufz/ogs/commit/6a0dcaa1e8ea6f3e923f7db4789f67c9b14a2db8

scripts/jenkins/gcc.groovy
chleh: Thanks.

bilke:
Again everything is green except:

```
145 - ogs-Adaptive_dt_ThermalConvectionFlow2D_Constant_Viscosity-time-vtkdiff (Failed)
```

bilke:
@chleh Conan migration guide [is hosted here](https://docs.opengeosys.org/docs/devguide/troubleshooting/conan).

endJunction:
Still conan problems:
New conan, ~/.conan removed.
```sh
conan remote list                    
ogs: https://api.bintray.com/conan/ogs/conan [Verify SSL: True]
conan-center: https://conan.bintray.com [Verify SSL: True]
conan-transit: https://conan-transit.bintray.com [Verify SSL: True]
```

then building with `conan install ../s -u --build=missing` builds VTK and bzip2, but fails to build zlib:
```
PROJECT: 
zlib/1.2.8@lasote/stable: ERROR: Package 'bfe54884bf4e83adb9a2749ddbd047b02be54c43' build failed
zlib/1.2.8@lasote/stable: WARN: Build folder /home/naumov/.conan/data/zlib/1.2.8/lasote/stable/build/bfe54884bf4e83adb9a2749ddbd047b02be54c43
ERROR: zlib/1.2.8@lasote/stable: Error in build() method, line 39
        if self.settings.compiler != "gcc" or self.settings.compiler.version > "4.1":
        TypeError: '>' not supported between instances of 'SettingsItem' and 'str'
```


Is it possible to use system's VTK and zlib and bzip?

bilke:
@endJunction zlib and bzip2 should not be necassary, see https://github.com/lasote/conan-boost/issues/52.

bilke:
@endJunction This is now ready to merge. The issues on envinf1 with gcc 6.2 could not be resolved so far, people on the internet even mention a bug in `ld` and suggest to upgrade it but this is not possible at envinf1 at least I think ..

chleh:
> The issues on envinf1 with gcc 6.2 could not be resolved so far

Does that mean that we won't be able to build on envinf1 with the "standard" set of libraries anymore once this is merged?

bilke:
@chleh gcc 6.2 and libraries loaded via `module` still works. Just the combination gcc 6.2 and VTK from Conan does not work. Haven't had time to investigate this further.

endJunction:
If using system libraries is still possible :fast_forward: from me.
@TomFischer will check it on his computer once more and merge it.

chleh:
@bilke thank you for the clarification.

bilke:
@endJunction @chleh @wenqing Thanks for your feedback! I have added another last commit (finally) which gives complete control over what is used directly from Conan binary packages and what is build locally. So for example if Conan VTK on envinf1 does not work with gcc 6.2 one can still use Conan but just build VTK locally:

```bash
cmake ../ogs -DOGS_USE_CONAN=ON -DOGS_CONAN_BUILD=VTK
```

TomFischer:
I checked out the branch and created a new build folder. Then in the build folder I did:
`
cmake ../sources/ -DOGS_USE_CONAN=On
` 
and obtain the following error message:
`
-- Conan adding ogs remote repositoy (https://api.bintray.com/conan/ogs/conan)                 
usage: conan remote [-h]                       
                    {list,add,remove,update,list_ref,add_ref,remove_ref,update_ref}            
                    ...                        
conan remote: error: unrecognized arguments: -i                                                
-- Conan adding community remote repositoy (https://api.bintray.com/conan/conan-community/conan)
usage: conan remote [-h]                       
                    {list,add,remove,update,list_ref,add_ref,remove_ref,update_ref}            
                    ...                        
conan remote: error: unrecognized arguments: -i                                                
-- Third-party libraries:                      
-- Conan ** WARNING** : This detection of settings from cmake is experimental and incomplete. Please check 'conan.cmake' and contribute
-- Conan executing: conan install -g cmake -s build_type=Debug -s os=Linux -s compiler=gcc -s compiler.version=7.1 -s compiler.libcxx=libstdc++11 --build=missing --update
ERROR: Invalid setting '7.1' is not a valid 'settings.compiler.version' value.                 
Possible values are ['4.1', '4.4', '4.5', '4.6', '4.7', '4.8', '4.9', '5.1', '5.2', '5.3', '5.4', '6.1', '6.2', '6.3']
Read "http://docs.conan.io/en/latest/faq/troubleshooting.html#error-invalid-setting"           
CMake Error at scripts/cmake/conan/conan.cmake:240 (message):                                  
  Conan install failed='1' 
`

What is wrong here? It seems the problem is that my compiler is gcc 7.1.
@bilke Some time ago there was a small tutorial in this PR how to use the new scheme when conan was used before. Unfortunately, I can not find it anymore.

bilke:
@TomFischer Thanks for testing! Your Conan is way too old! Please upgrade it (via `pip install conan --upgrade`) to at least 0.26.0 (with pip you should get the latest version which is 0.26.1). 

gcc 7.1 should ok with the upgraded Conan but please be aware that for the moment there are no prebuilt packages for gcc 7.x so it will build them locally.

- [ ] add a Conan version check
- [ ] build gcc 7.x Conan packages

chleh:
Just out of curiosity: Why do we need gcc 7.1 if conan provides older versions of gcc prebuilt?

chleh:
@TomFischer:
> @chleh Conan migration guide [is hosted here](https://docs.opengeosys.org/docs/devguide/troubleshooting/conan).

chleh:
Btw: Did anybody test this PR under Windows?

bilke:
@chleh I have to update the docs. The migration should not be necessary anymore. Jenkins tested this PR on Windows 😜..

@TomFischer I have built gcc 7.1 packages for VTK. A gcc 7.2 build environment (a Docker container) for creating the packages is not yet available from Conan. Although 7.1 packages should be compatible with the 7.2 compiler I guess Conan will complain about this version mismatch so maybe you have to built them locally anyway...

chleh:
I'm so glad, Jenkins always cares for us 😄 (and @bilke always cares for Jenkins!).

endJunction:
@bilke What was the necessary conan remote for the VTK and other OGS specialities? I don't see it neither here nor in the docs page any longer... ;(

bilke:
@endJunction You can find the remote here: https://github.com/ufz/ogs/blob/master/scripts/cmake/ConanSetup.cmake#L67.

I have to check what we can do if we use a compiler which is not yet known by Conan...

PR #1895, opened 2017-08-03
Phasefield_wip

xingyuanmiao:
The mechanical phase-field process was implemented. Currently, the implementation has a 3D homogeneous unconfined compression test. Other three large tests(3D cracking beam; tension of single-edge-notch test; shear of single-edge-notch test ) will be added soon.

PR #1908, opened 2017-08-10
Modified return type of function createLinearElasticIsotropic as template type

xingyuanmiao:
This general function will be used in mechanical processes, PR #1895.


wenqing:
Closed because there is a better solution.

PR #1910, opened 2017-08-10
Two improvements

wenqing:
1. Removed LocalDataInitializer.h and  createLocalAssemblers.h from ThermoMechanics and let ThermoMechanics use these two files in SmallDeformation.
 The reason is that LocalAssemblers in ThermoMechanics and SmallDeformation have the same interface. 
  _Note: In ThermoMechanics, elements for deformation equation and heat conduction equation have the same order. It's Ok under the fact that T and M are weakly coupled._

2. Modified the warning messages when cell/node property input files do not exist in running ogs6#PETSc.
   The reason to this change is that Jobst got the following warning in running ogs6#PETSc, and he could not know whether it indicates an error or just a warning
     ```
    warning: Could not open file 'meshing_007_partitioned_node_properties_cfg4.bin in binary model'.
     ```
    The modified warning gives the following message
    ```
    warning: Could not open file 'meshing_007_partitioned_node_properties_cfg4.bin'.
	       You can ignore this warning if the mesh does not contain node-wise property data.
   ```


chleh:
:+1:

PR #1920, opened 2017-08-22
Set coupling term as a member of class Process

wenqing:
For the new change of velocity calculation (PR #1874), the coupling term is not considered. As discussed with Christoph, the coupling term could be set as one member of class Process in order not to change the interface  of velocity calculation.

In this PR,
1) a pointer ~weak_ptr~ to StaggeredCouplingTerm, which only contains reference type members, is added as a member of  class Process
2) a pointer to an instance of StaggeredCouplingTerm, which is created  on fly before global assembly, is passed to process class.
~2) a shared_ptr to StaggeredCouplingTerm is created  on fly before global assembly and passed to instance of process class.~
3) StaggeredCouplingTerm type argument is removed from all member functions that have it.
~4) Avoided if-condition about coupling inside element-wise loop during global assembly by adding two members in VectorMatrixAssembler.~

~Note: The first two commits are only related to code format, and they are put in another PR.~

### DIFF
--- a/NumLib/ODESolver/NonlinearSolver.cpp
+++ b/NumLib/ODESolver/NonlinearSolver.cpp
@@ -21,14 +21,13 @@
 namespace NumLib
 {
 void NonlinearSolver<NonlinearSolverTag::Picard>::assemble(
-    GlobalVector const& x,
-    ProcessLib::StaggeredCouplingTerm const& coupling_term) const
+    GlobalVector const& x) const
 {
-    _equation_system->assemble(x, coupling_term);
+    _equation_system->assemble(x);
 }
 
 bool NonlinearSolver<NonlinearSolverTag::Picard>::solve(
-    GlobalVector& x, ProcessLib::StaggeredCouplingTerm const& coupling_term,
+    GlobalVector& x,
     std::function<void(unsigned, GlobalVector const&)> const& postIterationCallback)
 {
     namespace LinAlg = MathLib::LinAlg;
@@ -59,7 +58,7 @@ bool NonlinearSolver<NonlinearSolverTag::Picard>::solve(
 
         BaseLib::RunTime time_assembly;
         time_assembly.start();
-        sys.assemble(x, coupling_term);
+        sys.assemble(x);
         sys.getA(A);
         sys.getRhs(rhs);
         INFO("[time] Assembly took %g s.", time_assembly.elapsed());
@@ -159,17 +158,16 @@ bool NonlinearSolver<NonlinearSolverTag::Picard>::solve(
 }
 
 void NonlinearSolver<NonlinearSolverTag::Newton>::assemble(
-    GlobalVector const& x,
-    ProcessLib::StaggeredCouplingTerm const& coupling_term) const
+    GlobalVector const& x) const
 {
-    _equation_system->assemble(x, coupling_term);
+    _equation_system->assemble(x);
     // TODO if the equation system would be reset to nullptr after each
     //      assemble() or solve() call, the user would be forced to set the
     //      equation every time and could not forget it.
 }
 
 bool NonlinearSolver<NonlinearSolverTag::Newton>::solve(
-    GlobalVector& x, ProcessLib::StaggeredCouplingTerm const& coupling_term,
+    GlobalVector& x,
     std::function<void(unsigned, GlobalVector const&)> const& postIterationCallback)
 {
     namespace LinAlg = MathLib::LinAlg;
@@ -202,7 +200,7 @@ bool NonlinearSolver<NonlinearSolverTag::Newton>::solve(
 
         BaseLib::RunTime time_assembly;
         time_assembly.start();
-        sys.assemble(x, coupling_term);
+        sys.assemble(x);
         sys.getResidual(x, res);
         sys.getJacobian(J);
         INFO("[time] Assembly took %g s.", time_assembly.elapsed());
--- a/ProcessLib/ComponentTransport/ComponentTransportProcess.cpp
+++ b/ProcessLib/ComponentTransport/ComponentTransportProcess.cpp
@@ -56,22 +56,20 @@ void ComponentTransportProcess::assembleConcreteProcess(
     GlobalVector const& x,
     GlobalMatrix& M,
     GlobalMatrix& K,
-    GlobalVector& b,
-    StaggeredCouplingTerm const& coupling_term)
+    GlobalVector& b)
 {
     DBUG("Assemble ComponentTransportProcess.");
 
     // Call global assembler for each local assembly item.
     GlobalExecutor::executeMemberDereferenced(
         _global_assembler, &VectorMatrixAssembler::assemble, _local_assemblers,
-        *_local_to_global_index_map, t, x, M, K, b, coupling_term);
+        *_local_to_global_index_map, t, x, M, K, b, *_coupling_term.lock());
 }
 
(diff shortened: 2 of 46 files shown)


ProcessLib/ComponentTransport/ComponentTransportProcess.cpp
```diff
@@ -56,30 +56,50 @@ void ComponentTransportProcess::assembleConcreteProcess(
...
+    GlobalVector& b, GlobalMatrix& Jac)
 {
     DBUG("AssembleWithJacobian ComponentTransportProcess.");
 
     // Call global assembler for each local assembly item.
-    GlobalExecutor::executeMemberDereferenced(
-        _global_assembler, &VectorMatrixAssembler::assembleWithJacobian,
-        _local_assemblers, *_local_to_global_index_map, t, x, xdot, dxdot_dx,
-        dx_dx, M, K, b, Jac, coupling_term);
+    // Call global assembler for each local assembly item.
```
endJunction: duplicate comment

wenqing: Removed by amending a commit.

ProcessLib/Process.h
```diff
@@ -200,6 +203,10 @@ class Process
 
     VectorMatrixAssembler _global_assembler;
 
+    /// Coupled processes.
+    mutable std::weak_ptr<StaggeredCouplingTerm> _coupling_term;
```
TomFischer: Maybe it's a stupid question, but does a process couple always with only one other process? If there would be more than one other process to couple with, shouldn't the `_coupling_term` be a vector of weak_ptr's?

wenqing: It is a common question.
Such case is included in struct StaggeredCouplingTerm, which has only two members to store the references to the coupled processes and the references to the solutions of the coupled processes:
```
    /// References to the coupled processes are distinguished by the keys of
    /// process types.
    std::unordered_map<std::type_index, Process const&> const&
        coupled_processes;

    /// References to the current solutions of the coupled processes.
    /// The coupled solutions are distinguished by the keys of process types.
    std::unordered_map<std::type_index, GlobalVector const&> const& coupled_xs;
```

chleh: IMHO it could be a `StaggeredCouplingTerm* _coupling_term`, and it probably does not have to be mutable:
* no weak_ptr, because it is unlikely that the time loop destroys processes while they are still running
* not mutable, because if you have
```
struct S { int* i; };
S const obj;
```
you could still do: `obj->i = 8`. I.e.: The value of `i` is mutable although `obj` is not. Cf. `int const*` vs `int* const`.

ProcessLib/UncoupledProcessesTimeLoop.cpp
```diff
@@ -946,12 +940,15 @@ bool UncoupledProcessesTimeLoop::solveCoupledEquationSystemsByStaggeredScheme(
                 spd->process.preTimestep(x, t, dt);
             }
 
-            StaggeredCouplingTerm coupling_term(
-                spd->coupled_processes,
-                _solutions_of_coupled_processes[pcs_idx], dt);
+            std::shared_ptr<StaggeredCouplingTerm> coupling_term =
```
chleh: Maybe no need for `std::shared_ptr` here. Related to some of my other comments.

NumLib/ODESolver/NonlinearSolver.cpp
```diff
@@ -21,14 +21,13 @@
 namespace NumLib
 {
 void NonlinearSolver<NonlinearSolverTag::Picard>::assemble(
-    GlobalVector const& x,
-    ProcessLib::StaggeredCouplingTerm const& coupling_term) const
+    GlobalVector const& x) const
 {
-    _equation_system->assemble(x, coupling_term);
+    _equation_system->assemble(x);
```
chleh: Nice! It now looks more like an ODE again.

ProcessLib/VectorMatrixAssembler.cpp
```diff
@@ -58,6 +58,42 @@ VectorMatrixAssembler::VectorMatrixAssembler(
 void VectorMatrixAssembler::assemble(
     const std::size_t mesh_item_id, LocalAssemblerInterface& local_assembler,
     const NumLib::LocalToGlobalIndexMap& dof_table, const double t,
+    const GlobalVector& x, GlobalMatrix& M, GlobalMatrix& K, GlobalVector& b)
+{
+    auto const indices = NumLib::getIndices(mesh_item_id, dof_table);
+    auto const local_x = x.get(indices);
```
chleh: I'm confused: This whole bunch of code looks like it is copied from somewhere. But from where?

ProcessLib/ComponentTransport/ComponentTransportProcess.cpp
```diff
@@ -61,9 +61,19 @@ void ComponentTransportProcess::assembleConcreteProcess(
     DBUG("Assemble ComponentTransportProcess.");
 
     // Call global assembler for each local assembly item.
-    GlobalExecutor::executeMemberDereferenced(
-        _global_assembler, &VectorMatrixAssembler::assemble, _local_assemblers,
-        *_local_to_global_index_map, t, x, M, K, b, *_coupling_term.lock());
+    if (_coupling_term.lock())
+    {
+        GlobalExecutor::executeMemberDereferenced(
```
chleh: I'm not sure if avoiding the `if (_coupling_term.lock())` inside the `VectorMatrixAssembler` is worth that much code duplication...

ProcessLib/Process.h
wenqing: Changed to pointer type.

ProcessLib/UncoupledProcessesTimeLoop.cpp
wenqing: Removed std::shared_ptr.

ProcessLib/VectorMatrixAssembler.cpp
wenqing: Copied from VectorMatrixAssembler::assemble and VectorMatrixAssembler::assembleWithJacobian. Now, these changes are dropped.

ProcessLib/ComponentTransport/ComponentTransportProcess.cpp
wenqing: The changes are dropped by removing a commit.

wenqing:
The description of this PR is modified due to the current changes in PR.

ProcessLib/UncoupledProcessesTimeLoop.cpp
```diff
@@ -946,12 +940,14 @@ bool UncoupledProcessesTimeLoop::solveCoupledEquationSystemsByStaggeredScheme(
                 spd->process.preTimestep(x, t, dt);
             }
 
-            StaggeredCouplingTerm coupling_term(
+            StaggeredCouplingTerm coupling_term = StaggeredCouplingTerm(
```
chleh: Minor: That could stay as it was.

wenqing: Recovered by amending a commit.

wenqing: also initialized _coupling_term.

wenqing:
@endJunction Could you please restart the test on jenkins? On jenkins, the failure message is
`Log parsing has failed`.

PR #1921, opened 2017-08-22
Removed multiple private and protected in class Processs

wenqing:
~Removed multiple private and protected in class Processs, and~ applied clang-format to Process.h.

This PR contains the firs two commits in PR #1920, and it is expected to be merged before  PR #1920.

### DIFF
--- a/ProcessLib/Process.h
+++ b/ProcessLib/Process.h
@@ -93,7 +93,6 @@ public:
     }
 
     MeshLib::Mesh& getMesh() const { return _mesh; }
-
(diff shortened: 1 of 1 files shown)


chleh:
Clang-format is OK. Personally I'd prefer that the members are not moved around in the class declaration. That makes it much easier to track changes, e.g., when rebasing old feature branches.

wenqing:
@chleh Only the change by  clang-format is left.

chleh:
Thanks. :+1:

PR #1923, opened 2017-08-23
[CMake] Copy CXX to C flags.

bilke:
This adds the flags setup in `CompilerSetup.cmake` for C libraries such `metis` too. Tom needed this to add cpu arch flags when compiling on cluster systems.


chleh:
I'm not really happy with that: The CompilerSetup.cmake already looks rather intransparent to me. This PR doesn't make things (much) worse, but also does not improve things.
But since the feature added in this PR is obviously convenient: :+1:

endJunction:
Technically ok. Not sure about it though, because the c++-flags could be specific to c++ and not compatible to the c-flags.

PR #1924, opened 2017-08-23
Enforced git hooks

bilke:
### PR is for discussion

This enables enforced (after running CMake) git hooks. I added a pre-commit hook which checks for correctly committing files with git-lfs. Can be extended to more checks (e.g. whitespace, styling, commit message style, ...). Remember that you can always disable checks by committing with `--no-verify` or `-n`. Haven't tested if this works on Windows too (as the hooks are bash scripts).

### DIFF
--- a/.gitconfig
+++ b/.gitconfig
@@ -1,0 +1,4 @@
+[core]
+    hooksPath = .hooks
+[submodule]
+	fetchJobs = 4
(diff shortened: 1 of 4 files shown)


chleh:
I think it's good that you share your git hooks. But I don't know if it is the best idea to automatically set them up. At least we should thorougly test them before doing so.

endJunction:
As discussed, I'd also add hooks for ogs development, but do not enforce them automatically.

endJunction:
Let's first update this PR such that the hooks are available for everyone. This means only the `.hooks/pre-commit` file remains. After this is tested by several developers, preferably on different systems, we reopen the "make it automatic" discussion.

PR #1929, opened 2017-08-28
Richards Component Transport

TomFischer:
~~Please do not review - this PR is to test the generation of the documentation.~~

Please review!


chleh:
There is a huge ParaView state file in this PR. Can it be removed, or moved to the data repo?

waltherm:
@chleh You are right, this state file should be in data... ( @TomFischer can you move this?)

Documentation/ProjectFile/prj/processes/process/RichardsComponentTransport/t_specific_body_force.md
```diff
@@ -0,0 +1,3 @@
+Specific body forces applied to fluid.
+
+It is usually used to apply gravitational forces. Given as a 2d or 3d vector.
```
endJunction: Nit picking: it's probably copy-paste, but I'd think it also applies to 1d problems...

ProcessLib/HT/HTFEM.h
```diff
@@ -306,6 +306,8 @@ class LocalAssemblerData : public HTLocalAssemblerInterface
             auto const& N = ip_data.N;
             auto const& dNdx = ip_data.dNdx;
 
+            pos.setIntegrationPoint(ip);
```
endJunction: This can be directly be pushed to the master. Has nothing to do with Richards...

ProcessLib/RichardsComponentTransport/RichardsComponentTransportFEM.h
```diff
@@ -0,0 +1,444 @@
...
+            _integration_method.getNumberOfPoints();
+
+        auto const indices = NumLib::getIndices(_element_id, dof_table);
+        assert(!indices.empty());
+        auto const local_x = current_solution.get(indices);
+
+        cache.clear();
+        auto cache_vec = MathLib::createZeroedMatrix<
+            Eigen::Matrix<double, 1, Eigen::Dynamic>>(
+            cache, 1, n_integration_points);
```
endJunction: Not sure if the mapping is needed, because the cache is accessed only once by index (of integration point). But you might want to keep it if this style already became idiomatic.
The access below would be `cache[ip] = Sw`.

ProcessLib/RichardsComponentTransport/RichardsComponentTransportFEM.h
```diff
@@ -0,0 +1,444 @@
...
+        auto local_b = MathLib::createZeroedVector<LocalVectorType>(
+            local_b_data, local_matrix_size);
+
+        unsigned const n_integration_points =
+            _integration_method.getNumberOfPoints();
+
+        SpatialPosition pos;
+        pos.setElementID(_element_id);
+
+        auto const num_nodes = ShapeFunction::NPOINTS;
```
endJunction: Just a comment about variable naming: shape function's number of points is not always equal to the number of nodes of an element. Maybe `n_points`? Feel free to ignore the suggestion.

endJunction: In some other multi variable processes, e.g. [HM]( https://github.com/ufz/ogs/blob/master/ProcessLib/HydroMechanics/HydroMechanicsFEM.h#L587-L591), there are class variables X_index and X_size for each variable X (pressure, displacement, temperature,...). Then the submatrix creation looks like:
```c++
auto KCC = local_K.template block<C_size, C_size>(C_index, C_index);
auto KCp = local_K.template block<C_size, p_size>(C_index, p_index);
```

endJunction:
I'd recommend to split the Process.h into declarations and an implementation parts, where the latter Process-impl.h is _only_ included in the Process.cpp. This could save some compilation time.
See, for example HM process again with [template instantiation declarations](https://github.com/ufz/ogs/blob/master/ProcessLib/HydroMechanics/HydroMechanicsProcess.h#L83-L84) and the explicit instantiations in the [HMProcess.cpp](https://github.com/ufz/ogs/blob/master/ProcessLib/HydroMechanics/HydroMechanicsProcess.cpp).

TomFischer:
I don't understand your last comment concerning the splitting of the process into declarations and implementations. The RichardsComponentTransport process isn't a template. Maybe you want to suggest to split the FEM?

ProcessLib/RichardsComponentTransport/RichardsComponentTransportFEM-impl.h
```diff
@@ -0,0 +1,352 @@
...
+        auto const& dNdx = ip_data.dNdx;
+        auto const& w = ip_data.integration_weight;
+
+        double C_int_pt = 0.0;
+        double p_int_pt = 0.0;
+        // Order matters: First C, then p!
+        NumLib::shapeFunctionInterpolate(local_x, N, C_int_pt, p_int_pt);
+
+        double const pc_int_pt = -p_int_pt;
+        double const Sw = (pc_int_pt > 0)
```
wenqing: You can remove this ternary condition because  (pc_int_pt > 0) is already checked inside 
```
_process_data.porous_media_properties
                           .getCapillaryPressureSaturationModel(t, pos)
                                 .getSaturation(pc_int_pt)
```
as
```
double BrooksCoreyCapillaryPressureSaturation::getSaturation(
    const double capillary_pressure) const
{
    const double pc =
        (capillary_pressure < 0.0) ? _minor_offset : capillary_pressure;
     ...
    return MathLib::limitValueInInterval(S, _saturation_r + _minor_offset,
                                         _saturation_max - _minor_offset);
```

ProcessLib/RichardsComponentTransport/RichardsComponentTransportFEM-impl.h
```diff
@@ -0,0 +1,352 @@
...
+                              ? _process_data.porous_media_properties
+                                    .getCapillaryPressureSaturationModel(t, pos)
+                                    .getSaturation(pc_int_pt)
+                              : 1.0;
+
+        double const dSw_dpc =
```
wenqing: The same reason to remove this ternary condition. With the change, the benchmark results will be changed slightly if the given maximum saturation is less than 1.

ProcessLib/RichardsComponentTransport/RichardsComponentTransportFEM-impl.h
```diff
@@ -0,0 +1,352 @@
...
+        auto const mu = _process_data.fluid_properties->getValue(
+            MaterialLib::Fluid::FluidPropertyType::Viscosity, vars);
+
+        double C_int_pt = 0.0;
+        double p_int_pt = 0.0;
+        NumLib::shapeFunctionInterpolate(local_x, N, C_int_pt, p_int_pt);
+
+        // saturation
+        double const pc_int_pt = -p_int_pt;
+        double const Sw = (pc_int_pt > 0)
```
wenqing: The same here

ProcessLib/RichardsComponentTransport/RichardsComponentTransportFEM-impl.h
```diff
@@ -0,0 +1,352 @@
...
+        auto const& ip_data = _ip_data[ip];
+        auto const& N = ip_data.N;
+
+        double C_int_pt = 0.0;
+        double p_int_pt = 0.0;
+        NumLib::shapeFunctionInterpolate(local_x, N, C_int_pt, p_int_pt);
+
+        // saturation
+        double const pc_int_pt = -p_int_pt;
+        double const Sw = (pc_int_pt > 0)
```
wenqing: Here, the same.

ProcessLib/RichardsComponentTransport/RichardsComponentTransportFEM-impl.h
TomFischer: Thanks for the hint. Changed all locations.

Issue #1926, opened 2017-08-25
SegFault with not found BC point in mesh


waltherm:
OGS crashes with a SegFault if a point, where a boundary condition is defined (in .prj), cannot be defined on the mesh.

In the enclosed example, the second point (name "top") is at x=1, while the mesh ends at x=0.25. Changing the point in .gml to x=0.25 will make the example work.

[saturated.zip](https://github.com/ufz/ogs/files/1251522/saturated.zip)

endJunction:
error: Unknown process type: RichardsComponentTransport at ProjectData.cpp, line 530

waltherm:
@endJunction ups... yes, this is based on one of @TomFischer 's branches. Here is the PR: https://github.com/ufz/ogs/pull/1929
(sorry)

endJunction:
I'd say it's a feature, not a bug. If the gml does not match the underlying mesh closely, it's a user error.
We could improve the error message though.

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/1926).

PR #1932, opened 2017-08-29
[MGTL] BoundaryElems; Change assertion to error.

endJunction:
The BoundaryElementsAtPoint algorithm expects the given
point to be found in the mesh. If this fails the preconditions
for the algorithm are not fulfilled.

Closes https://github.com/ufz/ogs/issues/1926

### DIFF
--- a/MeshGeoToolsLib/BoundaryElementsAtPoint.cpp
+++ b/MeshGeoToolsLib/BoundaryElementsAtPoint.cpp
@@ -25,7 +25,18 @@ BoundaryElementsAtPoint::BoundaryElementsAtPoint(
     : _mesh(mesh), _point(point)
 {
     auto const node_ids = mshNodeSearcher.getMeshNodeIDs(_point);
-    assert(node_ids.size() == 1);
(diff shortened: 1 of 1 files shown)


MeshGeoToolsLib/BoundaryElementsAtPoint.cpp
```diff
@@ -25,7 +25,11 @@ BoundaryElementsAtPoint::BoundaryElementsAtPoint(
     : _mesh(mesh), _point(point)
 {
     auto const node_ids = mshNodeSearcher.getMeshNodeIDs(_point);
-    assert(node_ids.size() == 1);
+    if (node_ids.size() != 1)
```
chleh: Can `node_ids.size()` be bigger than one? E.g. if the search length is too big. If so, I suggest that there be two distinct cases and two different messages:
`if (node_ids.empty()) ... not found ...`
and
`if (node_ids.size() > 1) ... maybe search length too big ...`

waltherm:
:+1: super!

Issue #1839, opened 2017-06-16
Add prj-file option vtk output mode


bilke:
Add option in project files for vtk ascii output.

https://github.com/ufz/ogs/blob/master/ProcessLib/ProcessOutput.cpp#L184-L185

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/1839).

PR #1779, opened 2017-04-24
Apply clang modernize fixes

endJunction:
Almost all of the changes vere done automatically by clang-tidy followed by clang-format.

Used clang-tidy checks:
 - modernize-loop-convert
 - modernize-pass-by-value
 - modernize-use-auto
 - modernize-use-emplace
 - modernize-use-equals-default
 - modernize-use-equals-delete
 - modernize-use-nullptr
 - modernize-use-override
 - modernize-use-using

Have a look on each commit to see the style of the changes.

The code is compiled for analysis, so all parts inaccessible in compilation (like petsc code) are not changed.

PR #1935, opened 2017-08-29
 [PL] Add Ascii, Binary, and Appended output option

endJunction:
Closes https://github.com/ufz/ogs/issues/1839

### DIFF
--- a/ProcessLib/Output.h
+++ b/ProcessLib/Output.h
@@ -84,20 +84,20 @@ private:
         MeshLib::IO::PVDFile pvd_file;
     };
 
-    Output(std::string output_directory, std::string prefix, bool const compress_output,
-           bool const output_nonlinear_iteration_results)
-        : _output_directory(std::move(output_directory)),
-          _output_file_prefix(std::move(prefix)),
-          _output_file_compression(compress_output),
-          _output_nonlinear_iteration_results(
-              output_nonlinear_iteration_results)
-    {}
+    Output(std::string output_directory, std::string prefix,
+           bool const compress_output, std::string const& data_mode,
+           bool const output_nonlinear_iteration_results);
 
     std::string const _output_directory;
     std::string const _output_file_prefix;
 
     //! Enables or disables zlib-compression of the output files.
     bool const _output_file_compression;
+
+    //! Chooses vtk's data mode for output following the enumeration given in
+    /// the vtkXMLWriter: {Ascii, Binary, Appended}.  See vtkXMLWriter
+    /// documentation http://www.vtk.org/doc/nightly/html/classvtkXMLWriter.html
+    int const _output_file_data_mode;
     bool const _output_nonlinear_iteration_results;
 
     //! Describes after which timesteps to write output.
--- a/ProcessLib/ProcessOutput.cpp
+++ b/ProcessLib/ProcessOutput.cpp
@@ -42,6 +42,7 @@ ProcessOutput::ProcessOutput(BaseLib::ConfigTree const& output_config)
 
 void doProcessOutput(std::string const& file_name,
                      bool const compress_output,
+                     int const data_mode,
                      const double t,
                      GlobalVector const& x,
                      MeshLib::Mesh& mesh,
@@ -215,8 +216,7 @@ void doProcessOutput(std::string const& file_name,
 
     // Write output file
     DBUG("Writing output to \'%s\'.", file_name.c_str());
-    MeshLib::IO::VtuInterface vtu_interface(&mesh, vtkXMLWriter::Binary,
-                                            compress_output);
+    MeshLib::IO::VtuInterface vtu_interface(&mesh, data_mode, compress_output);
     vtu_interface.writeToFile(file_name);
 }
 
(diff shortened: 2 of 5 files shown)


ProcessLib/ProcessOutput.cpp
```diff
@@ -215,8 +216,7 @@ void doProcessOutput(std::string const& file_name,
 
     // Write output file
     DBUG("Writing output to \'%s\'.", file_name.c_str());
-    MeshLib::IO::VtuInterface vtu_interface(&mesh, vtkXMLWriter::Binary,
-                                            compress_output);
+    MeshLib::IO::VtuInterface vtu_interface(&mesh, data_mode, compress_output);
```
chleh: Do we have to specify the precision for Ascii output?

endJunction: It seems to me the ascii output uses `%g` from printf, so it is switching btw. `X.Y` and `X.YeZ` forms while outputting 11 significant digits for X and Y together.
I don't find any documentation about setting the output precision in vtk docu...
I guess the ascii output format is given by vtk and not customizable.

endJunction: Some example:
```xml
<DataArray type="Float64" Name="sigma_xy" format="ascii" RangeMin="-7.4359310551e-15" RangeMax="2.5418831431e-15">
-1.2844131096e-15 -2.8590975224e-15 ... 2.2671674321e-15 -6.4106780642e-15
</DataArray>
<DataArray type="Float64" Name="sigma_yy" format="ascii" RangeMin="-165.79792808" RangeMax="-165.79792808">
-165.79792808 -165.79792808 ...  -165.79792808 -165.79792808
</DataArray>
```

endJunction: Comparing ascii and binary outputs of same simulation:
```sh
➜  ogs/d bin/vtkdiff -a displacement -b displacement a.vtu cube_1e0_pcs_0_ts_100_t_0.010000.vtu 
Comparing data array `displacement' from file `a.vtu' to data array `displacement' from file `cube_1e0_pcs_0_ts_100_t_0.010000.vtu'.
Computed difference between data arrays:
abs l1 norm      = 3.093700426046175e-14
abs l2-norm^2    = 1.196238634558976e-28
abs l2-norm      = 1.093726946984016e-14
abs maximum norm = 3.866969678495424e-15

rel l1 norm      = 9.043003552305743e-11
rel l2-norm^2    = 1.022159699263067e-21
rel l2-norm      = 3.197123237010214e-11
rel maximum norm = 1.130371587200067e-11
```

ProcessLib/Output.h
```diff
@@ -84,20 +84,20 @@ class Output
...
-    Output(std::string output_directory, std::string prefix, bool const compress_output,
-           bool const output_nonlinear_iteration_results)
-        : _output_directory(std::move(output_directory)),
-          _output_file_prefix(std::move(prefix)),
-          _output_file_compression(compress_output),
-          _output_nonlinear_iteration_results(
-              output_nonlinear_iteration_results)
-    {}
+    Output(std::string output_directory, std::string prefix,
+           bool const compress_output, std::string const data_mode,
```
TomFischer: Why not `std::string const& data_mode` instead of `std::string const data_mode`?

chleh: This is the "take easy-to-move types by value, and move them" idiom @endJunction introduced some weeks ago. Personally I'd prefer the old-fashioned `std::string const&`...
Note: `std::string` here is not const!

ProcessLib/ProcessOutput.cpp
chleh: Thanks for checking this. Your observations seem to be correct:
https://gitlab.kitware.com/vtk/vtk/commit/47248b0d194a299e35c67af890fa350121e343c2
https://gitlab.kitware.com/vtk/vtk/blob/master/IO/XML/vtkXMLWriter.cxx#L828
and also all "stream"-related methods are protected:
http://www.vtk.org/doc/nightly/html/classvtkXMLWriter.html
So there's probably nothing we can do about that. Should we add a comment to the docs about that? In practice the precision of 11 shouldn't be a problem. There only might be some issues if somebody saves some simulation results to Ascii and then tries to start a second simulation from those results, applying too strict mesh searcher tolerances.

bilke:
👍

ProcessLib/Output.h
endJunction: Forgot the `&`. For move idiom it must be non-const.

chleh: OK. I confused `data_mode` with the other string arguments in my comment above. IMHO passing by value or by reference deliberately depending on some arbitrary notion of consistency is not the best option: IMHO either move idiom for all strings or for none.

endJunction: Move for all strings which are copied. I think that all cases were fixed in https://github.com/ufz/ogs/pull/1779 https://github.com/ufz/ogs/pull/1779/commits/88426db960647ea4fc24c2efde6981bb9121a09f.

Still (as changed in the comment above) It would be not optimal here, because there is no copy at all.

PR #1936, opened 2017-08-29
Material forces

endJunction:
In preparation.

TODO:

- [x] 3D test project file is needed @fparisio 
- [x] docu @fparisio After a submitted paper is published it will be referenced in the documentation.
- [x] ~Check internal vars output @endJunction~ Don't remember what it was about...
- [x] Free energy computation for Lubby2 material model required @nagelt

### DIFF
--- a/ProcessLib/Deformation/LinearBMatrix.h
+++ b/ProcessLib/Deformation/LinearBMatrix.h
@@ -78,6 +78,5 @@ BMatrixType computeBMatrix(DNDX_Type const& dNdx,
 
     return B;
 }
-
 }  // namespace LinearBMatrix
 }  // namespace ProcessLib
--- a/ProcessLib/Deformation/MaterialForces.h
+++ b/ProcessLib/Deformation/MaterialForces.h
@@ -1,0 +1,187 @@
+/**
+ * \file
+ *
+ * \copyright
+ * Copyright (c) 2012-2017, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#pragma once
+
+#include <cassert>
+#include <vector>
+
+#include "MathLib/LinAlg/Eigen/EigenMapTools.h"
+#include "NumLib/DOF/DOFTableUtil.h"
+#include "ProcessLib/Deformation/GMatrix.h"
+#include "ProcessLib/Parameter/SpatialPosition.h"
+#include "ProcessLib/Utils/InitShapeMatrices.h"
+
+namespace ProcessLib
+{
+namespace SmallDeformation
+{
+struct MaterialForcesInterface
+{
+    virtual std::vector<double> const& getMaterialForces(
+        std::vector<double> const& local_x,
+        std::vector<double>& nodal_values) = 0;
+
+    virtual ~MaterialForcesInterface() = default;
+};
+
+template <int DisplacementDim, typename ShapeFunction,
+          typename ShapeMatricesType, typename NodalForceVectorType,
+          typename NodalDisplacementVectorType, typename GradientVectorType,
+          typename GradientMatrixType, typename IPData,
+          typename IntegrationMethod>
+std::vector<double> const& getMaterialForces(
+    std::vector<double> const& local_x, std::vector<double>& nodal_values,
+    IntegrationMethod const& _integration_method, IPData const& _ip_data,
+    MeshLib::Element const& element, bool const is_axially_symmetric)
+{
+    unsigned const n_integration_points =
+        _integration_method.getNumberOfPoints();
+
+    nodal_values.clear();
+    auto local_b = MathLib::createZeroedVector<NodalDisplacementVectorType>(
+        nodal_values, ShapeFunction::NPOINTS * DisplacementDim);
+
+    for (unsigned ip = 0; ip < n_integration_points; ip++)
+    {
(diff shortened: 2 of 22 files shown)


endJunction:
@nagelt Do you have a formula for the free energy density of the Lubby2 model?

nagelt:
@endJunction Lubby2 added. Also the paper which will serve as docu is accepted now.

ProcessLib/SmallDeformation/SmallDeformationProcess-impl.h
```diff
@@ -69,6 +71,13 @@ void SmallDeformationProcess<DisplacementDim>::initializeConcreteProcess(
...
 
+    _material_forces->resize(DisplacementDim * mesh.getNumberOfNodes());
+
+    Base::_secondary_variables.addSecondaryVariable(
+        "free_energy_density",
+        makeExtrapolator(1, getExtrapolator(), _local_assemblers,
+                         &LocalAssemblerInterface::getIntPtFreeEnergyDensity));
+
     Base::_secondary_variables.addSecondaryVariable(
         "sigma",
```
nagelt: if one name is "free_energy_density" then the other name should be "stress" instead of "sigma". But since sigma is a very widespread symbol for stress, I'm also ok with that.

ProcessLib/SmallDeformation/Tests.cmake
```diff
@@ -84,6 +84,21 @@ AddTest(
...
+    EXECUTABLE_ARGS bar.prj
+    WRAPPER time
+    TESTER vtkdiff
+    REQUIREMENTS NOT OGS_USE_MPI
+    ABSTOL 1e-14 RELTOL 1e-15
+    DIFF_DATA
+    expected_bar_out_pcs_0_ts_2_t_1.000000.vtu bar_out_pcs_0_ts_2_t_1.000000.vtu displacement displacement
+    expected_bar_out_pcs_0_ts_2_t_1.000000.vtu bar_out_pcs_0_ts_2_t_1.000000.vtu MaterialForces MaterialForces
+    expected_bar_out_pcs_0_ts_2_t_1.000000.vtu bar_out_pcs_0_ts_2_t_1.000000.vtu free_energy_density free_energy_density
+)
```
nagelt: I'd to see both the 3D and the 2D version of the code tested. Test case can be the same (bar) with the beam's axis lying in a different coordinate plane in the 3D case compared to the 2D case. Simply because we manually coded the Eshelby stress component-wise separately for 3D and 2D, and the tests should ensure functionality of both implementations.

ProcessLib/Deformation/MaterialForces.h
```diff
@@ -0,0 +1,188 @@
...
+
+    GlobalExecutor::executeDereferenced(
+        [](const std::size_t mesh_item_id,
+           LocalAssemblerInterface& local_assembler,
+           const NumLib::LocalToGlobalIndexMap& dof_table,
+           GlobalVector const& x,
+           std::vector<double>& node_values) {
+            auto const indices = NumLib::getIndices(mesh_item_id, dof_table);
+            std::vector<double> local_data;
+            auto const local_x = x.get(indices);
```
wenqing: need to call MathLib::LinAlg::setLocalAccessibleVector(x) before x.get.

ProcessLib/Deformation/MaterialForces.h
```diff
@@ -0,0 +1,188 @@
...
+
+            local_assembler.getMaterialForces(local_x, local_data);
+
+            assert(local_data.size() == indices.size());
+            for (std::size_t i = 0; i < indices.size(); ++i)
+                node_values[indices[i]] += local_data[i];
```
wenqing: ``indices`` contains the global indices. 

Just checked the dropped ProcessLib/SmallDeformationCommon/Common.h again, where nodal force vector was change to a global vector in the first commit of PR #1838. Hereby  ``node_values`` is a local one in partition.  Unfortunately, we do not store local indices of partition. Is it possible to change the type of ``node_values`` to GlobalVector?
 
PS: I remembered that we had discussed to change the property vector type to GlobalVector.

wenqing: The output is already tailed for the global vector, where the global values are copied to a local vector. See
https://github.com/ufz/ogs/blob/master/ProcessLib/ProcessOutput.cpp#L58

ProcessLib/Deformation/MaterialForces.h
```diff
@@ -0,0 +1,188 @@
...
+                               (DisplacementDim == 2 ? 1 : 0));
+
+        if (DisplacementDim == 3)
+        {
+            eshelby_stress[0] = eshelby_stress[DisplacementDim + 1] =
+                eshelby_stress[8] = psi;
+
+            eshelby_stress[0] -= sigma[0] * grad_u[0] +
+                                 sigma[3] / std::sqrt(2) * grad_u[3] +
+                                 sigma[5] / std::sqrt(2) * grad_u[6];
```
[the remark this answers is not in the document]
endJunction: The assumption is not true ;) sqrt(2) is evaluated at compile time.

ProcessLib/Deformation/LinearBMatrix.h
```diff
@@ -79,5 +79,69 @@ BMatrixType computeBMatrix(DNDX_Type const& dNdx,
     return B;
 }
 
+/// Fills a G-matrix based on given shape function dN/dx values.
```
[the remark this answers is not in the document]
endJunction: Moved to `GMatrix.h` in `ProcessLib::Deformation` namespace.

PR #1930, opened 2017-08-28
THM

grubbymoon:
ThermoHydroMechanical process with homogeneous test,  bimaterial test and point heat source consolidation test.

PR #1937, opened 2017-08-29
Restructure ThermoMechanics

endJunction:
Same as #1877 and #1855 to keep the processes similar.

TODO:

- [x] Remove `<dimension>` from the project files.


wenqing:
THM of PR #1930  has the same structure as that of ThermoMechanics in the master branch.

endJunction:
@wenqing Yes, I'll have to update THM too after it is merged. Same goes for phase field process.

PR #1939, opened 2017-08-30
Vector extrapolation part2

wenqing:
Continue on Christoph's implementation of vectorial extrapolation, this PR changes the velocity calculation in several other classes of processes. The changes are
1. CL:     CompontTransport, Hydromechnics, Richards.
2. WW:  Coupled term, LiquidFlow.

### DIFF
--- a/ProcessLib/ComponentTransport/ComponentTransportFEM.h
+++ b/ProcessLib/ComponentTransport/ComponentTransportFEM.h
@@ -285,6 +285,8 @@ public:
             auto const& N = ip_data.N;
             auto const& dNdx = ip_data.dNdx;
 
+            pos.setIntegrationPoint(ip);
+
             auto const& K =
                 _process_data.porous_media_properties.getIntrinsicPermeability(
                     t, pos);
--- a/ProcessLib/HydroMechanics/HydroMechanicsFEM.h
+++ b/ProcessLib/HydroMechanics/HydroMechanicsFEM.h
@@ -499,7 +499,7 @@ public:
         auto const local_x = current_solution.get(indices);
 
         cache.clear();
-        auto cache_vec = MathLib::createZeroedMatrix<Eigen::Matrix<
+        auto cache_matrix = MathLib::createZeroedMatrix<Eigen::Matrix<
             double, DisplacementDim, Eigen::Dynamic, Eigen::RowMajor>>(
             cache, DisplacementDim, num_intpts);
 
@@ -510,8 +510,6 @@ public:
             Eigen::Map<typename ShapeMatricesTypePressure::template VectorType<
                 pressure_size> const>(local_x.data() + pressure_index,
                                       pressure_size);
-        using GlobalDimVectorType =
-            typename ShapeMatricesTypePressure::GlobalDimVectorType;
 
         unsigned const n_integration_points =
             _integration_method.getNumberOfPoints();
@@ -529,7 +527,7 @@ public:
 
             // Compute the velocity
             auto const& dNdx_p = _ip_data[ip].dNdx_p;
-            cache_vec.col(ip).noalias() =
+            cache_matrix.col(ip).noalias() =
                 -K_over_mu * dNdx_p * p - K_over_mu * rho_fr * b;
         }
 
--- a/ProcessLib/HydroMechanics/HydroMechanicsFEM.h
+++ b/ProcessLib/HydroMechanics/HydroMechanicsFEM.h
@@ -15,6 +15,7 @@
 #include "MaterialLib/SolidModels/KelvinVector.h"
 #include "MaterialLib/SolidModels/LinearElasticIsotropic.h"
 #include "MathLib/LinAlg/Eigen/EigenMapTools.h"
+#include "NumLib/DOF/DOFTableUtil.h"
 #include "NumLib/Fem/ShapeMatrixPolicy.h"
 #include "ProcessLib/Deformation/BMatrixPolicy.h"
 #include "ProcessLib/Deformation/LinearBMatrix.h"
@@ -364,43 +365,6 @@ public:
         }
     }
 
-    void postTimestepConcrete(std::vector<double> const& local_x) override
-    {
-        double const& t = _process_data.t;
-
-        auto p =
-            Eigen::Map<typename ShapeMatricesTypePressure::template VectorType<
-                pressure_size> const>(local_x.data() + pressure_index,
-                                      pressure_size);
-        using GlobalDimVectorType =
-            typename ShapeMatricesTypePressure::GlobalDimVectorType;
-
-        unsigned const n_integration_points =
-            _integration_method.getNumberOfPoints();
-
-        SpatialPosition x_position;
-        x_position.setElementID(_element.getID());
-        for (unsigned ip = 0; ip < n_integration_points; ip++)
-        {
-            x_position.setIntegrationPoint(ip);
-            double const K_over_mu =
-                _process_data.intrinsic_permeability(t, x_position)[0] /
-                _process_data.fluid_viscosity(t, x_position)[0];
-
-            auto const rho_fr = _process_data.fluid_density(t, x_position)[0];
-            auto const& b = _process_data.specific_body_force;
-
-            // Compute the velocity
-            auto const& dNdx_p = _ip_data[ip].dNdx_p;
-            GlobalDimVectorType const darcy_velocity =
-                -K_over_mu * dNdx_p * p - K_over_mu * rho_fr * b;
-            for (unsigned d = 0; d < DisplacementDim; ++d)
-            {
-                _darcy_velocities[d][ip] = darcy_velocity[d];
-            }
-        }
-    }
-
     Eigen::Map<const Eigen::RowVectorXd> getShapeMatrix(
         const unsigned integration_point) const override
     {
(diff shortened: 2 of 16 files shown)


ProcessLib/LiquidFlow/LiquidFlowLocalAssembler-impl.h
```diff
@@ -241,37 +243,73 @@ void LiquidFlowLocalAssembler<ShapeFunction, IntegrationMethod, GlobalDim>::
...
-    computeSecondaryVariableConcrete(double const t,
-                                     std::vector<double> const& local_x)
+std::vector<double> const&
+LiquidFlowLocalAssembler<ShapeFunction, IntegrationMethod, GlobalDim>::
+    getIntPtDarcyVelocity(const double t,
+                          GlobalVector const& current_solution,
+                          NumLib::LocalToGlobalIndexMap const& dof_table,
+                          std::vector<double>& veloctiy_cache) const
 {
+    // auto const num_nodes = ShapeFunction_::NPOINTS;
```
endJunction: remove if not needed.

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler-impl.h
```diff
@@ -241,37 +243,73 @@ void LiquidFlowLocalAssembler<ShapeFunction, IntegrationMethod, GlobalDim>::
...
+    veloctiy_cache.clear();
+    auto veloctiy_cache_vectors = MathLib::createZeroedMatrix<
+        Eigen::Matrix<double, GlobalDim, Eigen::Dynamic, Eigen::RowMajor>>(
+        veloctiy_cache, GlobalDim, num_intpts);
+
     SpatialPosition pos;
     pos.setElementID(_element.getID());
     const int material_id = _material_properties.getMaterialID(pos);
-    const Eigen::MatrixXd& permeability = _material_properties.getPermeability(
+    const Eigen::MatrixXd& perm = _material_properties.getPermeability(
```
endJunction: minor: permeability was a good name...

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler.h
```diff
@@ -39,23 +42,27 @@ class LiquidFlowLocalAssemblerInterface
...
 
-    virtual std::vector<double> const& getIntPtDarcyVelocityZ(
+    virtual std::vector<double> const& getIntPtDarcyVelocity(
         const double /*t*/,
         GlobalVector const& /*current_solution*/,
         NumLib::LocalToGlobalIndexMap const& /*dof_table*/,
         std::vector<double>& /*cache*/) const = 0;
+
+protected:
+    /// Pointer that gets from Process class
```
endJunction: Maybe "Pointer that is set from a Process class" would be better....

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler.h
```diff
@@ -72,6 +79,9 @@ class LiquidFlowLocalAssembler : public LiquidFlowLocalAssemblerInterface
     using NodalVectorType = typename LocalAssemblerTraits::LocalVector;
     using GlobalDimVectorType = typename ShapeMatricesType::GlobalDimVectorType;
 
+    using LocalMatrixType = Eigen::Map<
+        Eigen::Matrix<double, GlobalDim, Eigen::Dynamic, Eigen::RowMajor>>;
```
endJunction: Just for better understanding for discussion later:
 - Why is one dimension dynamic?
 - Why is it a map and not simply a matrix?

ProcessLib/LiquidFlow/LiquidFlowProcess.cpp
```diff
@@ -66,35 +66,20 @@ void LiquidFlowProcess::initializeConcreteProcess(
         pv.getShapeFunctionOrder(), _local_assemblers,
         mesh.isAxiallySymmetric(), integration_order, _gravitational_axis_id,
         _gravitational_acceleration, _reference_temperature,
-        *_material_properties);
+        *_material_properties, this->_coupling_term);
```
endJunction: Not sure, but is `this` required in this context?

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler-impl.h
```diff
@@ -397,22 +426,17 @@ template <typename ShapeFunction, typename IntegrationMethod,
           unsigned GlobalDim>
 void LiquidFlowLocalAssembler<ShapeFunction, IntegrationMethod, GlobalDim>::
     IsotropicCalculator::calculateVelocity(
-        std::vector<std::vector<double>>& darcy_velocities,
+        unsigned ip, LocalMatrixType& darcy_velocity_at_ips,
```
endJunction: - const ip.
 - -Would be nice if the output arguments were either in the beginning or the end of the argument list.

ProcessLib/RichardsFlow/RichardsFlowFEM.h
```diff
@@ -240,9 +226,44 @@ class LocalAssemblerData : public RichardsFlowLocalAssemblerInterface
...
+        return _saturation;
+    }
+
+    std::vector<double> const& getIntPtDarcyVelocity(
+        const double t,
+        GlobalVector const& current_solution,
+        NumLib::LocalToGlobalIndexMap const& dof_table,
+        std::vector<double>& cache) const override
+    {
+        // auto const num_nodes = ShapeFunction_::NPOINTS;
```
endJunction: remove if not needed

ProcessLib/UncoupledProcessesTimeLoop.cpp
```diff
@@ -828,9 +854,23 @@ bool UncoupledProcessesTimeLoop::loop()
         {
             auto& pcs = spd->process;
             auto const& x = *_process_solutions[pcs_idx];
-            _output->doOutputLastTimestep(pcs, spd->process_output,
-                                          accepted_steps + rejected_steps, t,
-                                          x);
+
+            if (is_staggered_coupling)
```
endJunction: This if-then-else looks almost identical to an above version with dt=0. Could be this deduplicated?

ProcessLib/HydroMechanics/HydroMechanicsFEM.h
```diff
@@ -522,34 +486,52 @@ class HydroMechanicsLocalAssembler : public LocalAssemblerInterface
...
-    std::vector<double> const& getIntPtDarcyVelocityZ(
-        const double /*t*/,
-        GlobalVector const& /*current_solution*/,
-        NumLib::LocalToGlobalIndexMap const& /*dof_table*/,
-        std::vector<double>& /*cache*/) const override
-    {
-        assert(_darcy_velocities.size() > 2);
-        return _darcy_velocities[2];
+        cache.clear();
+        auto cache_vec = MathLib::createZeroedMatrix<Eigen::Matrix<
```
chleh: Minor: I'd prefer `cache_mat` since it's a matrix. Probably I introduced it wrongly.

chleh: This is HM. @wenqing just told me that it's going to be split to a separate PR.

ProcessLib/ComponentTransport/ComponentTransportFEM.h
```diff
@@ -285,6 +285,8 @@ class LocalAssemblerData : public ComponentTransportLocalAssemblerInterface
             auto const& N = ip_data.N;
             auto const& dNdx = ip_data.dNdx;
 
+            pos.setIntegrationPoint(ip);
```
chleh: Good observation!

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler-impl.h
```diff
@@ -241,37 +243,73 @@ void LiquidFlowLocalAssembler<ShapeFunction, IntegrationMethod, GlobalDim>::
...
-                                     std::vector<double> const& local_x)
+std::vector<double> const&
+LiquidFlowLocalAssembler<ShapeFunction, IntegrationMethod, GlobalDim>::
+    getIntPtDarcyVelocity(const double t,
+                          GlobalVector const& current_solution,
+                          NumLib::LocalToGlobalIndexMap const& dof_table,
+                          std::vector<double>& veloctiy_cache) const
 {
+    // auto const num_nodes = ShapeFunction_::NPOINTS;
+    auto const num_intpts = _shape_matrices.size();
```
chleh: Minor: If there is `_integration_method`, then this could be `_integration_method.getNumberOfPoints();`.

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler.h
chleh: As far as I can see this is used for storing a (D x #integration_points) matrix for providing the "input data" of the velocity extrapolation. Therefore:
* Map because the data is stored in a std::vector
* Dynamic: Because of dynamic #int_pts

However, maybe it should be named `VelocityAtIntegrationPointsMatrix` or similar.

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler.h
```diff
@@ -81,8 +91,10 @@ class LiquidFlowLocalAssembler : public LiquidFlowLocalAssemblerInterface
         int const gravitational_axis_id,
         double const gravitational_acceleration,
         double const reference_temperature,
-        LiquidFlowMaterialProperties const& material_propertries)
-        : _element(element),
+        LiquidFlowMaterialProperties const& material_propertries,
+        StaggeredCouplingTerm* coupling_term)
```
chleh: It's unfortunate that now we need to store the coupling term in the local assemblers.
I believe that one can circumvent this; it's not hard, but certainly not trivial.
@wenqing if you want to change this now and if you see the solution without much thought, please make the change. Otherwise, could you please add a `// TODO remove additional member` to the source code? Then I can change it in the future.

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler-impl.h
wenqing: Removed.

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler-impl.h
wenqing: Changed.

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler.h
wenqing: Changed accordingly.

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler.h
wenqing: @chleh Thanks for explanation. The name is changed to MatrixOfVelocityAtIntegrationPoints.

ProcessLib/LiquidFlow/LiquidFlowProcess.cpp
wenqing: Removed.

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler-impl.h
wenqing: Changed accordingly.

ProcessLib/RichardsFlow/RichardsFlowFEM.h
wenqing: removed.

ProcessLib/UncoupledProcessesTimeLoop.cpp
wenqing: Added a generic member function to UncoupledProcessesTimeLoop to handle the ouput, with which all related code duplications is gone. Please review it in the commit entitled with ``[Output] Added a generic output member to UncoupledProcessesTimeLoop``

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler-impl.h
wenqing: Changed.

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler.h
wenqing: Yes a pointer type is added to all assembler instances. 
I prefer to move the integration data out of the local assembler and make the  local assembler unique for each process. By this way, the repeated computation and memory allocation for  integration data among different process, which use the same element order, can be avoid and consequently this coulping term pointer is only assigned once for each process.

I added a TODO comment  before the declaration of _coupling_term with the meanings of the above words.

wenqing:
The vectorial extrapolation for HM, which requires element order switch,  will be presented in next PR

ProcessLib/ComponentTransport/ComponentTransportFEM.h
TomFischer: Maybe extraction to separate PR?

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler.h
```diff
@@ -39,23 +42,30 @@ class LiquidFlowLocalAssemblerInterface
...
-    virtual std::vector<double> const& getIntPtDarcyVelocityY(
-        const double /*t*/,
-        GlobalVector const& /*current_solution*/,
-        NumLib::LocalToGlobalIndexMap const& /*dof_table*/,
-        std::vector<double>& /*cache*/) const = 0;
+    void setStaggeredCouplingTerm(std::size_t const /*mesh_item_id*/,
+                                  StaggeredCouplingTerm* const coupling_term)
+    {
+        _coupling_term = coupling_term;
+    }
```
endJunction: discussion point: There are no special checks or other procedures of the coupling term, so the `_coupling_term` could be made public saving the "set" method.

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler.h
endJunction: Agree to the fact that coupling term being in local assemblers. We shall see how it develops.

ProcessLib/ComponentTransport/ComponentTransportFEM.h
wenqing: After discussed with Tom, we keep it as it is in this PR.

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler.h
wenqing: Just discussed and keep it as it is.

ProcessLib/HydroMechanics/HydroMechanicsFEM.h
wenqing: @chleh Found the problem in the HM part. The bug is not caused by the order of shape function but by setting a wrong number of the components in  ``makeExtrapolator`` for velocity. With the bug fixing, the HM part is kept in this PR, and I am not going to bloat the source code with hundreds line change any more :smiley: . 

BTW, cache_vec is changed to cache_matrix as well.

PR #1943, opened 2017-09-07
Various fixes from #1907

bilke:
To cleanup #1907 a little bit...


.gitmodules
```diff
@@ -16,6 +16,7 @@
 [submodule "Tests/Data"]
 	path = Tests/Data
 	url = https://github.com/ufz/ogs-data
+	shallow = true
```
chleh: Although this PR is already merged: I understand the rationale behind setting `shallow = true`, but I assume that this can cause some problems, e.g., if a user just pushed something to a feature branch in ogs/data, then the shallow clone (with depth of one) of the submodule won't fetch the latest master commit, which will lead to nasty error messages at least. Or am I wrong?

bilke: I guess you are right..

endJunction: I don't think it will lead to problems because cmake scripts will fetch the missing data with depth=1.

chleh: Then I'm fine with it.

PR #1944, opened 2017-09-07
Restructure PhaseField

endJunction:
Same as #1877 and #1855 and https://github.com/ufz/ogs/pull/1937  to keep the processes similar.

### DIFF
--- a/ProcessLib/PhaseField/LocalAssemblerInterface.h
+++ b/ProcessLib/PhaseField/LocalAssemblerInterface.h
@@ -14,8 +14,6 @@
 #include "NumLib/Extrapolation/ExtrapolatableElement.h"
 #include "ProcessLib/LocalAssemblerInterface.h"
 
-#include "PhaseFieldProcessData.h"
-
 namespace ProcessLib
 {
 namespace PhaseField
--- a/ProcessLib/PhaseField/LocalAssemblerInterface.h
+++ b/ProcessLib/PhaseField/LocalAssemblerInterface.h
@@ -1,0 +1,102 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2017, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#pragma once
+
+#include <vector>
+
+#include "NumLib/Extrapolation/ExtrapolatableElement.h"
+#include "ProcessLib/LocalAssemblerInterface.h"
+
+#include "PhaseFieldProcessData.h"
+
+namespace ProcessLib
+{
+namespace PhaseField
+{
+
+struct PhaseFieldLocalAssemblerInterface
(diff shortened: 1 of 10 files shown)


ProcessLib/PhaseField/PhaseFieldFEM.h
```diff
@@ -419,20 +338,22 @@ class PhaseFieldLocalAssembler : public PhaseFieldLocalAssemblerInterface
...
+                                                              phasefield_index)
             .noalias() += Kdd + Ddd / dt;
 
         // phasefield equation, displacement part.
-        local_Jac.template block<phasefield_size, displacement_size>(
-                     phasefield_index, displacement_index)
+        local_Jac
+            .template block<phasefield_size, displacement_size>(
+                phasefield_index, displacement_index)
             .noalias() += Kdu;
```
endJunction: @nagelt Above in line 317, in the if-condition (history_prev < strain_energy), the value of `Kdu` is set depending on the last integration point values for the history_prev and strain_energy. Is it correct? Seems to me like the results would depend on the ordering of the integration points.

ProcessLib/PhaseField/LocalAssemblerInterface.h
```diff
@@ -0,0 +1,100 @@
...
+namespace ProcessLib
+{
+namespace PhaseField
+{
+
+struct PhaseFieldLocalAssemblerInterface
+    : public ProcessLib::LocalAssemblerInterface,
+      public NumLib::ExtrapolatableElement
+{
+    virtual std::vector<double> const& getIntPtSigmaXX(
```
chleh: I suggest to also switch to vectorial output in this PR.

endJunction: There will be a separate PR switching mechanics processes to vectorial output.

chleh: That's also OK.

ProcessLib/PhaseField/PhaseFieldFEM.h
xingyuanmiao: both the history_variable_prev and strain_energy_tensile are integration point data. The if-condition solves Kdu on each integration point.

nagelt: @xingyuanmiao not quite ... the if-condition determines whether or not the term is relevant in each integration point .. that's ok ... but the assigment you make then takes the entire K_ud transpose which has been assembled for all previous integration points (+=) ... so after your if you add more than you need ... and this leads to the randomness Dima describes ...

endJunction: I made the corresponding change in this PR. See commit `[PL] PhaseField: Fix computation of Kdu.`

endJunction:
@bilke In Jenkins one of the builds could not connect to a docker; appveyor shows same behaviour with the conan's remote ogs. Could you please have a look?

bilke:
@endJunction Conan Boost issue should be resolved: Conan cache directory was not writable by Jenkins on the new `azure-docker1`. I restarted the job..

PR #1950, opened 2017-09-18
Ccache versions

bilke:


bilke:
14.04. (used on Travis) ships with 3.1.x

chleh:
@bilke thanks for the explanation.

PR #1951, opened 2017-09-22
[CMake] Fixed shared libs installation on Win and Linux.

bilke:
Reported by @chleh.

Also:

- Fixed web
- Fixed appveyor
- Fixed VS out of heap mem (using 64-bit compiler by default now)


scripts/cmake/conan/conan.cmake
```diff
@@ -81,6 +81,11 @@ function(conan_cmake_settings result)
         list(GET VERSION_LIST 1 MINOR)
         conan_cmake_detect_gnu_libcxx(_LIBCXX)
         set(_SETTINGS ${_SETTINGS} -s compiler=gcc -s compiler.version=${MAJOR}.${MINOR} -s compiler.libcxx=${_LIBCXX})
+    elseif (${CMAKE_CXX_COMPILER_ID} STREQUAL Intel)
```
chleh: Why? Is it, because conan does not provide pre-built libs for icc?

scripts/cmake/packaging/GetDependencies.cmake
```diff
@@ -11,6 +11,12 @@ foreach(TARGET ${INSTALL_DEPENDENCIES})
     endif()
 
     if(EXISTS ${TARGET_EXE})
+        # Run CMake again after target was built to collect dependencies
```
chleh: So, that collects the shared lib dependencies?

Applications/CLI/CMakeLists.txt
```diff
@@ -34,3 +34,8 @@ cpack_add_component(ogs_cli
     DESCRIPTION "The command line interface for OpenGeoSys."
     GROUP Applications
 )
+
+if(NOT APPLE)
```
chleh: Does it fail on Mac or is it unnecessary on Mac? (Maybe add a comment in the cmake file?)

scripts/cmake/conan/conan.cmake
bilke: Ups sorry wrong branch!

bilke:
@endJunction There is something wrong on Jenkins.. (in the web-stage and conan-job) Will fix it next week..

scripts/cmake/packaging/GetDependencies.cmake
bilke: This runs CMake again with the (temporary) option `PRE_INSTALL_RUN=ON` which queries the already built executables for its dependencies.

chleh:
@bilke Were all the changes in this PR really necessary to solve the shared libs issue? The final size was 2700 lines of changes!
Apparently there were some web-related changes, too, and there is a huge yarn.lock file, and one changed line reads
> title = "Saturated Variable-Density Flow and Mass Transport (Elder)"

Probably everything is alright. I'm only a bit concerned that something wrong sneaked in.

endJunction:
The actual change is much smaller, only the web/yarn.lock (which is autogenerated/updated) is 2,668 lines.

PR #1953, opened 2017-09-26
LIE compression

endJunction:
Prevents negative apertures

Review can be done commit-wise.

TODO:

- [x] Include provided documentation into web/content
- [x] Add the two new projects (negative aperture and displacement controlled) to ctests.

### DIFF
--- a/MaterialLib/FractureModels/FractureModelBase.h
+++ b/MaterialLib/FractureModels/FractureModelBase.h
@@ -66,6 +66,7 @@ public:
      *
      * @param t           current time
      * @param x           current position in space
+     * @param aperture0   initial fracture's aperture
      * @param w_prev      fracture displacement at previous time step
      * @param w           fracture displacement at current time step
      * @param sigma_prev  stress at previous time step
@@ -76,6 +77,7 @@ public:
     virtual void computeConstitutiveRelation(
             double const t,
             ProcessLib::SpatialPosition const& x,
+            double const aperture0,
             Eigen::Ref<Eigen::VectorXd const> w_prev,
             Eigen::Ref<Eigen::VectorXd const> w,
             Eigen::Ref<Eigen::VectorXd const> sigma_prev,
--- a/MaterialLib/FractureModels/FractureModelBase.h
+++ b/MaterialLib/FractureModels/FractureModelBase.h
@@ -67,6 +67,7 @@ public:
      * @param t           current time
      * @param x           current position in space
      * @param aperture0   initial fracture's aperture
+     * @param sigma0      initial stress
      * @param w_prev      fracture displacement at previous time step
      * @param w           fracture displacement at current time step
      * @param sigma_prev  stress at previous time step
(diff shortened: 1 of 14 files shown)


nagelt:
@endJunction Documentation here ...
[doku_lie_comp.pdf](https://github.com/ufz/ogs/files/1334650/doku_lie_comp.pdf)

nagelt:
@endJunction You can use our test. Note that the cutoff should be negative (see last Eq. in docu). Rationale is that our test would lead to negative apertures but doesn't with this implementation.

nagelt:
@renchao-lu could you have a look at this PR, please?

renchao-lu:
@nagelt @endJunction It's my pleasure. Also, it's a good exercise for me to get familiar with this branch.

MaterialLib/FractureModels/FractureModelBase.h
```diff
@@ -66,6 +66,7 @@ class FractureModelBase
      *
      * @param t           current time
      * @param x           current position in space
+     * @param aperture0   initial fracture's aperture
```
renchao-lu: I would name this variable as aperture_bm0 instead which physically means the initial mechanical aperture at a certain node or initial mean mechanical aperture at a representative element, thereby leaving some naming space for other relevant variables, for instance, aperture_bh0 standing for initial hydraulic aperture.

renchao-lu: @endJunction @nagelt The variable "aperture0" introduced here keeps consistent with the pre-existing definition in the header file "FractureProperty.h". Therefore, just leave my previous immature thinking out.

MaterialLib/FractureModels/LogPenalty.h
```diff
@@ -0,0 +1,69 @@
...
+/// A penalty function for negative aperture supression used as a multiplier to
+/// the normal fracture stiffness.
+///
+/// The derivative is continuous at aperture = aperture0 and aperture =
+/// aperture_cutoff.
+inline double logPenaltyDerivative(double const aperture0,
+                                   double const aperture,
+                                   double const aperture_cutoff)
+{
+    if (aperture >= aperture0)
```
wenqing: Is it good to compare two real numbers with  ``>=``?

MaterialLib/FractureModels/LogPenalty.h
```diff
@@ -0,0 +1,69 @@
...
+               2 * penalty / aperture_cutoff *
+                   (2 * aperture - aperture_cutoff - aperture0);
+    }
+};
+
+inline double logPenalty(double const aperture0,
+                         double const aperture,
+                         double const aperture_cutoff)
+{
+    if (aperture >= aperture0)
```
wenqing: the same here.

MaterialLib/FractureModels/LogPenalty.h
endJunction: I don't see a problem; Comparing with `if (!(aperture < aperuture0))` yields the same assembly using the [ucomisd](http://www.felixcloutier.com/x86/UCOMISD.html) instruction.

At the boundary `aperture == aperture0` the functions result is the same, but the computation (`return 1`) is way easier.

wenqing: I see. Thanks for the explanation.

endJunction:
@renchao-lu Can this PR be merged (after the tests pass, of course) or do you have some additional comments?

renchao-lu:
Thanks for your kindly asking me suggestions. I have no further comment. Please feel free.

PR #1960, opened 2017-10-12
Fix bugs in ResetPropertiesInPolygonalRegion tool

TomFischer:


waltherm:
Thanks for fixing this, Tom. 
Just as a general comment for discussion: shouldn't these tools also get some testing to omit loss of functionality?

rinkk:
:+1:

Applications/Utils/MeshEdit/ResetPropertiesInPolygonalRegion.cpp
```diff
@@ -192,54 +187,31 @@ int main (int argc, char* argv[])
...
-        resetMeshElementProperty(*mesh, polygon, property_name, int_property_val);
+        resetMeshElementProperty(*mesh, polygon, property_name,
+                                 int_property_arg.getValue());
     }
 
     if (bool_property_arg.isSet()) {
         bool bool_property_val(bool_property_arg.getValue());
-        resetMeshElementProperty(*mesh, polygon, property_name, bool_property_val);
+        resetMeshElementProperty(*mesh, polygon, property_name,
+                                 bool_property_arg.getValue());
```
endJunction: Not sure here, but it looks like a bug to me:
I just stumbled on a compiler warning "The `bool_property_val` is not used." Maybe the second parameter in the function call should be `bool_property_val.getValue()` not `..._arg.getValue()`?

PR #1963, opened 2017-10-18
Fixed mem issue with raster mapping

rinkk:


MeshGeoToolsLib/GeoMapper.cpp
```diff
@@ -58,7 +57,7 @@ void GeoMapper::mapOnDEM(GeoLib::Raster *const raster)
         ERR("Geometry \"%s\" does not exist.", _geo_name.c_str());
         return;
     }
-    _raster = raster;
+    _raster.reset(raster);
```
endJunction: On the caller side you could move the raster and then use assignment here.
```c++
geo_mapper.mapOnDEM(std::move(raster));
```
The `mapOnDEM` interface changes to
```c++
void GeoMapper::mapOnDEM(std::unique_ptr<GeoLib::Raster> raster);
```

The `get()` and the `release()` calls are usually needed to the legacy code interfaces or low-level memory management (like memory pool).

PR #1967, opened 2017-10-19
Small GMSH improvements.

endJunction:
Two comments by Tom, using unique_ptr in two occasions, better function parameter names.


endJunction:
Jenkins fails only because of the doxygen, not related to this PR. Doxygen is fixed in #1968.

PR #1970, opened 2017-10-26
New coupling assembly computations

### LINKED ISSUE #1814: LiquidFlow <-> HeatConduction circular dependency
Both processes reference each other, which is not good. It should be possible in my opinion to compile the liquid flow process w/o it knowing about any other coupled process.

LiquidFlow references HeatConductionProcess in [LiquidFlowLocalAsm<>::assembleWithCoupledTerm()](https://github.com/ufz/ogs/blob/master/ProcessLib/LiquidFlow/LiquidFlowLocalAssembler-impl.h#L78)

HeatConduction references LiquidFlow in [...::assembleWithCoupledTerm()](https://github.com/ufz/ogs/blob/master/ProcessLib/HeatConduction/HeatConductionFEM-impl.h#L118)

wenqing:
This implementation changed the coupling assembly computations from that are performed  across the different Process classes to that is done within  a single Process class, which is for monolithic scheme originally. With the changes, the original Process class that for  monolithic scheme originally can now handle both of the monolithic and staggered schemes. 

With the new implementation,
1. coupled process class reference is removed from the coupled term class, which gets a new name of ``CoupledSolutionsForStaggeredScheme``, this   fixes #1814
2. material properties are shared by the coupled processes within one single instance of class Process. Considered that the global solutions have already being held by class UncoupledProcessesTimeLoop, there is no need to introduce a ``global state space`` for material properties or global solutions. Because now, the material properties are encapsulated into the proper process classes whichever the monolithic scheme or the staggered scheme being applied., and the solutions are encapsulated into the family of classes of the equations system. 
3. only one DOF table  and one shape function cache are created for the staggered decoupling if the modelling of the coupled processes employs the same order element.
4. only one vtu file at the required time is created for the staggered decoupling computation  instead of multiple vtu file output  for all coupled processes.   

The source code change is a little bit big (more than 2k lines). However, the changes are easy to follow.
For reviewing, the main changes are in:
1. coupled term classes.
2. class Process, ProcessVariable, ProcessOutput, and the classes in boundary conditions.
3.  project file for staggered scheme. (One can modify the project file for the monolithic scheme by just using two process definition in  the element of <time_loop>).
4. the old coupled assemblers in classes of LiquidFlow and HeatCondution are gone.  LiquidFlow and HeatCondution are now for pure single process.

What is untouched is the global framework of the staggered scheme, which is in class UncoupledProcessesTimeLoop.

As a first application of this decoupling to the coupled process,  the assembly classes of HT have been modified as
1. the local assembler is re-designed into two assemblers as  StaggeredHTFEM and MonolithicHTFEM, which are derived from a base class HTFEM.
2. some material properties are changed. for example,  the input of the thermal_dispersivity is set as optional, the fluid thermal expansion computation is newly added as an optional term. The reference fluid density is dropped. In the previous code, the so called reference fluid density is used in the computation of  the heat capacity and  the thermal_dispersivity related term. This is not corrected physically because the  current fluid density value must be used for computation. I guess that  the use of the reference fluid density is just to make the benchmark results being the same as what are obtained by other software.

The benchmark in HT is used to test this implementation without changing the time step sizes. Obviously, with the same time stepping, the results of the test by the staggered scheme has less accuracy than that by the monolithic scheme. See:
1. contour plot
contour_comparison_ht_staggeredmonolithicscheme
2. profile along the vertical line in the middle of the domain.
profile_comparison_ht_staggeredmonolithicscheme

The benchmarks about HT coupling in LiquidProcess will be immigrated to this implementation  in the next PR.

The next changes are HydroMechanics and ThermoMechanics.


wenqing:
Just committed new changes, which are mainly about the  the BCs, sources terms (STs) and the process variables (PV) in the coupling scheme.
Compared to the initial implementation of the BCs, STs and PVs for coupling in this PR, the BCs, STs and PVs are now categorized by process in class Process as
```
    std::vector<std::vector<std::reference_wrapper<ProcessVariable>>>
        _process_variables;
    std::vector<BoundaryConditionCollection> _boundary_conditions;
    std::vector<std::vector<std::unique_ptr<NodalSourceTerm>>> _source_terms;
``` 
with which, the BCs, STs and  PVs for the staggered coupling could be more flexible. For example THM could be T-H-M, TH-M, or T-HM.

ProcessLib/ComponentTransport/CreateComponentTransportProcess.cpp
```diff
@@ -35,19 +35,43 @@ std::unique_ptr<Process> createComponentTransportProcess(
...
+
+        for (int i = 0; i < 2; i++)
         {
-        //! \ogs_file_param_special{prj__processes__process__ComponentTransport__process_variables__concentration}
-        "concentration",
-        //! \ogs_file_param_special{prj__processes__process__ComponentTransport__process_variables__pressure}
-        "pressure"});
+            auto per_process_variables =
+                findProcessVariables(variables, pv_config, {variable_names[i]});
+            process_variables.push_back(std::move(per_process_variables));
```
endJunction: The for loop could be a for-rage, because `i` is not used but to access the variable_names.
This should work w/o variable_names array:
```c++
using std::string_literals;
for (auto const& variable_name : {"concentration"s, "pressure"s}
{
   // ...
}

File: ProcessLib/CoupledSolutionsForStaggeredScheme.cpp [hunk 2]
```diff
@@ -18,48 +18,46 @@
...
-    for (auto const& coupled_process_pair : coupled_processes)
+    for (std::size_t i = 0; i < number_of_coupled_solutions; i++)
     {
-        auto const& coupled_pcs = coupled_process_pair.second;
-        auto const prevous_time_x = coupled_pcs.getPreviousTimeStepSolution();
-        if (prevous_time_x)
-        {
-            MathLib::LinAlg::setLocalAccessibleVector(*prevous_time_x);
-        }
+        auto const& x_t0 = *cpl_xs.coupled_xs_t0[i];
```
endJunction: for-range-loop again; I'm not sure about the type of '(*cpl_xs.coupled_xs_t0)' but if it has `begin` and `end`, then:
```c++
for (auto const& x_t0 : *cpl_xs.coupled_xs_t0)
{
   local_xs_t0...
}

File: ProcessLib/CoupledSolutionsForStaggeredScheme.cpp [hunk 3]
```diff
@@ -18,48 +18,46 @@
...
-    // Get local nodal solutions of the coupled equations.
-    for (auto const& global_coupled_x_pair : global_coupled_xs)
+    for (std::size_t i = 0; i < number_of_coupled_solutions; i++)
     {
-        auto const& coupled_x = global_coupled_x_pair.second;
-        auto const local_coupled_x = coupled_x.get(indices);
-        BaseLib::insertIfTypeIndexKeyUniqueElseError(
-            local_coupled_xs, global_coupled_x_pair.first, local_coupled_x,
-            "local_coupled_x");
+        auto const& x_t1 = cpl_xs.coupled_xs[i].get();
```
endJunction: dto.

ProcessLib/HT/CreateHTProcess.cpp
```diff
@@ -36,17 +37,42 @@ std::unique_ptr<Process> createHTProcess(
 
     DBUG("Create HTProcess.");
 
+    auto const staggered_scheme =
+        //! \ogs_file_param{prj__processes__process__HT__coupling_scheme}
+        config.getConfigParameterOptional<std::string>("coupling_scheme");
+    const bool use_monolithic_scheme =
+        (staggered_scheme && (*staggered_scheme == "staggered")) ? false : true;
```
endJunction: The `&&` and `==` already return a boolean; no need to have the triple operator:
```c++
const bool smth = (a == b) && !c;
```

ProcessLib/HT/HTFEM.h
```diff
@@ -116,178 +79,82 @@ class LocalAssemblerData : public HTLocalAssemblerInterface
...
-            auto const thermal_dispersivity_longitudinal =
-                _process_data.thermal_dispersivity_longitudinal(t, pos)[0];
-            auto const thermal_dispersivity_transversal =
-                _process_data.thermal_dispersivity_transversal(t, pos)[0];
+    double getHeatEnergyCoefficient(const double t, const SpatialPosition& pos,
+                                    const double porosity,
+                                    const double fluid_density,
+                                    const double specific_heat_capacity_fluid)
+    {
+        auto const& material_properties = this->_material_properties;
```
endJunction: Could you remind me please, why `this` is necessary here? I'd think it's not.

ProcessLib/HT/StaggeredHTFEM-impl.h
```diff
@@ -0,0 +1,286 @@
...
+    auto const& material_properties = this->_material_properties;
+
+    auto const& b = material_properties.specific_body_force;
+
+    MaterialLib::Fluid::FluidProperty::ArrayType vars;
+
+    unsigned const n_integration_points =
+        this->_integration_method.getNumberOfPoints();
+
+    for (std::size_t ip(0); ip < n_integration_points; ip++)
```
endJunction: No need to extend an unsigned to size_t here.

ProcessLib/HT/StaggeredHTFEM-impl.h
```diff
@@ -0,0 +1,286 @@
...
+            material_properties.fluid_properties->getValue(
+                MaterialLib::Fluid::FluidPropertyType::Density, vars);
+        auto const specific_heat_capacity_fluid =
+            material_properties.fluid_properties->getValue(
+                MaterialLib::Fluid::FluidPropertyType::HeatCapacity, vars);
+
+        // Assemble mass matrix
+        local_M.noalias() +=
+            w *
+            this->getHeatEnergyCoefficient(t, pos, porosity, fluid_density,
```
endJunction: Again, I'm not sure if `this` is required here...

ProcessLib/HT/StaggeredHTFEM.h
```diff
@@ -0,0 +1,92 @@
+/**
```
endJunction: Not to myself @endJunction, continue review here.

ProcessLib/Process.h
```diff
@@ -215,7 +215,7 @@ class Process
 
     VectorMatrixAssembler _global_assembler;
 
-    mutable bool _is_monolithic_scheme;
+    const bool _use_monolithic_scheme;
```
endJunction: This is much nicer than the mutable version!

ProcessLib/Process.cpp
```diff
@@ -58,56 +68,69 @@ void Process::initialize()
...
+        std::vector<std::unique_ptr<NodalSourceTerm>> per_process_source_terms;
+        for (std::size_t variable_id = 0;
+             variable_id < per_process_variables.size();
+             variable_id++)
+        {
+            per_process_BCs.addBCsForProcessVariables(
+                per_process_variables, *_local_to_global_index_map,
+                _integration_order);
+
+            ProcessVariable& pv = per_process_variables[variable_id];
```
endJunction: This is good candidate for the for-range loop
```c++
for (auto& pv : per_process_variables) { ... }
```

ProcessLib/Output.cpp
```diff
@@ -123,10 +123,40 @@ void Output::addProcess(ProcessLib::Process const& process, const unsigned pcs_i
         BaseLib::joinPaths(_output_directory, _output_file_prefix + "_pcs_" + std::to_string(pcs_idx) + ".pvd");
     _single_process_data.emplace(std::piecewise_construct,
                                  std::forward_as_tuple(&process),
-                                 std::forward_as_tuple(pcs_idx, filename));
+                                 std::forward_as_tuple(filename));
 }
 
+Output::SingleProcessData Output::findSingleProcessData(
+    Process const& process, const unsigned process_index) const
```
endJunction: process_id. Could be int.

ProcessLib/Process.cpp
```diff
@@ -24,17 +24,29 @@ Process::Process(
...
     : _mesh(mesh),
       _secondary_variables(std::move(secondary_variables)),
       _named_function_caller(std::move(named_function_caller)),
       _global_assembler(std::move(jacobian_assembler)),
+      _use_monolithic_scheme(use_monolithic_scheme),
       _coupled_solutions(nullptr),
       _integration_order(integration_order),
       _process_variables(std::move(process_variables)),
-      _boundary_conditions(parameters)
+      _boundary_conditions([&](const size_t number_of_processes)
```
endJunction: std::size_t

ProcessLib/Process.cpp
```diff
@@ -169,23 +201,46 @@ void Process::constructDofTable()
...
     {
+        // Assuming that all equations of the coupled process use the same
+        // element order. Other cases can be considered by overloading this
+        // member function in the derived class.
+
+        // Collect the mesh subsets in a vector.
         std::generate_n(
             std::back_inserter(all_mesh_subsets),
-            pv.getNumberOfComponents(),
+            _process_variables[0][0].get().getNumberOfComponents(),
```
endJunction: get() is not needed.

ProcessLib/ComponentTransport/CreateComponentTransportProcess.cpp
wenqing: Changed accordingly.

ProcessLib/CoupledSolutionsForStaggeredScheme.cpp
wenqing: Changed accordingly.

ProcessLib/CoupledSolutionsForStaggeredScheme.cpp
wenqing: Changed accordingly.

ProcessLib/HT/CreateHTProcess.cpp
wenqing: Changed by using clang-tidy

ProcessLib/HT/HTFEM.h
wenqing: Removed.

ProcessLib/HT/StaggeredHTFEM-impl.h
wenqing: Changed accordingly.

ProcessLib/HT/StaggeredHTFEM-impl.h
wenqing: It is needed in the this derived class.

ProcessLib/Process.cpp
wenqing: Took it.

ProcessLib/Output.cpp
wenqing: Changed.

ProcessLib/Process.cpp
wenqing: Added.

ProcessLib/Process.cpp
wenqing: `_process_variables[0][0]` has the type of  `std::reference_wrapper`. Removing `get()``results in a compilation error. Do you have any other suggestion?

endJunction: Oh, my mistake. I thought it would be a `unique_ptr`. All fine.

PR #1973, opened 2017-10-29
Remove old eps/sigma output

endJunction:
Replacing the reference vtu files with those containing vectorial sigma and epsilon fields using the script [convert.py](https://gist.github.com/endJunction/2b1da96c8cb8135652e607602bd82bb3).

Some fields were not tested before, so I took the current simulation results for reference files. There were few mistakes either in the project files or in the reference data, which I corrected on the way.

This is done only for SmallDeformation, because the conversion and the _verification_ take some rather long time (paraview on the old and new data and comparison).


endJunction:
@bilke Can you help me with the failing windows builds, please? I don't understand why the ctest call is not correct for that particular test. I'll come to you for discussion.

PR #1988, opened 2017-11-16
[Jenkins] Removed Win 32-bit builds.

bilke:
This will fix current master as current VTK Conan package is not available for 32-bit anymore.

This will also remove official Windows XP support...


endJunction:
Maybe there are few other places using/documenting 32-bit Windows builds---these should be updated then.
:fast_forward:

bilke:
Clarified 64-bit requiremen on the [devguide](https://docs.opengeosys.org/docs/devguide/getting-started/prerequisites). There are no 32-bit traces left ...

PR #1995, opened 2017-11-21
Fixed a compilation warning

wenqing:
as titled.


endJunction:
It would be better, I think, to update other code to use `int` instead.
What is the warning? (I'm asking because I don't get any...)

wenqing:
@endJunction 
due to 
``virtual unsigned getNumberOfComponents() const = 0;``
https://github.com/ufz/ogs/blob/master/ProcessLib/Parameter/Parameter.h#L70

endJunction:
OK, I'll have a look until Friday.

wenqing:
@endJunction Sorry, closed by mistake

PR #1996, opened 2017-11-21
Name change and new argument for a member function of process classes

wenqing:
1. Changed the names of structs,  variables that are related to the coupled terms.
2. Changed the name and the type of HTProcessData to HTMaterialProperties (std::unique_ptr).
3. Added an argument to a member function of classes of processes.
4. removed an unused include.

This PR is a part of PR #1970.


wenqing:
@endJunction @chleh In order to reduce the time of rebases for reviewers and developers, I decide to split PR #1970 into only two PRs. For this reason, I added three more commits to this PR.
@endJunction Please take a look at the three new commits.

PR #1997, opened 2017-11-22
Update getNumberOfComponents to return an int.

endJunction:
Closes #1995

### DIFF
--- a/Tests/NumLib/LocalToGlobalIndexMap.cpp
+++ b/Tests/NumLib/LocalToGlobalIndexMap.cpp
@@ -41,8 +41,8 @@ protected:
     std::unique_ptr<MeshLib::MeshSubset const> nodesSubset;
 
     //data component 0 and 1 are assigned to all nodes in the mesh
-    static std::size_t const comp0_id = 0;
-    static std::size_t const comp1_id = 1;
+    static int const comp0_id = 0;
+    static int const comp1_id = 1;
     std::vector<MeshLib::MeshSubsets> components;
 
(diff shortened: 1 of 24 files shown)


Tests/NumLib/LocalToGlobalIndexMap.cpp
```diff
@@ -163,7 +163,7 @@ TEST_F(NumLibLocalToGlobalIndexMapTest, DISABLED_MultipleVariablesMultipleCompon
         NumLib::ComponentOrder::BY_COMPONENT);
 
     ASSERT_EQ(30, dof_map->dofSizeWithGhosts());
-    ASSERT_EQ(3u, dof_map->getNumberOfComponents());
+    ASSERT_EQ(3, dof_map->getNumberOfComponents());
     ASSERT_EQ(2u, dof_map->getNumberOfVariables());
```
wenqing: dof_map->getNumberOfVariables() also returns int. Therefore, 2u --> 2.

endJunction: Thanks! I found few more places... Unfortunately I was not able to find/create an automatic substitution.

wenqing:
@endJunction New warning in `assert(component < _n_components);` in  PropertyVector.h because that the type of  _n_components is changed to `int`, e.g. in line
https://github.com/ufz/ogs/blob/master/MeshLib/PropertyVector.h#L71

endJunction:
Ah, I see. Unfortunately jenkins didn't show it. I'll fix it first thing in the morning.

@bilke We need a warning warning system ;)

PR #1999, opened 2017-11-24
Small fixes

endJunction:
Just few annoying warnings reported by gcc.

### DIFF
--- a/MeshLib/PropertyVector.h
+++ b/MeshLib/PropertyVector.h
@@ -66,7 +66,7 @@ public:
     }
 
     //! Returns the value for the given component stored in the given tuple.
-    PROP_VAL_TYPE& getComponent(std::size_t tuple_index, std::size_t component)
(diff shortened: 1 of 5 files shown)


wenqing:
Just rebased and compiled. There comes one new similar warning

```
ogs6/MeshLib/MeshGenerators/QuadraticMeshGenerator.cpp:32:23: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
     for (int i = 0; i < n_base_nodes; i++)
```
It is because that n_base_nodes is an unsigned variable getting from  `Element::getNumberOfBaseNodes()`.

By the way: class Element (Element.h) has several members that return ``unsigned ``.

endJunction:
Thanks @wenqing ; I just have seen it in the jenkins warnings. Will fix that one shortly.
I wouldn't touch the MeshLib yet, it's simply too large change w/o obvious benefits.

PR #2002, opened 2017-11-30
Extend addTest macro

endJunction:
The `addTest` cmake macro now supports tolerances for _each_ of the `DIFF_DATA` lines.

The `Tests.cmake` files were converted with awk:
```sh
find -name 'Tests.cmake' | \
    while read i; do \
        awk '/ABSTOL/{a = $2; r = $4; next};/VIS/{print; next}/vtu/{print $0,a,r; next}{print}' "${i}" > a.new && \
        mv a.new "${i}";\
    done
```

The old version is still available, but emits a "deprecated" message.


endJunction:
The failing jenkins job is so far I've seen not related to the PR's changes and the windows build is checked in appveyor.

PR #2005, opened 2017-12-03
Add ode system preAssemble() call.

endJunction:
In some processes (using non-local methods, for example (upcoming SmallDeformationNonlocal PR)) the assembly is split in two parts: computation of local quantities and actual (non-local) assembly (integration) into the local matrices.

This enhances the interface by a `preAssemble()` function.


chleh:
It looks innocent, though it's hard to judge from this PR, why that's needed at all, and why it can't be simply merged with the assemble() methods...
@endJunction, could you tell us briefly (maybe orally) what's the roadmap this PR belongs to?

PR #2016, opened 2017-12-08
Staggered scheme for coupled processes with different orders of elements

wenqing:
This PR presents the framework for the staggered scheme for the case of  coupled processes with different orders of elements.

As a first implementation within the framework, the staggered scheme is realized in HydroMechanicsProcess. The implementation is tested by solving an example with both of the staggered and monolithic  schemes. 

Besides, the calculation of stress and strain for output and also for the coupling is corrected.

Since the convergence criterion of the coupling loop is changed in this PR, the solutions of the existing benchmark about the staggered scheme have slightly difference from that by the code in the master branch. Therefore, the solutions of the benchmark in Parabolic/HT/StaggeredCoupling/ConstViscosity have been updated in this PR too.


wenqing:
@bilke 
What is the error in CI and appveyor tests with the following message?
```
VTK/7.1.0@bilke/stable: WARN: Can't find a 'VTK/7.1.0@bilke/stable' package for the specified options and settings:

- Settings: arch=x86_64, build_type=Release, compiler=gcc, compiler.libcxx=libstdc++, compiler.version=4.9, os=Linux

- Options: fPIC=False, mpi=False, qt=False, shared=False, x11=True
```

bilke:
@wenqing Should be fixed now!

web/content/docs/benchmarks/hydro-mechanics/InjectionProduction.md
```diff
@@ -0,0 +1,73 @@
+---
```
bilke: This file need to have a so called frontmatter at the beginning. See [other](https://raw.githubusercontent.com/ufz/ogs/master/web/content/docs/benchmarks/elliptic/groundwater-flow-robin.md) for reference.

wenqing:
@bilke  Thanks for fixing.

web/content/docs/benchmarks/hydro-mechanics/InjectionProduction.md
wenqing: I added a formatter manually,  please take a look.

bilke:
@wenqing There are still some error I have to fix left.

Documentation/ProjectFile/prj/time_loop/global_process_coupling/convergence_criteria/t_convergence_criterion.md
```diff
@@ -0,0 +1,2 @@
+Defines the convergence criteria of the global un-coupling loop of the staggered
```
endJunction: "un-coupling loop" is not a common phrase. Do you mean weakly coupled processes:
"Defines the convergence criterion for each weakly coupled process." --- would be this correct?

NumLib/ODESolver/TimeDiscretizedODESystem.h
```diff
@@ -76,12 +76,12 @@ class TimeDiscretizedODESystem<ODESystemTag::FirstOrderImplicitQuasilinear,
      * \param ode the ODE to be wrapped.
      * \param time_discretization the time discretization to be used.
      */
-    explicit TimeDiscretizedODESystem(ODE& ode, TimeDisc& time_discretization);
+    explicit TimeDiscretizedODESystem(const int equation_id, ODE& ode,
+                                      TimeDisc& time_discretization);
```
endJunction: `equation_id` must be commented. There are corresponding doxygen warnings.

ProcessLib/AnalyticalJacobianAssembler.h
```diff
@@ -34,6 +37,14 @@ class AnalyticalJacobianAssembler final : public AbstractJacobianAssembler
         const double dx_dx, std::vector<double>& local_M_data,
         std::vector<double>& local_K_data, std::vector<double>& local_b_data,
         std::vector<double>& local_Jac_data) override;
+
+    void assembleWithJacobianAndCoupling(
```
endJunction: docs missing.
Maybe it would be better to call this `assembleWithJacobianWeakCoupling` because "Coupling" is not a product of this function like "Jacobian".

ProcessLib/CoupledSolutionsForStaggeredScheme.h
```diff
@@ -76,9 +76,13 @@ struct LocalCoupledSolutions
 
 std::vector<std::vector<double>> getPreviousLocalSolutions(
     const CoupledSolutionsForStaggeredScheme& cpl_xs,
-    const std::vector<GlobalIndexType>& indices);
+    const std::vector<
+        std::reference_wrapper<const std::vector<GlobalIndexType>>>&
+        indices);
 
 std::vector<std::vector<double>> getCurrentLocalSolutions(
```
endJunction: Just for better understanding: why is Solution**s** used in plural form? Could this be somehow documented?

I understand that this is not introduced in this PR.

ProcessLib/GroundwaterFlow/GroundwaterFlowProcess.h
```diff
@@ -69,9 +70,12 @@ class GroundwaterFlowProcess final : public Process
             MeshLib::addPropertyToMesh(*_balance_mesh, _balance_pv_name,
                                        MeshLib::MeshItemType::Cell, 1,
                                        init_values);
+            const int process_id = 0;
```
endJunction: This I don't understand: on one side process id is 0 but the passed function's argument is not used.... I'd expect that the callee is deciding which `process_id` is being used...

ProcessLib/HT/HTProcess.cpp
```diff
@@ -46,7 +48,10 @@ void HTProcess::initializeConcreteProcess(
     MeshLib::Mesh const& mesh,
     unsigned const integration_order)
 {
-    ProcessLib::ProcessVariable const& pv = getProcessVariables()[0];
+    // For the staggered scheme, both processes are assumed to use the same
+    // element order.
+    const int process_id = 0;
+    ProcessLib::ProcessVariable const& pv = getProcessVariables(process_id)[0];
```
endJunction: How is the above comment related to the `process_id`? I'd expect the general comments on implementation to be in the header file and in doxygen comments....

ProcessLib/HT/HTProcess.cpp
```diff
@@ -160,5 +165,54 @@ void HTProcess::setCoupledTermForTheStaggeredSchemeToLocalAssemblers()
         _local_assemblers, _coupled_solutions);
 }
 
+NumLib::LocalToGlobalIndexMap* HTProcess::getDOFTableForExtrapolatorData(
+    bool& manage_storage) const
```
endJunction: Because manage_storage is used only as output parameter, it would be better to return it too:
```c++
std::tuple<bool, DOF*> f()
{
   if (A) return {false, dof};
   return {true, other_dof};
}
```

ProcessLib/HT/HTProcess.cpp
```diff
@@ -160,5 +165,54 @@ void HTProcess::setCoupledTermForTheStaggeredSchemeToLocalAssemblers()
...
+    }
+
+    // Otherwise construct a new DOF table.
+    std::vector<MeshLib::MeshSubsets> all_mesh_subsets_single_component;
+    all_mesh_subsets_single_component.emplace_back(
+        _mesh_subset_all_nodes.get());
+
+    manage_storage = true;
+
+    return new NumLib::LocalToGlobalIndexMap(
```
endJunction: Please use automatic storage management for on-heap allocated objects.

ProcessLib/HT/HTProcess.cpp
```diff
@@ -160,5 +165,54 @@ void HTProcess::setCoupledTermForTheStaggeredSchemeToLocalAssemblers()
...
+{
+    const auto number_of_coupled_solutions =
+        _coupled_solutions->coupled_xs.size();
+    _coupled_solutions->coupled_xs_t0.clear();
+    _coupled_solutions->coupled_xs_t0.reserve(number_of_coupled_solutions);
+    const int process_id = _coupled_solutions->process_id;
+    for (std::size_t i = 0; i < number_of_coupled_solutions; i++)
+    {
+        const auto x_t0 = _xs_previous_timestep[process_id].get();
+        if (x_t0 == nullptr)
```
endJunction: Minor:
if the type of `_xs_previous_timestep[process_id]` is `std::unique_ptr` or similar, there is no need to `get` the pointer for comparison or later access:
```c++
auto const& x_t0 = xs[id];
if (!x_t0) OGS_FATAL(...);
```

ProcessLib/HT/HTProcess.h
```diff
@@ -90,6 +88,13 @@ class HTProcess final : public Process
                                     double const dt,
                                     const int process_id) override;
 
+    /// Set the solutions of the previous time step to the coupled term.
+    /// It only performs for the staggered scheme.
+    void setCoupledSolutionsOfPreviousTimeStep();
+
+    NumLib::LocalToGlobalIndexMap* getDOFTableForExtrapolatorData(
```
endJunction: docs missing.

ProcessLib/HydroMechanics/HydroMechanicsFEM.h
```diff
@@ -566,6 +582,20 @@ class HydroMechanicsLocalAssembler : public LocalAssemblerInterface
         return cache;
     }
 
+    void assembleWithJacobianForDeformationEquations(
```
endJunction: docs missing here and below for pressure eqs.

ProcessLib/HydroMechanics/HydroMechanicsFEM.h
```diff
@@ -595,3 +625,5 @@ class HydroMechanicsLocalAssembler : public LocalAssemblerInterface
 
 }  // namespace HydroMechanics
 }  // namespace ProcessLib
+
+#include "HydroMechanicsFEM-impl.h"
```
endJunction: I'd either keep all implementations in the HMFEM.h or move all implementations to HMFEM-impl.h, but not halfway, with my preference being the former case.

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
```diff
@@ -0,0 +1,288 @@
...
+template <typename ShapeFunctionDisplacement, typename ShapeFunctionPressure,
+          typename IntegrationMethod, int DisplacementDim>
+void HydroMechanicsLocalAssembler<ShapeFunctionDisplacement,
+                                  ShapeFunctionPressure, IntegrationMethod,
+                                  DisplacementDim>::
+    postNonLinearSolverConcrete(std::vector<double> const& local_x,
+                                double const t,
+                                bool const use_monolithic_scheme)
+{
+    const int offset = use_monolithic_scheme ? displacement_index : 0;
```
endJunction: better to call it displacement_offset to be consistent with displacement_size and displacement_index.

ProcessLib/HydroMechanics/HydroMechanicsProcess-impl.h
```diff
@@ -48,6 +50,27 @@ bool HydroMechanicsProcess<DisplacementDim>::isLinear() const
     return false;
 }
 
+template <int DisplacementDim>
+MathLib::MatrixSpecifications
+HydroMechanicsProcess<DisplacementDim>::getMatrixSpecifications(
+    const int equation_id) const
+{
+    // For the staggered scheme, equation_id == 1 indicates that the matrix
+    // specifications are for the momentum balance equation (deformation).
```
endJunction: this belongs to the `equation_id` parameter doxygen documentation.

ProcessLib/HydroMechanics/HydroMechanicsProcess-impl.h
```diff
@@ -84,101 +153,109 @@ void HydroMechanicsProcess<DisplacementDim>::initializeConcreteProcess(
...
 {
+    const int mechinical_process_id = _use_monolithic_scheme ? 0 : 1;
+    const int deformation_variable_id = _use_monolithic_scheme ? 1 : 0;
     ProcessLib::HydroMechanics::createLocalAssemblers<
         DisplacementDim, HydroMechanicsLocalAssembler>(
         mesh.getDimension(), mesh.getElements(), dof_table,
         // use displacment process variable for shapefunction order
-        getProcessVariables()[1].get().getShapeFunctionOrder(),
+        getProcessVariables(mechinical_process_id)[deformation_variable_id]
+            .get()
```
endJunction: Use dereference operator instead of get(). A unique pointer is a pointer too with all the related semantics.

ProcessLib/HydroMechanics/HydroMechanicsProcess-impl.h
```diff
@@ -84,101 +153,109 @@ void HydroMechanicsProcess<DisplacementDim>::initializeConcreteProcess(
...
 }
 
 template <int DisplacementDim>
 void HydroMechanicsProcess<DisplacementDim>::assembleConcreteProcess(
     const double t, GlobalVector const& x, GlobalMatrix& M, GlobalMatrix& K,
     GlobalVector& b)
 {
-    DBUG("Assemble HydroMechanicsProcess.");
+    // Not available because that only the Newton-Raphson method is available
+    // for HydroMechanics
```
endJunction: Something is wrong with the english sentence.

endJunction: Also not clear what is "not available".

ProcessLib/HydroMechanics/HydroMechanicsProcess-impl.h
```diff
@@ -188,43 +265,113 @@ void HydroMechanicsProcess<DisplacementDim>::assembleConcreteProcess(
...
+            "HydroMechanics for the staggered scheme.");
+    }
 
-    // Call global assembler for each local assembly item.
     GlobalExecutor::executeMemberDereferenced(
         _global_assembler, &VectorMatrixAssembler::assembleWithJacobian,
         _local_assemblers, *_local_to_global_index_map, t, x, xdot, dxdot_dx,
-        dx_dx, M, K, b, Jac, _coupled_solutions);
+        dx_dx, M, K, b, Jac, _coupled_solutions,
+        _local_to_global_index_map_with_base_nodes.get());
```
endJunction: Why `get()`? Either it exists already, then a reference is perfectly fine, or it does not and there should be a check before usage. Or I'm missing something ...

ProcessLib/HydroMechanics/HydroMechanicsProcess-impl.h
```diff
@@ -188,43 +265,113 @@ void HydroMechanicsProcess<DisplacementDim>::assembleConcreteProcess(
...
     DBUG("PreTimestep HydroMechanicsProcess.");
 
     _process_data.dt = dt;
     _process_data.t = t;
 
-    GlobalExecutor::executeMemberOnDereferenced(
-        &LocalAssemblerInterface::preTimestep, _local_assemblers,
-        *_local_to_global_index_map, x, t, dt);
+    // If monolithic scheme is used or the equation of deformation is solved in
+    // the staggered scheme.
```
endJunction: For understanding: How is it guaranteed, that `preTimestep` is called first in the case of staggered scheme before the other (staggered) process is executed?

endJunction:
For discussion: As I have seen the current implementation adds logic to the processes depending on the process being solved monolithically or staggered. It would be nice if this logic would not "intrude" into the FEM part. Also there are several places checking for various combinations of monolithic-staggered/process_ids which are easily to get wrong in implementation of more complex processes.

endJunction:
Note to myself: I'm half way through the review. Next is ProcessLib/Process.cpp

endJunction:
Update of the copyright note to 2018 is easy with
```sh
ag 'Copyright \(c\) 2012-' -l | xargs sed -i 's/Copyright (c) 2012-2017/Copyright (c) 2012-2018/'
```

ProcessLib/Process.cpp
```diff
@@ -149,7 +162,8 @@ void Process::setInitialConditions(const unsigned pcs_id, double const t,
     }
 }
 
-MathLib::MatrixSpecifications Process::getMatrixSpecifications() const
+MathLib::MatrixSpecifications Process::getMatrixSpecifications(
+    const int /*equation_id*/) const
```
endJunction: What is a difference between equation_id and process_id?

ProcessLib/Process.cpp
```diff
@@ -50,6 +50,38 @@ Process::Process(
 {
 }
 
+/// Initialize the boundary conditions for single PDE
```
endJunction: doxygen comment -> header.

ProcessLib/Process.cpp
```diff
@@ -50,6 +50,38 @@ Process::Process(
...
+
+        std::move(sts.begin(), sts.end(),
+                  std::back_inserter(per_process_source_terms));
+    }
+    _source_terms.push_back(std::move(per_process_source_terms));
+}
+
+void Process::initializeBoundaryConditions()
+{
+    for (std::size_t pcs_id = 0; pcs_id < _process_variables.size(); pcs_id++)
```
endJunction: Mismatch: counting of **process ids** but using **process variable**s' size. I guess it is correct, but the names are misleading.

ProcessLib/Process.cpp
```diff
@@ -50,6 +50,38 @@ Process::Process(
 {
 }
 
+/// Initialize the boundary conditions for single PDE
+void Process::initializeBoundaryConditionPerPDE(
+    const NumLib::LocalToGlobalIndexMap& dof_table, const int process_id)
```
endJunction: Naming issue: A process could consist of multiple PDEs. I think initializeProcessBoundaryCondition or similar would be a better name.

ProcessLib/Process.h
```diff
@@ -57,7 +57,10 @@ class Process
                      const double delta_t, const int process_id);
 
     /// Postprocessing after a complete timestep.
-    void postTimestep(GlobalVector const& x);
+    void postTimestep(GlobalVector const& x, int const processs_id);
+
+    void postNonLinearSolver(GlobalVector const& x, const double t,
```
endJunction: Docu.
Just for understanding, why is this function now necessary, and it was not required before?

ProcessLib/Process.h
```diff
@@ -147,13 +141,22 @@ class Process
         return _extrapolator_data.getDOFTable();
     }
 
+    /// Initialize the boundary conditions for single PDE. It is called by
+    /// initializeBoundaryConditions().
+    void initializeBoundaryConditionPerPDE(
+        const NumLib::LocalToGlobalIndexMap& dof_table, const int process_id);
```
endJunction: can be private, I guess...

ProcessLib/UncoupledProcessesTimeLoop.cpp
```diff
@@ -134,6 +134,8 @@ struct SingleProcessData
     //! cast of \c tdisc_ode_sys to NumLib::InternalMatrixStorage
     NumLib::InternalMatrixStorage* mat_strg = nullptr;
 
+    int equation_id = 0;
```
endJunction: Docu.

Still not clear to me is the difference btw. an equation id and a process id, could you explain, please?

ProcessLib/UncoupledProcessesTimeLoop.cpp
```diff
@@ -330,16 +332,29 @@ std::unique_ptr<UncoupledProcessesTimeLoop> createUncoupledProcessesTimeLoop(
         //! \ogs_file_param{prj__time_loop__global_process_coupling}
         = config.getConfigSubtreeOptional("global_process_coupling");
 
-    std::unique_ptr<NumLib::ConvergenceCriterion> coupling_conv_crit = nullptr;
+    std::vector<std::unique_ptr<NumLib::ConvergenceCriterion>>
+        _global_coupling_conv_criteria;
```
endJunction: underscores for members only...

ProcessLib/UncoupledProcessesTimeLoop.cpp
```diff
@@ -375,7 +390,7 @@ std::vector<GlobalVector*> setInitialConditions(
 {
     std::vector<GlobalVector*> process_solutions;
 
-    unsigned pcs_idx = 0;
+    int pcs_idx = 0;
```
endJunction: Not from this PR, but there is a difference btw. an index and an id. Which one is used here?

ProcessLib/UncoupledProcessesTimeLoop.cpp
```diff
@@ -440,22 +455,28 @@ bool solveOneTimeStepOneProcess(const unsigned process_index,
...
+                                                  process_data.process_output,
+                                                  timestep, t, x, iteration);
     };
 
     bool nonlinear_solver_succeeded =
         nonlinear_solver.solve(x, post_iteration_callback);
 
+    if (nonlinear_solver_succeeded)
+    {
+        process.postNonLinearSolver(x, t, process_id);
```
endJunction: Please add the info about this function being executed only on nonlinear solver success to the `postNonlinearSolver` member documentation.

ProcessLib/UncoupledProcessesTimeLoop.cpp
```diff
@@ -663,13 +684,14 @@ bool UncoupledProcessesTimeLoop::loop()
             auto& pcs = spd->process;
             _output->addProcess(pcs, pcs_idx);
 
+            spd->equation_id = pcs_idx;
```
endJunction: Naming.

ProcessLib/UncoupledProcessesTimeLoop.h
```diff
@@ -72,7 +72,8 @@ class UncoupledProcessesTimeLoop
     /// Maximum iterations of the global coupling.
     const unsigned _global_coupling_max_iterations;
     /// Convergence criteria of the global coupling iterations.
-    std::unique_ptr<NumLib::ConvergenceCriterion> _global_coupling_conv_crit;
+    std::vector<std::unique_ptr<NumLib::ConvergenceCriterion>>
```
endJunction: The docu needs an update about the vector.

ProcessLib/VectorMatrixAssembler.cpp
```diff
@@ -57,8 +58,18 @@ void VectorMatrixAssembler::assemble(
...
+        indices_of_all_coupled_processes.reserve(cpl_xs->coupled_xs.size());
+        for (std::size_t i = 0; i < cpl_xs->coupled_xs.size(); i++)
+        {
+            indices_of_all_coupled_processes.emplace_back(std::ref(indices));
+        }
+
+        auto local_coupled_xs0 = getPreviousLocalSolutions(
+            *cpl_xs, indices_of_all_coupled_processes);
+        auto local_coupled_xs =
+            getCurrentLocalSolutions(*cpl_xs, indices_of_all_coupled_processes);
```
endJunction: I'd think it is possible to pass `indices` instead of `indices_of_all_coupled_processes` because all of the entries do contain the same `indices` just several times. In the `getCurrentLocalSolutions` it could be removed also... Maybe I'm missing something.

ProcessLib/VectorMatrixAssembler.cpp
```diff
@@ -95,9 +106,20 @@ void VectorMatrixAssembler::assembleWithJacobian(
     NumLib::LocalToGlobalIndexMap const& dof_table, const double t,
     GlobalVector const& x, GlobalVector const& xdot, const double dxdot_dx,
     const double dx_dx, GlobalMatrix& M, GlobalMatrix& K, GlobalVector& b,
-    GlobalMatrix& Jac, const CoupledSolutionsForStaggeredScheme* cpl_xs)
+    GlobalMatrix& Jac, const CoupledSolutionsForStaggeredScheme* cpl_xs,
+    NumLib::LocalToGlobalIndexMap const* base_dof_table)
 {
-    auto const indices = NumLib::getIndices(mesh_item_id, dof_table);
+    // If base_dof_table != nullptr, then the coupled processes contains the
+    // mechanical process, which is alway placed in the end of the coupled
```
endJunction: typo; always.

ProcessLib/VectorMatrixAssembler.cpp
```diff
@@ -95,9 +106,20 @@ void VectorMatrixAssembler::assembleWithJacobian(
...
+    // process and always user higher order element than other process in the
```
endJunction: typo: uses;

ProcessLib/VectorMatrixAssembler.cpp
```diff
@@ -114,17 +136,32 @@ void VectorMatrixAssembler::assembleWithJacobian(
...
-
-        ProcessLib::LocalCoupledSolutions local_coupled_solutions(
-            cpl_xs->dt, cpl_xs->process_id, std::move(local_coupled_xs0),
-            std::move(local_coupled_xs));
-
-        _jacobian_assembler->assembleWithJacobianAndCoupling(
-            local_assembler, t, local_xdot, dxdot_dx, dx_dx, _local_M_data,
-            _local_K_data, _local_b_data, _local_Jac_data,
-            local_coupled_solutions);
+        if (base_dof_table == nullptr)
```
endJunction: Needs discussion: Does it also cover the T weakly coupled to HM case?

ProcessLib/VectorMatrixAssembler.cpp
```diff
@@ -160,4 +197,34 @@ void VectorMatrixAssembler::assembleWithJacobian(
...
+    const double dx_dx, CoupledSolutionsForStaggeredScheme const* cpl_xs)
+{
+    std::vector<std::reference_wrapper<const std::vector<GlobalIndexType>>>
+        indices_of_all_coupled_processes;
+    indices_of_all_coupled_processes.reserve(cpl_xs->coupled_xs.size());
+    for (std::size_t i = 0; i < cpl_xs->coupled_xs.size() - 1; i++)
+    {
+        indices_of_all_coupled_processes.emplace_back(std::ref(base_indices));
+    }
+    indices_of_all_coupled_processes.emplace_back(std::ref(full_indices));
```
endJunction: Just curious, does using of ("slightly" awkward) reference_wrapper pays of? I'd think the indices vector is rather short, isn't it?

Tests/Data/HydroMechanics/StaggeredScheme/InjectionProduction1D/InjectionProduction1D.prj
```diff
@@ -0,0 +1,296 @@
...
+        <global_process_coupling>
+            <max_iter> 6 </max_iter>
+            <convergence_criteria>
+                <!-- convergence criterion for the first process -->
+                <convergence_criterion>
+                    <type>DeltaX</type>
+                    <norm_type>NORM2</norm_type>
+                    <reltol>1.e-10</reltol>
+                </convergence_criterion>
+                <!-- convergence criterion for the second process -->
```
endJunction: How is it decided which process is the "first" and which is "second"? XML is unordered.

Tests/Data/HydroMechanics/StaggeredScheme/InjectionProduction1D/InjectionProduction1D.prj
```diff
@@ -0,0 +1,296 @@
...
+                    <type>DeltaX</type>
+                    <norm_type>NORM2</norm_type>
+                    <reltol>1.e-10</reltol>
+                </convergence_criterion>
+            </convergence_criteria>
+        </global_process_coupling>
+
+        <processes>
+            <!--For the equations of pressure-->
+            <process ref="HM">
```
endJunction: what does the "ref=HM" mean?

Tests/Data/HydroMechanics/StaggeredScheme/InjectionProduction1D/InjectionProduction1D.prj
```diff
@@ -0,0 +1,296 @@
...
+            <eigen>
+                <solver_type>CG</solver_type>
+                <precon_type>DIAGONAL</precon_type>
+                <max_iteration_step>10000</max_iteration_step>
+                <error_tolerance>1e-10</error_tolerance>
+            </eigen>
+        </linear_solver>
+        <linear_solver>
+            <name>linear_solver_u</name>
+            <lis>-i bicg -p jacobi -tol 1e-16 -maxiter 10000</lis>
```
endJunction: Did you test it with LIS or is it just leftover from copying?

Tests/Data/HydroMechanics/StaggeredScheme/InjectionProduction1D/RerefenceSolutionByMonolithicScheme/InjectionProduction1DMono.prj
```diff
@@ -0,0 +1,250 @@
...
+            <type>Constant</type>
+            <values>0.</values>
+        </parameter>
+        <parameter>
+            <name>dirichlet0</name>
+            <type>Constant</type>
+            <value>0</value>
+        </parameter>
+        <parameter>
+            <name>oberburden</name>
```
endJunction: What does 'oberburden' mean?

ProcessLib/CoupledSolutionsForStaggeredScheme.h
```diff
@@ -76,9 +76,13 @@ struct LocalCoupledSolutions
 
 std::vector<std::vector<double>> getPreviousLocalSolutions(
```
endJunction: Docu missing for both 'previous' and 'current', especially now it is extended by a non-trivial parameter.

web/content/docs/benchmarks/hydro-mechanics/InjectionProduction.md
```diff
@@ -0,0 +1,93 @@
...
++++
+
+{{< data-link >}}
+
+---
+#     Consolidation example based on fluid injection and production    application
+
+{{< img src="../InjectionProduction.png" >}}
+
+This benchmark is based on an example about injection and
```
endJunction: English. Need to go over together.

ProcessLib/HydroMechanics/Tests.cmake
```diff
@@ -84,3 +86,31 @@ AddTest(
...
+    PATH HydroMechanics/StaggeredScheme/InjectionProduction1D/RerefenceSolutionByMonolithicScheme
+    EXECUTABLE ogs
+    EXECUTABLE_ARGS InjectionProduction1DMono.prj
+    WRAPPER time
+    TESTER vtkdiff
+    REQUIREMENTS NOT OGS_USE_MPI
+    DIFF_DATA
+    InjectionProduction1D_pcs_1_ts_100_t_8640000.000000.vtu InjectionProduction1D_Mono_pcs_0_ts_100_t_8640000.000000.vtu displacement displacement 1e-11 1e-11
+    InjectionProduction1D_pcs_1_ts_100_t_8640000.000000.vtu InjectionProduction1D_Mono_pcs_0_ts_100_t_8640000.000000.vtu pressure pressure 1e-11 1e-11
+)
```
endJunction: Velocity, sigma, and eps are present in the results but are not verified in the test.

ProcessLib/HydroMechanics/HydroMechanicsFEM.h
wenqing: If it is the former case, we have a large header file mixed with declarations and definitions. I  will move the definitions of some large members to -impl.h file.

wenqing:
@endJunction  I will answer your comments later.

Documentation/ProjectFile/prj/time_loop/global_process_coupling/convergence_criteria/t_convergence_criterion.md
wenqing: Changed accordingly.

NumLib/ODESolver/TimeDiscretizedODESystem.h
wenqing: Commented.

ProcessLib/AnalyticalJacobianAssembler.h
wenqing: Docs was added and the name was changed as well.

ProcessLib/CoupledSolutionsForStaggeredScheme.h
wenqing: Documented these two functions.

ProcessLib/GroundwaterFlow/GroundwaterFlowProcess.h
wenqing: Added a comment.

ProcessLib/HT/HTProcess.cpp
wenqing: Added more a detailed description.

ProcessLib/HT/HTProcess.cpp
wenqing: Changed the return type to std::tuple.

ProcessLib/HT/HTProcess.cpp
wenqing: Explained in the discussion.

ProcessLib/HT/HTProcess.cpp
wenqing: removed `.get()`.

ProcessLib/HT/HTProcess.h
wenqing: Added a documentation.

ProcessLib/HydroMechanics/HydroMechanicsFEM.h
wenqing: Added.

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
wenqing: Changed accordingly.

ProcessLib/HydroMechanics/HydroMechanicsProcess-impl.h
wenqing: Replaced this comment with a simple one. Besides, `equation_id` is documented in the header file.

ProcessLib/HydroMechanics/HydroMechanicsProcess-impl.h
wenqing: Changed the description.

ProcessLib/HydroMechanics/HydroMechanicsProcess-impl.h
wenqing: ``_local_to_global_index_map_with_base_nodes.get())`` is removed now.

ProcessLib/HydroMechanics/HydroMechanicsProcess-impl.h
wenqing: If is called before a time step. For HydroMechanicsProcess, the function only updates strain, stress.

ProcessLib/Process.cpp
wenqing: Unified to use `process_id`.

ProcessLib/Process.cpp
wenqing: Removed.

ProcessLib/Process.cpp
wenqing: Added a description above the loop.

ProcessLib/Process.cpp
wenqing: Changed the name accordingly.

ProcessLib/Process.h
wenqing: This  corrects stress and strain for output and for coupling. In the present master branch, the strain together with stress are calculated by using the displacement of one iteration step forward  before nonlinear solver convergences.

ProcessLib/Process.h
wenqing: It is also called by  HydroMechanicsProcess, the derived class. Therefore, it is in the protected section.

ProcessLib/UncoupledProcessesTimeLoop.cpp
wenqing: Documented, and changed it to `process_id`.

ProcessLib/UncoupledProcessesTimeLoop.cpp
wenqing: Changed.

ProcessLib/UncoupledProcessesTimeLoop.cpp
wenqing: Changed to `process_id`.

ProcessLib/UncoupledProcessesTimeLoop.cpp
wenqing: Modified the documentation of postNonLinearSolver as
```
    /// Calculates secondary variables, e.g. stress and strain for deformation
    /// analysis, only after nonlinear solver being successfully conducted.
```

ProcessLib/UncoupledProcessesTimeLoop.cpp
wenqing: Changed

ProcessLib/UncoupledProcessesTimeLoop.h
wenqing: Added.

ProcessLib/VectorMatrixAssembler.cpp
wenqing: In VectorMatrixAssembler.cpp, the part about fetching indices has been simplified, and it is now more general for coupling and easy to understand.

ProcessLib/VectorMatrixAssembler.cpp
wenqing: Removed. Also see my previous answer
>In VectorMatrixAssembler.cpp, the part about fetching indices has been simplified, and it is now more general for coupling and easy to understand.

ProcessLib/VectorMatrixAssembler.cpp
wenqing: The comment is gone.

ProcessLib/VectorMatrixAssembler.cpp
wenqing: It is gone. Now a vector of DOF tables is passed to this function, which makes coupling more flexible. See the commit entitled with
``
[Coupling] Passed a vector of DOF tables to VectorMatrixAssembler ``

ProcessLib/VectorMatrixAssembler.cpp
wenqing: Changed. See my previous reply.

Tests/Data/HydroMechanics/StaggeredScheme/InjectionProduction1D/InjectionProduction1D.prj
wenqing: Added an attribution of `id`.

Tests/Data/HydroMechanics/StaggeredScheme/InjectionProduction1D/InjectionProduction1D.prj
wenqing: HydroMechanics. Changed it to InjectionProduction1D.

Tests/Data/HydroMechanics/StaggeredScheme/InjectionProduction1D/InjectionProduction1D.prj
wenqing: removed.

Tests/Data/HydroMechanics/StaggeredScheme/InjectionProduction1D/RerefenceSolutionByMonolithicScheme/InjectionProduction1DMono.prj
wenqing: Corrected it to overburden.

ProcessLib/CoupledSolutionsForStaggeredScheme.h
wenqing: Added.

ProcessLib/UncoupledProcessesTimeLoop.cpp
```diff
@@ -330,16 +335,37 @@ std::unique_ptr<UncoupledProcessesTimeLoop> createUncoupledProcessesTimeLoop(
...
+            auto const id =
+                coupling_convergence_criterion_config.getConfigAttribute<int>(
+                    "id");
+            criterion_ids.push_back(id);
+
+            global_coupling_conv_criteria.push_back(
+                NumLib::createConvergenceCriterion(
+                    coupling_convergence_criterion_config));
+        }
+        BaseLib::reorderVector(global_coupling_conv_criteria, criterion_ids);
```
chleh: Do the `criterion_ids` have any additional meaning, or are they just meant to put the `global_coupling_conv_criteria` into the right order?

ProcessLib/UncoupledProcessesTimeLoop.cpp
```diff
@@ -916,24 +961,26 @@ bool UncoupledProcessesTimeLoop::solveCoupledEquationSystemsByStaggeredScheme(
...
+                MathLib::LinAlg::axpy(x_old, -1.0, x);  // save dx to x_old
+                _global_coupling_conv_crit[process_id]->checkDeltaX(x_old, x);
+                coupling_iteration_converged =
+                    coupling_iteration_converged &&
+                    _global_coupling_conv_crit[process_id]->isSatisfied();
+            }
             MathLib::LinAlg::copy(x, x_old);
 
-            if (coupling_iteration_converged)
+            if (coupling_iteration_converged && global_coupling_iteration > 0)
```
chleh: Maybe it's better to remove this break statement and the `&& global_coupling_iteration > 0` conditions. Let's discuss.

ProcessLib/UncoupledProcessesTimeLoop.cpp
chleh: Is there a check that the number of convergence criteria == the number of processes?
Maybe it's better to drop the `criterion_ids` and take the `global_coupling_conv_criteria` in the order  they are read.

ProcessLib/VectorMatrixAssembler.cpp
```diff
@@ -39,12 +40,22 @@ void VectorMatrixAssembler::preAssemble(
...
-    auto const indices = NumLib::getIndices(mesh_item_id, dof_table);
+    std::vector<std::vector<GlobalIndexType>> indices_of_processes;
+    indices_of_processes.reserve(dof_tables.size());
+    for (std::size_t i = 0; i < dof_tables.size(); i++)
+    {
+        indices_of_processes.emplace_back(
+            NumLib::getIndices(mesh_item_id, dof_tables[i].get()));
+    }
 
+    auto const& indices = (dof_tables.size() == 1 && cpl_xs == nullptr)
```
chleh: Drop the `dof_tables.size() == 1` ?

Tests/Data/HydroMechanics/StaggeredScheme/InjectionProduction1D/InjectionProduction1D.prj
chleh: I just said the contrary in my review. :smile:
Cf. https://stackoverflow.com/a/1131547

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
```diff
@@ -0,0 +1,576 @@
+/**
```
chleh: `git diff aa11acf3ab6e0c2aeab6f0efb5ccf48246e65cba^ upstream/master -- ProcessLib/HydroMechanics/HydroMechanicsFEM.h` tells me that there are only very small changes in this file.

ProcessLib/HydroMechanics/HydroMechanicsProcess-impl.h
```diff
@@ -84,146 +158,231 @@ void HydroMechanicsProcess<DisplacementDim>::initializeConcreteProcess(
...
+                         _local_assemblers,
+                         &LocalAssemblerInterface::getIntPtDarcyVelocity));
+}
+
+template <int DisplacementDim>
+void HydroMechanicsProcess<DisplacementDim>::initializeBoundaryConditions()
+{
+    if (_use_monolithic_scheme)
+    {
+        const int process_id_of_up = 0;
```
chleh: Minor: `process_id_of_u_and_p` might be a tiny bit more intuitive to understand.

ProcessLib/HydroMechanics/HydroMechanicsProcess-impl.h
```diff
@@ -84,146 +158,231 @@ void HydroMechanicsProcess<DisplacementDim>::initializeConcreteProcess(
...
+                "HydroMechanics for the staggered scheme.");
+        }
+        else
+        {
+            DBUG(
+                "Assemble the Jacobian equations of mechanical process in "
+                "HydroMechanics for the staggered scheme.");
+        }
+        std::vector<std::reference_wrapper<NumLib::LocalToGlobalIndexMap>>
+            dof_tables = {std::ref(*_local_to_global_index_map_with_base_nodes),
```
chleh: this is a newly declared variable, which will be destroyed in the next line at the end of the scope.

chleh: Is there a staggered HM test?

ProcessLib/HydroMechanics/HydroMechanicsProcess-impl.h
```diff
@@ -84,146 +158,231 @@ void HydroMechanicsProcess<DisplacementDim>::initializeConcreteProcess(
...
         &LocalAssemblerInterface::postTimestep, _local_assemblers,
-        *_local_to_global_index_map, x);
+        getDOFTable(process_id), x);
+}
+
+template <int DisplacementDim>
+void HydroMechanicsProcess<DisplacementDim>::postNonLinearSolverProcess(
+    GlobalVector const& x, const double t, const int process_id)
+{
+    if (!_use_monolithic_scheme && process_id == 0)
```
chleh: Maybe there should be a method `! isLastProcess()` for this condition.

Tests/Data/HydroMechanics/StaggeredScheme/InjectionProduction1D/InjectionProduction1D.prj
wenqing: removed  `id`.

ProcessLib/GroundwaterFlow/GroundwaterFlowProcess.h
```diff
@@ -69,9 +70,13 @@ class GroundwaterFlowProcess final : public Process
             MeshLib::addPropertyToMesh(*_balance_mesh, _balance_pv_name,
                                        MeshLib::MeshItemType::Cell, 1,
                                        init_values);
+            //For this single process, process_id is always zero.
+            const int process_id = 0;
```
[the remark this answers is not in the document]
wenqing: Changed accordingly.

ProcessLib/HT/HTProcess.cpp
```diff
@@ -78,9 +85,12 @@ void HTProcess::assembleConcreteProcess(const double t,
                                         GlobalMatrix& K,
                                         GlobalVector& b)
 {
+    std::vector<std::reference_wrapper<NumLib::LocalToGlobalIndexMap>>
+        dof_tables;
     if (_use_monolithic_scheme)
     {
         DBUG("Assemble HTProcess.");
+        dof_tables.emplace_back(std::ref(*_local_to_global_index_map));
```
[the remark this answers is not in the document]
wenqing: Changed accordingly.

ProcessLib/HT/HTProcess.cpp
```diff
@@ -160,5 +180,54 @@ void HTProcess::setCoupledTermForTheStaggeredSchemeToLocalAssemblers()
         _local_assemblers, _coupled_solutions);
 }
 
+std::tuple<NumLib::LocalToGlobalIndexMap*, bool>
+    HTProcess::getDOFTableForExtrapolatorData() const
```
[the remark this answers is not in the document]
wenqing: It can be common to  all processes by specify one DOF table for extrapolation. To do this we have to  replace `Process::_local_to_global_index_map` with  `Process::_local_to_global_index_map_single_component` 
I prefer to  try it later in other PR.

ProcessLib/HT/HTProcess.cpp
```diff
@@ -160,5 +180,54 @@ void HTProcess::setCoupledTermForTheStaggeredSchemeToLocalAssemblers()
...
+    const int process_id = _coupled_solutions->process_id;
+    for (std::size_t i = 0; i < number_of_coupled_solutions; i++)
+    {
+        const auto& x_t0 = _xs_previous_timestep[process_id];
+        if (x_t0 == nullptr)
+        {
+            OGS_FATAL(
+                "Memory is not allocated for the global vector "
+                "of the solution of the previous time step for the ."
+                "staggered scheme.\n It can be done by overloading "
```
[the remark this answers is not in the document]
wenqing: Changed.

ProcessLib/HT/StaggeredHTFEM.h
```diff
@@ -62,7 +62,7 @@ class StaggeredHTFEM : public HTFEM<ShapeFunction, IntegrationMethod, GlobalDim>
     {
     }
 
-    void assembleWithCoupledTerm(
+    void assembleForStaggeredScheme(
```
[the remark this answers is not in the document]
wenqing: Yes it is possible. There is a member of `assembleWithJacobianForStaggeredScheme` in LocalAssemblerInterface for overriding in the derived process class. In this PR, HydroMechanicsProcess uses  `assembleWithJacobianForStaggeredScheme`.

ProcessLib/HydroMechanics/HydroMechanicsProcess-impl.h
wenqing: It is changed to `process_id_of_hydromechancs`

ProcessLib/HydroMechanics/HydroMechanicsProcess-impl.h
wenqing: As told in discussion, it was corrected and there is a test.

ProcessLib/HydroMechanics/HydroMechanicsProcess-impl.h
wenqing: Introduced a new member:  
`bool HydroMechanicsProcess::hasMechanicalProcess(int const process_id) `.

ProcessLib/LocalAssemblerInterface.h
```diff
@@ -78,6 +78,11 @@ class LocalAssemblerInterface
                               NumLib::LocalToGlobalIndexMap const& dof_table,
                               GlobalVector const& x);
 
+    virtual void postNonLinearSolver(std::size_t const mesh_item_id,
+                              NumLib::LocalToGlobalIndexMap const& dof_table,
+                              GlobalVector const& x, double const t,
+                              bool const use_monolithic_scheme);
```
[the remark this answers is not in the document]
wenqing: removed `virtual`.

ProcessLib/PhaseField/PhaseFieldProcess-impl.h
```diff
@@ -142,10 +142,12 @@ void PhaseFieldProcess<DisplacementDim>::assembleConcreteProcess(
 {
     DBUG("Assemble PhaseFieldProcess.");
 
+    std::vector<std::reference_wrapper<NumLib::LocalToGlobalIndexMap>>
+       dof_table = {std::ref(*_local_to_global_index_map)};
     // Call global assembler for each local assembly item.
     GlobalExecutor::executeMemberDereferenced(
         _global_assembler, &VectorMatrixAssembler::assemble, _local_assemblers,
-        *_local_to_global_index_map, t, x, M, K, b, _coupled_solutions);
+        dof_table, t, x, M, K, b, _coupled_solutions);
```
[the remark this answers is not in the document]
wenqing: Changed.

ProcessLib/Process.cpp
```diff
@@ -50,6 +50,38 @@ Process::Process(
 {
 }
 
+void Process::initializeProcessBoundaryCondition(
```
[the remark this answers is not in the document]
wenqing: Changed.

ProcessLib/Process.cpp
```diff
@@ -69,32 +101,16 @@ void Process::initialize()
...
+    initializeBoundaryConditions();
 }
 
-void Process::setInitialConditions(const unsigned pcs_id, double const t,
+void Process::setInitialConditions(const int process_id, double const t,
                                    GlobalVector& x)
 {
-    auto const& per_process_variables = _process_variables[pcs_id];
+    // getDOFTableOfProcess can be overloaded by the specific process.
+    auto const dof_table_of_process = getDOFTable(process_id);
```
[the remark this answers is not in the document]
wenqing: Changed it to reference type.

ProcessLib/Process.cpp
```diff
@@ -325,10 +346,17 @@ void Process::preTimestep(GlobalVector const& x, const double t,
...
     MathLib::LinAlg::setLocalAccessibleVector(x);
-    postTimestepConcreteProcess(x);
+    postTimestepConcreteProcess(x, process_id);
+}
+
+void Process::postNonLinearSolver(GlobalVector const& x, const double t,
+                                  int const process_id)
+{
+    MathLib::LinAlg::setLocalAccessibleVector(x);
+    postNonLinearSolverProcess(x, t, process_id);
```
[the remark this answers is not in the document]
wenqing: Changed.

ProcessLib/RichardsFlow/RichardsFlowProcess.cpp
```diff
@@ -43,7 +43,8 @@ void RichardsFlowProcess::initializeConcreteProcess(
     MeshLib::Mesh const& mesh,
     unsigned const integration_order)
 {
-    ProcessLib::ProcessVariable const& pv = getProcessVariables()[0];
+    const int process_id = 0;
```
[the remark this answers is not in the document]
wenqing: I prefer to keep this name because there is only one process in  RichardsFlowProcess, and the coupling scheme type is not necessary for  it.

ProcessLib/UncoupledProcessesTimeLoop.cpp
wenqing: Added an assertion for `global_coupling_conv_criteria` and dropped `id`.

ProcessLib/UncoupledProcessesTimeLoop.cpp
wenqing: Done as what discussed.

wenqing:
@chleh  All your comments have been addressed.

wenqing:
@endJunction  and @chleh  Thanks your review and comments.

ProcessLib/HT/HTProcess.cpp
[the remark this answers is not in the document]
chleh: OK, later PR is totally fine and it's not too important.

ProcessLib/HydroMechanics/HydroMechanicsProcess-impl.h
chleh: OK, for some occasions the usage of this method is more mechanics related (e.g. in getDOFtable), and sometimes it is more is-last-process related (e.g. in post timestep).

ProcessLib/UncoupledProcessesTimeLoop.cpp
```diff
@@ -343,6 +360,11 @@ std::unique_ptr<UncoupledProcessesTimeLoop> createUncoupledProcessesTimeLoop(
         //! \ogs_file_param{prj__time_loop__processes}
         config.getConfigSubtree("processes"), processes, nonlinear_solvers);
 
+    if (coupling_config)
+    {
+        assert(global_coupling_conv_criteria.size() == per_process_data.size());
```
chleh: Thanks for adding the check. Actually I'd prefer using OGS_FATAL, because this is a cheap check which is performed only once and the outcome depends on user input. Therefore the check should also be performed in release mode.

wenqing: Changed to use OGS_FATAL.

ProcessLib/UncoupledProcessesTimeLoop.cpp
```diff
@@ -360,6 +360,18 @@ std::unique_ptr<UncoupledProcessesTimeLoop> createUncoupledProcessesTimeLoop(
         //! \ogs_file_param{prj__time_loop__processes}
         config.getConfigSubtree("processes"), processes, nonlinear_solvers);
 
+    if (coupling_config)
+    {
+        if (global_coupling_conv_criteria.size() != per_process_data.size())
+        {
+            OGS_FATAL(
```
wenqing: Changed to  OGS_FATAL

ProcessLib/RichardsFlow/RichardsFlowProcess.cpp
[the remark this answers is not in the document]
endJunction: @chleh I think, that `process_id` is good. Maybe `Process` could be renamed into `CoupledProcess` in far feature.
Locally, of course, if both, the staggered and the monolithic process id appear in the same context, one could distinguish them better using `monolithic_process_id` or `staggerd_H_process_id` or similar.

PR #2027, opened 2017-12-15
Mechanics CTest updates

endJunction:
See commit messages for explanations.
Reason for changes is that it is not passing on my computer.


endJunction:
The CI failures are not related to the changes and the ctests passed where they were tested... merging.

PR #2032, opened 2017-12-22
[Num] Added a function in NaturalCoordinatesMapping.cpp to check the determinant of Jacobian.

wenqing:
The function gives more detailed message about the reason to failure.  Besides, the checking of the determinant of Jacobian  is moved to position before using the determinant.

### DIFF
--- a/NumLib/Fem/CoordinatesMapping/NaturalCoordinatesMapping.cpp
+++ b/NumLib/Fem/CoordinatesMapping/NaturalCoordinatesMapping.cpp
@@ -96,6 +96,28 @@ computeMappingMatrices(
 {
 }
 
+static void checkJacobianDeterminant(const double detJ)
+{
+    if (detJ <= .0)
+    {
+        if (detJ == .0)
+        {
+            OGS_FATAL(
+                "det J is zero. Please check whether:\n"
+                "\t element nodes may have the same coordinates,\n"
+                "\t or the coordinates of all nodes are not given in x "
+                "component for 1D problem,\n"
+                "\t or the coordinates of all nodes are not given in x-y "
+                "components for 2D problem,\n");
+        }
+
+        OGS_FATAL(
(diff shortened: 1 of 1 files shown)


NumLib/Fem/CoordinatesMapping/NaturalCoordinatesMapping.cpp
```diff
@@ -96,6 +96,28 @@ computeMappingMatrices(
 {
 }
 
+static void checkJacobianDeterminant(const double detJ)
+{
+    if (detJ <= .0)
+    {
+        if (detJ == .0)
```
endJunction: The cases could be simplified while reducing nestedness:
```c++
if (detJ < 0)
   OGS_FATAL(...);
if (detJ == 0)
   OGS_FATAL(...);
```

NumLib/Fem/CoordinatesMapping/NaturalCoordinatesMapping.cpp
```diff
@@ -96,6 +96,28 @@ computeMappingMatrices(
...
+        {
+            OGS_FATAL(
+                "det J is zero. Please check whether:\n"
+                "\t element nodes may have the same coordinates,\n"
+                "\t or the coordinates of all nodes are not given in x "
+                "component for 1D problem,\n"
+                "\t or the coordinates of all nodes are not given in x-y "
```
endJunction: by 'x component' and 'x-y component' you mean the x-axis and x-y-plane?

endJunction:
@wenqing please have a look at PR https://github.com/wenqing/ogs/pull/14 which relates to the above comments.

NumLib/Fem/CoordinatesMapping/NaturalCoordinatesMapping.cpp
wenqing: Yes, it is (as your changes in your PR).

PR #2033, opened 2018-01-03
Use Qt 5.9.2 (conan) and disabled Travis.

bilke:
Also: removed testing on Travis  for the moment, see commit messages. Everything else is green.


endJunction:
@bilke I suspect our `git diff --check` disappeared with disabled travis. At least I could not find the check in Jenkinsfile....

PR #2034, opened 2018-01-05
Fix boundary element search hex20 bug

endJunction:
The second face (`face 1`) of a hex 20 was wrong; should contain node 16, not 11. This is fixed in the first commit and is the actual bugfix. A test is added.

Other small things (comments and makeVectorUnique) popped up during debugging session.

Thanks to Peter for reporting and to Tom for debug session. :1st_place_medal:

### DIFF
--- a/Tests/MeshLib/TestBoundaryElementSearch.cpp
+++ b/Tests/MeshLib/TestBoundaryElementSearch.cpp
@@ -34,10 +34,17 @@ public:
         _quad_mesh(MeshGenerator::generateRegularQuadMesh(_geometric_size, _number_of_subdivisions_per_direction))
     {}
 
+    ~MeshLibBoundaryElementSearchInSimpleQuadMesh() override
+    {
+        for (auto p : _pnts)
+            delete p;
+    }
+
 protected:
     const double _geometric_size;
     const std::size_t _number_of_subdivisions_per_direction;
     std::unique_ptr<Mesh> _quad_mesh;
+    std::vector<GeoLib::Point*> _pnts;
 };
 
 class MeshLibBoundaryElementSearchInSimpleHexMesh : public testing::Test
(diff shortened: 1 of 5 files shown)


Tests/MeshLib/TestBoundaryElementSearch.cpp
```diff
@@ -175,3 +176,68 @@ TEST_F(MeshLibBoundaryElementSearchInSimpleHexMesh, SurfaceSearch)
...
+    sfc_bottom.addTriangle(0, 2, 3);
+
+    GeoLib::Surface sfc_front(pnts);
+    sfc_front.addTriangle(0, 1, 4);
+    sfc_front.addTriangle(0, 4, 5);
+
+    // perform search on the bottom surface
+    MeshGeoToolsLib::MeshNodeSearcher mesh_node_searcher(
+        *mesh,
+        // std::make_unique<MeshGeoToolsLib::HeuristicSearchLength>(*mesh),
```
chleh: Delete that line?

Tests/MeshLib/TestBoundaryElementSearch.cpp
```diff
@@ -175,3 +176,68 @@ TEST_F(MeshLibBoundaryElementSearchInSimpleHexMesh, SurfaceSearch)
         delete p;
 }
 
+// This is identical to the above
+// MeshLibBoundaryElementSearchInSimpleHexMesh.SurfaceSearch test but creates a
+// quadratic mesh from the original hex mesh.
```
chleh: Maybe then extract the common part of the tests to some function?

endJunction: Some restructuring happened, though not very generalized, but code duplication at the top level is avoided.

PR #2036, opened 2018-01-09
[VTU] Fixed pvd output

### LINKED ISSUE #2020: pvd contains only last step
at least in TaskB.prj checked for current master c9abfb8c1

wenqing:
This PR fixes #2020.

### DIFF
--- a/ProcessLib/Output.cpp
+++ b/ProcessLib/Output.cpp
@@ -134,35 +134,6 @@ void Output::addProcess(ProcessLib::Process const& process,
                                  std::forward_as_tuple(filename));
 }
 
-Output::SingleProcessData Output::findSingleProcessData(
-    Process const& process, const unsigned process_id) const
-{
-    if (process.isMonolithicSchemeUsed())
-    {
-        auto spd_it = _single_process_data.find(&process);
-        if (spd_it == _single_process_data.end()) {
-            OGS_FATAL("The given process is not contained in the output"
-                      " configuration. Aborting.");
-        }
-        return spd_it->second;
-    }
-
-    auto spd_range = _single_process_data.equal_range(&process);
-    unsigned counter = 0;
-    for (auto spd_it=spd_range.first; spd_it!=spd_range.second; ++spd_it)
-    {
-        if(counter == process_id)
-        {
-            return spd_it->second;
-        }
-        counter++;
-    }
-
-    OGS_FATAL("The given process is not contained in the output"
-                      " configuration. Aborting.");
-}
-
-
 void Output::doOutputAlways(Process const& process,
                             const int process_id,
                             ProcessOutput const& process_output,
@@ -170,10 +141,32 @@ void Output::doOutputAlways(Process const& process,
                             const double t,
                             GlobalVector const& x)
 {
+    const bool make_output =
+        !(process_id < static_cast<int>(_single_process_data.size()) - 1 &&
+          !(process.isMonolithicSchemeUsed()));
+    if (!make_output)
+        return;
+ 
     BaseLib::RunTime time_output;
     time_output.start();
 
-    auto spd = findSingleProcessData(process, process_id);
+    auto spd_range = _single_process_data.equal_range(&process);
+    int counter = 0;
+    SingleProcessData* spd_ptr = nullptr;
+    for (auto spd_it=spd_range.first; spd_it!=spd_range.second; ++spd_it)
+    {
+        if(counter == process_id)
+        {
+            spd_ptr = &spd_it->second;
+            break;
+        }
+        counter++;
+    }
+    if (spd_ptr == nullptr)
+    {
+        OGS_FATAL("The given process is not contained in the output"
+                  " configuration. Aborting.");
+    }
 
     std::string const output_file_name =
             _output_file_prefix + "_pcs_" + std::to_string(process_id)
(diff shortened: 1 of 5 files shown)


ProcessLib/Output.cpp
```diff
@@ -134,46 +134,39 @@ void Output::addProcess(ProcessLib::Process const& process,
...
 void Output::doOutputAlways(Process const& process,
                             const int process_id,
                             ProcessOutput const& process_output,
                             unsigned timestep,
                             const double t,
                             GlobalVector const& x)
 {
+    const bool make_output =
+        !(process_id < static_cast<int>(_single_process_data.size()) - 1 &&
+          !(process.isMonolithicSchemeUsed()));
```
chleh: I think the expression would be more readable if it was rephrased as:
```
process_id == static_cast<int>(_single_process_data.size()) - 1 
         || process.isMonolithicSchemeUsed();
```

Btw.: Is there a special reason why the last process writes output (i.e. `process_id == static_cast<int>(_single_process_data.size()) - 1`), and not the first one?

ProcessLib/Output.cpp
```diff
@@ -134,46 +134,39 @@ void Output::addProcess(ProcessLib::Process const& process,
...
     BaseLib::RunTime time_output;
     time_output.start();
 
-    auto spd = findSingleProcessData(process, process_id);
+    auto spd_range = _single_process_data.equal_range(&process);
+    int counter = 0;
+    SingleProcessData* spd_ptr = nullptr;
+    for (auto spd_it=spd_range.first; spd_it!=spd_range.second; ++spd_it)
+    {
+        if(counter == process_id)
```
chleh: Just curious: Is there a one-to-one correspondence between a `process_id` and a `process` in `_single_process_data`?

ProcessLib/Output.cpp
```diff
@@ -249,10 +235,32 @@ void Output::doOutputNonlinearIteration(Process const& process,
         return;
     }
 
+    const bool make_output =
+        !(process_id < static_cast<int>(_single_process_data.size()) - 1 &&
+          !(process.isMonolithicSchemeUsed()));
```
chleh: Could also be simplified.

ProcessLib/Output.cpp
```diff
@@ -249,10 +235,32 @@ void Output::doOutputNonlinearIteration(Process const& process,
...
+    if (!make_output)
+        return;
+
     BaseLib::RunTime time_output;
     time_output.start();
 
-    findSingleProcessData(process, process_id);
+    auto spd_range = _single_process_data.equal_range(&process);
```
chleh: Maybe this entire green block could be moved to a separate method `SPD* getSingleProcessData(int process_id)` and also used above.

ProcessLib/Output.cpp
```diff
@@ -262,16 +270,12 @@ void Output::doOutputNonlinearIteration(Process const& process,
             + ".vtu";
     std::string const output_file_path = BaseLib::joinPaths(_output_directory, output_file_name);
     DBUG("output iteration results to %s", output_file_path.c_str());
-    const bool make_out =
-        !(process_id < static_cast<int>(_single_process_data.size()) - 1 &&
-          !(process.isMonolithicSchemeUsed()));
-    doProcessOutput(output_file_path, make_out, _output_file_compression,
+    doProcessOutput(output_file_path, make_output, _output_file_compression,
```
chleh: Is it really/still necessary that `doProcessOutput()` is passed the argument `make_output`?

ProcessLib/Output.cpp
wenqing: Changed. A comment was added as the reason as
```
    // For the staggered scheme for the coupling, only the last process, which
    // gives the latest solution within a coupling loop, is allowed to make
    // output.
```

ProcessLib/Output.cpp
wenqing: Yes, they are. However In the present code, each process outputs VTU file with all variables of all processes.  We may need an option that allows one process only output its own variables in future. Therefore, I keep Output::SingleProcessData for this purpose. Otherwise, 
1.  struct  Output::SingleProcessData  can be removed by enabling MeshLib::IO::PVDFile pvd_file as a member of class Output directly.
2. ` std::multimap<Process const*, SingleProcessData> _single_process_data;` can be dropped too.

ProcessLib/Output.cpp
wenqing: Done.

ProcessLib/Output.cpp
wenqing: Added a new member function for that.  Since the new member can not have const modifier,  the const modifer of `doOutputNonlinearIteration` was dropped.

ProcessLib/Output.cpp
wenqing: Removed.

wenqing: Added it again.  Because doProcessOutput adds variable to vtu although the output can be skipped.

wenqing:
@endJunction  It is OK if all benchmarks pass the test.

bilke:
@wenqing Jenkins is fine again.

ProcessLib/Output.cpp
```diff
@@ -134,35 +134,30 @@ void Output::addProcess(ProcessLib::Process const& process,
...
-                      " configuration. Aborting.");
-        }
-        return spd_it->second;
-    }
-
     auto spd_range = _single_process_data.equal_range(&process);
-    unsigned counter = 0;
+    int counter = 0;
+    SingleProcessData* spd_ptr = nullptr;
     for (auto spd_it=spd_range.first; spd_it!=spd_range.second; ++spd_it)
```
endJunction: There is `safe_advance` solution on SO https://stackoverflow.com/a/1057782 .
Could be implemented in BaseLib or be done later.

ProcessLib/Output.cpp
chleh: OK, thanks.

wenqing:
@bilke  Thanks.

ProcessLib/Output.cpp
chleh: > However In the present code, each process outputs VTU file with all variables of all processes.

But since only the last process writes output, everything is written only once, right?

> We may need an option that allows one process only output its own variables in future.

Not sure, because then the grid would be written even more often than it's written now.

wenqing: >> However In the present code, each process outputs VTU file with all variables of all processes.

> But since only the last process writes output, everything is written only once, right?

Yes, it is.


> >We may need an option that allows one process only output its own variables in future.

>Not sure, because then the grid would be written even more often than it's written now.

If it is not needed, I can remove ``struct Output::SingleProcessData`` in another PR.

ProcessLib/Output.cpp
chleh: I see: Writing the output file is skipped if it's not the last process.
Anyway, thanks for your effort.

ProcessLib/Output.cpp
chleh: Yes, a later PR is fine. Maybe, once everything settled, a cleanup of the output logic would be appropriate anyway.

wenqing: Good. Another PR to clean up  the output logic including to remove ``struct Output::SingleProcessData``.

ProcessLib/Output.cpp
wenqing: This loop will be remove after removing `Output::SingleProcessData`, which will be presented in another PR.

Thomas-TK:
Thank you!

PR #2043, opened 2018-01-11
Check header for standalone compilation

bilke:
Added check functionality to compile all headers standalone to check for correct includes.

## Usage

```
$ rm CMakeFiles/CMakeError.log
$ cmake . -DOGS_CHECK_HEADER_COMPILATION=ON
[see CMake output for failing headers]
$ cat CMakeFiles/CMakeError.log # for details what went wrong
```

**OR**

```
make check-header
```

## Status

Top-level libs are checked; no errors on standard config.
Checks are done on Jenkins when `ufz/master` gets build (`Check-Header`-job).

### DIFF
--- a/GeoLib/CMakeLists.txt
+++ b/GeoLib/CMakeLists.txt
@@ -34,5 +34,3 @@ endif()
 
 include(${PROJECT_SOURCE_DIR}/scripts/cmake/packaging/InstallXmlSchemaFiles.cmake)
 InstallXmlSchemaFiles("IO/XmlIO/*.xsd")
-
-#check_header_compilation(GeoLib)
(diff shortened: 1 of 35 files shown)


GeoLib/CMakeLists.txt
```diff
@@ -34,3 +34,5 @@ endif()
 
 include(${PROJECT_SOURCE_DIR}/scripts/cmake/packaging/InstallXmlSchemaFiles.cmake)
 InstallXmlSchemaFiles("IO/XmlIO/*.xsd")
+
+#check_header_compilation(GeoLib)
```
endJunction: ?

PR #2044, opened 2018-01-12
changed deviator trace assertion

Scinopode:
Used diagonal of tensor and fixed norm

### DIFF
--- a/MathLib/KelvinVector-impl.h
+++ b/MathLib/KelvinVector-impl.h
@@ -16,7 +16,7 @@ double Invariants<KelvinVectorSize>::equivalentStress(
     Eigen::Matrix<double, KelvinVectorSize, 1> const& deviatoric_v)
 {
     assert(std::abs(trace(deviatoric_v)) <=
-           1e-16 * std::abs(deviatoric_v.squaredNorm()));
+           5e-14 * diagonal(deviatoric_v).norm());
     return std::sqrt(3 * J2(deviatoric_v));
 }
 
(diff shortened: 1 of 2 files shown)


MaterialLib/SolidModels/KelvinVector-impl.h
```diff
@@ -16,7 +16,7 @@ double Invariants<KelvinVectorSize>::equivalentStress(
     Eigen::Matrix<double, KelvinVectorSize, 1> const& deviatoric_v)
 {
     assert(std::abs(trace(deviatoric_v)) <=
-           1e-16 * std::abs(deviatoric_v.squaredNorm()));
```
chleh: `deviatoric_v` is a column vector, so `deviatoric_v.diagonal()` is only the first entry of that vector. Is that intended?
Does this small check lead to failing tests/simulations and therefore has to be changed?

endJunction: Seems like I made a mistake ;)

MathLib/KelvinVector-impl.h
```diff
@@ -36,16 +36,23 @@ double Invariants<KelvinVectorSize>::J3(
...
+    return v.template topLeftCorner<3, 1>();
+}
+
 /// Trace of the corresponding tensor.
 template <int KelvinVectorSize>
 double Invariants<KelvinVectorSize>::trace(
     Eigen::Matrix<double, KelvinVectorSize, 1> const& v)
 {
-    return v.template topLeftCorner<3, 1>().sum();
+    return diagonal(v).sum();
```
chleh: The former solution might have been more efficient, because `diagonal()` creates a temporary.

endJunction: Good point! I'll check this.

endJunction: Looks like there is no temporary. I made a small setup
```c++
    KelvinVectorType<3> const kv = KelvinVectorType<3>::Random();
    Invariants<6> const invariants;
    volatile double t = invariants.trace(kv);
```
and changed the implementation of trace btw. topLeftCorner and diagonal. There is no difference in the assembly code (in release mode).

I'd guess that because the diagonal method is inline (being a template), or return value optimization works, or both....


Anyway no difference.

If there are no further comments, I'll merge it tonight.

PR #2046, opened 2018-01-15
Fixed Qt GML writer to guarantee decimal output for coordinates

rinkk:


bilke:
Jenkins is fine, don't know whats wrong with envinf1...

PR #2049, opened 2018-01-17
[MPI/PETSc] Enable rank output.

TomFischer:


BaseLib/TemplateLogogFormatterSuppressedGCC-impl.h
```diff
@@ -16,10 +16,7 @@ ::TemplateLogogFormatterSuppressedGCC(MPI_Comm mpi_comm)
 {
     int rank = 0;
     MPI_Comm_rank(mpi_comm, &rank);
-    int size = 0;
-    MPI_Comm_size(mpi_comm, &size);
-    int digits = size/10 + 1;
-    _str_mpi_rank = "rank " + padLeft(std::to_string(rank), digits) + "| ";
+    _str_mpi_rank = "[" + std::to_string(rank) + "] ";
```
chleh: The former code didn't work?

Applications/CLI/ogs.cpp
```diff
@@ -144,6 +145,11 @@ int main(int argc, char *argv[])
                 vtkSmartPointer<vtkMPIController>::New();
             controller->Initialize(&argc, &argv, 1);
             vtkMPIController::SetGlobalController(controller);
+
+            logog_setup.setFormatter(
+                std::make_unique<BaseLib::TemplateLogogFormatterSuppressedGCC<
+                    TOPIC_LEVEL_FLAG | TOPIC_FILE_NAME_FLAG |
+                    TOPIC_LINE_NUMBER_FLAG>>());
```
chleh: Will that automatically nicely format the `[mpi_rank]` line prefixes?

TomFischer: The output looks like:
```
[9] Assemble HTProcess.
[1] Assemble HTProcess.
[3] Assemble HTProcess.
[0] Assemble HTProcess.
```

BaseLib/TemplateLogogFormatterSuppressedGCC-impl.h
TomFischer: I think the padding was not computed correctly. Furthermore, my aim was to reduce the size of the message as much as possible but containing the relevant information.

PR #2050, opened 2018-01-22
LIE_HM fix nodal aperture output

endJunction:
The first commit contains a bug fix for wrong global index access.

Small improvement in a vector access and updating the expected results.


renchao-lu:
I found a similar problem in the nodal output of stress. @endJunction @chleh

endJunction:
@renchao-lu Nice! Would you provide us some more information? And then you could open a new issue.

renchao-lu:
Not quite sure - solution would be to dynamically change the iteration times when performing extrapolateElement in void LocalLinearLeastSquaresExtrapolator::extrapolate() at LocalLinearLeastSquaresExtrapolator.cpp.

I could provide a simple test case with LIE implementation.

PR #2060, opened 2018-01-29
Collect Kelvin vector code and function in single place.

endJunction:
Commit-wise will be faster....

There are still many places having component-wise sigma/epsilon output, I'm aware of this, but it will take another five PRs to replace all of them. (Help is welcome;)


endJunction:
Any ideas why the msvc compiler fails and how to fix it?

endJunction:
The failing build is due to a different Doxygen problem, not related to these changes.

PR #2069, opened 2018-02-08
Doxygen material lib references

endJunction:
for fluid, solid/constitutive_relation, and fracture_model.
Now these tags are expanded in each process yielding better documentation, have a look [here](
https://jenkins.opengeosys.org/job/User/job/endJunction/job/ogs/job/DoxygenMaterialLibReferences_j/Doxygen/ogs_file_param__prj__processes__process__HYDRO_MECHANICS__constitutive_relation__Lubby2.html)
Didn't touch the porous_media because the docu is different in some processes.

### DIFF
--- a/scripts/doc/linked-xml-file.py
+++ b/scripts/doc/linked-xml-file.py
@@ -38,8 +38,6 @@ outdir = os.path.join(docauxdir, "dox", "CTestProjectFiles")
 # "prj__processes__process": "process",
 # See the expansion table in the append-xml-tags.py too.
 tag_path_expansion_table = {
-        "prj__processes__process__constitutive_relation" : "material__solid__constitutive_relation",
-        "prj__processes__process__material_property" : "material",
         "material__porous_medium__porous_medium" : "material__porous_medium",
         }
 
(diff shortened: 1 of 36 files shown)


scripts/doc/linked-xml-file.py
```diff
@@ -40,6 +40,7 @@
 tag_path_expansion_table = {
         "prj__processes__process__constitutive_relation" : "material__solid__constitutive_relation",
         "prj__processes__process__material_property" : "material",
+        "prj__processes__process__fracture_model" : "material__fracture_model",
```
chleh: Is that necessary? I'm wondering if that line matches anything. E.g. there is a parameter
[prj__processes__process__HYDRO_MECHANICS_WITH_LIE__fracture_model](https://jenkins.opengeosys.org/job/User/job/endJunction/job/ogs/job/DoxygenMaterialLibReferences_j/Doxygen/ogs_file_param__prj__processes__process__HYDRO_MECHANICS_WITH_LIE__fracture_model.html),
which does not match the key.
The other entries in the expansion table might be obsolete, too. But maybe I forgot why they are necessary.

endJunction: I think the expansion happened anyway, but it's indeed no longer needed. I was not sure about the `material__porous_medium`, but that whole subtree needs special attention anyway.

Removing the other expansions.

PR #2071, opened 2018-02-11
Integration point data output as VTK's field data

endJunction:
Adding output for integration point data, the general structure and an implementation in small deformation process for sigma. The integration point data stored as field data can be used for restart/initial conditions purpose.

There is metadata associated with the output and it is stored in simple JSON format. The header only json parser is added not as a submodule, because its tests and benchmarks result in 250MiB repo.


endJunction:
@bilke It seems that the windows jenkins' node is sleeping at nights

bilke:
@endJunction I do not know.. I have used subtree years ago and haven't heard much about it since. I am fine with adding the source directly.

endJunction:
@bilke Mac and Conan do not like each other anymore...

PR #2075, opened 2018-02-12
Kelvin vector symmetric tensor conversion

endJunction:
Follow up of #2060 with fixes to the symmetric tensor to/from Kelvin vector conversion, now tested.

This came up during [Integration point data to VTK's field data output PR](https://github.com/ufz/ogs/pull/2071).

### DIFF
--- a/MathLib/KelvinVector.h
+++ b/MathLib/KelvinVector.h
@@ -9,6 +9,7 @@
 #pragma once
 
 #include <Eigen/Dense>
+#include "BaseLib/Error.h"
 
 namespace MathLib
 {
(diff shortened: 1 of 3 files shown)


MathLib/KelvinVector.h
```diff
@@ -143,6 +152,37 @@ kelvinVectorToSymmetricTensor(Eigen::Matrix<double,
...
+/// symmetric 3x3 matrix to a Kelvin vector.
+///
+/// This is opposite of the kelvinVectorToSymmetricTensor()
+///
+/// Only implementations for KelvinVectorSize 4 and 6, and dynamic size vectors
+/// are provided.
+template <typename Derived>
+Eigen::Matrix<double, Eigen::MatrixBase<Derived>::RowsAtCompileTime, 1>
+symmetricTensorToKelvinVector(Eigen::MatrixBase<Derived> const& v)
+{
```
chleh: Maybe add `static_assert(Eigen::MatrixBase<Derived>::ColsAtCompileTime == 1, "ERR");`

MathLib/KelvinVector.h
```diff
@@ -143,6 +152,37 @@ kelvinVectorToSymmetricTensor(Eigen::Matrix<double,
...
+    Eigen::Matrix<double, Eigen::MatrixBase<Derived>::RowsAtCompileTime, 1>
+        result;
+    if (v.size() == 4)
```
chleh: This if-clause also covers dynamically allocated matrices, but don't you have to specify their size manually before filling them with the `<<` operator?

PR #2078, opened 2018-02-14
More analyzers (cppcheck, clang-tidy, iwyu)

bilke:
Added more analyzers to a master-only subjob:

- [cppcheck](http://cppcheck.sourceforge.net)
- [clang-tidy](http://clang.llvm.org/extra/clang-tidy/)
- [iwyu - include what you use](http://include-what-you-use.org/)
- link what you use (integrated into CMake)

See [example output](https://jenkins.opengeosys.org/job/User/job/bilke/job/ogs/job/iwyu_j/19/execution/node/35/log/). Unfortunately the warnings are not reported / summarized somewhere at the moment. CDash implements that and is under investigation ...

### Other changes:

- Use [Conan LIS package](https://github.com/bilke/conan-lis)
- Fixed two reported problems

### To discuss:

iwyu suggests to include 

```
#include <logog/include/macro.hpp>                     // for INFO, WARN
#include <logog/include/message.hpp>                   // for Message
```

instead of of just `logog/include/logog.hpp`. Should we do it like this, otherwise we have to somehow silence this warning.


ThirdParty/.clang-tidy
```diff
@@ -0,0 +1,3 @@
+---
+# Disable all checks, one check has to be enabled (hack)
```
endJunction: why the hack?

endJunction:
"somehow silence." I don't remember the options, but iwyu can be told to use that specific include file instead of the others... Maybe over a config file. Or enforce keeping that include, smth like:
```c++
#include "logog.h" // IWYU_special_keep_this_always.
```

ThirdParty/.clang-tidy
bilke: If there is no enabled check clang-tidy (at least in the way it is invoked by CMake) prints out its usage instructions.

bilke: I did not find another way to specify that this folder should be ignored.

endJunction: I see. I neither do know. ;(
Just add this comment to the "hack", so we are not wondering why...

PR #2079, opened 2018-02-15
Specific times in output

wenqing:
This PR adds a functionality to make output of results at given times. For this purpose, an optional input of specified times is introduced to createOutput, with which the specified times for output can be given in project file. For example
```
        <output>
             <type>VTK</type>
            <prefix>richards</prefix>
            <timesteps>
                <pair>
                    <repeat>1</repeat>
                    <each_steps>100000000</each_steps>
                </pair>
            </timesteps>
            <specified_times> 50.0 100.0 500.</specified_times>
            <output_iteration_results>false</output_iteration_results>
        </output>
```
For computation, the specified times given in the output data are merged to the specified times in the time stepper. One can input the  specified times in any order or even with duplicated data (e.g. by mistakes), however their order is sorted in the descending way and the data are made unique after input. In the computation, the time stepping is conducted at every specified time. So far, the functionality is for the adaptive time stepper, EvolutionaryPIDcontroller.

Test Parabolic/Richards/RichardsFlow_2d_small_adaptive_dt.prj is modified to test the presented functionality.

### DIFF
--- a/NumLib/TimeStepping/Algorithms/EvolutionaryPIDcontroller.h
+++ b/NumLib/TimeStepping/Algorithms/EvolutionaryPIDcontroller.h
@@ -91,6 +91,10 @@ public:
     /// Get a flag to indicate that this algorithm need to compute
     /// solution error.
     bool isSolutionErrorComputationNeeded() override { return true; }
+    /// Add specific times
+    void addSpecificTimes(
+        std::vector<double> const& extra_specific_times) override;
+
 private:
     const double _kP = 0.075;  ///< Parameter. \see EvolutionaryPIDcontroller
     const double _kI = 0.175;  ///< Parameter. \see EvolutionaryPIDcontroller
--- a/NumLib/TimeStepping/Algorithms/EvolutionaryPIDcontroller.h
+++ b/NumLib/TimeStepping/Algorithms/EvolutionaryPIDcontroller.h
@@ -55,7 +55,7 @@ public:
                               const double h0, const double h_min,
                               const double h_max, const double rel_h_min,
                               const double rel_h_max,
-                              const std::vector<double>&& specific_times,
+                              const std::vector<double>&& specified_times,
                               const double tol)
         : TimeStepAlgorithm(t0, t_end),
           _h0(h0),
@@ -63,7 +63,7 @@ public:
           _h_max(h_max),
           _rel_h_min(rel_h_min),
           _rel_h_max(rel_h_max),
-          _specific_times(std::move(specific_times)),
+          _specified_times(std::move(specified_times)),
           _tol(tol),
           _e_n_minus1(0.),
           _e_n_minus2(0.),
@@ -92,8 +92,8 @@ public:
     /// solution error.
     bool isSolutionErrorComputationNeeded() override { return true; }
     /// Add specific times
-    void addSpecificTimes(
-        std::vector<double> const& extra_specific_times) override;
+    void addSpecifiedTimes(
+        std::vector<double> const& extra_specified_times) override;
 
 private:
     const double _kP = 0.075;  ///< Parameter. \see EvolutionaryPIDcontroller
@@ -110,7 +110,7 @@ private:
     const double _rel_h_max;
 
     // Given times that steps have to reach.
-    std::vector<double> _specific_times;
+    std::vector<double> _specified_times;
 
     const double _tol;
 
--- a/NumLib/TimeStepping/Algorithms/EvolutionaryPIDcontroller.h
+++ b/NumLib/TimeStepping/Algorithms/EvolutionaryPIDcontroller.h
@@ -55,7 +55,7 @@ public:
                               const double h0, const double h_min,
                               const double h_max, const double rel_h_min,
                               const double rel_h_max,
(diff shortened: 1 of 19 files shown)


NumLib/TimeStepping/Algorithms/EvolutionaryPIDcontroller.h
```diff
@@ -91,6 +91,10 @@ class EvolutionaryPIDcontroller final : public TimeStepAlgorithm
     /// Get a flag to indicate that this algorithm need to compute
     /// solution error.
     bool isSolutionErrorComputationNeeded() override { return true; }
+    /// Add specific times
+    void addSpecificTimes(
```
endJunction: It would be better to add comment on what this function does aside the info given in the function's name.

NumLib/TimeStepping/Algorithms/TimeStepAlgorithm.h
```diff
@@ -85,6 +85,12 @@ class TimeStepAlgorithm
     /// Get a flag to indicate whether this algorithm needs to compute
     /// solution error. The default return value is false.
     virtual bool isSolutionErrorComputationNeeded() { return false; }
+
+    /// Add specific times
+    virtual void addSpecificTimes(std::vector<double> const& /*specific_times*/)
```
endJunction: Actually the general description would be here, and the overriding function in PID controller would provide additional info...

ProcessLib/Output/Output.cpp
```diff
@@ -70,16 +47,52 @@ int convertVtkDataMode(std::string const& data_mode)
...
+        }
+    }
+
+    bool make_output = timestep % each_steps == 0;
+
+    if (_specific_times.empty())
+        return make_output;
+
+    const double specific_time = _specific_times.back();
+    const double zero_threshold = std::numeric_limits<double>::epsilon();
```
endJunction: epsilon is not zero. :-)

ProcessLib/Output/Output.cpp
```diff
@@ -70,16 +47,52 @@ int convertVtkDataMode(std::string const& data_mode)
...
+
+    return make_output;
+}
+
 Output::Output(std::string output_directory, std::string prefix,
                bool const compress_output, std::string const& data_mode,
                bool const output_nonlinear_iteration_results,
-               std::vector<PairRepeatEachSteps> repeats_each_steps)
+               std::vector<PairRepeatEachSteps> repeats_each_steps,
+               const std::vector<double>&& specific_times)
```
endJunction: Just curious, why the const r-value?

Tests/Data/Parabolic/Richards/RichardsFlow_2d_small_adaptive_dt.prj
```diff
@@ -109,6 +109,7 @@
                     <each_steps>100000000</each_steps>
                 </pair>
             </timesteps>
+            <specific_times> 50.0 100.0 500.</specific_times>
```
endJunction: I'm not sure about the "specific" here. Just my opinion, other comments are welcome.
I'd rather call this `fixed_output_times`, because for me, the "specific" alone does not necessary mean "output" in this context.

NumLib/TimeStepping/Algorithms/EvolutionaryPIDcontroller.h
wenqing: Added a more detailed description.

NumLib/TimeStepping/Algorithms/TimeStepAlgorithm.h
wenqing: Use \copydoc

ProcessLib/Output/Output.cpp
wenqing: Changed to min, while changed the tolerance for  time step size calculation with specified times.

ProcessLib/Output/Output.cpp
wenqing: removed const from there. Just wonder, the code can be compiled with that const.

Tests/Data/Parabolic/Richards/RichardsFlow_2d_small_adaptive_dt.prj
wenqing: Changed the name to `specified times`.

endJunction:
Other ideas additionally to `fixed_output_times` mentioned above would be `forced_output`, `additional_timesteps`, and `additional_output`.

Suggestions are welcome!


`specified` is much better, than `specific`, but the other output times are also "specified", so it could be more clearly named....

Thomas-TK:
Concerning animations and tool coupling we need two things: 
a)  a forced output at user-defined specific times
b) an output with regular time step sizes (e.g. every hour),  but this can be also done by generating lists and using a) I guess. 
Therefore "fixed_output_times" sounds most intuitive to me.

wenqing:
Any other suggestion about the tag name?  If not, `fixed_output_times` will be taken.

endJunction:
@wenqing  Take `fixed_output_times`, rebase, and merge when green.

PR #2080, opened 2018-02-16
[Web] Added a figure about the result comparison of Theis' test

wenqing:
as titled


endJunction:
The new figure looks nice but there it is very difficult to see the differences between the solutions. Especially if the analytical solution is available, please add the difference graphs too.

wenqing:
@endJunction The three curves in the figure override each other.   I added a sentence about the comparison in the md file as
```
The figure shows that there is a good match between the analytical solution and
 the numerical solution obtained by using ogs5 or ogs6.
```

endJunction:
I see, that the solutions overlap and there is a good match. My point was about quantification of the differences.

endJunction:
I'm going to merge it, but I still think that a error plot/diff plot would be helpful to see the actual differences.

PR #2082, opened 2018-02-20
Enable extrapolation in PETSc

endJunction:
Although the extrapolation is not correct for partitioned domains, it is correct for non-partitioned projects and some people are using PETSc for its linear solvers in single threaded setup.

### DIFF
--- a/NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.cpp
+++ b/NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.cpp
@@ -68,6 +68,7 @@ void LocalLinearLeastSquaresExtrapolator::extrapolate(
         extrapolateElement(i, num_components, extrapolatables, t,
                            current_solution, dof_table, *counts);
     }
+    MathLib::LinAlg::finalizeAssembly(*_nodal_values);
 
     MathLib::LinAlg::componentwiseDivide(*_nodal_values, *_nodal_values,
                                          *counts);
@@ -104,6 +105,7 @@ void LocalLinearLeastSquaresExtrapolator::calculateResiduals(
         calculateResidualElement(i, num_components, extrapolatables, t,
                                  current_solution, dof_table);
     }
+    MathLib::LinAlg::finalizeAssembly(*_residuals);
 }
 
 void LocalLinearLeastSquaresExtrapolator::extrapolateElement(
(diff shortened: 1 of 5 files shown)


wenqing:
As a temporal solution,  you can change `#ifndef USE_PETSC` to `#ifdef USE_PETSC` with a constraint  as
```
#ifdef USE_PETSC
int petsc_np;
MPI_comm_size(PETSC_COMM_WORLD,  petsc_np);
if (petsc_np == 1)
{
#endif // end of #ifdef USE_PETSC`

...
piece of the source code for sequential computing
...

#ifdef USE_PETSC
}
#endif
```

endJunction:
Good point, Wenqing. I'll come back to this if the more correct implementation turns out to be more complex...

endJunction:
OK. Seems to be working, at least what is tested.
There is a problem with residuals which shows up in the TES-mpi ctest. The problem is about the access of values w/o finalize, or copy before finalize calls. I'd not trust the residual output for petsc now.

endJunction:
Some of the changes are just cleanup. see https://github.com/ufz/ogs/pull/2085

NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.cpp
```diff
@@ -47,12 +47,39 @@ void LocalLinearLeastSquaresExtrapolator::extrapolate(
 {
     auto const num_nodal_dof_result =
         _dof_table_single_component.dofSizeWithoutGhosts() * num_components;
+
+    std::vector<GlobalIndexType> ghost_indices;
```
wenqing: This is only for DDC. How about
```cpp
#ifndef USE_PETSC
    if (!_nodal_values ||  _nodal_values->size() != num_nodal_dof_result)
    {
         _nodal_values = MathLib::MatrixVectorTraits<GlobalVector>::newInstance(
            MathLib::MatrixSpecifications{
                num_nodal_dof_result, num_nodal_dof_result, nullptr, nullptr});         
    }
#else // For USE_PETSC
    std::vector<GlobalIndexType> ghost_indices;
    {  // Create num_components times version of ghost_indices arranged by
       // location. For example for 3 components and ghost_indices {5,6,10} we
       // compute {15, 16, 17,  18, 19, 20,  30, 31, 32}.
        auto const& single_component_ghost_indices =
            _dof_table_single_component.getGhostIndices();
        auto const single_component_ghost_indices_size =
            single_component_ghost_indices.size();
        ghost_indices.reserve(single_component_ghost_indices_size *
                              num_components);
        for (unsigned i = 0; i < single_component_ghost_indices_size; ++i)
        {
            for (unsigned c = 0; c < num_components; ++c)
            {
                ghost_indices.push_back(
                    single_component_ghost_indices[i] * num_components + c);
            }
        }
    }

     if (!_nodal_values ||  static_cast<std::size_t>(_nodal_values->getLocalSize() +
                                 _nodal_values->getGhostSize())
            != num_nodal_dof_result)
     {
         _nodal_values = MathLib::MatrixVectorTraits<GlobalVector>::newInstance(
           {num_nodal_dof_result, num_nodal_dof_result, &ghost_indices,
            nullptr});
     }
#endif
```

PR #2088, opened 2018-02-26
[Web] Migrated devguide website generator

bilke:
I have migrated the content of https://docs.opengeosys.org/docs/devguide to the website generator which is also used for the benchmark docs.

Once merged I will setup a redirection from the old to the new devguide.

After merge I will also migrate remaining content.


web/content/docs/devguide/development-workflows/branching-model.pandoc
```diff
@@ -0,0 +1,36 @@
...
+>
+> The main advantage of the Forking Workflow is that contributions can be integrated without the need for everybody to push to a single central repository. Developers push to their own server-side repositories, and only the project maintainer can push to the official repository. This allows the maintainer to accept commits from any developer without giving them write access to the official codebase.
+>
+> The result is a distributed workflow that provides a flexible way for large, organic teams (including untrusted third-parties) to collaborate securely. This also makes it an ideal workflow for open source projects.
+> <cite> https://www.atlassian.com/git/workflows#!workflow-forking </cite>
+>
+
+The workflow is summarized in the following image from the [GitHub blog](https://github.com/blog/2042-git-2-5-including-multiple-worktrees-and-triangular-workflows):
+
+![](https://cloud.githubusercontent.com/assets/1319791/8943755/5dcdcae4-354a-11e5-9f82-915914fad4f7.png)
```
endJunction: Just a notice; This image is not shown in the jenkins' artifacts, although it is a valid url pointing to right image.

bilke: Yes this is security feature of Jenkins. Only white-listed things can be fetched from remote locations. FYI the white-list is defined [here](https://gitlab.opengeosys.org/bilke/ogs-ansible/blob/master/roles/jenkins/defaults/main.yml#L4).

PR #2091, opened 2018-03-01
Small updates to coupled solutions

endJunction:
... discovered during staggered PhaseField process implementation.

### DIFF
--- a/ProcessLib/UncoupledProcessesTimeLoop.cpp
+++ b/ProcessLib/UncoupledProcessesTimeLoop.cpp
@@ -803,13 +803,16 @@ bool UncoupledProcessesTimeLoop::solveCoupledEquationSystemsByStaggeredScheme(
             {
                 MathLib::LinAlg::axpy(x_old, -1.0, x);  // save dx to x_old
                 if (process_id == last_process_id)
+                {
                     INFO(
                         "------- Checking convergence criterion for coupled "
                         "solution  -------");
-                _global_coupling_conv_crit[process_id]->checkDeltaX(x_old, x);
-                coupling_iteration_converged =
-                    coupling_iteration_converged &&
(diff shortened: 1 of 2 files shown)


ProcessLib/UncoupledProcessesTimeLoop.cpp
```diff
@@ -736,6 +736,7 @@ bool UncoupledProcessesTimeLoop::solveCoupledEquationSystemsByStaggeredScheme(
         bool nonlinear_solver_succeeded = true;
         coupling_iteration_converged = true;
         int process_id = 0;
+        int last_process_id = _per_process_data.size() - 1;
```
TomFischer: `last_process_id` could be `const`.

ProcessLib/UncoupledProcessesTimeLoop.cpp
```diff
@@ -838,6 +846,12 @@ bool UncoupledProcessesTimeLoop::solveCoupledEquationSystemsByStaggeredScheme(
             ++process_id;
             continue;
         }
+        CoupledSolutionsForStaggeredScheme coupled_solutions(
+            _solutions_of_coupled_processes, dt, process_id);
+
+        process_data->process.setCoupledSolutionsForStaggeredScheme(
+            &coupled_solutions);
```
TomFischer: `coupled_solutions` is a locally created object. I wonder if giving the address of a local created object to the method `setCoupledSolutionsForStaggeredScheme()` is a good idea?

endJunction: Exactly this is the bug, unfortunately. The reference can be easily stolen from the `process_data`.
The same code is in the assemble loop and was missing in the post timestep part.

PR #2097, opened 2018-03-12
CTest Fix

bilke:
Fixes missing benchmark runs.

Benchmarks where missing because of:

- `time`-wrapper was defined
- `time` command is not found on all platforms correctly (i.e. when it is implemented as a shell keyword)
- when `time` was not found the benchmark was silently disabled

Now a warning is printed and the wrapper is simply disabled but the benchmark is still run.


endJunction:
@bilke One of the problems for vtkdiff calls on windows is ["The command line is too long."](https://ci.appveyor.com/project/ogs/ogs/build/6.0.8.1602#L7369)

endJunction:
@bilke Aborted jenkins because win build hung there for 2h at 
```
cmd.exe /C "cd /D E:\jenkins2\workspace\7-JXVE3G3AAA@2\build && "C:\Program Files\CMake\bin\cmake.exe" -E remove Tests/ctest.log"
```

bilke:
@endJunction Ready for merge!

PR #2104, opened 2018-03-19
Stronger gravity input check liquid flow

endJunction:


bilke:
@endJunction In `.appveyor.yml` can you change line 62 to:

```yaml
  - cmake --build . --config %configuration% --target package
```

endJunction:
@bilke Unfortunately it is not sufficient to produce an "artifact".

endJunction:
@HBShaoUFZ I'm not sure about the build-up test, for which the comparison fails on one of the windows builds (on appveyor) but not on the other (on jenkings), when keeping the vtkdiff tolerances as in ufz/ogs/master. Relaxing them to 1e-4 (absolute) makes the test pass.
What shall we do?
 - Easy way is to set the tolerances to 1e-4---would you be happy with it? Because I'm not very satisfied.
 - Start investigating why the build on a particular machine produces so different result in this case. The difficulty is to reproduce the build somewhere locally.

@bilke Do you have another trick on how to get the build binary (and libraries if shared)?

PR #2107, opened 2018-03-23
Separate processes

endJunction:
- Introduce a cmake option (by default ON) for building each of the processes.
 - Separate the ProcessLib.so into multiple libraries, one for each process.

There are two questions below, please continue reading ;)

## Some timings
Procedure: on a given branch build ogs, remove shared libraries, measure build time of ogs target. Repeat the last two for three times and take the minimum total time;
```sh
ninja ogs
for i in `seq 3`; do rm lib/*.so && time ninja -j6 ogs; done
```

branch | threads | 1 | 2 | 3 | 4 | 5 | 6
--------- | ---------- | - | - | - | - | - | -
master | time (s) | **6.179** | 6.179 | 6.114 | 6.218 | 6.147 | 6.194
this PR | time (s) | 8.386 | **6.097** | 5.220 | 4.751 | 4.618 | **4.571**

So single threaded the master is faster, but does not scale at all. This PR is slower, but scales OK up to three threads.

**Question** to all: Do you like a single shared library, or multiple libraries, one per process?

### Compilation times

Compilation time reduces, of course, if only the required process(es) are selected. But changing the selection of processes introduce a recompilation of all object files because the definitions are passed on command line (ccache fails too). 

**Question**, Is it possible to avoid the recompilation?

If the process selection would be collected in single header file included in ApplicationsLib, then the recompilation would not touch for example the MathLib library (and other unrelated libraries). ccache still would not use the cached code, though.
Other ideas?

<sup><sup>You can peek up a traditional beverage in my office if you read this far and answered both questions</sup></sup>


CMakeLists.txt
```diff
@@ -85,6 +85,90 @@ if(OGS_BUILD_GUI)
 endif()
 option(OGS_BUILD_UTILS "Should the utilities programms be built?" OFF)
 
+
+# A list of processes to be build.
+option(OGS_BUILD_PROCESS_ComponentTransport "Build the component transport process." ON)
+if(OGS_BUILD_PROCESS_ComponentTransport)
+    add_definitions(-DOGS_BUILD_PROCESS_COMPONENTTRANSPORT)
```
endJunction: @bilke Shall I make the cmake option capital case too, same as the c-definitions? All capital case are harder to read, but would be consistent with the c-definitions.

bilke: I do not know whats best but for me the mixed case you did is better to read indead..

bilke:
My two answers to get the beverage:

1. I like one lib per process.
2. Limit the scope of the definition! See below for explanation.

Add the definition (e.g. `OGS_BUILD_PROCESS_COMPONENTTRANSPORT`) to just the place where it is needed. In this case `ProjectData.cpp`. In `Applications/ApplicationsLib/CMakeLists.txt` (not tested):

```cmake
if(OGS_BUILD_PROCESS_ComponentTransport)
    set_source_files_properties(ProjectData.cpp PROPERTIES
        COMPILE_DEFINITIONS -DOGS_BUILD_PROCESS_COMPONENTTRANSPORT)
endif()
...
```

CMakeLists.txt
endJunction: OK. Keeping mixed case.

endJunction:
@bilke All suggestions implemented!

@ all: Only one place for process names. Compile definitions for ProjectData.cpp only. Let's hope it compiles, then it can be merged.
I didn't test all of the possible combinations of on/off processes, so there could be some corner cases like the one I updated for the LIE tests and utils.

endJunction:
Working! The usual "build-up" test on appveyour is failing....

PR #2108, opened 2018-03-24
correction of density calculation in thermo-mechanics

xingyuanmiao:
As titled. This correction has been done in thermo-mechanical phase-field process.

### DIFF
--- a/Documentation/ProjectFile/prj/processes/process/THERMO_MECHANICS/t_reference_solid_density.md
+++ b/Documentation/ProjectFile/prj/processes/process/THERMO_MECHANICS/t_reference_solid_density.md
@@ -1,1 +1,1 @@
-Reference solid density. The density changes from this reference value.
+Reference solid density. The real density is expressed as: rho_s = rho_sr / (1 + 3 * alpha * dT ) under the assumption of incompressible solid, i.e. for an incompressible solid: dV / dp --> 0 (cf. Tykodi, R. J. (1991). Thermodynamics of an incompressible solid and a thermodynamic functional determinant. Journal of Chemical Education, 68(10), 830.).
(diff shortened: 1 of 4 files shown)


wenqing:
The corresponding benchmark failed.

wenqing:
@xingyuanmiao 
I just requested a change. Please don't make change before @endJunction  and @nagelt review it.

endJunction:
Setting WIP until the requested documentation is provided.

xingyuanmiao:
which kind of documentation should be provided?

xingyuanmiao:
The derivation of the calculation of real density is added in the code and documentation.

endJunction:
As mentioned before, please add the incompressibility assumption to the formulae. There must be some text describing the formulae...

> The computation of solid's density is correct under the assumption of incompressible solid grains. This **must** be mentioned in the code and documentation of the `rho_sr` parameter. Also the derivation of this formula is to be included.

xingyuanmiao:
provided a reference to explain the formula.

endJunction:
I, unfortunately, don't see how the formula is explained in the referenced paper, there is even no discussion on density computation. Furthermore, the paper shows that the assumption of solid's "incompressibility implies thermal inexpansibility" (see eq. 4 there). The latter is not the point of TM implementation.

The explanation of the solid's density computation is still to be explained.

For the references, please use `\cite` and proper bibtex entries.

nagelt:
The original Ansatz made is \rho_{SR} = \rho_{SR0} exp (- 3 \alpha_T \Delta T) for large thermal strains. Once we linearize around \Delta T = 0 and truncate the Taylor series after the first member, the end result is \rho_{SR} = \rho_{SR0} / ( 1 + 3 \alpha_T \Delta T). You can also arrive at the result by assuming a linear volume strain from the start: e = 3 \alpha_T \Delta T. The key assumption is that the solid density is independent of pressure, i.e. \rho_{SR} = \rho_{SR}(T).

PR #2114, opened 2018-04-20
TecPlot utility

rinkk:
Currently can

* split TecPlot files containing multiple "zones" (i.e. timesteps, subdomains, etc.) into seperate tecplot files
* convert TecPlot files representing raster data into ogs-meshes (one file per zone, containing all variables as scalar arrays)

### DIFF
--- a/Applications/Utils/FileConverter/TecPlotTools.cpp
+++ b/Applications/Utils/FileConverter/TecPlotTools.cpp
@@ -19,12 +19,14 @@
 #include "BaseLib/BuildInfo.h"
 #include "BaseLib/StringTools.h"
 
+#include "MeshLib/IO/VtkIO/VtuInterface.h"
 #include "MeshLib/Mesh.h"
 #include "MeshLib/MeshGenerators/RasterToMesh.h"
-#include "MeshLib/IO/VtkIO/VtuInterface.h"
 
 /// Returns the value for the given parameter name (i.e. for "x = 3" it returns "3")
-std::string getValue(std::string const& line, std::string const& val_name, bool is_string)
+std::string getValue(std::string const& line,
+                     std::string const& val_name,
+                     bool is_string)
 {
     std::string value;
     std::size_t start(line.find(val_name));
@@ -82,9 +84,9 @@ std::string trimVariable(std::string& var)
 /// Returns an vector of variable names from the "Variables"-description
 std::vector<std::string> getVariables(std::string const& line)
 {
-    std::string const var_str ("VARIABLES");
+    std::string const var_str("VARIABLES");
     std::size_t start(line.find(var_str));
-    std::string all_vars = line.substr(start+var_str.length(), std::string::npos);
+    std::string all_vars = line.substr(start + var_str.length(), std::string::npos);
     start = all_vars.find("=");
     all_vars = all_vars.substr(start + 1, std::string::npos);
     BaseLib::trim(all_vars);
@@ -96,7 +98,7 @@ std::vector<std::string> getVariables(std::string const& line)
         end = all_vars.find_first_of(" ");
         std::string var = all_vars.substr(0, end);
         variables.push_back(trimVariable(var));
-        all_vars = all_vars.substr(end+1, std::string::npos);
+        all_vars = all_vars.substr(end + 1, std::string::npos);
         BaseLib::trim(all_vars);
     }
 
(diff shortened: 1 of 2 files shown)


Applications/Utils/FileConverter/TecPlotTools.cpp
```diff
@@ -0,0 +1,410 @@
...
+/// Splits a TecPlot file containing multiple sections/zones into seperate files
+int splitFile(std::ifstream& in, std::string file_name)
+{
+    std::ofstream out;
+    std::string line("");
+    std::string name;
+    std::pair<std::size_t, std::size_t> dims(0,0);
+    std::size_t val_count(0), val_total(0);
+    std::size_t write_count(0);
+    while (getline(in, line))
```
TomFischer: `std::getline`?

Applications/Utils/FileConverter/TecPlotTools.cpp
```diff
@@ -0,0 +1,410 @@
...
+            (line.find("VARIABLES") != std::string::npos) ||
+            (line.find("ZONE") != std::string::npos))
+            return;
+    }
+}
+
+/// Splits a TecPlot file containing multiple sections/zones into seperate files
+int splitFile(std::ifstream& in, std::string file_name)
+{
+    std::ofstream out;
```
TomFischer: Where is `out` initialized? What is the purpose of `out`?

Applications/Utils/FileConverter/TecPlotTools.cpp
```diff
@@ -0,0 +1,410 @@
...
+    std::string line("");
+    std::string name;
```
TomFischer: The created `name` object is empty. It is used, for instance, in line 245 and 252, but will be initialized in line 262. Could this be a problem?

Applications/Utils/FileConverter/TecPlotTools.cpp
```diff
@@ -0,0 +1,410 @@
...
+    std::pair<std::size_t, std::size_t> dims(0,0);
```
TomFischer: `dims` could be more local, it is used as a temporary variabele while parsing the zone.

Applications/Utils/FileConverter/TecPlotTools.cpp
```diff
@@ -0,0 +1,410 @@
...
+            val_total = dims.first * dims.second;
+            val_count = 0;
+            continue;
+        }
+
+        out << line << "\n";
+        val_count++;
+    }
+    if (dataCountError(out, name, val_count, val_total))
+        return -3;
```
TomFischer: Why `-3`? It seems the method could return a bool  value.

Applications/Utils/FileConverter/TecPlotTools.cpp
```diff
@@ -0,0 +1,410 @@
...
+            continue;
+        }
+
+        double x;
+        std::stringstream iss(line);
+        std::size_t i(0);
+        std::size_t const n_scalars (scalars.size());
+        while (iss >> x)
+        {
+            if (i > n_scalars-1)
```
TomFischer: Is it possible that `n_scalars` is zero at this point? I'm not sure, but `n_scalars - 1` is probably very large. Is this okay?

Applications/Utils/FileConverter/TecPlotTools.cpp
rinkk: :heavy_check_mark:

Applications/Utils/FileConverter/TecPlotTools.cpp
rinkk: ```out``` is initalised / switched to the next file / closed in ```writeTecPlotSection()```

Applications/Utils/FileConverter/TecPlotTools.cpp
rinkk: removed all instances ```("")```

Applications/Utils/FileConverter/TecPlotTools.cpp
rinkk: Different values indicate different reasons for failing. ```-3``` is indicating an index error.

Applications/Utils/FileConverter/TecPlotTools.cpp
rinkk: Barring an invalid TecPlot-file, ```n_scalars``` must be larger than ```0```, otherwise the ```zone```-section would contain no data.

Applications/Utils/FileConverter/TecPlotTools.cpp
rinkk: I put an explanation on the exit codes in the documentation. Independent of that, a detailed error-message is output in each case.

PR #2115, opened 2018-04-26
Extension of element removal

rinkk:
* works with arbitrary scalar arrays now (currently limited to int + double)
* allows inclusion and exclusion of value ranges

### DIFF
--- a/Applications/DataExplorer/DataView/MeshElementRemovalDialog.cpp
+++ b/Applications/DataExplorer/DataView/MeshElementRemovalDialog.cpp
@@ -27,7 +27,6 @@
 #include "MeshLib/MeshSearch/ElementSearch.h"
 #include "MeshLib/Node.h"
 #include "MeshLib/Properties.h"
-#include "MeshLib/PropertyVector.h"
 
 /// Constructor
 MeshElementRemovalDialog::MeshElementRemovalDialog(
@@ -93,8 +92,15 @@ void MeshElementRemovalDialog::accept()
             min_val = this->insideScalarMinEdit->text().toDouble();
             max_val = this->insideScalarMaxEdit->text().toDouble();
         }
-        ex.searchByPropertyValueRange(array_name, min_val, max_val, outside);
-        anything_checked = true;
+        std::size_t n_marked_elements =
+            ex.searchByPropertyValueRange<double>(array_name, min_val, max_val, outside);
+
+        if (n_marked_elements == 0)
+            n_marked_elements =
+                ex.searchByPropertyValueRange<int>(array_name, min_val, max_val, outside);
+
+        if (n_marked_elements > 0)
+            anything_checked = true;
     }
     if (this->boundingBoxCheckBox->isChecked())
     {
@@ -157,11 +163,11 @@ std::size_t MeshElementRemovalDialog::addScalarArrays(MeshLib::Mesh const& mesh)
 {
     MeshLib::Properties const& properties = mesh.getProperties();
     std::vector<std::string> const& names = properties.getPropertyVectorNames();
-    for (auto name : names)
+    for (auto const& name : names)
     {
         if (properties.existsPropertyVector<int>(name))
         {
-            auto const p = properties.getPropertyVector<int>(name);
+            auto const& p = properties.getPropertyVector<int>(name);
             if (p->getMeshItemType() == MeshLib::MeshItemType::Cell)
             {
                 this->scalarArrayComboBox->addItem(QString::fromStdString(name));
@@ -170,7 +176,7 @@ std::size_t MeshElementRemovalDialog::addScalarArrays(MeshLib::Mesh const& mesh)
         }
         if (properties.existsPropertyVector<double>(name))
         {
-            auto const p = properties.getPropertyVector<double>(name);
+            auto const& p = properties.getPropertyVector<double>(name);
             if (p->getMeshItemType() == MeshLib::MeshItemType::Cell)
             {
                 this->scalarArrayComboBox->addItem(QString::fromStdString(name));
@@ -202,10 +208,7 @@ void MeshElementRemovalDialog::toggleScalarEdits(bool outside) const
 
 void MeshElementRemovalDialog::on_insideButton_toggled(bool is_checked)
 {
-    if (this->insideButton->isChecked())
-        toggleScalarEdits(false);
-    else
-        toggleScalarEdits(true);
+    toggleScalarEdits(!this->insideButton->isChecked());
 }
 
 void MeshElementRemovalDialog::on_boundingBoxCheckBox_toggled(bool is_checked)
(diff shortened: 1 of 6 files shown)


Applications/DataExplorer/DataView/MeshElementRemovalDialog.cpp
```diff
@@ -134,6 +153,61 @@ void MeshElementRemovalDialog::reject()
     this->done(QDialog::Rejected);
 }
 
+std::size_t MeshElementRemovalDialog::addScalarArrays(MeshLib::Mesh const& mesh) const
+{
+    MeshLib::Properties const& properties = mesh.getProperties();
+    std::vector<std::string> const& names = properties.getPropertyVectorNames();
+    for (auto name : names)
```
endJunction: a const ref would save some copies here: `auto const& name ...`

Applications/DataExplorer/DataView/MeshElementRemovalDialog.cpp
```diff
@@ -134,6 +153,61 @@ void MeshElementRemovalDialog::reject()
...
+    this->insideScalarMinEdit->setEnabled(!outside);
+    this->insideScalarMaxEdit->setEnabled(!outside);
+}
+
+void MeshElementRemovalDialog::on_insideButton_toggled(bool is_checked)
+{
+    if (this->insideButton->isChecked())
+        toggleScalarEdits(false);
+    else
+        toggleScalarEdits(true);
```
endJunction: shorter version, and IMO also more readable:
```c++
toggleScalarEdits(insideButton->isChecked());
```

Applications/DataExplorer/DataView/MeshElementRemovalDialog.cpp
```diff
@@ -202,24 +262,48 @@ void MeshElementRemovalDialog::on_meshNameComboBox_currentIndexChanged(int idx)
...
-        }
+    this->scalarArrayComboBox->clear();
+    this->outsideScalarMinEdit->setText("");
+    this->outsideScalarMaxEdit->setText("");
+    this->insideScalarMinEdit->setText("");
+    this->insideScalarMaxEdit->setText("");
+    if (this->scalarArrayCheckBox->isChecked())
+        on_scalarArrayCheckBox_toggled(true);
+    if (this->boundingBoxCheckBox->isChecked())
+        on_boundingBoxCheckBox_toggled(true);
```
endJunction: same applies here:
```c++
on_boundingBoxCheckBox_toggled(boundingBoxCheckBox->isChecked());
```

Applications/DataExplorer/DataView/MeshElementRemovalDialog.cpp
```diff
@@ -202,24 +262,48 @@ void MeshElementRemovalDialog::on_meshNameComboBox_currentIndexChanged(int idx)
...
+}
+
+void MeshElementRemovalDialog::on_scalarArrayComboBox_currentIndexChanged(int idx)
+{
+    Q_UNUSED(idx);
+    MeshLib::Mesh const* const mesh =
+        _project.getMesh(meshNameComboBox->currentText().toStdString());
+    MeshLib::Properties const& properties = mesh->getProperties();
```
endJunction: The mesh pointer is not checked before access, could be nullptr.

Applications/DataExplorer/DataView/MeshElementRemovalDialog.cpp
```diff
@@ -202,24 +262,48 @@ void MeshElementRemovalDialog::on_meshNameComboBox_currentIndexChanged(int idx)
...
 
-        materialIDCheckBox->setEnabled(true);
-        if (materialIDCheckBox->isChecked())
-            on_materialIDCheckBox_toggled(true);
+    std::string const vec_name(scalarArrayComboBox->currentText().toStdString());
+    if (vec_name == "")
```
endJunction:
```c++
if (vec_name.empty())
```

Applications/DataExplorer/DataView/MeshElementRemovalDialog.cpp
```diff
@@ -202,24 +262,48 @@ void MeshElementRemovalDialog::on_meshNameComboBox_currentIndexChanged(int idx)
...
+    else if (properties.existsPropertyVector<double>(vec_name))
+    {
+        MeshLib::PropertyVector<double> const& vec =
+            *properties.getPropertyVector<double>(vec_name);
+        auto min = std::min_element(vec.cbegin(), vec.cend());
+        auto max = std::max_element(vec.cbegin(), vec.cend());
+        this->outsideScalarMinEdit->setText(QString::number(*min));
+        this->outsideScalarMaxEdit->setText(QString::number(*max));
+        this->insideScalarMinEdit->setText(QString::number(*min));
+        this->insideScalarMaxEdit->setText(QString::number(*max));
```
endJunction: The `int` and `double` versions are equal up to the type. A lambda would remove the code duplication easily.
```c++
template <typename T> auto do_something = [&]() {
   MeshLib::PropertyVector<T> ...
}
```

Applications/Utils/MeshEdit/removeMeshElements.cpp
```diff
@@ -28,26 +28,46 @@
...
+void searchByPropertyValue(std::string const& property_name,
+                           std::vector<PROPERTY_TYPE> const& property_values,
+                           MeshLib::ElementSearch& searcher)
 {
-    for (auto const& property_value : property_values) {
-        const std::size_t n_marked_elements =
-            searcher.searchByPropertyValue(property_value, property_name);
+    for (auto const& property_value : property_values)
+    {
+        std::size_t n_marked_elements = searcher.searchByPropertyValue<double>(
```
endJunction: const

MeshLib/MeshSearch/ElementSearch.h
```diff
@@ -34,18 +34,43 @@ class ElementSearch final
...
-        PROPERTY_TYPE const property_value,
-        std::string const& property_name)
+    std::size_t searchByPropertyValue(std::string const& property_name,
+                                      PROPERTY_TYPE const property_value)
+    {
+        return searchByPropertyValueRange<PROPERTY_TYPE>(
+            property_name, property_value, property_value, false);
+    }
+
+    /// @tparam PROPERTY_TYPE integral type of the property
```
endJunction: I'm not sure I understand what you write here. What do you mean by "integral type"? I'm asking because in the above implementation you added `double` and `int`, where the former is not an "integral type".

MeshLib/MeshSearch/ElementSearch.h
```diff
@@ -34,18 +34,43 @@ class ElementSearch final
...
+    /// property value outside of the interval [min_property_value,
+    /// max_property_value].
+    /// @param property_name the name of the property the searching/marking is
+    /// based on
+    /// @param min_property_value minimum value of the given property for the
+    /// element not to be marked
+    /// @param max_property_value maximum value of the given property for the
+    /// element not to be marked
+    /// @param outside_of if true, all values outside of the given range or
+    /// markedc, if false, all values inside the given range are marked
```
endJunction: typo in the first "marked"

MeshLib/MeshSearch/ElementSearch.h
```diff
@@ -34,18 +34,43 @@ class ElementSearch final
...
+    /// all elements of the mesh for the property \c property_name with  a
+    /// property value outside of the interval [min_property_value,
+    /// max_property_value].
+    /// @param property_name the name of the property the searching/marking is
+    /// based on
+    /// @param min_property_value minimum value of the given property for the
+    /// element not to be marked
+    /// @param max_property_value maximum value of the given property for the
+    /// element not to be marked
+    /// @param outside_of if true, all values outside of the given range or
```
endJunction: the default value is not documented. I'd even recommend to remove the default value and always explicitly write the intentions.

MeshLib/MeshSearch/ElementSearch.h
```diff
@@ -63,12 +88,23 @@ class ElementSearch final
...
-                matchedIDs.push_back(i);
+            for (std::size_t i(0); i < pv->getNumberOfTuples(); ++i)
+            {
+                if ((*pv)[i] < min_property_value || (*pv)[i] > max_property_value)
+                    matchedIDs.push_back(i);
+            }
+        }
+        else
+        {
+            for (std::size_t i(0); i < pv->getNumberOfTuples(); ++i)
```
endJunction: This is `std::copy_if`

Applications/DataExplorer/DataView/MeshElementRemovalDialog.cpp
rinkk: Added the nullptr-check (although if there was a problem the programme would've crashed long before this point), but I'm not sure what you mean with your second remark. 

The ```if```` applies if the currently selected mesh has no scalar arrays. In this case, the combobox will be empty and therefore the returned string will be empty. No range values can be computed and all related widgets will remain disabled.

If the has *does* have scalar arrays, ```vec_name``` cannot be empty, thus range values are computed, widgets are enabled, etc.

Does that answer your comment?

Applications/DataExplorer/DataView/MeshElementRemovalDialog.cpp
rinkk: :heavy_check_mark:

Applications/Utils/MeshEdit/removeMeshElements.cpp
rinkk: :heavy_check_mark:

Applications/Utils/MeshEdit/removeMeshElements.cpp
```diff
@@ -28,26 +28,46 @@
...
+            property_name, property_value);
+        if (n_marked_elements == 0)
+            std::size_t n_marked_elements = searcher.searchByPropertyValue<int>(
```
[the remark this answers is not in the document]
rinkk: ```searchByPropertyValue``` just calls ```searchByPropertyValueRange```. I've just put this here for backwards compatibility because searching for a specific value was used in a few tests IIRC.

Additional types would need to be implemented her by adding a line with the respective type. The search-method itself just checks if the value for each element is smaller/larger than the range bounds. We just need to test the type because the property vector won't be returned without knowing the correct type.

The order of type-tests shouldn't matter because only one of them can ever return a nonzero value. All the wrong types aren't tested at all because they return a nullptr (or rather a boost optional, but it's essentially the same result) and thus will automatically return zero.

Applications/Utils/MeshEdit/removeMeshElements.cpp
```diff
@@ -116,9 +154,52 @@ int main (int argc, char* argv[])
...
+                (outside_property_arg.isSet() && inside_property_arg.isSet()))
+            {
+                ERR("Specify if the inside or the outside of the selected "
+                    "range should be removed.");
+                return EXIT_FAILURE;
+            }
+
+            bool outside = false;
+            if (outside_property_arg.isSet())
+                outside = true;
```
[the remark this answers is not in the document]
rinkk: :heavy_check_mark:

MeshLib/MeshSearch/ElementSearch.h
rinkk: I have no idea, actually. I just copied the comment from the previous method, because the range-search basically replaces the (single-)value search. I have no idea who wrote that comment originally.

I've removed the word "integral", is that okay with you?

MeshLib/MeshSearch/ElementSearch.h
rinkk: `:heavy_check_mark:

MeshLib/MeshSearch/ElementSearch.h
endJunction: Ah, the index ;)
Sorry, then the copy_if is not applicable. It looked so similar to `push_back(*pv[i])`...

MeshLib/MeshSearch/ElementSearch.h
endJunction: :heavy_check_mark:

Applications/DataExplorer/DataView/MeshElementRemovalDialog.cpp
endJunction: The second remark is only for the order of the lines; Should be
```c++
if (vec_name.empty()) return;
auto mesh = getMesh();
auto properties = getProperties(mesh);
```
The `properties` are not needed in case the `vec_name` is empty...

rinkk: I see. Fixed! :heavy_check_mark:

MeshLib/MeshSearch/ElementSearch.h
rinkk: Removed the default value.

PR #2117, opened 2018-05-06
Move process implementations into cpp

endJunction:
Merge unneeded file separation.

No changes, only move the -impl.h into the corresponding .cpp.


endJunction:
@chleh Not easy to generalize but in this particular context the cpp contains only the -impl.h include and the explicit instantiations; there is no advantage to separate the two explicit instantiations (plus the usual copyright, include, and namespace overhead).

PR #2118, opened 2018-05-07
Generalize RHS output

endJunction:
Output of nodal forces (already implemented in SmallDeformation) is extended to other processes (HM, LIE/HM) and generalized to other RHSs (for pressure or displacement jump).
In HM the RHS of p corresponds to the nodal volume fluxes, for example.


ProcessLib/HydroMechanics/Tests.cmake
```diff
@@ -10,18 +10,11 @@ AddTest(
...
-    expected_square_1e2_pcs_0_ts_1_t_5.000000.vtu square_1e2_pcs_0_ts_1_t_5.000000.vtu pressure pressure 1e-15 1e-15
-    expected_square_1e2_pcs_0_ts_20_t_100.000000.vtu square_1e2_pcs_0_ts_20_t_100.000000.vtu pressure pressure 1e-15 1e-15
-    expected_square_1e2_pcs_0_ts_120_t_1000.000000.vtu square_1e2_pcs_0_ts_120_t_1000.000000.vtu pressure pressure 1e-15 1e-15
-    expected_square_1e2_pcs_0_ts_420_t_4000.000000.vtu square_1e2_pcs_0_ts_420_t_4000.000000.vtu pressure pressure 1e-15 1e-15
-    expected_square_1e2_pcs_0_ts_1_t_5.000000.vtu square_1e2_pcs_0_ts_1_t_5.000000.vtu velocity velocity 1e-15 1e-15
-    expected_square_1e2_pcs_0_ts_20_t_100.000000.vtu square_1e2_pcs_0_ts_20_t_100.000000.vtu velocity velocity 1e-15 1e-15
-    expected_square_1e2_pcs_0_ts_120_t_1000.000000.vtu square_1e2_pcs_0_ts_120_t_1000.000000.vtu velocity velocity 1e-15 1e-15
-    expected_square_1e2_pcs_0_ts_420_t_4000.000000.vtu square_1e2_pcs_0_ts_420_t_4000.000000.vtu velocity velocity 1e-15 1e-15
+    GLOB square_1e2_pcs_0_ts_*.vtu displacement displacement 1e-15 1e-15
```
chleh: Globbing is possible here? I guess I missed that PR.
Maybe it would be better (because more explicit) to not use globbing here.

ProcessLib/HydroMechanics/HydroMechanicsProcess.cpp
```diff
@@ -313,6 +320,29 @@ void HydroMechanicsProcess<DisplacementDim>::
...
+        }
+    };
+    if (_use_monolithic_scheme || _coupled_solutions->process_id == 0)
+    {
+        copyRhs(0, *_hydraulic_flow);
+    }
+    if (_use_monolithic_scheme || _coupled_solutions->process_id == 1)
+    {
+        copyRhs(1, *_nodal_forces);
+    }
```
chleh: Is there a possibility to do the computation only prior to writing output?
What about the output of ICs? Will it contain the two fields? Will they be all-zero?

ProcessLib/Utils/GlobalVectorUtils.h
```diff
@@ -0,0 +1,53 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2018, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
```
chleh: Maybe this file can be merged with `NumLib/DOF/DOFTableUtil.h`, because there already are functions accessing nodal values, computing norms etc.

ProcessLib/HydroMechanics/HydroMechanicsProcess.cpp
```diff
@@ -10,11 +10,13 @@
 #include "HydroMechanicsProcess.h"
 
 #include <cassert>
+#include <functional>
```
chleh: include explicitly needed?

ProcessLib/LIE/HydroMechanics/HydroMechanicsProcess.cpp
```diff
@@ -403,6 +404,24 @@ void HydroMechanicsProcess<GlobalDim>::initializeConcreteProcess(
             MeshLib::MeshItemType::Node, 1);
         mesh_prop_nodal_p->resize(mesh.getNumberOfNodes());
         _process_data.mesh_prop_nodal_p = mesh_prop_nodal_p;
+
+        _process_data.nodal_forces = MeshLib::getOrCreateMeshProperty<double>(
```
chleh: Minor naming inconsistency: cf. ` _process_data.mesh_prop_nodal_p` and `_process_data.nodal_forces`.

chleh: Btw: In HM you do this initialization directly in the ctor.

ProcessLib/LIE/HydroMechanics/HydroMechanicsProcess.cpp
```diff
@@ -403,6 +404,24 @@ void HydroMechanicsProcess<GlobalDim>::initializeConcreteProcess(
...
+            const_cast<MeshLib::Mesh&>(mesh), "NodalForces",
+            MeshLib::MeshItemType::Node, GlobalDim);
+        assert(_process_data.nodal_forces->size() ==
```
chleh: For the future: We need an assertion that also covers release mode for such cheap checks.

ProcessLib/Utils/GlobalVectorUtils.h
```diff
@@ -0,0 +1,53 @@
...
+
+#include "MeshLib/Properties.h"
+
+#include "NumLib/DOF/LocalToGlobalIndexMap.h"
+
+namespace ProcessLib
+{
+/// Copies part of a global vector for the given variable into output vector
```
chleh: Minor: `into output_vector` (underscore) or `into the output vector`

ProcessLib/Utils/GlobalVectorUtils.h
```diff
@@ -0,0 +1,53 @@
...
+/// \attention The output_vector is accessed through node id and component,
+/// therefore multiple meshes are not supported.
+template <typename Functor>
+void transformVariableFromGlobalVector(
+    GlobalVector const& input_vector, int const variable_id,
+    NumLib::LocalToGlobalIndexMap const& local_to_global_index_map,
+    MeshLib::PropertyVector<double>& output_vector, Functor mapFunction)
+{
+    std::fill(begin(output_vector), end(output_vector),
+              std::numeric_limits<double>::quiet_NaN());
```
chleh: In HM the generated output for the pressure-related rhs will be a mix of NaNs and proper values, right?
Maybe for that case the rhs should be interpolated to the second-order nodes in that case?

ProcessLib/HydroMechanics/Tests.cmake
endJunction: For many time steps it is convenient, though I agree that the explicit version would catch missing files.

ProcessLib/HydroMechanics/HydroMechanicsProcess.cpp
endJunction: I'll check the compute-on-output.

The mesh properties are created, and initialized with zero by default, and written to the 0th timestep, that's correct.

ProcessLib/Utils/GlobalVectorUtils.h
endJunction: Yes, the nodes not participating in the process have NaN output.
Interpolation is not suitable IMO. I try to explain it in the sketch:
```
Force <-----  * ==== x ==== * ==== x ==== * .... ==== x ==== * -----> opposing Force
```
where `*` is a node with nodal force output, `x` is the node with NaN force, and `====` is the element connection.
The sum of the forces is 0 and can occur only on boundary nodes (injection nodes). Interpolation does not necessarily leads to balanced forces, only if the "secondary" nodes lie in the middle between the "primary" nodes.

ProcessLib/LIE/HydroMechanics/HydroMechanicsProcess.cpp
endJunction: Made consistent.
In HM it is better, but for historical reasons it is initialized later in LIE/{SD,HM} (and requires a `const_cast` on the mesh too).

endJunction:
I'd merge it tonight, if there are no further comments.

PR #2122, opened 2018-05-14
Point shape function for second order quadratic elements

endJunction:
Adding a missing point shape function. Tested with a Line3 mesh (RichardsComponentTransport, copy of Padilla_NaCl1 ctest).


ProcessLib/RichardsComponentTransport/Tests.cmake
```diff
@@ -11,6 +11,19 @@ AddTest(
...
+    NAME RichardsComponentTransport_1D_Padilla_NaCl1_quadratic
+    PATH Parabolic/RichardsComponentTransport/Padilla/Padilla_NaCl1
+    EXECUTABLE ogs
+    EXECUTABLE_ARGS Padilla_NaCl1_quadratic.prj
+    WRAPPER time
+    TESTER vtkdiff
+    REQUIREMENTS NOT OGS_USE_MPI
+    DIFF_DATA
+    GLOB Padilla_NaCl1_quadratic_pcs_0_ts_*.vtu concentration concentration 1e-2 1e-10
+    GLOB Padilla_NaCl1_quadratic_pcs_0_ts_*.vtu pressure pressure 1e-2 1e-10
```
chleh: Are you comparing to numerical reference solutions? If so, the tolerance of 1e-2 seems a bit coarse.

endJunction: Agree, the tolerances were a "little" loose. Fixed for all of the Padilla tests, let's hope other builds will arrive at similar differences.

Issue #2054, opened 2018-01-25
Darcy velocity calculation is wrong in HT


waltherm:
The darcy velocity calculation in direction of gravity is wrong in all HT process benchmarks (maybe this is only wrong for the output). 

@wenqing can you please check this after you changed the velocity calculation lately?

wenqing:
I am going to take a look.

wenqing:
Checked, I think it is corrected according to
v=-k(grad p + rho g grad z)

In the source code, v is calculated as
```
 -K_over_mu * dNdx * p_nodal_values + K_over_mu * rho_w * b;
```
where b  is assumed that it is directed 'downwards'.  In the project files of square_5500x5500.prj (monolithic) and square_5500x5500_staggered_scheme.prj, b is input as
```
<specific_body_force>0 -9.81</specific_body_force>
```
The negative sign in which shows that the input is corrected too.

![t](https://user-images.githubusercontent.com/1343839/36089559-d935618c-0fdc-11e8-965d-d5600d6923bf.png)

wenqing:
In HT, both of the staggered scheme and the monolithic scheme use the same function for velocity calculation.

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/2054).

PR #2124, opened 2018-05-16
[PL/HT] Fix computation of darcy velocity.

TomFischer:
Corrected access to the local pressure and temperature variables.

Fixes issue #2054.


TomFischer:
@wenqing I think we can let them as separate PRs.

PR #2125, opened 2018-05-16
BoundaryElementSearch: Return bulk element id and bulk element face id.

TomFischer:
Preparation PR for the constraint boundary condition PR that will be created soon.

### DIFF
--- a/MeshGeoToolsLib/BoundaryElementsAlongPolyline.cpp
+++ b/MeshGeoToolsLib/BoundaryElementsAlongPolyline.cpp
@@ -48,8 +48,12 @@ BoundaryElementsAlongPolyline::BoundaryElementsAlongPolyline(
             auto* edge = e->getEdge(i);
             // check if all edge nodes are along the polyline (if yes, store a distance)
             std::vector<std::size_t> edge_node_distances_along_ply;
-            if (includesAllEdgeNodeIDs(node_ids_on_poly, *edge, edge_node_distances_along_ply)) {
-                auto* new_edge = modifyEdgeNodeOrdering(*edge, ply, edge_node_distances_along_ply, node_ids_on_poly);
+            if (includesAllEdgeNodeIDs(node_ids_on_poly, *edge,
+                                       edge_node_distances_along_ply))
+            {
+                auto* new_edge = modifyEdgeNodeOrdering(
+                    *edge, ply, edge_node_distances_along_ply,
+                    node_ids_on_poly);
                 if (edge != new_edge)
                     delete edge;
                 _boundary_elements.push_back(new_edge);
--- a/MeshGeoToolsLib/BoundaryElementsAlongPolyline.cpp
+++ b/MeshGeoToolsLib/BoundaryElementsAlongPolyline.cpp
@@ -57,22 +57,52 @@ BoundaryElementsAlongPolyline::BoundaryElementsAlongPolyline(
                 if (edge != new_edge)
                     delete edge;
                 _boundary_elements.push_back(new_edge);
+                _bulk_ids.emplace_back(e->getID(), i);
             } else {
                 delete edge;
             }
         }
     }
 
-    // sort picked edges according to a distance of their first node along the polyline
-    std::sort(_boundary_elements.begin(), _boundary_elements.end(),
-            [&](MeshLib::Element*e1, MeshLib::Element*e2)
-            {
-                std::size_t dist1 = std::distance(node_ids_on_poly.begin(),
-                    std::find(node_ids_on_poly.begin(), node_ids_on_poly.end(), e1->getNodeIndex(0)));
-                std::size_t dist2 = std::distance(node_ids_on_poly.begin(),
-                    std::find(node_ids_on_poly.begin(), node_ids_on_poly.end(), e2->getNodeIndex(0)));
-                return (dist1 < dist2);
-            });
+    // zip _boundary_elements and _bulk_element_ids
+    std::vector<std::pair<MeshLib::Element*, std::pair<std::size_t, unsigned>>>
+        zipped_data;
+    zipped_data.reserve(_boundary_elements.size());
+    std::transform(_boundary_elements.cbegin(), _boundary_elements.cend(),
+                   _bulk_ids.cbegin(), std::back_inserter(zipped_data),
+                   [](MeshLib::Element* boundary_element,
+                      std::pair<std::size_t, unsigned> const& bulk_ids) {
+                       return std::make_pair(boundary_element, bulk_ids);
+                   });
+
+    // The sort was necessary in OGS-5 for some reason. I'm not sure if it is
+    // needed anymore in OGS-6.
+    // sort picked edges according to a distance of their first node along the
+    // polyline
+    std::sort(zipped_data.begin(), zipped_data.end(),
+              [&](std::pair<MeshLib::Element*,
+                            std::pair<std::size_t, unsigned>> const& a,
+                  std::pair<MeshLib::Element*,
+                            std::pair<std::size_t, unsigned>> const& b) {
+                  auto const* const e1 = a.first;
+                  auto const* const e2 = b.first;
+                  std::size_t dist1 = std::distance(
+                      node_ids_on_poly.begin(),
+                      std::find(node_ids_on_poly.begin(),
+                                node_ids_on_poly.end(), e1->getNodeIndex(0)));
+                  std::size_t dist2 = std::distance(
+                      node_ids_on_poly.begin(),
+                      std::find(node_ids_on_poly.begin(),
+                                node_ids_on_poly.end(), e2->getNodeIndex(0)));
+                  return (dist1 < dist2);
+              });
+
+    // unzip the sorted data
+    for (std::size_t k = 0; k < zipped_data.size(); ++k)
+    {
+        _boundary_elements[k] = zipped_data[k].first;
+        _bulk_ids[k] = zipped_data[k].second;
+    }
 }
 
 BoundaryElementsAlongPolyline::~BoundaryElementsAlongPolyline()
@@ -126,6 +156,12 @@ MeshLib::Element* BoundaryElementsAlongPolyline::modifyEdgeNodeOrdering(
 
     // Return the original edge otherwise.
     return const_cast<MeshLib::Element*>(&edge);
+}
+
+std::vector<std::pair<std::size_t, unsigned>> const&
+BoundaryElementsAlongPolyline::getBulkIDs() const
+{
+    return _bulk_ids;
 }
 
 } // end namespace MeshGeoToolsLib
--- a/MeshGeoToolsLib/BoundaryElementsAlongPolyline.cpp
+++ b/MeshGeoToolsLib/BoundaryElementsAlongPolyline.cpp
@@ -11,6 +11,7 @@
 #include <algorithm>
 #include <typeinfo>
 
+#include "BaseLib/quicksort.h"
 #include "GeoLib/Polyline.h"
 
 #include "MeshLib/Mesh.h"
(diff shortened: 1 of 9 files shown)


MeshGeoToolsLib/BoundaryElementsSearcher.h
```diff
@@ -56,20 +56,28 @@ class BoundaryElementsSearcher
      */
     std::vector<MeshLib::Element*> const& getBoundaryElements(GeoLib::GeoObject const& geoObj);
 
+    std::vector<std::pair<std::size_t, unsigned>> const& getBulkIDs(
```
chleh: Documentation?

MeshGeoToolsLib/BoundaryElementsSearcher.h
```diff
@@ -56,20 +56,28 @@ class BoundaryElementsSearcher
...
+        GeoLib::GeoObject const& geometry);
+
     /**
      * generate boundary elements at the given point.
      * @param point Search the mesh for given point
      * @return a vector of boundary elements
      */
     std::vector<MeshLib::Element*> const& getBoundaryElementsAtPoint(
         GeoLib::Point const& point);
+    std::vector<std::pair<std::size_t, unsigned>> const& getBulkIDsAtPoint(
```
chleh: Documentation?

MeshGeoToolsLib/BoundaryElementsSearcher.cpp
```diff
@@ -58,6 +58,28 @@ std::vector<MeshLib::Element*> const& BoundaryElementsSearcher::getBoundaryEleme
...
+        return getBulkIDsAlongPolyline(
+            *dynamic_cast<const GeoLib::Polyline*>(&geometry));
+        break;
+    case GeoLib::GEOTYPE::SURFACE:
+        return getBulkIDsOnSurface(
+            *dynamic_cast<const GeoLib::Surface*>(&geometry));
+        break;
+    default:
+        const static std::vector<std::pair<std::size_t, unsigned>> dummy;
+        return dummy;
```
chleh: Shouldn't that case rather be an error?

MeshGeoToolsLib/BoundaryElementsAtPoint.cpp
```diff
@@ -37,10 +37,19 @@ BoundaryElementsAtPoint::BoundaryElementsAtPoint(
...
+    auto& mesh_nodes =
+        const_cast<std::vector<MeshLib::Node*>&>(_mesh.getNodes());
+
     std::array<MeshLib::Node*, 1> const nodes = {{
         const_cast<MeshLib::Node*>(_mesh.getNode(node_ids[0]))}};
 
     _boundary_elements.push_back(new MeshLib::Point{nodes, node_ids[0]});
+    for (auto const* bulk_element : _mesh.getNode(node_ids[0])->getElements())
+    {
+        _bulk_ids.emplace_back(std::make_pair(
```
chleh: Minor Note: `emplace_back` shouldn't require `std::make_pair`.

MeshGeoToolsLib/BoundaryElementsAtPoint.cpp
```diff
@@ -37,10 +37,19 @@ BoundaryElementsAtPoint::BoundaryElementsAtPoint(
...
+            bulk_element->getID(),
+            bulk_element->identifyFace(mesh_nodes.data()+node_ids[0])));
```
chleh: (Not thir PR) That `identifyFace` does not have a const argument looks like a mistake.

MeshGeoToolsLib/BoundaryElementsAlongPolyline.cpp
```diff
@@ -57,22 +57,50 @@ BoundaryElementsAlongPolyline::BoundaryElementsAlongPolyline(
...
+    zipped_data.reserve(_boundary_elements.size());
+    std::transform(_boundary_elements.cbegin(), _boundary_elements.cend(),
+                   _bulk_ids.cbegin(), std::back_inserter(zipped_data),
+                   [](MeshLib::Element* boundary_element,
+                      std::pair<std::size_t, unsigned> const& bulk_ids) {
+                       return std::make_pair(boundary_element, bulk_ids);
+                   });
+
+    // sort picked edges according to a distance of their first node along the
+    // polyline
```
chleh: Is that necessary to speed up some subsequent linear searches?

MeshGeoToolsLib/BoundaryElementsSearcher.cpp
TomFischer: In principle you are right. I made it consistent with the existing code. The new method `getBulkIDs()` has the same behaviour as the similar method `getBoundaryElements()`. I don't want to break the consistency in this PR.  @chleh Is this okay for you?

MeshGeoToolsLib/BoundaryElementsAlongPolyline.cpp
TomFischer: The boundary elements have been sorted in the previous version of the implementation, too. I think the sort was necessary in OGS-5 for some reason. I'm not sure if it is needed anymore in OGS-6.

MeshGeoToolsLib/BoundaryElementsSearcher.cpp
chleh: OK.

MeshGeoToolsLib/BoundaryElementsAlongPolyline.cpp
chleh: Probably currently it's too much work to check if that code is still needed. But could you please add a comment that the sorting might be unnecessary. Because otherewise somebody who will refactor that piece of code later one might be pretty puzzled.

MeshGeoToolsLib/BoundaryElementsAlongPolyline.cpp
```diff
@@ -48,27 +48,61 @@ BoundaryElementsAlongPolyline::BoundaryElementsAlongPolyline(
...
-    std::sort(_boundary_elements.begin(), _boundary_elements.end(),
-            [&](MeshLib::Element*e1, MeshLib::Element*e2)
-            {
-                std::size_t dist1 = std::distance(node_ids_on_poly.begin(),
-                    std::find(node_ids_on_poly.begin(), node_ids_on_poly.end(), e1->getNodeIndex(0)));
-                std::size_t dist2 = std::distance(node_ids_on_poly.begin(),
-                    std::find(node_ids_on_poly.begin(), node_ids_on_poly.end(), e2->getNodeIndex(0)));
-                return (dist1 < dist2);
-            });
+    // zip _boundary_elements and _bulk_element_ids
```
endJunction: Very similar algorithm zip-sort-unzip is used in BaseLib::quicksort. Maybe the BaseLib version can be generalized and used for this purpose? The only difference there is that `begin` is a pointer and offset, and same for the `end`.

MeshGeoToolsLib/BoundaryElementsAlongPolyline.h
```diff
@@ -60,6 +60,8 @@ class BoundaryElementsAlongPolyline
      */
     std::vector<MeshLib::Element*> const& getBoundaryElements() const {return _boundary_elements; }
 
+    std::vector<std::pair<std::size_t, unsigned>> const& getBulkIDs() const;
```
endJunction: Short comment about what is the first and second pair's element meaning.

MeshGeoToolsLib/BoundaryElementsAlongPolyline.h
```diff
@@ -83,6 +85,8 @@ class BoundaryElementsAlongPolyline
     MeshLib::Mesh const& _mesh;
     GeoLib::Polyline const& _ply;
     std::vector<MeshLib::Element*> _boundary_elements;
+
+    std::vector<std::pair<std::size_t, unsigned>> _bulk_ids;
```
endJunction: Same comment here.

MeshGeoToolsLib/BoundaryElementsAtPoint.cpp
```diff
@@ -37,10 +37,19 @@ BoundaryElementsAtPoint::BoundaryElementsAtPoint(
             "one is expected.",
             node_ids.size(), _point[0], _point[1], _point[2]);
 
+    auto& mesh_nodes =
+        const_cast<std::vector<MeshLib::Node*>&>(_mesh.getNodes());
```
endJunction: Not sure here: Maybe the const cast is not necessary? It is only used in the identifyFace call through a pointer...

MeshGeoToolsLib/BoundaryElementsAtPoint.h
```diff
@@ -56,10 +56,13 @@ class BoundaryElementsAtPoint final
         return _boundary_elements;
     }
 
+    std::vector<std::pair<std::size_t, unsigned>> const& getBulkIDs() const;
+
 private:
     MeshLib::Mesh const& _mesh;
     GeoLib::Point const& _point;
     std::vector<MeshLib::Element*> _boundary_elements;
+    std::vector<std::pair<std::size_t, unsigned>> _bulk_ids;
```
endJunction: same comment here *etc.*

MeshGeoToolsLib/BoundaryElementsSearcher.cpp
```diff
@@ -74,6 +96,22 @@ BoundaryElementsSearcher::getBoundaryElementsAtPoint(GeoLib::Point const& point)
...
 }
 
+std::vector<std::pair<std::size_t, unsigned>> const&
+BoundaryElementsSearcher::getBulkIDsAtPoint(GeoLib::Point const& point)
+{
+    // Look for already saved points and return if found.
+    for (auto const& boundary_elements : _boundary_elements_at_point)
+    {
+        if (boundary_elements->getPoint() == point)
+            return boundary_elements->getBulkIDs();
```
endJunction: The STL version:
```c++
auto it = std::find_if(begin(_boundary_elements_at_point), end(_boundary_elements_at_point,
   [&point](auto const& boundary_element)  {
      return boundary_elements->getPoint() == point;
   });
if (it != _boundary_elements_at_point.end())
    return it->getBulkIDs();
```

... is not necessarily shorter

MeshGeoToolsLib/BoundaryElementsSearcher.h
```diff
@@ -56,27 +56,62 @@ class BoundaryElementsSearcher
...
 
+    /// A boundary element is generated out of a face of a bulk element. This
+    /// method returns the bulk element ids and the corresponding face ids.
+    /// @param geometry The boundary elements, bulk element ids and  bulk
+    /// element face ids will be extracted for the given geometry (point,
+    /// polyline or surface).
+    /// @return A vector of id pairs will be returned where the first item of
+    /// the pair is the element id and the second item of the pair is the face
+    /// id.
+    std::vector<std::pair<std::size_t, unsigned>> const& getBulkIDs(
```
endJunction: Ah, I found the docu. You might want to use `\copydoc` if it's all the same text.

MeshGeoToolsLib/BoundaryElementsAtPoint.cpp
TomFischer: `identifyFace` expects a non const array. As @chleh mentioned this is probably bad design in `identifyFace`.

MeshGeoToolsLib/BoundaryElementsSearcher.cpp
TomFischer: It is not shorter and not substantially more readable. I would like to keep my implementation. Is this okay for you?

endJunction:
Hope the changes still give same results. If not, drop the BL changes and merge.

PR #2127, opened 2018-05-18
Fixed  the calculation of the Darcy velocity in staggered TH

wenqing:
Follow up PR #2124, this PR fixes a bug  in the calculation of the Darcy velocity in the staggered TH. The bug is caused  by passing wrong parameters to the arguments of the  velocity calculation function.

To fix the bug and also make the source code more readable, process indices are introduced in the local assembler of  the staggered scheme for HT.

Due to the partitioning of nodal/element properties being available now,  the number  of parallel compute cores for the two related benchmarks are increased to three from one.


ProcessLib/HT/HTProcess.cpp
```diff
@@ -140,8 +143,8 @@ void HTProcess::assembleWithJacobianConcreteProcess(
     // Call global assembler for each local assembly item.
     GlobalExecutor::executeMemberDereferenced(
         _global_assembler, &VectorMatrixAssembler::assembleWithJacobian,
-        _local_assemblers, dof_tables, t, x, xdot, dxdot_dx,
-        dx_dx, M, K, b, Jac, _coupled_solutions);
+        _local_assemblers, dof_tables, t, x, xdot, dxdot_dx, dx_dx, M, K, b,
```
wenqing: Changes due to clang-format

ProcessLib/HT/HTProcess.cpp
```diff
@@ -181,7 +184,7 @@ void HTProcess::setCoupledTermForTheStaggeredSchemeToLocalAssemblers()
 }
 
 std::tuple<NumLib::LocalToGlobalIndexMap*, bool>
-    HTProcess::getDOFTableForExtrapolatorData() const
+HTProcess::getDOFTableForExtrapolatorData() const
```
wenqing: Change due to clang-format

ProcessLib/HT/HTProcess.cpp
```diff
@@ -199,9 +202,10 @@ std::tuple<NumLib::LocalToGlobalIndexMap*, bool>
 
     const bool manage_storage = true;
     return std::make_tuple(new NumLib::LocalToGlobalIndexMap(
-        std::move(all_mesh_subsets_single_component),
-        // by location order is needed for output
-        NumLib::ComponentOrder::BY_LOCATION), manage_storage);
+                               std::move(all_mesh_subsets_single_component),
```
wenqing: Change due to clang-format

ProcessLib/HT/StaggeredHTFEM-impl.h
```diff
@@ -126,9 +128,10 @@ void StaggeredHTFEM<ShapeFunction, IntegrationMethod, GlobalDim>::
         GlobalDimMatrixType K_over_mu = intrinsic_permeability / viscosity;
 
         // matrix assembly
-        local_M.noalias() += w * (porosity * dfluid_density_dp / fluid_density +
-                                  specific_storage) *
-                             N.transpose() * N;
+        local_M.noalias() +=
```
wenqing: Change due to clang-format

TomFischer:
Thanks @wenqing for fixing the darcy velocity calculation in the HT staggered implementation.  In PR https://github.com/ufz/ogs/pull/2124 some simple benchmarks were introduced (see Tests/Data/Parabolic/HT/SimpleSynthetics/) for the monolithic implementation. Please, could you add similar benchmarks for the staggered scheme?

wenqing:
@TomFischer  I will add that benchmark

wenqing:
@chleh That benchmark, which sounds like a sensitive analysis, gets slightly difference between the reference result and the result by the present master branch by using either the monolithic scheme or the staggered scheme.  Since that benchmark is marked as LARGE, it is not checked on jenkins.  For which change in the code for the slight discrepancy on results is still under investigating .  The other benchmarks are OK.

PR #2128, opened 2018-05-18
 Removed two compilation warnings

wenqing:
Removed two compilation warnings

### DIFF
--- a/NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.cpp
+++ b/NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.cpp
@@ -163,7 +163,7 @@ void LocalLinearLeastSquaresExtrapolator::extrapolateElement(
     // number of integration points in the element
     const auto num_int_pts = num_values / num_components;
 
-    if (num_int_pts < static_cast<decltype(num_int_pts)>(num_nodes))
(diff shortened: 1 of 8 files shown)


NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.cpp
```diff
@@ -70,12 +70,12 @@ void LocalLinearLeastSquaresExtrapolator::extrapolate(
...
     if (!_nodal_values ||
 #ifdef USE_PETSC
-        static_cast<std::size_t>(_nodal_values->getLocalSize() +
+        static_cast<int>(_nodal_values->getLocalSize() +
                                  _nodal_values->getGhostSize())
 #else
         _nodal_values->size()
 #endif
-            != num_nodal_dof_result)
+            != static_cast<int>(num_nodal_dof_result))
```
endJunction: Not sure about this casts. The sizes of a GlobalVector are probably the largest integers available, so the std::size_t used before here seems better sized.
I'd continue to use std::size_t here unless there are some other arguments to use an int.

endJunction: Using GlobalIndexType in one of the casts solves the issue. The change is already uploaded.

PR #2129, opened 2018-05-21
LIE/HM fixes and enhancements

endJunction:
If the PR is too long I can easily split it into two.

LIE/HM changes:
 - Darcy velocity output for fracture too.
 - Fix leak off into the matrix if the matrix flow is deactivated.
 - Add aperture0 interpolation: For now only constant aperture0 was possible. Now element-wise or node-wise parameter is working properly. The nodal values are interpolated using the corresponding shape functions.

For the last feature the Parameter class was extended by the `getNodalElementValues` function with corresponding tests.


ProcessLib/Parameter/Parameter.h
```diff
@@ -13,18 +13,14 @@
...
 
 namespace BaseLib
 {
 class ConfigTree;
 }  // BaseLib
 
-namespace MathLib
-{
-class PiecewiseLinearInterpolation;
-}  // MathLib
```
chleh: Hm. IMHO the forward declaration should suffice here. But then maybe the include has to be added to `Parameter.cpp`.

ProcessLib/Parameter/CurveScaledParameter.h
```diff
@@ -29,13 +29,24 @@ struct CurveScaledParameter final : public Parameter<T> {
     }
 
     bool isTimeDependent() const override { return true; }
+
+    void setParameter(ParameterBase const* parameter)
+    {
+        if (dynamic_cast<Parameter<double> const*>(parameter) == nullptr)
+        {
+            OGS_FATAL(
+                "Parameters other then for double type are not supported.");
```
chleh: But the CurveScaledParameter is parametrized by `T`.

ProcessLib/Parameter/CurveScaledParameter.h
```diff
@@ -29,13 +29,24 @@ struct CurveScaledParameter final : public Parameter<T> {
     }
 
     bool isTimeDependent() const override { return true; }
+
+    void setParameter(ParameterBase const* parameter)
```
chleh: This will be only used in the tests, right? Actually I think, we could live without this method.

ProcessLib/LIE/HydroMechanics/HydroMechanicsProcess.cpp
```diff
@@ -489,9 +490,22 @@ void HydroMechanicsProcess<GlobalDim>::computeSecondaryVariableConcrete(
...
     auto const& R = _process_data.fracture_property->R;
+    auto const& b0 = _process_data.fracture_property->aperture0;
     MeshLib::PropertyVector<double>& vec_w = *_process_data.mesh_prop_nodal_w;
     MeshLib::PropertyVector<double>& vec_b = *_process_data.mesh_prop_nodal_b;
 
+    auto compute_nodal_aperture = [&](std::size_t const node_id,
+                                      double const w_n) {
+        // skip aperture computation for element-wise defined b0 because there
+        // are jumps on the nodes between the element's values.
+        if (dynamic_cast<MeshElementParameter<double> const*>(b0))
```
chleh: Why not fail immediately, because the NaN will destroy the computation anyway?
Btw: Can't that be checked during the initialization?

ProcessLib/LIE/HydroMechanics/HydroMechanicsProcess.cpp
```diff
@@ -489,9 +490,22 @@ void HydroMechanicsProcess<GlobalDim>::computeSecondaryVariableConcrete(
 
     // compute nodal w and aperture
     auto const& R = _process_data.fracture_property->R;
+    auto const& b0 = _process_data.fracture_property->aperture0;
```
chleh: `auto *const` for clarity?

ProcessLib/LIE/HydroMechanics/HydroMechanicsProcess.cpp
```diff
@@ -489,9 +490,22 @@ void HydroMechanicsProcess<GlobalDim>::computeSecondaryVariableConcrete(
...
+    auto compute_nodal_aperture = [&](std::size_t const node_id,
+                                      double const w_n) {
+        // skip aperture computation for element-wise defined b0 because there
+        // are jumps on the nodes between the element's values.
+        if (dynamic_cast<MeshElementParameter<double> const*>(b0))
+            return std::numeric_limits<double>::quiet_NaN();
+
+        ProcessLib::SpatialPosition x;
+        x.setNodeID(node_id);
+        return w_n + (*b0)(0, x)[0];
```
chleh: What about time, just for completeness? `(0, x)` --> `(t, x)`

ProcessLib/Parameter/Parameter.h
```diff
@@ -68,6 +75,40 @@ struct Parameter : public ParameterBase
...
+    virtual Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic>
+    getNodalValuesOnElement(MeshLib::Element const& element,
+                            double const t) const
+    {
+        auto const n_nodes = static_cast<int>(element.getNumberOfNodes());
+        auto const n_components = getNumberOfComponents();
+        Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> result(n_components,
+                                                                n_nodes);
+
+        // Column vector of constant values, copied for each node.
```
chleh: `constant` values?

ProcessLib/Parameter/Parameter.h
```diff
@@ -68,6 +75,40 @@ struct Parameter : public ParameterBase
     //! Returns the parameter value at the given time and position.
     virtual std::vector<T> const& operator()(
         double const t, SpatialPosition const& pos) const = 0;
+
+    //! Returns a matrix of values for all nodes of the given element.
+    //
+    // The matrix is of the shape CxN, where C is the number of components and N
+    // the number of nodes, such that subsequent multiplication with shape
+    // functions matrix results in a vector of length C.
```
chleh: In many occasions the multiplication of shape (row) vectors `N` and nodal values (column vectors) `V` is currently done as `N * V`. Due to consistency reasons therefore it might be better to transpose the matrix order in this method. Furthermore, often we use row major storage.

ProcessLib/Parameter/Parameter.h
```diff
@@ -68,6 +75,40 @@ struct Parameter : public ParameterBase
...
+                                                                n_nodes);
+
+        // Column vector of constant values, copied for each node.
+        SpatialPosition x_position;
+        auto const nodes = element.getNodes();
+        for (int i = 0; i < n_nodes; ++i)
+        {
+            x_position.setAll(
+                nodes[i]->getID(), element.getID(), boost::none, boost::none);
+            auto const& row = this->operator()(t, x_position);
```
chleh: Why row if you later use it as col? Applies also to overrides.

ProcessLib/Parameter/MeshNodeParameter.h
```diff
@@ -57,6 +59,27 @@ struct MeshNodeParameter final : public Parameter<T> {
         return _cache;
     }
 
+    Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> getNodalValuesOnElement(
+        MeshLib::Element const& element, double const t) const override
+    {
```
chleh: Doesn't the base implementation do exactly the same for nodal params?

ProcessLib/Parameter/Parameter.h
endJunction: Unfortunately not, because the unique_ptr needs the definition of the destructor.

chleh: Even though it's passed as const reference? Strange. Another thing learned.

ProcessLib/LIE/HydroMechanics/LocalAssembler/HydroMechanicsLocalAssemblerFracture-impl.h
```diff
@@ -377,11 +379,13 @@ void HydroMechanicsLocalAssemblerFracture<ShapeFunctionDisplacement,
         ele_sigma_eff += ip.sigma_eff;
         ele_Fs = std::max(
             ele_Fs, ip.material_state_variables->getShearYieldFunctionValue());
+        ele_velocity += ip.darcy_velocity;
```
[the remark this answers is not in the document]
endJunction: Didn't think about it before, but you are probably right, especially in the cases where the weights are different for each point. (Doesn't happen for quads.)

ProcessLib/LIE/HydroMechanics/LocalAssembler/IntegrationPointDataFracture.h
```diff
@@ -50,6 +50,8 @@ struct IntegrationPointDataFracture final
     Eigen::MatrixXd C;
     double integration_weight;
 
+    Eigen::Vector3d darcy_velocity;
```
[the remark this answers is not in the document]
endJunction: Yes, it is not recalculated for the output (in the `computeSecondaryVariable...` call).

For the coffee discussion: It might be worth to add yet another function to the `Process` interface for updating the local states after a new solution was computed. This matters not only for output, as you mention, but maybe also for the staggered coupling.

ProcessLib/Parameter/CurveScaledParameter.h
endJunction: Yes, I added this for the testing, otherwise one would need to construct a vector of unique pointers to the parameters, s.t. the `initialize` function could be used.

Maybe moving the `findParameter` call and use the `setParameter` instead of `initialize` would result in better interface. The `initialize` function is implemented only in the CurveScaledParameter. (I guess the discussion is for another coffee ;)

ProcessLib/Parameter/CurveScaledParameter.h
endJunction: Fixing this part, but the actual problem is in the create method, where there is only double type allowed: https://github.com/ufz/ogs/blob/master/ProcessLib/Parameter/CurveScaledParameter.cpp#L38-L40

Same for all other types.

ProcessLib/LIE/HydroMechanics/HydroMechanicsProcess.cpp
endJunction: The NaN is for output only, the computation is not using this value.
One could check it in the initialization, but then it's another boolean indicating some special feature.
I'd like to keep this output-only thing here.

ProcessLib/LIE/HydroMechanics/HydroMechanicsProcess.cpp
endJunction: Here, it's other way around; b0 is b at time 0. Just for safety...
Would be nice if we could force the parameter to be time-independent....

ProcessLib/Parameter/MeshNodeParameter.h
endJunction: There is a small difference. If I find some time to measure the differences, I'll report here.

ProcessLib/Parameter/Parameter.h
endJunction: See [SO](https://stackoverflow.com/a/6089065) and the [reference](http://howardhinnant.github.io/incomplete.html) therein.
[cppreference](http://en.cppreference.com/w/cpp/memory/unique_ptr) (search "incomplete type") also has a paragraph about the incomplete type requirements of the unique_ptr, but more difficult to understand.

endJunction: Hmmm, now it's working :confused:

ProcessLib/Parameter/Parameter.h
endJunction: Changed the result to be NxC, same as *e.g.* the local_x scalar quantities T or p.

ProcessLib/LIE/HydroMechanics/LocalAssembler/HydroMechanicsLocalAssemblerFracture-impl.h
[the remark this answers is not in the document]
chleh: Anyway, it's probably only a minor thing. Let's leave it as it is for now.

ProcessLib/LIE/HydroMechanics/HydroMechanicsProcess.cpp
chleh: Maybe add a comment that time is zero deliberately?
Could be checked during initialization...

ProcessLib/Parameter/Parameter.h
chleh: OK, thanks.

ProcessLib/Parameter/MeshNodeParameter.h
chleh: No need to measure. OK for me, even though the difference is only in the saved `element.getID()` essentially.

ProcessLib/LIE/HydroMechanics/HydroMechanicsProcess.cpp
chleh: OK.

chleh:
To be discussed:
* [ ] initialize vs. setParameter in CurveScaledParameter
* [ ] It might be worth to add yet another function to the Process interface
* [ ] time-independent param check for b0

endJunction:
`setParameter` removed; b0 time-independency check added;
Process interface extension discussed; Current `computeSecondaryVars` is already fulfilling the requirements -> no change for now.

ProcessLib/Parameter/ConstantParameter.h
```diff
@@ -46,6 +46,24 @@ struct ConstantParameter final : public Parameter<T>
         return _values;
     }
 
+    Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> getNodalValuesOnElement(
```
TomFischer: Documentation needed? Maybe @copydoc?

ProcessLib/Parameter/MeshElementParameter.h
```diff
@@ -50,9 +55,30 @@ struct MeshElementParameter final : public Parameter<T> {
         return _cache;
     }
 
+    Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> getNodalValuesOnElement(
```
TomFischer: Docu? `@copydoc`?

ProcessLib/Parameter/MeshNodeParameter.h
```diff
@@ -50,9 +59,30 @@ struct MeshNodeParameter final : public Parameter<T> {
         return _cache;
     }
 
+    Eigen::Matrix<T, Eigen::Dynamic, Eigen::Dynamic> getNodalValuesOnElement(
```
TomFischer: Docu? `@copydoc`?

ProcessLib/Parameter/ConstantParameter.h
endJunction: https://doxygen.opengeosys.org/d4/d33/structProcessLib_1_1ConstantParameter.html#af1afed3f67f93de2ddbdc14e737a595c

PR #2132, opened 2018-05-28
Balance for HT process

TomFischer:
PR implements getFlux in the in the class HTProcess and the local assembler for the HT process. The new feature is checked by two new test cases.


ProcessLib/HT/HTFEM.h
```diff
@@ -88,6 +88,76 @@ class HTFEM : public HTLocalAssemblerInterface
...
+
+        typename ShapeMatricesType::ShapeMatrices shape_matrices(
+            ShapeFunction::DIM, GlobalDim, ShapeFunction::NPOINTS);
+
+        // Note: Axial symmetry is set to false here, because we only need dNdx
+        // here, which is not affected by axial symmetry.
+        fe.computeShapeFunctions(pnt_local_coords.getCoords(), shape_matrices,
+                                 GlobalDim, false);
+        std::vector<double> flux;
+        flux.resize(3);
```
endJunction: One could move it down to near the return statement.
I'd say, that returning a Eigen::Vector3d is safer, then the std::vector...

ProcessLib/HT/HTFEM.h
```diff
@@ -88,6 +88,76 @@ class HTFEM : public HTLocalAssemblerInterface
...
+
+        MaterialLib::Fluid::FluidProperty::ArrayType vars;
+        double T_int_pt = 0.0;
+        double p_int_pt = 0.0;
+
+        // local_x contains the local temperature and pressure values
+        NumLib::shapeFunctionInterpolate(local_x, shape_matrices.N, T_int_pt, p_int_pt);
+
+        // set the values into array vars
+        vars[static_cast<int>(MaterialLib::Fluid::PropertyVariableType::T)] =
```
endJunction: Would it make sense to pass the vars[T] directly to shapeFunctionInterpolate call?
```c++
NumLib::shapeFunctionInterpolate(local_x, shape_matrices.N, vars[static_cast<int>(MaterialLib::Fluid::PropertyVariableType::T)], vars[static_cast<int>(MaterialLib::Fluid::PropertyVariableType::p)]);
```

ProcessLib/HT/HTFEM.h
```diff
@@ -88,6 +88,76 @@ class HTFEM : public HTLocalAssemblerInterface
...
+            -K_over_mu * shape_matrices.dNdx * p_nodal_values;
+
+        if (this->_material_properties.has_gravity)
+        {
+            auto const rho_w =
+                this->_material_properties.fluid_properties->getValue(
+                    MaterialLib::Fluid::FluidPropertyType::Density, vars);
+            auto const b = this->_material_properties.specific_body_force;
+            // here it is assumed that the vector b is directed 'downwards'
+            q += K_over_mu * rho_w * b;
```
endJunction: I'm not getting the intentions of the comment. What is 'downwards'? Maybe it's copied from some other context...

ProcessLib/HT/HTProcess.cpp
```diff
@@ -229,5 +240,59 @@ void HTProcess::setCoupledSolutionsOfPreviousTimeStep()
...
+        OGS_FATAL(
+            "The condition of process_id = 0 must be satisfied for "
+            "monolithic HTProcess, which is a single process.");
+    }
+    if (!_use_monolithic_scheme && process_id != 1)
+    {
+        DBUG("This is the thermal part of the staggered HTProcess.");
+        return;
+    }
+    if (_balance_mesh)  // computing the balance is optional
```
endJunction:
```c++
if (!_balance_mesh)
{
    return;
}
```

ProcessLib/HT/HTProcess.cpp
```diff
@@ -229,5 +240,59 @@ void HTProcess::setCoupledSolutionsOfPreviousTimeStep()
...
+    {
+        auto* const balance_pv = MeshLib::getOrCreateMeshProperty<double>(
+            *_balance_mesh, _balance_pv_name, MeshLib::MeshItemType::Cell, 1);
+        // initialise the PropertyVector pv with zero values
+        std::fill(balance_pv->begin(), balance_pv->end(), 0.0);
```
endJunction: Not sure here;
If the properties are set, then use NaN for initialization. If the properties are '+='-ed, then setting them 0 is of course necessary.

TomFischer: The properties are  '+='-ed.

ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
```diff
@@ -105,7 +105,7 @@ class LocalAssemblerData : public GroundwaterFlowLocalAssemblerInterface
     /// Computes the flux in the point \c p_local_coords that is given in local
     /// coordinates using the values from \c local_x.
     // TODO add time dependency
-    std::vector<double> getFlux(
```
chleh: ~~Isn't the flux 2D in two dimentsions?~~

ProcessLib/HT/HTFEM.h
```diff
@@ -88,6 +88,73 @@ class HTFEM : public HTLocalAssemblerInterface
         return Eigen::Map<const Eigen::RowVectorXd>(N.data(), N.size());
     }
 
+    /// Computes the flux in the point \c pnt_local_coords that is given in
+    /// local coordinates using the values from \c local_x.
+    // TODO add time dependency
```
chleh: Yep, that would be nice (for the sake of completeness). Could you please add it now?

ProcessLib/HT/HTFEM.h
```diff
@@ -88,6 +88,73 @@ class HTFEM : public HTLocalAssemblerInterface
...
+            vars[static_cast<int>(MaterialLib::Fluid::PropertyVariableType::T)],
+            vars[static_cast<int>(
+                MaterialLib::Fluid::PropertyVariableType::p)]);
+
+        // TODO remove following line if time dependency is implemented
+        double const t = 0.0;
+        auto const K =
+            this->_material_properties.porous_media_properties
+                .getIntrinsicPermeability(t, pos)
+                .getValue(t, pos, 0.0,
```
chleh: Is the `0.0` correct here?

ProcessLib/HT/HTFEM.h
```diff
@@ -88,6 +88,73 @@ class HTFEM : public HTLocalAssemblerInterface
...
+        {
+            auto const rho_w =
+                this->_material_properties.fluid_properties->getValue(
+                    MaterialLib::Fluid::FluidPropertyType::Density, vars);
+            auto const b = this->_material_properties.specific_body_force;
+            q += K_over_mu * rho_w * b;
+        }
+
+        Eigen::Vector3d flux;
+        flux.head(GlobalDim) = q;
```
chleh: Maybe here the templated version `head>GlobalDim>()` would be a tiny bit better.
And I see: That's why flux is always 3D.

ProcessLib/HT/HTFEM.h
```diff
@@ -88,6 +88,73 @@ class HTFEM : public HTLocalAssemblerInterface
         return Eigen::Map<const Eigen::RowVectorXd>(N.data(), N.size());
     }
 
+    /// Computes the flux in the point \c pnt_local_coords that is given in
+    /// local coordinates using the values from \c local_x.
+    // TODO add time dependency
+    Eigen::Vector3d getFlux(
```
chleh: For the HT process the problem is that there are potentially two different interesting fluxes:
The mass/volume flux and the heat flux.
Maybe there should be different names for them?

ProcessLib/HT/HTFEM.h
endJunction: Regarding the flux being 3D always: I thought it would be easier in general to be it of constant size, like the node coordinates.

`flux.head<GlobalDim>()` would be better, that's true!

ProcessLib/HT/HTFEM.h
endJunction: Or one could pass the variable_id to getFlux?

TomFischer: You are right, there are potentially two different interesting fluxes. Maybe other processes have also interesting quantities fluxes. For this reason we would need rather general names. I opened an issue for discussion and for the time being I would keep the implementation of this PR. Is this okay for you?

ProcessLib/HT/HTFEM.h
TomFischer: Yes, I think so.

TomFischer:
Added the time to all getFlux and the balance functions, see commit a823565. For the second bigger comment I added the issue https://github.com/ufz/ogs/issues/2134 for discussion.

ProcessLib/HT/HTFEM.h
chleh: OK.

PR #2135, opened 2018-05-31
Remove MeshSubsets

endJunction:
~This is a preview, there is still a small memory leak in this (commented), but the remaining parts will stay as they are now.~ Memory leaks fixed.

A never-used feature: a variable defined on different meshes.
While removing the `MeshSubsets`, the `MeshSubset` class was modified too with notable changes:
 - No memory management of the vectors. Now the node vectors are managed solely by the caller/creator of the `MeshSubset`.
 - Subsetting via elements is removed, since it was not used anywhere.

The [first part](https://github.com/ufz/ogs/compare/master...78820ba) deals with the `MeshSubsets` removal, while the [second part](https://github.com/ufz/ogs/compare/78820ba...a2fc618) fixes a memory leak and updates the `MeshSubset`.

Preparation to named meshes created from geometries. The creation happens now in BCs (Neumann).


endJunction:
The failing jenkins/mac job is due to missing files on the website: #2136.

ProcessLib/Process.h
```diff
@@ -251,7 +251,7 @@ class Process
 
 protected:
     MeshLib::Mesh& _mesh;
-    std::unique_ptr<MeshLib::MeshSubset const> _mesh_subset_all_nodes;
+    std::unique_ptr<MeshLib::MeshSubset> _mesh_subset_all_nodes;
```
chleh: not const anymore?

ProcessLib/BoundaryCondition/GenericNonuniformNaturalBoundaryCondition.h
```diff
@@ -46,7 +46,7 @@ class GenericNonuniformNaturalBoundaryCondition final : public BoundaryCondition
 
     std::unique_ptr<MeshLib::Mesh> _boundary_mesh;
 
-    std::unique_ptr<MeshLib::MeshSubset const> _mesh_subset_all_nodes;
+    std::unique_ptr<MeshLib::MeshSubset> _mesh_subset_all_nodes;
```
chleh: Minor: Is the constness change necessary?

ProcessLib/BoundaryCondition/NormalTractionBoundaryCondition.h
```diff
@@ -57,7 +57,7 @@ class NormalTractionBoundaryCondition final : public BoundaryCondition
     /// defined.
     std::vector<MeshLib::Element*> _elements;
 
-    std::unique_ptr<MeshLib::MeshSubset const> _mesh_subset_all_nodes;
+    std::unique_ptr<MeshLib::MeshSubset> _mesh_subset_all_nodes;
```
chleh: likewise

ProcessLib/BoundaryCondition/GenericNaturalBoundaryCondition-impl.h
```diff
@@ -55,8 +55,8 @@ GenericNaturalBoundaryCondition<BoundaryConditionData,
     auto const& mesh_subset =
         dof_table_bulk.getMeshSubset(variable_id, component_id);
 
-    MeshLib::MeshSubset bc_mesh_subset =
-        mesh_subset.getIntersectionByNodes(nodes);
+    _nodes_subset = nodesNodesIntersection(mesh_subset.getNodes(), nodes);
+    MeshLib::MeshSubset bc_mesh_subset(mesh_subset.getMesh(), &_nodes_subset);
```
chleh: Wouldn't it be better to store the nodes set as a `shared_ptr` inside the subset instead of stroing it in the caller?

endJunction:
@chleh Thank you for the fast review. I'll try to add back the constness and update the comments.

The nodes vector is usually the same as used by the mesh, and only a const ref is stored in the mesh subset. The only different story is the node vector for bcs, but with introduction of boundary meshes, it will also be the same node vector as used by the (boundary) mesh.
LIE is the only special case.

Tests/NumLib/LocalToGlobalIndexMapMultiComponent.cpp
```diff
@@ -141,15 +143,18 @@ class NumLibLocalToGlobalIndexMapMultiDOFTest : public ::testing::Test
                   compute_global_index);
 
     std::unique_ptr<const MeshLib::Mesh> mesh;
-    std::unique_ptr<const MeL::MeshSubset> mesh_items_all_nodes;
+    std::unique_ptr<MeL::MeshSubset> mesh_items_all_nodes;
```
wenqing: You may have the same reason as that in above to remove const

Tests/NumLib/LocalToGlobalIndexMapMultiComponent.cpp
```diff
@@ -141,15 +143,18 @@ class NumLibLocalToGlobalIndexMapMultiDOFTest : public ::testing::Test
...
 
     GeoLib::GEOObjects geo_objs;
 
     std::unique_ptr<NL::LocalToGlobalIndexMap> dof_map;
     std::unique_ptr<NL::LocalToGlobalIndexMap> dof_map_boundary;
 
-    std::unique_ptr<MeL::MeshSubset const> mesh_items_boundary;
+    std::unique_ptr<MeL::MeshSubset> mesh_items_boundary;
```
wenqing: Here the same

MeshLib/MeshSubset.h
```diff
@@ -22,61 +22,38 @@
...
 
-    /// return the total number of mesh items
-    std::size_t getNumberOfTotalItems() const
-    {
-        return getNumberOfNodes() + getNumberOfElements();
-    }
+    return active_nodes;
+}
+
+/// A subset of nodes on a single mesh.
```
wenqing: We may use elements as a MeshSubset to perform deactivate/activating elements for different processes in a coupling or for excavation modelling. If only nodes are kept in this class, the class name, I think, it can be termed as  MeshNodeSubset. Then if we need the functionality  of deactivate/activating elements, we can introduce a class of  MeshElementSubset.

endJunction: I agree with renaming in general. I'd postpone it until we have the hypothetical MeshElementSubset because the rename touches 43 files.

chleh:
Thanks for the changes and the comment on the nodes vector.

PR #2140, opened 2018-06-08
Move BC meshes; preparations

endJunction:
A set of basic changes required for the actual "Move BC meshes #2141" PR.

### DIFF
--- a/MeshGeoToolsLib/BoundaryElementsAtPoint.cpp
+++ b/MeshGeoToolsLib/BoundaryElementsAtPoint.cpp
@@ -46,6 +46,16 @@ BoundaryElementsAtPoint::BoundaryElementsAtPoint(
     _boundary_elements.push_back(new MeshLib::Point{nodes, node_ids[0]});
     for (auto const* bulk_element : _mesh.getNode(node_ids[0])->getElements())
     {
+        // Special treatment of the non-line elements because the identifyFace
+        // implementation expects three valid node pointers in the call. We can
+        // guarantee only one valid node, and calling the identifyFace would
+        // most likely lead to a SEGV.
+        if (bulk_element->getGeomType() != MeshLib::MeshElemType::LINE)
+        {
+            _bulk_ids.emplace_back(bulk_element->getID(), -1);
+            continue;
+        }
+
         _bulk_ids.emplace_back(
             bulk_element->getID(),
             bulk_element->identifyFace(mesh_nodes.data() + node_ids[0]));
--- a/MeshLib/MeshSubset.h
+++ b/MeshLib/MeshSubset.h
@@ -53,7 +53,41 @@ public:
     /// \param vec_items Vector of Node pointers.
     MeshSubset(const Mesh& msh, std::vector<Node*> const& vec_items)
         : _msh(msh), _nodes(vec_items)
-    {}
+    {
+        // If the mesh nodes and the given nodes point to the same vector, they
+        // must be equal.
+        if (&_msh.getNodes() == &_nodes)
+        {
+            return;
+        }
+
+        //
+        // Testing if the given nodes belong to the mesh.
+        //
+        {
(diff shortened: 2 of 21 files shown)


MeshGeoToolsLib/BoundaryElementsAtPoint.cpp
```diff
@@ -46,6 +46,16 @@ BoundaryElementsAtPoint::BoundaryElementsAtPoint(
     _boundary_elements.push_back(new MeshLib::Point{nodes, node_ids[0]});
     for (auto const* bulk_element : _mesh.getNode(node_ids[0])->getElements())
     {
+        // Special treatment of the non-line elements because the identifyFace
+        // implementation expects three valid node pointers in the call. We can
+        // guarantee only one valid node, and calling the identifyFace would
+        // most likely lead to a SEGV.
```
chleh: Maybe I don't understand it right, but isn't that necessary only for points then?

MeshLib/MeshSubset.h
```diff
@@ -53,7 +53,24 @@ class MeshSubset
     /// \param vec_items Vector of Node pointers.
     MeshSubset(const Mesh& msh, std::vector<Node*> const& vec_items)
         : _msh(msh), _nodes(vec_items)
-    {}
+    {
+        // testing if the given nodes belong to the mesh.
+        auto node_is_part_of_mesh =
+            [mesh_nodes = _msh.getNodes()](MeshLib::Node* const& n) {
+                auto it = std::find(begin(mesh_nodes), end(mesh_nodes), n);
```
chleh: This is a pointer comparison. Is that correct?

MeshLib/MeshSubset.h
```diff
@@ -53,7 +53,24 @@ class MeshSubset
...
+                if (it == end(mesh_nodes))
+                {
+                    ERR("A node %d in mesh subset is not a part of the mesh.",
+                        n->getID());
+                    return false;
+                }
+                return true;
+            };
+        if (!std::all_of(begin(_nodes), end(_nodes), node_is_part_of_mesh))
```
chleh: The runtime complexity of that is `length(mesh_nodes) * length(mesh_subset)` isn't that terribly slow?

MeshLib/MeshSubset.h
endJunction: Yes. Just a logical check (not geometrical) if the subset of nodes is indeed a part of the mesh nodes vector.

MeshLib/MeshSubset.h
endJunction: You are right, although it is not that slow that it would have drawn my attention to it.
I'll update for more optimized version in this PR. The special case (also very fast to check) is the whole mesh subset, where the nodes are the same as the mesh's nodes.

MeshGeoToolsLib/BoundaryElementsAtPoint.cpp
endJunction: It is for points only (see function name: boundaryElementsAtPoint), but maybe I missed you point ;)

Note to myself: a similar case happens when searching for lines (having two points) in an element's face... but the case didn't happen yet.

MeshLib/MeshSubset.h
endJunction: Improving it, tested with GWF/square_1e5_neumann.prj

ufz/master time is 2.88s
the unsorted vector naive implementation runs in 4.95s
the sorted vector with binary search runs in 2.90s
and checking for the vector addresses before expensive search is indistinguishable from the ufz/master.

Taking the last one :)

MeshGeoToolsLib/BoundaryElementsAtPoint.cpp
endJunction: if there is something more to discuss, I'll prepare another PR, but for now we have to move on...

PR #2154, opened 2018-06-27
[web] OGS-5 page

bilke:
As requested by Marc: a short info page about OGS-5 with download links.


OlafKolditz:
This new branch is dedicated to OGS teaching activities at our partner universities (indicated as OGS Teaching). The first examples come from the Hydroinformatics lectures at TU Dresden. Interested students may jointly work on the Finite Difference Method exercises:
- Diffusion equation (1D)
- Open channel flow (1D)
- Groundwater flow (2D)

endJunction:
I'd say it's subtle enough. Maybe there could be a more prominent section in 'releases'.

bilke:
@endJunction Yes, with download page I actually meant 'releases'.

web/config.toml
```diff
@@ -29,6 +29,12 @@ staticDir = ["dist", "static"]
   identifier = "docs"
   weight = 4
 
+[[menu.main]]
+  name = "OGS-5"
```
TomFischer: Maybe in the top bar: 

 - OpenGeoSys__-6__ 
 - OpenGeoSys-5

for consistent naming.

rinkk:
I think this is confusing to everyone not familiar with OGS development. If it was up to me I would

* rename "OpenGeoSys-6" to "OpenGeoSys" (links to the start page anyway, so there's not really a reason to mention the "6"
* remove the "OpenGeoSys-5" link in the menu bar
* link to the OGS5 page from Releases and whereever else it is necessary (some books are based on OGS5 and there might be some docs, etc.)

Anyone familiar with OGS will know what to look for or ask if they don't find it.

Everyone else should be directed to the new version and not be confused by competing versions when visiting the page for the first time. The OGS-5 development has been important, there are things that can (currently) only be done with 5 and it should be kept available to anyone who needs it. But I don't think we are doing ourselves a favour by putting both on the front page.

bilke:
@rinkk I fully agree. Will make some more changes on Monday.

PR #2157, opened 2018-07-03
Added an element test as a LIE cohesive zone benchmark

renchao-lu:
This pull request is motivated to enrich secondary variables in the LIE/SmallDeformation as in the LIE/HydroMechanics, and add an element test for LIE cohesive zone model. The extended secondary variables consist of a fracture opening vector (w_n and w_s) and a fracture stress vector (f_stress_normal and f_stress_shear). With these output variables, it becomes more straightforward to make a comparison against the analytical solution in the element test.


renchao-lu:
- [to do] Clang-formatting

endJunction:
Few more small things:
 - remove the executable flag from test files, _e.g._ Tests/Data/LIE/Mechanics/cohesive_zone_load_path.gml (`chmod a-x Tests/Data/LIE/Mechanics/cohesive_zone_load_path.gml`)
 - remove the pvd file from tests

ProcessLib/LIE/SmallDeformation/Tests.cmake
```diff
@@ -89,10 +89,14 @@ AddTest(
     TESTER vtkdiff
     REQUIREMENTS NOT OGS_USE_MPI
     DIFF_DATA
-    GLOB cohesive_zone_load_path_pcs_0_ts_*.vtu displacement displacement 1e-16 0
```
endJunction: Put back the comparison of displacement, displacement_jump1, and sigma_*.
Use GLOB pattern, instead of writing out each time step.
The tolerances for the stress_n are rather loose; how come?

ProcessLib/LIE/SmallDeformation/SmallDeformationProcess.cpp
```diff
@@ -356,6 +356,18 @@ void SmallDeformationProcess<DisplacementDim>::initializeConcreteProcess(
         }
     }
 
+    auto mesh_prop_w_n = MeshLib::getOrCreateMeshProperty<double>(
+        const_cast<MeshLib::Mesh&>(mesh), "w_n",
+        MeshLib::MeshItemType::Cell, 1);
+    mesh_prop_w_n->resize(mesh.getNumberOfElements());
```
endJunction: The resize is not necessary. It's already done in the MeshLib::getOrCreateMeshProperty...

ProcessLib/LIE/SmallDeformation/SmallDeformationProcessData.h
```diff
@@ -94,6 +94,10 @@ struct SmallDeformationProcessData
 
     // mesh property for fracture aperture
     MeshLib::PropertyVector<double>* _mesh_prop_b = nullptr;
+    MeshLib::PropertyVector<double>* _mesh_prop_w_n = nullptr;
+    MeshLib::PropertyVector<double>* _mesh_prop_w_s = nullptr;
+    MeshLib::PropertyVector<double>* _mesh_prop_fracture_stress_shear = nullptr;
+    MeshLib::PropertyVector<double>* _mesh_prop_fracture_stress_normal = nullptr;
```
endJunction: I'd try to use either `_n` and `_s` or `_normal` and `_shear`, but not to mix the both.

ProcessLib/LIE/SmallDeformation/Tests.cmake
renchao-lu: The GLOB pattern sounds like making the file comparison with an identical file name in different folder. I am not sure.

endJunction: Yes, it does compare the files from the source folder (expected output) to the build folder (ctest output) and the files are named identically. Simply do as for displacement etc.

renchao-lu: The expected output contains the analytical solution 'f_stress_n_analytical' which is not included in the ctest output though...

renchao-lu:
I have locally run the failed LIE benchmarks in the series of sequential tests. The below are the error messages

Debugging starts
info: This is OpenGeoSys-6 version 6.1.0-681-gfd7ab9d.
info: OGS started on 2018-07-17 14:31:32+0200.
debug: Project configuration from file 'single_joint.prj' read.
ERROR: In /home/renchaolu/.conan/data/VTK/8.1.1/bilke/stable/build/3ae6769b6d7a0bea12b51d821c046d5f4d7c7298/source_subfolder/IO/XMLParser/vtkXMLParser.cxx, line 399
vtkXMLDataParser (0x5e98340): Error parsing XML in stream at line 1, column 0, byte index 0: syntax error

ERROR: In /home/renchaolu/.conan/data/VTK/8.1.1/bilke/stable/build/3ae6769b6d7a0bea12b51d821c046d5f4d7c7298/source_subfolder/IO/XML/vtkXMLReader.cxx, line 491
vtkXMLUnstructuredGridReader (0x5ead280): Error parsing input file.  ReadXMLInformation aborting.

ERROR: In /home/renchaolu/.conan/data/VTK/8.1.1/bilke/stable/build/3ae6769b6d7a0bea12b51d821c046d5f4d7c7298/source_subfolder/Common/ExecutionModel/vtkExecutive.cxx, line 784
vtkCompositeDataPipeline (0x5ea46e0): Algorithm vtkXMLUnstructuredGridReader(0x5ead280) returned failure for request: vtkInformation (0x5ea51a0)
  Debug: Off
  Modified Time: 76
  Reference Count: 1
  Registered Events: (none)
  Request: REQUEST_INFORMATION
  FORWARD_DIRECTION: 0
  ALGORITHM_AFTER_FORWARD: 1

error: Mesh "single_joint_beta50.vtu" contains zero points.
error: Could not read mesh from 'single_joint_beta50.vtu' file. No mesh added. at ProjectData.cpp, line 114
info: OGS terminated on 2018-07-17 14:31:32+020

Does anyone have some idea?

ProcessLib/LIE/SmallDeformation/LocalAssembler/SmallDeformationLocalAssemblerInterface.h
```diff
@@ -90,6 +90,21 @@ class SmallDeformationLocalAssemblerInterface
             "is not implemented");
     }
 
+    void computeSecondaryVariableConcrete(
+        const double t, std::vector<double> const& local_x_) override
+    {
+        if (!_dofIndex_to_localIndex.empty())
+        {
+            auto const local_dof_size = local_x_.size();
```
wenqing: You can directly use local_x_.size() in the following for loop.

ProcessLib/LIE/SmallDeformation/LocalAssembler/SmallDeformationLocalAssemblerInterface.h
```diff
@@ -90,6 +90,21 @@ class SmallDeformationLocalAssemblerInterface
...
+
+            _local_u.setZero();
+            for (unsigned i = 0; i < local_dof_size; i++)
```
wenqing: Change unsigned to std::size_t.

ProcessLib/LIE/SmallDeformation/LocalAssembler/SmallDeformationLocalAssemblerInterface.h
renchao-lu: Replaced accordingly.

ProcessLib/LIE/SmallDeformation/LocalAssembler/SmallDeformationLocalAssemblerInterface.h
renchao-lu: Changed accordingly.

ProcessLib/LIE/SmallDeformation/SmallDeformationProcess.h
```diff
@@ -45,6 +44,9 @@ class SmallDeformationProcess final : public Process
     bool isLinear() const override;
     //! @}
 
+    void computeSecondaryVariableConcrete(double const t,
```
chleh: Sorry for the late comment. AFAIK `computeSecondaryVariableConcrete()` is called after every single iteration.
In contrast `postTimestepConcreteProcess()` is called only after each timestep, so probably much less often. You seem to use the secondary variables only for output. Wouldn't `postTimestepConcreteProcess()` therefore suffice?
The only corner case that I am aware of is the output of iteration results. Or am I missing something?
For later developments, of course a `preOutput()` method would be a good option.

endJunction: I thought it would be the other way around, but it looks like the compute secondary vars is called just after postTimestep [UncoupledTimeLoop.cpp](https://github.com/ufz/ogs/blob/master/ProcessLib/UncoupledProcessesTimeLoop.cpp#L684).

I agree with the need for preOutput(). --> another PR.

I'd suggest to keep this implementation for now.

chleh: Yep, implementation can be kept. Then I was wrong. Maybe then in the future the two methods can be combined.

endJunction:
@renchao-lu I've seen you want to add a python-script to the tests/data, that's great! Just add it when you are ready as another PR.

PR #2163, opened 2018-07-17
[appveyor] Bumped Visual Studio to 2017, never build Conan packages.

bilke:
VTK 8.1 binary packages are built for 2017 only.


bilke:
@endJunction Can be merged once green.

PR #2164, opened 2018-07-18
[Solids] Added an arguments of T to the class member of integrateStress

wenqing:
as titled.

### DIFF
--- a/MaterialLib/PhysicalConstant.h
+++ b/MaterialLib/PhysicalConstant.h
@@ -20,6 +20,9 @@ namespace PhysicalConstant
 /// Zero degrees Celsius in Kelvin
 const double CelsiusZeroInKelvin = 273.15;
 
+/// Room temperature, which is used as a reference temperature.
+const double RoomTemperature = CelsiusZeroInKelvin + 20.0;
+
 /**
   Ideal gas constant in SI standard units (J \per{mol} \per{K})
 
--- a/MaterialLib/PhysicalConstant.h
+++ b/MaterialLib/PhysicalConstant.h
@@ -20,9 +20,6 @@ namespace PhysicalConstant
 /// Zero degrees Celsius in Kelvin
 const double CelsiusZeroInKelvin = 273.15;
 
-/// Room temperature, which is used as a reference temperature.
-const double RoomTemperature = CelsiusZeroInKelvin + 20.0;
-
 /**
   Ideal gas constant in SI standard units (J \per{mol} \per{K})
 
--- a/ProcessLib/HydroMechanics/CreateHydroMechanicsProcess.cpp
+++ b/ProcessLib/HydroMechanics/CreateHydroMechanicsProcess.cpp
@@ -183,7 +183,8 @@ std::unique_ptr<Process> createHydroMechanicsProcess(
     // Reference temperature
     const auto& reference_temperature =
         //! \ogs_file_param{prj__processes__process__HYDRO_MECHANICS__reference_temperature}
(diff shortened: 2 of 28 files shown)


MaterialLib/PhysicalConstant.h
```diff
@@ -20,6 +20,9 @@ namespace PhysicalConstant
 /// Zero degrees Celsius in Kelvin
 const double CelsiusZeroInKelvin = 273.15;
 
+/// Room temperature, which is used as a reference temperature.
+const double RoomTemperature = CelsiusZeroInKelvin + 20.0;
```
endJunction: Better to call it nist_standard_temperature, or normal_temperature, or NTP_temperature;
Not this PR, but the CelsiusZeroInKelvin should be celsius_zero_in_kelvin too.

ProcessLib/HydroMechanics/HydroMechanicsFEM.h
```diff
@@ -76,7 +77,8 @@ struct IntegrationPointData final
     {
         auto&& solution = solid_material.integrateStress(
             t, x_position, dt, eps_prev, eps, sigma_eff_prev,
-            *material_state_variables);
+            *material_state_variables,
+            MaterialLib::PhysicalConstant::RoomTemperature, 0.);
```
endJunction: I'd move it out into configuration file, because there are many different standards for temperature.
Also when using dimensionless form of equations, the absolute value is a no-go.

endJunction: I also understand, that the last 0 function argument is pressure, but 0Pa pressure is not used in any standard.
In *M we explicitly do have the pressure of the system.

wenqing: Do you mean to set it as an input parameter?

endJunction: yes, like the "reference_temperature" in TM.

MaterialLib/PhysicalConstant.h
chleh: Minor comment: All of PhysicalConstant.h is in CamelCase. IMHO we should keep that convention.

wenqing: Removed.

ProcessLib/HydroMechanics/HydroMechanicsFEM.h
wenqing: Now,  reference_temperature is set an optional input, and reference_fluid_pressure is dropped because it is only for some special models for unsaturated flow processes.

ProcessLib/HydroMechanics/CreateHydroMechanicsProcess.cpp
```diff
@@ -189,7 +194,8 @@ std::unique_ptr<Process> createHydroMechanicsProcess(
         biot_coefficient,
         porosity,
         solid_density,
-        specific_body_force};
+        specific_body_force,
+        *reference_temperature};
```
endJunction: I think this is undefined behaviour in case the config parameter is empty.
https://en.cppreference.com/w/cpp/utility/optional/operator*

wenqing: Do you mean to change the type of the argument to std::optional?

wenqing: or something like
```
(reference_temperature) ? *reference_temperature : [a default value]
```

PR #2165, opened 2018-07-18
[CMake] Updated cmake-conan to 0.12.

bilke:


chleh:
:+1:
Is it just an update to the latest version or does it solve some issues? Could this update lead to problems (e.g., cmake errors upon rebase)?

bilke:
It is just an update, but I wanted to have it as it allows for using then [virtualrunenv](
https://docs.conan.io/en/latest/mastering/virtualenv.html#virtualrunenv-generator)-generator. See my last commit.

PR #2171, opened 2018-08-01
Improve TM process

wenqing:
Use the temperature increment dT instead of the global temperature change T-T_0 in the thermal strain calculation. Also this PR corrects the thermal strain calculation with initial temperature.

In the current master, the thermal strain increment is calculated by 
dt0
Such calculation is simplified in  this PR as
dt1
and the parameter of reference_temperature is dropped.
  
Note: In the current repository, the calculation of total strain increment takes the same calculation approach as that for thermal strain increment, which uses extra more memory by storing vectors of eps_prev and eps_m at each Gauss point. This will be improved later for all mechanics related processes.

### DIFF
--- a/ProcessLib/ThermoMechanics/ThermoMechanicsFEM.h
+++ b/ProcessLib/ThermoMechanics/ThermoMechanicsFEM.h
@@ -50,7 +50,8 @@ struct IntegrationPointData final
     std::unique_ptr<typename MaterialLib::Solids::MechanicsBase<
         DisplacementDim>::MaterialStateVariables>
         material_state_variables;
-    double solid_density_current, solid_density_prev;
+    double solid_density;
+    double solid_density_prev;
 
     double integration_weight;
     typename ShapeMatricesType::NodalRowVectorType N;
@@ -60,7 +61,7 @@ struct IntegrationPointData final
     {
         eps_m_prev = eps_m;
         sigma_prev = sigma;
-        solid_density_prev = solid_density_current;
+        solid_density_prev = solid_density;
         material_state_variables->pushBackState();
     }
 
@@ -142,9 +143,9 @@ public:
 
             SpatialPosition x_position;
             x_position.setElementID(_element.getID());
-            ip_data.solid_density_current =
+            ip_data.solid_density =
                 _process_data.reference_solid_density(0, x_position)[0];
-            ip_data.solid_density_prev = ip_data.solid_density_current;
+            ip_data.solid_density_prev = ip_data.solid_density;
 
             ip_data.N = shape_matrices[ip].N;
             ip_data.dNdx = shape_matrices[ip].dNdx;
@@ -241,7 +242,7 @@ public:
             auto const alpha =
                 _process_data.linear_thermal_expansion_coefficient(
                     t, x_position)[0];
-            double const linear_thermal_strain = alpha * dT;
+            double const linear_thermal_strain_increment = alpha * dT;
 
             //
             // displacement equation, displacement part
@@ -255,7 +256,7 @@ public:
             // assume isotropic thermal expansion
             const double T_ip = N.dot(T);  // T at integration point
             eps_m.noalias() =
-                eps - linear_thermal_strain * Invariants::identity2;
+                eps - linear_thermal_strain_increment * Invariants::identity2;
             auto&& solution = _ip_data[ip].solid_material.integrateStress(
                 t, x_position, dt, eps_m_prev, eps_m, sigma_prev, *state, T_ip);
             eps_m.noalias() = eps;
@@ -288,9 +289,9 @@ public:
             // dV = 3 * alpha * dT * V_0
             // rho_s_{n+1} = rho_s_n / (1 + 3 * alpha * dT )
             // see reference solid density description for details.
-            double const rho_s = _ip_data[ip].solid_density_prev /
-                                 (1 + 3 * linear_thermal_strain);
-            _ip_data[ip].solid_density_current = rho_s;
+            auto& rho_s = _ip_data[ip].solid_density;
+            rho_s = _ip_data[ip].solid_density_prev /
+                                 (1 + 3 * linear_thermal_strain_increment);
 
             auto const& b = _process_data.specific_body_force;
             local_rhs
--- a/ProcessLib/ThermoMechanics/ThermoMechanicsFEM.h
+++ b/ProcessLib/ThermoMechanics/ThermoMechanicsFEM.h
@@ -50,6 +50,7 @@ struct IntegrationPointData final
     std::unique_ptr<typename MaterialLib::Solids::MechanicsBase<
         DisplacementDim>::MaterialStateVariables>
         material_state_variables;
+    double solid_density_current, solid_density_prev;
 
     double integration_weight;
     typename ShapeMatricesType::NodalRowVectorType N;
@@ -59,6 +60,7 @@ struct IntegrationPointData final
     {
         eps_m_prev = eps_m;
         sigma_prev = sigma;
+        solid_density_prev = solid_density_current;
         material_state_variables->pushBackState();
     }
 
@@ -137,6 +139,12 @@ public:
             ip_data.eps.setZero(kelvin_vector_size);
             ip_data.eps_m.setZero(kelvin_vector_size);
             ip_data.eps_m_prev.setZero(kelvin_vector_size);
+
+            SpatialPosition x_position;
+            x_position.setElementID(_element.getID());
+            ip_data.solid_density_current =
+                _process_data.reference_solid_density(0, x_position)[0];
+            ip_data.solid_density_prev = ip_data.solid_density_current;
 
             ip_data.N = shape_matrices[ip].N;
             ip_data.dNdx = shape_matrices[ip].dNdx;
@@ -227,14 +235,13 @@ public:
             auto& eps_m_prev = _ip_data[ip].eps_m_prev;
             auto& state = _ip_data[ip].material_state_variables;
 
-            double const delta_T =
-                N.dot(T) - _process_data.reference_temperature;
+            double const dT = N.dot(T_dot) * dt;
             // calculate thermally induced strain
             // assume isotropic thermal expansion
             auto const alpha =
                 _process_data.linear_thermal_expansion_coefficient(
                     t, x_position)[0];
-            double const linear_thermal_strain = alpha * delta_T;
+            double const linear_thermal_strain = alpha * dT;
 
             //
             // displacement equation, displacement part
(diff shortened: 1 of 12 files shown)


ProcessLib/ThermoMechanics/ThermoMechanicsFEM.h
```diff
@@ -246,11 +253,12 @@ class ThermoMechanicsLocalAssembler
                     DisplacementDim>::value>;
 
             // assume isotropic thermal expansion
+            const double T_ip = N.dot(T);  // T at integration point
```
chleh: Very minor note: `N * T` would work, because it's row times column vector.

ProcessLib/ThermoMechanics/ThermoMechanicsFEM.h
```diff
@@ -246,11 +253,12 @@ class ThermoMechanicsLocalAssembler
...
             eps_m.noalias() =
                 eps - linear_thermal_strain * Invariants::identity2;
             auto&& solution = _ip_data[ip].solid_material.integrateStress(
-                t, x_position, dt, eps_m_prev, eps_m, sigma_prev, *state,
-                delta_T + _process_data.reference_temperature);
+                t, x_position, dt, eps_m_prev, eps_m, sigma_prev, *state, T_ip);
+            eps_m.noalias() = eps;
```
chleh: This assignment seems to overwrite the assignment in the line above. Is that intended?

ProcessLib/ThermoMechanics/ThermoMechanicsFEM.h
```diff
@@ -276,13 +284,13 @@ class ThermoMechanicsLocalAssembler
...
+            // rho_s_{n+1} * (V_{n} + dV) = rho_s_n * V_n
             // dV = 3 * alpha * dT * V_0
-            // rho_s = rho_sr / (1 + 3 * alpha * dT )
+            // rho_s_{n+1} = rho_s_n / (1 + 3 * alpha * dT )
             // see reference solid density description for details.
-            auto const rho_sr =
-                _process_data.reference_solid_density(t, x_position)[0];
-            double const rho_s = rho_sr / (1 + 3 * linear_thermal_strain);
+            double const rho_s = _ip_data[ip].solid_density_prev /
+                                 (1 + 3 * linear_thermal_strain);
```
chleh: Not introduced in this PR, but is the number 3 related to the space dimension?

ProcessLib/ThermoMechanics/ThermoMechanicsFEM.h
```diff
@@ -227,14 +235,13 @@ class ThermoMechanicsLocalAssembler
...
-            double const delta_T =
-                N.dot(T) - _process_data.reference_temperature;
+            double const dT = N.dot(T_dot) * dt;
             // calculate thermally induced strain
             // assume isotropic thermal expansion
             auto const alpha =
                 _process_data.linear_thermal_expansion_coefficient(
                     t, x_position)[0];
-            double const linear_thermal_strain = alpha * delta_T;
+            double const linear_thermal_strain = alpha * dT;
```
chleh: Maybe a comment in the code would be appropriate that the linear thermal strain now is w.r.t. the state from the previous timestep and not w.r.t. a reference temperature (if I understood that correctly).

Tests/Data/ThermoMechanics/iglu_axisymmetric_plane_strain_quad.prj
```diff
@@ -237,9 +236,9 @@
             <lis>-i bicgstab -p jacobi -tol 1e-11 -maxiter 10000</lis>
             <eigen>
                 <solver_type>BiCGSTAB</solver_type>
-                <precon_type>DIAGONAL</precon_type>
```
chleh: A stronger preconditioner? Did the condition of the equation systems degrade in this PR?

ProcessLib/ThermoMechanics/ThermoMechanicsFEM.h
```diff
@@ -50,6 +50,7 @@ struct IntegrationPointData final
     std::unique_ptr<typename MaterialLib::Solids::MechanicsBase<
         DisplacementDim>::MaterialStateVariables>
         material_state_variables;
+    double solid_density_current, solid_density_prev;
```
endJunction: I'd write one definition per line...

chleh:
> Note:  In the current repository, the calculation of total strain increment  takes the same calculation approach as that for thermal strain  increment, which uses extra more memory by storing vectors of eps_prev  and eps_m at each Gauss point. This will be improved later for all  mechanics related processes.

I assume that this design (the extra memory for eps and eps_prev) was done on purpose by @nagelt and @endJunction to allow for most general mechanical material models. Maybe they should be asked before optimizing/simplifying their design.

ProcessLib/ThermoMechanics/ThermoMechanicsFEM.h
endJunction: Not sure, if a 1x1 matrix object is convertible to double....

ProcessLib/ThermoMechanics/ThermoMechanicsFEM.h
```diff
@@ -276,13 +284,13 @@ class ThermoMechanicsLocalAssembler
...
             // dV = 3 * alpha * dT * V_0
-            // rho_s = rho_sr / (1 + 3 * alpha * dT )
+            // rho_s_{n+1} = rho_s_n / (1 + 3 * alpha * dT )
             // see reference solid density description for details.
-            auto const rho_sr =
-                _process_data.reference_solid_density(t, x_position)[0];
-            double const rho_s = rho_sr / (1 + 3 * linear_thermal_strain);
+            double const rho_s = _ip_data[ip].solid_density_prev /
+                                 (1 + 3 * linear_thermal_strain);
+            _ip_data[ip].solid_density_current = rho_s;
```
endJunction: I'd rather write it as
```c++
auto& rho_s = _ip_data[ip].solid_density;

rho_s = ...;
```

ProcessLib/ThermoMechanics/ThermoMechanicsFEM.h
wenqing: Then I keep it.

ProcessLib/ThermoMechanics/ThermoMechanicsFEM.h
wenqing: Yes, it is intended.  Inside integrateStress, eps_m and  eps_m_prev is used to calculate d eps^e such that
``d eps^e =  eps_m - eps_m_prev``. In the previous implementation,   ``eps_m = B u^{n+1} - \alpha_T (T^{n+1}-T_{ref}) I``and  ``eps_m_prev = B u^{n} - \alpha_T (T^{n}-T_{ref}) I`` (by pushBackState()). With the current change,    ``eps_m = B u^{n+1} - \alpha_T dT I``, and ``eps_m_prev = B u^{n}``. In order to update ``eps_m_prev = B u^{n}`` by pushBackState(), we have to have the assignment.

This is also the reason that I mentioned in the introduction that eps_m and eps_m_prev can removed from _ip_data even under large deformation. We only need the strain increment in the stress integration. Since we already have access to displacement and the increment of displacement,  the strain increment can be computed directly without storing two step strains.

ProcessLib/ThermoMechanics/ThermoMechanicsFEM.h
wenqing: We set input of the thermal expansion, alpha_T,  as line thermal expansion. The means it is the  thermal expansion for strain component as 
d epsilon^T_{xx} = alpha_T dT
d epsilon^T_{yy} = alpha_T dT
d epsilon^T_{zz} = alpha_T dT

For the volume change due to the thermal expansion, we have
d epsilon^T_{xx} + d epsilon^T_{yy} + d epsilon^T_{zz} = 3 alpha_T dT.
This is valid for 1D, 2D (either plane strain or plane stress) and 3D problem.

ProcessLib/ThermoMechanics/ThermoMechanicsFEM.h
wenqing: Changed the name to `linear_thermal_strain_increment`.

ProcessLib/ThermoMechanics/ThermoMechanicsFEM.h
chleh: Thanks for the explanation. For me (as a non-expert), the explanation would also be very helpful directly in the code. But maybe people who know the topic don't need it, and maybe it's already said in some PDF explaining the governing equations.

Basically, the cause for my confusion was that I read the lines 257-260 wrongly. I thought, they were one statement, but actually they are two. Thanks anyway.

> This is also the reason that I mentioned in the introduction that eps_m and eps_m_prev can removed from _ip_data even under large deformation. We only need the strain increment in the stress integration. Since we already have access to displacement and the increment of displacement, the strain increment can be computed directly without storing two step strains.

I thought, the design choice was plasticity-related. But maybe I'm wrong.

Tests/Data/ThermoMechanics/iglu_axisymmetric_plane_strain_quad.prj
wenqing: For this benchmark,  the condition of the equation systems is sensitive to the present change in the source code. I could either decrease the tolerance the convergence of the linear solver or change the preconditioner, and I chose the later. Note, the equation systems of TM by the monolithic scheme is unsymmetrical.

ProcessLib/ThermoMechanics/ThermoMechanicsFEM.h
wenqing: Changed accordingly.

ProcessLib/ThermoMechanics/ThermoMechanicsFEM.h
wenqing: Changed accordingly.

chleh:
:+1:

ProcessLib/ThermoMechanics/ThermoMechanicsFEM.h
chleh: OK, thanks. But in shouldn't it be different in plane strain, when all deformation is confined to two dimensions, or am I confusing something, here?

wenqing: In 2D plane strain, e
which means
e2
In 2D plane stress (we do not need it)
\dot \epsilon_{zz} != 0, we still have
e3
where the superscript e means elastic, and T mean thermal.

ProcessLib/ThermoMechanics/ThermoMechanicsFEM.h
wenqing: For plastic model, we only need the total strain increment to be passed into integrateStress. Of course, I would not change the code for this before a discussion.

ProcessLib/ThermoMechanics/ThermoMechanicsFEM.h
wenqing: Your confusing leads a correction to volume thermal strain under  2D plane strain condition. Thanks. I will  change it to `1 + DisplacementDim  * linear_thermal_strain`.

wenqing: I withdraw my last answer.

chleh: As written two lines above the "offending" one I probably should `// see reference solid density description for details.`
Thanks for your explanations so far. If I'll have some time, I'll dive into the details to fully understand the formula.

ProcessLib/ThermoMechanics/ThermoMechanicsFEM.h
endJunction: https://godbolt.org/g/QhizDv

1x1 matrix is not convertible to double.

PR #2173, opened 2018-08-02
Fixed issue with limiting the date range on time series data

rinkk:
When visualising a time series as a diagram, limiting the time range works as intended now.

Also fixed a number of implicit data type conversion warnings.

### DIFF
--- a/Applications/DataExplorer/DataView/DiagramView/CMakeLists.txt
+++ b/Applications/DataExplorer/DataView/DiagramView/CMakeLists.txt
@@ -10,6 +10,7 @@ set(SOURCES
 
 set(HEADERS
     DiagramList.h
+    getDateTime.h
     QArrow.h
     QGraphicsGrid.h
     DiagramScene.h
--- a/Applications/DataExplorer/DataView/DiagramView/CMakeLists.txt
+++ b/Applications/DataExplorer/DataView/DiagramView/CMakeLists.txt
@@ -10,7 +10,7 @@ set(SOURCES
 
 set(HEADERS
     DiagramList.h
-    getDateTime.h
+    GetDateTime.h
     QArrow.h
     QGraphicsGrid.h
     DiagramScene.h
--- a/Applications/DataExplorer/DataView/DiagramView/DiagramList.cpp
+++ b/Applications/DataExplorer/DataView/DiagramView/DiagramList.cpp
@@ -13,7 +13,7 @@
  */
 
 #include "DiagramList.h"
-#include "getDateTime.h"
+#include "GetDateTime.h"
 
 #include <logog/include/logog.hpp>
 
--- a/Applications/DataExplorer/DataView/DiagramView/DiagramList.cpp
+++ b/Applications/DataExplorer/DataView/DiagramView/DiagramList.cpp
@@ -171,9 +171,6 @@ int DiagramList::readList(const QString &path, std::vector<DiagramList*> &lists)
         }
 
         bool first_loop(true);
-        int numberOfSecs(0);
-        double value(0);
-        QString stringDate("");
         QDateTime startDate, currentDate;
         unsigned line_count (1);
 
@@ -184,7 +181,7 @@ int DiagramList::readList(const QString &path, std::vector<DiagramList*> &lists)
             fields = line.split('\t');
             if (fields.size() >= (nLists + 1))
             {
-                stringDate = fields.takeFirst();
+                QString const stringDate = fields.takeFirst();
                 currentDate = getDateTime(stringDate);
                 if (first_loop)
                 {
@@ -194,15 +191,12 @@ int DiagramList::readList(const QString &path, std::vector<DiagramList*> &lists)
                     first_loop = false;
                 }
 
-                numberOfSecs = startDate.secsTo(currentDate);
-
+                float const numberOfSecs = static_cast<float>(startDate.secsTo(currentDate));
                 for (int i = 0; i < nLists; i++)
                 {
-                    value =
-                        strtod(BaseLib::replaceString(
-                                   ",", ".", fields.takeFirst().toStdString())
-                                   .c_str(),
-                               nullptr);
+                    float const value = static_cast<float>(strtod(
+                        BaseLib::replaceString(",", ".", fields.takeFirst().toStdString()).c_str(),
+                        nullptr));
                     lists[i]->addNextPoint(numberOfSecs,value);
                 }
             }
@@ -268,10 +262,10 @@ int DiagramList::readList(const SensorData* data, std::vector<DiagramList*> &lis
             l->setXUnit("day");
             QDateTime startDate(getDateTime(QString::fromStdString(BaseLib::int2date(time_steps[0]))));
             lists[i]->setStartDate(startDate);
-            int numberOfSecs(0);
             for (std::size_t j = 0; j < nValues; j++)
             {
-                numberOfSecs = startDate.secsTo(getDateTime(QString::fromStdString(BaseLib::int2date(time_steps[j]))));
+                QString const data_str = QString::fromStdString(BaseLib::int2date(time_steps[j]));
+                float numberOfSecs = static_cast<float>(startDate.secsTo(getDateTime(data_str)));
                 lists[i]->addNextPoint(numberOfSecs, (*time_series)[j]);
             }
         }
@@ -279,7 +273,8 @@ int DiagramList::readList(const SensorData* data, std::vector<DiagramList*> &lis
         {
             l->setXUnit("time step");
             for (std::size_t j = 0; j < nValues; j++)
-                lists[i]->addNextPoint(time_steps[j], (*time_series)[j]);
+                lists[i]->addNextPoint(static_cast<float>(time_steps[j]),
+                                       (*time_series)[j]);
         }
 
         lists[i]->update();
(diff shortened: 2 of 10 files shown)


Applications/DataExplorer/DataView/DiagramView/DiagramList.cpp
```diff
@@ -288,19 +283,38 @@ int DiagramList::readList(const SensorData* data, std::vector<DiagramList*> &lis
...
+    _coords.erase(std::remove_if(_coords.begin(), _coords.end(),
+        [&](std::pair<float, float> const& c){return (c.first<start_secs || c.first>end_secs);}),
+        _coords.end());
+    _startDate = start;
+    for (auto& c : _coords)
+        c.first -= start_secs;
+    update();
+}
 
+void DiagramList::setList(std::vector< std::pair<QDateTime, float> > coords)
```
TomFischer: Do you really want to copy the vector? Maybe `const &`?

Applications/DataExplorer/DataView/DiagramView/DiagramList.cpp
```diff
@@ -288,19 +283,38 @@ int DiagramList::readList(const SensorData* data, std::vector<DiagramList*> &lis
...
     std::size_t nCoords = coords.size();
     for (std::size_t i = 1; i < nCoords; i++)
     {
-        numberOfDays = this->_startDate.daysTo(coords[i].first);
-        _coords.emplace_back(static_cast<float>(numberOfDays),
-                             coords[i].second);
+        _coords.emplace_back(
+            static_cast<float>(_startDate.daysTo(coords[i].first)),
+            coords[i].second);
     }
```
TomFischer: Instead of lines 312--318 you could write:
```std::transform(coords.begin()++, coords.end(), std::back_inserter(_coords),
                   [this](auto const& coord) {
                       return std::make_pair(_startDate.daysTo(coord.first),
                                             coord.second);
                   });
```

Applications/DataExplorer/DataView/DiagramView/QArrow.h
```diff
@@ -25,26 +25,26 @@ const double PI = 3.14159265;
...
     ~QArrow() override;
 
     double getLength();
     double getAngle();
     void paint(QPainter* painter, const QStyleOptionGraphicsItem* option,
                QWidget* widget) override;
     QRectF boundingRect() const override;
-    void setAngle(double a);
-    void setLength(double l);
+    void setAngle(float a);
```
TomFischer: Why isn't the argument type qreal?

Applications/DataExplorer/DataView/DiagramView/DiagramList.cpp
TomFischer: For the suggested solution one has to check if there are at least 2 elements in the vector.

Applications/DataExplorer/DataView/DiagramView/DiagramList.cpp
```diff
@@ -288,19 +283,38 @@ int DiagramList::readList(const SensorData* data, std::vector<DiagramList*> &lis
...
+        _coords.end());
+    _startDate = start;
+    for (auto& c : _coords)
+        c.first -= start_secs;
+    update();
+}
 
+void DiagramList::setList(std::vector< std::pair<QDateTime, float> > coords)
+{
     this->_startDate = coords[0].first;
```
TomFischer: Check if the `coords` vector contains at least one element:
```
    if (coords.empty())
        return;
```

Applications/DataExplorer/DataView/DiagramView/DiagramPrefsDialog.cpp
```diff
@@ -67,47 +67,52 @@ DiagramPrefsDialog::~DiagramPrefsDialog()
...
-        if (_list[0]->size() > 0)
-        {
-            bool window_is_empty(false);
-            if (_window == nullptr)
-            {
-                _window = new DetailWindow();
-                _window->setAttribute(Qt::WA_DeleteOnClose);
-                window_is_empty = true;
-            }
+        OGSError::box("No data found...");
```
TomFischer: It could be helpful for the user if the error message would be more specific. Error reasons could be
 - an invalid start or end date, 
 - an illegal date range,
 - an empty data set.
So I would split the if condition. In the error message I would also mention in which part of the  program the error occurred.

Applications/DataExplorer/DataView/DiagramView/DiagramPrefsDialog.cpp
```diff
@@ -67,47 +67,52 @@ DiagramPrefsDialog::~DiagramPrefsDialog()
...
+        }
 
-            if (!window_is_empty)
-            {
-                _window->show();
-                this->done(QDialog::Accepted);
-            }
-            else
+        for (std::size_t i = 0; i < _list.size(); i++)
+            if (_visability[i]->isChecked())
```
TomFischer: In line 81 it is checked the size for the first item of the data container. Why isn't it checked for the other items in the `_list`.

Applications/DataExplorer/DataView/DiagramView/QArrow.h
rinkk: fixed (not sure why I didn't get a warning here)

Applications/DataExplorer/DataView/DiagramView/DiagramList.cpp
rinkk: added check :heavy_check_mark:

Applications/DataExplorer/DataView/DiagramView/DiagramList.cpp
rinkk: done :heavy_check_mark:

Applications/DataExplorer/DataView/DiagramView/DiagramList.cpp
rinkk: Is this significantly faster? Otherwise I'd rather keep my code because it's easier to read.

Applications/DataExplorer/DataView/DiagramView/DiagramPrefsDialog.cpp
rinkk: As discussed, there is a lot that can go wrong if the format of the input file doesn't fit what I expect. I'd have to add lots of error handling and analysis and would rather not do it at this time because it'll take forever and this part of the Data Explorer is fairly unimportant anyway.

Applications/DataExplorer/DataView/DiagramView/DiagramPrefsDialog.cpp
rinkk: I've added a comment above to make clear what the intent is here...

Applications/DataExplorer/DataView/DiagramView/DiagramScene.cpp
```diff
@@ -142,17 +142,17 @@ void DiagramScene::adjustAxis(float &min, float &max, int &numberOfTicks)
     numberOfTicks = int(ceil(max / step) - std::floor(min / step));
     if (numberOfTicks < MinTicks)
         numberOfTicks = MinTicks;
-    min = std::floor(min / step) * step;
-    max = ceil(max / step) * step;
+    min = (std::floor(min / step) * step);
+    max = (ceil(max / step) * step);
```
endJunction: `std::ceil`
outer parentheses not needed.

Applications/DataExplorer/DataView/DiagramView/DiagramList.h
```diff
@@ -142,6 +146,9 @@ class DiagramList
     /// Returns the width of the bounding box of all data points within the list.
     double width() const { return _maxX - _minX; }
 
+    /// Converts string into QDateTime-format
+    static QDateTime getDateTime(QString s);
```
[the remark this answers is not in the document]
rinkk: What do you mean by "free function"? Put it in BaseLib? Leave it there but take it of DiagramList?  
(In a way this is fairly specific for this class because I'm just checking a few date prepresentations. This is by no means a comprehensive, super-safe, multi-purpose function. Just saying.)

Applications/DataExplorer/DataView/DiagramView/DiagramPrefsDialog.cpp
```diff
@@ -67,47 +67,53 @@ DiagramPrefsDialog::~DiagramPrefsDialog()
...
 
-            for (std::size_t i = 0; i < _list.size(); i++)
-                if (this->_visability[i]->isChecked())
-                {
-                    _window->addList(_list[i]);
-                    window_is_empty = false;
-                }
+    // Data has been loaded.
+    // If loading the other lists fails at least nothing terrible will happen.
+    if (_list[0]->size() > 0)
```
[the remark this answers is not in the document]
rinkk: DiagramList does not have an empty()-function.  
(I mean, I could add one, but it'd just be used here and basically just look like the condition used now.)

Applications/DataExplorer/DataView/DiagramView/DiagramScene.cpp
rinkk: I'm not sure how these got there. Is it possible that clang-format does something like that?

Anyway, I've removed them again... :heavy_check_mark:

Applications/DataExplorer/DataView/DiagramView/DiagramList.h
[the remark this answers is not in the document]
endJunction: A free function is not a member function. If the function is specific to the translation unit (DiagramList.cpp in this case) add the function to the cpp file only, there no need to add it to the class interface.

Applications/DataExplorer/DataView/DiagramView/DiagramPrefsDialog.cpp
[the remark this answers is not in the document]
endJunction: OK. I thought it would be an container like in STL or QT.  I'm editing the first comment.
Early return is still a suggestion.

Applications/DataExplorer/DataView/DiagramView/DiagramScene.cpp
endJunction: no, clang-format does not add parentheses in this case... It would not change the formatting at all, because it is already correct... ;)

Applications/DataExplorer/DataView/DiagramView/DiagramList.h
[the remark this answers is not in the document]
rinkk: I need the method in both DiagramList and DiagramPrefsDialog. What's your preferred way to implement this then? As a free function in both files? A seperate file?

Applications/DataExplorer/DataView/DiagramView/DiagramPrefsDialog.cpp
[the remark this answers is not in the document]
rinkk: done :heavy_check_mark:

Applications/DataExplorer/DataView/DiagramView/CMakeLists.txt
```diff
@@ -10,6 +10,7 @@ set(SOURCES
 
 set(HEADERS
     DiagramList.h
+    getDateTime.h
```
endJunction: I'd use capital letter in the beginning for consistency: `GetDateTime.h`
And you forgot to upload the actual file, I guess...

PR #2178, opened 2018-08-09
Partition boundary meshes

endJunction:
Follow up of #2177 (MeshLib vertex point vtk conversion).

Further refactorings (mainly free-functions) to the NodeWiseMeshPartitioner.
Extension of the partmesh tool to partition boundary meshes (with ctests).

### DIFF
--- a/Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
+++ b/Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
@@ -125,6 +125,25 @@ void splitOffHigherOrderNode(std::vector<MeshLib::Node*> const& nodes,
     }
 }
 
+std::size_t nodeIdBulkMesh(
+    MeshLib::Node const& node,
+    std::vector<std::size_t> const* node_id_mapping = nullptr)
+{
+    return node_id_mapping ? (*node_id_mapping)[node.getID()] : node.getID();
+}
+
+std::size_t partitionLookup(
+    MeshLib::Node const& node,
+    std::vector<std::size_t> const& partition_ids,
+    std::vector<std::size_t> const* node_id_mapping = nullptr)
+{
+    auto node_id = [&node_id_mapping](MeshLib::Node const& n) {
+        return nodeIdBulkMesh(n, node_id_mapping);
+    };
+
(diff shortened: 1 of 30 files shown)


Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
```diff
@@ -397,6 +486,69 @@ void NodeWiseMeshPartitioner::partitionByMETIS(
...
+{
+    auto const& bulk_node_ids =
+        mesh.getProperties().getPropertyVector<std::size_t>("bulk_node_ids");
+
+    std::vector<Partition> partitions(_partitions.size());
+    for (std::size_t part_id = 0; part_id < _partitions.size(); part_id++)
+    {
+        auto& partition = partitions[part_id];
+        INFO("Processing partition: %d", part_id);
+        {  // processPartition(part_id, is_mixed_high_order_linear_elems)
```
TomFischer: Is this comment necessary?

Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
```diff
@@ -633,7 +782,8 @@ void getElementIntegerVariables(
 
     for (long i = 0; i < nn; i++)
     {
-        elem_info[counter++] = local_node_ids.at(elem.getNodeIndex(i));
+        auto const& n = *elem.getNode(i);
+        elem_info[counter++] = local_node_ids.at(n.getID());
```
TomFischer: I wonder why is the access into `local_node_ids` via the `at()` method?

endJunction: For safety. In a map, accessed via `op[]`, a default value would be created.

endJunction: Note to myself: local_node_ids could be passed as `const&` maybe...

TomFischer: Thanks for the explanation. I overlook that `local_node_ids` is a `map`.

PR #2181, opened 2018-08-13
Changed the type of paramter members of  CreepBGRa

wenqing:
The member type is changed from double to Parameter. One test of BGRa is changed to use different parameters of BGRa creep model with the new source code change.

### DIFF
--- a/MaterialLib/SolidModels/CreepBGRa.cpp
+++ b/MaterialLib/SolidModels/CreepBGRa.cpp
@@ -43,9 +43,16 @@ CreepBGRa<DisplacementDim>::integrateStress(
     KelvinVector sigma_try = sigma_prev + C * (eps - eps_prev);
     ResidualVectorType solution = sigma_try;
 
+    const double A = _a(t, x)[0];
+    const double n = _n(t, x)[0];
+    const double sigma0 = _sigma_f(t, x)[0];
+    const double Q = _q(t, x)[0];
+
+    const double coef = A * std::pow(1.5, 0.5 * (1 + n)) / std::pow(sigma0, n);
+
     const double b =
-        dt * _coef *
-        std::exp(-_q / (MaterialLib::PhysicalConstant::IdealGasConstant * T));
+        dt * coef *
+        std::exp(-Q / (MaterialLib::PhysicalConstant::IdealGasConstant * T));
     auto const& deviatoric_matrix = Invariants::deviatoric_projection;
 
     // In newton_solver.solve(), the Jacobian is calculated first, and then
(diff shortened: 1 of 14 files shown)


MaterialLib/SolidModels/CreepBGRa.cpp
```diff
@@ -43,9 +43,16 @@ CreepBGRa<DisplacementDim>::integrateStress(
     KelvinVector sigma_try = sigma_prev + C * (eps - eps_prev);
     ResidualVectorType solution = sigma_try;
 
+    const double A = _a(t, x)[0];
+    const double n = _n(t, x)[0];
+    const double sigma0 = _sigma_f(t, x)[0];
+    const double Q = _q(t, x)[0];
+
+    const double coef = A * std::pow(1.5, 0.5 * (1 + n)) / std::pow(sigma0, n);
```
endJunction: style: to name the variable after its meaning would be better.

The same function is implemented below; a single function for this computation might be better.

web/content/docs/benchmarks/creep-after-excavation-bgra/CreepAfterExcavation.md
```diff
@@ -53,7 +53,7 @@ assume a steady state of stress and temperature of excavation at the
 beginning of the current simulation. For this assumption, the boundary
 conditions are given as:
 
--   top boundary: $\tau_x = \sigma_x=0$, $\tau_y=\sigma_y=70$ MPa,
+-   top boundary: $\tau_x =$$\sigma_x$$_y=0$, $\tau_y=\sigma_y=-10$ MPa,
```
endJunction: Not sure about double $$ here, though the rendering looks ok. Maybe `$sigma_{xy} = 0$`....

MaterialLib/SolidModels/CreepBGRa.cpp
wenqing: Introduced a local function for this coefficient computation, and changed the variable name with meaning as well.

web/content/docs/benchmarks/creep-after-excavation-bgra/CreepAfterExcavation.md
wenqing: $sigma_{xy} = 0$ did not work (actually, {} in subscript does not work  occasionally). That's why I split it into three formulas as  `$\tau_x =$$\sigma_x$$_y=0$`. With the splitting, you can see that $$ are not doubled $ for one formulas.

By the way, the formulas in the benchmark documentation have already been corrected according to the comments on the report by @nagelt.

PR #2170, opened 2018-07-31
Python BCs

chleh:
Python code for BCs
================
* [Hertz contact](https://github.com/chleh/ogs6/blob/python-bcs-for-pr/Tests/Data/Mechanics/Linear/Python/hertz-contact-bc.py) // [docu](https://deploy-preview-2170--ogs.netlify.com/docs/benchmarks/python-bc/hertz-contact/hertz-contact/)
* [Laplace equation with BCs by given functions](https://github.com/chleh/ogs6/blob/python-bcs-for-pr/Tests/Data/Elliptic/square_1x1_GroundWaterFlow_Python/bcs_laplace_eq.py) // [docu](https://deploy-preview-2170--ogs.netlify.com/docs/benchmarks/python-bc/python-laplace-eq/)

Limitations (not solved in this PR)
===========================
* Only works if all physical fields have the same ansatz functions, i.e., not for HM.
* The normal traction BC is special and not covered by this PR.
* Constraint boundary conditions (#2145) cope with secondary variables (fluxes). Fluxes cannot be *passed to* the Python scripts by the code of this PR.

TODO
=====
* [x] finish Hertz contact test
* [x] test with nonlinear BC
* [x] test with multi-variable process
* [x] input file documentation
* [x] maybe fix possible Python version mismatch (presumably between VTK and OGS)
* [x] fix git lfs tricks (*.pnK)
* [x] squash

Review
======
* [ ] attribute used
* [x] VTK Python version mismatch
* [x] discuss NLSolver changes.
* [ ] Large tests & Python?

Issues to open
============
* [x] Maybe the two versions of deriveBCMap could be merged into one?
* [x] Python BC for normal traction
* [x] Python BC for mixed FEM ansatz functions
* [x] passing secondary variables to Python BC

~~***Don't review yet! Soon: review commit-wise.***~~

PR #2184, opened 2018-08-17
Python BCs: Hertz Contact test

chleh:
Extracted from #2170, follow-up of #2170.

In the Hertz contact test the BC decides to prescribe/release some Dirichlet BC d.o.f.s differently
from one iteration to another.
Because manipulating the solution vector is done after the assembly (except for the first iteration in each timestep) in the non-linear solver, the assembler does not "see" the right values at the Dirichlet boundary if the Dirichlet boundary changes between iterations.

Currently in this PR the issue is tried to be solved in a hackish way, namely the computation of the difference `delta_x` in the Newton solver is modified s.t. changed Dirichlet d.o.f.s will likely lead to a failing tolerance check and to additional non-linear solver iterations.

However, the full solution to the problem probably is:

```
for each iteration:
  set Dirichlet d.o.f.s to x     <-- not done currently      (*)
  assemble rhs, Jac using x
  set Dirichlet d.o.f.s to Jac, res, -delta_x
  solve Jac * (-delta_x) = res
```

(*), however, might require many changes to the whole ODE solver code, because the non-linear solver is time-agnostic, whereas the Dirichlet BC code requires time as an input.
(Currently the time is stored within the `TimeDiscretizedODESystem` and BCs are set with `applyKnownSolutions` and never returned to the non-linear solver.)

**Update**: It turned out that implementing the required changes was pretty straightforward.

~~Status~~
=====

~~The tolerance increases so far in this PR should be sufficient that all non-LARGE tests don't fail.~~

TODO
=====

* [x] also add forgotten Tests.cmake file.

Note
====

I flagged the added Hertz contact test as large, because it takes between one and two minutes on my laptop.


chleh:
@wenqing, you requested to change one sketch in the web docs. After @endJunction on our meeting this Monday, I think we agreed that the present figure is correct.

web/content/docs/benchmarks/python-bc/hertz-contact/hertz-contact.md
```diff
@@ -0,0 +1,100 @@
++++
```
endJunction: Could you rename the `.md` to a `hertz-contact.pandoc`, please. See https://github.com/ufz/ogs/pull/2205 for reason...

Tests/Data/Mechanics/Linear/PythonHertzContact/gen-unit-circle.py
```diff
@@ -0,0 +1,126 @@
...
+
+idcs = np.where(abs(R_squared - 1) < 1e-8)[0]
+idcs = idcs[np.argsort(phi[idcs])]  # sorted with ascending polar angle
+
+
+with open(out_geom, "w") as fh:
+    fh.write("""<?xml version="1.0" encoding="ISO-8859-1"?>
+<?xml-stylesheet type="text/xsl" href="OpenGeoSysGLI.xsl"?>
+
+<OpenGeoSysGLI xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://www.opengeosys.org/images/xsd/OpenGeoSysGLI.xsd" xmlns:ogs="http://www.opengeosys.org">
```
endJunction: The schema can be dropped, as in https://github.com/ufz/ogs/pull/2198

endJunction:
I tried to build OGS with python, w/o success. Shared libs seems to cause a problem.
In new directory:
```
> rm -rf *; cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON -DOGS_USE_PYTHON=ON ../s
...
-- Found PythonInterp: /usr/bin/python (found version "3.7") 
-- Found PythonLibs: /usr/lib/libpython3.7m.so
-- pybind11 v2.2.3
-- Successfully loaded VTK-m component Base
-- CXX target ApplicationsLib cotired without precompiled header. Too few applicable sources.
-- CXX target DataHolderLib cotired.
-- CXX target ApplicationsFileIO cotired.
CMake Error at Applications/CLI/CMakeLists.txt:49 (target_compile_definitions):
  target_compile_definitions called with invalid arguments
```

Building w/o shared libraries work though....

chleh:
@endJunction, at first look the problem should be easy to fix...

endJunction:
CMake works. Linker fails with:
```
/usr/bin/ld: lib/libogs_embedded_python.a(ogs_embedded_python.cpp.o): in function `pybind11_init_impl_OpenGeoSys':
ogs_embedded_python.cpp:(.text+0x3ca): undefined reference to `ProcessLib::pythonBindBoundaryCondition(pybind11::module&)'
```

chleh:
> 2 successful checks
Where is Jenkins?

chleh:
@endJunction, no, the commit you mentioned is there on purpose. Otherwise PETSc+Python compilation fails. We currently don't test this setting, but maybe in the future...

endJunction:
Never mind. There is a rename "bc_mesh" to "boundary_mesh"---I didn't realize... ;)

PR #2188, opened 2018-08-19
Richards and material models

endJunction:
A set of smaller fixes and extension of the unsaturated material model by permeability derivative and second derivative of capillary pressure wrt saturation. The derivatives are used in the RichardsMechanics process implementation in the following PR.

### DIFF
--- a/MaterialLib/PorousMedium/UnsaturatedProperty/CapillaryPressure/VanGenuchtenCapillaryPressureSaturation.cpp
+++ b/MaterialLib/PorousMedium/UnsaturatedProperty/CapillaryPressure/VanGenuchtenCapillaryPressureSaturation.cpp
@@ -15,6 +15,7 @@
 #include <cmath>
 
 #include "MathLib/MathTools.h"
+#include "BaseLib/Error.h"
 
 namespace MaterialLib
 {
@@ -84,6 +85,30 @@ double VanGenuchtenCapillaryPressureSaturation::getdPcdS(
         ((S - _saturation_r) / (_saturation_max - _saturation_r)), -1.0 / _m);
     const double val2 = std::pow(val1 - 1.0, -_m);
     return _pb * (_m - 1.0) * val1 * val2 / (_m * (S - _saturation_r));
+}
+
+double VanGenuchtenCapillaryPressureSaturation::getd2PcdS2(
+    const double saturation) const
+{
+    if (_has_regularized)
+    {
+        OGS_FATAL(
+            "Second derivative of regularized van-Genuchten saturation "
+            "pressure relation is not implemented.");
+    }
+    if (saturation < _saturation_r)
+        return 0;
+    if (saturation > _saturation_max)
+        return 0;
+
+    const double S =
+        MathLib::limitValueInInterval(saturation, _saturation_r + _minor_offset,
+                                      _saturation_max - _minor_offset);
+    const double val1 = std::pow(
+        ((S - _saturation_r) / (_saturation_max - _saturation_r)), 1.0 / _m);
+    return -_pb / (_m * _m * (S - _saturation_r) * (S - _saturation_r)) *
+           std::pow(1 - val1, -_m - 1) * std::pow(val1, _m - 1) *
+           ((1 - _m * _m) * val1 + _m - 1);
 }
 /// Regularized van Genuchten capillary pressure-saturation Model
 double VanGenuchtenCapillaryPressureSaturation::getPcBarvGSg(double Sg) const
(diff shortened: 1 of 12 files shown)


MaterialLib/PorousMedium/UnsaturatedProperty/CapillaryPressure/VanGenuchtenCapillaryPressureSaturation.cpp
```diff
@@ -51,10 +52,8 @@ double VanGenuchtenCapillaryPressureSaturation::getCapillaryPressure(
 double VanGenuchtenCapillaryPressureSaturation::getSaturation(
     const double capillary_pressure) const
 {
-    const double pc =
-        (capillary_pressure < 0.0) ? _minor_offset : capillary_pressure;
-    double Se = std::pow(pc / _pb, 1.0 / (1.0 - _m)) + 1.0;
-    Se = std::pow(Se, -_m);
+    const double pc = std::max(_minor_offset, capillary_pressure);
```
TomFischer: The previous expression and the new one are not equal. For instance if `capillary_pressure = 0.1` and `_minor_offset = 0.5` the previous expression set `pc` to `0.1` while the new expression results in `pc = 0.5`. Is this intended?

MaterialLib/PorousMedium/UnsaturatedProperty/CapillaryPressure/VanGenuchtenCapillaryPressureSaturation.cpp
```diff
@@ -77,6 +76,11 @@ double VanGenuchtenCapillaryPressureSaturation::getdPcdS(
 
         return -getdPcdSvGBar(1 - _saturation_r);
     }
+    if (saturation < _saturation_r)
+        return 0;
```
TomFischer: Is `_saturation_r` a lower bound? Then I would expect to set the saturation to `_saturation_r`.

MaterialLib/PorousMedium/UnsaturatedProperty/CapillaryPressure/VanGenuchtenCapillaryPressureSaturation.cpp
```diff
@@ -77,6 +76,11 @@ double VanGenuchtenCapillaryPressureSaturation::getdPcdS(
...
+    if (saturation > _saturation_max)
+        return 0;
```
TomFischer: Why not `_saturation_max`?

MaterialLib/PorousMedium/UnsaturatedProperty/CapillaryPressure/VanGenuchtenCapillaryPressureSaturation.cpp
```diff
@@ -85,6 +89,41 @@ double VanGenuchtenCapillaryPressureSaturation::getdPcdS(
...
     return _pb * (_m - 1.0) * val1 * val2 / (_m * (S - _saturation_r));
 }
+
+double VanGenuchtenCapillaryPressureSaturation::getd2PcdS2(
+    const double saturation) const
+{
+    if (_has_regularized)
+    {
+        OGS_FATAL("Not implemented.");
+        /*
```
TomFischer: What is the plan for the commented code? Is it a suggestion for the implementation?

MaterialLib/PorousMedium/UnsaturatedProperty/CapillaryPressure/VanGenuchtenCapillaryPressureSaturation.cpp
endJunction: This one is indeed correct: there is a cut-off in the function for s < s_r, so the derivative (this function) is 0.

MaterialLib/PorousMedium/UnsaturatedProperty/CapillaryPressure/VanGenuchtenCapillaryPressureSaturation.cpp
endJunction: Intended. This moves the cutoff to correct point; the pc value is set to "minor_offset" not only for non-positive values, but for pc < minor_offset.
With capillary_pressure(x) = x, minor_offset = 0.1, the pc(x = 0.05) = 0.05 in the previous implementation, now it is does not have this "negative tooth" between x = 0 and 0.1.

MaterialLib/PorousMedium/UnsaturatedProperty/CapillaryPressure/VanGenuchtenCapillaryPressureSaturation.cpp
endJunction: Same cut-off in function => derivative = 0

MaterialLib/PorousMedium/UnsaturatedProperty/CapillaryPressure/VanGenuchtenCapillaryPressureSaturation.cpp
endJunction: Fixing...

PR #2189, opened 2018-08-19
RichardsMechanics implementation.

endJunction:
Based on #2188 

Monolithic scheme. (Staggered scheme functions are prepared, but not implemented.)
Current implementation relies on the RichardsFlow material and solid material model implementations. Some of the properties are repeated therefore, and not all properties (of RichardsFlow material) are used, storage term, for example...
Tests for coupled and decoupled saturated flow are provided. Decoupled unsaturated test (with comparison to the RichardsFlow implementation) is given. There is no test for coupled unsaturated flow for now.


ProcessLib/RichardsMechanics/RichardsMechanicsFEM-impl.h
```diff
@@ -0,0 +1,819 @@
...
+                (porosity / K_LR + a0 * 3 * S_L + dS_L_dp_cap * p_cap_ip * a0);
+
+        storage_p.noalias() +=
+            N_p.transpose() * rho_LR * specific_storage * N_p * w;
+
+        local_Jac
+            .template block<pressure_size, pressure_size>(pressure_index,
+                                                          pressure_index)
+            .noalias() -=
+            N_p.transpose() * rho_LR * dspecific_storage_dp_cap * N_p * w;
```
wenqing: as discussed, dot p_c is missing.

ProcessLib/RichardsMechanics/RichardsMechanicsFEM-impl.h
```diff
@@ -0,0 +1,819 @@
...
+        double const specific_storage =
+            dS_L_dp_cap * (p_cap_ip * S_L * a0 - porosity) +
+            S_L * (porosity / K_LR + S_L * a0);
+
+        double const dspecific_storage_dp_cap =
+            d2S_L_dp_cap_2 * (p_cap_ip * S_L * a0 - porosity) +
+            dS_L_dp_cap *
+                (porosity / K_LR + a0 * 3 * S_L + dS_L_dp_cap * p_cap_ip * a0);
+
+        storage_p.noalias() +=
```
wenqing: I see that the pressure part is not scaled with rho_LR. It is OK.

endJunction:
TODO's after merge:
 - Use GPML (need some time with NG)
 - Swelling is not included.
 - Test of coupled unsaturated flow (Vin?)

ProcessLib/RichardsMechanics/RichardsMechanicsFEM-impl.h
```diff
@@ -0,0 +1,832 @@
...
+            N_p.transpose() * rho_LR * specific_storage * N_p * w;
+
+        local_Jac
+            .template block<pressure_size, pressure_size>(pressure_index,
+                                                          pressure_index)
+            .noalias() -= N_p.transpose() * rho_LR * p_cap_dot_ip *
+                          dspecific_storage_dp_cap * N_p * w;
+
```
wenqing: `-=` should be `+=`, because
d (alpha dot p)/dp = d (alpha)/dp dot p + alpha d(dot p)/dp.  In which  
```d (alpha)/dp dot p = d (alpha)/dp_c dp_c/dp dot p =  d (alpha)/dp_c (-1) (-dot p_c) =  d (alpha)/dp_c dot p_c)```

endJunction: What you say, is correct, though the convergence is worse. I'm looking into this, but if there is no improvement, I'll omit the term completely for now...

endJunction:
Jacobian fixed and corresponds to the numerical Jacobian in RM and RF.
RF example with 2nd order integration delivers slightly different results and was changed to use the 3rd order for comparison with RM. Since density is cancelled out in RF, for comparison it is set to 1.

PR #2192, opened 2018-08-21
[MeL] Change checkNonlinearNodeId msg to warning.

endJunction:
Since (most) parts of the Mesh are not dependent on this property, this is merely a warning.

It would be better, though, to run this check only for algorithms relying on this property. Does anybody knows such algorithms? @rinkk , @TomFischer ?
There, the check must be a hard error, not just a warning.

Related to https://github.com/ufz/ogs/pull/1495

### DIFF
--- a/MeshLib/Mesh.cpp
+++ b/MeshLib/Mesh.cpp
@@ -306,9 +306,10 @@ void Mesh::checkNonlinearNodeIDs() const
             if (e->getNodeIndex(i) >= getNumberOfBaseNodes())
                 continue;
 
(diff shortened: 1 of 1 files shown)


rinkk:
I have just checked and this has been added by Nori about two years ago. It is called just once (in the constructor for Mesh) and just gives the warning but doesn't actually do anything. 

Maybe it was planned as a helper function for something that has never been implemented? Based on the current implementation I would argue that we can remove it. (Does Nori still contribute? If so we could ask him. If not, no one is ever going to use this, I guess.

Issue #2169, opened 2018-07-30
All condition files and geometric files  that are saved from the Modelling Data View are invalid


TanLanglang:
I am reading Computational Hydrology: I Groundwater Flow Modeling .I can not get the gml-file and cnd-file that need to convert to gli-file and bc/ic/st-files by OGSfileConverter when I learn to Case Study. 
   XMLInterface:isValid()-XML File is invalid (in reference to schema./OpenGeoSysCND.xsd).

TanLanglang:
Is there anyone help me?

endJunction:
Hi Tan,
sorry for the delay, many people are on vacation now...

We would need some more information on the issue, so we can try to reproduce your problem.
First of all, what is the exact ogs version? You can see it when starting ogs with
```sh
yourBuildDirectory/bin/ogs --version
```
On which OS are you working?

TomFischer:
Which version of the DataExplorer do use?

Are you talking about the chapter 5? If yes, did you download all the data as described in the chapter?

rinkk:
Do you get an error message what exactly the problem is?

We are currently having issues with XML validaten after switching to HTTPS on our file server. Possibly, the file converter (and the Data Explorer) don't find the remote files to validate your GML. You could try opening your GML-file in a text editor and remove ```xsi:noNamespaceSchemaLocation="https://www.opengeosys.org/images/xsd/OpenGeoSysGLI.xsd"```.

That way the programme should use your local files for validaten instead of looking for the remote ones.
If this doesn't fix your issue, please let us know which version of the Data Explorer you use and -- if possible -- what the error message is.

TanLanglang:
Sorry for delay. I'm sure I‘ve downloaded all the data as described in the chapter. I have experienced it by OGS-5 and OGS-6.1.0. 
Thanks to all people
According to rinkk's opinion,The issue has been resolved. @rinkk You did me a great favor!

bilke:
See also: https://www.opengeosys.org/docs/userguide/troubleshooting/#xsderror-loaded-schema-file-is-invalid

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/2169).

Issue #2136, opened 2018-06-02
xsd files missing on the website


endJunction:
Since #2123 the XSD files like http://www.opengeosys.org/images/xsd/OpenGeoSysGLI.xsd are missing.

Currently all PR's will fail (Jenkins/Mac) with an error in the tests target:
```
[ RUN      ] TestGmlInterface.QtXmlGmlWriterReaderTest

Error XSDError in http://www.opengeosys.org/images/xsd/OpenGeoSysGLI.xsd, at line 1, column 66: Document is not a XML schema.

Error XSDError in file:///Volumes/datadrive/jenkins/workspace/5-FHNEQE5DIY/build/bin/testrunner, at line 3, column 194: Loaded schema file is invalid.

../Tests/FileIO_Qt/TestQtGmlInterface.cpp:43: Failure
```

I'd think the following files need to be uploaded:
- ./Applications/FileIO/XmlIO/OpenGeoSysCND.xsd
- ./Applications/FileIO/XmlIO/OpenGeoSysNum.xsd
- ./Applications/FileIO/XmlIO/OpenGeoSysProject.xsd
- ./GeoLib/IO/XmlIO/OpenGeoSysGLI.xsd
- ./GeoLib/IO/XmlIO/OpenGeoSysSTN.xsd

endJunction:
Could it be that the DE app will not run from now on?

endJunction:
Unfortunately the suggested change does not fix the problem. The files are still not on the web page. There must be a different mechanism. If the problem is not solved on Wednesday, I'll push a change disabling the checks in the OGS-6 code/DE because nobody can work with that code when referencing the xsd files.

A quick workaround suggested by @waltherm is to remove the xsd file reference from the GLI/GML/STN/etc. files.

bilke:
@endJunction Move them into `static` instead of `content`:

` web/static/images/...`

See also https://gohugo.io/content-management/static-files/.

Currently `static` is git-ignored, you may have to add them with `-f`.

endJunction:
@bilke Added, but the files are not pushed to the web site.
https://github.com/ufz/ogs/tree/master/web/static/images/xsd
https://www.opengeosys.org/images/xsd/OpenGeoSysCND.xsd

endJunction:
Resolved.

bilke:
See also https://www.opengeosys.org/docs/userguide/troubleshooting/#xsderror-loaded-schema-file-is-invalid

[...]

PR #2198, opened 2018-08-28
Remove writing of xsd header in XML files

bilke:
As the existence of this line:

```
xsi:noNamespaceSchemaLocation="http://www.opengeosys.org/images/xsd/OpenGeoSysXXX.xsd"
```

in OpenGeoSys XML files still (see #2169, #2136 and 57821a958) leads to validation errors it is no longer written to OpenGeoSys generated XML files. Local validation (either in source or also when using Tools from precompiled zip-archives, e.g. from the website) should still work. The only functionality which will be lost is if you open a XML file standalone in an editor (which supports validation) the file is not validated.

For existing XML files it is strongly advised to remove the line manually to prevent validation errors with current OGS tools.

The XSD files are still served from `opengeosys.org/images/xsd/..` for refercence and are now automatically updated (by simply using symlinks).


bilke:
Failed unit test on Windows:

```
[ RUN      ] ShapeFunctionTest/10.PartitionOfUnity

e:\jenkins\workspace\8-b3ql3ysxj6\thirdparty\autocheck\include\autocheck\reporter.hpp(91): error: Value of: AUTOCHECK_SUCCESS

  Actual: false

Expected: true

Falsifiable, after 80 tests:

([-0.36244,-0.103832,0.272988,])



[  FAILED  ] ShapeFunctionTest/10.PartitionOfUnity, where TypeParam = class NumLib::ShapeHex20 (1 ms)
```

GeoLib/IO/XmlIO/Boost/BoostXmlGmlInterface.cpp
```diff
@@ -42,7 +42,6 @@ bool BoostXmlGmlInterface::readFile(const std::string &fname)
 
     // ignore attributes related to XML schema
     doc->ignoreConfigAttribute("xmlns:xsi");
-    doc->ignoreConfigAttribute("xsi:noNamespaceSchemaLocation");
```
chleh: Removing that line will probably lead to errors with existing files that contain that attribute. Maybe better leave in order to not frustrate users?

bilke: Good point, thanks!

PR #2199, opened 2018-08-31
Use multiple mesh infrastructure for balance calculation.

TomFischer:


ProcessLib/CalculateSurfaceFlux/Balance.h
```diff
@@ -0,0 +1,120 @@
...
+
+#pragma once
+
+#include <logog/include/logog.hpp>
+
+#include <memory>
+#include <string>
+
+#include "MeshLib/IO/readMeshFromFile.h"
+// TODO used for output, if output classes are ready this has to be changed
```
endJunction: TODO -> TODO (fischer)

ProcessLib/CalculateSurfaceFlux/Balance.h
```diff
@@ -0,0 +1,120 @@
...
+#include "MeshLib/IO/writeMeshToFile.h"
+
+#include "ProcessLib/CalculateSurfaceFlux/CalculateSurfaceFlux.h"
+
+namespace ProcessLib
+{
+struct Balance
```
endJunction: The previous name was including "flux", which was more descriptive. I'd also think, there is no balancing in here, but only calculation of flux; Needs better name at the end.

ProcessLib/CalculateSurfaceFlux/Balance.h
```diff
@@ -0,0 +1,120 @@
...
+        std::string const fname =
+            BaseLib::dropFileExtension(output_mesh_file_name) +
+            "_t_" + std::to_string(t) + ".vtu";
+        MeshLib::IO::writeMeshToFile(surface_mesh, fname);
+    }
+
+    MeshLib::Mesh & surface_mesh;
+    std::string const mesh_name;
+    std::string const property_vector_name;
+    std::string const output_mesh_file_name;
```
endJunction: If there is no access to the members, they (or some) could be made private...

PR #2205, opened 2018-09-12
[web] Renamed .md to .pandoc.

bilke:
Fixes formula rendering errors. Pandoc is also a superset of Markdown
with lots of useful features. For the future we should use just Pandoc
as format.

__Addendum__: The markdown code like:
```
This is a formula $\alpha_a = \beta_b$.
```
would be incorrectly rendered because the pair of underscores would be replaced with `<em> ... </em>` pair, which hinders proper LaTeX code rendering, which happens to a later point in time.


endJunction:
@bilke Had to rebase after CreepBGRa.md changes. force-pushed.

chleh:
Nice. Unforunately this change destroys the rendering of https://deploy-preview-2205--ogs.netlify.com/docs/benchmarks/python-bc/laplace-equation/python-laplace-eq/.
There I manually escaped all underscores in formulas `_` → `\_` to get proper results with the `*.md` format.

endJunction:
@chleh I'll fix it.

bilke:
@endJunction Thanks!

PR #2206, opened 2018-09-13
[Jenkins] New warnings publisher syntax

bilke:
See https://jenkins.io/blog/2018/09/11/speaker-blog-warnings-plugin/

Provides a better overview over warnings. I suggest to

- [x] Fix warnings
- [x] Exclude remaining warnings
- [x] Set warning thresholds (to 0 or a specified number) to mark build as unstable

### DIFF
--- a/Applications/CLI/ogs.cpp
+++ b/Applications/CLI/ogs.cpp
@@ -173,8 +173,8 @@ int main(int argc, char* argv[])
             //! \ogs_file_param{prj__insitu}
             if (auto t = project_config->getConfigSubtreeOptional("insitu"))
             {
-                //! \ogs_file_param{prj__insitu__scripts}
                 InSituLib::Initialize(
+                    //! \ogs_file_param{prj__insitu__scripts}
                     t->getConfigSubtree("scripts"),
                     BaseLib::extractPath(project_arg.getValue()));
                 isInsituConfigured = true;
--- a/Applications/CLI/ogs_embedded_python.cpp
+++ b/Applications/CLI/ogs_embedded_python.cpp
@@ -28,7 +28,7 @@ PYBIND11_EMBEDDED_MODULE(OpenGeoSys, m)
 template <typename T>
 void mark_used(T p)
 {
-    volatile T vp = (volatile T)p;
+    volatile T vp = p;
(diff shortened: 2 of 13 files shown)


endJunction:
All of the doxygen warnings are ignorable. There two types max_graph_nodes too small, and a warning in [functional.h](https://jenkins.opengeosys.org/job/ufz/job/ogs/job/PR-2206/1/doxygen/fileName.125649096/source.8ef67070-9a7d-4e06-a376-04501992c3e7/#126) about recursion.

bilke:
It seems that filtering by warning message is not possible yet: https://github.com/jenkinsci/warnings-plugin/blob/master/doc/Documentation.md#filtering-issues

Before we filtered the Doxygen warnings. (`.*DOT_GRAPH_MAX_NODES.` and `*potential recursive class relation.*`)

chleh:
> One of the MSVC warnings (['void (__cdecl *volatile )(void)': top-level volatile in cast is ignored](https://msdn.microsoft.com/en-us/library/ydet06y0.aspx)) could be potentially fixed by
> 
> ```diff
> diff --git a/Applications/CLI/ogs_embedded_python.cpp b/Applications/CLI/ogs_embedded_python.cpp
> index 2da7a23d1..39b6e3e28 100644
> --- a/Applications/CLI/ogs_embedded_python.cpp
> +++ b/Applications/CLI/ogs_embedded_python.cpp
> @@ -28,7 +28,7 @@ PYBIND11_EMBEDDED_MODULE(OpenGeoSys, m)
>  template <typename T>
>  void mark_used(T p)
>  {
> -    volatile T vp = (volatile T)p;
> +    volatile T vp = p;
>      vp = vp;
>  }
> ```
> 
> but since I'm not able to use shared libraries with python, I cannot test its correctness.

@endJunction, that code is intended to capture static libs + Python. Is there a problem for you using shared libs and Python?

endJunction:
@chleh Yes. See also https://github.com/ufz/ogs/pull/2184#issuecomment-420928307 . Let's continue the discussion there. I'll be in the office after lunch.

bilke:
@endJunction Thanks for your fixes!

PR #2208, opened 2018-09-14
Fixes for Eve module and shared lib installations

bilke:
- Migrate to Lmod module syntax
- Install shared libraries to `lib`
- Set `RPATH` appropriately


web/content/docs/devguide/getting-started/build-configuration.pandoc
```diff
@@ -37,9 +37,9 @@ With Conan one build directory corresponds to one configuration. If you want to
 $ [assuming you are at the same directory where the source code directory is located]
 $ mkdir ogs-build && cd ogs-build
 $ mkdir debug && cd debug
-$ cmake ../../ogs -G "Visual Studio 14 2015 Win64" -DOGS_USE_CONAN=ON -DCMAKE_BUILD_TYPE=Debug
+$ cmake ../../ogs -G "Visual Studio 14 2015 Win64" -DCMAKE_BUILD_TYPE=Debug
 $ cd .. && mkdir release && cd release
-$ cmake ../../ogs -G "Visual Studio 14 2015 Win64" -DOGS_USE_CONAN=ON -DCMAKE_BUILD_TYPE=Release
+$ cmake ../../ogs -G "Visual Studio 14 2015 Win64" -DCMAKE_BUILD_TYPE=Release
```
endJunction: Another PR: I think it would be better to provide a list of targets for different visual studio's? And use the latest version in the default text?

PR #2210, opened 2018-09-18
CMake args tracking in exe and installations

bilke:
Please review commit-wise.


scripts/cmake/CMakeSetup.cmake
```diff
@@ -1,14 +1,22 @@
+# Disallow in-source builds
```
endJunction: Why?

scripts/cmake/CMakeSetup.cmake
```diff
@@ -1,14 +1,22 @@
...
+if("${PROJECT_SOURCE_DIR}" STREQUAL "${PROJECT_BINARY_DIR}")
```
chleh: Wouldn't a more general check, e.g. `if ("${PROJECT_BINARY_DIR}/" STRING_STARTSWITH "${PROJECT_SOURCE_DIR}")` be even better. (the search operator is probably not defined.)

CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ endif()
 project( OGS-6 )
 
 include(scripts/cmake/CMakeSetup.cmake)
+include(ParseCMakeArgs)
```
chleh: no "scripts/cmake/" needed?

scripts/cmake/ParseCMakeArgs.cmake
```diff
@@ -0,0 +1,19 @@
+# Captures not-yet cached CMake variables.
+# On first CMake run via cmake-cli this works as expected.
+# Once the variables are cached this will not work anymore.
```
chleh: ... and the use-case for this is overriding certain settings...?

scripts/cmake/packaging/Pack.cmake
```diff
@@ -126,3 +126,6 @@ endif()
 
 configure_file(Documentation/README.txt.in ${PROJECT_BINARY_DIR}/README.txt)
 install(FILES ${PROJECT_BINARY_DIR}/README.txt DESTINATION .)
+
+install(FILES ${PROJECT_BINARY_DIR}/CMakeCache.txt DESTINATION ${CMAKE_INSTALL_INFODIR})
```
chleh: For reproducible builds? Nice.

Applications/CLI/ogs.cpp
```diff
@@ -57,7 +57,9 @@ int main(int argc, char* argv[])
         "See accompanying file LICENSE.txt or "
         "http://www.opengeosys.org/project/license\n"
         "version: " +
-            BaseLib::BuildInfo::git_describe,
+            BaseLib::BuildInfo::git_describe + "\n" +
+        "CMake arguments: " +
+            BaseLib::BuildInfo::cmake_args,
```
chleh: As far as I understand, in the past some user/developer reported/experienced some problems caused by manually set cmake variables, which now are reported by `ogs --help`. That looks like nice feature.

Maybe also some people have problems caused by "dirty" source directories. Therefore (different PR) it might also be helpful to write a status message like "Warning: Source dir dirty, there were changes in the following files: ..." to the screen (and maybe also to each vtu file that we write).

BaseLib/BuildInfo.h
```diff
@@ -32,6 +32,7 @@ namespace BuildInfo
 
     extern BASELIB_EXPORT const std::string git_describe;
     extern BASELIB_EXPORT const std::string ogs_version;
+    extern BASELIB_EXPORT const std::string cmake_args;
```
chleh: Btw. BASELIB_EXPORT seems to be an empty macro. Can't we just remove it?

Applications/CLI/ogs.cpp
endJunction: the '-dirty' postfix is attached by [git describe](https://git-scm.com/docs/git-describe#git-describe---dirtyltmarkgt) in that case...

scripts/cmake/CMakeSetup.cmake
bilke: It disallows running cmake in the source root directory only as this clutters your git project. You can of course still have `source/build` or similar. `build*/` is even git-ignored.

scripts/cmake/packaging/Pack.cmake
bilke: Not for reproducing but for checking which config one actually has..

scripts/cmake/ParseCMakeArgs.cmake
bilke: No I just want to have the info on the CMake configuration inside the OGS exe. As I said this only works of you specify everything on the first CMake run (as we in Jenkins, so the Jenkins provided binaries contain the correct information).

BaseLib/BuildInfo.h
bilke: I thought this was required for shared libs on Windows (.dlls)?

scripts/cmake/CMakeSetup.cmake
bilke: But this would also exclude `source/build/..` which I find handy. See also comment above.

Applications/CLI/ogs.cpp
chleh: Ah, OK.

BaseLib/BuildInfo.h
chleh: Yes, probably. I got confused. After some googling I now roughly understand how this works.
https://cmake.org/cmake/help/v3.0/module/GenerateExportHeader.html

bilke:
@endJunction Don't merge I will explain a bit more..

PR #2211, opened 2018-09-18
[MGTL] Transform OGS_FATAL to DBUG in parallel version.

bilke:
In parallel execution it must not to be necessarily a fatal error if the BoundaryElementsAtPoint searcher doesn't find a mesh node.

Allows for running the cube_ht benchmark on eve.


endJunction:
The node searcher was not intended to work for partitioned meshes. Correct solution would include an allreduce check on nodes.empty(), s.t. at least one rank finds something, and an extension to polyline and surface searchers.
The actual problem (a certain benchmark not running) is solvable when using multiple meshes in the input, like it is done in Tests/Data/EllipticPETSc/*neumann.prj.

As discussed, the second solution is preferred. I'll update some of the documentation today.

PR #2215, opened 2018-09-20
[web/BGRa] Recovered the display of the formula of the elastic tensor and a table

wenqing:
1. CreepBRGa.pandoc:  $\mathbf{C}: = \lambda  \mathcal{J} + 2G \mathbf I \otimes \mathbf I$ cannot be displayed after recent commits. Change it to $\mathbf{C} = \lambda  \mathcal{J} + 2G \mathbf I \otimes \mathbf I$
2. CreepAfterExcavation.pandoc: the table of material properties.


endJunction:
Thanks, Wenqing!

PR #2216, opened 2018-09-20
MaterialIDs for solid models

endJunction:
Does not change the current behaviour: A single constitutive relation is used independent of MaterialIDs being present or not, if there is only one `<constitutive_relation>` section.
Adding an "id" attribute to it `<constitutive_relation id="42">` __and__ having multiple constitutive relations requires the MaterialIDs to be available and there must be a one-to-one correspondence.

Added a simple test gravity with two different materials having different E moduli.
Modified a CreepBGRa test, which was using a special case of the BGRa model (A = 0) to simulate the linear elastic case. Now there are two different materials.


MaterialLib/SolidModels/CreateConstitutiveRelation.cpp
```diff
@@ -33,47 +33,83 @@ createConstitutiveRelation(
...
+        //! \ogs_file_param{material__solid__constitutive_relation}
+        config.getConfigSubtreeList("constitutive_relation");
+
+    std::map<int, std::unique_ptr<MechanicsBase<DisplacementDim>>>
+        constitutive_relations;
+
+    for (auto const& constitutive_relation_config :
+         constitutive_relation_configs)
+    {
+        int material_id =
```
endJunction: const

ProcessLib/HydroMechanics/CreateHydroMechanicsProcess.cpp
```diff
@@ -101,9 +101,14 @@ std::unique_ptr<Process> createHydroMechanicsProcess(
...
-    auto material =
-        MaterialLib::Solids::createConstitutiveRelation<DisplacementDim>(
+    auto const material_ids =
+        mesh.getProperties().existsPropertyVector<int>("MaterialIDs")
+            ? mesh.getProperties().getPropertyVector<int>("MaterialIDs")
+            : nullptr;
+
+    // Constitutive relations.
+    auto solid_materials =
+        MaterialLib::Solids::createConstitutiveRelations<DisplacementDim>(
```
endJunction: One could collect `material_ids` and `solid_materials` in single class and add the above `select...` function to it...

ProcessLib/LIE/SmallDeformation/LocalAssembler/SmallDeformationLocalAssemblerMatrixNearFracture-impl.h
```diff
@@ -81,9 +81,12 @@ SmallDeformationLocalAssemblerMatrixNearFracture<ShapeFunction,
     _ip_data.reserve(n_integration_points);
     _secondary_data.N.resize(n_integration_points);
 
+    auto const material_id = _process_data.material_ids == nullptr
```
endJunction: Use select function here.

ProcessLib/RichardsMechanics/CreateRichardsMechanicsProcess.cpp
```diff
@@ -174,14 +174,15 @@ std::unique_ptr<Process> createRichardsMechanicsProcess(
         //! \ogs_file_param{prj__processes__process__RICHARDS_MECHANICS__material_property}
         config.getConfigSubtree("material_property");
 
-    boost::optional<MeshLib::PropertyVector<int> const&> material_ids;
-    if (mesh.getProperties().existsPropertyVector<int>("MaterialIDs"))
+    auto const material_ids =
+        mesh.getProperties().existsPropertyVector<int>("MaterialIDs")
+            ? mesh.getProperties().getPropertyVector<int>("MaterialIDs")
+            : nullptr;
```
endJunction: use select here too.

ProcessLib/RichardsMechanics/RichardsMechanicsFEM-impl.h
```diff
@@ -54,10 +54,12 @@ RichardsMechanicsLocalAssembler<ShapeFunctionDisplacement,
                           IntegrationMethod, DisplacementDim>(
             e, is_axially_symmetric, _integration_method);
 
+    auto const material_id = _process_data.material_ids == nullptr
```
endJunction: use select here too.

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
```diff
@@ -53,10 +54,12 @@ HydroMechanicsLocalAssembler<ShapeFunctionDisplacement, ShapeFunctionPressure,
                           IntegrationMethod, DisplacementDim>(
             e, is_axially_symmetric, _integration_method);
 
+    auto& solid_material = MaterialLib::Solids::selectSolidConstitutiveRelation(
```
wenqing: const?

ProcessLib/HydroMechanics/HydroMechanicsProcessData.h
```diff
@@ -82,10 +72,14 @@ struct HydroMechanicsProcessData
     //! Assignments are not needed.
     void operator=(HydroMechanicsProcessData&&) = delete;
 
+    MeshLib::PropertyVector<int> const* material_ids;
```
wenqing: Can the type be reference?

ProcessLib/LIE/SmallDeformation/CreateSmallDeformationProcess.cpp
```diff
@@ -98,9 +98,14 @@ std::unique_ptr<Process> createSmallDeformationProcess(
         process_variables;
     process_variables.push_back(std::move(per_process_variables));
 
+    auto const material_ids =
```
wenqing: Can it be a member of mesh class? It appears in several functions.

wenqing: I mean to add a member to Mesh as
```
PropertyVector<int> Mesh::getMaterialIDs() const
{
return _properties.existsPropertyVector<int>("MaterialIDs")
            ? _properties.getPropertyVector<int>("MaterialIDs")
            : nullptr;
}
```

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
endJunction: This is actually possible. Did mark the integrateStress and create*State functions as const, then this one and in the IPData too...

ProcessLib/HydroMechanics/HydroMechanicsProcessData.h
endJunction: No, not with current implementation. It is allowed not to have the 'material_ids'.

ProcessLib/LIE/SmallDeformation/CreateSmallDeformationProcess.cpp
endJunction: What do other think if we add a free function in MeshLib namespace?
```c++
PropertyVector<int>* materialIDs(Mesh const&);
```

wenqing: Sounds good.

ProcessLib/HydroMechanics/CreateHydroMechanicsProcess.cpp
```diff
@@ -187,15 +192,11 @@ std::unique_ptr<Process> createHydroMechanicsProcess(
...
-        fluid_density,
-        biot_coefficient,
-        porosity,
-        solid_density,
-        specific_body_force,
+        material_ids,           std::move(solid_materials),
+        intrinsic_permeability, specific_storage,
+        fluid_viscosity,        fluid_density,
+        biot_coefficient,       porosity,
+        solid_density,          specific_body_force,
```
TomFischer: Is the formatting okay?

endJunction: Looks good to clang-format...

PR #2219, opened 2018-09-23
Warning and error fixes

endJunction:
Fixes for some of the warnings from a [static analyzer run](https://jenkins.opengeosys.org/job/ufz/job/ogs/job/master/702/clang/).

See commit messages for some explanations, so maybe review commit-wise is better.

I definitely changed the logic in most places, but not the end results (so far they are tested ;), so please check carefully.

### DIFF
--- a/Applications/FileIO/TetGenInterface.cpp
+++ b/Applications/FileIO/TetGenInterface.cpp
@@ -492,8 +492,9 @@ bool TetGenInterface::parseElements(std::ifstream& ins,
         }
         // insert new element into vector
         auto** tet_nodes = new MeshLib::Node*[4];
-        for (unsigned k(0); k<4; k++) {
-            tet_nodes[k] = nodes[ids[k]];
+        for (int n = 0; n < 4; n++)
+        {
+            tet_nodes[n] = nodes[ids[n]];
         }
         elements.push_back (new MeshLib::Tet(tet_nodes));
         materials.push_back(region);
--- a/Applications/FileIO/TetGenInterface.cpp
+++ b/Applications/FileIO/TetGenInterface.cpp
@@ -430,8 +430,8 @@ bool TetGenInterface::parseElements(std::ifstream& ins,
                                     bool region_attribute) const
 {
     std::string line;
-    auto* ids(static_cast<std::size_t*>(
-        alloca(sizeof(std::size_t) * n_nodes_per_tet)));
+    std::vector<std::size_t> ids(n_nodes_per_tet);
+
     elements.reserve(n_tets);
     materials.reserve(n_tets);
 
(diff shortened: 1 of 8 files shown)


Tests/GeoLib/TestPolygon.cpp
```diff
@@ -85,14 +85,14 @@ TEST_F(PolygonTest, isPntInPolygonCheckCorners)
...
-        for (double t(0.0); t<1.0; t+=0.001) {
-            EXPECT_TRUE(_polygon->isPntInPolygon(
-                GeoLib::Point(
-                    (*_pnts[k])[0] + t*((*_pnts[k+1])[0]-(*_pnts[k])[0]),
-                    (*_pnts[k])[1] + t*((*_pnts[k+1])[1]-(*_pnts[k])[1]),
-                    (*_pnts[k])[2] + t*((*_pnts[k+1])[2]-(*_pnts[k])[2])
-                ))
-            );
+        double t = 0;
+        while (t < 1.0)
```
chleh: This is basically the same as before. It's only called something else than `for`. Since 0.001 is not exactly representable as double
```
>>> print("{:.50f}".format(0.001))
0.00100000000000000002081668171172168513294309377670
```

it's not intuitively clear how many time the loop will iterate (1000 vs. 1001 times). This is not a big deal here,
but maybe using a true integer loop variable would be a better solution.

Tests/MeshLib/TestUniqueMeshId.cpp
```diff
@@ -23,13 +23,14 @@ TEST(MeshLib, UniqueMeshId)
     //
     // Test mesh counter increments.
     //
-    Mesh* m1 = new Mesh("second", std::vector<Node*>(), std::vector<Element*>());
```
chleh: Time to forbid `new`? ^^

Applications/FileIO/TetGenInterface.cpp
```diff
@@ -430,8 +430,8 @@ bool TetGenInterface::parseElements(std::ifstream& ins,
                                     bool region_attribute) const
 {
     std::string line;
-    auto* ids(static_cast<std::size_t*>(
-        alloca(sizeof(std::size_t) * n_nodes_per_tet)));
+    std::vector<std::size_t> ids(n_nodes_per_tet);
```
chleh: Can `n_nodes_per_tet` be anything other than 4?

endJunction: Could be quadratic...

Applications/FileIO/TetGenInterface.cpp
```diff
@@ -492,8 +492,9 @@ bool TetGenInterface::parseElements(std::ifstream& ins,
         }
         // insert new element into vector
         auto** tet_nodes = new MeshLib::Node*[4];
-        for (unsigned k(0); k<4; k++) {
-            tet_nodes[k] = nodes[ids[k]];
+        for (int n = 0; n < 4; n++)
```
TomFischer: If the tet is of quadratic order is looping only over the four nodes still okay?

endJunction: I guess there is no test for this.
Concerning the for loop change, it is consistent though with the tet_nodes size.

PR #2224, opened 2018-09-27
HC-Goswami benchmark update

jbathmann:
Changes:
-new input mesh(es) for the benchmark
-setup slightly changed
-benchmark-doc-pandoc updated


Tests/Data/Parabolic/ComponentTransport/goswami/goswami_input.prj
```diff
@@ -1,284 +1,226 @@
...
-                <max_iteration_step>20000</max_iteration_step>
-                <error_tolerance>1e-11</error_tolerance>
-            </eigen>
-            <petsc>
-                <prefix>hc</prefix>
-                <parameters>-hc_ksp_type bcgs -hc_pc_type bjacobi -hc_ksp_rtol 1e-11 -hc_ksp_max_it 20000</parameters>
-            </petsc>
-        </linear_solver>
-    </linear_solvers>
+	<meshes>
```
endJunction: Would be nice to see the original formatting. Four spaces instead of tabs. Then the diff would make more sense, because now it looks like the whole file was rewritten.

TomFischer: The tool `xmlstarlet` can format xml files.

endJunction:
@waltherm If this is fine with you, the PR can be merged.

waltherm:
Thanks for the heads-up @endJunction. 

Comparing the currently presented solution in the PR with the previous OGS6 (or OGS5) result is rather bad. I plotted this in the attached figure: white lines OGS5, green lines OGS6 (previous solution), red dots lab experiments, color scale red-blue current PR. 
I think that 1) the solution deviates too much from the experiments (and the previous numerical solutions), and 2) the isolines wiggle too much (mesh too coarse?).

compare_ogs6-exp_ogs5

jbathmann:
> Thanks for the heads-up @endJunction.
> 
> Comparing the currently presented solution in the PR with the previous OGS6 (or OGS5) result is rather bad. I plotted this in the attached figure: white lines OGS5, green lines OGS6 (previous solution), red dots lab experiments, color scale red-blue current PR.
> I think that 1) the solution deviates too much from the experiments (and the previous numerical solutions), and 2) the isolines wiggle too much (mesh too coarse?).
> 
> compare_ogs6-exp_ogs5

Thanks for the comments @waltherm! 
The static state case might be a better scenario for the test simulation as changed in the last commit. This was not the case beforehand and hence the comparison between a arbitrary simulation step and the steady state was showing huge differences. Additionally, I added a finer grid. Would you agree with the changed setup?

waltherm:
> Would you agree with the changed setup?

Will take a look tomorrow...

jbathmann:
@waltherm I incorporated your comments except the suggestion of all three transient states run in the long benchmarks for two reasons. 1) the plots from the doc webpage show the ability of the current version to correctly reproduce the temporal development of the salt water front and hence 2) the temporal evolution of the system is already tested by the .vtu files within the current benchmark version. Hence a longer simulation setup would not lead to any additional insights about the codes performance but cost computation time on the tests. 
Do you agree with this?

waltherm:
:+1:

PR #2228, opened 2018-09-30
RM/HM/LIE-HM interpolated pressure

endJunction:
Move LIE/HM interpolate_pressure to `NumLib::interpolateToHigherOrderNodes`.

Use in RichardsMechanics and HydroMechanics processes.


NumLib/Function/Interpolation.h
```diff
@@ -85,4 +88,52 @@ void shapeFunctionInterpolate(
                                         interpolated_values...);
 }
 
+/// Interpolates \c node_values given in lower order element nodes (e.g. the
+/// base nodes) to higher order element's nodes (e.g. quadratic nodes) and
+/// writes the result into the global property vector.
+///
+/// The base nodes' values are copied.  For each higher order node the shape
+/// matrices are evaluated for the lower order element (the base nodes), and
+/// used for the the scalar quantity interpolation.
```
chleh: Maybe it should be mentioned that the node_values belong to a scalar field.

NumLib/Function/Interpolation.h
```diff
@@ -85,4 +88,52 @@ void shapeFunctionInterpolate(
...
+
+    // Shape matrices storage reused in the interpolation loop.
+    ShapeMatrices shape_matrices{SF::DIM, GlobalDim, SF::NPOINTS};
+
+    // Interpolate values for higher order nodes.
+    for (int n = number_base_nodes; n < number_all_nodes; ++n)
+    {
+        // Evaluated at higher order nodes' coordinates.
+        fe.template computeShapeFunctions<ShapeMatrixType::N>(
+            NaturalCoordinates<HigherOrderMeshElementType>::coordinates[n]
```
chleh: The `HigherOrderMeshElementType` is the "true" type of `element`, isn't it? Maybe that could be commented
or asserted: `assert(dynamic_cast<HigherOrderMeshElementType*>(&element));`.

NumLib/Function/Interpolation.h
```diff
@@ -85,4 +88,52 @@ void shapeFunctionInterpolate(
...
+                .data(),
+            shape_matrices, GlobalDim, is_axially_symmetric);
+
+        std::size_t const global_index = element.getNodeIndex(n);
```
chleh: Does that work with domain decomposition?

endJunction: I think so, because the node numbering is local for each partition, and so are the mesh properties...

PR #2234, opened 2018-10-04
Python source terms

TomFischer:
A first version of Python source term implementation.

ToDo:
 - [x] Check constness, final, forward declarations, ...
 - [x] Add a more complicated example than constant source terms.
 - [x] Documentation of project file tags.
 - [x] Website documentation.

Thanks to @chleh for the support!


ProcessLib/Process.cpp
```diff
@@ -189,7 +189,7 @@ void Process::assemble(const double t, GlobalVector const& x, GlobalMatrix& M,
         (_coupled_solutions) != nullptr ? _coupled_solutions->process_id : 0;
     _boundary_conditions[pcs_id].applyNaturalBC(t, x, K, b, nullptr);
 
-    _source_term_collections[pcs_id].integrate(t, b);
+    _source_term_collections[pcs_id].integrate(t, x, b, nullptr);
```
endJunction: Could you add a comment to this and foregoing `nullptr`, please? Something like `/* unused jacobian */`

ProcessLib/SourceTerms/Python/CreatePythonSourceTerm.cpp
```diff
@@ -0,0 +1,88 @@
...
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#include "CreatePythonSourceTerm.h"
+
+#include <pybind11/pybind11.h>
+#include <iostream>
```
endJunction: Logog for cout... Maybe you need it for flushing, though.

ProcessLib/SourceTerms/Python/PythonSourceTerm.cpp
```diff
@@ -0,0 +1,105 @@
...
+class FlushStdoutGuard
+{
+public:
+    //! Optionally flushes C++ stdout before running Python code.
+    explicit FlushStdoutGuard(bool const flush) : _flush(flush)
+    {
+        if (!flush)
+            return;
+
+        LOGOG_COUT << std::flush;
```
endJunction: std::flush is in `<ostream>`...

ProcessLib/SourceTerms/Python/PythonSourceTerm.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include "ProcessLib/Utils/CreateLocalAssemblers.h"
+#include "ProcessLib/Utils/ProcessUtils.h"
+#include "PythonSourceTermLocalAssembler.h"
+
+namespace
+{
+//! Optionally flushes the standard output upon creation and destruction.
+//! Can be used to improve the debug output readability when printing debug
+//! messages both from OGS and from Python.
+class FlushStdoutGuard
```
endJunction: final, just for safety

ProcessLib/SourceTerms/Python/PythonSourceTerm.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+    try
+    {
+        GlobalExecutor::executeMemberOnDereferenced(
+            &PythonSourceTermLocalAssemblerInterface::assemble,
+            _local_assemblers, *_dof_table_source_term, t, x, b, Jac);
+    }
+    catch (MethodNotOverriddenInDerivedClassException const& /*e*/)
+    {
+        DBUG("Method `getFlux' not overridden in Python script.");
```
endJunction: I would either call `OGS_FATAL` or rethrow the exception, since the whole class' purpose is to call that `getFlux`...

ProcessLib/SourceTerms/Python/PythonSourceTerm.h
```diff
@@ -0,0 +1,77 @@
...
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#pragma once
+
+#include "NumLib/DOF/LocalToGlobalIndexMap.h"
+#include "NumLib/IndexValueVector.h"
```
endJunction: Looks like this particular one is not needed. Can you try to put some of the includes into cpp and use fwd decls here?
These should be possible to put as fwd decls:
PythonSourceTermPythonSideInterface
PythonSourceTermLocalAssemblerInterface.h
LocalToGlobalIndexMap

ProcessLib/SourceTerms/Python/PythonSourceTerm.cpp
endJunction: I might be completely wrong here....

ProcessLib/SourceTerms/Python/PythonSourceTermLocalAssembler.h
```diff
@@ -0,0 +1,231 @@
...
+            ShapeMatrixPolicyType<ShapeFunction, GlobalDim>;
+        using FemType =
+            NumLib::TemplateIsoparametric<ShapeFunction, ShapeMatricesType>;
+        FemType fe(*static_cast<const typename ShapeFunction::MeshElement*>(
+            &_element));
+
+        unsigned const num_integration_points =
+            _integration_method.getNumberOfPoints();
+        auto const num_var = _data.dof_table_bulk.getNumberOfVariables();
+        auto const num_nodes = _element.getNumberOfNodes();
```
endJunction: Use ShapeFunction::NPOINTS.

ProcessLib/SourceTerms/Python/PythonSourceTermLocalAssembler.h
```diff
@@ -0,0 +1,231 @@
...
+                            "e.g., the HM process.",
+                            bulk_node_id, var, comp);
+                    }
+                    primary_variables_mat(element_node_id, global_component) =
+                        x[dof_idx];
+                }
+            }
+        }
+
+        Eigen::VectorXd local_rhs = Eigen::VectorXd::Zero(num_nodes);
```
endJunction: NodalVectorType, or does it have to be dynamic for the python bindings?

ProcessLib/SourceTerms/Python/PythonSourceTermLocalAssembler.h
```diff
@@ -0,0 +1,231 @@
...
+        Eigen::MatrixXd local_Jac =
+            Eigen::MatrixXd::Zero(num_nodes, num_nodes * num_comp_total);
+        std::vector<double> prim_vars_data(num_comp_total);
+        auto prim_vars = MathLib::toVector(prim_vars_data);
+
+        for (unsigned ip = 0; ip < num_integration_points; ip++)
+        {
+            auto const ip_data = _ip_data[ip];
```
endJunction: should be a ref, not a copy. Same for the following definitions in the for-loop scope.

web/content/docs/benchmarks/elliptic/poisson_equation.pandoc
```diff
@@ -0,0 +1,164 @@
...
+info: OGS terminated on 2018-10-10 09:22:17+020
+```
+
+## Results and evaluation
+
+### Comparison of the numerical and analytical solution
+
+{{< img src="../square_1e3_poisson_sin_x_sin_y_sourceterm_Pressure_PythonSourceTerm.png" >}}
+The above picture shows the numerical result. The solution conforms in the edges
+to the prescribed boundary conditions.
```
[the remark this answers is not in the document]
endJunction: Nevermind. The solution is 1 in corners. All fine.

ProcessLib/SourceTerms/Python/PythonSourceTerm.cpp
chleh: I think `MethodNotOverriddenInDerivedClassException` won't be thrown anyway, b/c @TomFischer removed that "feature".

ProcessLib/SourceTerms/Python/CreatePythonSourceTerm.cpp
TomFischer: Yes, it is needed for flushing to keep the C++ output and the python output in the correct sequence.

ProcessLib/SourceTerms/Python/PythonSourceTerm.cpp
TomFischer: Will remove the try-catch block.

ProcessLib/SourceTerms/Python/PythonSourceTerm.cpp
TomFischer: Since the macro LOGOG_COUT is defined as `#define LOGOG_COUT  std::cout` in `ThirdParty/logog/include/string.hpp:39` it is necessary to include iostream.

TomFischer:
Thanks @chleh for the support!
Thanks @chleh and @endJunction for the review!

PR #2235, opened 2018-10-08
[L-IE] supporting fracture intersections in mechanics

norihiro-w:
supporting branch/junction intersections of fractures in 2D M problems


norihiro-w:
i can't find out why Jenkins is failing. anyone can help me?

endJunction:
I think it's a temporary internet access error:
`Get https://registry-1.docker.io/v2/library/ubuntu/manifests/16.04: received unexpected HTTP status: 503 Service Unavailable`
Just restarted appveyor and jenkins.

norihiro-w:
@endJunction can you please restart the Jenkins again?

endJunction:
The jenkins run also fine, but has new warnings; Github cannot display the "unstable" build, so it is displaying a failure. The warnings are:
on windows: for postutils.cpp: 59 and 67: `'return': conversion from '__int64' to 'int', possible loss of data` and `'return': conversion from 'const __int64' to 'unsigned int', possible loss of data`
Then there are new warning for doxygen in levelsetfunction.h; there the @param is wrong, should be `@param`.

The builds automatically starts, if there is a change in commit hash, e.g. by changing the last commit's date.

norihiro-w:
i think now this PR is ready for reviewing

ProcessLib/LIE/Common/BranchProperty.h
```diff
@@ -0,0 +1,32 @@
...
+struct BranchProperty
+{
+	int node_id;
+	Eigen::Vector3d coords;
+	int master_fracture_ID;
+	int slave_fracture_ID;
+	// unit vector normal to the master fracture in a direction to the slave
+	Eigen::Vector3d normal_vector_branch;
+
+    virtual ~BranchProperty() = default;
```
endJunction: - Since nothing is derived from this class, I'd drop the dtor and mark the class `final`.
 - Reorder the members as {coords, normal, node_id, further ints} s.t. to reduce the holes in the memory because of alignment.
 - Not sure, but might be necessary to have aligned new operator http://eigen.tuxfamily.org/dox/group__TopicStructHavingEigenMembers.html

ProcessLib/LIE/Common/FractureProperty.h
```diff
@@ -54,11 +59,28 @@ inline void setFractureProperty(unsigned dim, MeshLib::Element const& e,
...
     frac_prop.R.resize(dim, dim);
     computeRotationMatrix(e, frac_prop.normal_vector, dim, frac_prop.R);
 }
 
+
+inline void setBranchProperty(
+	MeshLib::Node const& branchNode,
+	FractureProperty const& master_frac,
+	FractureProperty const& slave_frac,
+	BranchProperty& branch)
```
endJunction: Consideration: I'd rather return a new Branch object, or maybe create a constructor to the Branch class.

ProcessLib/LIE/Common/LevelSetFunction.cpp
```diff
@@ -10,14 +10,22 @@
...
 #include "FractureProperty.h"
+#include "JunctionProperty.h"
 
 namespace
 {
 // Heaviside step function
 inline double Heaviside(double v)
 {
-    return (v < 0.0) ? 0.0 : 1.0;
+	return (v < 0.0) ? 0.0 : 1.0;
```
endJunction: It seems, that the clang-format is not using the `.clang-format` from the source root...

ProcessLib/LIE/Common/LevelSetFunction.cpp
```diff
@@ -10,14 +10,22 @@
...
+}
+
+template <class T_VEC, class T_V>
+inline bool inList(T_VEC const& vec, T_V const& v)
```
endJunction: Would suggest to move it to BaseLib/Algorithm.h, and rename to `contains` or similar, because it is not acting on lists only.

ProcessLib/LIE/Common/LevelSetFunction.cpp
```diff
@@ -26,11 +34,101 @@ namespace ProcessLib
...
+
+
+std::vector<double> du_global_enrichments(
+	std::size_t this_frac_id,
+	std::vector<FractureProperty*> const& frac_props,
+	std::vector<JunctionProperty*> const& junction_props,
+	std::unordered_map<int,int> const& fracID_to_local,
+	Eigen::Vector3d const& x)
+{
+	auto this_frac_local_index = fracID_to_local.at(this_frac_id);
```
endJunction: Not sure if bound checked access is intended; if not use `operator[]`.

ProcessLib/LIE/Common/PostUtils.cpp
```diff
@@ -16,11 +16,67 @@ namespace ProcessLib
...
+bool includesNodeID(std::vector<MeshLib::Node*> const& vec_nodes,
+                       std::size_t node_id)
+{
+    auto itr2 = std::find_if(
+        vec_nodes.begin(), vec_nodes.end(),
+        [&](MeshLib::Node const* node) { return node->getID() == node_id; });
+    return (itr2 != vec_nodes.end());
+}
+
+std::vector<int> const& getmatids(
```
endJunction: longer names are preferred and use camelCase for function names, *e.g.* `getMaterialIdsForNode(...)`

ProcessLib/LIE/SmallDeformation/SmallDeformationProcess.cpp
```diff
@@ -96,6 +100,73 @@ SmallDeformationProcess<DisplacementDim>::SmallDeformationProcess(
...
+            *_process_data._vec_fracture_property
+                 [_process_data._map_materialID_to_fractureID[slave_matId]];
+
+        BranchProperty* branch = new BranchProperty();
+        setBranchProperty(*mesh.getNode(vec_branch_nodeID_matIDs[i].first),
+                          master_frac, slave_frac, *branch);
+
+        master_frac.branches_master.emplace_back(branch);
+
+        BranchProperty* branch2 = new BranchProperty();
```
endJunction: In the new code `new` is not welcome (unless you have to as in copying the MeshLib's nodes). I also didn't see the reason to store pointers in the `*.branches_master/slave` vectors instead of the branch objects.
In any case this is a memory leak.

ProcessLib/LIE/Common/BranchProperty.h
```diff
@@ -0,0 +1,32 @@
...
+
+#include <Eigen/Eigen>
+
+namespace ProcessLib
+{
+namespace LIE
+{
+struct BranchProperty
+{
+	int node_id;
```
wenqing: These members can be set as const by introducing a constructor.

ProcessLib/LIE/Common/FractureProperty.h
```diff
@@ -54,11 +59,28 @@ inline void setFractureProperty(unsigned dim, MeshLib::Element const& e,
...
     for (int j = 0; j < 3; j++)
-        frac_prop.point_on_fracture[j] = e.getNode(0)->getCoords()[j];
+        frac_prop.point_on_fracture[j] = e.getCenterOfGravity().getCoords()[j];
     computeNormalVector(e, dim, frac_prop.normal_vector);
     frac_prop.R.resize(dim, dim);
     computeRotationMatrix(e, frac_prop.normal_vector, dim, frac_prop.R);
 }
 
+
+inline void setBranchProperty(
```
wenqing: can be converted to a constructor of BranchProperty.

ProcessLib/LIE/Common/BranchProperty.h
norihiro-w: corrected

ProcessLib/LIE/Common/PostUtils.cpp
norihiro-w: done

ProcessLib/LIE/Common/FractureProperty.h
norihiro-w: done

ProcessLib/LIE/Common/LevelSetFunction.cpp
norihiro-w: done.

norihiro-w: update: i remember it now. i had to use `at()` because `fracID_to_local` is const. `operator[]` cannot be used because it is non-const.

ProcessLib/LIE/SmallDeformation/SmallDeformationProcess.cpp
norihiro-w: Because the number of branches varies depending on problem,  i store them as a vector of its pointers. i am not  sure if I understand your question.

ProcessLib/LIE/Common/BranchProperty.h
norihiro-w: yes it can be const but i don't see a definitive reason for it.

ProcessLib/LIE/Common/FractureProperty.h
norihiro-w: here i intentionally separate data and functions (to let myself focus on the data structure during coding).

ProcessLib/LIE/Common/FractureProperty.h
```diff
@@ -54,11 +60,43 @@ inline void setFractureProperty(unsigned dim, MeshLib::Element const& e,
...
     // a fracture is not curving
     for (int j = 0; j < 3; j++)
-        frac_prop.point_on_fracture[j] = e.getNode(0)->getCoords()[j];
+        frac_prop.point_on_fracture[j] = e.getCenterOfGravity().getCoords()[j];
     computeNormalVector(e, dim, frac_prop.normal_vector);
     frac_prop.R.resize(dim, dim);
     computeRotationMatrix(e, frac_prop.normal_vector, dim, frac_prop.R);
 }
 
+inline BranchProperty* createBranchProperty(MeshLib::Node const& branchNode,
```
wenqing: I see that this function is called to fill ``_process_data._vec_junction_property`` which is a vector of unique_ptr. I think that you can return `std::unique_ptr<BranchProperty>` in this function.  Then use std::move in line 139 of file
ProcessLib/LIE/SmallDeformation/SmallDeformationProcess.cpp.

ProcessLib/LIE/Common/FractureProperty.h
```diff
@@ -54,11 +60,43 @@ inline void setFractureProperty(unsigned dim, MeshLib::Element const& e,
...
+                                            FractureProperty const& master_frac,
+                                            FractureProperty const& slave_frac)
+{
+    BranchProperty* branch = new BranchProperty();
```
wenqing: new can be replace with make_unique if the return type is change to unique_ptr.

ProcessLib/LIE/Common/FractureProperty.h
```diff
@@ -54,11 +60,43 @@ inline void setFractureProperty(unsigned dim, MeshLib::Element const& e,
...
+    branch->slave_fracture_ID = slave_frac.fracture_id;
+    // set a normal vector from the master to the slave fracture
+    Eigen::Vector3d branch_vector =
+        slave_frac.point_on_fracture - branch->coords;
+    double sign = (branch_vector.dot(master_frac.normal_vector) < 0) ? -1 : 1;
+    branch->normal_vector_branch = sign * master_frac.normal_vector;
+    return branch;
+}
+
+inline JunctionProperty* createJunctionProperty(
```
wenqing: Here the same.

ProcessLib/LIE/Common/LevelSetFunction.cpp
```diff
@@ -26,11 +30,107 @@ namespace ProcessLib
...
+        for (std::size_t j = 0; j < frac->branches_slave.size(); j++)
+            enrich *= Heaviside(levelset_branch(*frac->branches_slave[j], x));
+        enrichments[i] = enrich;
+    }
+
+    // junctions
+    for (std::size_t i = 0; i < junction_props.size(); i++)
+    {
+        auto const* junction = junction_props[i];
+        auto fid1 = fracID_to_local.at(junction->fracture_IDs[0]);
```
wenqing: It is good to add const to these three definitions.

ProcessLib/LIE/Common/LevelSetFunction.cpp
```diff
@@ -26,11 +30,107 @@ namespace ProcessLib
...
+        for (auto const& branch : this_frac.branches_master)
+        {
+            if (branch->master_fracture_ID != this_frac.fracture_id)
+                continue;
+
+            if (fracID_to_local.find(branch->slave_fracture_ID) ==
+                fracID_to_local.end())
+                continue;
+
+            double singned = boost::math::sign(
```
wenqing: const

ProcessLib/LIE/Common/LevelSetFunction.cpp
```diff
@@ -26,11 +30,107 @@ namespace ProcessLib
...
+    }
+
+    // junctions
+    for (unsigned i = 0; i < junction_props.size(); i++)
+    {
+        auto const* junction = junction_props[i];
+        if (!BaseLib::contains(junction->fracture_IDs, this_frac.fracture_id))
+            continue;
+
+        auto another_frac_id =
```
wenqing: const

ProcessLib/LIE/HydroMechanics/LocalAssembler/HydroMechanicsLocalAssemblerMatrixNearFracture.h
```diff
@@ -72,6 +76,11 @@ class HydroMechanicsLocalAssemblerMatrixNearFracture
 
     static const int displacement_jump_index =
         displacement_index + displacement_size;
+
+    std::vector<FractureProperty*> _fracture_props;
```
wenqing: Might be good to use std::vector<std::reference_wrapper<const unique_ptr<FractureProperty>> > because the vector is pushed with process_data.fracture_property.get().

ProcessLib/LIE/HydroMechanics/LocalAssembler/HydroMechanicsLocalAssemblerMatrixNearFracture-impl.h
```diff
@@ -75,9 +79,9 @@ void HydroMechanicsLocalAssemblerMatrixNearFracture<
 
     // levelset value of the element
     // remark: this assumes the levelset function is uniform within an element
-    auto const& fracture_props = *_process_data.fracture_property;
```
wenqing: `fracture_props` was set as member of the class as `_fracture_props`. However I could not found where  `_fracture_props` is initialized. May be I am wrong.

ProcessLib/LIE/Common/FractureProperty.h
wenqing: I see.

ProcessLib/LIE/Common/LevelSetFunction.cpp
norihiro-w: not important as the scope is limited

ProcessLib/LIE/Common/LevelSetFunction.cpp
norihiro-w: not important as the scope is limited

ProcessLib/LIE/HydroMechanics/LocalAssembler/HydroMechanicsLocalAssemblerMatrixNearFracture.h
norihiro-w: thanks for the suggestion. however that way is too complicated for me

ProcessLib/LIE/HydroMechanics/LocalAssembler/HydroMechanicsLocalAssemblerMatrixNearFracture-impl.h
norihiro-w: `fracture_property` is initialized in CreateSmallDeformationProcess.cpp and moved to the process data. see https://github.com/ufz/ogs/blob/master/ProcessLib/LIE/SmallDeformation/CreateSmallDeformationProcess.cpp#L143

wenqing: I see that is initialized after creating its owner.

ProcessLib/LIE/Common/FractureProperty.h
wenqing: It is just related to the operator of `new`. If a `new` is used, one may try to find a `delete`. If you prefer to use it, it is OK. All other stuffs are OK for me too.

wenqing: benchmarks failed.

PR #2241, opened 2018-10-12
Fix volumetric source terms

TomFischer:
- ~~Fix wrong sign~~
 - ~~Change tests according to this fix and~~ update the documentation
 - Added non-constant source term test similar to the python non-constant source term test


ProcessLib/SourceTerms/VolumetricSourceTermFEM.h
```diff
@@ -105,7 +105,7 @@ class VolumetricSourceTermLocalAssembler final
             pos.setIntegrationPoint(ip);
             auto const st_val = _volumetric_source_term(t, pos)[0];
 
-            _local_rhs.noalias() +=
```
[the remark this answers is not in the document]
TomFischer: You are right. I used the wrong equation for the derivation of the analytical solution. I reversed the changes and updated the documentation.

ProcessLib/GroundwaterFlow/Tests.cmake
```diff
@@ -706,6 +706,18 @@ AddTest(
...
+AddTest(
+    NAME GroundWaterFlowProcess_VolumetricSourceTerm_sin_x_sin_y_square_1e3
+    PATH Elliptic/square_1x1_GroundWaterFlow
+    EXECUTABLE ogs
+    EXECUTABLE_ARGS square_1e3_volumetricsourcetermdataarray.prj
+    WRAPPER time
+    TESTER vtkdiff
+    REQUIREMENTS NOT (OGS_USE_LIS OR OGS_USE_MPI)
+    DIFF_DATA
+    square_1x1_quad_1e3_volumetricsourcetermdataarray.vtu square_1e3_volumetricsourcetermdataarray_pcs_0_ts_1_t_1.000000.vtu analytical_solution pressure 2e-2 1e-16
```
chleh: This compares to an analytical solution? (Maybe sin(x)*sin(y))? Otherwise the absolute tolerance would be pretty bad. Maybe you could mention the fact that you compare to an analytical solution, and also to which one.

TomFischer:
Thanks @chleh  and @endJunction for the review!

ProcessLib/GroundwaterFlow/Tests.cmake
```diff
@@ -706,6 +706,20 @@ AddTest(
...
+AddTest(
+    NAME GroundWaterFlowProcess_VolumetricSourceTerm_sin_x_sin_y_square_1e3
+    PATH Elliptic/square_1x1_GroundWaterFlow
+    EXECUTABLE ogs
+    EXECUTABLE_ARGS square_1e3_volumetricsourcetermdataarray.prj
+    WRAPPER time
+    TESTER vtkdiff
+    REQUIREMENTS NOT (OGS_USE_LIS OR OGS_USE_MPI)
+    # the analytical solution is: sin(2*Pi*x-Pi/2)*sin(2*Pi*y-Pi/2)
+    # the source term in the data array was set to: -2*(2*Pi)^2 * sin(2*Pi*x-Pi/2)*sin(2*Pi*y-Pi/2)
```
chleh: Thanks for adding this comment. Could be helpful if somebody has to touch this test case in the future.

PR #2243, opened 2018-10-23
Liquid flow improvements

TomFischer:
Use IntegrationPointData to accelerate the assembly. This saves ~ 3 % of the total run time for the two large benchmark cases.

### DIFF
--- a/ProcessLib/LiquidFlow/LiquidFlowLocalAssembler.h
+++ b/ProcessLib/LiquidFlow/LiquidFlowLocalAssembler.h
@@ -33,10 +33,8 @@
 
 namespace ProcessLib
 {
-
 namespace LiquidFlow
 {
-
 template <typename NodalRowVectorType, typename GlobalDimNodalMatrixType>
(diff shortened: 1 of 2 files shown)


chleh:
:+1:

ProcessLib/LiquidFlow/LiquidFlowLocalAssembler.h
```diff
@@ -33,9 +33,26 @@
...
-
 namespace LiquidFlow
 {
+template <typename NodalRowVectorType, typename GlobalDimNodalMatrixType>
+struct IntegrationPointData final
+{
+    explicit IntegrationPointData(NodalRowVectorType const& N_,
+                                  GlobalDimNodalMatrixType const& dNdx_,
+                                  double const& integration_weight_)
+        : N(std::move(N_)),
```
endJunction: no move will happen, most likely, because I doubt there is a constructor taking an `Eigen::Matrix<...> const&&`.
Drop the move or pass by value.

PR #2244, opened 2018-10-23
Deactivate  subdomains

wenqing:
This PR adds a feature to deactivate elements in the specified subdomains by material IDs. As a first case, SmallDeformaton is selected for the approach.

A benchmark is included in this PR as well. The  benchmark is a duplication of disc_with_hole but the mesh. Its mesh has two material subdomains, and the subdomain of  the quarter disk (see the left of the following figure) is deactivated for the assembly. The following figure compares the results of the present benchmark (left) and  disc_with_hole (right):
comparison

### DIFF
--- a/ProcessLib/SmallDeformation/CreateSmallDeformationProcess.cpp
+++ b/ProcessLib/SmallDeformation/CreateSmallDeformationProcess.cpp
@@ -109,6 +109,9 @@ createSmallDeformationProcess(
                 "deactivated_subdomains", std::vector<int>{});
         if (!deactivated_subdomains.empty())
         {
+            std::sort(deactivated_subdomains.begin(),
+                      deactivated_subdomains.end());
+
             // TODO: Modify constructDofTable() to exclude the nodes in the
             //       the deactivated subdomains from the mesh subset,
             //       and consequently remove the following warning.
--- a/ProcessLib/SmallDeformation/SmallDeformationProcessData.h
+++ b/ProcessLib/SmallDeformation/SmallDeformationProcessData.h
@@ -68,10 +68,9 @@ struct SmallDeformationProcessData
         if (!material_ids)
             return false;
 
-        return (std::find(deactivated_material_subdomains.begin(),
+        return (std::binary_search(deactivated_material_subdomains.begin(),
                           deactivated_material_subdomains.end(),
-                          (*material_ids)[element_id]) !=
-                deactivated_material_subdomains.end());
+                          (*material_ids)[element_id]));
     }
 
     MeshLib::PropertyVector<int> const* const material_ids;


ProcessLib/SmallDeformation/SmallDeformationProcessData.h
```diff
@@ -60,8 +63,25 @@ struct SmallDeformationProcessData
...
+            return false;
+        if (deactivated_material_subdomains.empty())
+            return false;
+
+        return (std::find(deactivated_material_subdomains.begin(),
+                          deactivated_material_subdomains.end(),
+                          (*material_ids)[element_id]) !=
+                deactivated_material_subdomains.end())
+                   ? true
+                   : false;
```
endJunction: Result of `!=` is already true or false...

ProcessLib/SmallDeformation/SmallDeformationFEM.h
```diff
@@ -209,6 +209,9 @@ class SmallDeformationLocalAssembler
         auto local_b = MathLib::createZeroedVector<NodalDisplacementVectorType>(
             local_b_data, local_matrix_size);
 
+        if (_process_data.isElementDeactivated(_element.getID()))
```
endJunction: I'd move the lookup into the constructor, since the element id is constant as well as the material ids and the `deactivated_material_ids`.

ProcessLib/SmallDeformation/SmallDeformationProcessData.h
```diff
@@ -60,8 +63,25 @@ struct SmallDeformationProcessData
...
+    }
+
     MeshLib::PropertyVector<int> const* const material_ids;
 
+    std::vector<int> deactivated_material_subdomains;
```
endJunction: const...

Tests/Data/Mechanics/Linear/square_with_deactivated_hole.prj
```diff
@@ -0,0 +1,165 @@
...
+        <process>
+            <name>SD</name>
+            <type>SMALL_DEFORMATION</type>
+            <integration_order>2</integration_order>
+            <constitutive_relation id="0">
+                <type>LinearElasticIsotropic</type>
+                <youngs_modulus>E</youngs_modulus>
+                <poissons_ratio>nu</poissons_ratio>
+            </constitutive_relation>
+            <constitutive_relation id="1">
```
endJunction: If id=1 is deactivated, there is no need to have the constitutive relation, I'd say.

Documentation/ProjectFile/prj/processes/process/SMALL_DEFORMATION/t_deactivated_subdomains.md
```diff
@@ -0,0 +1,3 @@
+Set of material subdomain indices. The elements in the subdomains by these
+indices will be skipped for local assembly.
```
endJunction: Set of material ids, which are excluded from local assembly.

endJunction:
How is the DOF table adapted? Isn't the matrix singular (because of zero rows/columns)?

Documentation/ProjectFile/prj/processes/process/SMALL_DEFORMATION/t_deactivated_subdomains.md
wenqing: Changed to
```
Set of material indices, the subdomains by which are excluded from local assembly.
```

Tests/Data/Mechanics/Linear/square_with_deactivated_hole.prj
wenqing: removed.

ProcessLib/SmallDeformation/SmallDeformationProcessData.h
wenqing: added.

ProcessLib/SmallDeformation/SmallDeformationFEM.h
wenqing: Move it to the constructor.

wenqing:
@endJunction There is no need to modify the DOF table. The only restriction is that there must be no boundary condition at any node inside the deactivated subdomains. The direct solver does not work with this approach. Since there is no need to modify the assignment of  BC, I added a running time warning about the restrictions.

ProcessLib/SmallDeformation/CreateSmallDeformationProcess.cpp
```diff
@@ -112,6 +112,16 @@ createSmallDeformationProcess(
             std::copy(deactivated_subdomains_optional->begin(),
                       deactivated_subdomains_optional->end(),
                       std::back_inserter(deactivated_subdomains));
+            WARN(
+                "There are subdoamins being deactivated."
```
endJunction: typo: subdomains

ProcessLib/SmallDeformation/CreateSmallDeformationProcess.cpp
```diff
@@ -112,6 +112,16 @@ createSmallDeformationProcess(
...
+                "Please make sure that there is no boundary condition "
+                "at any node inside these sudomains, "
+                "and only the iterative linear solver is utilized.");
+            INFO("The indices of the deactivated sudomains are");
```
endJunction: typo: subdomains.

endJunction:
@wenqing I see. I didn't check it, but I'd expect the direct solver is not able to solve a singular system.

Since the "deactivated subdomains" are known when constructing the DOF table, it should be possible to get a correct linear system of equations... There are nodes and elements selected for the DOF table construction; in case of deactivated elements, these are to be removed from the lists...

Documentation/ProjectFile/prj/processes/process/SMALL_DEFORMATION/t_deactivated_subdomains.md
endJunction: "by which are" yields zero hits in google.

ProcessLib/SmallDeformation/CreateSmallDeformationProcess.cpp
wenqing: corrected by amending the corresponding commit.

wenqing:
@endJunction For the sake of  efficiency, the DOF table have to be modified, i.e changing mesh subset of nodes for each process. The dimension of the system equations is reduced to that only determined by the number of the activated node. Then the restriction in the present implementation is gone.  For this, void Process::constructDofTable() has to be changed to a pure virtual member function or empty function, and its functionality has to be shifted to   its override   member of the derived classes. I would like to do it in future.

wenqing:
@endJunction  Besides, the change of constructDofTable() as that addressed above will not influence the changes in the local assembly in this PR.

ProcessLib/SmallDeformation/SmallDeformationFEM.h
```diff
@@ -209,6 +210,9 @@ class SmallDeformationLocalAssembler
         auto local_b = MathLib::createZeroedVector<NodalDisplacementVectorType>(
             local_b_data, local_matrix_size);
 
+        if (_dectivated)
```
endJunction: typo: deactivated.

ProcessLib/SmallDeformation/CreateSmallDeformationProcess.cpp
endJunction: There are still few of the 'su_domains'

ProcessLib/SmallDeformation/SmallDeformationProcessData.h
```diff
@@ -60,8 +63,23 @@ struct SmallDeformationProcessData
     //! Assignments are not needed.
     void operator=(SmallDeformationProcessData&&) = delete;
 
+    bool isElementDeactivated(const std::size_t element_id)
+    {
+        if (!material_ids)
+            return false;
+        if (deactivated_material_subdomains.empty())
```
endJunction: Didn't see first time. The case is already handled by find() == end.

endJunction:
I see that removing the elements from the assembly requires more modifications and also touches the output routines. I suggest, for at least to keep the linear equation system solvable, to put 1 on the diagonal.

wenqing:
@endJunction Actually this leads to a non singular subsystem term that can be solved by iterative linear solver if the elements of RHS outside the subsystem are zero. This is because that  the iterative solver obtains the solution by a serial multiplication computations of A and vectors, and the inverse of A is not involved. Then there are two options:
1)  as it is in the present PR,  not setting one on the diagonal entries associated with the inactive nodes. If there are boundary conditions are assigned at the inactive nodes, or  the elements of RHS outside the subsystem are not zero, the subsystem is extended to a singular one and the linear solver fails immediately. Under such condition, we have an abnormal  program exiting, which can prompt us to check the log for the reason. For this, I will add one more warning message in the code.

2)  setting one on the diagonal entries associated with the inactive nodes. This is easy to implement, e.g, assigning the identity to the local Jacobian if the element is deactivated. However, this leads to an always solvable system equations.  If  there are boundary conditions are assigned at the inactive nodes, the obtained solution is meaning less and we get no warnings from the computation.

Of course, if we modify the the DOF table for the case of activated/deactivated, this problem will be gone.

In the current stage, I prefer to take option 1).

ProcessLib/SmallDeformation/SmallDeformationFEM.h
```diff
@@ -209,6 +210,9 @@ class SmallDeformationLocalAssembler
         auto local_b = MathLib::createZeroedVector<NodalDisplacementVectorType>(
             local_b_data, local_matrix_size);
 
+        if (_is_deactivated)
+            return;
```
chleh: Maybe add a comment `// return after local_Jac and local_b have been resized and zeroed in order to avoid complaints from the VectorMatrixAssembler` (if that is the case)

ProcessLib/SmallDeformation/SmallDeformationProcessData.h
```diff
@@ -60,8 +63,25 @@ struct SmallDeformationProcessData
...
+            return false;
+
+        return (std::find(deactivated_material_subdomains.begin(),
```
chleh: maybe the vector could be ordered and a binary search could be used.

ProcessLib/SmallDeformation/SmallDeformationProcessData.h
```diff
@@ -60,8 +63,25 @@ struct SmallDeformationProcessData
     //! Assignments are not needed.
     void operator=(SmallDeformationProcessData&&) = delete;
 
+    bool isElementDeactivated(const std::size_t element_id)
```
chleh: Minor: could be const.

ProcessLib/SmallDeformation/CreateSmallDeformationProcess.cpp
```diff
@@ -98,8 +98,26 @@ createSmallDeformationProcess(
...
+    {
+        auto deactivated_subdomains_optional =
+            //! \ogs_file_param{prj__processes__process__SMALL_DEFORMATION__deactivated_subdomains}
+            config.getConfigParameterOptional<std::vector<int>>(
+                "deactivated_subdomains");
+        if (deactivated_subdomains_optional)
+        {
+            std::copy(deactivated_subdomains_optional->begin(),
+                      deactivated_subdomains_optional->end(),
+                      std::back_inserter(deactivated_subdomains));
```
chleh: Why not simply `deactivated_subdomains = std::move(*deactivated_subdomains_optional)`?

chleh: or `auto deactivated_subdomains = config.getConfigParameter(
                "deactivated_subdomains", std::vector<int>{});`

ProcessLib/SmallDeformation/SmallDeformationFEM.h
```diff
@@ -474,6 +475,9 @@ class SmallDeformationLocalAssembler
     SecondaryData<typename ShapeMatrices::ShapeType> _secondary_data;
     bool const _is_axially_symmetric;
 
+    /// Flag to indicate whether the local assembly is skipped or not.
+    bool const _is_deactivated;
```
chleh: Hm... What if, in the future, the set of deactivated subdomains changes over time?

Documentation/ProjectFile/prj/processes/process/SMALL_DEFORMATION/t_deactivated_subdomains.md
```diff
@@ -1,3 +1 @@
-Set of material subdomain indices. The elements in the subdomains by these
-indices will be skipped for local assembly.
-
+Set of material indices, the subdomains by which are excluded from local assembly.
```
chleh: Maybe better: `Set of material IDs. The subdomains associated to those material IDs are excluded from the local assembly.`

ProcessLib/SmallDeformation/SmallDeformationProcessData.h
```diff
@@ -33,6 +33,7 @@ struct SmallDeformationProcessData
 {
     SmallDeformationProcessData(
         MeshLib::PropertyVector<int> const* const material_ids_,
+        std::vector<int>&& deactivated_material_subdomains_,
```
chleh: Deactivation of subdomains looks like a very general feature. IMHO the proper solution would be to implement this feature on a higher level for all processes, not specifically for SD.
The current PR is a good starting point, but one should not forget future extensions.
Or is that feature only application relevant for SD?

ProcessLib/SmallDeformation/CreateSmallDeformationProcess.cpp
```diff
@@ -119,7 +119,10 @@ createSmallDeformationProcess(
                 "There are subdomains being deactivated."
                 "Please make sure that there is no boundary condition "
                 "at any node inside these subdomains, "
-                "and only the iterative linear solver is utilized.");
+                "and only the iterative linear solver is utilized.\n"
+                "If  there are boundary conditions  at the nodes inside these "
+                "subdomains, the program will exit by the failure of the "
+                "iterative linear.");
```
chleh: Maybe better: ` the program will terminate with linear solver failure.`

ProcessLib/SmallDeformation/CreateSmallDeformationProcess.cpp
```diff
@@ -119,7 +119,10 @@ createSmallDeformationProcess(
                 "There are subdomains being deactivated."
                 "Please make sure that there is no boundary condition "
                 "at any node inside these subdomains, "
-                "and only the iterative linear solver is utilized.");
+                "and only the iterative linear solver is utilized.\n"
+                "If  there are boundary conditions  at the nodes inside these "
```
chleh: Does `inside` mean "entirely within the deactivated region"?
In other words: Are BCs at the interface between activated and deactivated regions possible?

endJunction:
The problem with a matrix filled with zero rows and columns is that, the only usable solver is CG type with no preconditioner.
Neither LIS, nor PETSc, nor Eigen solvers in other combinations than CG/none are working properly. (Didn't test all of them, of course, but all the combinations of CG/BiCG + None/Jacobi/ILUT + direct solver if available.)
I didn't look, why Eigen CG/Jacobi works, but this is not the point.

TomFischer:
@wenqing:
_setting one on the diagonal entries associated with the inactive nodes. This is easy to implement, e.g, assigning the identity to the local Jacobian if the element is deactivated. However, this leads to an always solvable system equations. If there are boundary conditions are assigned at the inactive nodes, the obtained solution is meaning less and we get no warnings from the computation._

What do you mean by: _the obtained solution is meaning less_? Are all entries of the solution vector wrong or only the ones associated to the inactive nodes to which Dirichlet boundary conditions are set?

What are the difficulties to detect and omit the setting of boundary conditions to inactive nodes?

wenqing:
@endJunction  You are right, only Jacobian preconditoner of eigen works. The thing is that ILUT of eigen dose not  skip zero diagonal during the incomplete LU decomposition, for example as
 ```
   ... ILU loops
  {
     ....
      if (std::fabs(a_kk) > 0.0)
     {
         ...
          a_ik /= a_kk;
         ...
    }
   ....
}
```

PETSc ksp and pc (or  the third party pc, hypre) should work because they are tested in ogs5.

Unfortunately,  only  Jacobian preconditoner of Eigen works for the present approach.  I will consider your suggestion of "to put 1 on the diagonal". For this, I found that to set local Jacobian matrix as identity for deactivated element is not a right solution. If we perform "to put 1 on the diagonal" locally for a deactivated element, only the diagonal entries of local Jacobian that are  associated with the inactive nodes have to be set to one. This leads to more local computations. Therefore I will think about to put such assignment as a general computation after global assembly.

Still, I do not want to change the DOF table for the presented element deactivation approach because the present approach gives a flexible solution for  the set of deactivated subdomains changes over time, a  functionality mentioned by Christoph in his comments.

Under such situation, I would like to set this PR as TODO after replying Christoph's comments, and I will come back to this PR again once I  finish some ogs 5 stuff.

**Comment by endJunction** [comment]
Another approach (again suggested by Christoph) would be to remove the deactivated DOF's from the linear system. This could be extended to the Dirichlet BCs too. The DOF table does not need modification then.

**Comment by wenqing** [comment]
@TomFischer 
_What do you mean by: _the obtained solution is meaning less_? Are all entries of the solution vector wrong or only the ones associated to the inactive nodes to which Dirichlet boundary conditions are set?_

For this case, all the diagonal entries  that are associated inactive nodes are set to one.  If some inactive nodes are  accidentally assigned with  boundary conditions, their corresponding global RHS entries get values changed. If such changed values are non-zero, linear system will gives an unexpected solution.

_What are the difficulties to detect and omit the setting of boundary conditions to inactive nodes?_
It is not so difficult. For that, we have to change the DOF table for all processes.

File: ProcessLib/SmallDeformation/SmallDeformationFEM.h (hunk 10 above)
Comment (wenqing) [reply ProcessLib/SmallDeformation/SmallDeformationFEM.h]: Changed.

File: ProcessLib/SmallDeformation/SmallDeformationProcessData.h (hunk 11 above)
Comment (wenqing) [reply ProcessLib/SmallDeformation/SmallDeformationProcessData.h]: Done.

File: ProcessLib/SmallDeformation/SmallDeformationProcessData.h (hunk 12 above)
Comment (wenqing) [reply ProcessLib/SmallDeformation/SmallDeformationProcessData.h]: Added.

File: ProcessLib/SmallDeformation/CreateSmallDeformationProcess.cpp (hunk 13 above)
Comment (wenqing) [reply ProcessLib/SmallDeformation/CreateSmallDeformationProcess.cpp]: Changed.

File: ProcessLib/SmallDeformation/SmallDeformationFEM.h (hunk 14 above)
Comment (wenqing) [reply ProcessLib/SmallDeformation/SmallDeformationFEM.h]: Will consider it later.

File: Documentation/ProjectFile/prj/processes/process/SMALL_DEFORMATION/t_deactivated_subdomains.md (hunk 15 above)
Comment (wenqing) [reply Documentation/ProjectFile/prj/processes/process/SMALL_DEFORMATION/t_deactivated_subdomains.md]: Changed accordingly.

File: ProcessLib/SmallDeformation/SmallDeformationProcessData.h (hunk 16 above)
Comment (wenqing) [reply ProcessLib/SmallDeformation/SmallDeformationProcessData.h]: will move it to Process.h

Comment (wenqing) [reply ProcessLib/SmallDeformation/SmallDeformationProcessData.h]: or to ProcessData

File: ProcessLib/SmallDeformation/CreateSmallDeformationProcess.cpp (hunk 17 above)
Comment (wenqing) [reply ProcessLib/SmallDeformation/CreateSmallDeformationProcess.cpp]: Changed accordingly.

File: Tests/Data/Mechanics/Linear/square_with_deactivated_hole.prj [hunk 19]
```diff
@@ -0,0 +1,160 @@
...
+            <lis>-i cg -p jacobi -tol 1e-16 -maxiter 10000</lis>
+            <eigen>
+                <solver_type>CG</solver_type>
+                <precon_type>DIAGONAL</precon_type>
+                <max_iteration_step>10000</max_iteration_step>
+                <error_tolerance>1e-16</error_tolerance>
+            </eigen>
+            <petsc>
+                <prefix>sd</prefix>
+                <parameters>-sd_ksp_type cg -sd_pc_type bjacobi -sd_ksp_rtol 1e-16 -sd_ksp_max_it 10000</parameters>
```
endJunction: This PETSc configuration does not run (on my compilation at least) with error:
```
Divergence detected, ...
```

ProcessLib/SmallDeformation/CreateSmallDeformationProcess.cpp
wenqing: The nodes at the interface  between activated and deactivated regions  are still active, and they are be assigned with BCs.

wenqing:
Thanks for your comments!

PR #2250, opened 2018-10-29
[web/BGRa] Corrected two typos and made a consistency of one symbol within the context

wenqing:
as titled. Thanks  @KeitaYoshioka for pointing out these.


wenqing:
@endJunction  I am going to merge this because it is just for a tiny correction of benchmark documentation.

PR #2251, opened 2018-10-30
Prepare TM process for upcoming MFront integration.

chleh:
The TM process passes the mechanical strain to the solid material model. So far this mechanical strain has been a mix of total strain and the thermal strain increment.
That worked until now for the OGS material models, because there this "mixture" cancelled out in the end. However, it won't cancel out if we use material models provided by MFront.

In this PR the mechanical strain is properly computed with the total strain increment and the thermal strain increment.
This solution has been proposed by @fparisio.

In the TM-Phasefield process no change seems to be necessary, because there no incremental thermal strains seem to be used. Maybe somebody (@KeitaYoshioka) could confirm this.

### DIFF
--- a/ProcessLib/ThermoMechanics/ThermoMechanicsFEM.h
+++ b/ProcessLib/ThermoMechanics/ThermoMechanicsFEM.h
@@ -43,8 +43,13 @@ struct IntegrationPointData final
     {
     }
 
+    // total stress
     typename BMatricesType::KelvinVectorType sigma, sigma_prev;
-    typename BMatricesType::KelvinVectorType eps;
+
+    // total strain
+    typename BMatricesType::KelvinVectorType eps, eps_prev;
+
+    // mechanical strain
     typename BMatricesType::KelvinVectorType eps_m, eps_m_prev;
 
     MaterialLib::Solids::MechanicsBase<DisplacementDim> const& solid_material;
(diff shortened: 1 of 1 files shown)


ProcessLib/ThermoMechanics/ThermoMechanicsFEM.h
```diff
@@ -285,10 +295,11 @@ class ThermoMechanicsLocalAssembler
             // assume isotropic thermal expansion
             const double T_ip = N.dot(T);  // T at integration point
             eps_m.noalias() =
-                eps - linear_thermal_strain_increment * Invariants::identity2;
+                eps_m_prev + eps - eps_prev -
+                linear_thermal_strain_increment * Invariants::identity2;
+
             auto&& solution = _ip_data[ip].solid_material.integrateStress(
```
wenqing: For the stress integration, I think we only need two variables of strain, i.e. the increment of strain that including thermal strain increment, and the mechanical strain of the previous time step for some special material model. I will try it in future. In the current structure of the stress integration, the changes presented in the PR is quite good.

chleh: But in order that `pushBackState()` can update the state maybe one has to keep four strain variables anyway.

PR #2252, opened 2018-10-30
[MGTL] Allow multiple bulk element ids in identify

endJunction:
Upon request for Dirichlet BCs inside of the domain defined along lines.

@Thomas-TK Could you provide an example?

### DIFF
--- a/MeshGeoToolsLib/IdentifySubdomainMesh.cpp
+++ b/MeshGeoToolsLib/IdentifySubdomainMesh.cpp
@@ -92,19 +92,19 @@ std::vector<std::size_t> findElementsInMesh(
     return element_ids;
 }
 
-/// Tries to find unique elements' ids in the bulk mesh.
+/// Tries to find all elements' ids in the bulk mesh.
 /// The subdomain elements' nodes are mapped to the bulk mesh and are then used
 /// to identify the bulk mesh elements.
-std::vector<std::size_t> identifySubdomainMeshElements(
+std::vector<std::vector<std::size_t>> identifySubdomainMeshElements(
     MeshLib::Mesh const& subdomain_mesh, MeshLib::Mesh const& bulk_mesh)
 {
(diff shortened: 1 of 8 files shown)


Thomas-TK:
Of course, it's my pleasure !

Thomas-TK:
Here we go: an unstructured triangle mesh and a Y-line going through the domain (typical for river BCs).  Additionally, an according prism mesh. 
Thank you 
Thomas

[SimpleRiverExample.zip](https://github.com/ufz/ogs/files/2538198/SimpleRiverExample.zip)


PS: that's how it looks like
riverpointallocation

MeshGeoToolsLib/IdentifySubdomainMesh.cpp
```diff
@@ -133,21 +133,8 @@ std::vector<std::size_t> identifySubdomainMeshElements(
             OGS_FATAL(
                 "Expect exactly one element to be found in the bulk mesh.");
```
TomFischer: Is the error message still correct?

Thomas-TK:
Looks good with triangle mesh. . .will test it for prism meshes as well 
riverpointallocationresult

MeshGeoToolsLib/IdentifySubdomainMesh.cpp
endJunction: Fixed: "exactly one" to at "least one".

Thomas-TK:
I works for the 3D prism model as well. !!!!

Thomas-TK:
I tested triangles, line elements and individual points as  elements on a triangle and prism mesh. Works very well as BCs inside the domain. 
riverpointallocationresultprism

endJunction:
Put it in a ctest and add some docu on the benchmark page. Then, and only then, can you rely on these features being correct in the feature.

PR #2254, opened 2018-11-03
[App] Add project dir to python module path.

endJunction:
When the python modules are not in the current working directory but
in the projects directory, the user modules would not be found.
Adding project's directory to the python's module path allows to run ogs
from a different directory than the project file.

The error occurs, for example, if running the python piston  benchmark from the build directory like
```sh
bin/ogs ../ogs_source_dir/Tests/Data/Mechanics/Linear/PythonPiston/piston.prj
```
It would fail to load the `chamber` module from the [project directory](https://github.com/ufz/ogs/tree/master/Tests/Data/Mechanics/Linear/PythonPiston).

### DIFF
--- a/Applications/ApplicationsLib/ProjectData.cpp
+++ b/Applications/ApplicationsLib/ProjectData.cpp
@@ -200,6 +200,11 @@ ProjectData::ProjectData(BaseLib::ConfigTree const& project_config,
     {
 #ifdef OGS_USE_PYTHON
         namespace py = pybind11;
+
+        // Append project's directory to python's module search path.
+        py::module::import("sys").attr("path").attr("append")(
+            project_directory);
+
         auto const script_path =
             BaseLib::copyPathToFileName(*python_script, project_directory);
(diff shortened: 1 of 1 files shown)


chleh:
The solution looks OK. But two things:
* adding the project dir is not necessary to import the `OpenGeoSys` module. (Sure, you never claimed that.)
* maybe we will want to use the project dir also for other things, e.g., loading additional assets such as csv files. Therefore  maybe the project dir should be exposed explicitly, e.g., in the `OpenGeoSys` module.

NB: One can get the path of the python script with
`p = os.path.realpath(__file__)` (https://stackoverflow.com/a/595332)
and then one could do something like
`sys.path.append(os.path.dirname(p))` purely from within the python script. That, however, is probably pretty user unfriendly.

endJunction:
The `__file__` variable is not set. Adding project dir to `OpenGeoSys` module is good idea. Noted elsewhere.

PR #2255, opened 2018-11-04
OgsTester. Add test definitions to project files.

endJunction:
Move test definitions into the project files and add a "test mode" to the ogs executable.

PR #2257, opened 2018-11-06
Misc for OgsTester

endJunction:
Extracted from [OgsTester PR](https://github.com/ufz/ogs/pull/2255).

Few changes found during the OgsTester implementation. The extracted function in the VtuInterface is reused in OgsTester PR.


endJunction:
Hmmm.... Something is wrong there. Thinking aloud....
Of the three modified test the axisymmetric_pipe.prj is executed normally with Eigen, and it does not matter if the `solid_density` is there or not. But the hollow_sphere.prj (which is enabled for LIS only) does not run w/o `solid_density`.

endJunction:
`hollow_sphere` test runs successfully with Eigen using CG/Diagonal linear solver.
Now all of the changed tests are Eigen enabled and do run (on my machine at least ;).

@bilke singularity1 hit a singularity ;(

PR #2258, opened 2018-11-06
Modified and renamed Test for identifySubdomains.

Thomas-TK:
The test includes geometric points, lines and surfaces and tests their bulk node assignment for triangle and prism meshes.


Applications/Utils/Tests.cmake
```diff
@@ -69,29 +69,41 @@ AddTest(
 )
 
 AddTest(
-    NAME identifySubdomains_riverTriangleMesh
+    NAME identifySubdomains_PntLineSFC_DomainTRIMesh
```
endJunction: PointLineSurface would be more readable than mixing three types of naming.
I'd call the whole thing `identifySubdomains_pointLineSurfaceOnTriangleMesh`.

Applications/Utils/Tests.cmake
```diff
@@ -69,29 +69,41 @@ AddTest(
...
     PATH MeshGeoToolsLib/IdentifySubdomains
     EXECUTABLE identifySubdomains
-    EXECUTABLE_ARGS -m river_domain_triangle.vtu -o ${Data_BINARY_DIR}/MeshGeoToolsLib/IdentifySubdomains/triangle_ -- river_bc.vtu
+    EXECUTABLE_ARGS -m PntLineSFC_DomainTRIMesh.vtu -o ${Data_BINARY_DIR}/MeshGeoToolsLib/IdentifySubdomains/TRI_ -- PntLineSFC_BC_Point.vtu PntLineSFC_BC_Line.vtu PntLineSFC_BC_Surface.vtu
```
endJunction: Same here; why not `triangle_mesh.vtu` and `point.vtu`, `line.vtu`, `surface.vtu`?
Keep in mind, other people have to read the text many times---readability is important.

Thomas-TK: Thanks, I will rename it again.

PR #2259, opened 2018-11-06
Use Solid Material Models from MFront

chleh:
TODO
* [x] Reference solutions to remaining test cases
* [x] squash commits

Follow-up
* MFront parameters (currently not supported by MGIS)
* Read/write internal state variables
* Maybe initially compute self-consistent internal state based on ICs
* Enable CI-Tests
* Non-isotropic materials (essentially only tests needed)
* Other modelling hypotheses than 3D and plane strain


chleh:
Btw: Thanks to @thelfer for MGIS, and for @fparisio for discovering MGIS and for support with the tests.

thelfer:
Hi everyone,
I am quite impressed that @chleh was able to use MGIS with has been developped quite recently without much focus on documentation !
Do not hesitate to make feed-backs and request new features, I will be happy to help !
I am planning to add proper support for orthotropic behaviours (by providing the free functions to rotate tensors). I'll do this just after finishing the `python` bindings. BTW, I also know that the `generic` interface has a few shortcomings to I need to tackle.
Regards,
Thomas

MaterialLib/SolidModels/MFront/MFront.cpp
```diff
@@ -0,0 +1,336 @@
...
+    static_assert(Derived::ColsAtCompileTime != Eigen::Dynamic, "Error");
+
+    typename Derived::PlainObject n;
+
+    // optimal for row-major storage order
+    for (std::ptrdiff_t r = 0; r < Eigen::DenseBase<Derived>::RowsAtCompileTime;
+         ++r)
+    {
+        auto const R = OGSToMFront(r);
+        for (std::ptrdiff_t c = 0;
```
endJunction: Just curious, why ptrdiff here, not an int (which is the type for rows/cols at compile time)?

chleh: It's somewhat mixed. Eigen's `operator()` uses [`Eigen::Index`](http://eigen.tuxfamily.org/dox/classEigen_1_1DenseCoeffsBase_3_01Derived_00_01WriteAccessors_01_4.html#ae6ba07bad9e3026afe54806fdefe32bb),
Eigen::Index by default is [`std::ptrdiff_t`](https://bitbucket.org/eigen/eigen/src/e58451e50e9c4ea60cf1a054c4fdc376cdcbd217/Eigen/src/Core/util/Macros.h?at=default&fileviewer=file-view-default#Macros.h-33).

PR #2261, opened 2018-11-09
Make local source term definition possible

TomFischer:
This PR restricts the definition of the source term to a sub-domain of the entire simulation domain.


ProcessLib/SourceTerms/CreateNodalSourceTerm.h
```diff
@@ -34,8 +34,8 @@ namespace ProcessLib
 {
 std::unique_ptr<SourceTerm> createNodalSourceTerm(
     BaseLib::ConfigTree const& config, MeshLib::Mesh const& st_mesh,
-    const NumLib::LocalToGlobalIndexMap& dof_table, std::size_t mesh_id,
-    const int variable_id, const int component_id,
+    std::unique_ptr<NumLib::LocalToGlobalIndexMap>&& dof_table,
```
endJunction: [discussion] I know, we have it in several other places done the same way, but I think that `&&` is not needed for a `unique_ptr` because it is move only anyway. Passing it by value is sufficient.

TomFischer: Yes, this works. I changed it in nodal, volumetric and the Python source term implementations and rebased.

chleh: IMHO the rule "Add `&&` to a type if you don't allow copies." is simpler "Add `&&` to a type if you don't allow copies. But if the type is move-only, you can skip the `&&`."

ProcessLib/SourceTerms/CreateSourceTerm.cpp
```diff
@@ -22,35 +22,60 @@ namespace ProcessLib
...
+            config.config, config.mesh, std::move(dof_table_source_term),
+            parameters, integration_order, shapefunction_order, variable_id,
             *config.component_id);
     }
 
     if (type == "Python")
     {
 #ifdef OGS_USE_PYTHON
+        // Create local DOF table from the source term mesh subset for the
+        // given variable and component id.
```
chleh: In Python probably for all vars/comps.
Probably, these comments can all be removed. They just describe the already very descriptive line of code that follows them.

Tests/Data/Elliptic/square_1x1_GroundWaterFlow/square_1e2_volumetricsourceterm.prj
```diff
@@ -2,6 +2,7 @@
 <OpenGeoSysProject>
     <meshes>
         <mesh>square_1x1_quad_1e2.vtu</mesh>
+        <mesh>square_1x1_quad_1e2_volumetric_source_term.vtu</mesh>
```
chleh: This is still the entire mesh, isn't it? Is it special in any way? E.g., does it have an element property like `original_element_ids`? Btw. I didn't find code changes related to mapping d.o.f.s on the source term mesh to d.o.f.s on the whole domain. Does the current code work if the source term is applied to parts of the mesh only (in particular if the ST subset does not start from element ID 0)?

ProcessLib/SourceTerms/CreateSourceTerm.cpp
```diff
@@ -29,6 +29,35 @@ std::unique_ptr<SourceTerm> createSourceTerm(
...
 
+    // check basic data consistency
+    if (variable_id >=
+            static_cast<int>(dof_table_bulk.getNumberOfVariables()) ||
+        *config.component_id >=
+            dof_table_bulk.getNumberOfVariableComponents(variable_id))
+    {
+        OGS_FATAL(
+            "Variable id or component id too high. Actual values: (%d, "
+            "%d), maximum values: (%d, %d).",
```
chleh: Btw.: the "maximum" is off by one. What is printed is in C style one past maximum. This is confusing. It's wrong everywhere, and I think I made it wrong for the first time. :smile:

Tests/Data/Elliptic/square_1x1_GroundWaterFlow/square_1e2_volumetricsourceterm.prj
TomFischer: Created test for strict subset.

ProcessLib/SourceTerms/CreateSourceTerm.cpp
TomFischer: In `dof_table_bulk.getNumberOfVariables()` the `size - 1` is returned, so the message for the number of variables should be consistent. `dof_table_bulk.getNumberOfVariableComponents(variable_id)` returns a difference where the off by one should be canceled out.

ProcessLib/HeatConduction/Tests.cmake
```diff
@@ -55,3 +55,15 @@ AddTest(
...
+AddTest(
+    NAME 1D_HeatConduction_dirichlet_SourceTerm
+    PATH Parabolic/T/1D_dirichlet_source-term
+    EXECUTABLE ogs
+    EXECUTABLE_ARGS line_1_line_1e2_source_term.prj
+    TESTER vtkdiff
+    DIFF_DATA
+    line_1_line_1e2_pcs_0_ts_500_t_39062500.000000_reference.vtu line_1_line_1e2_pcs_0_ts_500_t_39062500.000000.vtu temperature temperature 1e-11 0.0
+    REQUIREMENTS NOT OGS_USE_MPI
+)
```
wenqing: Could you please add a test for PETSc as
```
# test the source term on a subdomain with the PETSc embedded executable file
AddTest(
    NAME 1D_HeatConduction_dirichlet_SourceTerm_PETSc
    PATH Parabolic/T/1D_dirichlet_source-term
    EXECUTABLE_ARGS line_1_line_1e2_source_term.prj
    WRAPPER mpirun
    WRAPPER_ARGS -np 1
    TESTER vtkdiff
    REQUIREMENTS OGS_USE_MPI
    DIFF_DATA
    line_1_line_1e2_pcs_0_ts_500_t_39062500.000000_reference.vtu line_1_line_1e2_pcs_0_ts_500_t_39062500.000000_0.vtu temperature temperature 1e-11 0.0
)
```

Tests/Data/Elliptic/square_1x1_GroundWaterFlow/square_1e2_volumetricsourceterm.prj
chleh: OK, thanks.

PR #2263, opened 2018-11-13
Fix compiling with boost

fparisio:


endJunction:
Has nothing to do with boost, but with missing uniform initialization of tuple in gcc prior 4.9.x. Implemented in later versions.

PR #2264, opened 2018-11-14
Use new overloaded getPropertyVector

TomFischer:
Using the new overloaded `Properties::getPropertyVector()`  the user can specify additionally to the data type and the name also the `MeshItemType` and the number of components.  These features don't need to be tested later in the user code.

### DIFF
--- a/Applications/Utils/FileConverter/generateMatPropsFromMatID.cpp
+++ b/Applications/Utils/FileConverter/generateMatPropsFromMatID.cpp
@@ -52,31 +52,38 @@ int main (int argc, char* argv[])
     cmd.parse( argc, argv );
 
     // read mesh
-    std::unique_ptr<MeshLib::Mesh> mesh(MeshLib::IO::readMeshFromFile(mesh_arg.getValue()));
-    if (!mesh) {
+    std::unique_ptr<MeshLib::Mesh> mesh(
+        MeshLib::IO::readMeshFromFile(mesh_arg.getValue()));
+
+    if (!mesh)
+    {
         INFO("Could not read mesh from file \"%s\".", mesh_arg.getValue().c_str());
         return EXIT_FAILURE;
     }
-    if (!mesh->getProperties().existsPropertyVector<int>("MaterialIDs"))
+
+    MeshLib::PropertyVector<int>* materialIds = nullptr;
+    try
     {
-        ERR("Mesh contains no int-property vector named \"MaterialIds\".");
+        materialIds = mesh->getProperties().getPropertyVector<int>(
+            "MaterialIDs", MeshLib::MeshItemType::Cell, 1);
+    }
+    catch (std::runtime_error const& e)
+    {
+        WARN("%s", e.what());
         return EXIT_FAILURE;
     }
-    auto materialIds = mesh->getProperties().getPropertyVector<int>("MaterialIDs");
 
     std::size_t const n_properties(materialIds->size());
-    if (n_properties != mesh->getNumberOfElements()) {
-        ERR("Size mismatch: number of elements (%u) != number of material "
-            "properties (%u).", mesh->getNumberOfElements(), n_properties);
-        return EXIT_FAILURE;
-    }
-    std::string const name = BaseLib::extractBaseNameWithoutExtension(mesh_arg.getValue());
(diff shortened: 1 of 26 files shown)


Applications/Utils/FileConverter/generateMatPropsFromMatID.cpp
```diff
@@ -52,47 +52,58 @@ int main (int argc, char* argv[])
...
+    try
     {
-        for (std::size_t i=0; i<n_properties; ++i)
-            out_prop << i << "\t" << (*materialIds)[i] << "\n";
-        out_prop.close();
+        auto materialIds = mesh->getProperties().getPropertyVector<int>(
+            "MaterialIDs", MeshLib::MeshItemType::Cell, 1);
+
+        std::size_t const n_properties(materialIds->size());
+        if (n_properties != mesh->getNumberOfElements())
```
endJunction: This seems to be already checked by getPropertyVector because of the Cell requirement...

Applications/Utils/FileConverter/generateMatPropsFromMatID.cpp
```diff
@@ -52,47 +52,58 @@ int main (int argc, char* argv[])
...
+        INFO("Writing mesh to file \"%s\".", new_mshname.c_str());
+        MeshLib::IO::writeMeshToFile(*mesh, new_mshname);
+
+        INFO("New files \"%s\" and \"%s\" written.", new_mshname.c_str(),
+             new_matname.c_str());
+
+        return EXIT_SUCCESS;
     }
-    else
+    catch (std::runtime_error const& e)
```
endJunction: to be discussed: I'd try to wrap only the getPropertyVector call.

MeshLib/Properties.h
```diff
@@ -103,6 +111,14 @@ class Properties
                                                MeshItemType const item_type,
                                                int const n_components) const;
 
+    /// Non-const version of getPropertyVector returns a property vector with
+    /// given \c name, \c item_type and \c number_of_components or calls
+    /// OGS_FATAL if no such property vector exists.
+    template <typename T>
+    PropertyVector<T>* getPropertyVector(std::string const& name,
```
endJunction: for discussion: I think the `existsPropertyVector` is obsolete now, since a returned nullptr is indicating exactly the same...

Applications/Utils/FileConverter/generateMatPropsFromMatID.cpp
TomFischer: Will change it to assert.

MeshLib/Properties.h
TomFischer: In some cases it is necessary to check the existence before accessing the property vector. Maybe using new techniques instead of exceptions, for instance, returning a tuple will be implemented in the future.

MeshLib/MeshEditing/ElementValueModification.cpp
```diff
@@ -71,20 +76,24 @@ bool ElementValueModification::replace(MeshLib::Mesh &mesh,
...
 {
-    auto* property_value_vector =
-        mesh.getProperties().getPropertyVector<int>("MaterialIDs");
-
-    if (!property_value_vector)
+    MeshLib::PropertyVector<int>* property_value_vector = nullptr;
+    try
+    {
+        property_value_vector = mesh.getProperties().getPropertyVector<int>(
+            "MaterialIDs", MeshLib::MeshItemType::Node, 1);
```
endJunction: Must be `Cell`.

MeshLib/MeshEditing/ElementValueModification.cpp
```diff
@@ -97,18 +106,23 @@ std::size_t ElementValueModification::condense(MeshLib::Mesh &mesh)
...
 {
-    auto* const property_value_vector =
-        mesh.getProperties().getPropertyVector<int>("MaterialIDs");
-
-    if (!property_value_vector)
+    MeshLib::PropertyVector<int>* property_value_vector = nullptr;
+    try
+    {
+        property_value_vector = mesh.getProperties().getPropertyVector<int>(
+            "MaterialIDs", MeshLib::MeshItemType::Node, 1);
```
endJunction: dto.

PR #2268, opened 2018-11-16
Impl. of getFlux() in ComponentTransport process

TomFischer:
In order to calculate the water flux on a subset surface of the simulation domain the user has to provide the surface as a vtu file.  The PR contains a test example computing the water flux for the ComponentTransport process.


TomFischer:
One test fails at the moment in some tested configurations. It is not clear to me why.  I can not reproduce this failing tests on my laptop, on envinf1 using my usual user account and on envinf1 as jenkins user. The reason for the failing test is that the number of data array tuples in the reference file (1089) and the simulation output (1331). I haven't any idea what is the reason for this.

ProcessLib/ComponentTransport/ComponentTransportFEM.h
```diff
@@ -336,6 +336,66 @@ class LocalAssemblerData : public ComponentTransportLocalAssemblerInterface
...
+
+        FemType fe(*static_cast<const typename ShapeFunction::MeshElement*>(
+            &_element));
+
+        typename ShapeMatricesType::ShapeMatrices shape_matrices(
+            ShapeFunction::DIM, GlobalDim, ShapeFunction::NPOINTS);
+
+        // Note: Axial symmetry is set to false here, because we only need dNdx
+        // here, which is not affected by axial symmetry.
+        fe.computeShapeFunctions(pnt_local_coords.getCoords(), shape_matrices,
```
endJunction: There is a template parameter of computeShapeFunctions to compute dNdx only.

ProcessLib/ComponentTransport/ComponentTransportFEM.h
```diff
@@ -336,6 +336,66 @@ class LocalAssemblerData : public ComponentTransportLocalAssemblerInterface
...
+
+        if (_process_data.has_gravity)
+        {
+            auto const rho_w =
+                _process_data.fluid_properties->getValue(
+                    MaterialLib::Fluid::FluidPropertyType::Density, vars);
+            auto const b = _process_data.specific_body_force;
+            q += K_over_mu * rho_w * b;
+        }
+        Eigen::Vector3d flux;
```
endJunction: Must be initialized with zero, otherwise the last component(s) could contain garbage.

ProcessLib/ComponentTransport/ComponentTransportProcess.cpp
```diff
@@ -88,6 +92,46 @@ void ComponentTransportProcess::assembleWithJacobianConcreteProcess(
         dx_dx, M, K, b, Jac, _coupled_solutions);
 }
 
+Eigen::Vector3d ComponentTransportProcess::getFlux(std::size_t element_id,
```
endJunction: const element_id

ProcessLib/ComponentTransport/ComponentTransportFEM.h
```diff
@@ -336,6 +336,66 @@ class LocalAssemblerData : public ComponentTransportLocalAssemblerInterface
         return Eigen::Map<const Eigen::RowVectorXd>(N.data(), N.size());
     }
 
+    Eigen::Vector3d getFlux(MathLib::Point3d const& pnt_local_coords,
+                            double const t,
+                            std::vector<double> const& local_x) const override
+    {
+        // eval dNdx and invJ at given point
```
endJunction: A latter comment says that only dNdx is needed...

TomFischer: Fixes comment.

ProcessLib/ComponentTransport/ComponentTransportFEM.h
```diff
@@ -336,6 +336,70 @@ class LocalAssemblerData : public ComponentTransportLocalAssemblerInterface
...
+        {
+            auto const rho_w =
+                _process_data.fluid_properties->getValue(
+                    MaterialLib::Fluid::FluidPropertyType::Density, vars);
+            auto const b = _process_data.specific_body_force;
+            q += K_over_mu * rho_w * b;
+        }
+        Eigen::Vector3d flux(0.0, 0.0, 0.0);
+        flux.head<GlobalDim>() = q;
+        return flux;
```
wenqing: Just one question: 
For most of cases, the flux through the surface is interesting. Is it good to return a scalar number of  q * n? where n is the normal to the surface at the point.

TomFischer: @wenqing This method is not integrating over a face. It returns only the flux at the specific point `pnt_local_coords` of the current process. The integration you suggested is done in the more general SurfaceFlux process.
Each process that wants to output fluxes has more or less only to provide `getFlux()`.

wenqing: I see. The current implementation is good for its purpose.  :fast_forward: . 
I just thought about that the surface could be an interface of domains, where the normal flux at node without  integration is needed for some coupling applications. Of course, it is not for this PR.

TomFischer: This method return the flux vector in the point `pnt_local_coords`. If you need a scalar value (q * n) then you have to integrate over the surface elements gauss points. This is done in class `SurfaceFlux`.

PR #2271, opened 2018-11-27
BHE_1U flow properties update and bugfixes.

endJunction:
Use already extracted code. This prevents code duplication for 2U and the CXA/CXC implementations.

@ChaofanChen Could you please review this?

### DIFF
--- a/ProcessLib/HeatTransportBHE/BHE/Pipe.h
+++ b/ProcessLib/HeatTransportBHE/BHE/Pipe.h
@@ -29,10 +29,19 @@ struct Pipe
     double const wall_thickness;
     double const wall_thermal_conductivity;
 
+    /// Area of the pipe's inside without the wall.
     double area() const
     {
         constexpr double pi = boost::math::constants::pi<double>();
         return pi * diameter * diameter / 4;
+    }
+
+    /// Area of the pipe's outside including the wall thickness.
+    double outsideArea() const
+    {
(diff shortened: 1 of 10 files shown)


ChaofanChen:
@endJunction I'd love to. Thank you very much.

ChaofanChen:
@endJunction Could you please move the implementation of the constructor and other functions in "CreateFlowAndTemperatureControl.h" to the cpp file? In the case of the CXA/CXC or 2U BHE, we also need to call the function named "createFlowAndTemperatureControl", and this will cause an error of multiple definition when compiling. Besides, we also need to enrich flow and temperature boundary conditions including coupled with heat pump. I am not sure if my understanding is right, but please consider it.

ProcessLib/HeatTransportBHE/BHE/Pipe.h
```diff
@@ -29,11 +29,20 @@ struct Pipe
...
     {
         constexpr double pi = boost::math::constants::pi<double>();
         return pi * diameter * diameter / 4;
     }
+
+    /// Area of the pipe's outside including the wall thickness.
+    double outsideArea() const
+    {
+        constexpr double pi = boost::math::constants::pi<double>();
+        double const d = diameter + wall_thickness;
```
TomFischer: 2 * wall_thickness?

PR #2272, opened 2018-11-27
Dirichlet boundary condition within a time interval

wenqing:
This PR presents a Dirichlet boundary condition within a time interval.

The benchmark is about a square domain under  liquid pressure on its top and bottom. On the top boundary, pressure is always 1.0e+5, while on the bottom boundary, pressure is  1.0e+6 within a time interval of [0, 10] (see the following figure):
tintervalbc
```
            Left, t<=10.0;                            right, t>10
```


NumLib/TimeStepping/TimeInterval.h
```diff
@@ -0,0 +1,41 @@
...
+ */
+#pragma once
+
+namespace NumLib
+{
+/*!
+ * Class for a time interval, which has a member to check whether the given time
+ * is in this time interval.
+ */
+class TimeInterval
```
endJunction: final

NumLib/TimeStepping/TimeInterval.h
```diff
@@ -0,0 +1,41 @@
...
+    TimeInterval(const double start_time, const double end_time)
+        : _start_time(start_time), _end_time(end_time)
+    {
+    }
+
+    bool isInThisTimeInterval(const double current_time) const
+    {
+        return (current_time >= _start_time && current_time <= _end_time)
+                   ? true
+                   : false;
```
endJunction: The result of comparison is already a boolean, I'd drop the `true:false`

NumLib/TimeStepping/TimeInterval.h
```diff
@@ -0,0 +1,41 @@
...
+ */
+class TimeInterval
+{
+public:
+    TimeInterval(const double start_time, const double end_time)
+        : _start_time(start_time), _end_time(end_time)
+    {
+    }
+
+    bool isInThisTimeInterval(const double current_time) const
```
endJunction: Maybe `isActive`, s.t.
```c++
TimeInterval time_interval{start, end};
...
if (time_interval.isActive(t)) { ... }
```

ProcessLib/BoundaryCondition/DirichletBoundaryConditionWithinTimeInterval.cpp
```diff
@@ -0,0 +1,101 @@
...
+{
+}
+
+void DirichletBoundaryConditionWithinTimeInterval::getEssentialBCValues(
+    const double t, GlobalVector const& x,
+    NumLib::IndexValueVector<GlobalIndexType>& bc_values) const
+{
+    if (_time_interval->isInThisTimeInterval(t))
+    {
+        return getEssentialBCValuesLocal(t, x, bc_values);
```
endJunction: The return type of the function is `void`...

ProcessLib/BoundaryCondition/DirichletBoundaryConditionWithinTimeInterval.cpp
```diff
@@ -0,0 +1,101 @@
...
+    }
+#endif  // USE_PETSC
+
+    const double start_time =
+        //! \ogs_file_param{prj__process_variables__process_variable__boundary_conditions__boundary_condition__DirichletWithinTimeInterval__start_time}
+        config.getConfigParameter<double>("start_time");
+
+    const double end_time =
+        //! \ogs_file_param{prj__process_variables__process_variable__boundary_conditions__boundary_condition__DirichletWithinTimeInterval__end_time}
+        config.getConfigParameter<double>("end_time");
```
endJunction: Just a suggestion: I'd rather extract the config parsing into `createTimeInterval()` and have the `<time_interval>` tag in the project file:
```xml
...
<time_interval>
    <start>0</start>
    <end>1</end>
</time_interval>
```
Because parsing of the time interval is not a task for a boundary condition, in my opinion.

NumLib/TimeStepping/TimeInterval.h
```diff
@@ -0,0 +1,41 @@
...
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ * File:   TimeInterval.h
+ *
+ * Created on November 26, 2018, 4:44 PM
+ */
+#pragma once
+
+namespace NumLib
```
TomFischer: Why is it in NumLib? I would put it in BaseLib.

NumLib/TimeStepping/TimeInterval.h
wenqing: added.

NumLib/TimeStepping/TimeInterval.h
wenqing: changed.

NumLib/TimeStepping/TimeInterval.h
wenqing: changed as  `if (time_interval.contains(t))`.

ProcessLib/BoundaryCondition/DirichletBoundaryConditionWithinTimeInterval.cpp
wenqing: changed.

ProcessLib/BoundaryCondition/DirichletBoundaryConditionWithinTimeInterval.cpp
wenqing: Added.

NumLib/TimeStepping/TimeInterval.h
wenqing: At beginning of my implementation, I had put it in BaseLib. Later on, I thought that this class is related to the time stepping only. Therefore I moved it to  NumLib/TimeStepping. If you still think BaseLib is the best choice for it after reading this reply, I will move it  BaseLib again.

wenqing:
@endJunction Introduced two functions to avoid class inheritance while to avoid source code duplication.

NumLib/TimeStepping/TimeInterval.h
endJunction: Could be BaseLib because it's rather general, or ProcessLib/BC/ because this is the only place where it is used now.

wenqing: Since both of you prefer to move to BaseLib, I have  moved it to there but kept the path of parameter documentation.

PR #2275, opened 2018-11-28
BHE; Move the function body in "CreateFlowAndTemperatureControl.h" to cpp file.

ChaofanChen:
When adding cases of the CXA/CXC or 2U BHE, we also need to call the function named "createFlowAndTemperatureControl", and this will cause an error that multiple inclusion of the same definition in header file when compiling. Besides, we also need to enrich flow and temperature boundary conditions including the BHE coupled with heat pump.

### DIFF
--- a/ProcessLib/HeatTransportBHE/BHE/CreateFlowAndTemperatureControl.cpp
+++ b/ProcessLib/HeatTransportBHE/BHE/CreateFlowAndTemperatureControl.cpp
@@ -10,7 +10,7 @@
  */
 
 #include "BaseLib/ConfigTree.h"
-#include "BaseLib/Error.h"
+#include "MathLib/InterpolationAlgorithms/PiecewiseLinearInterpolation.h"
 #include "BaseLib/Algorithm.h"
 
 #include "CreateFlowAndTemperatureControl.h"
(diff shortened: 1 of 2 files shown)


ProcessLib/HeatTransportBHE/BHE/CreateFlowAndTemperatureControl.h
```diff
@@ -11,6 +11,9 @@
 
 #pragma once
 
+#include "BaseLib/ConfigTree.h"
+#include "BaseLib/Error.h"
+#include "FlowAndTemperatureControl.h"
 #include "RefrigerantProperties.h"
```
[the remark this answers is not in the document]
ChaofanChen: Thanks for the comment, modified as suggested.

ProcessLib/HeatTransportBHE/BHE/CreateFlowAndTemperatureControl.cpp
```diff
@@ -0,0 +1,90 @@
...
+namespace BHE
+{
+FlowAndTemperatureControl createFlowAndTemperatureControl(
+    BaseLib::ConfigTree const& config,
+    std::map<std::string,
+             std::unique_ptr<MathLib::PiecewiseLinearInterpolation>> const&
+        curves,
+    RefrigerantProperties const& refrigerant)
+{
+    auto find_curve_or_error = [&](std::string const& name,
```
TomFischer: You can use the very similar function `getOrError()` from `BaseLib/Algorithm.h`.

ChaofanChen: Thanks for your comment. Using the existed function getOrError().

PR #2278, opened 2018-11-30
Restructure output functionality.

TomFischer:
Move the ProcessOutput from ProcessData into class Output. The tags in the project files have also to 
be moved. This is done by the following  gist https://gist.github.com/TomFischer/fa68620af095e67d1e3bb9c2f670365e that uses `xmlstarlet`.


endJunction:
@ufz/members Please be prepared to update your project files once this is merged. Tom has prepared a short script (see PR description) and as an example the changes to the project files. There is no simple way to keep both types of the output configuration.

The PR merge is planned for Friday, the 7.Dec.2018, evening.

Issue #2276, opened 2018-11-29
Bug in the Neumann BC for the Taylor-Hood element


wenqing:
For processes that use Taylor-Hood element pair, the submesh for  the Neumann BC is created with high order element only, i.e. the number of the base nodes is identical to the number of total nodes. This causes vector sizes mismatching in NeumannBoundaryCondition::assemble() when the Neumann BC for the process with lower order element is assembled. 

The bug can be shown by running benchmark /HydroMechanics/Linear/Unconfined_Compression_late/square_1e2_UC_late.prj with a replacement of the Dirichlet boundary condition of pressure to Neumann in line 182 of the project file
https://github.com/ufz/ogs/blob/master/Tests/Data/HydroMechanics/Linear/Unconfined_Compression_late/square_1e2_UC_late.prj#L182

wenqing:
So far, there is no benchmark for the Neumman BC that is applied to the processes using the Taylor-Hood element.

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/2276).

Issue #2279, opened 2018-11-30
DoF construction without checking the shape function order?


norihiro-w:
i wonder whether a dof table is being constructed without checking the shape function order given in the input file. By mistake, I've set the order=1 and used a quadratic order mesh in a input file. then ogs is using the second order in the calculation of shape functions. maybe, this happens only with L-IE... i am not sure.

endJunction:
@norihiro-w Could you try again with the changes in #2280, please? It could be the same bug, if not, I'll take a closer look.

norihiro-w:
no, it didn't work. it seems this is caused by LocalDataInitializer or related codes in LIE. i will check it later

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/2279).

PR #2280, opened 2018-11-30
Fix DOF table for lower order variables

endJunction:
Fixes https://github.com/ufz/ogs/issues/2276

Maybe also related to the https://github.com/ufz/ogs/issues/2279.

### DIFF
--- a/NumLib/DOF/LocalToGlobalIndexMap.cpp
+++ b/NumLib/DOF/LocalToGlobalIndexMap.cpp
@@ -94,7 +94,13 @@ void LocalToGlobalIndexMap::findGlobalIndices(
                 continue;
             MeshLib::Location l(
                 mesh_id, MeshLib::MeshItemType::Node, (*n)->getID());
-            indices.push_back(_mesh_component_map.getGlobalIndex(l, comp_id));
+            auto const global_index =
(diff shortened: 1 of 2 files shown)


NumLib/DOF/LocalToGlobalIndexMap.cpp
```diff
@@ -94,7 +94,13 @@ void LocalToGlobalIndexMap::findGlobalIndices(
                 continue;
             MeshLib::Location l(
                 mesh_id, MeshLib::MeshItemType::Node, (*n)->getID());
-            indices.push_back(_mesh_component_map.getGlobalIndex(l, comp_id));
+            auto const global_index =
+                _mesh_component_map.getGlobalIndex(l, comp_id);
+            if (global_index == std::numeric_limits<GlobalIndexType>::max())
```
TomFischer: Just out of curiosity: In which cases this will happen?

endJunction: The mesh component map returns -1 if there is no DOF associated with the given location and the component. In case of a variable (like pressure) defined only in the base nodes and given quadratic mesh elements, the higher order node locations don't have that variable associated there.

PR #2297, opened 2018-12-13
Subdomain deactivation within time intervals

wenqing:
This PR presents an important feature of subdomain deactivation within time intervals. As a benefit of the feature, the modelling of time dependent excavation is possible.

Based on PR #2244, this PR extends the subdomain deactivation as a general approach for all processes and removes the linear solver restriction. All comments and suggestions to PR #2244 were considered. The implementation includes:
1. subdomain deactivation data are set as members of `ProcessVariable`.
2. nodes in the deactivated subdomains,  which are excluded those on the borders among deactivated and activated subdomains, are assigned with zero Dirichlet BC in order to guarantee that the system of equations is not singular. For this, a constant zero Parameter is created. 
3.  a boolean vector (belong to  `ProcessVariable`) is used to record the element status of   deactivation. The vector is only initialized once for each time interval.
4. two new members, `executeSelectedMemberDereferenced` and `executeMemberOnDereferenced`,
were added to `struct SerialExecutor` to use the boolean vector of  `ProcessVariable` to skip the elements that are deactivated.
5. In all process classes,   `executeMemberDereferenced` and `executeOnDereferenced`  are replaced with `executeSelectedMemberDereferenced` and `executeMemberOnDereferenced`, respectively.

So far, only  SmallDeformation gets a benchmark for this feature in this PR. The benchmark is the same as that in PR #2244, a comparison of  the results of the present benchmark (left) and disc_with_hole (right):
47353534-ca3d2280-d6bc-11e8-90fb-5838b9e42af8

 Benchmarks for other processes with this feature will be presented later on.


ProcessLib/BoundaryCondition/DirichletBoundaryConditionWithinTimeInterval.cpp
```diff
@@ -34,20 +34,44 @@ DirichletBoundaryConditionWithinTimeInterval::
         int const variable_id, int const component_id)
     : _parameter(parameter),
       _bc_mesh(bc_mesh),
+      _nodes_in_bc_mesh(bc_mesh.getNodes()),
```
endJunction: For discussion: Why is this needed.

ProcessLib/DeactivatedSubdomain.cpp
```diff
@@ -0,0 +1,170 @@
...
+    //! \ogs_file_param{prj__process_variables__process_variable__deactivated_subdomains__deactivated_subdomain__time_interval}
+    config.peekConfigParameter<std::string>("time_interval");
+    auto time_interval = BaseLib::createTimeInterval(config);
+
+    std::vector<int> deactivated_subdomain_material_ids;
+    deactivated_subdomain_material_ids =
+        //! \ogs_file_param{prj__process_variables__process_variable__deactivated_subdomains__deactivated_subdomain__material_ids}
+        config.getConfigParameter<std::vector<int>>("material_ids",
+                                                    std::vector<int>{});
+    if (!deactivated_subdomain_material_ids.empty())
```
endJunction: Invert the if condition => reduced indentation

Applications/ApplicationsLib/ProjectData.cpp
```diff
@@ -24,6 +24,8 @@
 
 #include "BaseLib/Algorithm.h"
 #include "BaseLib/FileTools.h"
+// include due to the forward declaration in ProcessVariable.h
+#include "BaseLib/TimeInterval.h"
```
endJunction: This looks to me to be at wrong place. It should be in `DeactivatedSubdomain.h` because of the unique_ptr used there; see https://stackoverflow.com/a/6089065

ProcessLib/DeactivatedSubdomain.cpp
wenqing: done.

Applications/ApplicationsLib/ProjectData.cpp
wenqing: took it.

BaseLib/TimeInterval.h
```diff
@@ -30,6 +30,12 @@ class TimeInterval final
     {
     }
 
+    explicit TimeInterval(const TimeInterval& time_inverval)
```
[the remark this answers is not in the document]
wenqing: took `= default`. Why need the assignment operator?

ProcessLib/ProcessVariable.h
```diff
@@ -94,6 +112,17 @@ class ProcessVariable
     ///
     /// \sa MeshLib::CellRule MeshLib::FaceRule MeshLib::EdgeRule.
     unsigned _shapefunction_order;
+
+    std::vector<std::unique_ptr<DeactivatedSubdomain const>>
+        _deactivated_subdomains;
+
+    mutable std::vector<bool> _element_deactivation_flags;
```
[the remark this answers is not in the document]
wenqing: Since it is a simple change, I will make it after you finish reviewing the present commits.

wenqing:
@bilke  What is the following error in git-check?
```
[MSVC] -> found 428 issues (skipped 46 duplicates)
[MSVC] [ERROR] Can't resolve absolute paths for 417 files:
[MSVC] [ERROR] - ../ThirdParty/metis/GKlib/util.c
...
```

BaseLib/TimeInterval.h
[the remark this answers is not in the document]
wenqing: added a default assignment operator.

ProcessLib/BoundaryCondition/DirichletBoundaryConditionWithinTimeInterval.cpp
wenqing: As explained.

ProcessLib/ProcessVariable.h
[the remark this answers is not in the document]
wenqing: Changed for a test (see commit f6afbe4).  A little acceleration in assembly loop but memory increases 64 times (on 64 bit machine) with the replacement of a boolean vector with a  std::size_t vector. See the efficiency comparison below.

NumLib/Assembler/SerialExecutor.h
```diff
@@ -37,16 +36,16 @@ struct SerialExecutor
     /// \tparam C    input container type.
     /// \tparam Args types additional arguments passed to \c f.
     ///
-    /// \param f    a function that accepts a a reference to container's elements,
+    /// \param f    a function that accepts a a reference to container's
```
endJunction: double "a" before reference.

wenqing:
### Comparison of different members of struct SerialExecutor

**Test cases**
1. original one, `SerialExecutor::executeMemberDereferenced`.
2. new member, `SerialExecutor::executeSelectedMemberOnDereferenced`, which uses `std::vector<bool> container_selector` to distinguish which container members have to been skipped. The member contains
      ```
         assert(container.size() == container_selector.size());
         for (std::size_t i = 0; i < container.size(); i++)
         {
            if (container_selector[i])
                continue;
            (object.*method)(i, *container[i], std::forward<Args>(args)...);
         }
      ```
      For the test, all entries of container_selector are set to false.
3. The same new member as 2. However,  the argument of std::vector<bool> container_selector was replaced with  `std::vector<std::size_t> const& active_container_ids`, which takes `ProcessVariable::_ids_of_active_elements` in  computations (the last commit: f6afbe4). The content of `executeSelectedMemberOnDereferenced` was changed to
     ```
        for (std::size_t i = 0; i < active_container_ids.size(); i++)
        {
            (object.*method)(i, *container[active_container_ids[i]],
                             std::forward<Args>(args)...);
        }
      ```

    which drops the if-condition inside the loop that exists in the implementation in 2.
    For the tests, active_container_ids is set to contain all element IDs.
 
**Benchmark**
The assembly part of Mechanics/Linear/square_1e5.prj.

**Run time comparison**
The comparison uses the release mode compiled exactable files.
| Case 1  | Case 2   | Case 3   |   
|---|---|---|
| 0.383908 s | 0.382762 s | 0.382998 s |

One can see that there is no much differences in the runtime of the three cases.

**Additional comparison**
The tests in the above comparison  do not have deactivated subdomains. In this comparison, the benchmark for this PR, which contains deactivated elements, is used. Therefore only case 2 and case 3 are compared. The run times of the entire computations by Case 2 and Case 3 are:
| Case 2  | Case 3   |   
|---|---|
| 0.48 s | 0.37 s |
Relatively the computation time with Case 3 decreases `(0.48-0.37)/0.48=0.22916666666666666` due to the condensed assembly loop in Case 3. The cost is that the memory for the element deactivation increases 2768 * 64 times of that by Case 2 (2768 * 1 bit of boolean) , where 2768 is the number of the elements in the benchmark.

**Conclusions**
1 If there is not any deactivated subdomain in computation, the changes of SerialExecutor in this PR does not influence the computational efficiency distinctively.
2. If there  are  deactivated  subdomains in computation, Case 3 shows a significant run time reduction compared to that by Case 2 but with a little bit memory usage increase.

NumLib/Assembler/SerialExecutor.h
wenqing: removed by amending the last commit.

ProcessLib/DeactivatedSubdomain.cpp
```diff
@@ -0,0 +1,164 @@
...
+        std::vector<MeshLib::Element*> deactivated_elements;
+        std::vector<MeshLib::Node*> deactivated_nodes;
+
+        // temporary vector to enhance node searching.
+        std::vector<bool> deactivation_flag_of_nodes(mesh.getNumberOfNodes(),
+                                                     false);
+
+        for (std::size_t i = 0; i < mesh.getNumberOfElements(); i++)
+        {
+            if (ids != (*material_ids)[i])
```
endJunction: Please add an `OGS_FATAL` if the `material_ids` are needed but not available. The pointer might be null.

ProcessLib/DeactivatedSubdomain.cpp
```diff
@@ -0,0 +1,164 @@
...
+                continue;
+
+            auto* element = mesh.getElement(i);
+            deactivated_elements.push_back(
```
endJunction: Would you please have a look if `deactivated_elements` could be of `std::vector<MeshLib::Element const*>` type?
Same question to the `deactivated_nodes`.

ProcessLib/LIE/HydroMechanics/HydroMechanicsProcess.cpp
```diff
@@ -568,13 +574,17 @@ void HydroMechanicsProcess<GlobalDim>::assembleWithJacobianConcreteProcess(
 {
     DBUG("AssembleWithJacobian HydroMechanicsProcess.");
 
+    const int process_id =
+        _use_monolithic_scheme ? 0 : _coupled_solutions->process_id;
```
endJunction: For another PR; I'd suggest to extract a function `currentProcessId()` into the base Process class.

ProcessLib/LIE/HydroMechanics/HydroMechanicsProcess.cpp
```diff
@@ -568,13 +574,17 @@ void HydroMechanicsProcess<GlobalDim>::assembleWithJacobianConcreteProcess(
...
+    ProcessLib::ProcessVariable const& pv = getProcessVariables(process_id)[0];
```
endJunction: For another PR; Maybe also a `processVariable(int)` function too with appropriate safety checks.

ProcessLib/ProcessVariable.cpp
```diff
@@ -202,11 +219,99 @@ ProcessVariable::createBoundaryConditions(
...
+{
+    auto& parameter = findParameter<double>(
+        DeactivatedSubdomain::name_of_paramater_of_zero, parameters, 1);
+
+    for (auto const& deactivated_subdomain : _deactivated_subdomains)
+    {
+        auto const& deactivated_sudomain_meshes =
+            (*deactivated_subdomain).deactivated_sudomain_meshes;
+        for (auto const& deactivated_sudomain_mesh :
+             deactivated_sudomain_meshes)
```
endJunction: typo; subdomain

ProcessLib/ProcessVariable.cpp
```diff
@@ -202,11 +219,99 @@ ProcessVariable::createBoundaryConditions(
...
+        {
+            for (int component_id = 0;
+                 component_id < dof_table.getNumberOfComponents();
+                 component_id++)
+            {
+                // Copy the time interval.
+                std::unique_ptr<BaseLib::TimeInterval> time_interval =
+                    std::make_unique<BaseLib::TimeInterval>(
+                        (*(*deactivated_subdomain).time_interval));
```
endJunction: I'd think that `(*(*deactivated_subdomain).time_interval)` is equivalent to `*deactivated_subdomain->time_interval`.

ProcessLib/ProcessVariable.cpp
```diff
@@ -202,11 +219,99 @@ ProcessVariable::createBoundaryConditions(
...
+                    continue;
+                }
+#endif  // USE_PETSC
+                bcs.push_back(std::move(bc));
+            }
+        }
+    }
+}
+
+void ProcessVariable::checkElementDeactivation(double const time)
```
endJunction: It was confusing me above in the for-loop where this check is called; The function does not only check but update some internal variables. Then it's better to call it `updateDeactivatedSubdomains` or similar.

ProcessLib/ProcessVariable.cpp
```diff
@@ -9,14 +9,23 @@
 
 #include "ProcessVariable.h"
 
+#include <iostream>
```
TomFischer: Is iostream needed?

ProcessLib/DeactivatedSubdomain.cpp
wenqing: Moved the checking of material_ids form the ctor of Process to this function.

ProcessLib/DeactivatedSubdomain.cpp
wenqing: Tried but have to keep as they are. Reasons:

1) `std::vector<MeshLib::Elemen *>`:  MeshLib::createMeshFromElementSelection needs this type  
2) `std::vector<MeshLib::Node*>`: bc_mesh.getNodes() returns this type of variable, which is needed to initialize  `DeactivatedSubdomain:: std::vector<MeshLib::Node*> const inactive_nodes`

If const is added, several more files have to to changed as well. This can be done in another PR.

ProcessLib/LIE/HydroMechanics/HydroMechanicsProcess.cpp
wenqing: TODO once this PR is merged.

ProcessLib/LIE/HydroMechanics/HydroMechanicsProcess.cpp
wenqing: TODO once this PR is merged too.

ProcessLib/ProcessVariable.cpp
wenqing: corrected.

ProcessLib/ProcessVariable.cpp
wenqing: Changed.

ProcessLib/ProcessVariable.cpp
wenqing: Changed.

wenqing:
@endJunction  This feature does not work under PETSc so far. Added TODO in the source code.

@TomFischer Made the replacement in the remaining member of PhaseField, and all associated members of SmallDeformationNonlocal. For the  mentioned files about source terms and boundary conditions, the replacement is skipped because the nodes in the deactivated nodes are assigned with zero Dirichlet BC.

ProcessLib/ProcessVariable.cpp
wenqing: Changed the remaining stuffs by amending commit b4200a7.

ProcessLib/ProcessVariable.cpp
wenqing: removed.

PR #2301, opened 2018-12-19
[Parameter Doc] Fixed some errors

wenqing:
as titled


ProcessLib/CreateProcessData.cpp
```diff
@@ -81,7 +81,7 @@ std::vector<std::unique_ptr<ProcessData>> createPerProcessData(
             //! \ogs_file_param{prj__time_loop__processes__process__convergence_criterion}
             pcs_config.getConfigSubtree("convergence_criterion"));
 
-        //! \ogs_file_special
+        //! \ogs_file_param{prj__time_loop__processes__process__output}
         auto output = pcs_config.getConfigSubtreeOptional("output");
```
wenqing: It is better to remove this config.

endJunction: We keep it for a year or so until the external project files have been converted.

PR #1955, opened 2017-09-28
Gpml

Scinopode:
A new method to handle material properties. It creates a single object that holds all the properties of a medium, its phases and their components. This object is a private attribute of ProjectData and can be passed forward to any process. The structure of this material object is process independent, thus it should be adequate to satisfy all processes.

PR #2303, opened 2018-12-21
[MatL] Multi-phase/-component library.

endJunction:
This is part of #1955 which is used in multi component transport (#2304). Mainly infrastructure, since only constant properties were used so far in the ComponentTransportProcess.


MaterialLib/MPL/Enums.h
```diff
@@ -0,0 +1,311 @@
...
+
+/// Very simple tensor data type for holding
+/// tensor components.
+using Tensor = std::array<double, 9>;
+
+/// Variables is simply a list of all commonly used variables that are used to
+/// determine the size of the VariableArray. If the variable of your choice is
+/// missing, simply add it somewhere at the list, but above the last entry
+/// 'numberOfVariables'
+enum Variables : std::size_t
```
TomFischer: Why not 'enum class'?

MaterialLib/MPL/Enums.h
```diff
@@ -0,0 +1,311 @@
...
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#pragma once
+
+#include <boost/algorithm/string/predicate.hpp>
+#include <boost/variant.hpp>
```
TomFischer: Can we already use `std::variant`?

MaterialLib/MPL/Enums.h
```diff
@@ -0,0 +1,311 @@
...
+}
+
+/// PropertyType is an enumerator list of all known properties of a substance.
+/// This includes all properties on all scales (i.e. component, phase, and
+/// medium scales). It is used as an indexer for the PropertyArray of the
+/// materials. If a necessary property is not in the list, simply add a new one
+/// in alphabetical order (of course, except for the last entry
+/// 'number_of_property_enums'). Please note that any of these entries must also
+/// appear in below convert functions.
+enum PropertyType : std::size_t
```
TomFischer: Why not `enum class`?

MaterialLib/MPL/Enums.h
```diff
@@ -0,0 +1,311 @@
...
+    {
+        return transveral_dispersivity;
+    }
+    if (boost::iequals(inString, "viscosity"))
+    {
+        return viscosity;
+    }
+
+    OGS_FATAL(
+        "The property name \"%s\" does not correspond to any known "
```
TomFischer: To be consistent with other OGS-6 code please use `'` instead `\"`.

MaterialLib/MPL/MaterialSpatialDistributionMap.h
```diff
@@ -0,0 +1,43 @@
...
+}  // namespace MeshLib
+
+namespace MaterialPropertyLib
+{
+class Medium;
+
+class MaterialSpatialDistributionMap
+{
+private:
+    std::map<int, std::unique_ptr<Medium>> const& _media;
```
TomFischer: As far as I understand the domain of the map is [0, #number of media]. Maybe a vector can be used here?

endJunction: Note to myself; compare with error/warning message code.

MaterialLib/MPL/Medium.h
```diff
@@ -0,0 +1,80 @@
...
+#include "Phase.h"
+#include "Property.h"
+
+namespace MaterialPropertyLib
+{
+/// This class is for material objects on the Medium scale.
+///
+/// A Medium consists of an arbitrarily long vector of phases and an array of
+/// properties.
+class Medium
```
TomFischer: Can the class Medium be final or is it planned in another PR to derive from Medium?

MaterialLib/MPL/Medium.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+#include "BaseLib/Algorithm.h"
+#include "Properties/Properties.h"
+
+#include "Component.h"
+#include "Phase.h"
+
+namespace MaterialPropertyLib
+{
+Medium::Medium(BaseLib::ConfigTree const& config)
```
wenqing: By 
1) making createProperties, createPhases be local functions only in this file, 
2) letting createProperties return PropertyArray, and let createPhases return std::vector<std::unique_ptr<Phase>> 
 
 ` _phases` and  `_properties` can be initialized in the initialization list.

MaterialLib/MPL/Component.h
```diff
@@ -0,0 +1,62 @@
...
+namespace MaterialPropertyLib
+{
+/// \brief This class defines components (substances).
+/// \details The Component class is a base class used for not further specified
+/// components. Components are specified by the property 'name'. For specified
+/// components we derive special classes from this class (for clarity they are
+/// located in the 'components' subfolder).
+class Component
+{
+protected:
```
wenqing: It is better to place protected or private section after public.

MaterialLib/MPL/Phase.cpp
```diff
@@ -0,0 +1,134 @@
...
+                  phase_name) == allowed_phase_names.end())
+    {
+        ERR("Phase name should be one of:");
+        for (auto const name : allowed_phase_names)
+        {
+            ERR(name.c_str());
+        }
+        OGS_FATAL("Wrong phase name '%s' given.", phase_name.c_str());
+    }
+    _properties[name] = std::make_unique<Constant>(phase_name);
```
wenqing: The same question: can _properties be initialized in the initialization list of this ctor?

MaterialLib/MPL/Phase.h
```diff
@@ -0,0 +1,75 @@
...
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+#pragma once
+
+#include <string>
+#include <vector>
+
+#include "BaseLib/ConfigTree.h"
```
wenqing: Use the forward declaration.

MaterialLib/MPL/Phase.h
```diff
@@ -0,0 +1,75 @@
...
+
+#include "Component.h"
```
wenqing: It is possible to use forward declaration for Property too.

MaterialLib/MPL/Properties/Constant.cpp
```diff
@@ -0,0 +1,25 @@
+/**
+ * \file
```
wenqing: remove it or leave a blank line after it.

MaterialLib/MPL/Properties/Properties.h
```diff
@@ -0,0 +1,15 @@
...
+ * \copyright
+ * Copyright (c) 2012-2018, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+#pragma once
+
+#include "Constant.h"
```
wenqing: This file only contains one include.

MaterialLib/MPL/Property.h
```diff
@@ -0,0 +1,192 @@
...
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+#pragma once
+
+#include <array>
+#include <boost/variant.hpp>
+#include <string>
+
+#include "BaseLib/ConfigTree.h"
```
wenqing: Forward declaration.

Tests/MaterialLib/TestMPL.h
```diff
@@ -0,0 +1,21 @@
+/**
+ * \file
```
wenqing: Leave a blank line after it.

Tests/MaterialLib/TestMPLParseMaterial.cpp
```diff
@@ -0,0 +1,297 @@
...
+
+#include "TestMPL.h"
+#include "Tests/TestTools.h"
+
+#include "MaterialLib/MPL/mpMedium.h"
+
+namespace MPL = MaterialPropertyLib;
+
+//----------------------------------------------------------------------------
+// MPL::Medium createTestMaterial(std::string const& xml)
```
wenqing: Are these lines still needed?

MaterialLib/MPL/Component.cpp
```diff
@@ -0,0 +1,69 @@
+/**
+ * \file
+ * \author Norbert Grunwald
+ * \date   Sep 7, 2017
+ *
+ * \copyright
+ * Copyright (c) 2012-2018, OpenGeoSys Community (http://www.opengeosys.org)
```
wenqing: 2018--> 2019

wenqing: The same for the other new files in this PR.

MaterialLib/MPL/Property.h
```diff
@@ -0,0 +1,192 @@
...
+{
+    nScalar,
+    nPair,
+    nVector,
+    nSymmTensor,
+    nTensor
+};
+
+using PropertyDataType =
+    boost::variant<double, Pair, Vector, SymmTensor, Tensor, std::string>;
```
TomFischer: I think std::variant can be used.

MaterialLib/MPL/MaterialSpatialDistributionMap.cpp
```diff
@@ -0,0 +1,24 @@
+/**
+ * \file
```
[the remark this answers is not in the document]
endJunction: I see, but this is a little trickier: `\file` should be there, otherwise doxygen puts documentation from different files together like [here](https://doxygen.opengeosys.org/d8/df9/GroundwaterFlowFEM_8h.html), see namespace section. The `\file` is missing in https://github.com/ufz/ogs/blob/master/ProcessLib/TwoPhaseFlowWithPP/TwoPhaseFlowWithPPLocalAssembler-impl.h

Filename is optional and is the current file name, whatever it is. Adding explicitly a file name requires also an update of the `\file xxx.h` to `\file yyy.h` when moving files.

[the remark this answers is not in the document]
endJunction: You are right about the blank line, thanks! Will change.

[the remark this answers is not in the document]
endJunction: I tried four options:
1. no `\file`
2. `\file` followed by `\author` on new line.
3. `\file` followed by empty line.
4. `\file` followed by XXX on new line.
5. `\file` followed by XXX on same line.

Number 1 and 5 do not produce correct output. So, it seems, that new line is sufficient. Will keep the original version (2).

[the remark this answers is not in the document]
wenqing: or \file followed by the file name.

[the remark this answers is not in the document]
endJunction: Adding explicitly a file name requires also an update of the \file xxx.h to \file yyy.h when moving files.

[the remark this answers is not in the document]
wenqing: I see.

MaterialLib/MPL/Enums.h
endJunction: Unfortunately not until gcc-7.1. And also the macOS's clang in version 10 does not support it fully yet. But (fasten your belts!!!) the latest visual studio would work!

Also tried new syntax for "structured binding declarations" instead of `std::tie`
```c++
auto [a, b, c] = function_returning_a_tuple_of_tree();
```
working only since gcc-7

MaterialLib/MPL/MaterialSpatialDistributionMap.h
endJunction: @TomFischer the media ids follow numbering from MaterialIDs and can be rather arbitrary, especially negative. We might have a look on it later, if it looks like a bottleneck.

MaterialLib/MPL/Enums.h
endJunction: The implicit conversion to `int` (I changed the underlying type) is quite handy.

Applications/ApplicationsLib/ProjectData.cpp
```diff
@@ -303,6 +307,49 @@ void ProjectData::parseParameters(BaseLib::ConfigTree const& parameters_config)
...
+            std::make_unique<MaterialPropertyLib::Medium>(medium_config);
+    }
+
+    if (_media.empty())
+        OGS_FATAL("No entity is found inside <media>.");
+
+    if (_media.rbegin()->first != static_cast<int>(_media.size()) - 1)
+           OGS_FATAL(
+               "The ids in the porous media definitions in the project file have "
+               "to be sequential, starting with the id zero.");
```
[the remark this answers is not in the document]
endJunction: I dropped the whole if-condition. The requirement of media to be numbered from 0 to n sequentially, does not match with the freedom of the MaterialIDs, which can be any `int`, especially negative.

MaterialLib/MPL/Component.cpp
```diff
@@ -0,0 +1,69 @@
...
+{
+    return *_properties[p];
+}
+
+std::unique_ptr<Component> newComponent(std::string const& component_name,
+                                        bool& isCustomComponent)
+{
+    // If a name is given, it must conform with one of the derived component
+    // names in the following list:
+    if (boost::iequals(component_name, "water"))
```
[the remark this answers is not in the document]
endJunction: Norbert's argument is that it is difficult to explain people, why "Water" is not good enough for the specification of water. Same goes with "CO2" etc.

MaterialLib/MPL/Component.h
```diff
@@ -0,0 +1,62 @@
...
+ * \copyright
+ * Copyright (c) 2012-2018, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+#pragma once
+
+#include "BaseLib/ConfigTree.h"
```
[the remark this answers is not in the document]
endJunction: Whole creation of media, phases, components, and properties is now moved into corresponding CreateX.h/cpp files.

MaterialLib/MPL/Component.h
```diff
@@ -0,0 +1,62 @@
...
+/// \brief This class defines components (substances).
+/// \details The Component class is a base class used for not further specified
+/// components. Components are specified by the property 'name'. For specified
+/// components we derive special classes from this class (for clarity they are
+/// located in the 'components' subfolder).
+class Component
+{
+protected:
+    /// The property array of the component.
+    PropertyArray _properties;
```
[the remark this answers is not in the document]
endJunction: No, some of the properties are set after the creation. It would require some more coding to be able to make `_properties` `const`.

MaterialLib/MPL/Component.h
```diff
@@ -0,0 +1,62 @@
...
+ * Copyright (c) 2012-2018, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+#pragma once
+
+#include "BaseLib/ConfigTree.h"
+#include "Property.h"
```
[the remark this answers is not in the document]
endJunction: Unfortunately not, because the implementation of `value<T>()` etc. requires access to Property members.
I split the Enums.h in PropertyType.h and VariableType.h though.

MaterialLib/MPL/Components/Components.h
```diff
@@ -0,0 +1,15 @@
...
+ * \copyright
+ * Copyright (c) 2012-2018, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+#pragma once
+
+#include "Water.h"
```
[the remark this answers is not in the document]
endJunction: In the full implementation there are many more components here. This file is the included once in the process.

MaterialLib/MPL/Medium.h
```diff
@@ -0,0 +1,80 @@
...
+{
+private:
+    /// The vector that holds the phases.
+    std::vector<std::unique_ptr<Phase>> _phases;
```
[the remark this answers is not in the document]
endJunction: _phases yes, _properties not, as explained above.

MaterialLib/MPL/Medium.h
```diff
@@ -0,0 +1,80 @@
...
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+#pragma once
+
+#include <memory>
+#include <vector>
+#include "BaseLib/ConfigTree.h"
+
+#include "Phase.h"
```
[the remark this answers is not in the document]
endJunction: Only Property. Phase is used in unique_ptr, and compiler asks for `sizeof(Phase)`.

MaterialLib/MPL/Medium.cpp
endJunction: Followed your suggestion, though only `_phases` are initialized, since `_properties` are set in the ctor body.

MaterialLib/MPL/Properties/Properties.h
endJunction: Same as for components, the full implementation has many, and the properties include is specified once in the process.
Other properties (and also components) which are not tested in the follow-up PR, are excluded.

MaterialLib/MPL/PropertyType.h
```diff
@@ -0,0 +1,274 @@
...
+    viscosity,
+    number_of_properties
+};
+
+/// This function converts a string (e.g. a string from the configuration-tree)
+/// into one of the entries of the PropertyType enumerator. To avoid confusion,
+/// I suggest that the syntax of the properties in the input-files (i.e. the
+/// strings) have to be identical to the syntax of the entries in the
+/// enumerator.
+inline PropertyType convertStringToProperty(std::string const& inString)
```
wenqing: If we want to the elements of  PropertyType, we have to to change strings in  `convertStringToProperty` and `property_enum_to_string` as well.

It might be good to 
1) create a enum to std::string convertor in BaseLib (see https://stackoverflow.com/questions/28828957/enum-to-string-in-modern-c11-c14-c17-and-future-c20).
2) use the  enum to std::string convertor to initialize `std::array<std::string, PropertyType::number_of_properties>
    property_enum_to_string`
3) create a std::map<std::string, PropertyType> by using of std::string convertor to replace `convertStringToProperty` 

With that,   we only need to change the elements of PropertyType if such changed are needed.

The current implementation is also OK because  there might be less change in `PropertyType`.

wenqing: 1) is not good. One still has to use the enum name to get its string type. It seems that each name of enum element must  appears at least two times in the source code. Therefore, please go forward with the present implementation.

MaterialLib/MPL/Component.h
```diff
@@ -0,0 +1,87 @@
...
+    /// Default constructor of Component. This constructor is used
+    /// when the component is not specified via the 'name'-tag.
+    Component();
+
+    /// Constructor for a custom component
+    Component(std::string const& component_name,
+              std::unique_ptr<PropertyArray>&& properties);
+    virtual ~Component() = default;
+
+    /// A get-function for retrieving a cartain property.
```
TomFischer: cartain -> certain

MaterialLib/MPL/Component.h
```diff
@@ -0,0 +1,87 @@
...
+    Property const& property(PropertyType const& /*p*/) const;
+
+    template <typename T>
+    T value(PropertyType const p) const
+    {
+        return property(p).template value<T>();
+    }
+
+    template <typename T>
```
TomFischer: Is my understanding correct: the variable_array is used to pass information for property value calculation?

MaterialLib/MPL/CreatePhase.cpp
```diff
@@ -0,0 +1,112 @@
...
+    if (std::none_of(allowed_phase_types.begin(),
+                     allowed_phase_types.end(),
+                     [&phase_type](std::string const& type) {
+                         return phase_type == type;
+                     }))
+    {
+        ERR("Phase type should be one of:");
+        for (auto const type : allowed_phase_types)
+        {
+            ERR(type.c_str());
```
TomFischer: Is this valid? I thought it must be ```ERR("%s", type.c_str())```.

MaterialLib/MPL/CreateProperty.cpp
```diff
@@ -0,0 +1,160 @@
...
+    // corresponding value is needed.
+    if (property_type == "Constant")
+    {
+        //! \ogs_file_param{properties__property__Constant__value}
+        auto const sValue = config.getConfigParameter<std::string>("value");
+
+        std::stringstream issValue(sValue);
+
+        std::vector<double> values;
+        double dummy(0);
```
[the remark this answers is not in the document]
TomFischer: `dummy` cann't be const since it is used to read from the stringstream `issValues`. So maybe 'dummy' isn't so dummy ;-)

MaterialLib/MPL/Property.h
```diff
@@ -0,0 +1,116 @@
...
+};
+
+/// This method returns the 0-based index of the variant data types. Can be
+/// enhanced by using enums.
+inline std::size_t getType(Property const& p)
+{
+    return p.value().which();
+}
+
+inline void overwriteExistingProperties(PropertyArray& properties,
```
TomFischer: Since in line 111 `new_properties` entries are moved, the `new_properties` aren't valid anymore, isn't it? Maybe the name of the function is a bit misleading, but I haven't a good suggestion.

MaterialLib/MPL/VariableType.h
```diff
@@ -0,0 +1,66 @@
...
+/// a pair of values.
+using Pair = std::array<double, 2>;
+
+/// Very simple vector data type for holding
+/// vector components.
+using Vector = std::array<double, 3>;
+
+/// Simple symmetric tensor data type for holding
+/// xx, yy, zz, xy, xz, yz tensor components.
+using SymmTensor = std::array<double, 9>;
```
TomFischer: The documentation says there are 6 entries, the array has size 9. Is this correct?

Tests/MaterialLib/TestMPLParseMaterial.cpp
```diff
@@ -0,0 +1,274 @@
...
+};
+
+/// Short method creating a number of blanks used for indentation.
+std::string indent(std::size_t ind)
+{
+    return std::string(ind, ' ');
+}
+
+/// This method creates an XML-snippet of a sigle property.
+std::string makeProperty(std::size_t ind, std::size_t index,
```
TomFischer: Why not `... makeProperty(std::size_t const ind, std::size_t const index, std::string const& property)`?

Tests/MaterialLib/TestMPLParseMaterial.cpp
```diff
@@ -0,0 +1,274 @@
...
+                         std::string property)
+{
+    std::stringstream sProperty;
+
+    if (property == "")
```
TomFischer: Maybe `property.empty()` would be more readable.

Tests/MaterialLib/TestMPLParseMaterial.cpp
```diff
@@ -0,0 +1,274 @@
...
+        return "";
+
+    sProperty << indent(ind) << "<property>\n";
+    sProperty << indent(ind + 2) << "<name>"
+              << MPL::property_enum_to_string[index] << "</name>\n";
+
+    // What follows is a C-style string-to-double-conversion...
+    // seems so easy compared to c++-standard; Feel free to
+    // change it to latest standard..
+    if (double value = atof(property.c_str()))
```
TomFischer: `std::atof` please.

Tests/MaterialLib/TestMPLParseMaterial.cpp
```diff
@@ -0,0 +1,274 @@
...
+    for (auto c : components)
+    {
+        std::string component = makeComponent(c);
+        sComponents << component;
+    }
+    sComponents << indent(10) << "</components>\n";
+    return sComponents.str();
+}
+
+/// This method creates an XML-snippet of a sigle material
```
TomFischer: sigle -> single

Tests/MaterialLib/TestMPLParseMaterial.cpp
```diff
@@ -0,0 +1,274 @@
...
+    phase << indent(10) << "<type>" << p.property[MPL::name] << "</type>\n";
+    phase << phaseComponents;
+    phase << phaseProperties;
+    phase << indent(8) << "</phase>\n";
+    return phase.str();
+}
+
+/// This method creates an XML-snippet of the material
+/// phases.
+std::string makePhases(std::vector<Phase> phases)
```
TomFischer: `std::vector<Phase> const& phases`?

Tests/MaterialLib/TestMPLParseMaterial.cpp
```diff
@@ -0,0 +1,274 @@
...
+        sPhases << phase;
+    }
+    sPhases << indent(6) << "</phases>\n";
+    return sPhases.str();
+}
+
+/// This method creates the entire XML-tree structure from the string-
+/// based medium specification object. I know, indentation is not
+/// necessary for this, but my OCD kicked in...
+std::string makeMedium(Medium m)
```
TomFischer: Medium const& m

Tests/MaterialLib/TestMPLParseMaterial.cpp
```diff
@@ -0,0 +1,274 @@
...
+
+/// A method used to obtain the name of a medium, phase, or component of a
+/// material or of a specifier and to store them in two vectors for later
+/// comparison.
+void getNames(std::string const& observed_name, std::string const& expectation,
+              std::string const& defaultName, std::vector<std::string>* obs,
+              std::vector<std::string>* exp)
+{
+    obs->push_back(observed_name);
+    if (expectation == "")
```
TomFischer: IMHO `expectation.empty()` would be more readable.

MaterialLib/MPL/Phase.h
```diff
@@ -0,0 +1,66 @@
...
+
+    /// A get-function for retrieving the number of components in this phase.
+    std::size_t numberOfComponents() const;
+
+    std::string name() const;
+
+private:
+    std::vector<std::unique_ptr<Component>> const _components;
+
+    /// Here, all for all properties listed in the Properties enumerator are
```
TomFischer: ~~all for~~

MaterialLib/MPL/Component.h
endJunction: Yes.

MaterialLib/MPL/CreatePhase.cpp
endJunction: It's just a char array, with substitution or without.

MaterialLib/MPL/Property.h
endJunction: Good point. Need a name.

MaterialLib/MPL/VariableType.h
endJunction: Good catch! fixing.

MaterialLib/MPL/CreateProperty.cpp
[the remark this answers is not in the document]
endJunction: Didn't see the reading of `std::vector<double>` from the config. ConfigTree supports `getConfigParameter<std::vector<T>>(...)`.

endJunction:
@TomFischer I polished the TestMPLParseMaterial.cpp with all the const, &, typos, etc. being fixed now.

endJunction:
Two open points remain: enum <-> string, and name for overwriteProperties.

endJunction:
When there are no further comments, and the builds pass, this can be merged.
The enum <-> string conversion needs additional discussions first.

wenqing:
:fast_forward:

PR #2307, opened 2019-01-07
[Jenkins] Use new xunit()-syntax.

bilke:


endJunction:
Note to myself: googletest 1.6.0 is not compatible with c++17.

PR #2318, opened 2019-01-14
Iteration number based time stepping

endJunction:
The main algorithm was implemented quite some time already, but there was no driver for it. Here it goes.

I'd suggest commit-wise review.


ProcessLib/RichardsFlow/Tests.cmake
```diff
@@ -46,11 +46,25 @@ AddTest(
...
 )
 
+AddTest(
+    NAME 2D_RichardsFlow_h_us_quad_small_iteration_adaptive_dt
+    PATH Parabolic/Richards
+    EXECUTABLE ogs
+    EXECUTABLE_ARGS RichardsFlow_2d_small_iteration_adaptive_dt.prj
+    TESTER vtkdiff
+    REQUIREMENTS NOT OGS_USE_MPI
+    # No vtkdiff comparison here, because of the different file names for
```
bilke: Then I guess you have to remove `TESTER vtkdiff`-line.

NumLib/TimeStepping/Algorithms/IterationNumberBasedTimeStepping.cpp
```diff
@@ -59,50 +62,61 @@ bool IterationNumberBasedAdaptiveTimeStepping::next(
...
+        }
+        dt = _ts_prev.dt() * multiplier;
     }
 
     // check whether out of the boundary
-    if (dt < _min_ts)
-        dt = _min_ts;
-    else if (dt > _max_ts)
-        dt = _max_ts;
+    if (dt < _min_dt)
```
wenqing: dt = std::max(dt, _min_dt);

NumLib/TimeStepping/Algorithms/IterationNumberBasedTimeStepping.cpp
```diff
@@ -59,50 +62,61 @@ bool IterationNumberBasedAdaptiveTimeStepping::next(
...
+    {
+        dt = _min_dt;
+    }
+    else if (dt > _max_dt)
```
wenqing: dt = std::min(dt, _max_dt);

NumLib/TimeStepping/Algorithms/IterationNumberBasedTimeStepping.cpp
```diff
@@ -19,31 +20,33 @@
...
+    double const max_dt, double const initial_dt,
+    std::vector<int>&& iter_times_vector,
+    std::vector<double>&& multiplier_vector)
+    : TimeStepAlgorithm(t_initial, t_end),
+      _iter_times_vector(std::move(iter_times_vector)),
+      _multiplier_vector(std::move(multiplier_vector)),
+      _min_dt(min_dt),
+      _max_dt(max_dt),
+      _initial_dt(initial_dt),
+      _max_iter(_iter_times_vector.empty() ? 0 : _iter_times_vector.back())
```
TomFischer: Sorry, I'm not so familiar with the time stepping. Could you give a simple and short example what data is stored in  `_iter_times_vector`?

NumLib/TimeStepping/CreateTimeStepper.cpp
```diff
@@ -29,33 +30,31 @@ std::unique_ptr<TimeStepAlgorithm> createTimeStepper(
...
+        return NumLib::createIterationNumberBasedTimeStepping(config);
     }
-
-    return timestepper;
+    OGS_FATAL(
+        "Unknown time stepping type: '%s'. The available types are: "
+        "\n\tSingleStep,"
+        "\n\tFixedTimeStepping,"
+        "\n\tEvolutionaryPIDcontroller,",
+        "\n\t IterationNumberBasedTimeStepping\n",
```
TomFischer: Indentation of IterationNumberBasedTimeStepping is inconsistent with other time stepping schemas.

jbathmann:
I did some tests with the setup attached to this post on this and found some issues.

- The algorithm might fail if nonlinear_solvers/nonlinear_solver/max_iter == max(number_iterations), because last element of list does not seem to be taken into account

- The algorithm fails, if max_iterations (same as above)  > max(number_iterations), because multiplier is set to 1

- I needed some time to figure out, what number_iterations and multiplier tag mean. Maybe some comments on this in the .md files would be really nice. Or did I miss on finding the documentation?

-The fixed output times from the project file are not hit. 

Hope this helps?
[testsetup.zip](https://github.com/ufz/ogs/files/2768393/testsetup.zip)

endJunction:
As @jbathmann mentioned above there are some issues with the repeating of the time steps for certain combinations of `number_of_iterations` vector and the `max_iter` in the non-linear solver; the simulations stops, although it could have succeeded. Increasing the `max_iter` seems to resolve this kind of problems.

Real fix need a major rewrite of the time loop.

Please review from the "[doc] Add IterationNumberBasedTS prj tags docu." commit 'til the end, especially the changes in the `ProcessLib/UncoupledProcessesTimeLoop.cpp`

ProcessLib/UncoupledProcessesTimeLoop.cpp
```diff
@@ -431,7 +425,8 @@ double UncoupledProcessesTimeLoop::computeTimeStepping(
         }
         else
         {
-            if (t < _end_time)
+            if (t < _end_time || std::abs(t - _end_time) <
+                                     std::numeric_limits<double>::epsilon())
```
wenqing: Just find a std function: std::islessequal.

ProcessLib/UncoupledProcessesTimeLoop.cpp
```diff
@@ -451,7 +446,8 @@ double UncoupledProcessesTimeLoop::computeTimeStepping(
...
```
wenqing: It can be replace with  std::islessequal.

ProcessLib/UncoupledProcessesTimeLoop.cpp
endJunction: Thanks, I didn't know that function. But from [cppreference](https://en.cppreference.com/w/cpp/numeric/math/islessequal) it seems that the only difference to `operator<=(x, y)` is that the floating point exceptions are not raised (and we do like exceptions).

wenqing: I see.

PR #2321, opened 2019-01-15
BHE Pipline Network Coupling through Python Interface

HBShaoUFZ:
In this PR, the feature pipeline network is coupled with the HEAT_TRANSPORT_BHE process through a python interface. This feature is mainly developed by Shuang Chen. 

The benchmark case "3D_3BHE_array" is  located at \Tests\Data\Parabolic\T\3D_3BHE_array
The running of the benchmark requires the following configurations: 

- In the cmake configuration, the option"OGS_USE_PYTHON" is turned on; 
- The execution environment is set to the same folder where the prj file is located; 
- In the python environment, the package Numpy, CoolProp and TESPy have to be installed. 

Source code is ready for review. The documentation of the benchmark case will be uploaded in the next days.


ProcessLib/BoundaryCondition/Python/BHEInflowPythonBoundaryCondition.cpp
```diff
@@ -0,0 +1,96 @@
...
+        if (dataframe_node_id[i] == boundary_node_id)
+        {
+            bc_values.values[0] = dataframe_Tin_val[i];
+            break;
+        }
+    }
+}
+
+std::unique_ptr<BHEInflowPythonBoundaryCondition>
+createBHEInflowPythonBoundaryCondition(
```
endJunction: Move in own `CreateBHEInflowPythonBoundaryCondition.cpp`. Same with `.h`.

ProcessLib/BoundaryCondition/Python/PythonBoundaryConditionModule.cpp
```diff
@@ -59,5 +59,4 @@ void pythonBindBoundaryCondition(pybind11::module& m)
              &PythonBoundaryConditionPythonSideInterface::getDirichletBCValue);
     pybc.def("getFlux", &PythonBoundaryConditionPythonSideInterface::getFlux);
 }
-
```
endJunction: Please, don't change unrelated files.

ProcessLib/HeatTransportBHE/CreateHeatTransportBHEProcess.cpp
```diff
@@ -19,6 +19,8 @@
 #include "HeatTransportBHEProcess.h"
 #include "HeatTransportBHEProcessData.h"
 
+#include <pybind11/pybind11.h>
```
endJunction: The process should not depend on python, in my opinion. Why is the dependency necessary and can it be moved in the boundary condition?

ProcessLib/HeatTransportBHE/CreateHeatTransportBHEProcess.cpp
```diff
@@ -184,23 +190,82 @@ std::unique_ptr<Process> createHeatTransportBHEProcess(
...
-                                             density_fluid,
-                                             density_gas,
-                                             std::move(bhes)};
+    //! Python object computing BC values.
+    HeatTransportBHEProcessData* process_data;
 
+    if (if_bhe_network_exist_python_bc == false)
+    {
+        process_data =
+            new HeatTransportBHEProcessData(thermal_conductivity_solid,
```
endJunction: `new` and `delete` are nowadays replaced by `unique_ptr` (or other smart pointers). In the particular case I don't see why a pointer is needed: there is no base class used here.

ProcessLib/HeatTransportBHE/HeatTransportBHEProcess.cpp
```diff
@@ -197,6 +199,44 @@ void HeatTransportBHEProcess::computeSecondaryVariableConcrete(
...
+                std::get<1>(_process_data.py_bc_object->dataframe_network),
+                std::get<2>(_process_data.py_bc_object->dataframe_network));
+        auto const cur_Tin = std::get<2>(tespy_result);
+
+        auto const if_convergence = std::get<1>(tespy_result);
+        if (if_convergence == true)
+            return NumLib::IterationResult::SUCCESS;
+        else
+            for (std::size_t i = 0; i < n_bc_nodes; i++)
+                std::get<1>(_process_data.py_bc_object->dataframe_network)[i] =
```
endJunction: Braces around (even single line) if/else/for branches/loops missing.

ProcessLib/HeatTransportBHE/CreateHeatTransportBHEProcess.cpp
```diff
@@ -184,23 +190,82 @@ std::unique_ptr<Process> createHeatTransportBHEProcess(
...
+                                            heat_capacity_solid,
+                                            heat_capacity_fluid,
+                                            heat_capacity_gas,
+                                            density_solid,
+                                            density_fluid,
+                                            density_gas,
+                                            std::move(bhes),
+                                            if_bhe_network_exist_python_bc);
+    }
+    // creat a pythonBoundaryCondition object
```
endJunction: create

ProcessLib/HeatTransportBHE/HeatTransportBHEProcessData.h
```diff
@@ -82,6 +89,12 @@ struct HeatTransportBHEProcessData
     std::unordered_map<int, int> _map_materialID_to_BHE_ID;
 
     std::vector<BHE::BHETypes> _vec_BHE_property;
+
+    // if bhe network exists python boundary condition
+    bool if_bhe_network_exist_python_bc;
+
+    //! Python object computing BC values.
+    BHEInflowPythonBoundaryConditionPythonSideInterface* py_bc_object;
```
endJunction: No plain pointers, please. This is a memory leak.

Tests/Data/Parabolic/T/3D_3BHE_array/pre/3bhes.py
```diff
@@ -0,0 +1,152 @@
...
+bhe2 = cmp.heat_exchanger_simple('BHE2')
+bhe3 = cmp.heat_exchanger_simple('BHE3')
+
+mg = cmp.merge('merge',num_in = 3)
+
+cons = cmp.heat_exchanger_simple('consumer')
+# two btes subsystems (pipes in parallel flow):
+# 10 pipes for outer ring, 5 pipes inner ring
+#bo = bp('btes outer', 4)
+#bi = bp('btes inner', 2)
```
endJunction: There is quite a bit of commented code in the python modules. Is it needed?

Tests/Data/Parabolic/T/3D_3BHE_array/pre/3bhes.py
```diff
@@ -0,0 +1,152 @@
...
+# cycle closer (from consumer)
+fc_in = cmp.source('from consumer inflow')
+fc_out = cmp.sink('from consumer outflow')
+
+p1 = cmp.pump('pump')
+
+sp = cmp.splitter('splitter',num_out = 3)
+
+#bhe:
+bhe1 = cmp.heat_exchanger_simple('BHE1')#!name must be same as the user setting in inp_network.py
```
endJunction: Don't assume, the name is same as somewhere else, write an assertion.

endJunction:
Main questions for discussion: what is the difference to the existing PythonBoundaryCondition and why is the BHE network boundary condition not derived from the existing implementation?

ProcessLib/HeatTransportBHE/CreateHeatTransportBHEProcess.cpp
HBShaoUFZ: fixed in commit 87cce675

ProcessLib/HeatTransportBHE/HeatTransportBHEProcess.cpp
HBShaoUFZ: fixed in commit 87cce675

ProcessLib/HeatTransportBHE/HeatTransportBHEProcess.cpp
```diff
@@ -197,6 +199,44 @@ void HeatTransportBHEProcess::computeSecondaryVariableConcrete(
         getDOFTable(process_id), t, x, _coupled_solutions);
 }
 
+NumLib::IterationResult HeatTransportBHEProcess::postIterationConcreteProcess(
+    GlobalVector const& x)
+{
+    // if the process use python boundary conditon
+    if (_process_data.if_bhe_network_exist_python_bc == false)
+        return NumLib::IterationResult::SUCCESS;
+    else
```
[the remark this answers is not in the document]
HBShaoUFZ: fixed in commit 87cce675

ProcessLib/HeatTransportBHE/HeatTransportBHEProcess.cpp
```diff
@@ -9,8 +9,8 @@
 
 #include "HeatTransportBHEProcess.h"
 
-#include <iostream>
 #include <cassert>
+#include <iostream>
```
TomFischer: I don't think the iostream include is needed.

ProcessLib/HeatTransportBHE/HeatTransportBHEProcess.cpp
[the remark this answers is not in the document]
endJunction: It looks like there are more if-else cases with returns.

ProcessLib/HeatTransportBHE/CreateHeatTransportBHEProcess.cpp
endJunction: All typos of this type should be fixed.

ProcessLib/HeatTransportBHE/HeatTransportBHEProcess.cpp
```diff
@@ -239,18 +277,37 @@ void HeatTransportBHEProcess::createBHEBoundaryConditionTopBottom(
...
+                    // Top, inflow.
+                    bcs.addBoundaryCondition(
+                        ProcessLib::createBHEInflowDirichletBoundaryCondition(
+                            get_global_bhe_bc_indices(bc_top_node_id,
+                                                      in_out_component_id),
+                            bhe));
+
+                    // Bottom, outflow.
+                    bcs.addBoundaryCondition(
+                        ProcessLib::createBHEBottomDirichletBoundaryCondition(
```
endJunction: Code duplication for the bottom BC is not needed.

Tests/Data/Parabolic/T/3D_3BHE_array/pre/3bhes.py
ShuangChen88: No, it is no needed. The unnecessary commented codes in python modules will be cleaned.

Tests/Data/Parabolic/T/3D_3BHE_array/pre/3bhes.py
ShuangChen88: The assertion code will be added to ensure the name of BHE is correct.

ProcessLib/HeatTransportBHE/HeatTransportBHEProcess.cpp
HBShaoUFZ: fixed in commit #95f4087el

ProcessLib/HeatTransportBHE/HeatTransportBHEProcess.cpp
HBShaoUFZ: removed.

ProcessLib/HeatTransportBHE/HeatTransportBHEProcessData.h
HBShaoUFZ: fixed in commit #b23114c6

ProcessLib/HeatTransportBHE/CreateHeatTransportBHEProcess.cpp
HBShaoUFZ: fixed in commit #32114c54

ProcessLib/HeatTransportBHE/CreateHeatTransportBHEProcess.cpp
HBShaoUFZ: This dependency cannot be moved to the boundary condition class. The reason is, the OGS-TESPy coupling is not only executed on the boundary condition level. At the end of each linear solver execution, inside of the HeatTransportBHEProcess::postIterationConcreteProcess function, the Python script will compare the simulated temperatures of each BHE and decide whether the OGS iteration should continue or not. This means, the process itself needs to be linked to pybind11 library.

HBShaoUFZ:
> Main questions for discussion: what is the difference to the existing PythonBoundaryCondition and why is the BHE network boundary condition not derived from the existing implementation?

I just discussed this with Shuang and our answer is "probably not". As you see in my previous answer about "pybind11", the coupling requires actions in the post iteration level. This is beyond the boundary condition class and related directly to the process control. Also, in cases where there are hundreds of BHEs (which we will have), the amount of data exchange is not negligible. In the current version, the data from each BHE are collected altogether in a continuous memory space and exchanged with external Python script. In case we use or derive from the PythonBoundaryCondition class, the exchangeable data will be scattering on each BHE inflow boudary condition and treated one after another. Notice that the determination of whether the iteration should go on or not has to be done only in the postIterationConcrete function. We do not gain anything in our favor if we stick to  PythonBoundaryCondition class or its variant.

ProcessLib/HeatTransportBHE/CreateHeatTransportBHEProcess.cpp
bilke: @HBShaoUFZ I have disabled building the process when `OGS_USE_PYTHON=OFF`. This fixed the failing Jenkins jobs.

HBShaoUFZ:
We are not happy with the current status of the code. Dima is correct that the HeatTransportBHE process should not depend on Python interface. Please close this PR for now. Once the code is ready, we will send another PR.

PR #2324, opened 2019-01-17
[ThermoMechanics] support reloading of strains in restart

norihiro-w:
These changes support reloading of strains from previous simulation results. Reloading of stress is already implemented.

### DIFF
--- a/ProcessLib/ThermoMechanics/ThermoMechanicsProcess.cpp
+++ b/ProcessLib/ThermoMechanics/ThermoMechanicsProcess.cpp
@@ -39,7 +39,7 @@ ThermoMechanicsProcess<DisplacementDim>::ThermoMechanicsProcess(
       _process_data(std::move(process_data))
 {
     _integration_point_writer.emplace_back(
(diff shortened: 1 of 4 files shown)


ProcessLib/ThermoMechanics/ThermoMechanicsProcess.cpp
```diff
@@ -54,6 +55,26 @@ ThermoMechanicsProcess<DisplacementDim>::ThermoMechanicsProcess(
                     result[i] = local_asm.getSigma();
                 }
 
+                return result;
+            }));
+
+    _integration_point_writer.emplace_back(
+        std::make_unique<SigmaIntegrationPointWriter>(
+            "epsilon_ip",
```
endJunction: I'd suggest to rename the SigmaIntegrationPointWriter in KelvinVectorIntegrationPointWriter; it looks strange to feed the sigma writer with epsilon_ip.

norihiro-w: done

PR #2325, opened 2019-01-17
support mathematical expressions in Parameter

norihiro-w:
as titled. The implementation is based on exprtk (http://www.partow.net/programming/exprtk/index.html).
So far I named the new parameter type as "Function", but you might have better idea about the naming. let me know...

Remarks
- Currently "x", "y", "z" are supported as variables.
- Example of input files is
```
        <parameter>
            <name>initial_pressure</name>
            <type>Function</type>
            <expression>1000.0*9.81*(-z)</expression>
        </parameter>
```
- For non-scalar parameters, `<expression>` definition should be given for each component. The number of parameter components is set by the number of given `<expression>`.

### DIFF
--- a/Documentation/ProjectFile/prj/parameters/parameter/Function/t_expression.md
+++ b/Documentation/ProjectFile/prj/parameters/parameter/Function/t_expression.md
@@ -1,1 +1,1 @@
-Mathematical expression of the function (currently x,y,z are supported variales). For non-scalar values, an expression should be given for each component.
+Mathematical expression of the function (currently x,y,z are supported variables). For non-scalar values, an expression should be given for each component.
--- a/Documentation/ProjectFile/prj/parameters/parameter/Function/t_expression.md
+++ b/Documentation/ProjectFile/prj/parameters/parameter/Function/t_expression.md
@@ -1,0 +1,1 @@
+Mathematical expression of the function (currently x,y,z are supported variales). For non-scalar values, an expression should be given for each component.
--- a/ProcessLib/Parameter/FunctionParameter.cpp
+++ b/ProcessLib/Parameter/FunctionParameter.cpp
@@ -1,0 +1,37 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2019, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#include "FunctionParameter.h"
+
+#include "BaseLib/ConfigTree.h"
+#include "MeshLib/Mesh.h"
+
+namespace ProcessLib
+{
+std::unique_ptr<ParameterBase> createFunctionParameter(
+    std::string const& name, BaseLib::ConfigTree const& config,
+    MeshLib::Mesh const& mesh)
+{
+    //! \ogs_file_param{prj__parameters__parameter__type}
+    config.checkConfigParameter("type", "Function");
+
+    std::vector<std::string> vec_expressions;
+
+    //! \ogs_file_param{prj__parameters__parameter__Function__expression}
+    for (auto p : config.getConfigSubtreeList("expression"))
+    {
+        std::string const expression_str = p.getValue<std::string>();
+        vec_expressions.emplace_back(expression_str);
+    }
+
+    return std::make_unique<FunctionParameter<double>>(
+        name, mesh, vec_expressions);
+}
+
+}  // ProcessLib
--- a/ProcessLib/Parameter/FunctionParameter.cpp
+++ b/ProcessLib/Parameter/FunctionParameter.cpp
@@ -24,7 +24,7 @@ std::unique_ptr<ParameterBase> createFunctionParameter(
     std::vector<std::string> vec_expressions;
 
     //! \ogs_file_param{prj__parameters__parameter__Function__expression}
-    for (auto p : config.getConfigSubtreeList("expression"))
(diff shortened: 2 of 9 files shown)


endJunction:
I'd add a git submodule instead. Same source code by same author here: https://github.com/ArashPartow/exprtk

"Function" is OK. If we add python functions later, they can be called "Python(Function)"; no clash here.
Maybe "SpatialFunction" would be more precise. With possible variations on "TemporalFunction" or "SpatioTemporalFunction". More suggestions are welcome.

ProcessLib/Parameter/FunctionParameter.h
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+private:
+    MeshLib::Mesh const& _mesh;
+    std::vector<std::string> _vec_expression_str;
+
+    mutable double _x, _y, _z;
+    symbol_table_t _symbol_table;
+    std::vector<expression_t> _vec_expression;
+    mutable std::vector<T> _cache;
```
endJunction: The mutable members are not compatible with openMP parallelization. Better to use local variables in the `operator()`.

ProcessLib/Parameter/FunctionParameter.h
```diff
@@ -0,0 +1,117 @@
...
+        _symbol_table.add_variable("z", _z);
+
+        _vec_expression.resize(_vec_expression_str.size());
+        for (unsigned i=0; i<_vec_expression_str.size(); i++)
+        {
+            _vec_expression[i].register_symbol_table(_symbol_table);
+            parser_t parser;
+            if (!parser.compile(_vec_expression_str[i], _vec_expression[i]))
+            {
+                ERR("Error: %s\tExpression: %s\n", parser.error().c_str(),
```
endJunction: Replace the ERR with OGS_FATAL. Somebody is going to miss that error message otherwise.

ProcessLib/Parameter/FunctionParameter.cpp
```diff
@@ -0,0 +1,37 @@
...
+std::unique_ptr<ParameterBase> createFunctionParameter(
+    std::string const& name, BaseLib::ConfigTree const& config,
+    MeshLib::Mesh const& mesh)
+{
+    //! \ogs_file_param{prj__parameters__parameter__type}
+    config.checkConfigParameter("type", "Function");
+
+    std::vector<std::string> vec_expressions;
+
+    //! \ogs_file_param{prj__parameters__parameter__Function__expression}
```
endJunction: the tag documentation files are missing. See https://github.com/ufz/ogs/tree/master/Documentation/ProjectFile/prj/parameters/parameter

ProcessLib/Parameter/FunctionParameter.h
```diff
@@ -0,0 +1,117 @@
...
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#pragma once
+
+#include <utility>
+#include <vector>
+
+#include <exprtk.hpp>
```
wenqing: The file is large with template definitions. When it was compiled in OGS5 with visual studio 2010, the option of /bigobj is needed. Since the building in appveyor succeeded, I think the new visual studio can handle it without using any additional option.

ProcessLib/Parameter/FunctionParameter.h
```diff
@@ -0,0 +1,117 @@
...
+#include "MeshLib/Node.h"
+
+#include "Parameter.h"
+#include "ProcessLib/Utils/ProcessUtils.h"
+
+namespace ProcessLib
+{
+
+/// A parameter class evaluating a functon defined by
+/// a user-provided maethmatical expression.
```
wenqing: maethmatical --> mathematical

ProcessLib/Parameter/FunctionParameter.h
norihiro-w: @endJunction I will change it. FYI other parameters also have mutable members.

ProcessLib/Parameter/FunctionParameter.h
norihiro-w: i tested with VS2017 and so far it's ok.

ProcessLib/Parameter/FunctionParameter.h
norihiro-w: @endJunction no it's no easy. Because the `operator()` returns a reference, local variables cannot be used here.

ProcessLib/Parameter/FunctionParameter.h
norihiro-w: done

ProcessLib/Parameter/FunctionParameter.h
norihiro-w: done

ProcessLib/Parameter/FunctionParameter.cpp
norihiro-w: done

Documentation/ProjectFile/prj/parameters/parameter/Function/t_expression.md
```diff
@@ -0,0 +1 @@
+Mathematical expression of the function (currently x,y,z are supported variales). For non-scalar values, an expression should be given for each component.
```
TomFischer: variales -> variables

ProcessLib/Parameter/FunctionParameter.cpp
```diff
@@ -0,0 +1,37 @@
...
+    for (auto p : config.getConfigSubtreeList("expression"))
```
TomFischer: `for (auto const& p : config.getConfigSubtreeList("expression"))` avoid copies.

PR #2334, opened 2019-01-29
[PL] Infrastructure for setInitialConditions().

endJunction:
Some processes might require setup of the initial conditions dependent variables in local assemblers.
The new interface allows such implementations.

See https://github.com/endJunction/ogs/tree/RMSaturation branch changes of RichardsFlow process for usage.


ThieJan:
Thanks!
Could the interface be extended adding the current time (initial time) to the interface as well? This enables correct evaluation of parameters or other time dependent properties.

PR #2335, opened 2019-01-30
Refactored Doxygen setup

bilke:
Full Doxygen setup is now done in `DocumentationSetup.cmake` instead of `Doxyfile.in`.

Uses `doxygen_add_docs()` from [FindDoxygen.cmake module](https://cmake.org/cmake/help/latest/module/FindDoxygen.html) which required CMake 3.9 (bumped requirement too).

Made some changes to CMake due to the new minimum version:

- Use imported MPI target
- Added summary of found packages to CMake output


bilke:
@endJunction Ready to merge! One test fails but is unrelated:

```
[ RUN      ] FileIO.TestGmsInterface
GMSInterface::readGMS3DMMesh(): Could not read expected file header.
../Tests/FileIO/TestGmsInterface.cpp:24: Failure
Value of: mesh != nullptr
```

endJunction:
@bilke Git LFS initialized. But shouldn't it happen in the docker?

bilke:
@endJunction No the initial checkout happens before Docker.

PR #2337, opened 2019-01-30
[PL] remove mutable member variables in Parameter classes

norihiro-w:
as suggested in #2325

### DIFF
--- a/ProcessLib/Parameter/CurveScaledParameter.h
+++ b/ProcessLib/Parameter/CurveScaledParameter.h
@@ -35,7 +35,6 @@ struct CurveScaledParameter final : public Parameter<T> {
     {
         _parameter =
             &findParameter<T>(_referenced_parameter_name, parameters, 0);
-        _cache.resize(_parameter->getNumberOfComponents());
     }
 
     int getNumberOfComponents() const override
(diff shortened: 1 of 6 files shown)


endJunction:
This looks good. Though I tested it and it has some overhead now due to copy the vector resulting in 5% or so performance drop. Easily fixable (for most cases) by introducing an `operator()(t, pos, index)`, s.t. the resulting call is `some_parameter(t, pos, 0)` instead of the mostly used `some_parameter(t, pos)[0]`. Not this PR anyway.

I didn't look into the reasons for the few failing ctests.

norihiro-w:
@endJunction In current codes, there are some cases using a returned value of  `Parameter::operator()`   like
```
        _intrinsic_permeability_tensor =
            Eigen::Map<Eigen::Matrix<double,
                                     Eigen::Dynamic,
                                     Eigen::Dynamic,
                                     Eigen::RowMajor> const>(
                _permeability_parameter(t, pos).data(), _dimension, _dimension);
```
In this example, I guess the returned vector of `_permeability_parameter()` is dead after defining `_intrinsic_permeability_tensor `. How do you think?

ProcessLib/Parameter/CurveScaledParameter.h
```diff
@@ -35,32 +35,31 @@ struct CurveScaledParameter final : public Parameter<T> {
...
 
-    std::vector<T> const& operator()(double const t,
+    std::vector<T> operator()(double const t,
                                      SpatialPosition const& pos) const override
     {
         auto const& tup = (*_parameter)(t, pos);
         auto const scaling = _curve.getValue(t);
 
         auto const num_comp = _parameter->getNumberOfComponents();
+        std::vector<T> cache(num_comp);
```
wenqing: May be this change makes a little slow down in the performance. I checked the maximum Total Test time (real) in the building history:
This PR: 494 s
PR #2334:  464.40 s
PR #2338:  457.47 s

TomFischer: The numbers are not very reliable because of changing load while tests run.

PR #2344, opened 2019-02-06
[CMake] Added option OGS_USE_CVODE.

bilke:


bilke:
@endJunction Ready to merge.

endJunction:
Thanks Lars!

General note for users. The current sundials version 4 is not compatible with the OGS implementation.

PR #2348, opened 2019-02-08
[PL] Remove monolithic phasefield implementation.

endJunction:
Current staggered scheme implementation proved to be
much better than the monolithic implementation, which
fails to predict failure in a shear test.


endJunction:
@wenqing I think you refer to the handling of the monolithic *vs.* staggered scheme, like dof table creation/access *etc.* I kept them, because the main compilation "burden" was the `assembleWithJacobian`.
If you think, more of the monolithic scheme should be removed, I'd like to know which parts, and I'll drop them too.

wenqing:
@endJunction  I referred to the following portions. if you like to keep them as a framework of coupling for future development, it is OK.
PhaseFieldProcess.cpp
* L66 and L85: `if (_use_monolithic_scheme || process_id == 0)` -->`if (process_id == 0)`
* L113-130: `if (_use_monolithic_scheme)`
* L182-188: `if (_use_monolithic_scheme)`
* L212-217: `if (_use_monolithic_scheme)`
* L237, L282, L291 and L411  are OK
* L257-261: `if (_use_monolithic_scheme)`

endJunction:
Dropped almost all of the monolithic scheme conditions. Added one more check in process' constructor for the monolithic scheme, s.t. it aborts if monolithic scheme is used.

wenqing:
Looks good.

PR #2352, opened 2019-02-11
Don't reset node element ids always.

endJunction:
For the vtk meshes, where the node and element ids are already set (starting from 0, no gaps, etc) it is better to keep the original numbering. This comes in handy for pre- and post-processing, and sparse matrix bandwidth reduction.

### DIFF
--- a/MeshLib/Mesh.cpp
+++ b/MeshLib/Mesh.cpp
@@ -50,6 +50,10 @@ Mesh::Mesh(std::string name,
     assert(_n_base_nodes <= _nodes.size());
     this->resetNodeIDs();
     this->resetElementIDs();
+    if (_n_base_nodes == 0)
+    {
+        recalculateMaxBaseNodeId();
+    }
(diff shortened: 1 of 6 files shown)


MeshLib/Mesh.cpp
```diff
@@ -119,18 +165,24 @@ void Mesh::addElement(Element* elem)
...
-        for (Element const* e : _elements)
-            for (unsigned i=0; i<e->getNumberOfBaseNodes(); i++)
-                max_basenode_ID = std::max(max_basenode_ID, e->getNodeIndex(i));
-        _n_base_nodes = max_basenode_ID + 1;
+        for (std::size_t i = 0; i < e->getNumberOfBaseNodes(); i++)
+        {
+            max_basenode_ID = std::max(max_basenode_ID, e->getNodeIndex(i));
+        }
     }
+    _n_base_nodes = max_basenode_ID + 1;
```
TomFischer: I know that you only refactored the code. But why this '+ 1'?

PR #2353, opened 2019-02-11
RichardsMechanics anisotropic permeability

endJunction:
Reuse permeability implementation from RichardsFlow process. Anisotropy is included.


ProcessLib/Parameter/ConstantParameter.cpp
```diff
@@ -25,12 +25,20 @@ std::unique_ptr<ParameterBase> createConstantParameter(
...
         if (value)
         {
-            DBUG("Using value %g for constant parameter.", *value);
-            return std::make_unique<ConstantParameter<double>>(name, *value);
+            if (value->size() != 1)
+            {
+                OGS_FATAL(
+                    "Expected to read exactly one value, but %d were given.",
+                    value->size());
+            }
```
TomFischer: Maybe I misunderstood something. But why is only a scalar allowed? IMHO a constant vector/matrix is a vector/matrix with constant entries.

ProcessLib/RichardsMechanics/RichardsMechanicsFEM-impl.h
```diff
@@ -194,9 +219,8 @@ void RichardsMechanicsLocalAssembler<
         auto const mu = _process_data.flow_material->getFluidViscosity(
             -p_cap_ip, temperature);
 
-        double const K_intrinsic =
-            _process_data.intrinsic_permeability(t, x_position)[0];
-        double const K_over_mu = K_intrinsic / mu;
```
TomFischer: Why not keep `double const K_over_mu = k_rel / mu * K_intrinsic;` This expression is used twice below. Maybe pre-calculating the value can save some time?

ProcessLib/Parameter/ConstantParameter.cpp
endJunction: You are right. Multiple values are also possible but with \<value**s**\> tag. for the \<value\> tag only scalar is allowed.

ProcessLib/RichardsMechanics/RichardsMechanicsFEM-impl.h
```diff
@@ -232,10 +256,11 @@ void RichardsMechanicsLocalAssembler<
         K.template block<pressure_size, pressure_size>(pressure_index,
                                                        pressure_index)
             .noalias() +=
-            dNdx_p.transpose() * rho_LR * k_rel * K_over_mu * dNdx_p * w;
+            dNdx_p.transpose() * rho_LR * k_rel / mu * K_intrinsic * dNdx_p * w;
```
wenqing: ` rho_LR * k_rel * K_intrinsic  / mu` is repeated several times.

ProcessLib/RichardsMechanics/RichardsMechanicsFEM-impl.h
```diff
@@ -508,17 +533,18 @@ void RichardsMechanicsLocalAssembler<ShapeFunctionDisplacement,
...
-                          dk_rel_dS_l * dS_L_dp_cap * N_p * w;
+            .noalias() += dNdx_p.transpose() * rho_LR / mu * K_intrinsic *
+                          grad_p_cap * dk_rel_dS_l * dS_L_dp_cap * N_p * w;
 
         local_Jac
             .template block<pressure_size, pressure_size>(pressure_index,
                                                           pressure_index)
-            .noalias() += dNdx_p.transpose() * rho_LR * rho_LR * K_over_mu * b *
-                          dk_rel_dS_l * dS_L_dp_cap * N_p * w;
+            .noalias() += dNdx_p.transpose() * rho_LR * rho_LR / mu *
```
wenqing: Sounds missing krel.

ProcessLib/RichardsMechanics/RichardsMechanicsFEM-impl.h
```diff
@@ -642,16 +671,16 @@ std::vector<double> const& RichardsMechanicsLocalAssembler<
...
+            t, x_position, material_id, *_process_data.flow_material);
         auto const rho_LR = _process_data.flow_material->getFluidDensity(
             -p_cap_ip, temperature);
         auto const& b = _process_data.specific_body_force;
 
         // Compute the velocity
         auto const& dNdx_p = _ip_data[ip].dNdx_p;
-        cache_matrix.col(ip).noalias() =
-            -K_over_mu * dNdx_p * p_L - K_over_mu * rho_LR * b;
+        cache_matrix.col(ip).noalias() = -1 / mu * K_intrinsic * dNdx_p * p_L -
```
wenqing: Sounds missing krel.

ProcessLib/RichardsMechanics/RichardsMechanicsFEM-impl.h
```diff
@@ -508,17 +533,18 @@ void RichardsMechanicsLocalAssembler<ShapeFunctionDisplacement,
         local_Jac
             .template block<pressure_size, pressure_size>(pressure_index,
                                                           pressure_index)
-            .noalias() += dNdx_p.transpose() * rho_LR * K_over_mu * grad_p_cap *
-                          dk_rel_dS_l * dS_L_dp_cap * N_p * w;
+            .noalias() += dNdx_p.transpose() * rho_LR / mu * K_intrinsic *
```
wenqing: Sounds missing krel.

ProcessLib/RichardsMechanics/RichardsMechanicsFEM-impl.h
endJunction: I would think that here dk_rel/dS_l is used instead of k_rel. Same for the other derivative expression in line ~530.

ProcessLib/RichardsMechanics/RichardsMechanicsFEM-impl.h
endJunction: Here it is missing indeed.

PR #2341, opened 2019-02-01
Implemented staggered scheme in single component transport process

renchao-lu:
As titled.

All the benchmarks in the folder _SimpleSynthetics_ have been reproduced in this pull request.

To-do list:
- [x] Remove coding for PETSc;
- [x] Assign more strict tolerance for vtkdiff;
- [x] Improve naming of variables;

PR #2357, opened 2019-02-13
Extended Staggered Scheme in Multi-component Transport Process

renchao-lu:
This pull request is based upon #2341. With this implementation, staggered scheme is ready for multi-component transport simulation.


ProcessLib/ComponentTransport/ComponentTransportFEM.h
```diff
@@ -358,27 +373,329 @@ class LocalAssemblerData : public ComponentTransportLocalAssemblerInterface
...
+
+            auto const local_p = Eigen::Map<const NodalVectorType>(
+                &local_x[pressure_index], pressure_size);
+            int const concentration_index = 1 * ShapeFunction::NPOINTS;
+            auto const local_C = Eigen::Map<const NodalVectorType>(
+                &local_x[concentration_index], concentration_size);
+
+            return calculateIntPtDarcyVelocity(t, local_p, local_C, cache);
+        }
+        else  // staggered scheme
```
endJunction: no else after return...

ProcessLib/ComponentTransport/ComponentTransportFEM.h
```diff
@@ -358,27 +373,329 @@ class LocalAssemblerData : public ComponentTransportLocalAssemblerInterface
...
 
-        // Assuming that fluid density always depends on the concentration of
-        // the component placed at the first.
-        auto const concentration_index = 1 * ShapeFunction::NPOINTS;
-        // get local_C and local_p
-        auto const C_nodal_values = Eigen::Map<const NodalVectorType>(
-            &local_x[concentration_index], concentration_size);
-        auto const p_nodal_values = Eigen::Map<const NodalVectorType>(
-            &local_x[pressure_index], pressure_size);
+        if (_coupled_solutions == nullptr)  // monolithic scheme
```
endJunction: if (_use_monolithic_scheme)

renchao-lu:
Done. Please review.

ProcessLib/ComponentTransport/ComponentTransportProcess.h
```diff
@@ -146,7 +146,7 @@ class ComponentTransportProcess final : public Process
         _local_assemblers;
 
     /// Solutions of the previous time step
-    std::array<std::unique_ptr<GlobalVector>, 2> _xs_previous_timestep;
+    std::vector<std::unique_ptr<GlobalVector>> _xs_previous_timestep;
```
TomFischer: Is this a bug fix?

ProcessLib/ComponentTransport/ComponentTransportFEM.h
```diff
@@ -499,6 +499,7 @@ class LocalAssemblerData : public ComponentTransportLocalAssemblerInterface
         std::vector<double>& /*local_b_data*/,
         LocalCoupledSolutions const& coupled_xs)
     {
+        auto const& transport_process_id = coupled_xs.process_id;
```
TomFischer: Is it correct that this change is only relevant in the staggered scheme? If so the variable name should reflect this special use. What do you think about `transport_process_id_staggered`?

ProcessLib/ComponentTransport/ComponentTransportProcess.h
renchao-lu: Last pull request (#2341) achieved functionality of staggered coupling scheme in single component transport process which has two primary variables, i.e., one for hydraulic pressure and the other for component concentration. Currently, we would like to use staggered scheme for multi-component transport simulation.

ProcessLib/ComponentTransport/ComponentTransportFEM.h
renchao-lu: `transport_process_id` was set as a static constant variable in previous implementation (#2341) where single component transport process is given consideration. This change is to get a particular transport process id from local coupled solutions where solutions of multiple transport processes are stored.

TomFischer: What do you think about `particular_transport_process_id` or `coupled_solutions_transport_process_id`? Please let the reader of your code know that this is not the usual `some_process_id`, but it is used here in the staggered context.

renchao-lu: Changed to `particular_transport_process_id`.

ProcessLib/ComponentTransport/ComponentTransportFEM.h
```diff
@@ -80,7 +80,7 @@ class LocalAssemblerData : public ComponentTransportLocalAssemblerInterface
     // When staggered scheme is adopted, nodal pressure and nodal concentration
     // are accessed by process id.
     static const int hydraulic_process_id = 0;
-    static const int transport_process_id = 1;
+    static const int first_transport_process_id = 1;
```
TomFischer: IMHO, for consistent naming scheme, it would be better to revert this name change.

renchao-lu: `first_transport_process_id` is used for calculating fluid density, with the assumption that fluid density is dependent upon the concentration of the component which is placed uppermost. There is no such an issue in single component transport simulation.

endJunction: Then I recommend to add a comment explaining the variable purpose (if it cannot be explained by variable name).
```c++
    // TODO (renchao-lu): The variable is used in the calculation of the fluid's
    // density and flux indicating the transport process id. For now it is
    // assumed that these quantities depend on the first occurring transport
    // process only. The density and flux calculations have to be extended to
    // all processes.
```
*sic*!

ProcessLib/ComponentTransport/ComponentTransportFEM.h
```diff
@@ -499,8 +499,9 @@ class LocalAssemblerData : public ComponentTransportLocalAssemblerInterface
         std::vector<double>& /*local_b_data*/,
         LocalCoupledSolutions const& coupled_xs)
     {
+        auto const& particular_transport_process_id = coupled_xs.process_id;
```
endJunction: In my opinion, this is the only line (aside the vector and resize) relevant to the described changes.
```c++
auto const& transport_process_id = coupled_xs.process_id;
```
All of the renames can be dropped.

Issue #2347, opened 2019-02-08
Pushing request to a forked repo needs id and pwd of gitlab.opengeosys.org


zhangning737:
When I try to push some notes to my forked repo, it shows a ID and pwd of ```gitlab.opengeosys.org``` are needed.

Any else needs to set for my forked repo?

Thanks.

endJunction:
The documentation is somewhat hidden: https://www.opengeosys.org/docs/devguide/advanced/gitlab/
Your account is setup; Set your password, before you push.

For forking/cloning/downloading the git-lfs access to the gitlab server is not needed, only if you push lfs data.

zhangning737:
@endJunction ID is unblocked, thanks.

bilke:
@zhangning737 That is indeed problematic for first time contributors, we have to find another solution..

@endJunction We could use https://ogs.jfrog.io/ which allows non-authenticated pushes for git lfs repos. Migration should be seamless (hopefully) ... What do you think?

endJunction:
@bilke Good idea. Would improve the developers usability. Let's try to start it on a Thursday evening s.t. it runs smoothly by mid Monday.

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/2347).

PR #2359, opened 2019-02-14
Migrate LFS from GitLab to Artifactory

bilke:
Our Artifactory at https://ogs.jfrog.io/ allows non-authenticated pushes for git lfs files in contrast to GitLab where user always needed to authenticate which was quite uncomfortable especially for new contributors.

We have no size limit on Artifactory so we should be safe for the future.

Related to #2347.

TODO after merge:

- [ ] Switch off git lfs on GitLab to make sure nobody pushes to the old repo


bilke:
@ufz/members !
I switched off the old GitLab-based Git LFS server. Please rebase your branches on `origin/master` to get the right lfs server!

PR #2366, opened 2019-02-18
[Jenkins] Added [ci skip] functionality.

bilke:
When a commit message contains `[ci skip]` the Jenkins build will be skipped. Can be useful to not overload our CI system when testing is not necessary.


endJunction:
Of course it add the burden to clean up the commit messages afterwards.

PR #2367, opened 2019-02-18
Mesh rasterisation

rinkk:
Takes a 2D mesh and constructs an asc-raster by sampling elevation values at user-specified equidistant intervals.


Applications/Utils/FileConverter/MeshToRaster.cpp
```diff
@@ -0,0 +1,205 @@
+/**
+ * \file moveMeshNodes.cpp
```
endJunction: Wrong file name. Just use `\file`

Applications/Utils/FileConverter/MeshToRaster.cpp
```diff
@@ -0,0 +1,205 @@
...
+ * 2012/03/07 KR Initial implementation
+ *
+ * \copyright
+ * Copyright (c) 2012-2018, OpenGeoSys Community (http://www.opengeosys.org)
```
endJunction: 2019

Applications/Utils/FileConverter/MeshToRaster.cpp
```diff
@@ -0,0 +1,205 @@
...
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/LICENSE.txt
+ */
+
+#include <memory>
+#include <string>
+#include <iostream>
```
endJunction: Don't include iostream, use logog.

Applications/Utils/FileConverter/MeshToRaster.cpp
```diff
@@ -0,0 +1,205 @@
...
+#include "GeoLib/AABB.h"
+#include "GeoLib/AnalyticalGeometry.h"
+#include "MeshLib/IO/readMeshFromFile.h"
+#include "MeshLib/Mesh.h"
+#include "MeshLib/Node.h"
+#include "MeshLib/Elements/Tri.h"
+#include "MeshLib/Elements/Quad.h"
+#include "MeshLib/MeshSearch/MeshElementGrid.h"
+
+/// Returns the index of the element p is located in when projected onto a mesh.
```
endJunction: Probably just me, but what is the "element p"?

Applications/Utils/FileConverter/MeshToRaster.cpp
```diff
@@ -0,0 +1,205 @@
...
+                std::array<double, 4> const x_off{{-half_cell, half_cell, -half_cell, half_cell}};
+                std::array<double, 4> const y_off{{-half_cell, -half_cell, half_cell, half_cell}};
+                double sum (0);
+                std::size_t nonzero_count (0);
+                // test for all of the pixel's corner if any are within an element
+                for (std::size_t i=0; i<4; ++i)
+                {
+                    MeshLib::Node const node(x+x_off[i], y+y_off[i], 0);
+                    std::size_t const elem_idx = getProjectedElement(elems, node);
+                    if (elem_idx != std::numeric_limits<std::size_t>::max())
```
endJunction: This if condition seems to be always false, because it is in the else branch of the same test in line 169

Applications/Utils/FileConverter/MeshToRaster.cpp
```diff
@@ -0,0 +1,205 @@
...
+    std::size_t const nNodes (mesh->getNumberOfNodes());
+    std::vector<MeshLib::Node*> const& nodes_vec(mesh->getNodes());
+    GeoLib::AABB bounding_box(nodes_vec.begin(), nodes_vec.end());
+    MathLib::Point3d const& min(bounding_box.getMinPoint());
+    MathLib::Point3d const& max(bounding_box.getMaxPoint());
+    std::size_t const n_cols = static_cast<std::size_t>(std::ceil((max[0] - min[0]) / cellsize));
+    std::size_t const n_rows = static_cast<std::size_t>(std::ceil((max[1] - min[1]) / cellsize));
+
+    // raster header
+    std::ofstream out(output_arg.getValue(), std::ios::out);
```
endJunction: ofstream is already having the default argument for the open mode set to `std::ios::out`, no need to mention it here.

Applications/Utils/FileConverter/MeshToRaster.cpp
```diff
@@ -0,0 +1,205 @@
...
+        ERR("The programme requires a mesh containing two-dimensional elements "
+            "(i.e. triangals or quadrilaterals.");
+        return 2;
+    }
+
+    double const cellsize = (cell_arg.isSet()) ? cell_arg.getValue()
+                                               : mesh->getMinEdgeLength();
+    INFO ("Cellsize set to %f", cellsize);
+
+    std::size_t const nNodes (mesh->getNumberOfNodes());
```
endJunction: unused

GeoLib/AnalyticalGeometry.h
```diff
@@ -52,6 +53,17 @@ Orientation getOrientation (const GeoLib::Point* p0,
                             const GeoLib::Point* p1,
                             const GeoLib::Point* p2);
 
+
+/**
+ * Tests if p is right of the line defined by a and b in 2D.
+ * (Note: This functionality is similar to getOrientation(). It is probably
+ * not quite as exact in numerically challenging situations but requires only
+ * about half the runtime.)
```
endJunction: I'd recommend to prefer the correctness over runtime in such cases.

Applications/Utils/FileConverter/MeshToRaster.cpp
```diff
@@ -0,0 +1,205 @@
...
+        MeshLib::IO::readMeshFromFile(input_arg.getValue()));
+    if (mesh == nullptr)
+    {
+        ERR("Error reading mesh file.");
+        return 1;
+    }
+    if (mesh->getDimension() != 2)
+    {
+        ERR("The programme requires a mesh containing two-dimensional elements "
+            "(i.e. triangals or quadrilaterals.");
```
endJunction: triangles.

Applications/Utils/FileConverter/MeshToRaster.cpp
```diff
@@ -0,0 +1,205 @@
...
+        "o", "output-file", "Raster output file (*.asc)", true, "", "string");
+    cmd.add(output_arg);
+    TCLAP::ValueArg<double> cell_arg(
+        "c", "cellsize", "edge length of raster cells in result", false, 1, "real");
+    cmd.add(cell_arg);
+
+    cmd.parse(argc, argv);
+
+    INFO("Rasterising mesh...");
+    if (!input_arg.isSet())
```
endJunction: Isn't tclap already testing this? The third last argument to the constructor of the argument is the requirement, isn't it?

Applications/Utils/FileConverter/MeshToRaster.cpp
```diff
@@ -0,0 +1,205 @@
...
+#include "MeshLib/Node.h"
+#include "MeshLib/Elements/Tri.h"
+#include "MeshLib/Elements/Quad.h"
+#include "MeshLib/MeshSearch/MeshElementGrid.h"
+
+/// Returns the index of the element p is located in when projected onto a mesh.
+std::size_t getProjectedElement(std::vector<const MeshLib::Element*> const& elems,
+                                MeshLib::Node const& node)
+{
+   // std::vector<MeshLib::Element*> const& elems = mesh.getElements();
```
endJunction: drop unused code...

Applications/Utils/FileConverter/MeshToRaster.cpp
rinkk: :heavy_check_mark:

Applications/Utils/FileConverter/MeshToRaster.cpp
rinkk: :heavy_check_mark:

Applications/Utils/FileConverter/MeshToRaster.cpp
rinkk: :heavy_check_mark:

Applications/Utils/FileConverter/MeshToRaster.cpp
rinkk: :heavy_check_mark:

Applications/Utils/FileConverter/MeshToRaster.cpp
rinkk: :heavy_check_mark:

Applications/Utils/FileConverter/MeshToRaster.cpp
rinkk: :heavy_check_mark:

Applications/Utils/FileConverter/MeshToRaster.cpp
rinkk: :heavy_check_mark:

Applications/Utils/FileConverter/MeshToRaster.cpp
rinkk: :heavy_check_mark:

Applications/Utils/FileConverter/MeshToRaster.cpp
rinkk: :heavy_check_mark:

GeoLib/AnalyticalGeometry.h
rinkk: This is probably my computer graphics POV but nothing bad happens if this gives a result that is wrong by a measure of epsilon outside of the elevation also being wrong by an epsilon (which is most likely less of an error than any rounding artifacts, measurement errors, etc that happened before already). Should I still remove it and use the slower method?

endJunction: @TomFischer your decision.

endJunction:
Is it possible to have a small example to be included in the ctests?

Applications/Utils/FileConverter/MeshToRaster.cpp
```diff
@@ -0,0 +1,189 @@
...
+    }
+    return std::numeric_limits<size_t>::max();
+}
+
+/// Returns the z-coordinate of a point projected onto the plane defined by a
+/// mesh element.
+double getElevation(MeshLib::Element const& elem, MeshLib::Node const& node)
+{
+    MathLib::Vector3 const n = MeshLib::FaceRule::getSurfaceNormal(&elem).getNormalizedVector();
+    MeshLib::Node const orig = *elem.getNode(0);
```
TomFischer: Why making a copy here?
```MeshLib::Node const& orig = *elem.getNode(0);```

File: Applications/Utils/FileConverter/MeshToRaster.cpp [hunk 13]
```diff
@@ -0,0 +1,189 @@
...
+            "(i.e. triangles or quadrilaterals.");
+        return 2;
+    }
+
+    double const cellsize = (cell_arg.isSet()) ? cell_arg.getValue()
+                                               : mesh->getMinEdgeLength();
+    INFO ("Cellsize set to %f", cellsize);
+
+    std::vector<MeshLib::Node*> const& nodes_vec(mesh->getNodes());
+    GeoLib::AABB bounding_box(nodes_vec.begin(), nodes_vec.end());
```
TomFischer: const

Applications/Utils/FileConverter/MeshToRaster.cpp
```diff
@@ -0,0 +1,189 @@
...
+#include "GeoLib/AnalyticalGeometry.h"
+#include "MeshLib/IO/readMeshFromFile.h"
+#include "MeshLib/Mesh.h"
+#include "MeshLib/Node.h"
+#include "MeshLib/Elements/Tri.h"
+#include "MeshLib/Elements/Quad.h"
+#include "MeshLib/MeshSearch/MeshElementGrid.h"
+
+/// Returns the index of the element the given node is located in when projected onto a mesh.
+std::size_t getProjectedElement(std::vector<const MeshLib::Element*> const& elems,
```
TomFischer: The function name doesn't express exactly what the comment says and what is returned.

Applications/Utils/FileConverter/MeshToRaster.cpp
```diff
@@ -0,0 +1,189 @@
...
+        {
+            MeshLib::Node const node(x, y, 0);
+            MathLib::Point3d min_vol{{x - max_edge, y - max_edge, -std::numeric_limits<double>::max()}};
+            MathLib::Point3d max_vol{{x + max_edge, y + max_edge, std::numeric_limits<double>::max()}};
+            std::vector<const MeshLib::Element*> const elems = grid.getElementsInVolume(min_vol, max_vol);
+            std::size_t const elem_idx = getProjectedElement(elems, node);
+            // centre of the pixel is located within a mesh element
+            if (elem_idx != std::numeric_limits<std::size_t>::max())
+                out << getElevation(*(mesh->getElement(elem_idx)), node) << " ";
+            // centre of the pixel isn't located within a mesh element
```
TomFischer: From the above comment it is already clear that the centre of the pixel isn't located within a mesh element. Maybe, it is better to write what you'll do in the following (checking the corner points of the element).

TomFischer:
You could use `GeoLib::Raster` and `Applications/FileIO/AsciiRasterInterface` for the implementation to reuse existing code.

Applications/Utils/FileConverter/MeshToRaster.cpp
rinkk: :heavy_check_mark:

Applications/Utils/FileConverter/MeshToRaster.cpp
rinkk: :heavy_check_mark:

Applications/Utils/FileConverter/MeshToRaster.cpp
rinkk: :heavy_check_mark:

Applications/Utils/FileConverter/MeshToRaster.cpp
rinkk: :heavy_check_mark:

Applications/Utils/FileConverter/MeshToRaster.cpp
rinkk: There are multiple comments below this line, *including* one that mentions the corner points.

GeoLib/AnalyticalGeometry.h
TomFischer: I would take the safe implementation.

rinkk:
The raster interface is hard to use in this case. For one thing, it'd require writing more code than I did. Most annoyingly, it'd require storing all the data first before I can start creating a raster object (currently I'm not storing anything!), then creating a header object and *then* writing all of it to a file.

rinkk:
I'm using the slower (but presumably more correct) method now for determining if a point is in a triangle.

For the record: I think it's extremely annoying to do this. The runtime is noticably longer now but the result for the test data is exactly the same. Not a single byte was changed. We're basically hobbling our legs here, nobody was ever going to notice a difference in the results...

TomFischer:
I would suggest to put the asc test file to git large filesystem.

rinkk:
I did the test precisely as both Dima and Lars said I should do it.

endJunction:
@rinkk  For discussion an refactored version of the tool https://github.com/endJunction/ogs/tree/master2raster.

TomFischer:
I don't insist to move the raster test data to git large file system.

rinkk:
:+1:

PR #2368, opened 2019-02-18
[PL] Move solution assignment of previous time step forward

renchao-lu:
As titled. Getting all the solutions of previous time step has to be done before looping over coupled processes, because in some coupled problems all the PDEs have the coupling terms.


renchao-lu:
Done. Moved out of the loop. @wenqing  @endJunction

renchao-lu:
Done. Cherry-picked the commit for fixing the HM benchmark.

PR #2369, opened 2019-02-18
Geo-point to station conversion

rinkk:
Converts geometry points into station objects.

A flag signals if all points should be converted or just the ones not used in any polyline or surface.

### DIFF
--- a/GeoLib/GEOObjects.cpp
+++ b/GeoLib/GEOObjects.cpp
@@ -554,9 +554,83 @@ void GEOObjects::renameGeometry(std::string const& old_name,
     }
 }
 
-const GeoLib::GeoObject* GEOObjects::getGeoObject(const std::string &geo_name,
-                                                            GeoLib::GEOTYPE type,
-                                                            const std::string &geo_obj_name) const
+void GEOObjects::markPointsAsUsed(std::string const& geo_name,
+                                  std::vector<bool>& flags)
+{
+    GeoLib::PolylineVec const* const ply_obj(getPolylineVecObj(geo_name));
+    if (ply_obj)
+    {
+        std::vector<GeoLib::Polyline*> const& lines(*ply_obj->getVector());
+        for (auto line : lines)
+        {
+            std::size_t const n_pnts(line->getNumberOfPoints());
+            for (std::size_t i = 0; i < n_pnts; ++i)
+                flags[line->getPointID(i)] = false;
+        }
+    }
+
+    GeoLib::SurfaceVec const* const sfc_obj(getSurfaceVecObj(geo_name));
+    if (sfc_obj)
+    {
+        std::vector<GeoLib::Surface*> const& surfaces = *sfc_obj->getVector();
+        for (auto sfc : surfaces)
+        {
+            std::size_t const n_tri(sfc->getNumberOfTriangles());
+            for (std::size_t i = 0; i < n_tri; ++i)
+            {
+                GeoLib::Triangle const& t = *(*sfc)[i];
+                flags[t[0]] = false;
+                flags[t[1]] = false;
+                flags[t[2]] = false;
+            }
+        }
+    }
+}
+
+void GEOObjects::geoPointsToStation(std::string const& geo_name,
+                                    std::string const& stn_name,
+                                    bool only_unused_pnts)
+{
+    GeoLib::PointVec const* const pnt_obj(getPointVecObj(geo_name));
+    if (pnt_obj == nullptr)
+    {
+        ERR("Point vector %s not found.", geo_name.c_str());
+        return;
+    }
+    std::vector<GeoLib::Point*> const& pnts = *pnt_obj->getVector();
+    if (pnts.empty())
+    {
+        ERR("Point vector %s is empty.", geo_name.c_str());
+        return;
+    }
+    std::size_t const n_pnts(pnts.size());
+    std::vector<bool> flags(n_pnts, true);
+    if (only_unused_pnts)
+        markPointsAsUsed(geo_name, flags);
+
+    std::unique_ptr<std::vector<GeoLib::Point*>> stations(
+        new std::vector<GeoLib::Point*>);
+    for (std::size_t i = 0; i < n_pnts; ++i)
+    {
+        if (!flags[i])
+            continue;
+        std::string name = pnt_obj->getItemNameByID(i);
+        if (name.empty())
+            name = "Station " + std::to_string(i);
+        stations->push_back(new GeoLib::Station((*pnts[i])[0], (*pnts[i])[1],
+                                                (*pnts[i])[2], name));
+    }
+    std::string vec_name = geo_name + " stations";
+    if (!stations->empty())
+        addStationVec(std::move(stations), vec_name);
+    else
+        WARN("No points found to convert.");
+}
+
+const GeoLib::GeoObject* GEOObjects::getGeoObject(
+    const std::string& geo_name,
+    GeoLib::GEOTYPE type,
+    const std::string& geo_obj_name) const
 {
     GeoLib::GeoObject *geo_obj(nullptr);
     switch (type) {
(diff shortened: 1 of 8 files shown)


GeoLib/GEOObjects.cpp
```diff
@@ -551,9 +551,83 @@ void GEOObjects::renameGeometry(std::string const& old_name,
...
+                flags[t[1]] = false;
+                flags[t[2]] = false;
+            }
+        }
+    }
+}
+
+void GEOObjects::geoPointsToStation(std::string const& geo_name,
+                                    std::string const& stn_name,
+                                    bool only_unused_pnts)
```
endJunction: const

GeoLib/GEOObjects.cpp
```diff
@@ -551,9 +551,83 @@ void GEOObjects::renameGeometry(std::string const& old_name,
     }
 }
 
-const GeoLib::GeoObject* GEOObjects::getGeoObject(const std::string &geo_name,
-                                                            GeoLib::GEOTYPE type,
-                                                            const std::string &geo_obj_name) const
+void GEOObjects::markPointsAsUsed(std::string const& geo_name,
+                                  std::vector<bool>& flags)
```
endJunction: better to return the flags instead of modifying.

endJunction: if the function does not modify anything, mark it as const.

GeoLib/GEOObjects.cpp
```diff
@@ -551,9 +551,83 @@ void GEOObjects::renameGeometry(std::string const& old_name,
...
+{
+    GeoLib::PolylineVec const* const ply_obj(getPolylineVecObj(geo_name));
+    if (ply_obj)
+    {
+        std::vector<GeoLib::Polyline*> const& lines(*ply_obj->getVector());
+        for (auto line : lines)
```
endJunction: Minor: better to write `auto* line` forcing the value to be a pointer. Same for surfaces.

GeoLib/GEOObjects.cpp
```diff
@@ -551,9 +551,83 @@ void GEOObjects::renameGeometry(std::string const& old_name,
...
+        return;
+    }
+    std::vector<GeoLib::Point*> const& pnts = *pnt_obj->getVector();
+    if (pnts.empty())
+    {
+        ERR("Point vector %s is empty.", geo_name.c_str());
+        return;
+    }
+    std::size_t const n_pnts(pnts.size());
+    std::vector<bool> flags(n_pnts, true);
```
endJunction: "flags" could be named somehow more meaningful. A vector of booleans is already saying that these are flags...

GeoLib/GEOObjects.cpp
```diff
@@ -551,9 +551,83 @@ void GEOObjects::renameGeometry(std::string const& old_name,
...
+    if (only_unused_pnts)
+        markPointsAsUsed(geo_name, flags);
+
+    std::unique_ptr<std::vector<GeoLib::Point*>> stations(
+        new std::vector<GeoLib::Point*>);
```
endJunction:
```
    auto stations = std::make_unique<std::vector<GeoLib::Point*>();
```

GeoLib/GEOObjects.cpp
TomFischer: Minor: Maybe the name of the method could be changed to 'markUsedPoints' that is a tiny bit more readable IMHO.

GeoLib/GEOObjects.cpp
```diff
@@ -551,9 +551,83 @@ void GEOObjects::renameGeometry(std::string const& old_name,
...
+{
+    GeoLib::PolylineVec const* const ply_obj(getPolylineVecObj(geo_name));
+    if (ply_obj)
+    {
+        std::vector<GeoLib::Polyline*> const& lines(*ply_obj->getVector());
+        for (auto line : lines)
+        {
+            std::size_t const n_pnts(line->getNumberOfPoints());
+            for (std::size_t i = 0; i < n_pnts; ++i)
+                flags[line->getPointID(i)] = false;
```
TomFischer: I would reverse the logic. From the method name I would expect that used points have the value true.

GeoLib/GEOObjects.cpp
```diff
@@ -551,9 +551,83 @@ void GEOObjects::renameGeometry(std::string const& old_name,
...
+        std::vector<GeoLib::Surface*> const& surfaces = *sfc_obj->getVector();
+        for (auto sfc : surfaces)
+        {
+            std::size_t const n_tri(sfc->getNumberOfTriangles());
+            for (std::size_t i = 0; i < n_tri; ++i)
+            {
+                GeoLib::Triangle const& t = *(*sfc)[i];
+                flags[t[0]] = false;
+                flags[t[1]] = false;
+                flags[t[2]] = false;
```
TomFischer: Again, I would reverse the logic.

GeoLib/GEOObjects.cpp
```diff
@@ -551,9 +551,83 @@ void GEOObjects::renameGeometry(std::string const& old_name,
...
+}
+
+void GEOObjects::geoPointsToStation(std::string const& geo_name,
+                                    std::string const& stn_name,
+                                    bool only_unused_pnts)
+{
+    GeoLib::PointVec const* const pnt_obj(getPointVecObj(geo_name));
+    if (pnt_obj == nullptr)
+    {
+        ERR("Point vector %s not found.", geo_name.c_str());
```
TomFischer: This error message and also the following messages are printed to the terminal. Is there also message/error box for the DataExplorer?

GeoLib/GEOObjects.cpp
```diff
@@ -551,9 +551,83 @@ void GEOObjects::renameGeometry(std::string const& old_name,
...
+    std::unique_ptr<std::vector<GeoLib::Point*>> stations(
+        new std::vector<GeoLib::Point*>);
+    for (std::size_t i = 0; i < n_pnts; ++i)
+    {
+        if (!flags[i])
+            continue;
+        std::string name = pnt_obj->getItemNameByID(i);
+        if (name.empty())
+            name = "Station " + std::to_string(i);
+        stations->push_back(new GeoLib::Station((*pnts[i])[0], (*pnts[i])[1],
```
TomFischer: You can use the `GeoLib::Station` constructor that takes a `GeoLib::Point`.

GeoLib/GEOObjects.cpp
rinkk: Renamed method and set const.

I can return the flags-array instead of returning it, but wouldn't that mean I'd create a copy of it? I need the vector in the calling method anyway, so if I return it here, I'd have the vector twice?!

GeoLib/GEOObjects.cpp
rinkk: :heavy_check_mark:

GeoLib/GEOObjects.cpp
rinkk: :heavy_check_mark:

GeoLib/GEOObjects.cpp
rinkk: :heavy_check_mark:

GeoLib/GEOObjects.cpp
rinkk: renamed the method :heavy_check_mark:

GeoLib/GEOObjects.cpp
rinkk: renamed the method :heavy_check_mark:

GeoLib/GEOObjects.cpp
rinkk: added error msg for DE :heavy_check_mark:

GeoLib/GEOObjects.cpp
rinkk: :heavy_check_mark:

GeoLib/GEOObjects.cpp
rinkk: :heavy_check_mark:

endJunction:
For the main new function in GeoLib it would be better to have a unit tests. Otherwise the code is difficult to maintain.

PR #2370, opened 2019-02-19
Local coordinate system

endJunction:
See the project file for the new possibilities. `<local_coordinate_system>` and a tag in permeability parameter "k" for usage.
The designated parameters will be transformed according to the given local coordinate system.
An additional if-condition is introduced in all parameters, but it has less than 1% impact on the runtime.
Review commit-wise is possible.

TODO:

- [x] Agree on the input file tags. (decided to keep basis vectors)
- [x] Documentation for the basis vector (which might change with the above tags todo).
- [x] Showcase for anisotropic tensor (3D would be preferable, because 2D is already tested in RM)


JobstM:
I just took a brief look on the project file - for me it's ok. One question: How to define multiple local coordinate systems?
Best

endJunction:
@JobstM Do you mean for different material groups? Or, indeed, multiple, space-dependent coordinate systems?

JobstM:
I mean the possibility to define several coordinate systems, which could be used eg in different areas (material groups) or different material models (eg permeability, elasticity, heat conductivity). Example: A model with 3 material groups, in material group 1, the permeability tensor is inclined by 20 deg, in material group 2 and 3 by 30 deg.

ProcessLib/Parameter/CoordinateSystem.cpp
```diff
@@ -0,0 +1,110 @@
...
+    }
+    if (_base[0]->isTimeDependent() || _base[1]->isTimeDependent())
+    {
+        OGS_FATAL("The parameters for the basis must not be time dependent.");
+    }
+    if (_base[0]->getNumberOfComponents() != 2 ||
+        _base[1]->getNumberOfComponents() != 2)
+    {
+        OGS_FATAL("The parameters for the 2D basis must have two components.");
+    }
```
wenqing: In case there are wrong input by mistakes, it is also good to check the values of the bases, such as e_i \dot e_j := (i==j) ? 1, 0.

ProcessLib/Parameter/CoordinateSystem.cpp
```diff
@@ -0,0 +1,110 @@
...
+        OGS_FATAL("The parameters for the basis must not be time dependent.");
+    }
+    if (_base[0]->getNumberOfComponents() != 3 ||
+        _base[1]->getNumberOfComponents() != 3 ||
+        _base[2]->getNumberOfComponents() != 3)
+    {
+        OGS_FATAL(
+            "The parameters for the 3D basis must have three components.");
+    }
+}
```
wenqing: Here the same, to check the values of the base vectors.

Tests/Data/RichardsMechanics/flow_fully_saturated_coordinate_system.prj
```diff
@@ -0,0 +1,291 @@
...
+                <variable>displacement</variable>
+                <variable>pressure</variable>
+                <variable>sigma</variable>
+                <variable>epsilon</variable>
+                <variable>velocity</variable>
+                <variable>saturation</variable>
+            </variables>
+        </output>
+    </time_loop>
+    <local_coordinate_system>
```
wenqing: There may be not only one local system in some cases. It might be good to use a nested elements as
```
<local_coordinate_system>
    <local_coordinate_system>
        <name> x1  </name>
        <basis_vector_0>e0</basis_vector_0>
        <basis_vector_1>e1</basis_vector_1>
    </local_coordinate_system>
</local_coordinate_system>
```

endJunction: @JobstM Material group dependent coordinate systems are not a problem with a single \<local_coordinate_system\> tag.This is because the parameters for the base vectors can be group based. An (untested) example:
```xml
<local_coordinate_system>
    <basis_vector_0>e0</basis_vector_0>
    <basis_vector_1>e1</basis_vector_1>
</local_coordinate_system>
<parameters>
    <parameter>
        <name>e0</name>
        <type>Group</type>
        <group_id_property>MaterialIDs</group_id_property>
        <index_values>
            <index>0</index>
            <values>1 0 0</value> <!-- e0 for first material -->
        </index_values>
        <index_values>
            <index>1</index>
            <value>0.707 0.707 0</value> <!-- e0 for second material -->
         </index_values>
    </parameter>
    <parameter>
        <name>e1</name>
        <type>Group</type>
        <group_id_property>MaterialIDs</group_id_property>
        <index_values>
            <index>0</index>
            <values>0 1 0</value> <!-- e1 for first material -->
        </index_values>
        <index_values>
            <index>1</index>
            <value>-0.707 0.707 0</value> <!-- e1 for second material -->
         </index_values>
    </parameter>
</parameters>
```

ThieJan: More local_coordinate_system-tags are then only necesary if different parameters for the same material group have distinct principle directions? Or is there a way to represent this in the structure, too?

wenqing: I see. The current implementation can handle multi-local systems

@ThieJan Like this (figure from the Internet)
untitled

endJunction: That's correct, the multiple \<local_coordinate_system\> tags would only be necessary if, for example, temperature diffusion lives in a different coordinate system than the hydraulic permeability.

Suggest to wait until such a project comes up; the implementation is simple, just a name and instead of boolean \<use_local_coordinate_system\> a specific coordinate system would be referenced.

endJunction: Another possible input could be:
```xml
<local_coordinate_system>
    <basis_matrix>basis</basis_matrix>
</local_coordinate_system>
<parameters>
    <parameter>
        <name>basis</name>
        <type>Group</type>
        <group_id_property>MaterialIDs</group_id_property>
        <index_values>
            <index>0</index>
            <values>1 0 0 0 1 0 0 0 1</value> <!-- basis for first material -->
        </index_values>
        <index_values>
            <index>1</index>
            <values>0.707 0.707 0 -0.707 0.707 0 0 0 1</values> <!--  basis for second material -->
         </index_values>
    </parameter>
</parameters>
```

ProcessLib/Parameter/CoordinateSystem.cpp
endJunction: Checking the determinant of the transformation matrix to be close to 1. In debug build only, though.

wenqing: I see.

endJunction:
@TomFischer or @wenqing I've added a 3D test for anisotropy but had to extend the GWF. Could you please review the last two commits?

ProcessLib/Utils/ProcessUtils.cpp
```diff
@@ -105,4 +105,24 @@ std::vector<std::reference_wrapper<ProcessVariable>> findProcessVariables(
...
+    // Find corresponding parameter by name.
+    auto const it = std::find_if(
+        parameters.cbegin(), parameters.cend(),
+        [&parameter_name](std::unique_ptr<ParameterBase> const& p) {
+            return p->name == parameter_name;
+        });
+
+    if (it == parameters.end())
+    {
+        return nullptr;
```
TomFischer: Is this a usual case? Maybe add also some debug output to detect if there is an error?

ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
```diff
@@ -28,6 +28,34 @@ namespace GroundwaterFlow
 {
 const unsigned NUM_NODAL_DOF = 1;
 
+template <int GlobalDim>
+Eigen::Matrix<double, GlobalDim, GlobalDim> hydraulicConductivity(
```
wenqing: Nice. It can be  a common function for similar tensor parameters.

ProcessLib/Utils/ProcessUtils.cpp
endJunction: The nullptr is checked later by the caller, which issues a warning or whatever.

ProcessLib/GroundwaterFlow/Tests.cmake
```diff
@@ -757,3 +757,5 @@ AddTest(
     DIFF_DATA
     square_1x1_quad_1e5.vtu square_1e5_volumetricsourceterm_pcs_0_ts_1_t_1.000000.vtu analytical_solution pressure 0.75e-4 1e-16
 )
+
```
wenqing: if (NOT OGS_USE_MPI)
    OgsTest(PROJECTFILE "Elliptic/cube_1x1x1_GroundWaterFlow/cube_1e4_anisotropic.prj")
endif()

endJunction:
I've removed the split of findParameterOptional and findParameterByName, because there is a circular dependecy (on header files level) btw. the ProcessLib and MaterialLib. This is handled in a different PR.

PR #2371, opened 2019-02-19
Use exact predicates for orient2d

endJunction:
Review commit-wise possible.

Used in #2367 see https://github.com/endJunction/ogs/tree/master2raster branch with refactorings.


endJunction:
@TomFischer I'd think the it is also possible to implement GeoLib::parallel in terms of orient2D...

rinkk:
:+1:

PR #2374, opened 2019-02-21
Some improvements over the staggered scheme of TH

wenqing:
* Introduced two members of process IDs for the staggered scheme.
* Enabled more flexible process order within the staggered coupling.

### DIFF
--- a/ProcessLib/HT/HTProcess.h
+++ b/ProcessLib/HT/HTProcess.h
@@ -61,8 +61,9 @@ public:
         SecondaryVariableCollection&& secondary_variables,
         NumLib::NamedFunctionCaller&& named_function_caller,
         bool const use_monolithic_scheme,
-        std::unique_ptr<ProcessLib::SurfaceFluxData>&& surfaceflux);
-
+        std::unique_ptr<ProcessLib::SurfaceFluxData>&& surfaceflux,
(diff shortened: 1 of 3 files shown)


ProcessLib/HT/HTProcess.h
```diff
@@ -118,6 +119,9 @@ class HTProcess final : public Process
     std::array<std::unique_ptr<GlobalVector>, 2> _xs_previous_timestep;
 
     std::unique_ptr<ProcessLib::SurfaceFluxData> _surfaceflux;
+
+    const int _heat_transport_process_id;
+    const int _hydraulic_process_id;
```
endJunction: why not `static const int` or `constexpr int`?

wenqing: I just want to leave a possibility that the order of processes within the coupling can be optional, e.g. determined by input. Therefore these two members are initialized via ctor.

In the current implementation,  their values are given just after the variable names are parsed in CreateHTProcess. This  can also help to understand why _heat_transport_process_id=0 and _hydraulic_process_id=1.  I just added a comment in CreateHTProcess to highlight this purpose.

wenqing:
The phase-field related benchmarks failed.

PR #2380, opened 2019-02-24
[App/IO] Check gmsh return value.

endJunction:
The PR add testing of the return value of gmsh, which was silently omitted before. The check is related to the warnings about a return value of std::system call not being used.

No test case present. Needs checking by hand from data explorer.

### DIFF
--- a/Applications/FileIO/Legacy/createSurface.cpp
+++ b/Applications/FileIO/Legacy/createSurface.cpp
@@ -82,11 +82,12 @@ bool createSurface(GeoLib::Polyline const& ply,
     std::string gmsh_command =
         "gmsh -2 -algo meshadapt \"" + file_base_name + ".geo\"";
     gmsh_command += " -o \"" + file_base_name + ".msh\"";
-    // Temporarilly disable unused result warning
-    #pragma GCC diagnostic push
-    #pragma GCC diagnostic ignored "-Wunused-result"
-    std::system(gmsh_command.c_str());
-    #pragma GCC diagnostic pop
+    int const gmsh_return_value = std::system(gmsh_command.c_str());
+    if (gmsh_return_value != 0)
+    {
+        WARN("Call to '%s' returned non-zero value %d.", gmsh_command.c_str(),
+             gmsh_return_value);
+    }
     auto surface_mesh =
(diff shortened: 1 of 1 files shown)


rinkk:
It's not uncommon to import geometries with multiple surfaces, so it'd be helpful to include the name or ID of the currently meshed polygon in the warning.

endJunction:
@rinkk Please make suggestion.

The `ply_name` is "temporary_polyline_name", so this is not helpful.
Since the geo-objects are created in the scope of the function, the newly added polyline to be meshed, has always some constant id, so that's not helpful either.
There is not a single warning/info message having any of id/name information.

Probably I'm overlooking something, so a suggestion would be most welcome.

rinkk:
I see. Off the top of my head there are two things I can think of:

1. Adding a parameter to the method, explicitely giving the ID. When you call the method from, say, the Shape-interface, it's easy to determine the ID there.
2. Comparing the polyline-pointer with all other polyline-pointers in the given geometry. This shouldn't take too long given typical input geometries from previous projects (like, I don't think we ever had more than a few hundred polylines in a dataset. Usually it's a lot less than 100).

At the moment, the changes just add an OGS-error-message on top of the GMSH-error-message, which basically adds no useful information for the user. Knowing which Polygon, specifically, causes the problem, would help a lot with debugging the input data.

endJunction:
The required changes for the error message are out of the scope of this PR. Suggest to make another PR for better messages.

@rinkk OK?

rinkk:
We can do it that way, but I wonder: What is *this* PR actually good for? Specifically:

At the moment, GMSH is giving an error msg that something went wrong. This PR basically adds an error msg that says "GMSH said something went wrong." The original GMSH-msg is visible in the console so the newly added msg adds no information. The gmsh_command is always the same and the GMSH error code is already displayed in the GMSH error msg. So we could just as well *not* display an additional msg for less confusing information in the console? Maybe I'm missing something?

endJunction:
The PR add testing of the return value of gmsh, which was silently omitted before. The check is related to the warnings about a return value of std::system call not being used.

rinkk:
:fast_forward:

PR #2384, opened 2019-02-25
[eve] Updates for CentOS 7

### LINKED ISSUE #2383: Compilation problem on frontends of EVE: missing libSM.so
Error message
```
ninja: error: '/usr/lib64/libSM.so', needed by 'bin/testrunner', missing and no known rule to make it
```
On the two frontends libsm is not installed  in /usr/lib64.  There is one in  /usr/local/libsm/1.2.2-1/lib. A question: why testrunner needs libSM.so and also with a fixed path to it?

bilke:
Will re-enable envinf1 in this PR once it is back online.

Closes #2383.


wenqing:
ninja 1.9 error on eve:
```
ninja: /lib64/libc.so.6: version `GLIBC_2.15' not found (required by ninja)
```

bilke:
@wenqing envinf1 is still on CentOS 6. I reverted the module changes. See 1ddb61b.

wenqing:
@bilke  I see. Thanks.

PR #2386, opened 2019-02-26
[web] Fix download link

bilke:


bilke:
Nothing works, only deleting the cache on netlify:

https://www.netlify.com/docs/build-gotchas/#build-cache

PR #2389, opened 2019-02-27
Fix cppcheck warnings.

TomFischer:


Applications/FileIO/PetrelInterface.h
```diff
@@ -39,6 +39,11 @@ class PetrelInterface final
                     std::string &unique_model_name,
                     GeoLib::GEOObjects* obj);
 
+    PetrelInterface(PetrelInterface const& other) = delete;
+    PetrelInterface(PetrelInterface&& other) = delete;
+    PetrelInterface& operator=(PetrelInterface const&) = delete;
+    PetrelInterface& operator=(PetrelInterface&&) = delete;
```
endJunction: Just curious, why explicitly delete the ctor's/assignments?

Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
```diff
@@ -1101,7 +1101,7 @@ void NodeWiseMeshPartitioner::writeElementsASCII(
         for (const auto* node : partition.nodes)
         {
             nodes_local_ids[node->getID()] = node_local_id_offset;
-            node_local_id_offset++;
+            ++node_local_id_offset;
```
endJunction: What's the difference. Add a cppcheck message to the commit message, please. (Btw typo in the commit message)

Applications/FileIO/PetrelInterface.h
TomFischer: The class 'PetrelInterface' contains pointer to allocated memory. So a correct implementation would be necessary rather than a compiler generated. I thought it is better to mark ctor's/assignments as delete.

Applications/Utils/ModelPreparation/PartitionMesh/NodeWiseMeshPartitioner.cpp
TomFischer: The cppcheck message is:
*Prefix ++/-- operators should be preferred for non-primitive types. Pre-increment/decrement can be more efficient than post-increment/decrement. Post-increment/decrement usually involves keeping a copy of the previous value around and adds a little extra code.*

Typo is fixed.

endJunction: That's the reason, why clang-tidy is better here: `IntegerType` is a `long`. Doesn't matter.

PR #2390, opened 2019-02-28
Mesh on mesh mapping fix

rinkk:
Adjusts the mesh on mesh mapping tool to use the functionality introduced in #2367

### DIFF
--- a/Applications/Utils/MeshEdit/moveMeshNodes.cpp
+++ b/Applications/Utils/MeshEdit/moveMeshNodes.cpp
@@ -21,43 +21,26 @@
 #include "MeshLib/IO/writeMeshToFile.h"
 #include "MeshLib/Mesh.h"
 #include "MeshLib/Node.h"
+#include "MeshLib/MeshEditing/ProjectPointOnMesh.h"
 
-int find_closest_point(MeshLib::Node const*const point, std::vector<MeshLib::Node*> const& nodes, double const& max_dist)
+bool setClosestPointElevation(MeshLib::Node& p,
+                              std::vector<MeshLib::Node*> const& nodes,
+                              double const& max_dist)
 {
-    const std::size_t nNodes (nodes.size());
-    double sqr_shortest_dist (max_dist*2);
-    int idx = (sqr_shortest_dist<max_dist) ? 0 : -1;
-    const MeshLib::Node p (*point);
-
-    for (unsigned i=0; i<nNodes; i++)
+    double sqr_shortest_dist (max_dist);
+    bool pnt_found (false);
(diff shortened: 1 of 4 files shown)


Applications/Utils/MeshEdit/moveMeshNodes.cpp
```diff
@@ -151,7 +134,7 @@ int main (int argc, char* argv[])
     // maps the elevation of mesh nodes according to a ground truth mesh whenever nodes exist within max_dist
     if (current_key == "-MESH")
     {
-        if (argc < 5)
+        if (argc < 4)
```
endJunction: For another PR; this cli arguments handling could be upgraded to tclap.

MeshLib/MeshEditing/ProjectPointOnMesh.cpp
```diff
@@ -0,0 +1,75 @@
...
+#include "MeshLib/Elements/Tri.h"
+#include "MeshLib/MeshSearch/MeshElementGrid.h"
+
+namespace MeshLib
+{
+namespace ProjectPointOnMesh
+{
+
+/// Returns the element in which the given node is located when
+/// projected onto a mesh, or nullptr if no such element was found.
```
endJunction: Doxygen hasn't finished yet, but the function documentation appears twice in the code, once here, once in the header file; it could be that it will show up twice in the doxygen's function documentation. Needs a quick check, when doxygen is done.

Applications/Utils/MeshEdit/moveMeshNodes.cpp
rinkk: I know. However, this would require more time than I can currently spend on this. It's on the to-do list.

MeshLib/MeshEditing/ProjectPointOnMesh.cpp
rinkk: Removed the duplicate comments.

rinkk:
The change to the algorithm is really simple: The utility used to assign the elevation of the closest point in the ground truth mesh to all nodes in the mesh to be mapped. Now, it using the actual elevation at the location of each node (that's the functionality introduced in the mesh2raster tool) and only if no point is found, it checks for the closest point within a user-specified radius (which is useful if boundaries are slightly translated due to different precision used in preprocessing tools).

endJunction:
@rinkk This needs a fix for an unused 'offset' variable in moveMeshNodes.cpp:144
https://jenkins.opengeosys.org/job/ufz/job/ogs/job/PR-2390/3/clang-mac/

PR #2397, opened 2019-03-04
[GeoLib] cppcheck warnings cleanup

TomFischer:


GeoLib/DuplicateGeometry.cpp
```diff
@@ -23,7 +23,7 @@ namespace GeoLib
 {
 DuplicateGeometry::DuplicateGeometry(GeoLib::GEOObjects& geo_objects,
                                      std::string const& input_name,
-                                     std::string output_name)
```
endJunction: creation scenario: (creation with two moves) vs. (creation, pass-by-ref, and copy) => one copy less for value.
existing var scenario: (copy to ctor arg, move) vs (pass-by-ref, copy) => equal nr. of copies.

Tests/MeshLib/TestMeshValidation.cpp
```diff
@@ -46,9 +46,9 @@ detectHoles(MeshLib::Mesh const& mesh,
 TEST(MeshValidation, DetectHolesTri)
 {
     std::array<double, 12> pix = {{0,0.1,0.2,0.1,0,0,0.1,0,0,0,-0.1,0}};
-    GeoLib::RasterHeader const header = {
+    GeoLib::RasterHeader header = {
         4, 3, 1, MathLib::Point3d(std::array<double, 3>{{0, 0, 0}}), 1, -9999};
-    GeoLib::Raster const raster(header ,pix.begin(), pix.end());
+    GeoLib::Raster const raster(std::move(header), pix.begin(), pix.end());
```
endJunction: I'd agree here if RasterHeader would have a move ctor. But it is a simple aggregate, which have to be copied anyway.

GeoLib/StationBorehole.cpp
```diff
@@ -55,7 +55,7 @@ int StationBorehole::find(const std::string &str)
     std::size_t size = _soilName.size();
     for (std::size_t i = 0; i < size; i++)
     {
-        if (_soilName[i].find(str) == 0)
+        if (_soilName[i].compare(str) == 0)
```
endJunction: Isn't it simply same as `_soilName[i] == str`?
And then the whole StationBorehole::find() is simply a
```c++
std::find_first_of(begin(_soilName), end(_soilName), str) != _soilName.end()
```

PR #2400, opened 2019-03-05
[DataExplorer] Write gmsh geo format using fixed mesh density.

TomFischer:


TomFischer:
@rinkk Please could you check if this is the expected output?

waltherm:
Building does not finish for the MAC setup; @bilke can you have a look?

endJunction:
@waltherm The particular test fails on that particular mac machine. Wenqing is looking into the details.
Can be merged, when desired behaviour is confirmed. Though a test case would be good addition.

waltherm:
I cannot run the DataExplorer:
`./DataExplorer: error while loading shared libraries: libOGSFileConverterLib.so: cannot open shared object file: No such file or directory
`

TomFischer:
@waltherm I rebased just now onto the actual master where the problem:
```./DataExplorer: error while loading shared libraries: libOGSFileConverterLib.so: cannot open shared object file: No such file or directory```
should be fixed.  When jenkins has finished the new build its ready for the test.

waltherm:
I cannot write a gmsh (.geo) file. Enclosed is the log. And a test file (.shp).

[log.txt](https://github.com/ufz/ogs/files/2940847/log.txt)
[merged_simplify_reprojected_study_area.tar.gz](https://github.com/ufz/ogs/files/2940850/merged_simplify_reprojected_study_area.tar.gz)

TomFischer:
From the log I cannot see what is wrong. What steps did you do?

waltherm:
Strange... today it worked. At least saving the file. But I still have all the additional points saved in the .geo (gmsh) file. (File -> Save as -> .geo).

TomFischer:
Did you use the DataExplorer build in this pull request?

waltherm:
Sorry, @TomFischer, I really thought to have used this version last week. I tried again and it works: no extra points when importing a shp and exporting this as a gmsh file. Thanks!

:+1:

PR #2401, opened 2019-03-05
[ExtractSurface] Enable switch for binary or ascii output. 

TomFischer:


rinkk:
Keep in mind that writeMeshToFile() handles both *.msh and *.vtu files. A vtk-based compression parameter makes no sense here.

Instead, I would keep writeMeshToFile() as it is, remove the call in the utility and replace it with

```
MeshLib::IO::VtuInterface vtkIO(&_mesh, mode, compression);
vtkIO.writeToFile(file_name.toStdString().c_str());
```

where compression is either "true" or "false" and mode is either "vtkXMLWriter::Appended" or "vtkXMLWriter::Ascii", respectively.

TomFischer:
@rinkk 
> *A vtk-based compression parameter makes no sense here.*

Reverted this change.

rinkk:
:+1:

PR #2409, opened 2019-03-07
Open boundary implementation for HC process

jbathmann:
This implementation of a open boundary for the HC-Process is necessary, since a non-advective form of the governing equations is implemented. 

PR following a discussion with @renchao-lu.


ProcessLib/BoundaryCondition/HCOpenBoundaryCondition.cpp
```diff
@@ -0,0 +1,68 @@
...
+    unsigned const global_dim, Process const& process,
+    unsigned const shapefunction_order)
+{
+    DBUG(
+        "Constructing open boundary for Component Transport process from "
+        "config.");
+    //    //! \ogs_file_param{prj__processes__process__type}
+    //    config.checkConfigParameter("type", "ComponentTransport");
+    //    How can I check for process type == Component Transport? The method
+    //    above does not work.
```
endJunction: At this point you cannot. The process variables do not belong to a particular process. (A process uses some process variables, though.)

ProcessLib/BoundaryCondition/HCOpenBoundaryConditionLocalAssembler.h
```diff
@@ -0,0 +1,117 @@
...
+    {
+    }
+
+    void assemble(std::size_t const mesh_item_id,
+                  NumLib::LocalToGlobalIndexMap const& dof_table_boundary,
+                  double const t, const GlobalVector& x, GlobalMatrix& /*K*/,
+                  GlobalVector& b, GlobalMatrix* /*Jac*/) override
+    {
+        NodalVectorType _local_rhs(_local_matrix_size);
+        _local_rhs.setZero();
```
endJunction: NodalVectorType _local_rhs = NodalVectorType::Zero(_local_size);

ProcessLib/BoundaryCondition/HCOpenBoundaryConditionLocalAssembler.h
```diff
@@ -0,0 +1,117 @@
...
+        // Get element nodes for the interpolation from nodes to
+        // integration point.
+        NodalVectorType const boundary_permeability_node_values =
+            _data.boundary_permeability.getNodalValuesOnElement(Base::_element,
+                                                                t);
+        auto surface_element_normal =
+            MeshLib::FaceRule::getSurfaceNormal(&(Base::_element));
```
endJunction: Move to ctor; store Eigen::Vector3d.

ProcessLib/BoundaryCondition/HCOpenBoundaryConditionLocalAssembler.h
```diff
@@ -0,0 +1,117 @@
...
+        unsigned const n_integration_points =
+            Base::_integration_method.getNumberOfPoints();
+        auto const indices =
+            NumLib::getIndices(mesh_item_id, dof_table_boundary);
+
+        std::vector<double> const local_values = x.get(indices);
+
+        std::size_t bulk_element_id =
+            _data.bulk_element_ids[Base::_element.getID()];
+        std::size_t bulk_face_id = _data.bulk_face_ids[Base::_element.getID()];
```
endJunction: Also store as member; init in ctor.

renchao-lu:
I ran the benchmark of tracer test with the attached 3D geometrical model. It works, though the computation cost is somewhat high. Note that boundary condition of this type is quite common in component transport simulation.

Besides, I failed in testing with 2D geometrical model. I guess the failure may be caused by improper use of ExtractSurface Tool. Could you please instruct me when you are free. @jbathmann  

I checked local assember in ComponentTransportFEM.h again. We only need to change assembling of the convection term in KCC from `-dNdx.transpose() * velocity * density * N` to `N.transpose() * velocity.transpose() * density * dNdx`. I see this change will cause conflicts against your previous implementation. These are my two cents.

Tests/Data/Parabolic/ComponentTransport/SimpleSynthetics/surfaceflux_component-transport_cube_1e3.prj
```diff
@@ -2,7 +2,8 @@
 <OpenGeoSysProject>
     <meshes>
         <mesh>cube_1x1x1_hex_1e3.vtu</mesh>
-        <mesh>cube_1x1x1_hex_1e3_left.vtu</mesh>
+	<mesh>cube_1x1x1_hex_1e3_left.vtu</mesh>
+	<mesh>cube_1x1x1_hex_1e3_right.vtu</mesh>
```
renchao-lu: Small tips: duplicate mesh files are given.

jbathmann: changed.

ProcessLib/BoundaryCondition/HCOpenBoundaryCondition.cpp
jbathmann: Ok, so I drop it here.

ProcessLib/BoundaryCondition/HCOpenBoundaryConditionLocalAssembler.h
jbathmann: I am not sure about the "prettiness" of my updated solution... (?)

jbathmann:
> Besides, I failed in testing with 2D geometrical model. I guess the failure may be caused by improper use of ExtractSurface Tool. Could you please instruct me when you are free. @jbathmann
> 
As far as I know the ExtractSurface tool only works for 3D geometrical setups. I am not sure whether the surface flux can be calculated for 2D geometrical setups. @TomFischer are there any examples/benchmarks for surface flux calculation in 2D setups?

> I checked local assember in ComponentTransportFEM.h again. We only need to change assembling of the convection term in KCC from `-dNdx.transpose() * velocity * density * N` to `N.transpose() * velocity.transpose() * density * dNdx`. I see this change will cause conflicts against your previous implementation. These are my two cents.

IMHO we should maybe try to get more opinions on this issue of which formulation we choose. In any case, it might appears useful to me to choose the same form (advective or non-advective formulation) for both HC and HT process. However, before changing the process implementation, I think we need to update the deviation of the equation system (pdf-file).

renchao-lu:
> > Besides, I failed in testing with 2D geometrical model. I guess the failure may be caused by improper use of ExtractSurface Tool. Could you please instruct me when you are free. @jbathmann
> 
> As far as I know the ExtractSurface tool only works for 3D geometrical setups. I am not sure whether the surface flux can be calculated for 2D geometrical setups. @TomFischer are there any examples/benchmarks for surface flux calculation in 2D setups?
> I remembered some changes on calculation of normal vector are needed for running 2D case. I will fix it.

endJunction:
@jbathmann The parameters were moved in ParameterLib. I've updated the code after rebase.

jbathmann:
This PR is outdated by PR #2466 and can be removed.

PR #2411, opened 2019-03-08
SurfaceFlux calculates now specific flux

TomFischer:


ProcessLib/SurfaceFlux/SurfaceFluxData.h
```diff
@@ -68,8 +66,7 @@ struct SurfaceFluxData
 
         if (mesh_name.empty())
         {
-            return std::unique_ptr<ProcessLib::SurfaceFluxData>(
-                nullptr);
+            return std::unique_ptr<ProcessLib::SurfaceFluxData>(nullptr);
```
endJunction: Optional: Could be simply `return nullptr`

PR #2418, opened 2019-03-13
[PL] TM: Add NodalForces and HeatFlux.

endJunction:
Almost verbatim copy from HydroMechanics.
If you have a better name for "thermal forces" I'd be happy to improve it.

Waiting for @fparisio for validation.


vowinckel:
Quote from @endJunction: "If you have a better name for "thermal forces" I'd be happy to improve it."
How about "Thermal Heating"?

endJunction:
In some discussions the term "heat flux" was highly recommended.
I'm currently leaning towards the "heat flux" too.

endJunction:
@fparisio Thanks for testing.

Issue #2379, opened 2019-02-23
DataExplorer building error: netcdfcpp.h not found


endJunction:
Detection of where the netcdfcpp library is installed is not accurate.

```
Applications/DataExplorer/NetCdfDialog/NetCdfConfigureDialog.h:12:10: \
    fatal error: 'netcdfcpp.h' file not found
#include <netcdfcpp.h>
```
```
locate netcdfcpp.h
/usr/include/paraview-5.6/vtk_netcdfcpp.h
/usr/include/paraview-5.6/vtknetcdfcpp/netcdfcpp.h
```
but the build command (for NetCdfConfigureDialog.cpp) has
```
-isystem /usr/include/vtk/vtknetcdfcpp
```

Archlinux; VTK-8.2.0-1 (with netcdf dependency); paraview-5.6.0-2.

I'd think it is better to use the netcdf-cxx library directly, rather then the vtk/paraview internal wrapper.

endJunction:
started something here: https://github.com/endJunction/ogs/tree/NetCDF

bilke:
`netcdfcpp.h` does not come from the VTK interface but is an old implementation of a C++ interface:

https://github.com/Unidata/netcdf-c/tree/netcdf-4.1.3/cxx

Nowadays this one is used which has a slightly different interface:

https://github.com/Unidata/netcdf-cxx4

I prepared Conan packages for everything in #2450 so we should work on the interface migration once this is merged. @rinkk @endJunction

bilke:
@endJunction I think this can be closed as #2654 is merged, right?

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/2379).

PR #2427, opened 2019-03-15
Make NetCDF optional.

endJunction:
Adds an `OGS_USE_NETCDF` option with default being ON => no changes to current setup.
The NetCDF related code is wrapped in if conditions.

Background: The NetCDF library interface provided by paraview/vtk is used now, not the actual netcdf-cxx library. On some systems paraview/vtk installation does not provide the header files for the internally used interface. This leads to compilation problems of the DE. On the long run the netcdf-cxx library should be used directly, not the paraview/vtk interface to that library.

Closes https://github.com/ufz/ogs/issues/2379

### DIFF
--- a/Applications/DataExplorer/VtkVis/VtkCompositeTextureOnSurfaceFilter.cpp
+++ b/Applications/DataExplorer/VtkVis/VtkCompositeTextureOnSurfaceFilter.cpp
@@ -23,7 +23,9 @@
 #include <vtkUnstructuredGrid.h>
 #include "VtkGeoImageSource.h"
 #include "VtkRaster.h"
+#ifdef OGS_USE_NETCDF
 #include "NetCdfDialog/NetCdfConfigureDialog.h"
+#endif  // OGS_USE_NETCDF
 
 #include <QFileDialog>
 #include <QFileInfo>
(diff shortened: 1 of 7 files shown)


Applications/DataExplorer/VtkVis/VtkCompositeTextureOnSurfaceFilter.cpp
```diff
@@ -61,9 +63,14 @@ void VtkCompositeTextureOnSurfaceFilter::init()
...
-    QString fileName = QFileDialog::getOpenFileName(parent, "Select raster file to apply as texture",
-                                                    settings.value("lastOpenedTextureFileDirectory").toString(),
-                                                    "Raster files (*.asc *.grd *.bmp *.jpg *.png *.tif);;NetCDF files (*.nc);;");
+    QString fileName = QFileDialog::getOpenFileName(
+        parent, "Select raster file to apply as texture",
+        settings.value("lastOpenedTextureFileDirectory").toString(),
+        "Raster files (*.asc *.grd *.bmp *.jpg *.png *.tif);;"
+#ifdef OGS_USE_NETCDF
+        "NetCDF files "
+#endif  // OGS_USE_NETCDF
```
bilke: Should the `#endif` moved one line below?

CMakeLists.txt
```diff
@@ -95,6 +95,10 @@ option(OGS_BUILD_GUI "Should the Data Explorer be built?" OFF)
 if(OGS_BUILD_GUI)
     add_definitions(-DOGS_BUILD_GUI)
 endif()
+option(OGS_USE_NETCDF "Add NetCDF support." ON)
+if(OGS_USE_NETCDF)
+    add_definitions(-DOGS_USE_NETCDF)
```
bilke: Could also be moved to the Data Explorer CMakeLists.txt to narrow scope as we discussed today.

Applications/DataExplorer/VtkVis/VtkCompositeTextureOnSurfaceFilter.cpp
endJunction: Thanks!

CMakeLists.txt
endJunction: Also moved the `option(OGS_USE_NETCDF ...)` into `OGS_BUILD_GUI` condition and tested setting the variable OFF and then enabling the GUI; it works as anticipated in the discussion.

PR #2436, opened 2019-03-20
Expanded shape output to write scalar arrays to shape attribute table

rinkk:
* works only for 2D meshes
* works only for int/double arrays
* works only for cell-arrays (because the shape format can only assign value to shape-entities, i.e. polygons but not points ... this can be easily fixed via VTK point 2 cell, though)

### DIFF
--- a/Applications/FileIO/SHPInterface.cpp
+++ b/Applications/FileIO/SHPInterface.cpp
@@ -213,87 +213,110 @@ void SHPInterface::readPolygons(const SHPHandle& hSHP, int numberOfElements,
 
 bool SHPInterface::write2dMeshToSHP(const std::string &file_name, const MeshLib::Mesh &mesh)
 {
-    if (mesh.getDimension()!=2)
+    if (mesh.getDimension() != 2)
     {
-        ERR ("SHPInterface::write2dMeshToSHP(): Mesh to Shape conversion is only working for 2D Meshes.");
+        ERR("SHPInterface::write2dMeshToSHP(): Mesh to Shape conversion is "
+            "only working for 2D Meshes.");
         return false;
     }
 
-    unsigned nElements (mesh.getNumberOfElements());
-    if (nElements<1)
+    std::size_t const n_elements = mesh.getNumberOfElements();
+    if (n_elements < 1)
     {
-        ERR ("SHPInterface::write2dMeshToSHP(): Mesh contains no elements.");
+        ERR("SHPInterface::write2dMeshToSHP(): Mesh contains no elements.");
         return false;
     }
 
-    if (nElements>10E+7) // DBF-export requires a limit, 10 mio seems good for now
+    // DBF-export requires a limit of records because of the limits to the
+    // integer values. The exponent controls maximum number of elements and
+    // the maximum number of digits written in the DBF file.
+    std::size_t const max_exp = 8;
+    if (n_elements >= std::pow(10,max_exp))
     {
-        ERR ("SHPInterface::write2dMeshToSHP(): Mesh contains too many elements for currently implemented DBF-boundaries.");
+        ERR("SHPInterface::write2dMeshToSHP(): Mesh contains too many elements "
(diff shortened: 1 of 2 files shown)


Applications/FileIO/SHPInterface.cpp
```diff
@@ -213,87 +213,110 @@ void SHPInterface::readPolygons(const SHPHandle& hSHP, int numberOfElements,
...
+    if (n_elements < 1)
     {
-        ERR ("SHPInterface::write2dMeshToSHP(): Mesh contains no elements.");
+        ERR("SHPInterface::write2dMeshToSHP(): Mesh contains no elements.");
         return false;
     }
 
-    if (nElements>10E+7) // DBF-export requires a limit, 10 mio seems good for now
+    // DBF-export requires a limit of records, 10 mio elements seems good for now
+    std::size_t const max_exp = 8;
```
endJunction: minor: better not to mention code constants in the comment, because the comment diverges very fast from the actual code. I'd write combine the above comment with the latter one:
```c++
// DBF-export requires a limit of records because of the limits to the integer values.
// The exponent controls maximum number of elements and the maximum number of digits written in the DBF file.
```
Adjust accordingly.

Applications/FileIO/SHPInterface.cpp
```diff
@@ -213,87 +213,110 @@ void SHPInterface::readPolygons(const SHPHandle& hSHP, int numberOfElements,
...
-    double* padfX;
-    double* padfY;
-    double* padfZ;
-    for (unsigned i=0; i<nElements; ++i)
-    {
-        const MeshLib::Element* e (mesh.getElement(i));
+    // allows integers of length "8", i.e. 10mio-1 elements
+    int const elem_id_field = DBFAddField(hDBF, "Elem_ID", FTInteger, max_exp, 0);
 
+    //
```
endJunction: ?

endJunction:
:fast_forward: when green.

Applications/FileIO/SHPInterface.cpp
```diff
@@ -213,87 +213,109 @@ void SHPInterface::readPolygons(const SHPHandle& hSHP, int numberOfElements,
...
+    }
+    SHPClose(hSHP);
+
+    // Write scalar arrays to database file
+    MeshLib::Properties const& properties = mesh.getProperties();
+    std::vector<std::string> const& array_names =
+        properties.getPropertyVectorNames(MeshLib::MeshItemType::Cell);
+    int const n_recs = DBFGetRecordCount(hDBF);
+
+    for (std::string name : array_names)
```
TomFischer: Why not `for (std::string const& name : array_names)`?

Applications/FileIO/SHPInterface.cpp
```diff
@@ -213,87 +213,109 @@ void SHPInterface::readPolygons(const SHPHandle& hSHP, int numberOfElements,
...
-
+    INFO("Shape export of 2D mesh '%s' finished.", mesh.getName().c_str());
     return true;
 }
 
+SHPObject* SHPInterface::createShapeObject(MeshLib::Element const& e,
+                                           std::size_t const shp_record)
+{
+    unsigned const nNodes(e.getNumberOfBaseNodes());
+    double* padfX = new double[nNodes + 1];
```
TomFischer: New memory allocation - who deletes it?

TomFischer: Just now I saw the removed comment (previous line 287) that says that shapelib crashes when deleting the memory. Please, insert the comment somewhere in the method.

rinkk:
@TomFischer - Re: Mem Allocation: 

* padfX/Y/Z are parameters handed over to SHPCreateObject and are afterwards handled by said object (as you would assume, I guess)
* the object itself (to which the old comment referred) is now correctly handled after the changes by calling SHPDestroyObject

So I guess no additional comment is necessary.

PR #2444, opened 2019-03-29
re-inserting previously deleted method

rinkk:
Also added call to method in test so it doesn't get deleted again.

### DIFF
--- a/Applications/FileIO/CsvInterface.cpp
+++ b/Applications/FileIO/CsvInterface.cpp
@@ -22,6 +22,39 @@
 namespace FileIO {
 CsvInterface::CsvInterface() = default;
 
+std::vector<std::string> CsvInterface::getColumnNames(std::string const& fname,
+                                                      char const delim)
+{
+    std::ifstream in(fname.c_str());
+
+    if (!in.is_open())
+    {
+        ERR("CsvInterface::readPoints(): Could not open file %s.",
+            fname.c_str());
+        return std::vector<std::string>();
+    }
+    std::string line;
+    if (!getline(in, line))
+        return {};
+
+    std::list<std::string> fields;
+    if (delim != '\n')
+        fields = BaseLib::splitString(line, delim);
+
+    if (fields.size() < 2)
+    {
+        for (char const d : {'\t', ';', ','})
+        {
+            fields = BaseLib::splitString(line, d);
+            if (fields.size() > 1)
+            {
+                break;
+            }
+        }
+    }
+    return {begin(fields), end(fields)};
+}
+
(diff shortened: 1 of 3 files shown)


Applications/FileIO/CsvInterface.cpp
```diff
@@ -22,6 +22,44 @@
 namespace FileIO {
 CsvInterface::CsvInterface() = default;
 
+std::vector<std::string> CsvInterface::getColumnNames(std::string const& fname,
+                                                      char delim)
```
endJunction:
```suggestion
                                                      char const delim)
```

Applications/FileIO/CsvInterface.cpp
```diff
@@ -22,6 +22,44 @@
...
+
+    if (!in.is_open())
+    {
+        ERR("CsvInterface::readPoints(): Could not open file %s.",
+            fname.c_str());
+        return std::vector<std::string>();
+    }
+    std::string line;
+    if (!getline(in, line))
+        return std::vector<std::string>();
```
endJunction:
```suggestion
        return {};
```

Applications/FileIO/CsvInterface.cpp
```diff
@@ -22,6 +22,44 @@
...
+
+    std::list<std::string> fields;
+    if (delim != '\n')
+        fields = BaseLib::splitString(line, delim);
+
+    if (fields.size() < 2)
+    {
+        std::array<char, 3> delims = {'\t', ';', ','};
+        for (char d : delims)
```
endJunction:
```suggestion
        for (char const d : {'\t', ';', ','})
```

Applications/FileIO/CsvInterface.cpp
```diff
@@ -22,6 +22,44 @@
...
+        {
+            fields = BaseLib::splitString(line, d);
+            if (fields.size() > 1)
+            {
+                delim = d;
```
endJunction: Not needed.
The `delim` variable is not used after this assignment.

Applications/FileIO/CsvInterface.cpp
```diff
@@ -22,6 +22,44 @@
...
+                break;
+            }
+        }
+    }
+    std::vector<std::string> names;
+    names.reserve(names.size());
+    std::copy(fields.begin(), fields.end(), std::back_inserter(names));
+    return names;
```
endJunction:
```suggestion
    return {begin(fields), end(fields)};
```
no need to do all the work by hand.

endJunction:
The commit message could be more meaningful, like "Implement getColumnNames in csv IF."...

PR #2432, opened 2019-03-19
Use QVtkOpenGLWidget

endJunction:
The QVTKWidget is deprecated in VTK-8.1.

Fixes https://github.com/ufz/ogs/issues/2354

PR #2448, opened 2019-04-01
[DE] Fixed 3D update on item removal

bilke:
Probably since switching to the new QVTKWidget (#2432) the 3D view was not updated automatically upon item removal (e.g. mesh removal). I added a new signal/slot connect to fix this. I have not investigated why it worked before...


rinkk:
:+1:

PR #2449, opened 2019-04-02
Add Exponential property to MPL

TomFischer:
PR adds the property of exponential type.
 - A unit test covering the feature is included.
 - The feature is used in a benchmark.
 - The project file tags are documented.


MaterialLib/MPL/Properties/ExponentialProperty.cpp
```diff
@@ -0,0 +1,69 @@
...
+}
+
+PropertyDataType ExponentialProperty::d2Value(
+    VariableArray const& variable_array,
+    Variable const pv1,
+    Variable const pv2) const
+{
+    return _exponent_data.type == pv1 && _exponent_data.type == pv2
+               ? boost::get<double>(_value) *
+                     std::pow(boost::get<double>(_exponent_data.factor), 2) *
```
endJunction: Don't use `std::pow` for integer powers---it's slow. Use `boost::math::pow<2>(x)` in this case.

TomFischer: Did not know that the integer overloads of `std::pow` are gone and the ints were casted to floating point values. Will change it as suggested.

endJunction: I remember the optimization when I was refactoring Ehlers material model functions. Using templated boost version was much faster.
Now I tested some simple code and for std::pow(x,2) the assembly is the same as for x*x, but not for std::pow(x, 3) and x*x*x. For 3rd power the std::pow is ca. 2 times slower. http://quick-bench.com/smUihBRSJUMUXgBwKhEP_pDE14w

The overloads are there, but I'm not sure, why it is inlined in one case, but not in another.

PR #2454, opened 2019-04-07
Add ascii switch to VTK2OGS and createLayeredMeshFromRaster

TomFischer:
Feature request from Erik.


norihiro-w:
Just FYI:
When I tested it before, VTK ascii output wrote real data in fixed digit format, not in floating point expression. If it's still valid behavior in the latest and accuracy is important, one should use binary output.

TomFischer:
@norihiro-w : Thanks for pointing this out. I added some information to the command line help.

TomFischer:
Merged @endJunction suggestion.

PR #2457, opened 2019-04-09
[web] Added instructions on prj-file checking with xmllint

bilke:
As requested by @ErikNixdorf.

I also found out that xmllint provides some schema checking with [RELAX NG](https://relaxng.org) schema. This maybe worth a look...


endJunction:
@bilke RelaxNG isn't sufficient for our xml files, we would need something like [schematron](https://en.wikipedia.org/wiki/Schematron).

PR #2460, opened 2019-04-09
[DE] Added support for world files to georeference raster files

rinkk:
Image reader tries to find a world file when a raster is loaded to georeference the data correctly.

If libgeotiff is not found, vtkTIFFReader is now used instead of just giving an error msg. (vtkTiffReader is not great and has issues with lots of TIFF-variants so it's really just a fallback in case there is no alternative available.)


Applications/DataExplorer/VtkVis/VtkRaster.cpp
```diff
@@ -244,22 +252,101 @@ vtkImageReader2* VtkRaster::loadImageFromFile(const std::string &fileName)
...
+    std::string const no_ext = BaseLib::dropFileExtension(filename);
+
+    std::vector<std::string> const supported_extensions =
+    { ".pgw", ".pngw", ".pgwx",
+      ".jgw", ".jpgw", ".jgwx",
+      ".tfw", ".tifw", ".tfwx",
+      ".bpw", ".bmpw", ".bpwx",
+      ".wld" };
+
+    for (std::string ext : supported_extensions)
```
endJunction:
```suggestion
    for (std::string const& ext : supported_extensions)
```

PR #2461, opened 2019-04-09
[DE] Re-implementing strict double validation

rinkk:
Previous implementation didn't work with small value ranges for some reason (such as GMSH preferences where some params need to be in [0;1]).

This fixes an issue with unchangable parameter values in the GMSH dialog.
Also adjusts the minimum default value for parameters to something more reasonable.

(No entry to the changelog because this was supposed to work anyway. I *think* it *did* work in the past, possibly something was changed in QDoubleValidator?)

### DIFF
--- a/Applications/DataExplorer/DataView/GMSHPrefsDialog.cpp
+++ b/Applications/DataExplorer/DataView/GMSHPrefsDialog.cpp
@@ -43,11 +43,11 @@ GMSHPrefsDialog::GMSHPrefsDialog(GeoLib::GEOObjects const& geoObjects, QDialog* 
     param1->setValidator (max_number_of_points_in_quadtree_leaf_validator);
     // object will be deleted by Qt
     auto* mesh_density_scaling_pnts_validator(
-        new StrictDoubleValidator(1e-10, 1.0, 5, this->param2));
+        new StrictDoubleValidator(0, 1, 5, this->param2));
     param2->setValidator (mesh_density_scaling_pnts_validator);
     // object will be deleted by Qt#
     auto* mesh_density_scaling_stations_validator(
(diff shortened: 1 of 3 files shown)


Applications/DataExplorer/DataView/GMSHPrefsDialog.cpp
```diff
@@ -147,16 +147,17 @@ void GMSHPrefsDialog::accept()
...
+        double const min_scaling_factor (1e-10);
         max_number_of_points_in_quadtree_leaf = BaseLib::str2number<unsigned> (
                 param1->text().toStdString().c_str());
         if (max_number_of_points_in_quadtree_leaf == 0)
             max_number_of_points_in_quadtree_leaf = 10;
         mesh_density_scaling_pnts = fabs (param2->text().toDouble());
-        if (mesh_density_scaling_pnts < sqrt(std::numeric_limits<double>::epsilon()))
-            mesh_density_scaling_pnts = 0.5;
+        if (mesh_density_scaling_pnts < min_scaling_factor)
+            mesh_density_scaling_pnts = min_scaling_factor;
```
TomFischer: In the previous version the value was 10 x larger than the value for `mesh_density_scaling_stations`. Now, `mesh_density_scaling_pnts` equals `mesh_density_scaling_stations`. Is this change intended?

rinkk: Yes, it is intended. The smallest possible number allowed by the DoubleValidator is "0", but the algorithm requires a small positive value. So instead of setting it to the (relatively large) default value, it is now set to the previously allowed minimum value of 1^{-10}.

PR #2469, opened 2019-04-16
LIE Mohr-Coulomb update preparations

endJunction:
Zero net change with a notable exception of forgotten push back call in LIE/HM model, which is required for the upcoming Mohr-Coulomb fracture model update.

Upcoming changes: https://github.com/endJunction/ogs/tree/LIEMohrCoulombUpdate


MaterialLib/SolidModels/CreateCreepBGRa.cpp
```diff
@@ -62,8 +62,11 @@ createCreepBGRa(
         //! \ogs_file_param_special{material__solid__constitutive_relation__CreepBGRa__q}
         config, "q", parameters, 1);
 
+    auto const& nonlinear_solver_config =
+        //! \ogs_file_param{material__fracture_model__constitutive_relation__CreepBGRa__nonlinear_solver}
```
TomFischer: Did not find the documentation (also for other cases, for instance Ehlers). Since the the path to the documentation is not similar to the documentation path given in Ehlers case (and others) I wonder if the path is correct.

PR #2470, opened 2019-04-17
[DE] Renamed Msh to Mesh in filenames

rinkk:
*No actual changes, just renaming files and classes starting with "Msh..." to "Mesh..." to be consistent with the rest of OGS. Also removed unused files.*


Applications/DataExplorer/DataView/MshEditDialog.h
```diff
@@ -1,83 +0,0 @@
-/**
```
TomFischer: Is it correct to remove the mesh edit dialog?

Applications/DataExplorer/DataView/MeshView.cpp
```diff
@@ -38,34 +38,34 @@
...
 
-MshView::~MshView() = default;
+MeshView::~MeshView() = default;
 
-void MshView::updateView()
+void MeshView::updateView()
 {
     setAlternatingRowColors(true);
-    setColumnWidth(0,125);
+    setColumnWidth(0,150);
```
TomFischer: Is this size change intended?

Applications/DataExplorer/DataView/MeshTabWidgetBase.ui
```diff
@@ -1,20 +1,29 @@
...
- <widget class="QWidget" name="MshTabWidgetBase">
+ <class>MeshTabWidgetBase</class>
+ <widget class="QWidget" name="MeshTabWidgetBase">
   <property name="geometry">
    <rect>
     <x>0</x>
     <y>0</y>
     <width>278</width>
-    <height>500</height>
+    <height>511</height>
```
TomFischer: Is this size change intended?

Applications/DataExplorer/DataView/MeshView.cpp
rinkk: :-/

It's from a different branch. Not sure how it slipped in there.  
It doesn't matter, though...

Applications/DataExplorer/DataView/MeshTabWidgetBase.ui
rinkk: No idea how this happened. Some Qt-Designer magic?  
It doesn't really matter because the actual size is derived from the Data Explorer window. I'll change it back anyway...

PR #2471, opened 2019-04-17
[PL] Don't output bulk mesh if not specified.

endJunction:
Only if the output meshes list is empty, write it.

As requested by @ErikNixdorf 

1. [ ] ~Feature description was added to the [changelog](https://github.com/ufz/ogs/wiki/Release-notes-6.2.1)~ (Not important)
2. [ ] ~Tests covering your feature were added?~ (Tested locally. CTest is difficult because it must test for absence of a file.)


endJunction:
So far one problem discovered: an explicitly mentioned bulk mesh for output is not written. Working on this.

PR #2474, opened 2019-04-24
BGR Benchmarks

JohannesHerfurth:
Mechanics and LiquidFlow-Benchmarks
Selection of Tests published from Benchmark-Books


ThieJan:
As agreed some time ago we test against numerical reference solutions we checked against the analytical solutions locally. Some of the tests show a quite high numerical error on the azure ogsci test (one of the benchmarks with absolute error of order 1e-5). 

Do you think this is ok and we can adjust the tolerances accordingly to make the tests pass?

endJunction:
On azure node it seems there is different architecture. I run into the same issue before (don't remember the PR), where I had to set stricter tolerances on the nonlinear solver: 1e-14 instead of 1e-5. This reduced the solutions' spread and tighter tolerances for the tests could be set.
Also keep in mind, that derived quantities, like sigma, have lower order of convergence...

ThieJan:
Ok, good hint, We will check these tolerances.

ThieJan:
Even with smaller tolerances for the nonlinear solver I get large differences on the azure node for some tests.

Is there a way to get the computation results for the failing tests to investigate further on these differences?

endJunction:
Azure CI looks good! The absolute maximum errors are around 1e-10.
Need some adjustments to the ctests now.

ThieJan:
Our newly introduced tests are running now sucessfully (some with relaxed tolerances).

The failed test should not be related to the changes. Maybe you could check/ trigger a test restart?

endJunction:
@JohannesHerfurth @ThieJan Squashed all commits into one. Dropped files in erroneously included `build/Tests/Data/` folder. Original branch preserved.

endJunction:
@bilke Jenkins could not connect again to some nodes, in case you want to take a look on this.
Merge otherwise.

PR #2479, opened 2019-04-29
Output only explicitly given meshes

endJunction:
Replacing the previous attempt https://github.com/ufz/ogs/pull/2471.

Only if the output meshes list is empty, write it.
As requested by @ErikNixdorf

Some refactorings included avoiding code duplication.

The bulk mesh output and the observation meshes outputs are identical now but for the pvd files. The pvd files need some refactorings before => another PR.

### DIFF
--- a/ProcessLib/Output/Output.h
+++ b/ProcessLib/Output/Output.h
@@ -90,6 +90,13 @@ private:
         MeshLib::IO::PVDFile pvd_file;
     };
 
+    struct OutputFile;
+    void outputBulkMesh(OutputFile const& output_file,
+                        ProcessData* const process_data,
+                        MeshLib::Mesh const& mesh,
+                        double const t) const;
+
+private:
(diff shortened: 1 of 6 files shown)


ProcessLib/Output/Output.h
```diff
@@ -119,7 +122,7 @@ class Output
     ProcessData* findProcessData(Process const& process, const int process_id);
 
     //! Determines if there should be output at the given \c timestep or \c t.
-    bool shallDoOutput(unsigned timestep, double const t);
+    bool shallDoOutput(int timestep, double const t);
```
TomFischer: Is there any reason the timestep isn't `const`?

endJunction: Two points: in the declaration the `const` can be omitted anyway (even recommended by google's style guide), and secondly it is also non-const in the function's definition where it is used as a local variable.

PR #2487, opened 2019-05-09
[HT] Use MPL for solid phase properties.

TomFischer:
Using MPL for solid phase properties needs also changes in the project files. For the benchmarks this is done in this PR.

2. [x] All HT-Tests were changed, except the decovalex test since these tests use mesh parameters that are not implemented yet in the MPL. As soon as we can use a parameter as MPL::Property we can reenable the tests.


endJunction:
@bilke Merge, when ready.

PR #2491, opened 2019-05-12
Check BC parameters domains

endJunction:
Sometimes it happens, that a used parameter is defined on a different mesh, then the mesh node/element ids do not point to the right entities. This is difficult to detect at runtime (you have to look into the results) and leads to difficult to find errors.

This implements an additional function to check the domains of definition, and adds an optional call to the test when the parameter is found.

### DIFF
--- a/ParameterLib/Utils.h
+++ b/ParameterLib/Utils.h
@@ -33,13 +33,15 @@ ParameterBase* findParameterByName(
 /// \param parameters list of parameters in which it will be searched
 /// \param num_components the number of components of the parameters or zero if
 /// any number is acceptable
+/// \param mesh an optional mesh pointer used for test whether the parameter is
+/// defined on the given mesh. No test is performed if the pointer is a nullptr.
 ///
 /// \see The documentation of the other findParameter() function.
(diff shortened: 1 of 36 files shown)


ParameterLib/Utils.h
```diff
@@ -68,6 +70,19 @@ Parameter<ParameterDataType>* findParameterOptional(
...
 
+    // Test the parameter's mesh only if there is a "test"-mesh provided.
+    if (mesh != nullptr)
+    {
+        if (auto const error = isDefinedOnSameMesh(*parameter, *mesh))
+        {
+            OGS_FATAL(
+                "The found parameter is not suitable for the use on the "
+                "required mesh.\n%s",
+                *error);
```
TomFischer: Is it correct that `*error` is a string? I thought it is necessary to use the `c_str()` method for output?

endJunction: That's right; in the message there is a `%s`, and `*error` is a string argument.

bilke:
@endJunction Jenkins re-ran successfully. See #2497 for the correct (default) compiler.

PR #2492, opened 2019-05-14
Extract surface with scalar arrays

rinkk:
Extended mesh surface extraction for both DataExplorer + CL-tool to also copy scalar arrays (both point- and cell-based). Currently only int and double arrays are transferred and tuple sizes > 1 are ignored.

1. [ x] Feature description was added to the [changelog](https://github.com/ufz/ogs/wiki/Release-notes-6.2.1)
2. [ x] Tests covering your feature were added?

### DIFF
--- a/MeshLib/MeshSurfaceExtraction.cpp
+++ b/MeshLib/MeshSurfaceExtraction.cpp
@@ -104,8 +104,23 @@ MeshLib::Mesh* MeshSurfaceExtraction::getMeshSurface(
 
     // create new elements vector with newly created nodes (and delete
     // temp-elements)
-    std::vector<MeshLib::Element*> new_elements =
-        createSfcElementVector(sfc_elements, sfc_nodes, node_id_map);
+    std::vector<MeshLib::Element*> new_elements;
+    try
+    {
+        new_elements =
+            createSfcElementVector(sfc_elements, sfc_nodes, node_id_map);
+    }
+    catch (std::runtime_error const& err)
+    {
+        ERR("MeshSurfaceExtraction; could not create new surface "
+            "elements:\n%s.",
+            err.what());
+        std::for_each(sfc_elements.begin(), sfc_elements.end(),
+                      [](MeshLib::Element* e) { delete e; });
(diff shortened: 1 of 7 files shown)


MeshLib/MeshSurfaceExtraction.cpp
```diff
@@ -348,4 +327,100 @@ std::vector<MeshLib::Node*> MeshSurfaceExtraction::getSurfaceNodes(
...
+            new_nodes[k] =
+                sfc_nodes[node_id_map[sfc_element->getNode(k)->getID()]];
+        }
+        if (sfc_element->getGeomType() == MeshElemType::TRIANGLE)
+        {
+            new_elements.push_back(new MeshLib::Tri(new_nodes));
+        }
+        else
+        {
+            assert(sfc_element->getGeomType() == MeshElemType::QUAD);
```
endJunction: make it a strict check with `OGS_FATAL`...

MeshLib/MeshSurfaceExtraction.cpp
```diff
@@ -378,21 +374,52 @@ bool MeshSurfaceExtraction::createSfcMeshProperties(
...
+                std::vector<double> sfc_prop;
+                sfc_prop.reserve(vec_size[i]);
+                for (auto bulk_id : id_map[i])
+                {
+                    sfc_prop.push_back(org_vec[bulk_id]);
+                }
+                MeshLib::addPropertyToMesh<double>(sfc_mesh, name, type[i], 1, sfc_prop);
+                vectors_copied++;
+            }
+            else if (properties.existsPropertyVector<int>(name, type[i], 1))
```
endJunction: Handling of `int`-properties looks identical to the handling of the `double`-properties.
Reducing the code duplication by extracting the code in a function would be great.

rinkk: :heavy_check_mark:

MeshLib/MeshSurfaceExtraction.cpp
rinkk: :heavy_check_mark:

PR #2497, opened 2019-05-16
[Docker] Use default gcc (7.3) on Ubuntu.

bilke:
Corresponding change: https://github.com/ufz/ogs-container-maker/commit/79241adda9744ae4ee12ef88417a249807620599.

Merge when green.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -172,8 +172,7 @@ pipeline {
             dockerfile {
               filename 'Dockerfile.gcc.gui'
               dir 'scripts/docker'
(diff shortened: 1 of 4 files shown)


bilke:
@endJunction Everything is fine except that there is one new warning on GCC in`Lubby2.h:219`

```
'*((void*)& local_lubby2_properties +72)' is used uninitialized in this function [-Wuninitialized] eps_K.dot(sigma - eta_K * (eps_K - eps_K_prev) / dt) / 2;
```

Can you please have a look?

PR #2500, opened 2019-05-19
Rename time loop

endJunction:
Rename the time loop, extract the create method, rearrange private members.


wenqing:
or TimeDiscretization.

endJunction:
Discussed with Wenqing. Keeping the TimeLoop name for now.
Time discretization consists of three parts in OGS: The Euler/BDF/..., the Fixed/Adaptive/...Stepper, and the time loop itself.

PR #2501, opened 2019-05-20
Non-equilibrium initial state variables in SmallDeformation process.

endJunction:
Adds a possibility of non-equilibrium initial state in the small deformation process.

Two new methods for the finite element creation and coordinates interpolation added for convenience.

3. [x] Any new feature or behavior change was documented? (pdf in the Tests/Data together with the project files.)


ProcessLib/SmallDeformation/CreateSmallDeformationProcess.cpp
```diff
@@ -101,9 +101,25 @@ std::unique_ptr<Process> createSmallDeformationProcess(
...
+            //! \ogs_file_param_special{prj__processes__process__SMALL_DEFORMATION__nonequilibrium_state_variables__sigma}
+            "sigma", parameters,
+            MathLib::KelvinVector::KelvinVectorDimensions<
+                DisplacementDim>::value);
+    }
+
     SmallDeformationProcessData<DisplacementDim> process_data{
-        materialIDs(mesh), std::move(solid_constitutive_relations),
-        solid_density, specific_body_force, reference_temperature};
+        materialIDs(mesh),     std::move(solid_constitutive_relations),
```
TomFischer: Is this clang formatted?

ProcessLib/SmallDeformation/SmallDeformationFEM.h
```diff
@@ -272,6 +296,24 @@ class SmallDeformationLocalAssembler
             auto const& b = _process_data.specific_body_force;
             local_b.noalias() -=
                 (B.transpose() * sigma - N_u_op.transpose() * rho * b) * w;
+            if (_process_data.nonequilibrium_sigma)
+            {
```
TomFischer: Possible code duplication?

ProcessLib/SmallDeformation/SmallDeformationFEM.h
```diff
@@ -147,8 +149,30 @@ class SmallDeformationLocalAssembler
...
+                        std::numeric_limits<
+                            double>::quiet_NaN() /* time independent */,
+                        x_position);
+                typename BMatricesType::KelvinVectorType const sigma_ref =
+                    Eigen::Map<typename BMatricesType::KelvinVectorType>(
+                        sigma_ref_vector.data(),
+                        MathLib::KelvinVector::KelvinVectorDimensions<
+                            DisplacementDim>::value,
+                        1);
+                ip_data.sigma = sigma_ref;
```
wenqing: I think that sigma_ref can computed once at here and store it to ip_data.

ProcessLib/SmallDeformation/SmallDeformationFEM.h
```diff
@@ -272,6 +296,24 @@ class SmallDeformationLocalAssembler
             auto const& b = _process_data.specific_body_force;
             local_b.noalias() -=
                 (B.transpose() * sigma - N_u_op.transpose() * rho * b) * w;
+            if (_process_data.nonequilibrium_sigma)
+            {
+                x_position.setCoordinates(MathLib::Point3d(
```
wenqing: Here I think sigma_ref can be directly fetched from ip_data if it keeps  sigma_ref.

ProcessLib/SmallDeformation/SmallDeformationFEM.h
wenqing: initial_sigma or initial_stress could be a proper name.

norihiro-w:
perhaps you also need to include reference traction force, ie, neumann bc.

ProcessLib/SmallDeformation/CreateSmallDeformationProcess.cpp
```diff
@@ -101,9 +101,25 @@ std::unique_ptr<Process> createSmallDeformationProcess(
...
+    // Non-equilibrium variables
+    ParameterLib::Parameter<double> const* nonequilibrium_sigma = nullptr;
+    const auto& nonequilibrium_state_variables_config =
+        //! \ogs_file_param{prj__processes__process__SMALL_DEFORMATION__nonequilibrium_state_variables}
+        config.getConfigSubtreeOptional("nonequilibrium_state_variables");
+    if (nonequilibrium_state_variables_config)
+    {
+        nonequilibrium_sigma = &ParameterLib::findParameter<double>(
+            *nonequilibrium_state_variables_config,
+            //! \ogs_file_param_special{prj__processes__process__SMALL_DEFORMATION__nonequilibrium_state_variables__sigma}
```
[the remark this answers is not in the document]
wenqing: if (nonequilibrium_state_variables_config), set body force=zero.

wenqing:
@norihiro-w Good idea!. Since this gives a non-equilibrium initial state, I think, only a reference of body force has to be excluded. The traction bc can be considered  in the project file directly.

wenqing:
@norihiro-w The reference body force can be also excluded from the input data, e.g, if no change in body force, set  **0** in `<specific_body_force>`, otherwise, give the increment value.

norihiro-w:
@wenqing though I'm not checking the implementation, i guess here you want to solve the volume integration of `\nabla \cdot (stress - stress_ref) + (b-b_ref) = 0` (if i'm wrong, please correct it).  Then., in addition to `stress_ref`, you need to set `b_ref` (i forgot about this in previous comment) and boundary integrals of `stress_ref` which is the traction force (Nuemann bc). Of course, if `b=b_ref` (e.g. no density changes) or 't=t_ref'(e.g. constant crustal stress), you can simply set 'b=0' or 't=0' in the input file. However, I find the approach implicit and actual model setup becomes unclear for users.

wenqing:
@norihiro-w I think that you are right that this PR is aimed to solve `\nabla \cdot (stress - stress_ref) + (b-b_ref) = 0`. Where non-equilibrium means that the given reference stress and the reference body force  do not need to satisfy the momentum balance equation strictly, and they should be  excluded in  displacement change. Therefore the raised traction BC   of  `(stress - stress_ref) \cdot n` can be configured in the project file directly. Of course, the stress associated material parameters must be calculated by using the value of 'stress'. You know well that in some cases of geo-engineering application only measured normal stresses are available, which may does not strictly satisfy  `\nabla \cdot stress_ref + b_ref = 0`, and one only can or want to obtain the displacement from the initial status (0).

wenqing:
I think to add  an explanation in the document about how to set body force and  traction BC under the nonequilibrium condition is better than to change the corresponding source code. Because this can explicitly remind one the balance equation `\nabla \cdot (stress - stress_ref) + (b-b_ref) = 0`. This is just my opinion.

ProcessLib/SmallDeformation/SmallDeformationFEM.h
endJunction: Yes. Removing, and as suggested by @wenqing storing in the ip_data.

endJunction:
@norihiro-w We have mentioned some points in the pdf (see the test case folder). Especially the handling of the body-force term when density is not constant. We might miss some points and can improve the code when the need arises.
For this PR the density in the small deformation process is constant.

PR #2502, opened 2019-05-20
Fix unread variable warnings

endJunction:
Warning reported by cpp-check.

### DIFF
--- a/Applications/FileIO/SHPInterface.cpp
+++ b/Applications/FileIO/SHPInterface.cpp
@@ -193,7 +193,6 @@ void SHPInterface::readPolygons(const SHPHandle& hSHP, int numberOfElements,
     readPolylines(hSHP, numberOfElements, listName);
 
     auto const polylines = _geoObjects.getPolylineVec(listName);
-    auto sfc_vec = std::make_unique<std::vector<GeoLib::Surface*>>();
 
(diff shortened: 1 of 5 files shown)


BaseLib/FileTools.cpp
```diff
@@ -39,28 +39,6 @@ bool IsFileExisting(const std::string &strFilename)
     return (stat (strFilename.c_str(), &buffer) == 0);
 }
 
-double swapEndianness(double const& v)
```
TomFischer: This function is needed by the GocadSGridReader. Please, don't remove it.

endJunction: As discussed, there is a template version of this `T swapEndinness(T const&)`, and the code still compiles.
In the GocadGridReader two versions are called, one for doubles and another for floats, and implementations of both versions are different: float is written backwards [f3, f2, f1, f0], but double is split in two halves, which are written backwards each: [d3, d2, d1, d0, d7, d6, d5, d4].

Boost endian library handles only integers. A [Wikipedia article]( https://en.wikipedia.org/wiki/Endianness#Floating_point) on endianness says that the floating point numbers are treated as integers. The latter is contradicting the current implementation.

Dropping the change for now...

PR #2506, opened 2019-05-22
Explicit constructors

endJunction:
As reported by cppcheck, single argument ctors should be marked explicit.

Additionally add final specifier to the modified classes (where applicable).
Remove one defaulted dtor.


rinkk:
:+1:

PR #2522, opened 2019-06-09
Fixed a few cppcheck/clang-tidy warnings in chemistry library

renchao-lu:
As titled.

### DIFF
--- a/ChemistryLib/PhreeqcIO.h
+++ b/ChemistryLib/PhreeqcIO.h
@@ -80,6 +80,6 @@ private:
     std::unique_ptr<Output> const _output;
     std::vector<std::pair<int, std::string>> const&
         _process_id_to_component_name_map;
-    double _dt;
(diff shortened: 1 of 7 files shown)


ChemistryLib/PhreeqcIO.h
```diff
@@ -80,6 +80,6 @@ class PhreeqcIO
     std::unique_ptr<Output> const _output;
     std::vector<std::pair<int, std::string>> const&
         _process_id_to_component_name_map;
-    double _dt;
+    double _dt = std::numeric_limits<double>::quiet_NaN();
```
wenqing: It it better to move this to the initializer list of the ctor. The same for other similar changes.

endJunction: To be discussed. Class member initialization also has advantages.

wenqing: Since the member is not an instance of object, it is OK to set it as in class initialization.

Issue #2518, opened 2019-06-07
ChemistryLib -- PhreeqcInterface -- flag "-formula" in KINETICS bloc


joboog:
### Description

To define the stoichiometric relationships of components defined under SOLUTION with a KINETICS reactant, Phreeqc has the flag *-formula* in the KINETICS bloc. Setting this flag has not been included in the ogs6--PhreeqcInterface yet. It would provide the use of negative stoichiometric factors, which are required in complex kinetic reactions. This is not allowed by using  the PHASES  bloc in the phreeqc data base file *.dat to define the stoichiometry, which is currently the only option.

## Implementation

 The flag is described in the Phreeqc user manual on page [126](https://pubs.usgs.gov/tm/06/a43/pdf/tm6-A43.pdf).
It's function can easily be tested by modifying the already included test *KineticReactant*.
This test involves the kinetic reaction: *Productc = Synthetica + Syntheticb*, where the stoichiometry is defined in the PHASES bloc in the database file *.dat as:

*PHASES
Productc
SyntheticaSyntheticb = Synthetica+ + Syntheticb-
log_k -5*

This definition can be replaced  by setting the flag *-formula Synthetica 1   Syntheticb 1* under the KINETICS bloc when creating the preeqc input file ( within the PhreeqcInterface loop). Therefore, a sub-tag would have to be created in the *.prj file under the tag `<chemical_system><kinetic_reactants><kinetic_reactant>` where the flag can be defined, similar to the flags *-m* or *-parameters*. The simulation output of the modified *KineticReactant* test should be similar as to the original test (at temperature of  25°C, so *log_k -5* has no influence).

## Specifications

  - Version:  (6.2.0 -- current master)
  - Platform: (cross platform)

montoyav:
Good point! Definitely, the flag -formula should be include if not, kinetics reactions with only involving aqueous components would not be possible/correct.
As a note: negative stoichiometric coefficients means "reactant", positive number "product" in other words. Positive or negative number is only given information in which part of the "=" in the reaction* defined in SOLUTION is the aqueous component--> left (positive), right (negative)

* Productc = Synthetica + Syntheticb

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/2518).

PR #2523, opened 2019-06-09
Add an optional chemical formula flag under the subtree of kinetic reactant

renchao-lu:
As titled, this pull request meets the feature request #2518, along with a comparable simple benchmark test.


renchao-lu:
Could you please review the implementation? @joboog

ChemistryLib/PhreeqcIOData/CreateKineticReactant.cpp
```diff
@@ -29,17 +29,28 @@ std::vector<KineticReactant> createKineticReactants(
         //! \ogs_file_param{prj__chemical_system__kinetic_reactants__kinetic_reactant__name}
         auto name = reactant_config.getConfigParameter<std::string>("name");
 
+        auto chemical_formula_config =
+            //! \ogs_file_param{prj__chemical_system__kinetic_reactants__kinetic_reactant__chemical_formula}
+            reactant_config.getConfigParameterOptional<std::string>(
+                "chemical_formula");
```
endJunction:
```c++
reactant_config.getConfigParameter<std::string>(
```
```suggestion
                "chemical_formula", "");
```

And same for the parameters below. The if-condition is already included in the `getConfigParameter` function.

renchao-lu: I have not noticed the default setting contained in the `getConfigParameter` ever.

PR #2526, opened 2019-06-14
[PL] BC; Fix lower order element values access.

endJunction:
In cases when the underlying element is quadratic but the shape function
is of linear order, there is a dimension mismatch between the
`NodalVectorType` and the return value of the `getNodalValuesOnElement`.

Taking the first rows of the return matrix ensures correct dimensions.

Changing only for BC's which use the `getNodalValuesOnElement` and
are used by processes like HM/RM etc. where one of the process
variables is of lower order.

The error is present already in the master for quite some time and only shows
up when compiling with fixed size matrices in debug mode.


endJunction:
merging tonight...

PR #2527, opened 2019-06-15
Cleanup MaterialStateVariables in solid models

endJunction:
Originally pointed out by cppcheck about some assignment operator not being marked with 'override', but resulted in the following refactoring.


endJunction:
merging tonight...

PR #2531, opened 2019-06-17
[MatL/MPL] Check uniqueness of components and phases

renchao-lu:
Prevent creation of temporary variables in checking the uniqueness.

### DIFF
--- a/MaterialLib/MPL/CreateComponent.cpp
+++ b/MaterialLib/MPL/CreateComponent.cpp
@@ -72,10 +72,6 @@ std::vector<std::unique_ptr<Component>> createComponents(
     }
 
     std::vector<std::unique_ptr<Component>> components;
-
-    // Collect component's names to avoid duplicate components.
-    std::set<std::string> component_names;
-
     for (
         auto const& component_config :
         //! \ogs_file_param{prj__media__medium__phases__phase__components__component}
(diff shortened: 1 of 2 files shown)


MaterialLib/MPL/CreateComponent.cpp
```diff
@@ -72,28 +72,29 @@ std::vector<std::unique_ptr<Component>> createComponents(
...
         //! \ogs_file_param{prj__media__medium__phases__phase__components__component}
         config->getConfigSubtreeList("component"))
     {
         auto component = createComponent(component_config);
 
-        bool new_insertion = false;
-        std::tie(std::ignore, new_insertion) =
-            component_names.insert(component->name());
-        if (!new_insertion)
+        auto it = std::find_if(
```
endJunction: `BaseLib::findElementOrError()`

renchao-lu: At first glance, `BaseLib::findElementOrError()` is the first choice. This function returns custom error message if the element we are searching for is not found. It is opposite to what we want here - return error message if not found.

endJunction: Didn't see that! :fast_forward:

MaterialLib/MPL/CreateComponent.cpp
```diff
@@ -72,28 +72,29 @@ std::vector<std::unique_ptr<Component>> createComponents(
...
+            components.begin(), components.end(), [&component](auto const& c) {
```
endJunction: Just a comment: one can also capture a new variable in the lambda, to avoid repeated evaluations.
```c++
auto f = [name = component->name()](auto const& c) { return c->name() == name; };
```
or if you desperately want a reference, then
```c++
... [name = std::cref(component->name())] ...
```

renchao-lu: Good point! I see.

renchao-lu: Changed.

PR #2532, opened 2019-06-17
[CL] Improve data structure of Phreeqc Interface

renchao-lu:
- Use base class `ostream` for output; 
- Operator overloading on class templates;

### DIFF
--- a/ChemistryLib/Output.cpp
+++ b/ChemistryLib/Output.cpp
@@ -7,69 +7,69 @@
  *
  */
 
-#include <fstream>
+#include <ostream>
 
 #include "Output.h"
 
 namespace ChemistryLib
 {
-std::ofstream& operator<<(std::ofstream& out,
-                          BasicOutputSetups const& basic_output_setups)
+std::ostream& operator<<(std::ostream& os,
+                         BasicOutputSetups const& basic_output_setups)
 {
-    out << "-file " << basic_output_setups.output_file << "\n";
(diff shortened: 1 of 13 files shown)


ChemistryLib/PhreeqcIO.cpp
```diff
@@ -19,6 +19,18 @@
 
 namespace ChemistryLib
 {
+namespace
+{
+template <typename DataBlock>
+std::ostream& operator<<(std::ostream& os,
+                         std::vector<DataBlock> const& data_blocks)
+{
+    std::copy(data_blocks.begin(), data_blocks.end(),
```
endJunction: I'd think, it could be
```c++
return std::copy(...);
```

ChemistryLib/Output.cpp
```diff
@@ -7,69 +7,69 @@
  *
  */
 
-#include <fstream>
+#include <iostream>
```
endJunction:
```suggestion
#include <ostream>
```

renchao-lu: Changed as suggested except in the `PhreeqcIO.cpp` where operator `>>` (istream) is used.

ChemistryLib/PhreeqcIO.cpp
renchao-lu: May discuss tomorrow...

PR #2533, opened 2019-06-18
[CL] Move creation of Phreeqc instance in the ctor

renchao-lu:
With current commit changes, creation of phreeqc instance and database loading process will be executed only once in the ctor.


renchao-lu:
Rebased.

PR #2537, opened 2019-06-24
[Web] Fixes on web documentation of equilibrium phase benchmark

renchao-lu:
As titled.


web/content/docs/benchmarks/reactive-transport/calcite.pandoc
```diff
@@ -31,7 +31,7 @@ The operator-splitting (OS) model formulation (Strang, 1968) with a sequential n
 
 A comparison of the numerical data is shown in the figure below. The numerical results of OGS-6 coincide with those of OGS-5.
```
endJunction: Another useful information would be:

"... coincide with those of OGS-5 within 25% relative error."

Adjust the numbers as needed.

renchao-lu: Added absolute error information.

renchao-lu:
Rebased.

PR #2541, opened 2019-06-26
LIE/HM fracture permeability common changes

endJunction:
This are parts of an ancient PR #2014 .

The only significant change in my opinion is that negative apertures will be reset to 0 in assembly.
In the output any aperture will be taken.

### DIFF
--- a/ProcessLib/LIE/HydroMechanics/LocalAssembler/HydroMechanicsLocalAssemblerFracture-impl.h
+++ b/ProcessLib/LIE/HydroMechanics/LocalAssembler/HydroMechanicsLocalAssemblerFracture-impl.h
@@ -225,10 +225,11 @@ void HydroMechanicsLocalAssemblerFracture<ShapeFunctionDisplacement,
         b_m = ip_data.aperture0 + w[index_normal];
         if (b_m < 0.0)
         {
-            OGS_FATAL(
+            DBUG(
                 "Element %d, gp %d: Fracture aperture is %g, but it must be "
-                "non-negative.",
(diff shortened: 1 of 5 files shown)


ProcessLib/LIE/HydroMechanics/LocalAssembler/HydroMechanicsLocalAssemblerFracture-impl.h
```diff
@@ -355,9 +363,9 @@ void HydroMechanicsLocalAssemblerFracture<ShapeFunctionDisplacement,
...
         if (b_m < 0.0)
         {
-            OGS_FATAL(
-                "Element %d, gp %d: Fracture aperture is %g, but it must be "
-                "non-negative.",
+            DBUG(
+                "Element %d, gp %d: Fracture aperture is %g, but it is "
+                "expected to be non-negative.",
                 _element.getID(), ip, b_m);
         }
```
wenqing: Forgot to set b_m = 0.0;

endJunction: Not sure here, because it is only used in the output, so if the solution is with negative aperture, so let it be visible.
In the assembly above, it is set to zero explicitly to avoid non-admissible range.
I'd like to keep the output as it is (allowing negative values).

wenqing: I see.

PR #2547, opened 2019-06-28
[FileIO] Gocad TSurface interface

rinkk:
Adds an interface that converts *.ts files into one or more OGS meshes.
Nodes, Elements and scalar properties are parsed, various geographical and geological information as well as border-information is ignored (borders could potentially be used as boundary conditions, so maybe this is worth looking into in the future).

The interface is integrated into the Data Explorer and a simple command line tool has been added


Applications/Utils/Tests.cmake
```diff
@@ -243,3 +243,13 @@ MeshTest(
     REQUIREMENTS NOT OGS_USE_MPI
     DIFF_DATA Back.vtu Back.vtu 1e-16
 )
+
+AddTest(
```
endJunction: Better use MeshTest macro as above...

Applications/FileIO/GocadIO/GocadTSurfaceReader.h
```diff
@@ -0,0 +1,126 @@
...
+                   std::vector<MeshLib::Element*>& elems);
+
+    enum class NODE_TYPE
+    {
+        UNSPECIFIED,
+        VRTX,
+        PVRTX
+    };
+
+    std::vector<MeshLib::Mesh*> _mesh_vec;
```
endJunction: vector of unique pointers is recommended.

Applications/FileIO/GocadIO/GocadTSurfaceReader.h
```diff
@@ -0,0 +1,126 @@
...
+ * @copyright
+ * Copyright (c) 2012-2019, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/LICENSE.txt
+ */
+
+#pragma once
+
+#include <iostream>
```
endJunction: include iosfwd in the header file, if possible, an only the needed stream headers in the cpp.

Applications/FileIO/GocadIO/GocadTSurfaceReader.cpp
```diff
@@ -0,0 +1,525 @@
...
+MeshLib::Mesh* GocadTSurfaceReader::getData(std::size_t const idx) const
+{
+    if (_mesh_vec.empty())
+    {
+        ERR("Error: No mesh data available.");
+        return nullptr;
+    }
+    if (idx < _mesh_vec.size())
+        return _mesh_vec[idx];
+    ERR("Error: Mesh index (%d) out of range (0, %d).", idx, _mesh_vec.size());
```
TomFischer: The max index have to be `_mesh_vec.size() - 1`.

Applications/FileIO/GocadIO/GocadTSurfaceReader.cpp
```diff
@@ -0,0 +1,525 @@
...
+        ERR("Error: No mesh data available.");
+        return std::string();
+    }
+    if (idx < _mesh_vec.size())
+        return _mesh_vec[idx]->getName();
+    ERR("Error: Mesh index (%d) out of range (0, %d).", idx, _mesh_vec.size());
+    return std::string();
+}
+
+void GocadTSurfaceReader::writeData(std::string const& file_name,
```
TomFischer: Why does the reader need the possibility to write the meshes. This could be done also by external functionality, for  instance the DataExplorer.

Applications/FileIO/GocadIO/GocadTSurfaceReader.cpp
```diff
@@ -0,0 +1,525 @@
...
+
+        std::stringstream sstr(line);
+        if (line.substr(0, 4) == "VRTX" && t != NODE_TYPE::PVRTX)
+        {
+            nodes.push_back(createNode(sstr));
+        }
+        else if (line.substr(0, 5) == "PVRTX" && t != NODE_TYPE::VRTX)
+        {
+            nodes.push_back(createNode(sstr));
+            for (std::string name : array_names)
```
TomFischer: Why access by value? Why not `for (std::string const& name : array_names)`?

Applications/FileIO/GocadIO/GocadTSurfaceReader.h
rinkk: :heavy_check_mark:

Applications/Utils/Tests.cmake
rinkk: :heavy_check_mark:

Applications/FileIO/GocadIO/GocadTSurfaceReader.cpp
rinkk: The mesh writing functionality including all checks and path construction currently covers >40 LoC that otherwise would need to be duplicated in both DE and CLI. In my opinion it makes sense, that an interface offers functionality to save data with a single call if it outputs something more complex than a single mesh or something comparable. In this case it's probably not obvious which processing steps have to be performed to savely get and write the data (i.e. one mesh vs multiple meshes, path construction on windows vs linux, etc.). Other file interfaces do the same thing.

If you prefer that I move that functionality to another file (TSurfaceIO or whatever) that is certainly possible. But I won't include (and duplicate) all that code in the DE, that is not the place for IO functionality!

Applications/FileIO/GocadIO/GocadTSurfaceReader.cpp
rinkk: :heavy_check_mark:

Applications/FileIO/GocadIO/GocadTSurfaceReader.cpp
rinkk: :heavy_check_mark:

Applications/FileIO/GocadIO/GocadTSurfaceReader.h
endJunction: I'd like to merge the PR, but there still are memory leaks when executing the utility exe. Many of them are related to the meshes.
```sh
> bin/GocadTSurfaceReader -i ../s/Tests/Data/MeshLib/Top-Lower-Shaly.ts -o .
...
SUMMARY: AddressSanitizer: 4133962 byte(s) leaked in 71299 allocation(s).
```

rinkk: :heavy_check_mark:

Applications/FileIO/GocadIO/GocadTSurfaceReader.cpp
rinkk: :heavy_check_mark:

rinkk:
Technically, the ctor isn't needed anymore...

PR #2554, opened 2019-07-05
HM mass formulation

FZill:
Changed HM to trivial mass formulation and adapted tests where necessary


norihiro-w:
what is a motivation of this change?

endJunction:
We want to solve non-constant density HM problems. The  next step is indeed to add the density derivatives to the FEM.

norihiro-w:
@endJunction i see. thank you for the reply.

PR #2557, opened 2019-07-07
LIE Mohr-Coulomb; Use Newton-Raphson solver internally.

endJunction:
Change the explicit calculation of stress in plastic case to Newton-Raphson method. The new implementation yields far better and more stable results; compare the result of new load path test.

1. [ ] Feature description was added to the [changelog](https://github.com/ufz/ogs/wiki/Release-notes-6.2.1)
2. [x] Tests covering your feature were added? New load path test.


norihiro-w:
@endJunction thanks for improving this. I'm curious how much the new result is different from the previous one. Is the previous result completely wrong?

endJunction:
@norihiro-w Below a comparison for the load path test: While the loading up into the plastic area is comparable, the unloading with the explicit scheme is not physical. Don't hesitate to ask more questions.
image

endJunction:
For reviewers: I've to find out why the results on different architectures vary more than reasonable tolerances, but the code isn't going to change by any large extend.

wenqing:
Does that figure show stress - aperture (jump) curve? If it is, the results by the Newton looks reasonable.

The implementation looks good.

endJunction:
@wenqing Yes, the figure show fracture stress {n, s} in {red, blue} as well as the fracture opening in {n, s} direction in {magenta, black}.

norihiro-w:
@endJunction thank you for the figure.

PR #2558, opened 2019-07-08
Added temperture dependent to Lubby2

zhangning737:
- Update Lubby2 model for temperature dependenct
- Tested and added testing benchmark
- Compatible with previous benchmark in SmallDeformation Process

Everything works, but not good at optimizing the code, please help optimize the code, especially here,

https://github.com/zhangning737/ogs/blob/6d398b6cf7ce42f99e93f1339882ac2b519853d5/MaterialLib/SolidModels/CreateLubby2.h#L131

Thanks a lot...

### DIFF
--- a/MaterialLib/SolidModels/CreateLubby2.h
+++ b/MaterialLib/SolidModels/CreateLubby2.h
@@ -117,7 +117,7 @@ std::unique_ptr<Lubby2<DisplacementDim>> createLubby2(
     if (dependency_parameter_tref_name || dependency_parameter_mgt_name ||
         dependency_parameter_mkt_name || dependency_parameter_q_name)
     {
-        WARN("Lubby2 model lost essential thermal parameter(s), the thermal related properties are not activated.")
+        WARN("Lubby2 model lost essential thermal parameter(s), the thermal related properties are not activated.");
     }
 
     // Initialize the thermal related parameters 
--- a/MaterialLib/SolidModels/CreateLubby2.h
+++ b/MaterialLib/SolidModels/CreateLubby2.h
@@ -11,9 +11,8 @@
 #pragma once
 
 #include "CreateNewtonRaphsonSolverParameters.h"
-#include "ParameterLib/Utils.h"
-
 #include "Lubby2.h"
+#include "ParameterLib/Utils.h"
 
 namespace MaterialLib
 {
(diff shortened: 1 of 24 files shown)


MaterialLib/SolidModels/CreateLubby2.h
```diff
@@ -93,20 +94,79 @@ std::unique_ptr<Lubby2<DisplacementDim>> createLubby2(
...
+    {
+       auto parameter_empty_ptr = new ParameterLib::ConstantParameter<double>("NoName", 0.0);
+
+        Lubby2MaterialProperties mp{
+            kelvin_shear_modulus,     maxwell_shear_modulus,
+            maxwell_bulk_modulus,     kelvin_viscosity,
+            maxwell_viscosity,        dependency_parameter_mK,
+            dependency_parameter_mvK, dependency_parameter_mvM,
+            *parameter_empty_ptr, *parameter_empty_ptr,
+            *parameter_empty_ptr, *parameter_empty_ptr};
```
endJunction: I'd recommend to put the additional temperature related coefficients into a structure and pass them using `boost::optional`.
In the newly written code the use of plain pointers is prohibited, especially those causing memory leaks (like the above `parameter_empty_ptr`).

MaterialLib/SolidModels/Lubby2.h
```diff
@@ -72,10 +85,34 @@ struct LocalLubby2Properties
...
 
+    void update(double const s_eff, double const T)
+    {
+        double const dT(T - Tref);
+        double const GM0_s_eff = GM0 * s_eff;
+        GK = GK0 * std::exp(mK * GM0_s_eff);
+        etaK = etaK0 * std::exp(mvK * GM0_s_eff);
+        if (std::isnan(T))
+        {
+            etaM = etaM0 * std::exp(mvM * GM0_s_eff);
```
endJunction: Try to avoid any code duplication.
Especially this branch and all of the above computations are already implemented in the `update(s)` overload. Don't repeat them.

Tests/Data/ThermoMechanics/Lubby2/cube_1e0.prj
```diff
@@ -0,0 +1,327 @@
...
+            <name>temperature</name>
+            <components>1</components>
+            <order>1</order>
+            <initial_condition>temperature_ic</initial_condition>
+            <boundary_conditions>
+                <!-- <boundary_condition>
+                    <mesh>cube_1x1x1_geometry_top</mesh>
+                    <type>Dirichlet</type>
+                    <parameter>temperature_bc</parameter>
+                </boundary_condition> -->
```
endJunction: Drop unneeded commented blocks.

zhangning737: OK, removed.

MaterialLib/SolidModels/Lubby2.h
zhangning737: Here we have a problem, in the function ```integrateStress``` ```update(s)``` was changed to ```update(s, T)``` for TM process, while for ```SmallDeformation``` ```T``` is ```nan```, so add a check point here... Thinking how to due with that...

MaterialLib/SolidModels/CreateLubby2.h
```diff
@@ -93,20 +94,79 @@ std::unique_ptr<Lubby2<DisplacementDim>> createLubby2(
     DBUG("Use '%s' as dependency parameter mvM.",
          dependency_parameter_mvM.name.c_str());
 
-    Lubby2MaterialProperties mp{
-        kelvin_shear_modulus,     maxwell_shear_modulus,
-        maxwell_bulk_modulus,     kelvin_viscosity,
-        maxwell_viscosity,        dependency_parameter_mK,
-        dependency_parameter_mvK, dependency_parameter_mvM};
+    auto const dependency_parameter_tref_name = config.getConfigParameterOptional<std::string>("dependency_parameter_tref");
```
wenqing: - clang-format
- parameter documentation

MaterialLib/SolidModels/CreateLubby2.h
wenqing: or derive a class from Lubby2.

MaterialLib/SolidModels/CreateLubby2.h
zhangning737: @wenqing OK, will reformat and document...

MaterialLib/SolidModels/Lubby2.cpp
```diff
@@ -94,13 +94,15 @@ Lubby2<DisplacementDim>::integrateStress(
     // calculation of deviatoric parts
     auto const& P_dev = Invariants::deviatoric_projection;
     KelvinVector const epsd_i = P_dev * eps;
+    KelvinVector const epsd_t = P_dev * eps_prev;
 
-    // initial guess as elastic predictor
+    // initial guess as elastic predictor. Note: dimensionless stresses!
```
endJunction: `sigma_prev` two lines below is not dimensionless.

zhangning737: mis description, sigd_t is the dimensionless stress...

PR #2561, opened 2019-07-09
HydroMechanics non-equilibrium initial states.

endJunction:
Adding same non-equilibrium initial states as for the SmallDeformation process #https://github.com/ufz/ogs/pull/2501.

Meanwhile the theoretical ideas were developed further and this is only for the comparison with the next new implementation and also provides the project files/tests. The new idea is to include the non-equilibrium in the residuum.


Documentation/ProjectFile/prj/processes/process/HYDRO_MECHANICS/nonequilibrium_state_variables/i_nonequilibrium_state_variables.md
```diff
@@ -0,0 +1,3 @@
+Non-equilibrium initial state variables which are not explicitly in equilibrium
```
wenqing: not explicitly --> implicitly.

wenqing:
Will the changes in the local assembler be in another PR?

endJunction:
Yes, the F^{oob} (out of balance forces) formulation will follow in other PR.

ProcessLib/HydroMechanics/CreateHydroMechanicsProcess.cpp
```diff
@@ -188,13 +188,45 @@ std::unique_ptr<Process> createHydroMechanicsProcess(
         config.getConfigParameter<double>(
             "reference_temperature", std::numeric_limits<double>::quiet_NaN());
 
+    // Non-equilibrium variables
+    ParameterLib::Parameter<double> const* nonequilibrium_stress = nullptr;
```
TomFischer: Just out of curiosity: Why not ´unique_ptr`?

ProcessLib/HydroMechanics/HydroMechanicsProcessData.h
```diff
@@ -97,6 +101,12 @@ struct HydroMechanicsProcessData
     ParameterLib::Parameter<double> const& porosity;
     /// Solid's density. A scalar quantity, ParameterLib::Parameter<double>.
     ParameterLib::Parameter<double> const& solid_density;
+
+    /// Optional, non-equilibrium stress parameter.
+    ParameterLib::Parameter<double> const* const nonequilibrium_stress;
```
TomFischer: Why not ´unique_ptr`?

Tests/Data/HydroMechanics/InitialStates/displacement/into_initial_state_pcs_0.pvd
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0"?>
+<VTKFile type="Collection" version="0.1" byte_order="LittleEndian" compressor="vtkZLibDataCompressor">
+  <Collection>
+    <DataSet timestep="0" group="" part="0" file="into_initial_state_pcs_0_ts_0_t_0.000000.vtu"/>
+  </Collection>
+</VTKFile>
```
TomFischer: Is the pvd file needed?

TomFischer:
Are the gml files needed? I didn´t see references in the prooject files.

ProcessLib/HydroMechanics/CreateHydroMechanicsProcess.cpp
endJunction: It's a non-owning optional parameter.

wenqing:
I guess that Tests/Data/HydroMechanics/InitialStates/pressure/Docu_HM_Non_Equilibrium_Initial_States.pdf  is for the documentation of one benchmark. However the web content file, which should refer to that PDF is missing.

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
```diff
@@ -269,11 +302,11 @@ void HydroMechanicsLocalAssembler<ShapeFunctionDisplacement,
 
     // pressure equation
     local_rhs.template segment<pressure_size>(pressure_index).noalias() -=
-        laplace_p * p + storage_p * p_dot + Kpu * u_dot;
+        laplace_p * (p - p_neq) + storage_p * p_dot + Kpu * u_dot;
```
wenqing: I think that every simulation starts from the previous status, whether which is equilibrium (may from the previous simulation) or non-equilibrium (specfied). Therefore p_neq should be p0, and local_rhs should not be changed.  This is a suggestion for future change. The PR itself is OK with the current consideration.

nagelt:
This should be no longer needed with the CompensateInitial....

endJunction:
That's true, but I still didn't redo the test cases...

PR #2564, opened 2019-07-10
[CL] Output Solutions of Immobile Phases

renchao-lu:
As titled. The solutions of immobile phases defined in the chemical system, i.e., equilibrium phases and kinetic reactants, will be automatically written into the result files. Defining additional transport processes for outputting the solutions of immobile phases is not needed any more.

### DIFF
--- a/ChemistryLib/CreatePhreeqcIO.h
+++ b/ChemistryLib/CreatePhreeqcIO.h
@@ -13,6 +13,11 @@
 
 #include "PhreeqcIO.h"
 
+namespace MeshLib
+{
+class Mesh;
+}
+
 namespace BaseLib
 {
 class ConfigTree;
@@ -26,7 +31,7 @@ class Process;
 namespace ChemistryLib
 {
 std::unique_ptr<PhreeqcIO> createPhreeqcIO(
-    std::size_t const num_nodes,
+    MeshLib::Mesh const& mesh,
     std::vector<std::pair<int, std::string>> const&
         process_id_to_component_name_map,
     BaseLib::ConfigTree const& config,
--- a/ChemistryLib/PhreeqcIOData/CreateKineticReactant.h
+++ b/ChemistryLib/PhreeqcIOData/CreateKineticReactant.h
@@ -9,12 +9,17 @@
 
 #pragma once
 
-#include <boost/optional/optional.hpp>
+#include <boost/optional/optional_fwd.hpp>
 #include <vector>
 
 namespace BaseLib
 {
 class ConfigTree;
+}
+
+namespace MeshLib
+{
+class Mesh;
 }
 
 namespace ChemistryLib
@@ -25,5 +30,6 @@ struct KineticReactant;
 namespace ChemistryLib
 {
 std::vector<KineticReactant> createKineticReactants(
-    boost::optional<BaseLib::ConfigTree> const& config);
+    boost::optional<BaseLib::ConfigTree> const& config,
+    MeshLib::Mesh const& mesh);
 }  // namespace ChemistryLib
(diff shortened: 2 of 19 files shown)


renchao-lu:
The changes for outputting amounts of immobile phases (equilibrium phases and kinetic reactants) are constrained within the ChemistryLib. As you expected, there is definitely no need of intruding the ProcessLib::Output. In this way, least changes are necessary. Thanks for the guidance. @endJunction

ChemistryLib/PhreeqcIOData/CreateKineticReactant.h
```diff
@@ -12,18 +12,17 @@
 #include <boost/optional/optional.hpp>
 #include <vector>
 
+#include "KineticReactant.h"
```
endJunction: Why is the change from fwd-decl needed?

ChemistryLib/PhreeqcIOData/EquilibriumPhase.cpp
```diff
@@ -13,15 +13,10 @@
 
 namespace ChemistryLib
 {
-std::ostream& operator<<(std::ostream& os,
-                         EquilibriumPhase const& equilibrium_phase)
+void EquilibriumPhase::print(std::ostream& os,
+                             std::size_t const& chemical_system_id) const
```
endJunction: pass small things by value.

ChemistryLib/PhreeqcIOData/KineticReactant.cpp
```diff
@@ -13,28 +13,26 @@
 
 namespace ChemistryLib
 {
-std::ostream& operator<<(std::ostream& os,
-                         KineticReactant const& kinetic_reactant)
+void KineticReactant::print(std::ostream& os,
```
endJunction: Just curious, why change the output operator to `print`?

ChemistryLib/PhreeqcIOData/CreateKineticReactant.h
renchao-lu: I think the header file `KineticReactant.h` has to be included in the implementation file `CreateKineticReactant.cpp`, because the ctor of class `KineticReactant` is applied in this file for instance creation. If so, I can move this header file into `CreateKineticReactant.h` and drop the fwd-declaration. If my understanding is not correct, I can add the fwd-decl back and leave the header file in the implementation file.

ChemistryLib/PhreeqcIOData/KineticReactant.cpp
renchao-lu: Because the data type of `KineticReactant.amount` is changed to `MeshLib::PropertyVector<double>*`. `chemical system id`  is required from outside of the class `KineticReactant` to access elements by index when looping over the chemical systems. 

My preference is also to keep the operator overloading function herein. Can we keep it from drop-in replacement?

ChemistryLib/PhreeqcIOData/EquilibriumPhase.cpp
renchao-lu: Not fixed yet, as we may have better idea to retain the operator overloading function.

ChemistryLib/PhreeqcIOData/CreateKineticReactant.h
renchao-lu: Fixed.

ChemistryLib/PhreeqcIOData/EquilibriumPhase.cpp
renchao-lu: Fixed.

ChemistryLib/CreatePhreeqcIO.h
```diff
@@ -11,6 +11,7 @@
 
 #include <memory>
 
+#include "MeshLib/Mesh.h"
```
endJunction: fwd decl should work here too.

renchao-lu: Good catch. Fixed.

PR #2573, opened 2019-07-17
[Conan] Set build policy to outdated

bilke:
Ensures that the locally used package is always up-to-date
with the package definition on the remote.

Will lead to more local package builds but will hopefully
help in reducing errors.

1. [ ] Feature description was added to the [changelog](https://github.com/ufz/ogs/wiki/Release-notes-6.2.1)


bilke:
Everything is green now.

Jenkinsfile
```diff
@@ -108,6 +108,8 @@ pipeline {
                   "-DBUILD_SHARED_LIBS=${build_shared} " +
                   '-DOGS_CPU_ARCHITECTURE=generic ' +
                   '-DOGS_BUILD_UTILS=ON ' +
+                  '-DOGS_BUILD_UTILS=ON ' +
```
endJunction: OGS_BUILD_UTILS=ON was doubled...

bilke: Fixed in dc80cbc00.

PR #2575, opened 2019-07-18
RichardsM: Add missing saturation in bulk density computation

ThieJan:
Richards Mechanics process
-Missing saturation in bulk density computation added


nagelt:
Change is correct.

ThieJan:
> 
> 
> 👍
> It can be merged once the related reference results are updated.

We try with reference solutions computed here or shall we use the results from the azure/ jenkins nodes as references ?

endJunction:
Creating own references is easier, just run the ctests and overwrite the files in source/Tests/Data from build/Tests/Data/ with something like:
```sh
cd source/Tests/Data/X
for i in *_ts_*.vtu; do cp build/Tests/Data/X/${i} .; done
```
Then commit only changed files...

endJunction:
clang-format. See some info for pre-commit hooks: https://github.com/ufz/ogs-utils/tree/master/dev/code-formatting/clang-format-pre-commit-hook

PR #2583, opened 2019-07-29
Three updates in MPL

wenqing:
~1. Renamed thermal_expansivity as thermal_expansion~
2. Moved ProcessLib::HT::intrisincPermeabilty to a new file in MPL, and renamed the function to localTensor for general purpose.
3. Added a common function of effectiveThermalConductivity.

### DIFF
--- a/MaterialLib/MPL/Utils/FormEigenTensor.cpp
+++ b/MaterialLib/MPL/Utils/FormEigenTensor.cpp
@@ -1,0 +1,89 @@
+/*
+ * \file
+ * \copyright
+ * Copyright (c) 2012-2019, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ * Created on July 31, 2019, 11:28 AM
+ */
+
+#include "FormEigenTensor.h"
+
+#include <boost/variant/static_visitor.hpp>
+
+#include "MaterialLib/MPL/PropertyType.h"
+
+namespace MaterialPropertyLib
+{
+template <int GlobalDim>
+struct FormEigenTensor
+    : boost::static_visitor<Eigen::Matrix<double, GlobalDim, GlobalDim>>
+{
+    Eigen::Matrix<double, GlobalDim, GlobalDim> operator()(
+        double const& value) const
+    {
+        return Eigen::Matrix<double, GlobalDim, GlobalDim>::Identity() * value;
+    }
+
+    Eigen::Matrix<double, GlobalDim, GlobalDim> operator()(
+        MaterialPropertyLib::Vector const& values) const
+    {
+        return Eigen::Map<Eigen::Matrix<double, GlobalDim, 1> const>(
+                   values.data(), GlobalDim, 1)
+            .asDiagonal();
+    }
+
+    Eigen::Matrix<double, GlobalDim, GlobalDim> operator()(
+        MaterialPropertyLib::Tensor2d const& values) const
+    {
+        return Eigen::Map<Eigen::Matrix<double, GlobalDim, GlobalDim> const>(
+            values.data(), GlobalDim, GlobalDim);
+    }
+
+    Eigen::Matrix<double, GlobalDim, GlobalDim> operator()(
+        MaterialPropertyLib::Tensor const& values) const
+    {
+        return Eigen::Map<Eigen::Matrix<double, GlobalDim, GlobalDim> const>(
+            values.data(), GlobalDim, GlobalDim);
+    }
+
+    Eigen::Matrix<double, GlobalDim, GlobalDim> operator()(
+        MaterialPropertyLib::SymmTensor const& /*values*/) const
+    {
+        OGS_FATAL(
+            "The value of MaterialPropertyLib::SymmTensor is inapplicable");
+    }
+
+    Eigen::Matrix<double, GlobalDim, GlobalDim> operator()(
+        std::string const& /*values*/) const
+    {
+        OGS_FATAL("The value of std::string is inapplicable");
+    }
+
+    Eigen::Matrix<double, GlobalDim, GlobalDim> operator()(
+        MaterialPropertyLib::Pair const& /*values*/) const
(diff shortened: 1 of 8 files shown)


MaterialLib/MPL/MediumAuxiliaryFunction.h
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+#pragma once
+
+#include "MaterialLib/MPL/PropertyType.h"
+
+namespace MaterialPropertyLib
+{
+template <int GlobalDim>
+Eigen::Matrix<double, GlobalDim, GlobalDim> localTensor(
```
TomFischer: Why 'local'? Is there a global tensor? I would prefer the name 'makeTensor'.

MaterialLib/MPL/MediumAuxiliaryFunction.h
```diff
@@ -0,0 +1,63 @@
+/*
```
TomFischer: You could name the file also Utils.h to stay consistent with for instance ParameterLib. The ProcessLib has even a subfolder Utils that contains several files.

MaterialLib/MPL/MediumAuxiliaryFunction.h
```diff
@@ -0,0 +1,63 @@
...
+            boost::get<MaterialPropertyLib::Tensor>(values).data(), GlobalDim,
+            GlobalDim);
+    }
+    OGS_FATAL(
+        "The size of local tensor such intrinsic permeability, thermal "
+        "conductovity is neither one nor %d nor %d squared.",
+        GlobalDim, GlobalDim);
+}
+
+template <int GlobalDim>
```
TomFischer: MPL is a general concept of properties. The calculation of the effective thermal conductivity is process specific, i.e., it is used only by some of the processes. For this reason I suggest to leave the function in its old place. By the way, the new function isn't used anywhere.

MaterialLib/MPL/MediumAuxiliaryFunction.h
```diff
@@ -0,0 +1,63 @@
...
+
+#include "MaterialLib/MPL/PropertyType.h"
+
+namespace MaterialPropertyLib
+{
+template <int GlobalDim>
+Eigen::Matrix<double, GlobalDim, GlobalDim> localTensor(
+    MaterialPropertyLib::PropertyDataType const& values)
+{
+    if (boost::get<double>(&values))
```
endJunction: I suggest to rewrite the ifs with visitor, as described in [boost docu](https://www.boost.org/doc/libs/1_70_0/doc/html/variant/tutorial.html#variant.tutorial.basic)

MaterialLib/MPL/PropertyType.h
```diff
@@ -287,7 +287,7 @@ static const std::array<std::string, PropertyType::number_of_properties>
                              "specific_heat_capacity",
                              "storage",
                              "thermal_conductivity",
-                             "thermal_expansivity",
+                             "thermal_expansion",
```
endJunction: @wenqing @scinopode please discuss whether all of the items should be called one or the other way.
In my opinion expansivity is better because it describes how something would change, whereas expansion describes the actual change.

MaterialLib/MPL/MediumAuxiliaryFunction.h
endJunction: or maybe `eigenTensor`...

MaterialLib/MPL/PropertyType.h
wenqing: @endJunction Forgot to add `_coefficent`. Agree with you.  Will drop this change.

MaterialLib/MPL/MediumAuxiliaryFunction.h
wenqing: changed to `formTensor`.

MaterialLib/MPL/MediumAuxiliaryFunction.h
wenqing: Moved to ProcessLib/Utils

MaterialLib/MPL/MediumAuxiliaryFunction.h
wenqing: The function will be used in TH related processes, and its first application will be in THM. After moved to ProcessLib/Utils, I would like to keep the function in this file (now renamed as FormTensor.h)

MaterialLib/MPL/MediumAuxiliaryFunction.h
wenqing: Done.

MaterialLib/MPL/Utils/FormEigenTensor.cpp
```diff
@@ -0,0 +1,89 @@
...
+template <int GlobalDim>
+Eigen::Matrix<double, GlobalDim, GlobalDim> FormEigenTensor<GlobalDim>::
+operator()(MaterialPropertyLib::Pair const& /*values*/) const
+{
+    OGS_FATAL("The size of tensor is neither one nor %d nor %d squared.",
+              GlobalDim, GlobalDim);
+}
+template class FormEigenTensor<1>;
+template class FormEigenTensor<2>;
+template class FormEigenTensor<3>;
```
endJunction: No need for explicit instantiation, will be done by formEigenTensor() anyway.

MaterialLib/MPL/Utils/FormEigenTensor.h
```diff
@@ -0,0 +1,63 @@
...
+#include <boost/variant/static_visitor.hpp>
+#include <Eigen/Dense>
+
+#include "MaterialLib/MPL/Property.h"
+#include "MaterialLib/MPL/PropertyType.h"
+
+namespace MaterialPropertyLib
+{
+template <int GlobalDim>
+class FormEigenTensor
```
endJunction: The class definition goes into cpp too, because it is not needed outside. Only the formEigenTensor() declaration (with explicit template instantiation declarations) remains in the header file.

wenqing: Thanks. Fixed accordingly by amending the last commit.

PR #2584, opened 2019-07-31
[CL] Bugfix dropped_item_ids

joboog:
Fixes as data I/O phreeqc-ogs


renchao-lu:
Good catch. Thanks. @joboog

endJunction:
A test would be great, because later someone could change the code and reintroduce the same bug.
If an existing test can be modified, that would be sufficient.

renchao-lu:
I will modify the existing KineticReactant Benchmark to add two more kinetic reactants, as I need to reuse the test.dat file Vanessa and you created instead of the revised_llnl.dat file anyway. @joboog

renchao-lu:
Done! 
1) Compressed the original thermodynamic database 
2) Modified the KineticReactant Benchmark for testing
Please cherry-pick the latest commit from my branch `feat_bugfix_CL` (https://github.com/renchao-lu/ogs6/tree/feat_bugfix_CL). @joboog

montoyav:
I have had a fast look, and I see that all the products are considered as "PHASES" which means immobile mass. Probably, this was the intention (and numerically is correct), but I think this is not clear from the benchmark description.

joboog:
@montoyav : there is no description for the test KineticReactant. Do not mix up with the test `KineticReactant_AllAsComponents` that is described on the webpage.

montoyav:
OK, I see. My mistake.

endJunction:
@renchao-lu When OK from your side, please rebase and merge.

PR #2586, opened 2019-08-01
Expansion of GoCAD interface

rinkk:
Can now also parse GoCAD PLine data.

Also added a check if multiple data sets in the same file have the same name and if so append a UID, so files don't get overwritten on export.

### DIFF
--- a/Applications/FileIO/GocadIO/GocadAsciiReader.cpp
+++ b/Applications/FileIO/GocadIO/GocadAsciiReader.cpp
@@ -70,7 +70,7 @@ bool GocadAsciiReader::readFile(
     std::ifstream in(file_name.c_str());
     if (!in.is_open())
     {
-        ERR("GocadTSurfaceReader::readFile(): Could not open file %s.",
+        ERR("GocadAsciiReader::readFile(): Could not open file %s.",
             file_name.c_str());
         return false;
     }
@@ -190,7 +190,7 @@ MeshLib::Mesh* GocadAsciiReader::readData(std::ifstream& in,
         }
         else
         {
-            WARN("GocadTSurfaceReader::readMesh() - Unknown keyword found: %s",
+            WARN("GocadAsciiReader::readMesh() - Unknown keyword found: %s",
                  line.c_str());
         }
     }
@@ -371,7 +371,7 @@ bool GocadAsciiReader::parseLine(
         }
         else
         {
-            WARN("GocadTSurfaceReader::parseLine() - Unknown keyword found: %s", line.c_str());
+            WARN("GocadAsciiReader::parseLine() - Unknown keyword found: %s", line.c_str());
         }
     }
     ERR("%s", eof_error.c_str());
@@ -417,7 +417,7 @@ bool GocadAsciiReader::parseSurface(
         }
         else
         {
-            WARN("GocadTSurfaceReader::parseSurface() - Unknown keyword found: %s", line.c_str());
+            WARN("GocadAsciiReader::parseSurface() - Unknown keyword found: %s", line.c_str());
         }
     }
     ERR("%s", eof_error.c_str());
@@ -467,7 +467,7 @@ bool GocadAsciiReader::parseNodes(
         if (!(line.substr(0, 4) == "VRTX" || line.substr(0, 5) == "PVRTX" ||
               line.substr(0, 4) == "ATOM"))
         {
-            WARN("GocadTSurfaceReader::parseNodes() - Unknown keyword found: %s", line.c_str());
+            WARN("GocadAsciiReader::parseNodes() - Unknown keyword found: %s", line.c_str());
             continue;
         }
 
--- a/Applications/FileIO/GocadIO/GocadAsciiReader.cpp
+++ b/Applications/FileIO/GocadIO/GocadAsciiReader.cpp
@@ -13,10 +13,6 @@
 
 #include <fstream>
 
-#include <boost/uuid/uuid.hpp>
-#include <boost/uuid/uuid_generators.hpp>
-#include <boost/uuid/uuid_io.hpp>
-
 #include "Applications/FileIO/GocadIO/CoordinateSystem.h"
 #include "BaseLib/FileTools.h"
 #include "BaseLib/StringTools.h"
@@ -44,18 +40,16 @@ GocadAsciiReader::GocadAsciiReader(GocadDataType const t)
 /// another dataset with the same name exists.
 void checkMeshNames(std::vector<std::unique_ptr<MeshLib::Mesh>> const& meshes)
 {
-    boost::uuids::random_generator generator;
     std::size_t const n_meshes = meshes.size();
     for (std::size_t i=0; i<n_meshes; ++i)
     {
-        std::string name = meshes[i]->getName();
+        std::string const& name = meshes[i]->getName();
         for (std::size_t j=i+1; j<n_meshes; ++j)
         {
             if (meshes[j]->getName() == name)
             {
-                boost::uuids::uuid mesh_id = generator();
-                std::string id_str = boost::lexical_cast<std::string>(mesh_id);
-                meshes[i]->setName(name + "--" + id_str);
+                std::string const id_str = std::to_string(meshes[j]->getID());
+                meshes[i]->setName(name + "--importID-" + id_str);
                 break;
             }
         }
--- a/Applications/FileIO/GocadIO/GocadAsciiReader.cpp
+++ b/Applications/FileIO/GocadIO/GocadAsciiReader.cpp
@@ -28,7 +28,6 @@ namespace Gocad
 {
 namespace GocadAsciiReader
 {
-
 enum class NodeType
 {
     UNSPECIFIED,
@@ -98,8 +97,8 @@ bool skipToEND(std::ifstream& in)
 /// Checks if current line is a designated keyword for a GoCAD data set
 bool isKeyword(DataType const t, std::string const& line)
 {
-    std::size_t str_length = DataType2Str(t).length();
-    return (line.substr(0, str_length) == DataType2Str(t));
+    std::size_t str_length = dataType2String(t).length();
+    return (line.substr(0, str_length) == dataType2String(t));
 }
 
 /// Checks if a GoCAD data set begins at the current stream position.
@@ -217,14 +216,16 @@ bool parseProperties(std::ifstream& in,
             return true;
         }
 
-        // Currently all property parameters except array name and size are ignored.
+        // Currently all property parameters except array name and size are
+        // ignored.
         if (key == "ESIZES")
         {
             std::vector<std::string> prop_size = BaseLib::splitString(line);
 
             if (names.size() != prop_size.size())
             {
-                ERR("Error: Number of PROPERTY-names (%d) does not match number of ESIZES (%d)",
+                ERR("Error: Number of PROPERTY-names (%d) does not match "
+                    "number of ESIZES (%d)",
                     names.size(), prop_size.size());
                 return false;
             }
@@ -261,7 +262,8 @@ bool parseNodes(std::ifstream& in,
 {
     NodeType t = NodeType::UNSPECIFIED;
     double value;
-    std::vector<std::string> const array_names = mesh_prop.getPropertyVectorNames();
+    std::vector<std::string> const array_names =
+        mesh_prop.getPropertyVectorNames();
     std::streampos pos = in.tellg();
     std::string line;
     while (std::getline(in, line))
@@ -286,7 +288,8 @@ bool parseNodes(std::ifstream& in,
         if (!(line.substr(0, 4) == "VRTX" || line.substr(0, 5) == "PVRTX" ||
               line.substr(0, 4) == "ATOM"))
         {
-            WARN("GocadAsciiReader::parseNodes() - Unknown keyword found: %s", line.c_str());
+            WARN("GocadAsciiReader::parseNodes() - Unknown keyword found: %s",
+                 line.c_str());
             continue;
         }
 
@@ -314,7 +317,8 @@ bool parseNodes(std::ifstream& in,
             std::size_t ref_id;
             std::string keyword;
             sstr >> keyword >> new_id >> ref_id;
-            nodes.push_back(new MeshLib::Node(nodes[ref_id]->getCoords(), new_id));
+            nodes.push_back(
+                new MeshLib::Node(nodes[ref_id]->getCoords(), new_id));
         }
         node_id_map[nodes.back()->getID()] = nodes.size() - 1;
         pos = in.tellg();
@@ -330,7 +334,8 @@ bool parseLineSegments(std::ifstream& in,
                        std::map<std::size_t, std::size_t> const& node_id_map,
                        MeshLib::Properties& mesh_prop)
 {
-    MeshLib::PropertyVector<int>& mat_ids = *mesh_prop.getPropertyVector<int>(mat_id_name);
+    MeshLib::PropertyVector<int>& mat_ids =
+        *mesh_prop.getPropertyVector<int>(mat_id_name);
     int current_mat_id(0);
     if (!mat_ids.empty())
     {
@@ -406,7 +411,8 @@ bool parseLine(std::ifstream& in,
         {
             return true;
         }
-        WARN("GocadAsciiReader::parseLine() - Unknown keyword found: %s", line.c_str());
+        WARN("GocadAsciiReader::parseLine() - Unknown keyword found: %s",
+             line.c_str());
     }
     ERR("%s", eof_error.c_str());
     return false;
@@ -468,12 +474,11 @@ bool parseElements(std::ifstream& in,
 }
 
 /// Parses the surface information (nodes, triangles, properties)
-bool parseSurface(
-    std::ifstream& in,
-    std::vector<MeshLib::Node*>& nodes,
-    std::vector<MeshLib::Element*>& elems,
-    std::map<std::size_t, std::size_t>& node_id_map,
-    MeshLib::Properties& mesh_prop)
+bool parseSurface(std::ifstream& in,
+                  std::vector<MeshLib::Node*>& nodes,
+                  std::vector<MeshLib::Element*>& elems,
+                  std::map<std::size_t, std::size_t>& node_id_map,
+                  MeshLib::Properties& mesh_prop)
 {
     if (!parseNodes(in, nodes, node_id_map, mesh_prop))
     {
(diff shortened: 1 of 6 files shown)


Applications/FileIO/GocadIO/GocadAsciiReader.h
```diff
@@ -26,31 +26,44 @@ namespace FileIO
 {
 namespace Gocad
 {
-class GocadTSurfaceReader final
+
+enum class GOCAD_DATA_TYPE
```
endJunction:
```suggestion
enum class GocadDataType
```
Similar to the enums in MeshLib, for example: MeshElemType, CellType, MeshQualityType, etc.

Applications/FileIO/GocadIO/GocadAsciiReader.cpp
```diff
@@ -272,7 +340,45 @@ bool GocadTSurfaceReader::parseProperties(std::ifstream& in,
...
+    std::string line;
+    while (std::getline(in, line))
+    {
+        std::vector<std::string> str = BaseLib::splitString(line);
+        if (str[0] == "ILINE")
+        {
+            parseLine(in, nodes, elems, node_id_map, mesh_prop);
+            return true;
+        }
+        else if (line == "END")
```
endJunction: else after a return is not needed...

Applications/FileIO/GocadIO/GocadAsciiReader.cpp
```diff
@@ -27,11 +32,38 @@ namespace Gocad
...
+    for (std::size_t i=0; i<n_meshes; ++i)
+    {
+        std::string name = meshes[i]->getName();
+        for (std::size_t j=i+1; j<n_meshes; ++j)
+        {
+            if (meshes[j]->getName() == name)
+            {
+                boost::uuids::uuid mesh_id = generator();
+                std::string id_str = boost::lexical_cast<std::string>(mesh_id);
+                meshes[i]->setName(name + "--" + id_str);
```
endJunction: each mesh object has a globally unique id. See https://github.com/ufz/ogs/blob/master/MeshLib/Mesh.h#L41

Applications/FileIO/GocadIO/GocadAsciiReader.cpp
```diff
@@ -108,7 +160,21 @@ MeshLib::Mesh* GocadTSurfaceReader::readMesh(std::ifstream& in,
                 return nullptr;
             }
         }
-        else if (str[0] == "TFACE")
+        else if (type == GOCAD_DATA_TYPE::PLINE && str[0] == "ILINE")
```
endJunction: This is copy of the following if for TSURF. Please avoid code duplications.

Applications/FileIO/GocadIO/GocadAsciiReader.cpp
```diff
@@ -397,7 +504,62 @@ bool GocadTSurfaceReader::parseNodes(
...
 
-bool GocadTSurfaceReader::parseElements(
+bool GocadAsciiReader::parseLineSegments(
+    std::ifstream& in,
+    std::vector<MeshLib::Node*>& nodes,
+    std::vector<MeshLib::Element*>& elems,
+    std::map<std::size_t, std::size_t> const& node_id_map,
+    MeshLib::Properties& mesh_prop)
+{
+    std::string keyword;
```
endJunction: Please constrain variables to minimum possible scope.

Applications/FileIO/GocadIO/GocadAsciiReader.cpp
```diff
@@ -452,7 +614,7 @@ bool GocadTSurfaceReader::parseElements(
     return false;
 }
 
-bool GocadTSurfaceReader::skipModel3d(std::ifstream& in) const
+bool GocadAsciiReader::skipToEND(std::ifstream& in) const
```
endJunction:
```suggestion
bool GocadAsciiReader::skipToEnd(std::ifstream& in) const
```

Applications/FileIO/GocadIO/GocadAsciiReader.h
```diff
@@ -26,31 +26,44 @@ namespace FileIO
...
+    TSURF,
+    MODEL3D,
+    ALL
+};
+
+class GocadAsciiReader final
 {
 public:
     /**
      * Constructor takes as argument the Gocad .sg text file.
```
endJunction: Wrong comment. Default ctor does not take any arguments.

Applications/FileIO/GocadIO/GocadAsciiReader.h
```diff
@@ -26,31 +26,44 @@ namespace FileIO
...
-    explicit GocadTSurfaceReader();
+    explicit GocadAsciiReader();
+
+    explicit GocadAsciiReader(GOCAD_DATA_TYPE const t);
 
-    GocadTSurfaceReader(GocadTSurfaceReader&& src) = delete;
-    GocadTSurfaceReader(GocadTSurfaceReader const& src) = delete;
-    GocadTSurfaceReader& operator=(GocadTSurfaceReader&& rhs) = delete;
-    GocadTSurfaceReader& operator=(GocadTSurfaceReader const& rhs) = delete;
+    GocadAsciiReader(GocadAsciiReader&& src) = delete;
```
endJunction: Why are these `delete`d constructors/assignments needed?

Applications/Utils/FileConverter/GocadTSurfaceReader.cpp
```diff
@@ -51,21 +51,46 @@ int main(int argc, char* argv[])
...
+    cmd.add(export_lines_arg);
+
+    TCLAP::SwitchArg export_surfaces_arg(
+        "s", "surfaces-only",
+        "if set, only TSurf datasets will be parsed from the input file");
+    cmd.add(export_surfaces_arg);
+
     cmd.parse(argc, argv);
 
+    if (export_lines_arg.isSet() && export_surfaces_arg.isSet())
```
endJunction: Would it be possible to use tclap's functionality here? http://tclap.sourceforge.net/manual.html#XOR

endJunction:
clang-format

Applications/FileIO/GocadIO/GocadAsciiReader.cpp
```diff
@@ -27,11 +32,38 @@ namespace Gocad
...
+/// A GoCAD file may contain multiple datasets with the same name. To avoid
+/// conflicts when writing meshes, a unique id is appended to the mesh name if
+/// another dataset with the same name exists.
+void checkMeshNames(std::vector<std::unique_ptr<MeshLib::Mesh>> const& meshes)
 {
+    boost::uuids::random_generator generator;
+    std::size_t const n_meshes = meshes.size();
+    for (std::size_t i=0; i<n_meshes; ++i)
+    {
+        std::string name = meshes[i]->getName();
```
TomFischer: I think the copy is not necessary here. It should be possible to store a `std::string const&`.

Applications/FileIO/GocadIO/GocadAsciiReader.cpp
```diff
@@ -358,9 +467,7 @@ bool GocadTSurfaceReader::parseNodes(
         if (!(line.substr(0, 4) == "VRTX" || line.substr(0, 5) == "PVRTX" ||
               line.substr(0, 4) == "ATOM"))
         {
-            WARN(
-                "GocadTSurfaceReader::parseNodes() - Unknown keyword found: %s",
-                line.c_str());
+            WARN("GocadTSurfaceReader::parseNodes() - Unknown keyword found: %s", line.c_str());
```
TomFischer: GocadTSurfaceReaderb -> GocadAsciiReader?

Applications/FileIO/GocadIO/GocadAsciiReader.h
```diff
@@ -26,31 +26,44 @@ namespace FileIO
...
     /// Reads one mesh contained in the file (there may be more than one!)
-    MeshLib::Mesh* readMesh(std::ifstream& in, std::string& mesh_name);
+    MeshLib::Mesh* readData(std::ifstream& in, GOCAD_DATA_TYPE const& type, std::string& mesh_name);
 
     /// Checks if the current line is a comment
     bool isCommentLine(std::string const& str) const;
 
     /// Checks if a TSurf identifier is found at the current stream position.
-    bool TSurfaceFound(std::ifstream& in) const;
+    GOCAD_DATA_TYPE datasetFound(std::ifstream& in) const;
```
TomFischer: Docu needs an update.

rinkk: :heavy_check_mark:

Applications/FileIO/GocadIO/GocadAsciiReader.cpp
rinkk: :heavy_check_mark:

Applications/FileIO/GocadIO/GocadAsciiReader.h
rinkk: c/p from the sgrid reader. I figured it's supposed to be like that since an object copy doesn't really make sense.

Applications/FileIO/GocadIO/GocadAsciiReader.h
rinkk: :heavy_check_mark:

Applications/FileIO/GocadIO/GocadAsciiReader.cpp
rinkk: I used this notation because the tag in the gocad-file is actually called ```END```. If it's called "skipToEnd" it suggests that it jumps to the end of the file?

Applications/Utils/FileConverter/GocadTSurfaceReader.cpp
rinkk: I've added 

```
cmd.xorAdd(export_lines_arg, export_surfaces_arg);
```

but this causes the application to crash?!

Also, I am not sure if XOR means that one of these arguments *has* to set. For my app that isn't the case.

Applications/FileIO/GocadIO/GocadAsciiReader.h
endJunction: Just drop them.

Applications/FileIO/GocadIO/GocadAsciiReader.h
rinkk: :heavy_check_mark:

Applications/FileIO/GocadIO/GocadAsciiReader.h
rinkk: :heavy_check_mark:

Applications/FileIO/GocadIO/GocadAsciiReader.cpp
rinkk: :heavy_check_mark:

Applications/FileIO/GocadIO/GocadAsciiReader.cpp
rinkk: :heavy_check_mark:

Applications/FileIO/GocadIO/GocadAsciiReader.cpp
rinkk: :heavy_check_mark:

Applications/FileIO/GocadIO/GocadAsciiReader.cpp
rinkk: :heavy_check_mark:

Applications/FileIO/GocadIO/GocadAsciiReader.cpp
rinkk: :heavy_check_mark:

Applications/FileIO/GocadIO/GocadAsciiReader.cpp
```diff
@@ -142,35 +191,38 @@ bool GocadTSurfaceReader::TSurfaceFound(std::ifstream& in) const
...
+
+        if (line.substr(0, 10) == "GOCAD VSet")
         {
-            return true;
-        // No idea why this is allowed in a *.ts file.
-        // It should be a whole different file type.
+            return GocadDataType::VSET;
         }
-        if (line.substr(0, 13) == "GOCAD Model3d")
+        else if (line.substr(0, 11) == "GOCAD PLine")
```
endJunction: I'd move this string to enum conversion to the place where the enum is defined.

Also no else after return is needed.

Applications/FileIO/GocadIO/GocadAsciiReader.cpp
endJunction: The duplication is indeed resolved, but now the mesh creation logic is mixed up with the parsing of the cases. It is difficult to add new cases there.
It's up to you, but I'd have written something like
```c++
auto parse_mesh = [&](auto node_element_parser) {
    std::vector<MeshLib::Node*> nodes;
    std::vector<MeshLib::Element*> elems;
    std::map<std::size_t, std::size_t> node_id_map;
    if (!node_element_parser(in, nodes, elems, node_id_map, mesh_prop))
    {
        ERR(...);
        return nullptr;
    }
    return std::make_unique<MeshLib::Mesh>(mesh_name, ...);
};
```
And in the conditions:
```c++
if (type == TFACE)
{
    return parse_mesh(parseSurface);
}
if (type == PLINE)
{
    return parse_mesh(parseLine);
}
```

Applications/FileIO/GocadIO/GocadAsciiReader.cpp
endJunction: Same goes for `data` array.

Applications/FileIO/GocadIO/GocadAsciiReader.cpp
```diff
@@ -397,14 +485,68 @@ bool GocadTSurfaceReader::parseNodes(
...
+    {
+        if (line.empty() || isCommentLine(line))
+        {
+            continue;
+        }
+        if (line.substr(0, 3) == "SEG")
+        {
+            std::stringstream sstr(line);
+            std::string keyword;
+            sstr >> keyword >> data[0] >> data[1];
```
endJunction: Also not clear is why `data` is 3 long, but only two elements are used.

Applications/FileIO/GocadIO/GocadAsciiReader.cpp
```diff
@@ -27,48 +28,91 @@ namespace Gocad
 const std::string mat_id_name = "MaterialIDs";
 const std::string eof_error = "Error: Unexpected end of file.";
 
-GocadTSurfaceReader::GocadTSurfaceReader()
+GocadAsciiReader::GocadAsciiReader()
+: _export_type(GocadDataType::ALL) {}
+
+GocadAsciiReader::GocadAsciiReader(GocadDataType const t)
```
endJunction:
```suggestion
GocadAsciiReader::GocadAsciiReader(GocadDataType const t = GocadDataType::ALL)
```
... then the default ctor is not needed.

Applications/FileIO/GocadIO/GocadAsciiReader.cpp
```diff
@@ -27,48 +28,91 @@ namespace Gocad
...
+            skipToEND(in);
+            continue;
+        }
+
+        if (type == GocadDataType::VSET || type == GocadDataType::MODEL3D)
+        {
+            if (!skipToEND(in))
+            {
+                std::string const t = (type == GocadDataType::VSET) ? "VSet" : "Model3D";
+                ERR("Parsing of type %s is not implemented. Skipping section.", t);
```
endJunction:
```suggestion
                ERR("Parsing of type %s is not implemented. Skipping section.", t.c_str());
```

Applications/FileIO/GocadIO/GocadAsciiReader.cpp
rinkk: Honestly, I have no idea what happens there. And I know that it's difficult to add cases now, that's why it was written before such that cases were independent of each other.

Applications/FileIO/GocadIO/GocadAsciiReader.cpp
rinkk: :heavy_check_mark:

Applications/FileIO/GocadIO/GocadAsciiReader.cpp
rinkk: :heavy_check_mark:

Applications/FileIO/GocadIO/GocadAsciiReader.cpp
rinkk: :heavy_check_mark:

Applications/FileIO/GocadIO/GocadAsciiReader.cpp
rinkk: :heavy_check_mark:

Applications/FileIO/GocadIO/GocadAsciiReader.cpp
rinkk: Do you suggest that I should implement a enum -> string conversion for this one use?

The return does not follow "else" but "while" and is required in case the end of the file is reached.

Applications/FileIO/GocadIO/GocadAsciiReader.cpp
rinkk: To be more specific, I'd very much prefer the code the way it was originally written. That'd be six lines of duplicate code, which isn't a big issue in my opinion if it means the code is easier to read.

TomFischer:
The file `GocadEnums.h` is missing.
```
GocadAsciiReader.h:15:10: fatal error: 'Applications/FileIO/GocadIO/GocadEnums.h' file not found
```

Applications/FileIO/GocadIO/GocadEnums.h
```diff
@@ -0,0 +1,37 @@
...
+    UNDEFINED,
+    VSET,
+    PLINE,
+    TSURF,
+    MODEL3D,
+    ALL
+};
+
+/// Given a Gocad DataType this returns the appropriate string.
+std::string DataType2Str(DataType const t);
```
endJunction:
```suggestion
std::string dataType2String(DataType const t);
```

Applications/FileIO/GocadIO/GocadEnums.cpp
```diff
@@ -0,0 +1,69 @@
...
+    }
+    if (t == DataType::TSURF)
+    {
+        return "surface";
+    }
+    if (t == DataType::MODEL3D)
+    {
+        return "model";
+    }
+    return "[all data]";
```
endJunction: `all data` *vs.* `all types` above. Maybe wrong.

PR #2588, opened 2019-08-02
Time dependent heterogeneous parameter implementation

TomFischer:
Time dependent heterogeneous parameter implementation that can be used for instance for time dependent heterogeneous boundary conditions and source terms.

1. [x] Feature description is added to the [changelog](https://github.com/ufz/ogs/wiki/Release-notes-6.2.1)
2. [x] Tests covering your feature are added
3. [x] New feature is documented


ParameterLib/Parameter.cpp
```diff
@@ -80,6 +81,13 @@ std::unique_ptr<ParameterBase> createParameter(
         auto param = createMeshNodeParameter(name, config, mesh);
         return param;
     }
+    if (type == "TimeDependentHeterogeneousParameter")
+    {
+        INFO("TimeDependentHeterogeneousParameter: %s", name.c_str());
+        auto param =
```
endJunction:
```suggestion
        return
```

ParameterLib/TimeDependentHeterogeneousParameter.cpp
```diff
@@ -0,0 +1,152 @@
...
+#include "BaseLib/Error.h"
+#include "Utils.h"
+
+namespace ParameterLib
+{
+TimeDependentHeterogeneousParameter::TimeDependentHeterogeneousParameter(
+    std::string name,
+    std::vector<PairTimeParameterName>
+        time_parameter_name_mapping)
+    : Parameter<double>(name, nullptr),
```
endJunction:
```suggestion
    : Parameter<double>(std::move(name), nullptr),
```

ParameterLib/TimeDependentHeterogeneousParameter.h
```diff
@@ -0,0 +1,57 @@
...
+{
+public:
+    using PairTimeParameterName = std::pair<double, std::string>;
+    using PairTimeParameter = std::pair<double, Parameter<double> const* const>;
+
+    TimeDependentHeterogeneousParameter(std::string name,
+                                        std::vector<PairTimeParameterName>
+                                            time_parameter_name_mapping);
+
+    ~TimeDependentHeterogeneousParameter() override = default;
```
endJunction:
```suggestion
```

TomFischer: Removed.

ParameterLib/TimeDependentHeterogeneousParameter.cpp
TomFischer: Added.

ParameterLib/Parameter.cpp
TomFischer: Changed.

endJunction:
The .mp4 file must be 'lfs'ed.

wenqing:
Just curious: can it be possible to realize the same functionality by BC/ST * f(t), where f(t) is a time factor function.

TomFischer:
> Just curious: can it be possible to realize the same functionality by BC/ST * f(t), where f(t) is a time factor function.

Do you mean something like CurveScaledParameter? I think Erik was using CurveScaledParameters before, but it was very cumbersome since he needed many of them to reproduce the spatial heterogeneity. The implementation in this PR needs only one (possible spatial heterogeneous) parameter for each time step.

wenqing:
@TomFischer  I see.

PR #2589, opened 2019-08-05
[git] Set automatic line endings on all files

bilke:
Ensures native file endings on all platforms.

Should fix an issue where the GocadTSurface_Mesh_Test failed
on Windows as the reader cannot read .ts files on Windows
with Linux line endings.


endJunction:
The [gitattributes(5)](https://mirrors.edge.kernel.org/pub/software/scm/git/docs/gitattributes.html#_end_of_line_conversion) man page says something about renormalization of all files if central repo is used:

> Note
When text=auto conversion is enabled in a cross-platform project using push and pull to a central repository the text files containing CRLFs should be normalized.
From a clean working directory:

> $ echo "* text=auto" >.gitattributes
$ git add --renormalize .
$ git status        # Show files that will be normalized
$ git commit -m "Introduce end-of-line normalization"
If any files that should not be normalized show up in git status, unset their text attribute before running git add -u.
   manual.pdf      -text
Conversely, text files that Git does not detect can have normalization enabled manually.
  weirdchars.txt  text

PR #2590, opened 2019-08-05
RichardsMechanics: mass lumping

endJunction:


ProcessLib/RichardsMechanics/RichardsMechanicsFEM-impl.h
```diff
@@ -553,6 +560,11 @@ void RichardsMechanicsLocalAssembler<ShapeFunctionDisplacement,
             dNdx_p.transpose() * rho_LR * k_rel * rho_Ki_over_mu * b * w;
     }
 
+    if (_process_data.apply_mass_lumping)
+    {
+        storage_p = storage_p.colwise().sum().eval().asDiagonal();
```
wenqing: In future, the name of this term is better to be unified both here and where in assemble().

PR #2600, opened 2019-08-08
HM Compressible Fluids

FZill:
* Further updated the mass formulation to include the compressibility of the fluid
* specific storage is now not needed anymore and was removed
* added 3 tests which check that the pressure dependent compressibility (in the case for ideal gases) works as intended

the next step is to fetch the fluid data from MPL


ProcessLib/HydroMechanics/CreateHydroMechanicsProcess.cpp
```diff
@@ -183,18 +174,49 @@ std::unique_ptr<Process> createHydroMechanicsProcess(
     }
 
     // Reference temperature
-    const auto& reference_temperature =
+    double const reference_temperature =
         //! \ogs_file_param{prj__processes__process__HYDRO_MECHANICS__reference_temperature}
-        config.getConfigParameter<double>(
-            "reference_temperature", std::numeric_limits<double>::quiet_NaN());
+        config.getConfigParameter<double>("reference_temperature", 1);
```
endJunction: The default 1 (and similar 0 in other parameters) should either be replaced by NaN (as before), or (because this are required parameters) the be omitted.
No-implicit-magic-values-policy ;)

ProcessLib/HydroMechanics/CreateHydroMechanicsProcess.cpp
```diff
@@ -183,18 +174,49 @@ std::unique_ptr<Process> createHydroMechanicsProcess(
...
+        //! \ogs_file_param_special{prj__processes__process__HYDRO_MECHANICS__fluid_compressibility}
+        config.getConfigParameter<double>("fluid_compressibility", 0);
+    DBUG("Use '%s' as fluid compressibility parameter.", "fluid_compressibility");
+
+    auto const& fluid_type_str =
+        //! \ogs_file_param{prj__processes__process__HYDRO_MECHANICS__fluid_type}
+        config.getConfigParameter<std::string>("fluid_type", "incompressible_fluid");
+
+    unsigned const fluid_type =
+        (fluid_type_str == "incompressible_fluid") ? 0 :
```
endJunction: must become an `enum class`.
Don't use `unsigned`.

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
```diff
@@ -152,6 +152,10 @@ void HydroMechanicsLocalAssembler<ShapeFunctionDisplacement,
         ShapeMatricesTypePressure::NodalMatrixType::Zero(pressure_size,
                                                          pressure_size);
 
+    typename ShapeMatricesTypePressure::NodalMatrixType higher_order_p =
```
endJunction: What do you mean by higher order?

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
```diff
@@ -164,7 +168,18 @@ void HydroMechanicsLocalAssembler<ShapeFunctionDisplacement,
             pressure_size, displacement_size>::Zero(pressure_size,
                                                     displacement_size);
 
+    MaterialLib::Solids::MechanicsBase<DisplacementDim> const& solid_material =
+        *_process_data.solid_materials[0];
+
+    auto linear_elastic_mp =
+        static_cast<MaterialLib::Solids::LinearElasticIsotropic<
```
endJunction: This, of course, is invalid for all other materials.

Tests/Data/HydroMechanics/StaggeredScheme/InjectionProduction1D/RerefenceSolutionByMonolithicScheme/InjectionProduction1DMono.prj
```diff
@@ -36,6 +35,8 @@
                 <secondary_variable type="static" internal_name="velocity" output_name="velocity"/>
             </secondary_variables>
             <specific_body_force>0 0</specific_body_force>
+            <fluid_type>compressible_fluid</fluid_type>
+            <fluid_compressibility>3.33333333333e-5</fluid_compressibility>
```
endJunction: Just curious, what is this number? Specific fluid's value?

norihiro-w:
The specific storage has been used as a input parameter because one can easily change the storage models, e.g. Ss=phi * Cfluid, or Ss=phi * Cflud + (1-phi) * Cs, or Ss = phi * Cflud + (biot - phi) * Cbulk. As far as I understood, this PR supports only Ss=phi * Cfluid, which is rarely used in practical studies in my opinion. Is there any plan to support other models?

nagelt:
@norihiro-w This implementation uses Ss = phi * Cflud + (biot - phi) * Cbulk. As you say, the idea is to support different formulations (incl. a direct specification of the value) in the future. This one is the one consistent with the current model derivation and taken as a standard for now.

nagelt:
@endJunction Florian removed K_SR from the input file parameter definition. With Biot, K_S = f(E, nu), and K_SR the model was overparameterized (only two of the three are independent in this formulation) and it was easy to produce unphysical results, if the definitions don't match. So right now, the model calculates K_SR from K_S and alpha. Most of our models (currently all) contain E and nu or even K_S, so we could think about a general function that returns the bulk modulus. Or we decide to give the user the freedom back and live with the GIGO principle ;-)

norihiro-w:
@nagelt you are right. sorry i overlooked it. 

BTW why you have `beta_p` and `2.0 * rho_fr`  here?
```
 higher_order_p += rho_fr * beta_p * dNdx_p.transpose() * K_over_mu *
                          (dNdx_p * p - 2.0 * rho_fr * b) * N_p * w;
```

ProcessLib/HydroMechanics/CreateHydroMechanicsProcess.cpp
```diff
@@ -183,18 +174,49 @@ std::unique_ptr<Process> createHydroMechanicsProcess(
...
+        (fluid_type_str == "compressible_fluid") ? 1 :
+        (fluid_type_str == "ideal_gas") ? 2 : 3;
+
+    if (fluid_type == 3)
+    {
+        OGS_FATAL(
+            "The fluid type %s is unavailable.\n"
+            "The available fluid types are ideal_gas and "
```
TomFischer: Above also `compressible_fluid` type is handled, but not mentioned in the error message.

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
```diff
@@ -164,7 +168,18 @@ void HydroMechanicsLocalAssembler<ShapeFunctionDisplacement,
...
+            DisplacementDim> const&>(solid_material)
+            .getMaterialProperties();
+
     double const& dt = _process_data.dt;
+    double const R_s = _process_data.specific_gas_constant;
+    double const T = _process_data.reference_temperature;
```
TomFischer: Maybe `T_ref` to make clear later on that a reference temperature is meant?

ProcessLib/HydroMechanics/HydroMechanicsProcessData.h
```diff
@@ -104,8 +105,19 @@ struct HydroMechanicsProcessData
...
+    /// ParameterLib::Parameter<double>.
+    double const fluid_compressibility;
+
+    /// Reference Temperature. A scalar quantity.
     double const reference_temperature;
 
+    /// Specific gas constant. A scalar quantity.
+    double const specific_gas_constant;
+
+    /// Fluid type. 0 = incompressible_fluid, 1 = ideal_gas
```
TomFischer: Did you forget to mention the compressible_fluid type (somewhere it is handled in the code) to mention in the docu or is there a reason that it is not documented here?

endJunction:
@nagelt Regarding the E, nu; Could be implemented in the MechanicsBase interface.

nagelt:
@norihiro-w I think the two comes from a \rho_F^2 in the formulation (gravity term in mass flux) ... but the rho_FR may be too many (just have 2\beta_p) ... @FZill could you check again?

FZill:
@nagelt @norihiro-w I checked the term again and it should be correct, the units also add up correctly.

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
FZill: Yes that's better

nagelt:
@FZill I just checked our equations. You're right, the term is fine!

norihiro-w:
thank you for the check!

ProcessLib/HydroMechanics/CreateHydroMechanicsProcess.cpp
FZill: is now implemented

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
FZill: changed the name to add_p_derivative (additional_p_derivative) which suits it better.
origins in the p_fr derivative of rho_fr

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
FZill: is removed and now uses getBulkModulus

Tests/Data/HydroMechanics/StaggeredScheme/InjectionProduction1D/RerefenceSolutionByMonolithicScheme/InjectionProduction1DMono.prj
FZill: this is just the calculated fluid compressibility to get the same specific storage as before. In a later Version when there is also the possibility to directly specify the specific storage this can be changed back

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
endJunction: `drho_fr_dp_fr` ?

FZill: Yes that would work too, but it probably will be extended with extra terms, when pressure and stress dependent viscosity gets implemented.

endJunction:
@FZill Have a look at warnings on this page: https://jenkins.opengeosys.org/job/ufz/job/ogs/job/PR-2600/10/

ProcessLib/HydroMechanics/CreateHydroMechanicsProcess.cpp
```diff
@@ -183,18 +174,64 @@ std::unique_ptr<Process> createHydroMechanicsProcess(
     }
 
     // Reference temperature
-    const auto& reference_temperature =
+    double const reference_temperature =
         //! \ogs_file_param{prj__processes__process__HYDRO_MECHANICS__reference_temperature}
         config.getConfigParameter<double>(
             "reference_temperature", std::numeric_limits<double>::quiet_NaN());
+    DBUG("Use '%s' as reference temperature.", "reference_temperature");
```
TomFischer: You may want to output the value here?

```suggestion
    DBUG("Use '%d' as reference temperature.", reference_temperature);
```

ProcessLib/HydroMechanics/CreateHydroMechanicsProcess.cpp
```diff
@@ -183,18 +174,64 @@ std::unique_ptr<Process> createHydroMechanicsProcess(
...
+
+    //Specific gas constant
+    double const specific_gas_constant =
+        //! \ogs_file_param{prj__processes__process__HYDRO_MECHANICS__specific_gas_constant}
+        config.getConfigParameter<double>(
+            "specific_gas_constant", std::numeric_limits<double>::quiet_NaN());
+    DBUG("Use '%s' as specific gas constant.", "specific gas constant");
```
TomFischer:
```suggestion
    DBUG("Use '%f' as specific gas constant.", specific_gas_constant);
```

ProcessLib/HydroMechanics/CreateHydroMechanicsProcess.cpp
```diff
@@ -183,18 +174,64 @@ std::unique_ptr<Process> createHydroMechanicsProcess(
...
+
+    // Fluid compressibility
+    double const fluid_compressibility =
+        //! \ogs_file_param{prj__processes__process__HYDRO_MECHANICS__fluid_compressibility}
+        config.getConfigParameter<double>(
+            "fluid_compressibility", std::numeric_limits<double>::quiet_NaN());
+    DBUG("Use '%s' as fluid compressibility parameter.", "fluid_compressibility");
```
TomFischer:
```suggestion
    DBUG("Use '%f' as fluid compressibility parameter.", fluid_compressibility);
```

ProcessLib/HydroMechanics/FluidType.cpp
```diff
@@ -0,0 +1,17 @@
+#include "FluidType.h"
+
+namespace FluidType
+{
+
+FluidType strToFluidType(std::string s)
```
TomFischer: I think it is better to use a `const reference` for passing the parameter.

```suggestion
FluidType strToFluidType(std::string const& s)
```

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
```diff
@@ -197,15 +207,30 @@ void HydroMechanicsLocalAssembler<ShapeFunctionDisplacement,
...
-        auto const rho_fr = _process_data.fluid_density(t, x_position)[0];
+        // TODO get fluid properties from GPML
+        double const rho_fr =
+            (_process_data.fluid_type == FluidType::INCOMPRESSIBLE_FLUID)
+                ? _process_data.fluid_density(t, x_position)[0]
+                : (_process_data.fluid_type == FluidType::COMPRESSIBLE_FLUID)
+                      ? _process_data.fluid_density(t, x_position)[0]
+                      : (_process_data.fluid_type == FluidType::IDEAL_GAS)
+                            ? N_p.dot(p) / (R_s * T_ref)
+                            : 0;
```
TomFischer: I think this could be written more readable using a [lambda expression](https://en.cppreference.com/w/cpp/language/lambda).

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
```diff
@@ -197,15 +207,30 @@ void HydroMechanicsLocalAssembler<ShapeFunctionDisplacement,
...
+        double const beta_p =
+            (_process_data.fluid_type == FluidType::INCOMPRESSIBLE_FLUID)
+                ? 0
+                : (_process_data.fluid_type == FluidType::COMPRESSIBLE_FLUID)
+                      ? _process_data.fluid_compressibility
+                      : (_process_data.fluid_type == FluidType::IDEAL_GAS)
+                            ? 1.0 / N_p.dot(p)
+                            : 0;
```
TomFischer: Again I would use a lambda expression for the initialization of the variable.

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
```diff
@@ -390,16 +426,35 @@ void HydroMechanicsLocalAssembler<ShapeFunctionDisplacement,
...
         auto const& dNdx_p = _ip_data[ip].dNdx_p;
 
-        double const S = _process_data.specific_storage(t, x_position)[0];
         double const K_over_mu =
             _process_data.intrinsic_permeability(t, x_position)[0] /
             _process_data.fluid_viscosity(t, x_position)[0];
         auto const alpha_b = _process_data.biot_coefficient(t, x_position)[0];
-        auto const rho_fr = _process_data.fluid_density(t, x_position)[0];
+        // TODO get fluid properties from GPML
+        double const rho_fr =
```
TomFischer: Looks like the same initialization as above? Maybe you could re-use a small function?

nagelt: > @nagelt Regarding the E, nu; Could be implemented in the MechanicsBase interface.

There is a general way to obtain K_S (drained bulk modulus) that should work for all material models including anisotropic and inelastic ones (in case of anisotropy this will be an upper bound):

In Kelvin-representation we take 1/9 * I^T C I, with I being the second-order identity, and C our consistent material tangent. In 3D tensor notation this would be 1/9 \delta_{ij} C_{ijkl} \delta_{kl}.

Could maybe be a default-implementation unless someone specifies an alternative.

ProcessLib/HydroMechanics/CreateHydroMechanicsProcess.cpp
FZill: In all other DBUG messages the name of the parameter is given, not the value, so I would leave it here the same way.

ProcessLib/HydroMechanics/CreateHydroMechanicsProcess.cpp
```diff
@@ -183,18 +174,65 @@ std::unique_ptr<Process> createHydroMechanicsProcess(
...
     // Reference temperature
-    const auto& reference_temperature =
+    double const reference_temperature =
         //! \ogs_file_param{prj__processes__process__HYDRO_MECHANICS__reference_temperature}
         config.getConfigParameter<double>(
             "reference_temperature", std::numeric_limits<double>::quiet_NaN());
+    DBUG("Use '%s' as reference temperature.", "reference_temperature");
+
+    //Specific gas constant
+    double const specific_gas_constant =
```
wenqing: Since you only use the ideal gas law, you can use PhysicalConstant::IdealGasConstant (in MaterialLib/PhysicalConstant.h) directly.

ProcessLib/HydroMechanics/HydroMechanicsFEM.h
```diff
@@ -413,6 +413,41 @@ class HydroMechanicsLocalAssembler : public LocalAssemblerInterface
         std::vector<double>& local_b_data, std::vector<double>& local_Jac_data,
         LocalCoupledSolutions const& local_coupled_solutions);
 
+    double const getFluidDensity(
```
wenqing: Please move these two functions to HydroMechanicsProcessData.h, and add a comment such that 'It will be removed after the fluid properties are replaced with MPL'

wenqing:
If you have time, you may consider to use MPL library. With that you need:
1. Add  `class IdealGasLawDensity final : public Property`
2. Add a function of getFluidCompressibility {return a constant or return (d rho/dp) /rho} (c.f. MaterialLib/MPL/Components/GetThermalExpansivity.h)
3. Replace the fluid (or together with solid) material properties with MPL (c.f    https://github.com/ufz/ogs/blob/master/ProcessLib/ThermoHydroMechanics/ThermoHydroMechanicsFEM-impl.h)

And, a little bit more time is need to change the project files. 

It might be the content of a PR in future.

ProcessLib/HydroMechanics/CreateHydroMechanicsProcess.cpp
FZill: As discussed, this won't gain us anything since we still need to have the specific gas constant for the calculation.

ProcessLib/HydroMechanics/CreateHydroMechanicsProcess.cpp
TomFischer: Okay, if you want output the parameter name and it is 'hard coded' you could put it directly in the message.
```suggestion
    DBUG("Use 'fluid_compressibility' as fluid compressibility parameter.");
```

FZill:
@wenqing Using a Material Library is our plan for a coming PR. This was just an intermediate step.

MaterialLib/Fluid/FluidType/FluidType.cpp
```diff
@@ -0,0 +1,47 @@
+#include "FluidType.h"
```
FZill: copyright

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
```diff
@@ -390,16 +412,24 @@ void HydroMechanicsLocalAssembler<ShapeFunctionDisplacement,
         auto const& N_p = _ip_data[ip].N_p;
         auto const& dNdx_p = _ip_data[ip].dNdx_p;
 
-        double const S = _process_data.specific_storage(t, x_position)[0];
         double const K_over_mu =
             _process_data.intrinsic_permeability(t, x_position)[0] /
             _process_data.fluid_viscosity(t, x_position)[0];
         auto const alpha_b = _process_data.biot_coefficient(t, x_position)[0];
-        auto const rho_fr = _process_data.fluid_density(t, x_position)[0];
+        // TODO get fluid properties from GPML
```
FZill:
```suggestion
        // TODO (FZill) get fluid properties from GPML
```

PR #2602, opened 2019-08-10
Lubby2 incremental form

nagelt:
This is a minor change to Lubby2 switching from rate form to incremental form. Preparation for an upcoming model comparison.


endJunction:
Just curious, what is the advantage/difference?
Original (explicit?) formulation, rate form, and incremental form?

nagelt:
The rate and incremental forms make it easier to work with non-zero initial states and some coupling terms than with the absolute form. The incremental form avoids dividing by dt in the Jacobian which could be beneficial for some ratios between loading and relaxation rates.

PR #2605, opened 2019-08-14
Split pushBackState between init and post- timestep

endJunction:
In case when a time step is repeated `preTimestep` is called again. The `pushBackState` call must be executed only after successful time step, *i.e.* in `postTimestep`. Then an additional call in the initialization is needed; therefore a new interface function in local assemblers: `initialize`.

The first `pushBackState` call must be executed after the secondary variables are set, therefore a new function is needed. If you have better ideas, w/o adding a new interface function, I'd be happy to implement those.

Didn't apply the changes to LIE.

2. [x] Tests covering your feature were added? (Covered by current ctests)
3. [ ] Any new feature or behavior change was documented? Will not do, unless explicitly asked.

### DIFF
--- a/ProcessLib/HydroMechanics/HydroMechanicsProcess.cpp
+++ b/ProcessLib/HydroMechanics/HydroMechanicsProcess.cpp
@@ -245,6 +245,14 @@ void HydroMechanicsProcess<DisplacementDim>::initializeConcreteProcess(
         MeshLib::getOrCreateMeshProperty<double>(
             const_cast<MeshLib::Mesh&>(mesh), "pressure_interpolated",
             MeshLib::MeshItemType::Node, 1);
+
+    // Initialize local assemblers after all variables have been set.
+    const int process_id = 0;
+    ProcessLib::ProcessVariable const& pv = getProcessVariables(process_id)[0];
+
+    GlobalExecutor::executeSelectedMemberOnDereferenced(
+        &LocalAssemblerInterface::initialize, _local_assemblers,
+        pv.getActiveElementIDs(), *_local_to_global_index_map);
 }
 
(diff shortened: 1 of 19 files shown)


ProcessLib/HydroMechanics/HydroMechanicsProcess.cpp
```diff
@@ -245,6 +245,14 @@ void HydroMechanicsProcess<DisplacementDim>::initializeConcreteProcess(
         MeshLib::getOrCreateMeshProperty<double>(
             const_cast<MeshLib::Mesh&>(mesh), "pressure_interpolated",
             MeshLib::MeshItemType::Node, 1);
+
+    // Initialize local assemblers after all variables have been set.
+    const int process_id = 0;
```
wenqing: Suggestion:
- Remove lines 251-251.
- Change `GlobalExecutor::executeSelectedMemberOnDereferenced` to `GlobalExecutor::executeMemberOnDereferenced`
as 
```
  GlobalExecutor::executeSelectedMemberOnDereferenced(
        &LocalAssemblerInterface::initialize, _local_assemblers,
         *_local_to_global_index_map);
```

The same for other processes.

wenqing: using `GlobalExecutor::executeMemberOnDereferenced`

wenqing: Sorry: Remove lines 250-251.

endJunction: Thanks! This is better. Rebased. Changes in last commit.

wenqing:
:fast_forward:

PR #2610, opened 2019-08-15
Improve CMake find_package for Boost 

bilke:


bilke:
eve parallel job was green on the (identical) previous run

BaseLib/CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ include(GenerateExportHeader)
 generate_export_header(BaseLib)
 target_include_directories(BaseLib PUBLIC ${CMAKE_CURRENT_BINARY_DIR})
 
-target_link_libraries(BaseLib PUBLIC logog)
+target_link_libraries(BaseLib PUBLIC logog Boost::boost)
```
endJunction: Do you remember which part of boost is needed? I thought that all boost packages (used by OGS) were header-only...

bilke: We just use header-only libs. `Boost::boost` in newer CMake versions than 3.12 is an alias target to `Boost::headers`. See https://cmake.org/cmake/help/v3.15/module/FindBoost.html

To be 100 % correct we should link to this target from other OGS libs as well, but as `BaseLib` is linked from everywhere the dependency is implicitly also there...

PR #2616, opened 2019-08-19
[MPL] Add Property based on Parameter.

TomFischer:
PR implements a new Property based on Parameter. With this implementation spatial heterogeneous properties are available now.

  The disabled HT decovalex test was re-enable.

### DIFF
--- a/MaterialLib/MPL/Components/GetThermalExpansivity.cpp
+++ b/MaterialLib/MPL/Components/GetThermalExpansivity.cpp
@@ -14,14 +14,16 @@
 #include "GetThermalExpansivity.h"
 
 #include "MaterialLib/MPL/Phase.h"
+#include "MaterialLib/MPL/Phase.h"
 
 namespace MaterialPropertyLib
 {
 class Phase;
 
-double getThermalExpansivity(Phase const& phase,
-                             VariableArray const& vars,
-                             const double density)
+double getThermalExpansivity(Phase const& phase, VariableArray const& vars,
+                             const double density,
+                             ParameterLib::SpatialPosition const& pos,
+                             double const t)
 {
     auto const thermal_expansivity_ptr =
         &phase.property(MaterialPropertyLib::PropertyType::thermal_expansivity);
@@ -29,7 +31,7 @@ double getThermalExpansivity(Phase const& phase,
     // The thermal expansivity is explicitly given in the project file.
     if (thermal_expansivity_ptr)
     {
-        return (*thermal_expansivity_ptr).template value<double>(vars);
+        return (*thermal_expansivity_ptr).template value<double>(vars, pos, t);
     }
 
     // The thermal expansivity calculated by the density model directly.
--- a/MaterialLib/MPL/CreateProperty.cpp
+++ b/MaterialLib/MPL/CreateProperty.cpp
@@ -16,6 +16,8 @@
 #include <string>
 #include <vector>
 #include "BaseLib/ConfigTree.h"
+#include "ParameterLib/Parameter.h"
+#include "ParameterLib/Utils.h"
 
 #include "Properties/Properties.h"
 
@@ -26,7 +28,8 @@
 namespace
 {
 std::unique_ptr<MaterialPropertyLib::Property> createProperty(
-    BaseLib::ConfigTree const& config)
+    BaseLib::ConfigTree const& config,
+    std::vector<std::unique_ptr<ParameterLib::ParameterBase>> const& parameters)
 {
     using namespace MaterialPropertyLib;
     // Parsing the property type:
@@ -170,6 +173,17 @@ std::unique_ptr<MaterialPropertyLib::Property> createProperty(
             reference_value, exp_data);
     }
 
+    if (property_type == "Parameter")
+    {
+        std::string const& parameter_name =
+            //! \ogs_file_param{properties__property__Parameter__parameter_name}
+            config.getConfigParameter<std::string>("parameter_name");
+        auto const& parameter = ParameterLib::findParameter<double>(
+            parameter_name, parameters, 1, nullptr);
+        return std::make_unique<MaterialPropertyLib::ParameterProperty>(
+            parameter);
+    }
+
     /* TODO Additional properties go here, for example:
     if (boost::iequals(property_type, "BilinearTemperaturePressure"))
     {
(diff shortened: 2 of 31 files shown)


renchao-lu:
So great！

MaterialLib/MPL/Components/GetThermalExpansivity.cpp
```diff
@@ -13,23 +13,24 @@
...
 namespace MaterialPropertyLib
 {
 class Phase;
 
-double getThermalExpansivity(Phase const& phase,
-                             VariableArray const& vars,
-                             const double density)
+double getThermalExpansivity(Phase const& phase, VariableArray const& vars,
+                             const double density,
+                             ParameterLib::SpatialPosition const& pos, double t)
```
endJunction:
```suggestion
                             ParameterLib::SpatialPosition const& pos, double const t)
```

MaterialLib/MPL/CreateProperty.cpp
```diff
@@ -170,6 +173,17 @@ std::unique_ptr<MaterialPropertyLib::Property> createProperty(
             reference_value, exp_data);
     }
 
+    if (property_type == "parameter")
```
endJunction: parameter or Parameter? If the former, then in line 179 it should be
```
//! \ogs_file_param{properties__property__parameter__parameter_name}
```

MaterialLib/MPL/Properties/LinearProperty.h
```diff
@@ -34,9 +34,13 @@ class LinearProperty final : public Property
     /// of the base class Property to that value.
     LinearProperty(PropertyDataType const& property_reference_value,
                    IndependentVariable const& v);
+
     /// This method computes the value of a property depending linearly on
     /// the value of the given primary variable.
-    PropertyDataType value(VariableArray const& variable_array) const override;
+    PropertyDataType value(VariableArray const& variable_array,
+                           ParameterLib::SpatialPosition const& pos,
+                           double t) const override;
```
endJunction:
```suggestion
                           double const t) const override;
```

MaterialLib/MPL/Properties/ParameterProperty.cpp
```diff
@@ -0,0 +1,45 @@
...
+ParameterProperty::ParameterProperty(
+    ParameterLib::Parameter<double> const& parameter)
+    : _parameter(parameter)
+{
+}
+
+PropertyDataType ParameterProperty::value(
+    VariableArray const& /*variable_array*/,
+    ParameterLib::SpatialPosition const& pos,
+    double t) const
```
endJunction:
```suggestion
    double const t) const
```

MaterialLib/MPL/Property.h
```diff
@@ -69,11 +73,15 @@ class Property
     {
         return std::get<T>(value());
     }
+
     template <typename T>
-    T value(VariableArray const& variable_array) const
+    T value(VariableArray const& variable_array,
+            ParameterLib::SpatialPosition const& pos,
+            double t) const
```
endJunction:
```suggestion
            double const t) const
```

ProcessLib/HT/HTFEM.h
```diff
@@ -173,10 +173,9 @@ class HTFEM : public HTLocalAssemblerInterface
         _ip_data;
 
     double getHeatEnergyCoefficient(
-        MaterialPropertyLib::VariableArray const& vars,
-        const double porosity,
-        const double fluid_density,
-        const double specific_heat_capacity_fluid)
+        MaterialPropertyLib::VariableArray const& vars, const double porosity,
+        const double fluid_density, const double specific_heat_capacity_fluid,
+        ParameterLib::SpatialPosition const& pos, double t)
```
endJunction:
```suggestion
        ParameterLib::SpatialPosition const& pos, double const t)
```

Tests/Data/Parabolic/HT/StaggeredCoupling/ADecovalexTHMCBasedHTExample/th_decovalex.prj
```diff
@@ -73,6 +70,21 @@
                             <type>Constant</type>
                             <value>0.0</value>
                         </property>
+                        <property>
+                            <name>density</name>
+                            <type>parameter</type>
+                            <parameter_name>rho_solid</parameter_name>
```
endJunction: For discussion:
Could be also
```xml
<property>
    <name>density</name>
    <parameter>rho_solid</parameter>
</property>
```

What do you think?
@TomFischer I know, the parsing is slightly more difficult.

TomFischer: The advantage of your variant is a non-redundant, short description of a `ParameterProperty` object. The advantage of the implementation of the PR is that it is consistent with the specification of other parameter types.

@ @wenqing, @renchao-lu, @ErikNixdorf , @waltherm Opinions?

wenqing: I think that it is better to have the consistency with other types of input. Therefore I prefer to take the present one in this PR.

renchao-lu: In my mind, there are three types of property offered by far as follows: constant, linear dependency, and exponential dependency. I guess the initial design of the tag `<type>` intends to establish property dependency linking to primary process variables which will change over time. If a material property is not time-dependent, we assign the tag to constant. Otherwise, we assign to linear/exponential dependency. As to ParameterLib, I think the major business of ParameterLib is to tackle spatial heterogeneity, although it offers some time-associated functionalities. For this main reason, ParameterLib has its own tag for property type: function, group, meshnode, and meshelement (for spatial heterogeneity), curvescaled, and timedependentheterogeneousparameter (for time-dependency).

waltherm: I would vote for @TomFischer implementation, which seems more consistent with the existing, alternative options.

endJunction: Thanks for discussion.
:fast_forward:

ProcessLib/HT/HTFEM.h
```diff
@@ -199,7 +198,8 @@ class HTFEM : public HTLocalAssemblerInterface
     GlobalDimMatrixType getThermalConductivityDispersivity(
         MaterialPropertyLib::VariableArray const& vars, const double porosity,
         const double fluid_density, const double specific_heat_capacity_fluid,
-        const GlobalDimVectorType& velocity, const GlobalDimMatrixType& I)
+        const GlobalDimVectorType& velocity, const GlobalDimMatrixType& I,
+        ParameterLib::SpatialPosition const& pos, double t)
```
endJunction:
```suggestion
        ParameterLib::SpatialPosition const& pos, double const t)
```

PR #2625, opened 2019-08-21
Split BuildInfo from BaseLib into separate libs

bilke:
The info from `BaseLib/BuildInfo.h` is now split into:

- `InfoLib/CMakeInfo.h`
- `InfoLib/CompilerInfo.h` (info not used)
- `InfoLib/GitInfo.h`
- `InfoLib/TestInfo.h`

This should help in reducing relinking when e.g. a commit was done as before this change this modified the `BaseLib` and everything else depends on `BaseLib`. Now only stuff which really uses the Git information relinks.


bilke:
@endJunction Shall I remove unused variables? See `TestInfo.h` and all variables in `CompilerInfo.h`.

endJunction:
Weren't the compiler info strings included in the `--version` output? Maybe I mix up with cmake info.
Drop all unused stuff, otherwise.

PR #2628, opened 2019-08-23
Container generation

bilke:
~~Adds targets for container generation:~~

- ~~`docker`: Builds a Docker image from the local source code. **Important**: Uses the same CMake config as specified on the first CMake run. Once a `CMakeCache.txt` is in place changed CMake options are not considered. See the [docs](https://www.opengeosys.org/docs/devguide/advanced/docker/) for usage. Requires [Docker](https://www.docker.com)~~
- ~~`sif`: Converts the previously created Docker container into a **S**ingularity **I**mage **F**ile. See the [docs](https://www.opengeosys.org/docs/userguide/basics/container/#run-ogs-inside-a-container-called-from-outside) for usage. Requires [Singularity](https://sylabs.io).~~

- Adds CI builds of OGS in different configurations (different MPI versions) on [GitLab CI](https://gitlab.opengeosys.org/bilke/ogs/pipelines) (just for the fun of it)
- Adds CI builds of OGS in different configurations (different MPI versions) on Jenkins (master-only, Singularity image files are provided as build artifacts)

Other changes:

- Add https://github.com/ufz/ogs-container-maker as a submodule
- Made existence of local git repo optional for building OGS. In this PR this is used to optimize container build times (as the full source directory has to be transferred to Docker context) but should also allow building OGS from GitHub archive file (tar- or zip-file). In the later case you have to specify the CMake variable `OGS_VERSION`.


scripts/cmake/PreFind.cmake
```diff
@@ -9,15 +9,23 @@ find_program(GIT_EXECUTABLE
...
     RESULT_VARIABLE IS_GIT_REPO
     OUTPUT_QUIET)
 if(IS_GIT_REPO GREATER 0)
-    message(FATAL_ERROR "Source code at ${PROJECT_SOURCE_DIR} is not a git "
-        "repository! Please use git to obtain the source code! See "
-        "https://www.opengeosys.org/docs/devguide/getting-started/get-the-source-code/")
+    set(IS_GIT_REPO FALSE CACHE INTERNAL "")
+    if(DEFINED OGS_VERSION)
+        message(WARNING "Using user-provided OGS_VERSION: ${OGS_VERSION}!")
+        message(WARNING "Submodule setup is skipped!")
```
endJunction: better to move the warning into the if in SubmoduleSetup.cmake.

scripts/cmake/ProjectSetup.cmake
```diff
@@ -28,6 +28,10 @@ include_directories(${CMAKE_CURRENT_SOURCE_DIR})
 
 set(CMAKE_MACOSX_RPATH 1)
 
+if(NOT IS_GIT_REPO)
+    return()
```
endJunction: Could also issue a warning...

scripts/cmake/SubmoduleSetup.cmake
```diff
@@ -1,3 +1,7 @@
+if(NOT IS_GIT_REPO)
+    return()
```
endJunction:
```suggestion
    message(WARNING "Submodule setup is skipped!")
```

scripts/cmake/PreFind.cmake
bilke: This handles all possibilities regarding warning output. So it would leave this here and there is no additional output necessary in subsequent CMake files.

PR #2633, opened 2019-08-27
Global treatment of out of balance forces.

endJunction:
Follows #2635, where the `<initial_stress>` tag is added. Reenables the tests commented in #2635.


NumLib/ODESolver/NonlinearSolver.cpp
```diff
@@ -184,6 +184,17 @@ void NonlinearSolver<NonlinearSolverTag::Newton>::assemble(
     //      equation every time and could not forget it.
 }
 
+void NonlinearSolver<NonlinearSolverTag::Newton>::calculateOutOfBalanceForces(
```
norihiro-w: IMO, NumLib should be independent of physics.

norihiro-w:
what is "the global Foob's procedure"? If there is any paper about it, would you please give me citation?

NumLib/ODESolver/NonlinearSolver.cpp
endJunction: I just updated the documentation, there are some more details.
In the document we named the initial residuum F_oob -- out-of-balance forces. These are not the "physical" forces, but general external loads. For mechanics, these are indeed forces as
```tex
$F^oob = r_0 = \int\Omega B^T \sigma_0\; d\Omega$
```

web/content/docs/benchmarks/small-deformations/mechanics-linear-nonequilibrium-states.pandoc
```diff
@@ -0,0 +1,27 @@
...
+  [menu.benchmarks]
+    parent = "small-deformations"
+
++++
+
+{{< data-link >}}
+
+## Overview
+
+For detailed description see [thechnical note
```
TomFischer:
```suggestion
For detailed description see [technical note
```

endJunction:
> what is "the global Foob's procedure"?

"global" is just in the sense that the implementation is valid for all nonlinear problems, where the residuum needs to be adjusted, whereas (there is an open PR #2561 being superseded by this) non-equilibrium initial state treatment happened on the ("local") process scale.

NumLib/ODESolver/NonlinearSolver.cpp
norihiro-w: Thanks for the doc. Is there any way to see the documentation without doing git lfs? 

Do we need this calculation besides mechanics?  To me, "load", "force" sounds specific to mechanics.

endJunction: > the doc.

https://deploy-preview-2633--ogs.netlify.com/docs/benchmarks/small-deformations/mechanics-linear-nonequilibrium-states/

The preview to the benchmark sites is under the "deploy/netlify -- Details" in the checks of the PR. There in "docs", "Selected Benchmarks", you will find the above site.

> "load", "force" ...

Same arguments are valid for pressure or saturation. One could rename Foob to `initial_residuum`, `r_0`, or something similar to avoid mechanics relation...

nagelt: @endJunction The F^oob equation above also contains body forces and Neumann bc's right? I think you were just brief but want to make sure...

@norihiro-w I sent you the equations a while ago to replace the non-equ. stresses with a better more general solution. Indeed "forces" points to mechanics, like much of the FE nomenclature (e.g. stiffness matrix). So I wouldn't have a problem with that name. The most general alternative could be "out of balance fluxes". That applies to all processes. In case of mechanics, forces are linear momentum fluxes.... then we could even keep the F ;-)

nagelt: and yes, it is intended that this applies to all processes where needed ... we can write this for a general balance equation and Foob would always be the initial out of balance flux that a user may want to drive an initial process or not (and hence would have to compensate) ... 
time for dinner now in China ...

norihiro-w: @endJunction thank you for the tip. I could read it. 

@endJunction @nagelt Indeed I'm not suggesting renaming the function. I'm suggesting moving this functionality to somewhere in ProcessLib. NumLib should be a general pure numerical library and we should avoid contaminate it with physics except a concept of space and time. I guess you can use a post-assembly function caller in the nonlinear solver class without changing the general interface. If you still want to change NumLib, it would be better if you use a concept in math/numeric, such as residual shift/offset (i'm not sure this example is appropriate). 

@nagelt Is "Foob" a word you defined? I was asking Dima about the name because I'm surprised that this trivial algorithm has a name.

norihiro-w:
@nagelt why do you have to remove "r0" also from the mass balance equation in HM/RM? There should be no problem to calculate hydrostatic pressure. Maybe you have some applications demanding this, but I don't think it is usual case. It would be nicer if you introduce an option to activate or deactivate calculating r0 for non-mechanic process.

nagelt:
@norihiro-w It was a feature request to have this also for the mass balance. You're right, we want separate control over the individual process residuals, so that you can decide to only compensate for individual processes, e.g. mechanics. @endJunction will do this in a future PR.

NumLib/ODESolver/NonlinearSolver.cpp
nagelt: @norihiro-w maybe I misunderstand you and say something totally obvious to you now, but we just use "Foob" as a short-hand for $F_oob$ = out-of-balance-forces/fluxes. We could also call this initial residual if we leave physics out of the name ... but for most engineers the physical name implies the concept better than the purely mathematical one..

regarding ProcessLib: if we can move it there without much extra code into a general place, I agree. I'm not fully aware of the structure there and the effort involved, so I'll leave this call to @endJunction

NumLib/ODESolver/NonlinearSolver.h
```diff
@@ -108,11 +110,18 @@ class NonlinearSolver<NonlinearSolverTag::Newton> final
 
     void assemble(GlobalVector const& x) const override;
 
+    void calculateOutOfBalanceForces(GlobalVector const& x) override;
```
wenqing: It is better to move it to Process. In future, we can use CRTP idiom to encapsulate all mechanical process related stuff.

NumLib/ODESolver/NonlinearSolver.h
```diff
@@ -127,11 +136,18 @@ class NonlinearSolver<NonlinearSolverTag::Newton> final
     //! conservative approach.
     double const _damping;
 
+    GlobalVector* _f_oob = nullptr;      //!< Out-of-balance forces vector.
```
wenqing: Better to move it to Process.

NumLib/ODESolver/NonlinearSolver.h
```diff
@@ -127,11 +136,18 @@ class NonlinearSolver<NonlinearSolverTag::Newton> final
...
     std::size_t _J_id = 0u;              //!< ID of the Jacobian matrix.
     std::size_t _minus_delta_x_id = 0u;  //!< ID of the \f$ -\Delta x\f$ vector.
     std::size_t _x_new_id =
         0u;  //!< ID of the vector storing \f$ x - (-\Delta x) \f$.
+
+    /// Enables computation of the out-of-balance forces \f$ F_{\rm oob} \f$
+    /// before the first time step. The forces are zero if the external forces
+    /// are in equilibrium with the initial state/initial conditions. During the
+    /// simulation the new residuum reads \f$ \tilde r = r - F_{\rm oob} \f$.
+    bool _compensate_initial_out_of_balance_forces = false;
```
wenqing: It it better to move it to Process.

NumLib/ODESolver/NonlinearSolver.h
```diff
@@ -108,11 +110,18 @@ class NonlinearSolver<NonlinearSolverTag::Newton> final
...
     void assemble(GlobalVector const& x) const override;
 
+    void calculateOutOfBalanceForces(GlobalVector const& x) override;
+
     NonlinearSolverStatus solve(
         GlobalVector& x,
         std::function<void(int, GlobalVector const&)> const&
             postIterationCallback) override;
 
+    void compensateInitialOutOfBalanceForces(bool const value)
```
wenqing: It it better to move it Process.

ProcessLib/TimeLoop.cpp
```diff
@@ -167,6 +167,34 @@ std::vector<GlobalVector*> setInitialConditions(
     return process_solutions;
 }
 
+void calculateOutOfBalanceForces(
```
wenqing: can be a member of Process

endJunction:
Just rebased. I'll try to move the code to ProcessLib. I'm not sure whether this is good idea, because as I understood it, the procedure is independent of any process implementations and is a nonlinear solver feature:
```
r_0 = A(u_0,v) + b(u_0, v)
solve A(u,v) + b(u,v) - r_0 = 0;
```

NumLib/ODESolver/NonlinearSolver.cpp
```diff
@@ -232,6 +243,10 @@ NonlinearSolverStatus NonlinearSolver<NonlinearSolverTag::Newton>::solve(
         sys.getJacobian(J);
         INFO("[time] Assembly took %g s.", time_assembly.elapsed());
 
+        // Apply out-of-balance forces if set.
```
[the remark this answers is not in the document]
endJunction: Probably I'm missing something here, but I think it is not possible to move this part out of the nonlinear solver. The reason is the `sys.getResidual(x, res)` few lines above. Computation of the residuum depends on the time discretization scheme. Passing the time discretization scheme to the processes is not a good option.
The residuum can not be computed in the process in general case.

@norihiro-w @wenqing I'd like to see some sketches/ideas how to move the computation into the processes.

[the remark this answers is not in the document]
norihiro-w: @endJunction At least, for mechanics, you don't need to care the time discretization, because it's solving static problems. For dynamic problems, I have no idea how you can make F^oob consistent with any time discretization scheme.

endJunction:
> From the point view of numerics, the nonlinear solver should not included this part of computation. 
With this I need some help to understand it; the nonlinear solver is, until now, solving an equation of this type:
```
F(u) = 0
```
for `u`, using either fixed point iterations or some Newton scheme. Here we modify the equation to be solved to
```
F(u) - F(u_0) = 0
```
Now, I'm not against it if this can be moved to ProcessLib but I admit not to have an idea how to.

ThieJan:
I do not have the complete overview, but if its neither process lib nor nonlinear solver, maybe we have to add something in between. E.g., wrap the class responsible for the evaluation of `F(u)` returning the 'shifted' value?

endJunction:
Rebased. Renamed the "out-of-balance forces" to "non-equilibrium initial residuum".

PR #2640, opened 2019-09-02
[CL] New Phreeqc Interface

renchao-lu:
Current chemical solver interface between OGS-6 and Phreeqc works in a conventional way (via file-based approach). That is, in each time step, Phreeqc input file is prepared for speciation calculation and phreeqc output file is then parsed for result updates. This approach is not computationally efficient.

New Phreeqc Interface by direct memory access

Thanks to Dima's instruction and Wenqing's help, we hereby provide an alternative interface by direct memory access. This approach allows us not to process Phreeqc input and output files any more. This feature greatly cuts down computation costs. In the small-scale kinetic reactant benchmark added in this pull request (containing 201 nodes), computation time by Phreeqc reduces from ~0.05 seconds to ~0.03 seconds (in average) in each time step. We gain **60% speed-up**. In another intermediate-scale test (not included in this pull request, download by [IntermediateScaleKineticReactantBenchmark.zip](https://github.com/ufz/ogs/files/3566125/IntermediateScaleKineticReactantBenchmark.zip) containing 24461 nodes, with the same chemical configuration as in the small-scale test), Phreeqc runtime decreases from ~40 seconds to ~2 seconds. We acquire **20x speed-up**. Speed-up effect becomes more and more significant with the increase of number of nodes (also referred as number of chemical system).

In addition, using this alternative interface prevents from precision lost in truncation. 

Limitation
By far this new interface only supports for kinetic reaction process (equilibrium reaction process is not open yet).


ChemistryLib/PhreeqcKernel.cpp
```diff
@@ -0,0 +1,220 @@
...
+        // solutions after chemical calculation by Phreeqc
+        for (auto const& process_id_to_master_pair : _process_id_to_master_map)
+        {
+            auto& transport_process_id = process_id_to_master_pair.first;
+            auto& transport_process_solution =
+                process_solutions[transport_process_id];
+
+            auto& master_species = process_id_to_master_pair.second;
+            auto& element_name = master_species->elt->name;
+            if (strcmp(element_name, "H") == 0)
```
endJunction:
```suggestion
            if (element_name == "H")
```

ChemistryLib/PhreeqcKernel.cpp
```diff
@@ -0,0 +1,220 @@
...
+        new_solution = 1;
+        use.Set_n_solution_user(chemical_system_id);
+
+        if (!Rxn_kinetics_map.empty())
+        {
+            use.Set_kinetics_in(true);
+            use.Set_n_kinetics_user(chemical_system_id);
+        }
+
+        pr.all = false;
```
endJunction: What is `pr`?

ChemistryLib/PhreeqcKernel.h
```diff
@@ -0,0 +1,60 @@
...
+    void setTimeStep(double const dt) override;
+
+    void execute(std::vector<GlobalVector*>& process_solutions);
+
+    void updateNodalProcessSolutions(
+        std::vector<GlobalVector*> const& process_solutions,
+        std::size_t const node_id);
+
+private:
+    std::map<int, struct master*> _process_id_to_master_map;
```
endJunction: If keys are consecutive a vector would be better.

ChemistryLib/PhreeqcKernelData/CreateReactionRate.cpp
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    int count_rates = 0;
+    struct rate* rates = (struct rate*)malloc(sizeof(struct rate));
+    for (auto const& rate_config :
+         //! \ogs_file_param{prj__chemical_system__rates__rate}
+         config->getConfigSubtreeList("rate"))
+    {
+        if (count_rates > 0)
+            rates = (struct rate*)realloc(
```
endJunction: direct memory allocation is highly discouraged.
`unique_ptr`, `vector`.

ChemistryLib/PhreeqcKernelData/CreateReactionRate.cpp
```diff
@@ -0,0 +1,72 @@
...
+                rates, (std::size_t)(count_rates + 1) * sizeof(struct rate));
+
+        std::string* kinetic_reactant = new std::string(
```
endJunction: `new` is not allowed.

ChemistryLib/PhreeqcKernelData/InitialAqueousSolution.h
```diff
@@ -0,0 +1,30 @@
...
+#include "ThirdParty/iphreeqc/src/src/phreeqcpp/ISolution.h"
+
+namespace ChemistryLib
+{
+namespace PhreeqcKernelData
+{
+class Component final : public cxxISolutionComp
+{
+public:
+    Component(std::string const& name) { description = name; }
```
endJunction:
```suggestion
    explicit Component(std::string const& name) { description = name; }
```

ChemistryLib/PhreeqcKernelData/KineticReactant.cpp
```diff
@@ -0,0 +1,33 @@
...
+{
+    rate_name = name.c_str();
+    namecoef.add(name.c_str(), 1.0);
+    m = initial_amount;
+    m0 = initial_amount;
+}
+
+Kinetics::Kinetics(std::vector<KineticReactant> const& kinetic_reactants)
+{
+    for (auto const& kinetic_reactant : kinetic_reactants)
```
endJunction: std::copy or std::transform

joboog:
Is this replacing the existing OGS-6--PhreeqcInterface, or will there be an option to choose the interface type?

renchao-lu:
The latter. Switching the tag `<chemical_solver>` from `Phreeqc` to `PhreeqcKernel` is meant to call the new phreeqc interface.

ChemistryLib/PhreeqcKernelData/KineticReactant.cpp
renchao-lu: Fixed.

ChemistryLib/PhreeqcKernelData/InitialAqueousSolution.h
renchao-lu: Fixed herein as well as elsewhere.

montoyav:
Maybe interesting to know/mention how many cores were used for the mentioned simulations.

renchao-lu:
> Maybe interesting to know/mention how many cores were used for the mentioned simulations.

Parallelization not touched in this PR. Just single core. @montoyav

ChemistryLib/PhreeqcKernelData/CreateReactionRate.cpp
renchao-lu: Fixed.

ChemistryLib/PhreeqcKernelData/CreateReactionRate.cpp
renchao-lu: partially solved.

ChemistryLib/PhreeqcKernel.cpp
renchao-lu: Defined by third-party library Phreeqc. Moved into a private class method.

ChemistryLib/PhreeqcKernel.cpp
renchao-lu: Moved into a private class method.

ChemistryLib/PhreeqcKernel.h
renchao-lu: It depends upon how one sets up OGS project file. If the sequence of mobile components specified between `<process_variables>/<concentration>` and `<solution>/<components>` is consistent, the key is consecutive as you said. Otherwise the key is not consecutive. But it can be improved later on, I think.

ChemistryLib/PhreeqcKernelData/CreateReactionRate.cpp
renchao-lu: Memory reallocation appears once in the small private class method `void PhreeqcKernel::reinitializeRates()`.

ChemistryLib/PhreeqcKernel.h
```diff
@@ -0,0 +1,74 @@
...
+
+#pragma once
+
+#include <map>
+#include <vector>
+
+#include "ChemicalSolverInterface.h"
+#include "PhreeqcKernelData/AqueousSolution.h"
+
+#include "ThirdParty/iphreeqc/src/src/phreeqcpp/Phreeqc.h"
```
endJunction: If you move the implementations into cpp, this header file could be moved to cpp/replaced with fwd-decls.

ChemistryLib/PhreeqcIOData/ReactionRate.h
```diff
@@ -15,13 +15,15 @@
...
+namespace PhreeqcIOData
+{
 struct ReactionRate
 {
     ReactionRate(std::string kinetic_reactant_,
                  std::vector<std::string>
                      expression_statements_)
         : kinetic_reactant(std::move(kinetic_reactant_)),
-          expression_statements(std::move(expression_statements_))
+          expression_statements(expression_statements_)
```
endJunction: why no r-value anymore?

ChemistryLib/PhreeqcKernel.h
renchao-lu: May be not, even moving the implementation file. This header file, I think, is required for the declaration of the base class `Phreeqc`...

ChemistryLib/PhreeqcIOData/ReactionRate.h
renchao-lu: As the creation function `createReactionRates` is shared for creating instances of `PhreeqcIOData::ReactionRate` and `PhreeqcKernelData::ReactionRate`.

renchao-lu: Fixed.

ChemistryLib/PhreeqcIO.h
```diff
@@ -41,13 +44,14 @@ class PhreeqcIO
...
     void doWaterChemistryCalculation(
-        std::vector<GlobalVector*>& process_solutions, double const dt);
+        std::vector<GlobalVector*>& process_solutions,
+        double const dt) override;
 
     void setAqueousSolutionsOrUpdateProcessSolutions(
         std::vector<GlobalVector*> const& process_solutions,
         Status const status);
 
-    void setTimeStep(double const dt) { _dt = dt; }
```
endJunction: Not this PR; The `doWaterChemistryCalculation` has `dt` as an argument, and this function also handles `dt`. What's the difference?

ChemistryLib/PhreeqcIOData/CreateAqueousSolution.cpp
```diff
@@ -8,52 +8,35 @@
...
     if (means_of_adjusting_charge_in_str)
     {
         if (*means_of_adjusting_charge_in_str == "pH")
         {
-            means_of_adjusting_charge = MeansOfAdjustingCharge::pH;
+            return MeansOfAdjustingCharge::pH;
         }
         else if (*means_of_adjusting_charge_in_str == "pe")
         {
-            means_of_adjusting_charge = MeansOfAdjustingCharge::pe;
```
endJunction: drop `else` after return...

ChemistryLib/PhreeqcIOData/CreateAqueousSolution.cpp
```diff
@@ -65,10 +48,32 @@ AqueousSolution createAqueousSolution(BaseLib::ConfigTree const& config)
     }
     else
     {
-        means_of_adjusting_charge = MeansOfAdjustingCharge::Unspecified;
+        return MeansOfAdjustingCharge::Unspecified;
```
endJunction: Just a question; Is this valid and defined state or is it an error in general and a default will be chosen?

ChemistryLib/PhreeqcKernel.cpp
```diff
@@ -0,0 +1,257 @@
...
+    std::string const& database,
+    PhreeqcKernelData::AqueousSolution& aqueous_solution,
+    std::unique_ptr<Kinetics>
+        kinetic_reactants,
+    std::vector<ReactionRate>&& reaction_rates)
+    : Phreeqc(),
+      _initial_aqueous_solution(aqueous_solution.getInitialAqueousSolution()),
+      _reaction_rates(std::move(reaction_rates))
+{
+    do_initialize();
```
endJunction: drop the `do_` part, it's a verb already. Would be better, instead, to add something to the name like in `reinitializeRates`, which is a good name.

ChemistryLib/PhreeqcKernel.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+    execute(process_solutions);
+}
+
+void PhreeqcKernel::setAqueousSolutions(
+    std::vector<GlobalVector*> const& process_solutions)
+{
+    // Loop over chemical systems
+    auto const num_chemical_systems = process_solutions[0]->size();
+    for (auto chemical_system_id = 0; chemical_system_id < num_chemical_systems;
```
endJunction: This will complain about mismatched integer sizes: `num_chemical_systems` is most likely a `std::size_t`, and `chemical_system_id` is an `int` because the `0` is.

ChemistryLib/PhreeqcKernel.cpp
```diff
@@ -0,0 +1,257 @@
...
+            {
+                // Set pH value by hydrogen concentration.
+                double const pH = -std::log10(
+                    transport_process_solution->get(chemical_system_id));
+                aqueous_solution.Set_ph(pH);
+            }
+            else
+            {
+                // Set component concentrations.
+                auto const& concentration =
```
endJunction: if this is a `double`, drop the reference. Store simple and short types by value.

renchao-lu: Fixed.

ChemistryLib/PhreeqcKernel.cpp
renchao-lu: Fixed here and elsewhere.

ChemistryLib/PhreeqcKernel.cpp
renchao-lu: Fixed. Encapsulated into a function.

ChemistryLib/PhreeqcIOData/CreateAqueousSolution.cpp
renchao-lu: This return is a common case.

ChemistryLib/PhreeqcIOData/CreateAqueousSolution.cpp
renchao-lu: Improved.

ChemistryLib/PhreeqcIO.h
renchao-lu: Dropped `setTimeStep` function.

renchao-lu:
> There is some doxygen error resulting in 19 warnings -- should be fixed.
> As in #2660, please add `\file` in all headers -- there is a script in the PR's commits' comments.
> 
> There are some known deficiencies because of the way how Phreeqc is written, too long to describe here. I like that most of the Phreeqc calls are implemented in private functions and aren't mixed with the OGS code. Let's continue the journey to fast Phreeqc interface.

Thanks for the command.  Added `\file`.

ChemistryLib/CreateChemicalSolverInterface.cpp
```diff
@@ -0,0 +1,137 @@
...
+    // output
+    auto const& components = aqueous_solution.components;
+    auto const project_file_name = BaseLib::joinPaths(
+        output_directory,
+        BaseLib::extractBaseNameWithoutExtension(config.getProjectFileName()));
+    auto output = PhreeqcIOData::createOutput(
+        components, equilibrium_phases, kinetic_reactants, project_file_name);
+
+    return std::make_unique<PhreeqcIOData::PhreeqcIO>(
+        project_file_name, std::move(path_to_database),
```
TomFischer: I wonder why the `project_file_name` isn't moved, too?

ChemistryLib/CreateChemicalSolverInterface.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+    // rates
+    auto reaction_rates = createReactionRates<PhreeqcKernelData::ReactionRate>(
+        //! \ogs_file_param{prj__chemical_system__rates}
+        config.getConfigSubtreeOptional("rates"));
+
+    return std::make_unique<PhreeqcKernelData::PhreeqcKernel>(
+        mesh.getNumberOfBaseNodes(), process_id_to_component_name_map,
+        path_to_database, aqueous_solution, std::move(kinetic_reactants),
+        std::move(reaction_rates));
```
TomFischer: Is it worth to `std::move` the `path_to_database` and `aqueous_solution`? Or is there any reason not to do so?

ChemistryLib/PhreeqcKernel.h
```diff
@@ -0,0 +1,82 @@
...
+    void setConvergenceTolerance() { convergence_tolerance = 1e-12; }
+
+    void configureOutputSettings() { pr.all = false; }
+
+    cxxISolution* getOrCreateInitialAqueousSolution(
+        cxxSolution& aqueous_solution);
+
+    bool isHydrogen(char const* element) const
+    {
+        return strcmp(element, "H") == 0 ? true : false;
```
wenqing: directly `return strcmp(element, "H") == 0`

renchao-lu: Fixed.

ChemistryLib/CreateChemicalSolverInterface.cpp
renchao-lu: Enclosed in `std::move`.

ChemistryLib/CreateChemicalSolverInterface.cpp
renchao-lu: Fixed.

PR #2641, opened 2019-09-02
[Fix] Data Explorer netCDF-Interface using netcdf-cxx4

rinkk:
Adjusting CH's code to work the updated library.

### DIFF
--- a/Applications/DataExplorer/NetCdfDialog/NetCdfConfigureDialog.cpp
+++ b/Applications/DataExplorer/NetCdfDialog/NetCdfConfigureDialog.cpp
@@ -124,6 +124,7 @@ void NetCdfConfigureDialog::on_comboBoxDim3_currentIndexChanged(int id)
 //set up additional dimension
 void NetCdfConfigureDialog::on_comboBoxDim4_currentIndexChanged(int id)
 {
+    Q_UNUSED(id);
     if (_currentVar.getDimCount() > 3)
     {
         double firstValue=0, lastValue=0;
@@ -228,7 +229,7 @@ int NetCdfConfigureDialog::getDim4() const
 {
     NcVar const& dim3Var =
         _currentFile.getVar(comboBoxDim4->currentText().toStdString());
-    std::vector<std::size_t> start{{static_cast<std::size_t>(spinBoxDim4->value())}};
+    std::vector<std::size_t> start{static_cast<std::size_t>(spinBoxDim4->value())};
     int value(0);
     dim3Var.getVar(start, {1}, &value);
     if (value < 0)
@@ -344,7 +345,7 @@ QString NetCdfConfigureDialog::setName()
 std::string NetCdfConfigureDialog::getName()
 {
     std::string name = (lineEditName->text()).toStdString();
-    QString const date = dateTimeEditDim3->value();
+    QString const date = QString::number(dateTimeEditDim3->value());
     name.append(" - ").append(date.toStdString());
     return name;
 }
--- a/Applications/DataExplorer/NetCdfDialog/NetCdfConfigureDialog.cpp
+++ b/Applications/DataExplorer/NetCdfDialog/NetCdfConfigureDialog.cpp
@@ -139,12 +139,12 @@ int NetCdfConfigureDialog::setVariableSelect()
     int max_dim = 0;
     int max_dim_idx = 0;
     auto const& names =_currentFile.getVars();
-    for (auto it = names.cbegin(); it != names.cend(); ++it)
+    for (auto [name, var] : names)
     {
-        int const var_dim_count = it->second.getDimCount();
+        int const var_dim_count = var.getDimCount();
         if (var_dim_count > 1)
         {
-            comboBoxVariable->addItem(QString::fromStdString(it->first));
+            comboBoxVariable->addItem(QString::fromStdString(name));
             if (var_dim_count > max_dim)
             {
                 max_dim = var_dim_count;
@@ -223,8 +223,6 @@ void NetCdfConfigureDialog::getDimEdges(std::string const& name, unsigned& size,
 
 int NetCdfConfigureDialog::getTimeStep() const
 {
-    NcVar const& time_var =
-        _currentFile.getVar(comboBoxDim3->currentText().toStdString());
     return dateTimeEditDim3->value();
 }
 
--- a/Applications/DataExplorer/NetCdfDialog/NetCdfConfigureDialog.cpp
+++ b/Applications/DataExplorer/NetCdfDialog/NetCdfConfigureDialog.cpp
@@ -25,8 +25,11 @@ using namespace netCDF;
 
 // Constructor
 NetCdfConfigureDialog::NetCdfConfigureDialog(std::string const& fileName, QDialog* parent)
-    : QDialog(parent), _currentFile(fileName.c_str(), NcFile::read),
-  _currentInitialDateTime(QDateTime()), _currentMesh(nullptr), _currentRaster(nullptr), _currentPath(fileName)
+    : QDialog(parent),
+      _currentFile(fileName.c_str(), NcFile::read),
+      _currentMesh(nullptr),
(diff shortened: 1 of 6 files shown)


Applications/DataExplorer/NetCdfDialog/NetCdfConfigureDialog.cpp
```diff
@@ -158,17 +138,16 @@ int NetCdfConfigureDialog::setVariableSelect()
 {
     int max_dim = 0;
     int max_dim_idx = 0;
-    auto const n_vars = _currentFile->num_vars();
-    for (int i = 0; i < n_vars; i++)
+    auto const& names =_currentFile.getVars();
+    for (auto it = names.cbegin(); it != names.cend(); ++it)
```
endJunction: Could be
```suggestion
    for (auto [xxx, yyy] : names)
```
and later just use `xxx` and `yyy` instead of `it->first/second`. Choose proper names, of course.

Applications/DataExplorer/NetCdfDialog/NetCdfConfigureDialog.cpp
```diff
@@ -216,81 +200,55 @@ void NetCdfConfigureDialog::setDimensionSelect()
         spinBoxDim4->setValue(0);
 }
 
-void NetCdfConfigureDialog::getDimEdges(int dimId, unsigned &size, double &firstValue, double &lastValue)
+void NetCdfConfigureDialog::getDimEdges(std::string const& name, unsigned& size,
+                                        double& firstValue, double& lastValue)
```
endJunction: better to return a tuple.

Applications/DataExplorer/NetCdfDialog/NetCdfConfigureDialog.cpp
```diff
@@ -216,81 +200,55 @@ void NetCdfConfigureDialog::setDimensionSelect()
...
-{
-    NcVar* timeVar = _currentFile->get_var(comboBoxDim2->currentIndex());
-
-    double const datesToMinutes = convertDateToMinutes(_currentInitialDateTime,dateTimeEditDim3->date(),dateTimeEditDim3->time());
-
-    double timeArray[1] = {datesToMinutes};
-    double currentTime = timeVar->get_index(timeArray);
-    if (currentTime < 0) currentTime=0; //if the value isn't found in the array, set it to 0 as default...
-    return static_cast<int>(currentTime);
+    NcVar const& time_var =
```
endJunction: `time_var` is unused, is that right?

Applications/DataExplorer/NetCdfDialog/NetCdfConfigureDialog.cpp
rinkk: That syntax doesn't work for me for some reason.

FWIW, "names" is a multimap.

Applications/DataExplorer/NetCdfDialog/NetCdfConfigureDialog.cpp
rinkk: I'm returning three values here (size, first, last) ... putting these into one tuple seems strange. Also it'd require a ton of changes because the method is called quite often. Do you insist on this?

Applications/DataExplorer/NetCdfDialog/NetCdfConfigureDialog.cpp
rinkk: :heavy_check_mark:

Applications/DataExplorer/NetCdfDialog/NetCdfConfigureDialog.cpp
endJunction: no, it's a legacy code, so we keep it as is, until somebody needs to change it. Would be good opportunity to improve it now, though.

Applications/DataExplorer/NetCdfDialog/NetCdfConfigureDialog.cpp
```diff
@@ -21,16 +21,21 @@
 
 #include <vtkImageImport.h>
```
TomFischer: Maybe a stupid question, but is this include still necessary?

Applications/DataExplorer/NetCdfDialog/NetCdfConfigureDialog.cpp
```diff
@@ -56,21 +61,20 @@ void NetCdfConfigureDialog::accept()
...
 }
 
 // Instructions if the Cancel-Button has been pressed.
 void NetCdfConfigureDialog::reject()
 {
-    delete _currentFile;
     this->done(QDialog::Rejected);
 }
 
 void NetCdfConfigureDialog::on_comboBoxVariable_currentIndexChanged(int id)
```
TomFischer: The argument `id` isn't used in the implementation.

Applications/DataExplorer/NetCdfDialog/NetCdfConfigureDialog.cpp
```diff
@@ -396,7 +347,7 @@ QString NetCdfConfigureDialog::setName()
 std::string NetCdfConfigureDialog::getName()
 {
     std::string name = (lineEditName->text()).toStdString();
-    QString date = dateTimeEditDim3->date().toString(Qt::LocalDate);
+    QString date = dateTimeEditDim3->value();
```
TomFischer:
```suggestion
    QString const date = dateTimeEditDim3->value();
```

Applications/DataExplorer/NetCdfDialog/NetCdfConfigureDialog.cpp
```diff
@@ -216,81 +200,53 @@ void NetCdfConfigureDialog::setDimensionSelect()
...
-    int timeArray[1] = {spinBoxDim4->value()};
-    int currentValueDim3 = dim3Var->get_index(timeArray);
-    if (currentValueDim3 < 0) currentValueDim3=0; //if the value isn't found in the array, set it to 0 as default...
-    return currentValueDim3;
+    NcVar const& dim3Var =
+        _currentFile.getVar(comboBoxDim4->currentText().toStdString());
+    std::vector<std::size_t> start{{static_cast<std::size_t>(spinBoxDim4->value())}};
+    std::vector<std::size_t> length{{1}};
+    int value(0);
+    dim3Var.getVar(start, length, &value);
```
endJunction: Can you try this, please?
```suggestion
    dim3Var.getVar(start, {1}, &value);
```
If it works, same above around 220...

Applications/DataExplorer/NetCdfDialog/NetCdfConfigureDialog.cpp
rinkk: Yes, this is needed when we convert the netCDF data into image data.

Applications/DataExplorer/NetCdfDialog/NetCdfConfigureDialog.cpp
rinkk: It's used in for the signal that is implicitely created by this method.

I've added ```Q_UNUSED(id)``` so no warning will be displayed.

Applications/DataExplorer/NetCdfDialog/NetCdfConfigureDialog.cpp
rinkk: :heavy_check_mark:

Applications/DataExplorer/NetCdfDialog/NetCdfConfigureDialog.cpp
rinkk: It works :)  
:heavy_check_mark:

bilke:
I fixed the Windows build (missing Conan package) which ran successfully. The other are still not fixed!

rinkk:
Fixed the warnings. Everything's working now except that nc enum error on Docker which I have no idea what to do about...

bilke:
Implemented in #2654.

PR #2646, opened 2019-09-03
[PL/CT] MPL for component transport process

renchao-lu:
- Material properties involved in the component transport process all invoked from MPL.
- Linear property extended for linear dependency on multiple variables.

1. [x] Feature description was added to the [changelog](https://github.com/ufz/ogs/wiki/Release-notes-6.2.2)
2. [x] Tests covering your feature were added?
3. [ ] Any new feature or behavior change was documented?

PR #2649, opened 2019-09-04
[MatLib/MPL] Extend linear property

renchao-lu:
Following from Tom's suggestion, PR #2646 is split into two parts. As titled, this pull request extends for linear dependency on multiple variables.


MaterialLib/MPL/Properties/LinearProperty.cpp
```diff
@@ -25,21 +27,38 @@ PropertyDataType LinearProperty::value(
...
+    auto calculate_linearized_ratio =
+        [&variable_array](double initial_linearized_ratio, auto const& iv) {
+            return initial_linearized_ratio +
+                   std::get<double>(iv.slope) *
+                       (std::get<double>(
+                            variable_array[static_cast<int>(iv.type)]) -
+                        std::get<double>(iv.reference_condition));
+        };
+
+    double linearized_ratio_to_reference_value =
```
endJunction:
```suggestion
    double const linearized_ratio_to_reference_value =
```

MaterialLib/MPL/Properties/LinearProperty.cpp
```diff
@@ -25,21 +27,38 @@ PropertyDataType LinearProperty::value(
...
     double const /*t*/) const
 {
-    return std::get<double>(_value) *
-           (1 +
-            std::get<double>(_independent_variable.slope) *
-                (std::get<double>(variable_array[static_cast<int>(
-                     _independent_variable.type)]) -
-                 std::get<double>(_independent_variable.reference_condition)));
+    auto calculate_linearized_ratio =
+        [&variable_array](double initial_linearized_ratio, auto const& iv) {
```
endJunction:
```suggestion
        [&variable_array](double const initial_linearized_ratio, auto const& iv) {
```

MaterialLib/MPL/Properties/LinearProperty.cpp
```diff
@@ -25,21 +27,38 @@ PropertyDataType LinearProperty::value(
...
+
+    return std::get<double>(_value) * linearized_ratio_to_reference_value;
 }
 
 
 PropertyDataType LinearProperty::dValue(VariableArray const& /*variable_array*/,
                                         Variable const primary_variable) const
 {
-    return _independent_variable.type == primary_variable
+    auto independent_variable =
```
endJunction:
```suggestion
    auto const independent_variable =
```

MaterialLib/MPL/VariableType.h
```diff
@@ -42,6 +42,7 @@ using Tensor = std::array<double, 9>;
 /// is missing, simply add it somewhere at the list, but above the last entry.
 enum class Variable : int
 {
+    concentration,
```
endJunction:
```suggestion

```
Does not belong into this PR, I guess...

MaterialLib/MPL/Properties/LinearProperty.cpp
```diff
@@ -25,21 +27,38 @@ PropertyDataType LinearProperty::value(
...
+        std::find_if(_independent_variables.begin(),
```
wenqing: Maybe you can use an array like
```
using IndependentVariableArray =
    std::array<IndependentVariable, static_cast<int>(Variable::number_of_variables)>;
IndependentVariableArray const _independent_variables; /// With initial value for specified entries
``
Then you can get the value by the entry index of `static_cast<int>( primary_variable).

endJunction: @wenqing If I understood you correctly, instead of a `vector` an `array` should be used with maximum possible variables.
Then we would need to store another list of integeres indicating in which variables the LinearProperty is multi-linear.

TomFischer: The additional list of integers wouldn't make the code better readable. And I don't think we would gain much speedup.

Another possibility would be to set the slope of the independent variables in which the property is not linear to zero - but this implies more  computations. I would keep @renchao-lu implementation as it is.

MaterialLib/MPL/VariableType.h
renchao-lu: Moved this commit into PR #2646

MaterialLib/MPL/Properties/LinearProperty.cpp
renchao-lu: Fixed.

MaterialLib/MPL/Properties/LinearProperty.cpp
renchao-lu: Fixed.

MaterialLib/MPL/Properties/LinearProperty.cpp
renchao-lu: Fixed.

MaterialLib/MPL/Properties/LinearProperty.cpp
renchao-lu: By far I keep the implementation as it is.

wenqing: Ok.

endJunction:
@renchao-lu Please add the description to the changelog. Also in other PR's.

renchao-lu:
> @renchao-lu Please add the description to the changelog. Also in other PR's.

@endJunction  Description of this pull request added half an hour ago. ^-^

PR #2655, opened 2019-09-10
Cleanup of clang-tidy warnings

endJunction:
The clang-tidy warnings have exceeded a threshold of 200 warnings recently.
This causes the CI builds to fail.

A doxygen warning and potential for future warnings is also fixed.
Review commit-wise.


rinkk:
:+1:

(except that one bigger change in CreepBGRa.cpp which I didn't follow)

PR #2659, opened 2019-09-11
Fix failed large tests

renchao-lu:
As titled.

### DIFF
--- a/Documentation/ProjectFile/properties/property/LinearProperty/independent_variables/independent_variable/i_independent_variable.md
+++ b/Documentation/ProjectFile/properties/property/LinearProperty/independent_variables/independent_variable/i_independent_variable.md
@@ -1,1 +1,1 @@
-A linear constitutive property.
+Define specific variable type that linear material property depends upon.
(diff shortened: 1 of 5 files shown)


renchao-lu:
> As already mentioned, I like this better because the simple linear property does not require additional changes.
> Now I'd probably also not add any of the `<boundary_conditions` _etc._ but simply list the tags required number of times.

As discussed, I tried to use `getConfigParameterList` instead of `getConfigSubtreeList`... When doing the substitution work, I figured out `getConfigParameterList` could not be used here. `getConfigParameterList` does suit for single-valued tag and does not for those have children tags.

endJunction:
You are right about the get**Parameter**List. Somehow I didn't think about it.
A quick alternative is to fix the large tests ;)

renchao-lu:
> You are right about the get**Parameter**List. Somehow I didn't think about it.
> A quick alternative is to fix the large tests ;)

In order to get rid of the red light, it would be better to merge the fix-ups first. I will set up another pull request to drop the redundant parent tags, e.g., `<independent_variables>`...

PR #2663, opened 2019-09-12
Time loop refactorings

endJunction:
Moving and extracting functions.
One behavioural change is that now the `postTimestep` calls for monolithic processes happen in a batch after solve for all process as it is done in the staggered scheme.

### DIFF
--- a/ProcessLib/TimeLoop.cpp
+++ b/ProcessLib/TimeLoop.cpp
@@ -539,8 +539,6 @@ static NumLib::NonlinearSolverStatus solveMonolithicProcess(
 
     auto const nonlinear_solver_status = solveOneTimeStepOneProcess(
         process_data.process_id, x, timestep_id, t, dt, process_data, output);
-    pcs.postTimestep(x, t, dt, process_data.process_id);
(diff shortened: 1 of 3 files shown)


ProcessLib/TimeLoop.cpp
```diff
@@ -386,58 +379,70 @@ double TimeLoop::computeTimeStepping(const double prev_dt, double& t,
...
+                        &Output::doOutput);
     }
 
+}
+
+/*
+ * TODO:
+ * Now we have a structure inside the time loop which is very similar to the
+ * nonlinear solver. And admittedly, the control flow inside the nonlinear
+ * solver is rather complicated. Maybe in the future con can introduce an
```
TomFischer:
```suggestion
 * solver is rather complicated. Maybe in the future we can introduce an
```

endJunction:
I'd like to merge it today.

ProcessLib/TimeLoop.cpp
wenqing: `con can` --> we can

PR #2667, opened 2019-09-17
Time loop refactorings; Extract pre and post.

endJunction:
Extract `preTimestep` and `postTimestep` calls into individual functions merging the monolithic and staggered implementations.

No behavioural change for the staggered scheme.
For the monolithic scheme the `preTimestep` calls are now executed separately in the beginning.

### DIFF
--- a/ProcessLib/TimeLoop.cpp
+++ b/ProcessLib/TimeLoop.cpp
@@ -447,7 +447,7 @@ bool TimeLoop::loop()
     double t = _start_time;
     std::size_t accepted_steps = 0;
     std::size_t rejected_steps = 0;
-    NumLib::NonlinearSolverStatus nonlinear_solver_status{true, 0};
+    NumLib::NonlinearSolverStatus nonlinear_solver_status;
 
     double dt = computeTimeStepping(0.0, t, accepted_steps, rejected_steps);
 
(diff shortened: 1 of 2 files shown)


ProcessLib/TimeLoop.cpp
```diff
@@ -550,16 +561,45 @@ static std::string const nonlinear_fixed_dt_fails_info =
...
+    }
+}
+
 NumLib::NonlinearSolverStatus TimeLoop::solveUncoupledEquationSystems(
     const double t, const double dt, const std::size_t timestep_id)
 {
-    NumLib::NonlinearSolverStatus nonlinear_solver_status;
+    preTimestepForAllProcesses(t, dt, _per_process_data, _process_solutions);
+
+    NumLib::NonlinearSolverStatus nonlinear_solver_status{false, -1};
```
wenqing: For a readable source code, suggest to define the two variables for the arguments of the ctor of NumLib::NonlinearSolverStatus.  For example:

```suggestion
    const bool error_norms_met = false;
    const int number_iterations = -1
    NumLib::NonlinearSolverStatus nonlinear_solver_status{error_norms_met, number_iterations};
```

ProcessLib/TimeLoop.cpp
```diff
@@ -613,18 +648,9 @@ TimeLoop::solveCoupledEquationSystemsByStaggeredScheme(
...
-        {
-            auto& x = *_process_solutions[process_id];
-            process_data->process.preTimestep(x, t, dt, process_id);
-            ++process_id;
-        }
-    }
+    preTimestepForAllProcesses(t, dt, _per_process_data, _process_solutions);
 
-    NumLib::NonlinearSolverStatus nonlinear_solver_status{true, 0};
+    NumLib::NonlinearSolverStatus nonlinear_solver_status{false, -1};
```
wenqing: Here the same.

PR #2668, opened 2019-09-17
[HM] Allow principal stress output

FZill:
An extension of HM to allow the Output of the principal stresses and the corresponding vectors.
Stress and vectors are seperated, because multiplying the vector components with the principal stress results in losing the information about the sign of the principal stress.

The data is stored as a MeshLib::PropertyVector and is therfore saved automatically to the output. If this results in a significant performance loss a flag to disable it will be added.

2 Tests verifying the results, both being 2D-rotationally symmetric
(since the sign of the eigenvectors is ambiguos, testing with Tests.cmake-file can result in opposite directed vectors and therefore fail although the result is also valid. Thus it will not be testet. It could be testet if we implement testing only for the absolute values of the parameters)


ProcessLib/HydroMechanics/HydroMechanicsFEM.h
```diff
@@ -333,6 +360,34 @@ class HydroMechanicsLocalAssembler : public LocalAssemblerInterface
         return cache;
     }
 
+    std::vector<double> const& getIntPtPrincipleStress(
```
endJunction:
```suggestion
    std::vector<double> const& getIntPtPrincipalStress(
```

endJunction:
Adding principal vectors too...

FZill:
Current version of the test creates results on the server that deviate too much from the reference solution (10e-4). But the reference was deviating from the analytical solution quite a bit itself. Maybe Quad9-Elements will serve better for this kind of problem. But according to @wenqing there seems to be a bug in HM in using Quad9-elements.

FZill:
Quad9 Elemtents don't reduce the difference to the analytical solution. There seems to be something else which is causing the relatively large error for sigma_rr on the inner surface of the sphere.

FZill:
I am not 100% sure if this is the cause for the increased difference in stress at the inside of the sphere but it could explain it: In the derivation of the solution equations the book states that "In the current implementation, normal vector and surface area are evaluated on the undeformed geometry only. The pressure load is thus conservative and not following the (infinitesimal) deformation of the boundary contour." This was for OGS5 but I think it is still true for OGS6 and could therefore could be the solution for the problem.

nagelt:
@FZill The sphere stays a sphere, and the analytical solution also assumes infinitesimal strains, correct? So probably the conservative pressure load is not the cause. Approximation of the boundary contour could be one issue in deviating from the analytical solution (edgy sphere). Is your solution closer to the analytical than the reference solution? Where does the reference solution come from -- OGS-5? And is the 10e-4 relative or absolute?

FZill:
@nagelt That's correct. The simulation result on my unit was the reference solution which was already not matching the analytical solution very well. Especially the error spike on the inner radius is problematic :
![4k elements, p_out = 1e5] (https://drive.google.com/open?id=1wLyS_QGN1xONs4SzMeOAeE5PywDHF1Vc)
The 10e-4 was the relative difference  between server solution and reference.

I tried again with a pressure load on the outside p_out = 1e4 instead of 1e5 as before and the error spike on the inside went also down from 1e-1 to 1e-2:
![4k elements, p_out = 1e4] (https://drive.google.com/open?id=16Xwyaeu8S2-nouJ6iIdon02EzjDEbh1m)
So this reduces the relative error spike, but is obious not useful.
I tried your suggestion and quadrupeled the elements in radial and circumferential direction which brought the error down on the whole body but the spike remains:
![64k elements, p_out = 1e5] (https://drive.google.com/open?id=1Py0S-Fh8F_zcDxVtKWOY-EBB-cjj6GxU)
With that number of elements, lowering p_out again achieves the best result:
![64k elements, p_out = 1e4] (https://drive.google.com/open?id=1v9F2Pszz71KQoxobSknEer0eRdRCCfy1)

But needing such a high number of elements for such a seemingly simple problem seems a bit exaggerated...

FZill:
The sorting of the vectors  was abandoned since it would not have worked for all cases. So it is recommended when showing the vectors in paraview to represent them just as lines isntead of arrows. 

Also the "bad" agreement of numerical and analytical solution in the hollow sphere test was mainly due to the model having a "coarse" mesh. The PDF shows that the stresses and displacement do still converge even at a fine mesh. Since it is not favorable to make the test unnecessary long, the coarse variant is now the one being testet.

FZill:
@endJunction In the test with tube the displacement was already at about e-13 close to the analytical solution. We just didnt export the nodal values in paraview but instead only interpolated ones. 
And we were both right, the difference between Neumann and NormalTraction decreases with a finer mesh.

endJunction:
@FZill https://jenkins.opengeosys.org/blue/organizations/jenkins/ufz%2Fogs/detail/PR-2668/22/pipeline/  :

```
Comparing data array `sigma_1_vec' from file

[2019-10-12T16:08:50.440Z]   `/gpfs1/data/ogs/jenkins-frontend2/workspace/ufz_ogs_PR-2668/Tests/Data/HydroMechanics/Principle_Stress/Tube/output_pcs_0_ts_1_t_1.000000.vtu'
...
abs maximum norm = [3.333333333333342e+00,...
```

FZill:
Output is now per element and based on the average stress tensor.
Previously, the nodal results had fluctuations and gave no coherent result, probably due to the interpolation and thus could not properly be testet on the different platforms.

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
```diff
@@ -764,6 +764,35 @@ void HydroMechanicsLocalAssembler<ShapeFunctionDisplacement,
...
+
+    auto sigma_eff_sum = MathLib::KelvinVector::tensorToKelvin<DisplacementDim>(
+        Eigen::Matrix<double, 3, 3>::Zero());
+
+    for (auto const& ip_data : _ip_data)
+    {
+        sigma_eff_sum += ip_data.sigma_eff;
+    }
+
+    Eigen::Matrix<double, 3, 3, 0, 3, 3> sigma_avg =
```
TomFischer: Does this compile:
```suggestion
    Eigen::Matrix<double, 3, 3, 0, 3, 3> const sigma_avg =
```

FZill: yes, that works

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
```diff
@@ -764,6 +764,35 @@ void HydroMechanicsLocalAssembler<ShapeFunctionDisplacement,
...
+        e_s.eigenvalues();
+
+    auto eigen_vectors = e_s.eigenvectors();
+
+    for (auto i = 0; i < 3; i++)
+    {
+        Eigen::Map<Eigen::Vector3d>(
+            &(*_process_data.principal_stress_vector[i])[elem_id * 3], 3) =
+            eigen_vectors.col(i);
+    }
```
wenqing: Are the entries of eigen_vectors sorted?  σ_1 = max ( λ 1 , λ 2 , λ 3 ), σ 3 = min ( λ 1 , λ 2 , λ 3 ).

FZill: According to the Eigen docu the eigenvalues are sorted in increasing order and the eigenvectors correspond to the eigenvalue of the same number. So it would be opposite: σ_1 = min ( λ 1 , λ 2 , λ 3 ), σ 3 = max ( λ 1 , λ 2 , λ 3 )

PR #2221, opened 2018-09-26
Implementation of HeatTransportBHE process

HBShaoUFZ:
In this pull request, the HeatTransport BHE process is implemented.

PR #2671, opened 2019-09-19
Using MPL in Heat_Transport_BHE process

HBShaoUFZ:
In this PR, the source code and input files related to HEAT_TRANSPORT_BHE process were changed in order to use the MPL data structure for reading in soil parameters such as thermal conductivity and heat capacity. 

1. The MPL data structure was used when assembling the local matrix for soil part. 
2. All test files have been updated accordingly. 
3. When checking the number of material groups, the original FATAL error was replaced by a WARNING, as the BHEs are additional material groups than the number of medias defined.

### DIFF
--- a/MaterialLib/MPL/CreateMaterialSpatialDistributionMap.cpp
+++ b/MaterialLib/MPL/CreateMaterialSpatialDistributionMap.cpp
@@ -28,7 +28,7 @@ createMaterialSpatialDistributionMap(
 
     if (max_material_id > static_cast<int>(media.size() - 1))
     {
-        OGS_FATAL(
+        WARN(
             "The maximum value of MaterialIDs in mesh is %d. As the given "
             "number of porous media definitions in the project file is %d, the "
             "maximum value of MaterialIDs in mesh must be %d (index starts "
--- a/ProcessLib/HeatTransportBHE/LocalAssemblers/HeatTransportBHELocalAssemblerSoil-impl.h
+++ b/ProcessLib/HeatTransportBHE/LocalAssemblers/HeatTransportBHELocalAssemblerSoil-impl.h
@@ -13,6 +13,7 @@
 #include <valarray>
 #include <vector>
 
+#include "MaterialLib/MPL/Medium.h"
 #include "MathLib/LinAlg/Eigen/EigenMapTools.h"
 #include "NumLib/Fem/ShapeMatrixPolicy.h"
 #include "ProcessLib/HeatTransportBHE/HeatTransportBHEProcessData.h"
@@ -83,11 +84,16 @@ void HeatTransportBHELocalAssemblerSoil<ShapeFunction, IntegrationMethod>::
     auto local_K = MathLib::createZeroedMatrix<NodalMatrixType>(
         local_K_data, ShapeFunction::NPOINTS, ShapeFunction::NPOINTS);
 
-    unsigned const n_integration_points =
-        _integration_method.getNumberOfPoints();
-
     ParameterLib::SpatialPosition pos;
     pos.setElementID(_element_id);
+
+    auto const& medium = *_process_data.media_map->getMedium(_element_id);
+    auto const& solid_phase = medium.phase("Solid");
+
+    MaterialPropertyLib::VariableArray vars;
+
+    unsigned const n_integration_points =
+        _integration_method.getNumberOfPoints();
 
     for (unsigned ip = 0; ip < n_integration_points; ip++)
     {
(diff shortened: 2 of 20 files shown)


MaterialLib/MPL/CreateMaterialSpatialDistributionMap.cpp
```diff
@@ -28,7 +28,7 @@ createMaterialSpatialDistributionMap(
 
     if (max_material_id > static_cast<int>(media.size() - 1))
     {
-        OGS_FATAL(
+        WARN(
```
endJunction: Please provide explanation in the commit message.

ProcessLib/HeatTransportBHE/LocalAssemblers/HeatTransportBHELocalAssemblerSoil-impl.h
```diff
@@ -83,11 +84,20 @@ void HeatTransportBHELocalAssemblerSoil<ShapeFunction, IntegrationMethod>::
     auto local_K = MathLib::createZeroedMatrix<NodalMatrixType>(
         local_K_data, ShapeFunction::NPOINTS, ShapeFunction::NPOINTS);
 
-    unsigned const n_integration_points =
-        _integration_method.getNumberOfPoints();
-
     ParameterLib::SpatialPosition pos;
-    pos.setElementID(_element_id);
+    pos.setElementID(this->_element_id);
```
endJunction: Why `this`?

ProcessLib/HeatTransportBHE/LocalAssemblers/HeatTransportBHELocalAssemblerSoil-impl.h
```diff
@@ -96,20 +106,36 @@ void HeatTransportBHELocalAssemblerSoil<ShapeFunction, IntegrationMethod>::
...
+            solid_phase
+                .property(
+                    MaterialPropertyLib::PropertyType::specific_heat_capacity)
+                .template value<double>(vars, pos, t);
+
+        // auto const density_f =
+        // fluid_phase.property(MaterialPropertyLib::PropertyType::density).template
+        // value<double>(vars, pos, t); auto const density_g =
+        // gas_phase.property(MaterialPropertyLib::PropertyType::density).template
+        // value<double>(vars, pos, t);
```
endJunction: These "comments" were not touched for the last two years. I'd recommend to remove them.
If someday there is a need for those densities *etc.* it can be easily copied from the solid properties. No need to fix that dead "code"

Actually it was discussed in the original PR and nothing has changed: https://github.com/ufz/ogs/pull/2221#discussion_r231264509

Tests/Data/Parabolic/T/3D_2U_BHE/3D_2U_BHE.prj
```diff
@@ -63,6 +54,94 @@
...
+                        <!-- <property> -->
+                            <!-- <name>permeability</name> -->
+                            <!-- <type>Constant</type> -->
+                            <!-- <value>1.e-14 0 0 1.e-14</value> -->
+                        <!-- </property> -->
+                        <!-- <property> -->
+                            <!-- <name>porosity</name> -->
+                            <!-- <type>Constant</type> -->
+                            <!-- <value>0.001</value> -->
+                        <!-- </property> -->
```
endJunction: These properties seems to be unused currently. I strongly recommend to drop the unused, commented code parts.

ProcessLib/HeatTransportBHE/LocalAssemblers/HeatTransportBHELocalAssemblerSoil-impl.h
HBShaoUFZ: It's not needed. Removed.

ProcessLib/HeatTransportBHE/LocalAssemblers/HeatTransportBHELocalAssemblerSoil-impl.h
HBShaoUFZ: Although we will need them soon in order to include GW flow and partially saturated processes, they are not needed for now. Therefore removed.

Tests/Data/Parabolic/T/3D_2U_BHE/3D_2U_BHE.prj
HBShaoUFZ: Same as the above. Removed for now.

MaterialLib/MPL/CreateMaterialSpatialDistributionMap.cpp
HBShaoUFZ: Explanation was added in the commit [4ffd04f](https://github.com/ufz/ogs/pull/2671/commits/4ffd04f7a989526c049f36287c482fef489e0e10)

ProcessLib/HeatTransportBHE/LocalAssemblers/HeatTransportBHELocalAssemblerSoil-impl.h
```diff
@@ -83,33 +84,41 @@ void HeatTransportBHELocalAssemblerSoil<ShapeFunction, IntegrationMethod>::
     auto local_K = MathLib::createZeroedMatrix<NodalMatrixType>(
         local_K_data, ShapeFunction::NPOINTS, ShapeFunction::NPOINTS);
 
-    unsigned const n_integration_points =
-        _integration_method.getNumberOfPoints();
-
     ParameterLib::SpatialPosition pos;
     pos.setElementID(_element_id);
 
+    auto const& process_data = this->_process_data;
```
endJunction: Same here *etc*.

HBShaoUFZ: Actually this local process_data is not necessary. Removed.

MaterialLib/MPL/CreateMaterialSpatialDistributionMap.cpp
TomFischer: Why isn't this an OGS_FATAL? I would expect that bad things could be happen if something in the input project file is inconsistent with the mesh file.

HBShaoUFZ: The problem originates from the fact that we use the material IDs to differentiate the 1D line elements assigned to the BHE in a 3D mesh. The simplest case would be, we have one material ID assigned to 3D prism elements for all the soil part. Then we have the 2nd material ID assigned to the one BHE. In this case, our max_material_id  == 1. However, we can only define 1 media in the input files (which is the soil media). This means media.size() = 1. The if clause on line 29 is then true and we will trigger the block. If we keep the OGS_FATAL function, then we will not be able to run BHE simulations with different types of materials (soil layers). I discussed this with Dima and we agreed to change it to WARN.

TomFischer: Thanks for explaining the background. :fast_forward:

PR #2672, opened 2019-09-20
[PL/HT] Move the permeability specification from solid phase to medium scale.

TomFischer:
Used the following bash script for changing project files:
```
#!/usr/bin/env bash

number_of_medium_tags=`xmlstarlet sel -t -v "count(///*/medium)" $1`
number_of_medium_tags=$(($number_of_medium_tags - 1))

xmlstarlet ed --move "//OpenGeoSysProject/media/medium[@id=0]/phases/phase/properties/property[name='permeability']" "//OpenGeoSysProject/media/medium[@id=0]/properties" $1 > /tmp/t0.prj

for i in `seq 1 ${number_of_medium_tags}`;
do
    i_m_e=$(($i-1))
    #echo "${i}, ${i_m_e}"
    path="//OpenGeoSysProject/media/medium[@id="${i}"]"
    xmlstarlet ed --move "$path/phases/phase/properties/property[name='permeability']" "$path/properties" /tmp/t$i_m_e.prj > /tmp/t${i}.prj
done

#format the result file
xmlstarlet fo -s 4 /tmp/t${number_of_medium_tags}.prj > ${1}
```

### DIFF
--- a/ProcessLib/HT/HTFEM.h
+++ b/ProcessLib/HT/HTFEM.h
@@ -130,11 +130,9 @@ public:
         auto const& medium =
             *_process_data.media_map->getMedium(_element.getID());
         auto const& liquid_phase = medium.phase("AqueousLiquid");
-        auto const& solid_phase = medium.phase("Solid");
 
         auto const K = MaterialPropertyLib::formEigenTensor<GlobalDim>(
-            solid_phase
-                .property(MaterialPropertyLib::PropertyType::permeability)
+            medium.property(MaterialPropertyLib::PropertyType::permeability)
                 .value(vars, pos, t));
 
         auto const mu =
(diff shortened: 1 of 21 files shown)


ProcessLib/HT/HTFEM.h
```diff
@@ -294,8 +293,7 @@ class HTFEM : public HTLocalAssemblerInterface
                 MaterialPropertyLib::Variable::phase_pressure)] = p_int_pt;
 
             auto const K = MaterialPropertyLib::formEigenTensor<GlobalDim>(
-                solid_phase
```
endJunction: `solid_phase` is now unused in line 275

renchao-lu:
`Porosity` would be better also set as of medium property.

TomFischer:
> `Porosity` would be better also set as of medium property.

I'll prepare another PR for the changes.

ProcessLib/HT/HTFEM.h
endJunction: bump

PR #2673, opened 2019-09-20
Pass dt to assemble

endJunction:
Refactoring only.

The time and time increment are stored in different places like process data or coupled solutions. Another set of time and time increment is passed through the interfaces of assemble *etc.* functions.
To simplify (and to avoid unintentional mismatch between those) now `t` and `dt` are passed through interface to few more functions.

The bulk of the changes is to the interfaces. Some notable changes:
 - drop `dt` in coupled solutions: https://github.com/ufz/ogs/pull/2673/files#diff-81c8910638eee8d67ecaaca35bd29ac9
 - drop `t` and `dt` in all process data, *e.g.* in HM: https://github.com/ufz/ogs/pull/2673/files#diff-5caaa3748f350278bfadf3a7ef72be64
 - add getter for time increment in time discretizations: https://github.com/ufz/ogs/pull/2673/files#diff-1205c5023ac964f635a60cbde2984929
 - some changes in the time loop: https://github.com/ufz/ogs/pull/2673/files#diff-837505db22a18ac0f5a3d8dc15bb35cd

Changelog updated.

### DIFF
--- a/NumLib/ODESolver/ODESystem.h
+++ b/NumLib/ODESolver/ODESystem.h
@@ -49,9 +49,9 @@ public:
     virtual void preAssemble(const double t, GlobalVector const& x) = 0;
 
     //! Assemble \c M, \c K and \c b at the provided state (\c t, \c x).
-    virtual void assemble(const double t, GlobalVector const& x,
-                          GlobalMatrix& M, GlobalMatrix& K,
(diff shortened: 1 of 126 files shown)


endJunction:
Compiles. Failure due to conan.

PR #2675, opened 2019-09-23
Fix Conan Boost download

bilke:
Related Conan issue: https://github.com/conan-io/conan/issues/5780

If you encounter this locally on your workstation upgrade Conan (to 1.18.5):

```
pip install --upgrade conan
```

### DIFF
--- a/scripts/docker/Dockerfile.clang.full
+++ b/scripts/docker/Dockerfile.clang.full
@@ -82,12 +82,12 @@ RUN apt-get update -y && \
         python3-setuptools \
         python3-wheel && \
     rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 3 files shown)


endJunction:
Maybe a changelog entry for new minimum unless it is tracked elsewhere.

bilke:
@endJunction I do not know if this appears on older Conan versions. We have a Conan version enforcement in our CMake checks. Shall I set the min req. version to 1.18.5?

endJunction:
So the bug is maybe for specific version of conan but not for older versions.
Then let's keep it as it is for now unless users continue to encounter this issue with download, then we set a new minimum.

PR #2679, opened 2019-09-24
Eliminate process_id from coupled solutions class

endJunction:
Continuing the time loop refactorings, a second change (with big trail in the processes) eliminating `process_id` from the CoupledSolutionsForStaggeredScheme (and the local class therein).

Review commit-wise.

Main change in [ProcessLib/CoupledSolutionsForStaggeredScheme.h](https://github.com/ufz/ogs/pull/2679/files#diff-81c8910638eee8d67ecaaca35bd29ac9)

### DIFF
--- a/NumLib/ODESolver/NonlinearSolver.cpp
+++ b/NumLib/ODESolver/NonlinearSolver.cpp
@@ -22,15 +22,15 @@
 namespace NumLib
 {
 void NonlinearSolver<NonlinearSolverTag::Picard>::assemble(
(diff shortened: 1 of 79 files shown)


endJunction:
@wenqing I'd rather fix the other occurrences. The reason is to have constant arguments first, and then the output arguments, and not to mix them. Would that be OK? I'll open another PR for that part.

wenqing:
:fast-forward

PR #2686, opened 2019-09-27
Flatten local coupled solutions vector.

endJunction:
Yet another staggered coupling refactoring.

Instead of passing a vector of a vector with data, pass a flat vector of (local) solutions ordered the same as the processes.

Review commit-wise.

### DIFF
--- a/ProcessLib/CoupledSolutionsForStaggeredScheme.cpp
+++ b/ProcessLib/CoupledSolutionsForStaggeredScheme.cpp
@@ -12,6 +12,8 @@
 
 #include "CoupledSolutionsForStaggeredScheme.h"
 
+#include <numeric>
+
 #include "MathLib/LinAlg/LinAlg.h"
 #include "Process.h"
 
@@ -27,7 +29,7 @@ CoupledSolutionsForStaggeredScheme::CoupledSolutionsForStaggeredScheme(
     }
 }
 
-std::vector<std::vector<double>> getPreviousLocalSolutions(
+std::vector<double> getPreviousLocalSolutions(
     const CoupledSolutionsForStaggeredScheme& cpl_xs,
     const std::vector<std::vector<GlobalIndexType>>& indices)
 {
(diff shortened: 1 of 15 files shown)


ProcessLib/CoupledSolutionsForStaggeredScheme.cpp
```diff
@@ -27,48 +29,33 @@ CoupledSolutionsForStaggeredScheme::CoupledSolutionsForStaggeredScheme(
...
+
+    int process_id = 0;
+    for (auto const& x : global_solutions)
     {
-        local_xs_t0.emplace_back(x_t0->get(indices[coupling_id]));
-        coupling_id++;
+        auto const& values = x->get(indices[process_id]);
+        result.insert(cend(result), cbegin(values), cend(values));
+        process_id++;
     }
```
renchao-lu: How about writing as follows,
```
std::vector<double> local_solutions;
local_solutions.reserve(local_solution_size);

for (int process_id = 0; process_id < static_cast<int>(global_solutions); ++process_id)
{
    auto const& local_solution = global_solutions[process_id]->get(indices[process_id]);
    std::copy(local_solution.begin(), local_solution.end(), std::back_inserter(local_solutions));
}
```

endJunction: I see three suggestions:
 - rename result to `local_solutions`. That's easy.
 - use process_id as loop counter. Will think about it.
 - use copy not insert. Copy is slower because of push-back one-by-one. Insert works (in this case) on a chunk. But another thing can be done, is to use `std::make_move_iterator`.
Will push the changes in a minute.

endJunction:
bump

ProcessLib/PhaseField/PhaseFieldFEM.h
```diff
@@ -107,6 +107,14 @@ template <typename ShapeFunction, typename IntegrationMethod,
...
 class PhaseFieldLocalAssembler : public PhaseFieldLocalAssemblerInterface
 {
+private:
+    static constexpr int displacement_index = 0;
+    static constexpr int displacement_size =
+        ShapeFunction::NPOINTS * DisplacementDim;
+    static constexpr int phasefield_index =
+        displacement_index + displacement_size;
+    static constexpr int phasefield_size = ShapeFunction::NPOINTS;
+
```
TomFischer: Are there any technical reasons to move these private internal definitions to the top of the class declaration before the public part? As a user I first want to see the interface and then maybe I'm interested in the private details.

endJunction: Yes, technical reason. The values are needed for typedefs below...

ProcessLib/PhaseField/PhaseFieldFEM.h
```diff
@@ -117,6 +125,14 @@ class PhaseFieldLocalAssembler : public PhaseFieldLocalAssemblerInterface
 
     using NodalForceVectorType = typename BMatricesType::NodalForceVectorType;
 
+    using DeformationVector =
+        typename ShapeMatricesType::template VectorType<displacement_size>;
+    using PhaseFieldVector =
+        typename ShapeMatricesType::template VectorType<phasefield_size>;
+    using PhaseFieldMatrix =
+        typename ShapeMatricesType::template MatrixType<phasefield_size,
+                                                        phasefield_size>;
```
endJunction: @TomFischer ... here.

PR #2691, opened 2019-10-08
[PL] CT; Remove unneeded getFlux implementation.

endJunction:
There is another one, which is used.

### DIFF
--- a/ProcessLib/ComponentTransport/ComponentTransportFEM.h
+++ b/ProcessLib/ComponentTransport/ComponentTransportFEM.h
@@ -826,27 +826,6 @@ public:
         auto const local_C = Eigen::Map<const NodalVectorType>(
             &local_x[first_concentration_index], concentration_size);
 
-        return calculateFlux(pnt_local_coords, t, local_p, local_C);
-    }
-
-    Eigen::Vector3d getFlux(
-        MathLib::Point3d const& pnt_local_coords,
-        double const t,
-        std::vector<std::vector<double>> const& local_xs) const override
-    {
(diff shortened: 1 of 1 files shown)


ProcessLib/ComponentTransport/ComponentTransportFEM.h
```diff
@@ -826,27 +826,6 @@ class LocalAssemblerData : public ComponentTransportLocalAssemblerInterface
...
-        MathLib::Point3d const& pnt_local_coords,
-        double const t,
-        std::vector<std::vector<double>> const& local_xs) const override
-    {
-        auto const local_p = MathLib::toVector(local_xs[hydraulic_process_id]);
-        auto const local_C =
-            MathLib::toVector(local_xs[first_transport_process_id]);
-
-        return calculateFlux(pnt_local_coords, t, local_p, local_C);
-    }
```
renchao-lu: Based upon my limited understanding, the dropped function `Eigen::Vector3d getFlux(MathLib::Point3d const& pnt_local_coords, double const t, std::vector<std::vector<double>> const& local_xs) const override` is used in case of staggered coupling scheme, while the other retained function `Eigen::Vector3d getFlux(MathLib::Point3d const& pnt_local_coords, double const t, double const t, std::vector<double> const& local_x) const override` is in case of monolithic coupling scheme...

renchao-lu:
I have just one doubt. Other changes look fine to me.

ProcessLib/ComponentTransport/ComponentTransportFEM.h
endJunction: If it would be used, I think, it would not compile after removal, isn't it?

PR #2692, opened 2019-10-09
[Utils/Partmesh] Add fake property array 'bulk_element_ids' 

renchao-lu:
Fake property array 'bulk_element_ids' added in Cell Data if not found when executing Partmesh.


TomFischer:
@renchao-lu I use the identifySubdomain tool and it worked for me all the time. What isn't working? Please, give a reason for the changes.

renchao-lu:
In some cases, property array `number_bulk_elements` is added in the Cell Data and property array `bulk_element_ids` added in the Field Data, when one works with the utility tool `identifySubdomain`. Then, one performs domain decomposition on the mesh files with the utility tools `partmesh` and `mpmetis` for parallelization. In this step, property array `bulk_element_ids` that resides in the Cell Data is required, while it is now in the Field Data rather than in the Cell Data. Hopefully, my interpretive description is clear.

renchao-lu:
I figure out this workaround is problematic when heterogeneous material property is defined over the whole domain.

PR #2693, opened 2019-10-09
Pass solutions to pre and post

endJunction:
`preTimestep` now accepts a vector of solution together with a `process_id`.
In `postTimestep` only the vector of solutions has been changed.


endJunction:
@wenqing I think passing the vector (reference) via interface is more flexible. Storing the vector (pointer/reference) in the process is actually what is done with `CoupledSolutionsForStaggeredScheme`, which I try to avoid.

PR #2694, opened 2019-10-09
Pass current solutions to BC's

endJunction:
Main reason for modification is the getFlux function, which for
particular processes (ComponentTransport) relies on all solutions being
available.
This modification ripples through all of the code, especially through
the boundary conditions.
Modified `applyNaturalBC`, ~`preTimestep`~, `integrate`, and `assemble`
functions.
The `preTimestep` is now decoupled from the secondary vars/BCs since #2697

### DIFF
--- a/ProcessLib/BoundaryCondition/ConstraintDirichletBoundaryCondition.h
+++ b/ProcessLib/BoundaryCondition/ConstraintDirichletBoundaryCondition.h
@@ -58,10 +58,10 @@ public:
         bool const lower,
         std::function<Eigen::Vector3d(std::size_t const,
                                       MathLib::Point3d const&, double const,
-                                      GlobalVector const&)>
+                                      std::vector<GlobalVector*> const&)>
             getFlux);
 
-    void preTimestep(double t, std::vector<GlobalVector*> const& x,
+    void preTimestep(double const t, std::vector<GlobalVector*> const& x,
                      int const process_id) override;
 
     void getEssentialBCValues(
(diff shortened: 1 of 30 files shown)


endJunction:
Turned out to be tied to preIteration and secondary variables output through CachedSecondaryVariables used only in one process for one variable. And there is a bug in it.
So first removing/replacing that CachedSecondaryVariables, then will come back to this PR.

TomFischer:
Ready for review?

ProcessLib/BoundaryCondition/ConstraintDirichletBoundaryCondition.h
```diff
@@ -113,9 +113,10 @@ class ConstraintDirichletBoundaryCondition final : public BoundaryCondition
     MeshLib::Mesh const& _bulk_mesh;
 
     /// The function _getFlux calculates the flux through the boundary element.
-    std::function<Eigen::Vector3d(
-            std::size_t const, MathLib::Point3d const&, double const,
-            GlobalVector const&)> _getFlux;
+    std::function<Eigen::Vector3d(std::size_t const, MathLib::Point3d const&,
+                                  double const,
+                                  std::vector<GlobalVector*> const& x)>
```
TomFischer: To stay consistent with the other parameters:
```suggestion
                                  std::vector<GlobalVector*> const&)>
```

TomFischer: Or add to all the other parameters useful names.

PR #2698, opened 2019-10-14
[PaL] Explicitly instantiate rotate*() templates.

endJunction:
Moving the code into cpp.

This saves repeated instantiations of the same template. Almost 300 instatitations.

### DIFF
--- a/ParameterLib/CoordinateSystem.cpp
+++ b/ParameterLib/CoordinateSystem.cpp
@@ -116,4 +116,39 @@ Eigen::Matrix<double, 3, 3> CoordinateSystem::transformation<3>(
 #endif  // NDEBUG
     return t;
 }
+
+template <int Dimension>
+Eigen::Matrix<double, Dimension, Dimension> CoordinateSystem::rotateTensor(
+    std::vector<double> const& values, SpatialPosition const& pos) const
+{
+    assert(values.size() == Dimension * Dimension ||
+           "Input vector has wrong dimension; expected 4 or 9 entries.");
+    auto const tensor =
+        Eigen::Map<Eigen::Matrix<double, Dimension, Dimension> const>(
+            values.data(), Dimension, Dimension);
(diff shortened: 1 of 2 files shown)


ParameterLib/CoordinateSystem.cpp
```diff
@@ -116,4 +116,39 @@ Eigen::Matrix<double, 3, 3> CoordinateSystem::transformation<3>(
 #endif  // NDEBUG
     return t;
 }
+
+template <int Dimension>
+Eigen::Matrix<double, Dimension, Dimension> CoordinateSystem::rotateTensor(
+    std::vector<double> const& values, SpatialPosition const& pos) const
+{
+    assert(values.size() == Dimension * Dimension ||
+           "Input vector has wrong dimension; expected 4 or 9 entries.");
```
wenqing:
```suggestion
           "Input vector has wrong size; expected 4 or 9 entries.");
```

ParameterLib/CoordinateSystem.cpp
```diff
@@ -116,4 +116,39 @@ Eigen::Matrix<double, 3, 3> CoordinateSystem::transformation<3>(
...
+    return R * tensor * R.transpose();
+}
+
+template <int Dimension>
+Eigen::Matrix<double, Dimension, Dimension>
+CoordinateSystem::rotateDiagonalTensor(std::vector<double> const& values,
+                                       SpatialPosition const& pos) const
+{
+    assert(values.size() == Dimension ||
+           "Input vector has wrong dimension; expected 2 or 3 entries.");
```
wenqing:
```suggestion
           "Input vector has wrong size; expected 2 or 3 entries.");
```

endJunction: Same meaning at the end, though subtly different. Keeping "mathematical" `dimension`.

PR #2705, opened 2019-10-29
Pass xs to secondary variables computation

endJunction:
One more update here on the staggered scheme refactoring.
This is for secondary variables, because some processes (CT) need access to all solutions when computing the secondary variables.

Very repetitive. So I split the interesting part from interface changes. Review commit-wise. You might want to look at your favourite process' changes and into the TimeLoop.cpp changes.

### DIFF
--- a/NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.h
+++ b/NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.h
@@ -50,8 +50,9 @@ public:
     void extrapolate(const unsigned num_components,
                      ExtrapolatableElementCollection const& extrapolatables,
                      const double t,
-                     GlobalVector const& current_solution,
-                     LocalToGlobalIndexMap const& dof_table) override;
+                     std::vector<GlobalVector*> const& x,
+                     std::vector<NumLib::LocalToGlobalIndexMap const*> const&
+                         dof_table) override;
 
     /*! \copydoc Extrapolator::calculateResiduals()
      *
@@ -64,8 +65,9 @@ public:
         const unsigned num_components,
         ExtrapolatableElementCollection const& extrapolatables,
         const double t,
-        GlobalVector const& current_solution,
-        LocalToGlobalIndexMap const& dof_table) override;
+        std::vector<GlobalVector*> const& x,
+        std::vector<NumLib::LocalToGlobalIndexMap const*> const& dof_table)
+        override;
 
     GlobalVector const& getNodalValues() const override
     {
@@ -82,8 +84,9 @@ private:
     void extrapolateElement(
         std::size_t const element_index, const unsigned num_components,
         ExtrapolatableElementCollection const& extrapolatables, const double t,
-        GlobalVector const& current_solution,
-        LocalToGlobalIndexMap const& dof_table, GlobalVector& counts);
+        std::vector<GlobalVector*> const& x,
+        std::vector<NumLib::LocalToGlobalIndexMap const*> const& dof_table,
+        GlobalVector& counts);
 
     //! Compute the residuals for one element
     void calculateResidualElement(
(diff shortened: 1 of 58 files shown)


ProcessLib/ComponentTransport/ComponentTransportFEM.h
```diff
@@ -706,41 +706,42 @@ class LocalAssemblerData : public ComponentTransportLocalAssemblerInterface
...
+        std::vector<std::vector<double>> local_xs;
+        local_xs.reserve(n_processes);
+
+        for (std::size_t process_id = 0; process_id < n_processes; ++process_id)
         {
-            auto const local_x = current_solution.get(indices);
+            auto const indices =
+                NumLib::getIndices(_element.getID(), *dof_table[process_id]);
+            assert(!indices.empty());
+            local_xs.push_back(x[process_id]->get(indices));
```
renchao-lu: `x` and `local_xs` stand for the global and local solution. For consistency in concept, I would name the local solution as `local_x`. What do you think?

NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.h
```diff
@@ -82,17 +84,18 @@ class LocalLinearLeastSquaresExtrapolator : public Extrapolator
     void extrapolateElement(
         std::size_t const element_index, const unsigned num_components,
         ExtrapolatableElementCollection const& extrapolatables, const double t,
-        GlobalVector const& current_solution,
-        LocalToGlobalIndexMap const& dof_table, GlobalVector& counts);
+        std::vector<GlobalVector*> const& x,
+        std::vector<NumLib::LocalToGlobalIndexMap const*> const& dof_table,
+        GlobalVector& counts);
```
TomFischer: I don't know how `counts` is used, but I wonder why isn't `counts` not a `std::vector<GlobalVector*>&`?

endJunction: If I remember correctly, the counts vector is for averaging the node values, so more topology related then DOF table. Anyway it's either correct or not tested. ;)

ProcessLib/TES/TESProcess.cpp
```diff
@@ -322,7 +322,7 @@ GlobalVector const& TESProcess::computeVapourPartialPressure(
     std::vector<NumLib::LocalToGlobalIndexMap const*> const& dof_table,
     std::unique_ptr<GlobalVector>& result_cache)
 {
-    assert(&dof_table == _local_to_global_index_map.get());
+    assert(dof_table[0] == _local_to_global_index_map.get());
```
TomFischer: I would prefer somewhere above: ```constexpr int process_id = 0;```
and then accessing the `dof_table` via `process_id` in the entire initializeConcreteProcess.

ProcessLib/ComponentTransport/ComponentTransportFEM.h
endJunction: :heavy_check_mark:.

ProcessLib/TES/TESProcess.cpp
endJunction: :heavy_check_mark:

ProcessLib/HT/MonolithicHTFEM.h
```diff
@@ -204,14 +204,15 @@ class MonolithicHTFEM
 
     std::vector<double> const& getIntPtDarcyVelocity(
         const double t,
-        GlobalVector const& current_solution,
-        NumLib::LocalToGlobalIndexMap const& dof_table,
+        std::vector<GlobalVector*> const& x,
+        std::vector<NumLib::LocalToGlobalIndexMap const*> const& dof_table,
         std::vector<double>& cache) const override
     {
+        int const process_id = 0;  // monolithic case.
```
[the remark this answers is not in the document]
TomFischer: I would keep the `process_id` for readability reasons.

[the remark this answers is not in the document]
wenqing: Then the comment has to be modified.

PR #2712, opened 2019-11-08
Cleanup / Fix warnings / Correct formatting

TomFischer:

### DIFF
--- a/Applications/DataExplorer/mainwindow.cpp
+++ b/Applications/DataExplorer/mainwindow.cpp
@@ -837,48 +837,33 @@ void MainWindow::about()
 
 QMenu* MainWindow::createImportFilesMenu()
 {
-    auto* signal_mapper = new QSignalMapper(this);  // owned by MainWindow
     QMenu* importFiles = new QMenu("&Import Files", this);
-    QAction* feflowFiles = importFiles->addAction("&FEFLOW Files...");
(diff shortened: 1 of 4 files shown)


Applications/DataExplorer/mainwindow.cpp
```diff
@@ -839,48 +837,33 @@ void MainWindow::about()
...
-    signal_mapper->setMapping(rasterFiles, ImportFileType::RASTER);
+    importFiles->addAction("&Petrel Files...",
+                           [=] { loadPetrelFiles(); });
+    importFiles->addAction("&Raster Files...",
+                           [=] { open(ImportFileType::RASTER); });
 #if defined VTKFBXCONVERTER_FOUND
-    QAction* rasterPolyFiles = importFiles->addAction("R&aster Files as PolyData...");
-    connect(rasterPolyFiles, SIGNAL(triggered()), this, SLOT(map()));
-    _signal_mapper->setMapping(rasterPolyFiles, ImportFileType::POLYRASTER);
+    importFiles->addAction("R&aster Files as PolyData...",
```
endJunction: @rinkk Is the placement of '&' correct or should it be
```suggestion
    importFiles->addAction("&Raster Files as PolyData...",
```

rinkk: ```R&aster``` is correct here.

PR #2714, opened 2019-11-13
Correct Lubby2 equations in the pdf file in web content folder

zhangning737:
Correct Lubby2 eq 6, eq8 and eq 9:

1. [ ] Eq6: switch the \eta_M and G_M
2. [ ] Eq9: add an identity tensor symbol


endJunction:
In (8) last term should be `J_13`.
Otherwise the programmed Jacobian corresponds with the pdf.

zhangning737:
> In (8) last term should be `J_13`.
> Otherwise the programmed Jacobian corresponds with the pdf.

Corrected...

PR #2715, opened 2019-11-15
Fix bug in groundwater advection feature of HeatTransportBHE process

boyanmeng:
In the original PR#2711, when running in debug mode, there will be assertion failures since Eigen::Map performs an internal check of the row and cols. Now they are corrected as the same with Vector3d.


ProcessLib/HeatTransportBHE/LocalAssemblers/HeatTransportBHELocalAssemblerSoil-impl.h
```diff
@@ -139,7 +139,7 @@ void HeatTransportBHELocalAssemblerSoil<ShapeFunction, IntegrationMethod>::
                 .property(MaterialPropertyLib::PropertyType::phase_velocity)
                 .value(vars, pos, t);
         auto const velocity = Eigen::Map<Eigen::Vector3d const>(
-            std::get<MaterialPropertyLib::Vector>(velocity_arr).data(), 1, 3);
+            std::get<MaterialPropertyLib::Vector>(velocity_arr).data(), 3, 1);
```
endJunction: I'd think you don't need to specify any of the size arguments in this case.
Could you try it w/o size arguments, please?

boyanmeng: You are right. Changed.

PR #2717, opened 2019-11-18
[PL/ST/Python] bulk dof table and bulk mesh id passed

jbathmann:
Before this fix, python source terms were only applicable, if the bulk mesh and the source mesh were of identical. To find the correct dof_idx, a location with the bulk_mesh_id seems to need to be passed to the bulk_dof_table.


TomFischer:
Need explanation:
Since the dof table of the source term is derived from the dof table of the bulk mesh, I thought that the dof table of the source term contains all necessary information. Do I have a wrong understanding.

jbathmann:
> Need explanation:
> Since the dof table of the source term is derived from the dof table of the bulk mesh, I thought that the dof table of the source term contains all necessary information. Do I have a wrong understanding.

I am also wondering about this @TomFischer. However, this was the way I got the following example working. Please note, that it is not a meaningful example, but just a project file testing the source term assembly.
[Uploading testcase.zip…](Example)

PR #2721, opened 2019-11-21
Matched non-associated tension cut-off in MCAS.

nagelt:
Minor change in the Mohr-Coulomb Abbo-Sloan model for non-associated flow with apex smoothing.


endJunction:
`load_test_mc` fails, see [log](https://jenkins.opengeosys.org/blue/organizations/jenkins/ufz%2Fogs/detail/PR-2721/1/pipeline#step-268-log-682). Sigma (relative) needs a relaxation to `2e-12` in the `Tests.cmake`; see line 292.

PR #2725, opened 2019-11-25
Python source term DOF table fix.

endJunction:
@jbathmann If this works (I tested it with your example and it seems to work) for you, then it is a smaller bugfix, than in #2717.

Bugfix is only in the first commit. Second is cleanup.

#Closes #2717


jbathmann:
@endJunction Thank you very much. Thi fix works for me!

PR #2730, opened 2019-12-01
Repeat time step on assembly error

endJunction:
Introduce explicit exception handling in the time loop/nonlinear-solver for failed assembly. In case of exception the time step will be repeated (if adaptive time stepping is used) with smaller time step size.
The exception is thrown in MFront material model when local integration fails.

3. [x] Any new feature or behavior change was documented? Not needed.

### DIFF
--- a/MaterialLib/SolidModels/MFront/MFront.cpp
+++ b/MaterialLib/SolidModels/MFront/MFront.cpp
@@ -11,6 +11,8 @@
 
 #include <MGIS/Behaviour/Integrate.hxx>
 
+#include "NumLib/Exceptions.h"
+
 namespace
 {
 /// Converts between OGSes and MFront's Kelvin vector indices.
@@ -313,7 +315,8 @@ MFront<DisplacementDim>::integrateStress(
     auto const status = mgis::behaviour::integrate(v, _behaviour);
     if (status != 1)
     {
-        OGS_FATAL("Integration failed with status %i.", status);
+        throw NumLib::AssemblyException("MFront: integration failed with status"
+                + std::to_string(status) + ".");
     }
 
     KelvinVector sigma;
--- a/Tests/Data/Mechanics/MohrCoulombAbboSloan/oedometer.prj
+++ b/Tests/Data/Mechanics/MohrCoulombAbboSloan/oedometer.prj
@@ -32,10 +32,10 @@
                 <process_variable>displacement</process_variable>
             </process_variables>
             <secondary_variables>
-                <secondary_variable type="static" internal_name="ElasticStrain" output_name="ElasticStrain"/>
-                <secondary_variable type="static" internal_name="EquivalentPlasticStrain" output_name="EquivalentPlasticStrain"/>
-                <secondary_variable type="static" internal_name="sigma" output_name="sigma"/>
-                <secondary_variable type="static" internal_name="epsilon" output_name="epsilon"/>
+                <secondary_variable internal_name="ElasticStrain" output_name="ElasticStrain"/>
+                <secondary_variable internal_name="EquivalentPlasticStrain" output_name="EquivalentPlasticStrain"/>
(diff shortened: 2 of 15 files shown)


MaterialLib/SolidModels/MFront/MFront.cpp
```diff
@@ -11,6 +11,8 @@
 
 #include <MGIS/Behaviour/Integrate.hxx>
 
+#include "NumLib/Exceptions.h"
```
wenqing: It might be good to move `Exceptions.h` to `BaseLib` or to ProcessLib/Utils by the name of `AssemblyException`.

MaterialLib/SolidModels/MFront/MFront.cpp
```diff
@@ -313,7 +315,8 @@ MFront<DisplacementDim>::integrateStress(
     auto const status = mgis::behaviour::integrate(v, _behaviour);
     if (status != 1)
     {
-        OGS_FATAL("Integration failed with status %i.", status);
+        throw NumLib::AssemblyException("Integration failed with status" +
```
wenqing: Suggestion
`Integration  failed with ...` --> `MFront: Integration  failed with ...`

MaterialLib/SolidModels/MFront/MFront.cpp
endJunction: `BaseLib` would be too much exposure, because these (that) exception(s) are used only on `NumLib` level.
`ProcessLib/Utils` would not work, because of the include in `MaterialLib`, causing a circular dependency.

MaterialLib/SolidModels/MFront/MFront.cpp
endJunction: Good point. Will improve.

MaterialLib/SolidModels/MFront/MFront.cpp
wenqing: I see. Then just keep it as it is.

endJunction:
Test is added (oedometer test) but due to the nature of the output file name format the last time step has a different number.
Adding formatting possibility for the output file name would solve this problem....
Other ideas maybe?

Tests/Data/Mechanics/MohrCoulombAbboSloan/oedometer_fixed_timestepping.prj
```diff
@@ -0,0 +1,207 @@
...
+                </material_properties>
+            </constitutive_relation>
+            <solid_density>rho_sr</solid_density>
+            <specific_body_force>0 0</specific_body_force>
+            <reference_temperature>293.15</reference_temperature>
+            <process_variables>
+                <process_variable>displacement</process_variable>
+            </process_variables>
+            <secondary_variables>
+                <secondary_variable type="static" internal_name="ElasticStrain" output_name="ElasticStrain"/>
```
TomFischer:
```suggestion
                <secondary_variable internal_name="ElasticStrain" output_name="ElasticStrain"/>
```

Tests/Data/Mechanics/MohrCoulombAbboSloan/oedometer.prj
```diff
@@ -0,0 +1,277 @@
...
+                </material_properties>
+            </constitutive_relation>
+            <solid_density>rho_sr</solid_density>
+            <specific_body_force>0 0</specific_body_force>
+            <reference_temperature>293.15</reference_temperature>
+            <process_variables>
+                <process_variable>displacement</process_variable>
+            </process_variables>
+            <secondary_variables>
+                <secondary_variable type="static" internal_name="ElasticStrain" output_name="ElasticStrain"/>
```
TomFischer:
```suggestion
                <secondary_variable internal_name="ElasticStrain" output_name="ElasticStrain"/>
```

PR #2744, opened 2019-12-06
[PL] Add option to skip a staggered process.

endJunction:
Adding a new optional configuration parameter, skip_process_computation.
The value is used in staggered scheme and if set true, the particular
process computation is skipped throughout the simulation and the initial
value remains.


TomFischer:
Please, can you give a use case for skipping the computation of a process.

PR #2748, opened 2019-12-11
Convergence improvements

TomFischer:

Removes the Boussinesq approximation from the implementation of the hydraulic process.


ProcessLib/HT/HTProcess.cpp
```diff
@@ -64,6 +62,8 @@ void HTProcess::initializeConcreteProcess(
 
     if (_use_monolithic_scheme)
     {
+        checkMPLProperties(mesh, _process_data);
```
endJunction: Why only for monolithic scheme?

TomFischer: The staggered scheme uses other governing equations. For instance, the staggered scheme needs the specification of the storage.

PR #2755, opened 2019-12-30
SerialExcutor] Fixed another bug in the loop for the element deactiva…

wenqing:
As titled.


bilke:
@wenqing It looks like this PR introduced two new clang-tidy warnings:

https://jenkins.opengeosys.org/job/ufz/job/ogs/job/master/1337/clang-tidy/new/

endJunction:
Is fixed in #2754.

PR #2761, opened 2020-01-08
Serial Eve module with Python

### LINKED ISSUE #2759: Include python-bindings in ogs module on eve cluster?
It would be nice if the ogs module on the eve cluster is able to use python boundary conditions and source terms.
Hence, I would highly approve, if the module loaded with

`module load ogs`

would support python bc and source terms.
@bilke

bilke:
Closes #2759.


bilke:
Note to self: I think Python needs to be compiled with `--enable-shared` which is not the case for `python/3`-module.

bilke:
Closing this as #2765 is implemented (as a workaround).

PR #2767, opened 2020-01-20
[CMake] Added setup of CMAKE_BUILD_RPATH.

bilke:
Should fix runtime issues with mfront dylibs


endJunction:
@nagelt Can you (let it) test whether the `LD_LIBRARY_PATH` still would be required after this change?

nagelt:
Worked smoothly on my machine without any special tricks. Did a standard Release build without any changes (i.e. BUILD_SHARED_LIBS was OFF) and could run my MFront tests out of the box.

Verdict: approved ;-)

PR #2772, opened 2020-01-22
refactoring the MPL verification of HT process

Scinopode:
A more elegant way (IMHO) to verify the correct configuration of material properties. No new features, and in this process there is actually the same number of code lines, but it will be very useful in complex processes such as th2m where 18+ properties have to be checked.

### DIFF
--- a/ProcessLib/HT/HTProcess.cpp
+++ b/ProcessLib/HT/HTProcess.cpp
@@ -291,21 +291,19 @@ void checkMPLProperties(MeshLib::Mesh const& mesh,
 {
     DBUG("Check the media properties of HT process ...");
 
-    std::vector<MaterialPropertyLib::PropertyType> const requiredPropertyMedium =
-        {MaterialPropertyLib::PropertyType::porosity,
-         MaterialPropertyLib::PropertyType::permeability};
+    std::array const requiredPropertyMedium = {
+        MaterialPropertyLib::PropertyType::porosity,
+        MaterialPropertyLib::PropertyType::permeability};
 
-    std::vector<MaterialPropertyLib::PropertyType> const
-        requiredPropertyLiquidPhase = {
-            MaterialPropertyLib::PropertyType::viscosity,
-            MaterialPropertyLib::PropertyType::density,
-            MaterialPropertyLib::PropertyType::specific_heat_capacity};
+    std::array const requiredPropertyLiquidPhase = {
+        MaterialPropertyLib::PropertyType::viscosity,
+        MaterialPropertyLib::PropertyType::density,
+        MaterialPropertyLib::PropertyType::specific_heat_capacity};
 
-    std::vector<MaterialPropertyLib::PropertyType> const
-        requiredPropertySolidPhase = {
-            MaterialPropertyLib::PropertyType::specific_heat_capacity,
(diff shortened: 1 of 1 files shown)


ProcessLib/HT/HTProcess.cpp
```diff
@@ -291,66 +290,66 @@ void checkMPLProperties(MeshLib::Mesh const& mesh,
                         HTProcessData const& process_data)
 {
     DBUG("Check the media properties of HT process ...");
+
+    std::vector<MaterialPropertyLib::PropertyType> const reqiredPropertyMedium =
```
TomFischer:
```suggestion
    std::vector<MaterialPropertyLib::PropertyType> const requiredPropertyMedium =
```

ProcessLib/HT/HTProcess.cpp
```diff
@@ -291,66 +290,66 @@ void checkMPLProperties(MeshLib::Mesh const& mesh,
...
+        {MaterialPropertyLib::PropertyType::porosity,
+         MaterialPropertyLib::PropertyType::permeability};
+
+    std::vector<MaterialPropertyLib::PropertyType> const
+        reqiredPropertyLiquidPhase = {
```
TomFischer:
```suggestion
        requiredPropertyLiquidPhase = {
```

ProcessLib/HT/HTProcess.cpp
```diff
@@ -291,66 +290,66 @@ void checkMPLProperties(MeshLib::Mesh const& mesh,
...
+            MaterialPropertyLib::PropertyType::viscosity,
+            MaterialPropertyLib::PropertyType::density,
+            MaterialPropertyLib::PropertyType::specific_heat_capacity};
+
+    std::vector<MaterialPropertyLib::PropertyType> const
+        reqiredPropertySolidPhase = {
```
TomFischer:
```suggestion
        requiredPropertySolidPhase = {
```

ProcessLib/HT/HTProcess.cpp
```diff
@@ -291,66 +290,66 @@ void checkMPLProperties(MeshLib::Mesh const& mesh,
...
+        {MaterialPropertyLib::PropertyType::porosity,
+         MaterialPropertyLib::PropertyType::permeability};
+
+    std::vector<MaterialPropertyLib::PropertyType> const
+        reqiredPropertyLiquidPhase = {
+            MaterialPropertyLib::PropertyType::viscosity,
+            MaterialPropertyLib::PropertyType::density,
+            MaterialPropertyLib::PropertyType::specific_heat_capacity};
+
+    std::vector<MaterialPropertyLib::PropertyType> const
```
endJunction: An `array` would be better.

Scinopode: I thought the same, but then decided for a vector, since the array initialization requires not just the components (density, viscosity,...) but also the array size, which is somehow redundant. Using a vector, one just have to fill in the required properties, letting the compiler count the size.

Scinopode: is a const vector that different from an array?

endJunction: vector allocates memory dynamically, but array's size is known at compile time.

PR #2773, opened 2020-01-23
[HM] MPL Linkage

FZill:
Added Linkage to MPL with only porosity and viscosity being included for now.

The remaining parameters will follow in future PRs.
This is a preliminary step for properly introducing a new permeability model for HM
Also, the specific distinction of FluidTypes in HM will be removed and instead make use of MPL

Here is the Python Script I used for updating the benchmark files (one had to be adjusted manually):
[ConvertToMPL.txt](https://github.com/ufz/ogs/files/4103434/ConvertToMPL.txt)

### DIFF
--- a/MaterialLib/MPL/Phase.h
+++ b/MaterialLib/MPL/Phase.h
@@ -52,6 +52,20 @@ public:
 
     std::string name() const;
 
+    template <size_t N>
+    void checkRequiredProperties(
+        std::array<PropertyType, N> const& required_properties) const
+    {
+        for (auto const& p : required_properties)
+        {
+            if (!hasProperty(p))
+            {
+                OGS_FATAL("The property '%s' is missing in the %s phase.",
+                          property_enum_to_string[p].c_str(), name().c_str());
+            }
+        }
+    }
+
 private:
     std::vector<std::unique_ptr<Component>> const _components;
 
--- a/ProcessLib/HydroMechanics/CreateHydroMechanicsProcess.cpp
+++ b/ProcessLib/HydroMechanics/CreateHydroMechanicsProcess.cpp
@@ -12,6 +12,9 @@
 
 #include <cassert>
 
+#include "MaterialLib/MPL/CreateMaterialSpatialDistributionMap.h"
+#include "MaterialLib/MPL/MaterialSpatialDistributionMap.h"
+
 #include "MaterialLib/SolidModels/CreateConstitutiveRelation.h"
 #include "MaterialLib/SolidModels/MechanicsBase.h"
 #include "ParameterLib/Utils.h"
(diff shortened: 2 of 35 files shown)


ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
```diff
@@ -477,6 +512,11 @@ void HydroMechanicsLocalAssembler<ShapeFunctionDisplacement,
     ParameterLib::SpatialPosition x_position;
     x_position.setElementID(_element.getID());
 
+    auto const& medium = _process_data.media_map->getMedium(_element.getID());
+    auto const& gas_phase = medium->phase("Gas");
```
endJunction: Warning message says the variable is unused.

FZill:
@TomFischer You're right, but I believe this applies to all processes which use MPL. I would therefore implement the checking function in MPL which can be called when a process is created with the necessary parameters as an argument - in a later PR if that's ok.

FZill:
@TomFischer ok, I've got it finished and am adding it to this PR

endJunction:
:fast_forward:  when green.

ProcessLib/HydroMechanics/CreateHydroMechanicsProcess.cpp
```diff
@@ -164,6 +165,16 @@ std::unique_ptr<Process> createHydroMechanicsProcess(
     auto media_map =
         MaterialPropertyLib::createMaterialSpatialDistributionMap(media, mesh);
 
+    std::vector<MaterialPropertyLib::PropertyType> requiredGasParams{
```
TomFischer:
```suggestion
    std::vector<MaterialPropertyLib::PropertyType> const requiredGasParams{
```

ProcessLib/HydroMechanics/CreateHydroMechanicsProcess.cpp
```diff
@@ -164,6 +165,16 @@ std::unique_ptr<Process> createHydroMechanicsProcess(
...
+        MaterialPropertyLib::viscosity};
+    std::vector<MaterialPropertyLib::PropertyType> requiredSolidParams{
```
TomFischer:
```suggestion
    std::vector<MaterialPropertyLib::PropertyType> const requiredSolidParams{
```

TomFischer:
@FZill Thank you for the test.

ProcessLib/HydroMechanics/CreateHydroMechanicsProcess.cpp
```diff
@@ -165,9 +165,9 @@ std::unique_ptr<Process> createHydroMechanicsProcess(
     auto media_map =
         MaterialPropertyLib::createMaterialSpatialDistributionMap(media, mesh);
 
-    std::vector<MaterialPropertyLib::PropertyType> requiredGasParams{
+    std::vector<MaterialPropertyLib::PropertyType> const requiredGasParams{
```
endJunction: sorry to open discussion agai. In [#2772](https://github.com/ufz/ogs/pull/2772/files#diff-04eb6b0a4a7982ff1ca36675aa6f79abR293-R296) we have decided to use `std::array` for [several](https://github.com/ufz/ogs/pull/2772#pullrequestreview-346439926) reasons.

FZill: ok, I see the reasons. Going to fix it

endJunction: Thanks!

MaterialLib/MPL/Phase.h
```diff
@@ -52,8 +52,19 @@ class Phase final
 
     std::string name() const;
 
-    void checkRequiredParameters(
-        const std::vector<PropertyType> &properties) const;
+    template <size_t N>
+    void checkRequiredProperties(
+        std::array<PropertyType, N> const& required_properties) const
```
endJunction: FYI; Other PR, etc.
You can say
```c++
template <typename Container>
void checkRequiredProperties(Container const& properties) const
{
    for (auto const& p : properties) { ... }
}
```

PR #2774, opened 2020-01-23
Orthotropy in MFront solid material models

endJunction:
Orthotropy in MFront solid material models 2D and 3D cases.
Tested against OGS orthotropic elasticity implementation.


MaterialLib/SolidModels/MFront/MFront.cpp
```diff
@@ -293,19 +291,44 @@ MFront<DisplacementDim>::integrateStress(
...
+        }
+        return fourthOrderRotationMatrix(
+            _local_coordinate_system->transformation<DisplacementDim>(x));
+    }();
+
+    auto const eps_prev_MFront =
+        OGSToMFront(Q.transpose()
+                        .template topLeftCorner<
+                            KelvinVectorDimensions<DisplacementDim>::value,
+                            KelvinVectorDimensions<DisplacementDim>::value>() *
```
nagelt: is the topLeftCorner stuff needed now that we have a 2D (4x4) variant of Q?

MaterialLib/SolidModels/MFront/MFront.cpp
```diff
@@ -324,7 +347,10 @@ MFront<DisplacementDim>::integrateStress(
     {
         sigma[i] = behaviour_data.s1.thermodynamic_forces[i];
     }
-    sigma = MFrontToOGS(sigma);
+    sigma = Q.template topLeftCorner<
+                KelvinVectorDimensions<DisplacementDim>::value,
+                KelvinVectorDimensions<DisplacementDim>::value>() *
+            MFrontToOGS(sigma);
```
nagelt: We rotate sigma back into OGS coordinate system, so epsilon and sigma will be output in xyz. All MFront internal variables (ElasticStrain for example) will be output in the local coordinate system. in the _behaviour.symmetry != mgis::behaviour::Behaviour::Symmetry::ISOTROPIC case we could output a warning that his is the case and users have to be careful. Having the local coordinate system as an output in ParaView for use in transformations would be convenient (but not necessary).

MaterialLib/SolidModels/MFront/MFront.cpp
endJunction: To be brief, yes.

MaterialLib/SolidModels/MFront/MFront.cpp
endJunction: A todo at the place where the rotation should happen added.

PR #2778, opened 2020-01-28
Using generalized MPL checker in HM and HT processes.

TomFischer:

### DIFF
--- a/ProcessLib/HT/HTProcess.cpp
+++ b/ProcessLib/HT/HTProcess.cpp
@@ -309,46 +309,15 @@ void checkMPLProperties(MeshLib::Mesh const& mesh,
     {
         auto const element_id = element->getID();
 
-        // check if a definition of the porous media exists
         auto const& medium = *process_data.media_map->getMedium(element_id);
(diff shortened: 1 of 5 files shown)


ProcessLib/HT/HTProcess.cpp
```diff
@@ -24,6 +24,42 @@ namespace ProcessLib
...
+        MaterialPropertyLib::PropertyType::viscosity,
+        MaterialPropertyLib::PropertyType::density,
+        MaterialPropertyLib::PropertyType::specific_heat_capacity};
+
+    std::array const requiredPropertySolidPhase = {
+        MaterialPropertyLib::PropertyType::specific_heat_capacity,
+        MaterialPropertyLib::PropertyType::density,
+        MaterialPropertyLib::PropertyType::storage};
+
+    for (auto const& element : mesh.getElements())
```
endJunction: Not sure here. If we write `for (auto e : mesh.getElements())`, `e` has type `Element*` (optionally const qualified). In the current case there is additional reference... Not sure what a compiler will do out of `Element*&`...

PR #2779, opened 2020-01-28
[HM] Final MPL Link

FZill:
Linked fluid_density, permeabiltiy and ref. temperature in HM to MPL.
Fluid compressibiltiy is now properly calculated as drho_dp / rho.
Also filled the argument variables with temperature and pressure, so that MPL can calculate e.g. the IdealGasLaw.
Finally the FluidType class, which was only temporarily used by HM, could be removed.
HM-models which used the ideal gas law now have to state the molar_mass. To get the same results as before it can be calculated via: M = rho_fr * R * T / p with the fluid_density rho_fr, the specific_gas_constant R, the reference_temperature T and the initial_pressure p.

I updated the benchmarks with this script: 
[ConvertToMPL.txt](https://github.com/ufz/ogs/files/4159077/ConvertToMPL.txt)
It should ged rid of the now unnecessary parameters and add the needed properties.


endJunction:
I suggest to move the ideal gas laws in own PR. There are some unit tests and documentation associated. And is a stand alone feature.

MaterialLib/MPL/Properties/CompressibilityIdealGasLaw.cpp
```diff
@@ -0,0 +1,81 @@
...
+    {
+        return 2 / (pressure * pressure * pressure);
+    }
+    if ((primary_variable1 == Variable::temperature) ||
+        (primary_variable2 == Variable::temperature))
+    {
+        return 0;
+    }
+
+    OGS_FATAL(
```
TomFischer: What about mixed second variables derivatives (d2p/(dp dT) or d2p/(dT dp))? Should these combination result in an OGS_FATAL?

MaterialLib/MPL/Properties/CompressibilityIdealGasLaw.h
```diff
@@ -0,0 +1,72 @@
+/**
+ * \copyright
+ * Copyright (c) 2012-2020, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ *  \file
+ *  Created on November 29, 2017, 2:03 PM
```
TomFischer: Probably you can remove this line?

MaterialLib/MPL/Properties/CompressibilityIdealGasLaw.cpp
```diff
@@ -0,0 +1,81 @@
...
+{
+
+PropertyDataType CompressibilityIdealGasLaw::value(VariableArray const& variable_array,
+                                    ParameterLib::SpatialPosition const& pos,
+                                    double const t) const
+{
+    const double pressure = std::get<double>(
+        variable_array[static_cast<int>(Variable::phase_pressure)]);
+
+    return (1 / pressure);
```
wenqing: or
``` 
      // return 1/p  
     return 1.0 / std::get<double>(
        variable_array[static_cast<int>(Variable::phase_pressure)]);
```

FZill:
Since the compressibiltiy of the fluid can be calculated as drho_dp / rho, there was no need for the second IdealGasLaw-stuff. So this PR got a bit smaller and I made the commits more precise.

ProcessLib/HydroMechanics/CreateHydroMechanicsProcess.cpp
```diff
@@ -176,40 +162,11 @@ std::unique_ptr<Process> createHydroMechanicsProcess(
...
-
     HydroMechanicsProcessData<DisplacementDim> process_data{
         materialIDs(mesh),     std::move(media_map),
         std::move(solid_constitutive_relations),
-        initial_stress,        intrinsic_permeability,
-        fluid_density,         specific_body_force,
-        fluid_compressibility, reference_temperature,
-        specific_gas_constant, fluid_type};
+        initial_stress,        specific_body_force,
+        reference_temperature};
```
endJunction: The reference temperature is also available in MPL... https://github.com/ufz/ogs/blob/master/MaterialLib/MPL/PropertyType.h#L64

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
```diff
@@ -215,28 +216,34 @@ void HydroMechanicsLocalAssembler<ShapeFunctionDisplacement,
...
+
+        auto const mu = gas.property(MPL::PropertyType::viscosity)
+                            .template value<double>(vars, x_position, t);
+        auto const rho_fr = gas.property(MPL::PropertyType::density)
+                                .template value<double>(vars, x_position, t);
+        auto const beta_p = gas.property(MPL::PropertyType::density)
+                .template dValue<double>(vars, MPL::Variable::phase_pressure,
+                                         x_position, t) / rho_fr;
+
+        auto const S = gas.hasProperty(MPL::PropertyType::storage)
```
endJunction: let's call it `storage`.

FZill: okay and I noticed it has to be a medium property

ProcessLib/HydroMechanics/CreateHydroMechanicsProcess.cpp
FZill: also add this a medium property?

endJunction: Yes.

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
```diff
@@ -216,6 +214,13 @@ void HydroMechanicsLocalAssembler<ShapeFunctionDisplacement,
         auto& eps = _ip_data[ip].eps;
         auto const& sigma_eff = _ip_data[ip].sigma_eff;
 
+        auto const T_ref =
```
FZill: Has somebody a better idea?
I could make T_ref a required property, but since it is not needed in most of the cases I don't think that makes sense.

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
```diff
@@ -216,6 +214,13 @@ void HydroMechanicsLocalAssembler<ShapeFunctionDisplacement,
...
+            medium->hasProperty(MPL::PropertyType::reference_temperature)
+                ? medium->property(MPL::PropertyType::reference_temperature)
+                      .template value<double>(vars, x_position, t)
+                : std::numeric_limits<double>::quiet_NaN();
+
```
FZill: Does somebody have a better idea?
I could make T_ref a required property, but since it is not needed in most of the cases I don't think that makes sense

TomFischer: Is T_ref independent from the integration point? Then it could be evaluated only once outside of the integration loop.
If T_ref is additionally independent from time it could be evaluated in the constructor and stored in the local assembler. Or if it is even independent from the mesh element? Then, is it possible to move it out of the class and pass it as an argument to the local assembler?

FZill: Since the user could in theory make t_ref spatially dependent (e.g. cavern vertical temperature gradient) it is dependent of the IP, right?
I think with the time dependence it's the same. 
I could move it out of the loop when we assume it is constant per element.
But since it is probably a rare scenario having something else than a constant T_ref in a HM-Process, we maybe can move it further out. But then we could just leave it as a variable in process_data...

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
```diff
@@ -215,28 +214,41 @@ void HydroMechanicsLocalAssembler<ShapeFunctionDisplacement,
...
+        auto const mu = gas.property(MPL::PropertyType::viscosity)
+                            .template value<double>(vars, x_position, t);
+        auto const rho_fr = gas.property(MPL::PropertyType::density)
+                                .template value<double>(vars, x_position, t);
+        auto const beta_p = gas.property(MPL::PropertyType::density)
+                .template dValue<double>(vars, MPL::Variable::phase_pressure,
+                                         x_position, t) / rho_fr;
+
+        auto const storage = medium->hasProperty(MPL::PropertyType::storage)
```
Scinopode: This conditional means that you can either use a constant _storage_ parameter (combined fluid and solid compressibilities along with some poro-mechanical magic and porosity) OR to use the concept of compressibilities in the first place. This is OK, but I would highly recommend to get rid of the _storage_ parameter at all! Actually, there is no cost in doing so, since all of the parameters _storage_ is made of have to be provided anyway: fluid compressibility is given via the derivative of the density parameter, porosity and biot-coefficient are required, and solid phase compressibility can be taken from Youngs modulus (or defined directly). I see no benefit using a composite _storage_ parameter; otherwise, using _storage_ limits you in your choice of your other properties (you would have to reevaluate the value of _storage_ when changing porosity, biot-coeffient, or even the ambient temperature).
Please correct me if I'm wrong or if I missed something.

FZill: No you're right. we have just one test which used the specific storage until I replaced it with the corresponding compressibility but there was somebody which wanted the possibility back to give the storage as a user input and I thought I could now do exactly this. But you're right, it is not beneficial. I will revert this.

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
FZill: Consensus is, to make T_ref a required property

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
FZill: storage term is now calculated as before.
And since the compressibility is now fixed as drho_dp the benchmark InjectionProduction1D was inccompatible with the current implementation. I could recreate the same storage with the correct rho_fr = f(p) but since rho_fr is then not constant it influences the results. Thus I updated the solution files of this test. Since it mainly tests the equality of the monolithic and staggered theme solutions this shouldn't matter much.

Tests/Data/HydroMechanics/Linear/Confined_Compression/cube_1e3.prj
```diff
@@ -59,6 +61,18 @@
                     </properties>
                 </phase>
             </phases>
+            <properties>
+                <property>
+                    <name>reference_temperature</name>
+                    <type>Constant</type>
+                    <value>0</value>
```
endJunction: This is rather cold environment...
Should be mentioned at least in the PR. Better would be something around 20 degree Celsius.

Tests/Data/HydroMechanics/StaggeredScheme/InjectionProduction1D/InjectionProduction1D.prj
```diff
@@ -71,6 +76,18 @@
                     </properties>
                 </phase>
             </phases>
+            <properties>
+                <property>
+                    <name>reference_temperature</name>
+                    <type>Constant</type>
+                    <value>295.15</value>
```
endJunction: It's not 293.15K ?

Tests/Data/HydroMechanics/Verification/hm1_3Dgravity.prj
```diff
@@ -275,71 +340,13 @@
...
-                <value>0.25</value>
-            </index_values>
-            <index_values>
-                <index>3</index>
-                <value>0.25</value>
-            </index_values>
-        </parameter>
-        <parameter>
-            <name>k</name>
-            <type>Group</type>
```
endJunction: I'd keep the Group type parameter. There is no need to repeat all other properties four times, just because one parameter is heterogeneous.

Tests/Data/HydroMechanics/StaggeredScheme/InjectionProduction1D/InjectionProduction1D.prj
```diff
@@ -48,6 +43,16 @@
...
                             <value>1e-3</value>
                         </property>
+                        <property>
+                            <name>density</name>
+                            <type>Linear</type>
+                            <reference_value>1e4</reference_value>
+                                <independent_variable>
+                                    <variable_name>phase_pressure</variable_name>
+                                    <reference_condition>0</reference_condition>
+                                    <slope>3.3332941333333e-5</slope>
```
endJunction: Why is the slope different from the fluid's compressibility?

Tests/Data/HydroMechanics/Verification/hm1_3Dgravity.prj
FZill: But either way I have to mention in every medium either the constant value or the reference to the parameter, so the medium section will be large anyway. Then I thought, it would be best to choose the constant value, to get rid of the large parameter section.

Tests/Data/HydroMechanics/StaggeredScheme/InjectionProduction1D/InjectionProduction1D.prj
FZill: The reference_temp. is in K, so that means 20°C what's wring with this?

Tests/Data/HydroMechanics/Linear/Confined_Compression/cube_1e3.prj
FZill: Okay, then I go with the 293.15 K for all tests

Tests/Data/HydroMechanics/StaggeredScheme/InjectionProduction1D/InjectionProduction1D.prj
FZill: The compressebility of 3.333... before probably was misscalculated from my side but did not seem to influence the result enough to fail. This value now should correctly result in a storage of 1e-5.

FZill:
I updated the benchmarks to have a reasonable value for the reference temperature,
made use of #2786 to shrink down the definition of the 3Dgravity benchmark
and updated the checkRequiredProperties to do check on each element and also added another check to verify if a medium corresponding to the material_ID of the current element exists.
Without the latter, ogs can abort at this position when no media are defined for all materialID's in the mesh.

PR #2780, opened 2020-01-28
[MPL] Compressibility Ideal Gas Law

FZill:
Preceeding the #2779.

Added an Ideal Gas Law for Fluid compressibility.
The Ideal Gas Law which already existed and returned the value for the density is renamed in accordance to the other models with multiple return variables (e.g. RelPermBrooksCorey, SaturationBrooksCorey) -> DensityIdealGasLaw
The new file is named accordingly CompressibilityIdealGasLaw


endJunction:
@Scinopode Could you please review this?

FZill:
I overlooked the simple solution, that the dvalue of the (Density)IdealGasLaw already results in the compressibility, so no need for this PR.

PR #2783, opened 2020-01-29
 [CMake, web] Added version requirements file in json-format.

bilke:
The file is read by CMake (via new a json parser submodule) and
the web site generator and provides info on required clang,
gcc, msvc, conan and git versions. This info will be expanded in future commits.

Goal is to have a single source of truth for versions infos and
maybe other requirements.

json-file-excerpt:

```json
{
  "minimum_version": {
    "gcc": "7.3",
    "clang": "3.5",
    "msvc": {
      "year": "2017",
      "number": "15",
      "toolset": "141"
    }
  ...
  }
}
```

On CMake you access the info via `ogs.`-prefixed variables:

    ${ogs.minimum_version.gcc}

In Hugo you access the info via the new dataFile-shortcode:

    {{< dataFile "versions.minimum_version.gcc" >}}


scripts/cmake/SubmoduleSetup.cmake
```diff
@@ -14,7 +14,7 @@ set(REQUIRED_SUBMODULES
     ThirdParty/exprtk
     ThirdParty/googletest
     ThirdParty/iphreeqc/src
-    ThirdParty/jedbrown-cmake-modules
```
TomFischer: Isn't the 'jedbrown-cmake-modules' submodule necessary anymore?

bilke:
@endJunction The only file required is now in `scripts/cmake/jedbrown/FindPETSc.cmake`. I removed the submodule as it contains a `FindGit.cmake` which overrides the (better) default `FindGit.cmake` from CMake itself.

PR #2786, opened 2020-02-03
[MPL] Shared Media

FZill:
Changed the media map definition to shared_pointer, so that one medium definition can be assigned to multiple material ID's. This can be useful when a mesh has multiple material ID's defined, but they have the same properties. Or if they only differ in one property, this can be easier written by refering this property to a Group-Parameter.
PR #2779 will contain a benchmark benefitting from this.
This is how a .prj-file would look like, which benefits from it: 
[hm1_3Dgravity.prj.txt](https://github.com/ufz/ogs/files/4152636/hm1_3Dgravity.prj.txt)

### DIFF
--- a/Applications/ApplicationsLib/ProjectData.cpp
+++ b/Applications/ApplicationsLib/ProjectData.cpp
@@ -447,8 +447,8 @@ void ProjectData::parseMedia(
          //! \ogs_file_param{prj__media__medium}
          media_config->getConfigSubtreeList("medium"))
     {
-        //! \ogs_file_attr{prj__media__medium__id}
         auto material_id_string =
+            //! \ogs_file_attr{prj__media__medium__id}
             medium_config.getConfigAttribute<std::string>("id", "0");
         material_id_string.erase(
             remove_if(begin(material_id_string), end(material_id_string),
--- a/Applications/ApplicationsLib/ProjectData.cpp
+++ b/Applications/ApplicationsLib/ProjectData.cpp
@@ -14,11 +14,9 @@
 #include "ProjectData.h"
 
 #include <algorithm>
-#include <boost/algorithm/string.hpp>
-#include <boost/algorithm/cxx11/all_of.hpp>
-#include <set>
-
+#include <cctype>
 #include <logog/include/logog.hpp>
+#include <set>
 
 #ifdef OGS_USE_PYTHON
 #include <pybind11/eval.h>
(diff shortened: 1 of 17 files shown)


Applications/ApplicationsLib/ProjectData.cpp
```diff
@@ -448,19 +449,34 @@ void ProjectData::parseMedia(
...
     {
         //! \ogs_file_attr{prj__media__medium__id}
-        auto const material_id = medium_config.getConfigAttribute<int>("id", 0);
+        auto const material_id_string =
+            medium_config.getConfigAttribute<std::string>("id", "0");
+        std::vector<std::string> material_ids;
+        boost::split(material_ids, material_id_string, boost::is_any_of(","));
 
-        if (_media.find(material_id) != _media.end())
+        for (auto const& m_id : material_ids)
```
endJunction: Not sure here: In the for-loop, the `m_id`, doesn't it need white space stripping?

Applications/ApplicationsLib/ProjectData.cpp
```diff
@@ -448,19 +449,34 @@ void ProjectData::parseMedia(
...
         {
-            OGS_FATAL(
-                "Multiple media were specified for the same material id %d. "
-                "Keep in mind, that if no material id is specified, it is "
-                "assumed to be 0 by default.",
-                material_id);
-        }
+            if (m_id.find_first_not_of("0123456789") != std::string::npos)
```
endJunction: Minor; maybe `boost::algorithm::is_digit` is more idiomatic.

Applications/ApplicationsLib/ProjectData.cpp
```diff
@@ -448,19 +449,34 @@ void ProjectData::parseMedia(
...
+            {
+                OGS_FATAL(
+                    "Could not parse material ID's from '%s'. Please separate "
+                    "multiple material ID's by comma only.",
+                    material_id_string.c_str());
```
endJunction: I like the message. It would be nice to know which character was read, but is not a digit.

TomFischer:
> This can be useful when a mesh has multiple material ID's defined, but they have the same properties.

It is possible to modify the material IDs in the mesh such that it match the medium definitions. Then code changes aren't necessary. Do you have a special case in mind? Please, can you give a project file snippet how it should look like?

Applications/ApplicationsLib/ProjectData.cpp
FZill: At this state, if the string contains a whitespace it would cause the OGS_FATAL.
I can simply remove all whitespaces from the entire string before the loop.

Applications/ApplicationsLib/ProjectData.cpp
FZill: changed it to the boost solution

Applications/ApplicationsLib/ProjectData.cpp
FZill: added it, but used the find_first_not_of function. I wasn't able to devise an equally easy boost solution.

FZill:
@TomFischer I added the example .prj-file in the description. This code change should exactly prevent the user from needing to change the mesh. If you have a mesh with multiple materials defined in it, but you want to simulate it with only one material for the whole domain, it seems easier just to list the material ID's, than to change the mesh. Or you can - like in the example - define multiple media which differ in only few properties with the group-type parameter without the need to have 4 almost the same media definitions.

PR #2787, opened 2020-02-05
Use MPL in the RichardsMechanics project.

endJunction:
Similar to the #2779 , but with RichardsMechanics.
~Includes #2788 for the van Genuchten saturation and relative permeability models.~

4. [x] [Conversion script](https://github.com/ufz/ogs/files/4176827/convertToMPL.py.txt) for the project files.


Tests/Data/RichardsMechanics/RichardsFlow_2d_small.prj
```diff
@@ -13,67 +13,103 @@
...
         </process>
     </processes>
+    <media>
+        <medium>
+            <phases>
+                <phase>
+                    <type>AqueousLiquid</type>
+                    <properties>
+                        <property>
+                            <name>bulk_modulus</name>
```
Scinopode: The only thing I'd like to discuss here is wether we agree to use either bulk moduli _K_ or compressibilities _beta=1/K_ for fluids and/or for solid phases. Personally, I like the compressibility approach, but this is just a personal preference. But, I am worried that it may look inconsistent (in fact, it is) for a user, since some processes require _K_, some require _beta_. Like many other things in MPL, we should discuss that in the near future.

endJunction: Let's collect these things for the upcoming community meeting: https://github.com/ufz/ogs/wiki/Community-meeting-2019

PR #2789, opened 2020-02-07
Fixed a bug in the element deactivation approach

wenqing:
The bug: `DirichletBoundaryConditionWithinTimeInterval::_nodes_in_bc_mesh` stored the bulk nodes not the nodes of BC mesh before  this correction. Now it stores the nodes from BC mesh.  
(**Comment**: it would be better to avoid copying nodes and elements to sub meshes for BC from the bulk mesh )  

2. [x] A 3D Test covering the bug fixing was added.
3. [ ] Any new feature or behavior change was documented? TODO in another PR.

element_deactivation_3D


wenqing:
Commit 3253bd0  [Changed the search method] just changed the search method to find the inactive nodes. The previous implementation of  searching  inactive nodes work fines but with an additional memory of an array, the present one removes the additional array with a tiny increment in computation.

ProcessLib/DeactivatedSubdomain.cpp
```diff
@@ -89,49 +89,39 @@ std::unique_ptr<DeactivatedSubdomain const> createDeactivatedSubdomain(
...
+            if (std::find_if(
+                    connected_elements.begin(),
+                    connected_elements.end(),
+                    [&](auto const* const connected_elem) -> bool {
+                        return ids != (*material_ids)[connected_elem->getID()];
+                    }) != connected_elements.end())
+            {
+                continue;
+            }
+            else
```
endJunction: no need for an `else` branch after `return`, `continue`, `break`...

ProcessLib/DeactivatedSubdomain.cpp
```diff
@@ -89,55 +88,78 @@ std::unique_ptr<DeactivatedSubdomain const> createDeactivatedSubdomain(
...
+            {
+                continue;
+            }
+            deactivated_nodes.push_back(
+                const_cast<MeshLib::Node*>(nodes_in_bc_mesh[i]));
+
+            deactivated_bulk_node_ids.erase(
+                deactivated_bulk_node_ids.begin() +
+                std::distance(deactivated_bulk_node_ids.begin(),
+                              found_iterator));
```
endJunction: Isn't it same as
```c++
deactivated_bulk_node_ids.erase(found_iterator);
```
?

ProcessLib/DeactivatedSubdomain.cpp
wenqing: Forgot to remove that after removed lines except for `continue;` in if block. Removed the else by amending the commit.

ProcessLib/DeactivatedSubdomain.cpp
wenqing: You are right. Changed by amending the commit.

PR #2800, opened 2020-02-12
Install Conan settings from source

bilke:
Install conan settings from source.

https://docs.conan.io/en/latest/reference/commands/consumer/config.html#conan-config-install

This should fix errors with invalid compiler version settings. Downside is that we have to update the settings file from time to time.


bilke:
@endJunction Seems to work: https://jenkins.opengeosys.org/job/ufz/job/ogs/job/PR-2800/2/console

PR #2802, opened 2020-02-14
Fracture model rename

nagelt:
This simply corrects the naming of the fracture model 'MohrCoulomb' to 'Coulomb'.


Tests/MaterialLib/MFront.cpp
```diff
@@ -80,13 +80,12 @@ struct ElasticBehaviour
 };
 
 template <int Dim>
-struct MohrCoulombAbboSloanBehaviour
```
endJunction: This should not be changed...

TomFischer: If this issue is discussed/fixed and the test are green IMHO we can merge the PR.

nagelt: @endJunction @TomFischer Fixed.

PR #2804, opened 2020-02-16
RichardsMechanics; mass balance based porosity

endJunction:
#2801, and #2803 are included in this PR.

Adds variable porosity model to MPL used in the RichardsMechanics process.

This type of material models needs an initial (or reference) value to be available in the MPL and FEM. In this PR proposition is to add `initialValue(x, t)` function to MPL properties.


MaterialLib/MPL/Properties/PorosityFromMassBalance.h
```diff
@@ -0,0 +1,70 @@
...
+        }
+        else
+        {
+            OGS_FATAL(
+                "The property 'PorosityFromMassBalance' is "
+                "implemented on the 'phase' scales only.");
+        }
+    }
+
+    PropertyDataType initialValue(ParameterLib::SpatialPosition const& pos,
```
TomFischer: Why the `override` here?

ProcessLib/RichardsMechanics/RichardsMechanicsFEM-impl.h
```diff
@@ -194,17 +227,21 @@ void RichardsMechanicsLocalAssembler<
...
             liquid_phase.property(MPL::PropertyType::bulk_modulus)
                 .template value<double>(variables, x_position, t, dt);
-        auto const porosity =
-            solid_phase.property(MPL::PropertyType::porosity)
-                .template value<double>(variables, x_position, t, dt);
+
+        // Use previous time step porosity for porosity update, ...
+        variables[static_cast<int>(MPL::Variable::porosity)] =
+            _ip_data[ip].porosity_prev;
+        auto& porosity = _ip_data[ip].porosity;
```
TomFischer: Here you define the porosity and assign a value. In the next line you assign a different value. Why not like this:
```suggestion
        auto& porosity = solid_phase.property(MPL::PropertyType::porosity)
                       .template value<double>(variables, x_position, t, dt);
```

ProcessLib/RichardsMechanics/RichardsMechanicsFEM-impl.h
```diff
@@ -409,6 +450,10 @@ void RichardsMechanicsLocalAssembler<ShapeFunctionDisplacement,
         variables[static_cast<int>(MPL::Variable::capillary_pressure)] =
             p_cap_ip;
         variables[static_cast<int>(MPL::Variable::phase_pressure)] = -p_cap_ip;
+        variables[static_cast<int>(MPL::Variable::phase_pressure_rate)] =
+            N_p.dot(p_L_dot);
+        variables[static_cast<int>(MPL::Variable::volumetric_strain_rate)]
+            .emplace<double>(identity2.transpose() * B * u_dot);
```
nagelt: Looks easy to modify for inclusion of swelling/thermal strain. Nice.

MaterialLib/MPL/Properties/PorosityFromMassBalance.cpp
```diff
@@ -0,0 +1,55 @@
...
+        variable_array[static_cast<int>(Variable::volumetric_strain_rate)]);
+
+    double const p_dot = std::get<double>(
+        variable_array[static_cast<int>(Variable::phase_pressure_rate)]);
+
+    double const phi = std::get<double>(
+        variable_array[static_cast<int>(Variable::porosity)]);
+
+    double const w = dt * (e_dot + p_dot / K_SR);
+    return (phi + alpha_b * w) / (1 + w);
```
nagelt: I changed docu to implicit update, too

endJunction: So, this is correct one, isn't it?

ProcessLib/RichardsMechanics/RichardsMechanicsFEM-impl.h
endJunction: Because the `_ip_data[ip].porosity` wouldn't be updated.
`auto& porosity` is just a reference.

TomFischer: Sorry, I overlooked the reference.

endJunction:
@Scinopode Could you have a look on the prj files regarding the MPL? And also the `initialValue()`. We have already discussed most of the details, I think, so there should be no surprises.

PR #2805, opened 2020-02-17
Check MPL properties in ComponentTransport process.

TomFischer:

### DIFF
--- a/ProcessLib/ComponentTransport/ComponentTransportProcess.cpp
+++ b/ProcessLib/ComponentTransport/ComponentTransportProcess.cpp
@@ -20,6 +20,61 @@ namespace ProcessLib
 {
 namespace ComponentTransport
 {
+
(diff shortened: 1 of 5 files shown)


ProcessLib/ComponentTransport/ComponentTransportProcess.cpp
```diff
@@ -20,6 +20,61 @@ namespace ProcessLib
...
+        checkRequiredProperties(medium, required_properties_medium);
+
+        // check if liquid phase definition and the corresponding properties
+        // exists
+        auto const& liquid_phase = medium.phase("AqueousLiquid");
+        checkRequiredProperties(liquid_phase, required_properties_liquid_phase);
+
+        // check if components and the corresponding properties exists
+        for (std::size_t component_id = 0;
+             component_id < liquid_phase.numberOfComponents();
```
endJunction: minor: storing the number of components outside of the for-loop would save few instructions...

PR #2808, opened 2020-02-18
Updated submodule ThirdParty/cmake-modules

bilke:


TomFischer:
Two questions:
 - Could the jedbrown subfolder be a submodule?
 - I assume that the files in jedbrown aren't written by us. What license are these files under?

bilke:
It was a submodule, removed in https://github.com/ufz/ogs/pull/2783/commits/beaa7a275752fc3728cd25013b7d441c5209ea36. It is just the FindPETSc.cmake which is remaining.. I want to get rid of it too in the future.

In the meantime I will add the [license](https://github.com/jedbrown/cmake-modules/blob/master/LICENSE)

PR #2809, opened 2020-02-18
Fix bug in constraint Dirichlet-type boundary condition implementation

TomFischer:
@Thomas-TK  discovered a  bug. The constraint Dirichlet boundary condition was not applied since the location, needed for the global index calculation, wasn't constructed correctly.

Here is a [comparison of the 'old' (left side) and the 'new' (right side) behavior](https://user-images.githubusercontent.com/800065/74718221-0ee5b900-5232-11ea-86ff-7762b454e7e8.png) after one time step.


ProcessLib/HT/Tests.cmake
```diff
@@ -365,7 +365,7 @@ AddTest(
     TESTER vtkdiff
     REQUIREMENTS NOT OGS_USE_MPI
     DIFF_DATA
-    GLOB constraint_bc_1e3_pcs_0_ts_*.vtu T T 1e-14 1e-13
+    GLOB constraint_bc_1e3_pcs_0_ts_*.vtu T T 1e-4 1e-4
```
endJunction: Maybe new reference is due?

TomFischer: I haven't any idea why the tolerances have to be so "relaxed". It is only frontend2 that is complaining. Other configurations (docker, windows, mac, my home computer) pass with much tighter tolerances.
I have even to relax the tolerances for the sixth time step even more.

ProcessLib/BoundaryCondition/ConstraintDirichletBoundaryCondition.cpp
```diff
@@ -219,7 +219,8 @@ void ConstraintDirichletBoundaryCondition::getEssentialBCValues(
     bc_values.ids.emplace_back(current_id);
     bc_values.values.emplace_back(sum / cnt);
 
-    DBUG("Found %d dirichlet values.", bc_values.ids.size());
+    DBUG("Found %d constraint dirichlet boundary condition values.",
+         bc_values.ids.size());
     for (unsigned i = 0; i < bc_values.ids.size(); ++i)
     {
         DBUG("\tid: %d, value: %e", bc_values.ids[i], bc_values.values[i]);
```
endJunction: This and the above DBUG messages are quite noisy...

ProcessLib/HT/Tests.cmake
endJunction: I get for `p` at time step 1 difference of "abs maximum norm = [5.427110185059905e+04]"

endJunction:
At the end of (successful) simulation there is a warning:
```
warning: There are still some matrices and vectors in use. This might be an indicator of a possible waste of memory.
```

ProcessLib/HT/Tests.cmake
endJunction: The debug and release builds here yield different results. The above difference is for my release *vs.* the reference files.

TomFischer: I can not reproduce the divergence of the temperature values from the reference files. I compiled a (clang) debug build as well as a (clang) release build of ogs on my computer. In both versions the tolerances for the temperature was set to almost the machine precision:
```
GLOB constraint_bc_1e3_pcs_0_ts_*.vtu T T 1e-14 1e-13
```
In both cases (debug/release) the tests pass.

ProcessLib/BoundaryCondition/ConstraintDirichletBoundaryCondition.cpp
TomFischer: Removed the for loop. The debug message above (lines 222/223) is only outputted once per application of the constraint dirichlet bc. The output can be helpful if there are issues setting the boundary condition. I would keep this debug message.

ProcessLib/BoundaryCondition/ConstraintDirichletBoundaryCondition.cpp
```diff
@@ -139,8 +139,10 @@ void ConstraintDirichletBoundaryCondition::getEssentialBCValues(
     std::vector<std::pair<GlobalIndexType, double>> tmp_bc_values;
 
     auto isFlux = [&](const std::size_t element_id) {
-        return _lower ? _flux_values[element_id] < _constraint_threshold
-                      : _flux_values[element_id] > _constraint_threshold;
+        return _lower ? _constraint_threshold - _flux_values[element_id] >=
+                            std::numeric_limits<double>::epsilon()
+                      : _flux_values[element_id] - _constraint_threshold >=
+                            std::numeric_limits<double>::epsilon();
```
endJunction: Not sure, if this is a good idea...
I see how it solves the current ctest problem, though.
If this is what the boundary condition is required to do, OK, but I have a feeling, that this is not right way.

Documentation update might be needed for the changed behaviour.

TomFischer: The implementation requires that the user is familiar with correct (physical) setting the boundary condition.

Updated documentation.

TomFischer:
Only the Jenkins windows build is failing - but this seems to be a general problem at the moment and has probably nothing to do with the changes in this PR.

PR #2810, opened 2020-02-18
Updated eve modules

bilke:
Newer Boost module does not depend on deprecated modules.


bilke:
`drop`-commits will be deleted before merge.. I just disable some Jenkins checks to speed-up testing..

TomFischer:
> `drop`-commits will be deleted before merge.. I just disable some Jenkins checks to speed-up testing..

Thanks for explanation.

PR #2811, opened 2020-02-18
Update benchmark documentation.

TomFischer:


web/content/docs/benchmarks/elliptic/elliptic-dirichlet.pandoc
```diff
@@ -102,21 +105,4 @@ A major part of the output was produced by the linear equation solver (LIS in th
...
-0.2
-...
--0.6
--0.8
--1
-
-
-```
-(The next releases shall simplify this procedure.)
+The result can be visualized with Paraview.
```
endJunction: Reads easier in my opinion, if this is moved before the image:
```
The result can be visualized with Paraview:
 __
|  |
 --
```

TomFischer: Move text before image.

PR #2822, opened 2020-02-27
Avoided the infinite time loop when a rejected step gives no change in the step size

wenqing:
In the current ogs,  if a rejected step gives no change in the step size, the time loop becomes  infinite. This PR fixes this issue.

### DIFF
--- a/ProcessLib/TimeLoop.cpp
+++ b/ProcessLib/TimeLoop.cpp
@@ -399,6 +399,25 @@ double TimeLoop::computeTimeStepping(const double prev_dt, double& t,
         dt = _end_time - t;
     }
 
+    // Check whether the time stepping is stabilized
+    if (std::fabs(dt - prev_dt) < std::numeric_limits<double>::epsilon())
+    {
+        if (_last_step_rejected)
+        {
+            OGS_FATAL(
+                "The new step size of %g is the same as that of the previous "
+                "rejected time step. \nPlease re-run ogs with a proper "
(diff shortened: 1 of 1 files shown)


ProcessLib/TimeLoop.cpp
```diff
@@ -399,6 +399,25 @@ double TimeLoop::computeTimeStepping(const double prev_dt, double& t,
...
+            OGS_FATAL(
+                "The new step size of %g is the same as that of the previous "
+                "rejected time step. \nPlease re-run ogs with a proper "
+                "adjustment in the numerical settings, \ne.g those for "
+                "time stepper, local or global non-linear solver.",
+                dt);
+        }
+        else
+        {
+            WARN("The time stepping is stabilized with the step size of %g.\n",
```
endJunction: Maybe debug, but not a warning...

endJunction: I'd keep the changes to the output out of this PR. Some people might rely on the current output format...

wenqing: @endJunction WARN has been replaced with DBUG.

wenqing: @endJunction The last commit was removed.

endJunction:
@tengfeideng Could you check, if possible, whether the fix is working also on your system?

tengfeideng:
I have tested the infinite time loop. It worked in my project

> @tengfeideng Could you check, if possible, whether the fix is working also on your system?

I have tested it and it worked.

PR #2831, opened 2020-03-01
Fixed output times for iteration number based time stepping algorithm

endJunction:
2. [x] Tests covering your feature were added? New unit test. There are ctests using this feature.
3. [x] Any new feature or behavior change was documented? Documentation already available for this tag

### DIFF
--- a/NumLib/TimeStepping/Algorithms/IterationNumberBasedTimeStepping.cpp
+++ b/NumLib/TimeStepping/Algorithms/IterationNumberBasedTimeStepping.cpp
@@ -52,9 +52,9 @@ IterationNumberBasedTimeStepping::IterationNumberBasedTimeStepping(
     {
         OGS_FATAL("Vector of iteration numbers must be sorted.");
     }
-    //
-    // Remove possible duplicated elements and sort in descending order.
-    BaseLib::makeVectorUnique(_fixed_output_times, std::greater<double>());
+
+    // Remove possible duplicated elements. Result will be sorted.
+    BaseLib::makeVectorUnique(_fixed_output_times);
 }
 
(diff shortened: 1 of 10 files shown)


NumLib/TimeStepping/Algorithms/IterationNumberBasedTimeStepping.cpp
```diff
@@ -119,6 +126,35 @@ double IterationNumberBasedTimeStepping::getNextTimeStepSize() const
     return dt;
 }
 
+double IterationNumberBasedTimeStepping::checkSpecificTimeReached(const double h_new)
```
wenqing: Since only FixedTimeStepping dose not need ` _fixed_output_times` and `checkSpecificTimeReached`, we can move these to the base class `TimeStepAlgorithm` in future to avoid code duplication. For this PR, it is Ok.

TomFischer: The method changes the `_fixed_output_times` vector but the name doesn't is only checkSomething(). For me it is not obvious what the method does and the there isn't any documentation.

endJunction: My copy-paste laziness ;)
Fixing...

endJunction:
@TomFischer and @wenqing I just did some changes satisfying both requests.
There is only one implementation of the function. The function does not modify anything.

wenqing:
Tested with desaturated.prj with `t<initial_dt` or duplicated times, e.g.
` <fixed_output_times> 15.0 50 50 120 120 </fixed_output_times>` . It works fine!
:fast_forward:

PR #2832, opened 2020-03-01
[MPL] Fix saturation dependent swelling.

endJunction:
Divide by dt to get a rate instead of an increment.
Update the test references.


MaterialLib/MPL/Properties/SaturationDependentSwelling.cpp
```diff
@@ -97,7 +97,7 @@ PropertyDataType SaturationDependentSwelling::value(
             _lambda[i] * _p[i] * std::pow(S_eff, _lambda[i] - 1) * ei_otimes_ei;
     }
 
-    return (delta_sigma_sw * delta_S_eff).eval();
+    return (delta_sigma_sw * delta_S_eff / dt).eval();
```
wenqing: I see

PR #2833, opened 2020-03-02
Fix conan gui builds

bilke:
Disabled geotiff when `OGS_USE_CONAN=ON` and `OGS_BUILD_GUI=ON`.


endJunction:
@bilke With the visual studio version used in azure pipelines it was working from the beginning.
Lukily we will switch soon! :imp:

bilke:
@endJunction But on Azure it is 2017 too. Maybe updating the existing 2017 installation will fix it too? I can do it tomorrow morning...

endJunction:
@FZill 's installation also fails for the same reason. Let's wait for the big update in two weeks.

PR #2820, opened 2020-02-26
MPL in GroundwaterFlow process

TomFischer:
Use MPL data structures in GroundwaterFlow process.

1. [x] Feature description was added to the [changelog](https://github.com/ufz/ogs/wiki/Release-notes-6.3.0)
2. [x] Tests covering your feature were added? - Existing test are adapted.
3. [x] Any new feature or behavior change was documented? - There isn't any new feature.
4. [x] [Script](https://github.com/ufz/ogs/files/4257122/convertGroundWaterFlowToMPL.py.txt) for automatic prj file conversion.

PR #2835, opened 2020-03-03
Correct Darcy velocity calculation in GroundwaterFlow process

TomFischer:
Fixes wrong access to hydraulic conductivity in calculation of darcy velocity.

### DIFF
--- a/ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
+++ b/ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
@@ -155,7 +155,8 @@ public:
         // fetch hydraulic conductivity
         ParameterLib::SpatialPosition pos;
         pos.setElementID(_element.getID());
-        auto const k = _process_data.hydraulic_conductivity(t, pos)[0];
+        auto const k = hydraulicConductivity<GlobalDim>(
+            _process_data.hydraulic_conductivity(t, pos));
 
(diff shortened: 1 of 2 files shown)


ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
```diff
@@ -155,7 +155,8 @@ class LocalAssemblerData : public GroundwaterFlowLocalAssemblerInterface
         // fetch hydraulic conductivity
         ParameterLib::SpatialPosition pos;
         pos.setElementID(_element.getID());
-        auto const k = _process_data.hydraulic_conductivity(t, pos)[0];
+        auto const k = hydraulicConductivity<GlobalDim>(
```
wenqing: You can use MPL and formEigenTensor. With that the function of hydraulicConductivity can be removed.

TomFischer: PR #2820 will change it.

wenqing:
There are already changes in PR #2820. What it the purpose of this PR?

ProcessLib/GroundwaterFlow/GroundwaterFlowFEM.h
wenqing: I see this might be a quick fixing.

PR #2840, opened 2020-03-04
improving documentation

joergbuchwald:
Error message takes now into account the case  (detJ <0) if det J is very small.
Perhaps, we shouldn't distinguish those cases.


endJunction:
Is the ogs6py update part of this pr?
Looks good otherwise.

joergbuchwald:
I don't get why there is a diff for ogs6py. It is already up to date in ThirdParty.

PR #2843, opened 2020-03-05
Richards mechanics linear-linear shape functions

endJunction:
Add `ORDER` to shape functions.
Allows linear-linear shape functions in the RichardsMechanics process. So far only quadratic-linear combinations were possible.

Compilation times (for the RM process) are skyrocketing, as well as memory usage while compilation.


nagelt:
I ran a test on the B3 saturation benchmark. The two-fold DOF reduction from 45803 (Q2P1) to 19803 (Q1P1) yielded a reduction in simulation time by about a factor of 4. In addition to the DOF reduction I reduced IP number from 4x4 to 2x2. Results show only a slight shift, simulation is stable against pressure oscillations (storage and drainage sufficient). This can be a feature in HM, THM, etc. also!

PR #2848, opened 2020-03-05
HM permeability output

FZill:
Made the permeability in HM a tensor which will be needed to use certain permeability models.
To analyze the results of the permeability models the k-output was added.
Therefore, the variant visitor was implemented.

### DIFF
--- a/MaterialLib/MPL/Utils/GetSymmetricTensor.cpp
+++ b/MaterialLib/MPL/Utils/GetSymmetricTensor.cpp
@@ -1,0 +1,99 @@
+/*
+ * \file
+ * \copyright
+ * Copyright (c) 2012-2020, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ * Created on March 04, 2020, 5:20 PM
+ */
+
+#include "GetSymmetricTensor.h"
+
+#include "MaterialLib/MPL/PropertyType.h"
+
+namespace MaterialPropertyLib
+{
+template <int GlobalDim>
+struct GetSymmetricTensor
+{
+    SymmetricTensor<GlobalDim> operator()(double const& value) const
+    {
+        SymmetricTensor<GlobalDim> result = SymmetricTensor<GlobalDim>::Zero();
+        result.template head<3>() = Eigen::Vector3d::Constant(value);
+        return result;
+    }
+
+    SymmetricTensor<GlobalDim> operator()(Eigen::Vector2d const& values) const
+    {
+        SymmetricTensor<GlobalDim> result = SymmetricTensor<GlobalDim>::Zero();
+        result.template head<2>() = values;
+        return result;
+    }
+
+    SymmetricTensor<GlobalDim> operator()(Eigen::Vector3d const& values) const
+    {
+        SymmetricTensor<GlobalDim> result = SymmetricTensor<GlobalDim>::Zero();
+        result.template head<3>() = values;
+        return result;
+    }
+
+    SymmetricTensor<GlobalDim> operator()(Eigen::Matrix2d const& values) const
+    {
+        if constexpr (GlobalDim == 2)
+        {
+            SymmetricTensor<GlobalDim> result;
+            result << values(0, 0), values(1, 1), 0., values(0, 1);
+            return result;
+        }
+        OGS_FATAL("Cannot convert 2d matrix to 3d symmetric Tensor.");
+    }
+
+    SymmetricTensor<GlobalDim> operator()(Eigen::Matrix3d const& values) const
+    {
+        if constexpr (GlobalDim == 3)
+        {
+            SymmetricTensor<GlobalDim> result;
+            result << values(0, 0), values(1, 1), values(2, 2),
+                      values(0, 1), values(1, 2), values(0, 2);
+            return result;
+        }
+        OGS_FATAL("Cannot convert 3d matrix to 2d symmetric Tensor.");
+    }
+
+    SymmetricTensor<GlobalDim> operator()(
+        SymmetricTensor<2> const& values) const
+    {
+        if constexpr (GlobalDim == 2)
+        {
+            return values;
+        }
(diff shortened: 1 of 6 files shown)


MaterialLib/MPL/Utils/GetSymmetricTensor.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+    SymmetricTensor<GlobalDim> operator()(Eigen::Matrix2d const& values) const
+    {
+        if constexpr (GlobalDim == 2)
+        {
+            SymmetricTensor<GlobalDim> result;
+            result << values(0, 0), values(1, 1), 0., values(0, 1);
+            return result;
+        }
+        OGS_FATAL("Cannot convert 2d matrix to 3d symmetric Tensor.");
```
endJunction: Can you try this instead (and actually call a wrong version somewhere):
```suggestion
    else {
        static_assert(GlobalDim == 2, "Cannot convert 2d matrix to 3d symmetric Tensor.");
    }
```
I think a compile-time error is possible here.

ProcessLib/HydroMechanics/HydroMechanicsProcess.cpp
```diff
@@ -295,6 +295,11 @@ void HydroMechanicsProcess<DisplacementDim>::initializeConcreteProcess(
             const_cast<MeshLib::Mesh&>(mesh), "principal_stress_values",
             MeshLib::MeshItemType::Cell, 3);
 
+    _process_data.permeability = MeshLib::getOrCreateMeshProperty<double>(
```
wenqing: It would be nice  make it as an optional  initialization to save memory for  homogeneous or piece wise homogeneous material.

MaterialLib/MPL/Utils/GetSymmetricTensor.cpp
FZill: It doesn't seem to work. The error triggers only when the expr is false and since in this case GlobalDim has to be 2 it is always true, so no error is thrown.
But further, if I static_assert GlobalDim == 3 it will always fail to compile, since it compiles this function for GlobalDim = 2 and for 3 anyway, even if I don't call the wrong function.

ProcessLib/HydroMechanics/HydroMechanicsProcess.cpp
wenqing: Just a suggestion. The other parts looks good.

FZill: I added a commit to make the output optional

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
```diff
@@ -853,22 +858,65 @@ void HydroMechanicsLocalAssembler<ShapeFunctionDisplacement,
...
+    x_position.setElementID(elem_id);
+    unsigned const n_integration_points =
+        _integration_method.getNumberOfPoints();
+
+    if (_process_data.output_permeability)
+    {
+        auto const& medium = _process_data.media_map->getMedium(elem_id);
+        MPL::VariableArray vars;
+
+        // TODO (naumov) Temporary value not used by current material models.
```
TomFischer: TODO (naumov)? :smile:

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
```diff
@@ -853,22 +858,65 @@ void HydroMechanicsLocalAssembler<ShapeFunctionDisplacement,
         DisplacementDim>(_element, _is_axially_symmetric, p,
                          *_process_data.pressure_interpolated);
 
+    int const elem_id = _element.getID();
+    ParameterLib::SpatialPosition x_position;
```
TomFischer: The lines 862--865 could be move into the if condition block starting in 869.

ProcessLib/HydroMechanics/HydroMechanicsProcessData.h
```diff
@@ -57,11 +57,17 @@ struct HydroMechanicsProcessData
 
     bool const mass_lumping;
 
+    bool const output_permeability;
```
TomFischer: Shouldn't here be docu for the md-file? (\copydoc ProcessLib::HydroMechanics::HydroMechanicsProcessData::output_permeability)

ProcessLib/HydroMechanics/CreateHydroMechanicsProcess.cpp
```diff
@@ -169,7 +173,7 @@ std::unique_ptr<Process> createHydroMechanicsProcess(
         materialIDs(mesh),     std::move(media_map),
         std::move(solid_constitutive_relations),
         initial_stress,        specific_body_force,
-        mass_lumping};
+        mass_lumping,          output_permeability};
```
wenqing: Other suggestion: instead of  adding output_permeability as a member of `HydroMechanicsProcessData`, and point  `permeability` can be directly used for if-condition. To do this, permeability pointer has to be initialized before creating the instance of  HydroMechanicsProcessData and passed to HydroMechanicsProcessData as

```
      MeshLib::PropertyVector<double>* permeability = output_permeability ? MeshLib::getOrCreateMeshProperty<double>(
            const_cast<MeshLib::Mesh&>(mesh), "permeability",
            MeshLib::MeshItemType::Cell,
            MathLib::KelvinVector::KelvinVectorDimensions<
                DisplacementDim>::value) : nullptr;
```
```
    HydroMechanicsProcessData<DisplacementDim> process_data{
        materialIDs(mesh),     std::move(media_map),
        std::move(solid_constitutive_relations),
        initial_stress,        specific_body_force,
        mass_lumping, permeability};

```

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
FZill: We agreed to go without the optional output, so the if condition isn't there and this doesn't apply anymore. But before you are totally correct.

TomFischer: Ok.

PR #2849, opened 2020-03-05
improving help

joergbuchwald:
replacing true by 1, false by 0, identifySubdomains.cpp


joergbuchwald:
It's always the same. I cannot exclude ogs6py. Even tough it is already the newest version in master.

PR #2851, opened 2020-03-06
[CL] Distribute an independent mesh for chemical system

renchao-lu:
This pull request adds a mesh tag required under the tree of `<chemical system>`. With this change, one is able to restrict the place where chemical reactions take place within a specified subdomain. This utility prevents from performing unneeded chemical calculations within the subdomains where no chemical reactions take place, thereby boosting computational efficiency up.


TomFischer:
I don't expect problems for parallel models. But to be sure please also prepare a parallel test. If you want I can test an example also at JUWELS supercomputer.

ChemistryLib/PhreeqcIO.cpp
```diff
@@ -139,12 +142,18 @@ void PhreeqcIO::doWaterChemistryCalculation(
 void PhreeqcIO::setAqueousSolutionsOrUpdateProcessSolutions(
     std::vector<GlobalVector*> const& process_solutions, Status const status)
 {
-    std::size_t const num_chemical_systems = _aqueous_solutions.size();
+    std::size_t const num_chemical_systems = _mesh.getNumberOfBaseNodes();
+
+    auto const chemical_system_map =
+        *_mesh.getProperties().template getPropertyVector<std::size_t>(
+            "bulk_node_ids", MeshLib::MeshItemType::Node, 1);
```
endJunction: The default bulk mesh will not have the `bulk_node_ids`. (And it should not.)

renchao-lu: I haven't realized.

endJunction: => The chemical domain mesh must always be a subdomain mesh (not necessarily smaller).

Maybe it would be better to have a mesh property indicating where the chemical system should be solved?

renchao-lu:
> I don't expect problems for parallel models. But to be sure please also prepare a parallel test. If you want I can test an example also at JUWELS supercomputer.

As yet we don't have a benchmark run in parallel, because I have not yet touched parallel computing for local chemical calculation.

ChemistryLib/PhreeqcIO.cpp
renchao-lu: Good point! One need apply the mesh tool `identifySubdomain` to create a mesh file containing the property `bulk_node_ids`, even in the case that chemical reactions take place over the entire domain. Also, the tag `<mesh>` should no longer be optional.

renchao-lu:
The `gml` files are replaced with `vtu` files because by far the tag `<meshes>` and `<geomery>` cannot coexist.

TomFischer:
> > I don't expect problems for parallel models. But to be sure please also prepare a parallel test. If you want I can test an example also at JUWELS supercomputer.
> 
> As yet we don't have a benchmark run in parallel, because I have not yet touched parallel computing for local chemical calculation.

Please parallelize one of the benchmarks. If you have questions I can support you.

renchao-lu:
> > > I don't expect problems for parallel models. But to be sure please also prepare a parallel test. If you want I can test an example also at JUWELS supercomputer.
> > 
> > 
> > As yet we don't have a benchmark run in parallel, because I have not yet touched parallel computing for local chemical calculation.
> 
> Please parallelize one of the benchmarks. If you have questions I can support you.

Thanks in advance for your support. Parallel computing for solving chemical systems is on schedule. As yet i have no idea where to start. It would be great to get some instructions from your side. Let's discuss if you have time.

renchao-lu:
Rebased.

PR #2853, opened 2020-03-09
Move MPL porosity to solid phase in LiquidFlow

TomFischer:


renchao-lu:
I don't understand where of the liquid flow process is different to other processes. Why the porosity is no longer a medium property?

endJunction:
@renchao-lu The porosity, so far, in RM/HM/TM/THM has been a solid property. There are some arguments, why it is not. We will discuss this on the community meeting to get a consensus.
Feel free to extend the [discussions page](https://github.com/ufz/ogs/wiki/Community-meeting-2020)

montoyav:
In my opinion porosity should be medium property, specially for RT simulations. Unfortunately, I can not attend the community meeting, but I will be happy to know the outcome of this discussion :-). I will add some topics that I think should be discussed there...

PR #2854, opened 2020-03-10
Added a member of inverse_value to MPL::Property

wenqing:
For capillary saturation models, the newly added member function of `inverse_value` gets capillary pressure via saturation. In the existing classes for capillary saturation models, the member function of `value` represents function of S(p_c). This is not adequate to the conventional definition of such models. Here are suggestions for future changes for  capillary pressure model classes:
```
Property::value:  Pc(S)
Property::inverse_value:  S(Pc)
```

2. [x] Tests covering your feature were added? Yes
3. [x] Any new feature or behavior change was documented? yes

### DIFF
--- a/Documentation/ProjectFile/properties/property/SaturationBrooksCorey/t_maximum_capillary_pressure.md
+++ b/Documentation/ProjectFile/properties/property/SaturationBrooksCorey/t_maximum_capillary_pressure.md
@@ -1,0 +1,1 @@
+The maximum capillary pressure, which is an optional input.
--- a/Documentation/ProjectFile/properties/property/SaturationVanGenuchten/t_maximum_capillary_pressure.md
+++ b/Documentation/ProjectFile/properties/property/SaturationVanGenuchten/t_maximum_capillary_pressure.md
@@ -1,0 +1,1 @@
+The maximum capillary pressure, which is an optional input.
--- a/MaterialLib/MPL/Properties/CreateSaturationBrooksCorey.cpp
+++ b/MaterialLib/MPL/Properties/CreateSaturationBrooksCorey.cpp
@@ -34,8 +34,14 @@ std::unique_ptr<SaturationBrooksCorey> createSaturationBrooksCorey(
         //! \ogs_file_param{properties__property__SaturationBrooksCorey__entry_pressure}
         config.getConfigParameter<double>("entry_pressure");
 
-    return std::make_unique<SaturationBrooksCorey>(residual_liquid_saturation,
-                                                   residual_gas_saturation,
-                                                   exponent, entry_pressure);
+    // Optional input to get the maximum capillary pressure.
+    auto pc_max =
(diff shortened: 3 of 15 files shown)


endJunction:
Not sure about the implementation of the inverse for the van Genuchten saturation. For an inverse the identity f^{-1}(f(x)) = x must hold for the whole domain of definition. But maximum capillary pressure cuts it off. Maybe it is a process' internal property?

wenqing:
> Not sure about the implementation of the inverse for the van Genuchten saturation. For an inverse the identity f^{-1}(f(x)) = x must hold for the whole domain of definition. But maximum capillary pressure cuts it off. Maybe it is a process' internal property?

It should be a property of the water retention curve Pc(S). When S->0, Pc(S)->∞. `pc_max` can be used as a restriction for numerical stability.

endJunction:
> It should be a property of the water retention curve Pc(S). When S->0, Pc(S)->∞. `pc_max` can be used as a restriction for numerical stability.

numerical stability of the material property or the stability of the process?

wenqing:
> > It should be a property of the water retention curve Pc(S). When S->0, Pc(S)->∞. `pc_max` can be used as a restriction for numerical stability.
> 
> numerical stability of the material property or the stability of the process?

Like the saturation restriction, numerically it is for the stability in solving the nonlinear system, physically the capillary pressure can not be infinite. If the restriction is set in process level, meaning set a restriction in solution, it is meaningless both in physics and numerics.  

For the maximum capillary pressure restriction,  there are formulas on page 340 in Computational Methods in Environmental Fluid Mechanics by Olaf).

endJunction:
Maybe it is better to have a different property such as `capillary_pressure`?
I'm still against a function called `inverseValue` not being an inverse.
Another thought: In RM, for example, it is possible to set saturation to a constant value (*e.g.* 1 for simple setups). For such case it is not possible to define a useful inverse relation between saturation and capillary pressure.
So for now I tend to `capillary_pressure` property. Others here might have a better ideas.

wenqing:
> Maybe it is better to have a different property such as `capillary_pressure`?
> I'm still against a function called `inverseValue` not being an inverse.
> Another thought: In RM, for example, it is possible to set saturation to a constant value (_e.g._ 1 for simple setups). For such case it is not possible to define a useful inverse relation between saturation and capillary pressure.
> So for now I tend to `capillary_pressure` property. Others here might have better ideas.

Then I have the following idea with property type of `saturation_capillary_pressure` and `capillary_pressure_saturation`:
```
template <typename T>
class vanGenuchtenBase : public Property
{
    value()
    ....
};
 
class GenuchtenSaturationCapillaryPressure : public vanGenuchtenBase<Derived>
{
    value()
    ...
}; 
class GenuchtenCapillaryPressureSaturation : public vanGenuchtenBase<Derived>
{
    value()
    ...
};

```

PR #2858, opened 2020-03-11
Categorized the files of relative permeability and capillary pressure models

wenqing:
1. ~Renamed `PropertyType::saturation` to `PropertyType::saturation_capillary_pressure` to make it represent the model type and to avoid misleading, e.g. when the curve type is taken.~
2. Moved the files  of relative permeability and capillary pressure models to two subdirectories, respectively.


wenqing:
Any comments on this PR? I want to proceed based on these changes.

MaterialLib/MPL/PropertyType.h
```diff
@@ -311,7 +311,7 @@ static const std::array<std::string, PropertyType::number_of_properties>
                              "residual_gas_saturation",
                              "residual_liquid_saturation",
                              "retardation_factor",
-                             "saturation",
+                             "saturation_capillary_pressure",
```
endJunction: This is now ambiguous. Would this return saturation or pressure? Or both?
If there is need for capillary pressure model (*e.g.* from liquid saturation), I'd recommend to add a new property `capillary_pressure`.

wenqing: Yes it is ambiguous. If the property name refers to the meaning the returned value, `saturation` is a proper name. Then all properties that give saturation is covered by one property regardless their physical property.  Anyway, I am going to remove the commit of renaming.

endJunction: Let's try the simple way of new `capillary_pressure` property with a single `CapillaryPressureVanGenuchten` model. Then, when it is working together with H2-p-rho, we discuss how to generalize. What do you think?

wenqing: Agree. The commits of renaming have been removed.

MaterialLib/CMakeLists.txt
```diff
@@ -16,6 +16,8 @@ append_source_files(SOURCES Fluid/WaterVaporProperties)
 
 append_source_files(SOURCES MPL)
 append_source_files(SOURCES MPL/Properties)
+append_source_files(SOURCES MPL/Properties/CapillaryPressureModel)
```
endJunction: This is the new folder for the capillary pressure models. The saturation models shouldn't be here, isn't it?

wenqing: Physically, there is no saturation models, which in the source code are the inverse of  capillary pressure (saturation) model. I would prefer to put all  capillary pressure saturation model and their inverse models in this directory. The directory name would be CapillaryPressureSaturation.

Scinopode: As far as I understand this discussion, the question is whether a new directory is called 'capillary...' or 'saturation...'. From Dimas perspective, there should be two directories, one for capillary pressure models and one for saturation models. From Wenqings perspective, there is just one type of model, which returns either capillary pressure or saturation depending on the process. 

Personally, I would also prefer to have just one implementation for each model, which is then inverted or not depending on the process requirements. In that case, I agree to Dima that we need to make sure that the functions are always invertible (we need to regard cut-off pressures, etc. in that case). To avoid confusion, I suggest to rename the new directory to 'characteristic_curves', since this is the proper name for this type of model and it shows that those functions can be used in any direction.

wenqing: @Scinopode 
1. For the invertible, Dima suggested two classes for one model, one for direct and another one for invertible. I thought that it was good point because of the meaning of  the returned value of `...::value()` is consistent with the property name. My previous implementation is adding inverse_value to Property (see my closed PR #2854). Which one is better? I prefer to take Dima's suggection and have already implemented some.

2. For the name of new directory, since I want to moved all capillary pressure  saturation model into it, `characteristic_curves` seems can not represents those physical properties properly.

endJunction: As a simple suggestion, we could do the generalization (*e.g.* moving files around, creating inheritance trees, *etc.*) after the `CapillaryPressureVanGenuchten` model is added.

wenqing:
@endJunction @Scinopode Do you have any further comments?

MaterialLib/MPL/Properties/CapillaryPressureSaturation/SaturationVanGenuchten.h
```diff
@@ -15,20 +15,35 @@ namespace MaterialPropertyLib
 class Medium;
 class Phase;
 class Component;
-
-/// The van Genuchten soil characteristics function.
-///
-/// This property must be a medium property, it computes the saturation of the
-/// wetting phase as function of capillary pressure.
+/**
+ *   \brief van Genuchten water retention model
```
endJunction:
```suggestion
 *   \brief The van Genuchten water retention model.
```

endJunction: Not sure, if water is the only valid fluid...

MaterialLib/MPL/Properties/CapillaryPressureSaturation/SaturationVanGenuchten.h
```diff
@@ -15,20 +15,35 @@ namespace MaterialPropertyLib
...
+ * This property must be a medium property, it computes the saturation of the
+ * wetting phase as function of capillary pressure.
+ *
+ *   \f[p_c=p_b (S_e^{-1/m}-1)^{1-m}\f]
```
endJunction: Shouldn't it be other way around S(p_c)?

MaterialLib/MPL/Properties/CapillaryPressureSaturation/SaturationVanGenuchten.h
```diff
@@ -15,20 +15,35 @@ namespace MaterialPropertyLib
...
+ *    Note:
+ *     \f[m=1/(1-n)\f].
+ *
+ *    If \f$\alpha\f$ instead of \f$p_b\f$ is available, \f$p_b\f$ can
+ * be calculated
+ * as
+ *    \f[p_b=\rho g/\alpha\f]
+ *
+ * This property must be a medium property, it computes the saturation
+ *  of the wetting phase as function of capillary pressure.
```
endJunction:
```suggestion
```
Repetition.

wenqing:
One benchmark (`2D_Adaptive_dt_ThermalConvection_constviscosityStaggeredScheme` ) failed on dokcer, but failure  is not caused by the changes in this PR.

MaterialLib/MPL/Properties/CapillaryPressureSaturation/SaturationVanGenuchten.h
wenqing: Changed it to `The van Genuchten capillary pressure model.`

MaterialLib/MPL/Properties/CapillaryPressureSaturation/SaturationVanGenuchten.h
wenqing: Added a formula of S(pc) by amending the last commit.

MaterialLib/MPL/Properties/CapillaryPressureSaturation/SaturationVanGenuchten.h
wenqing: removed.

wenqing:
> One benchmark (`2D_Adaptive_dt_ThermalConvection_constviscosityStaggeredScheme` ) failed on dokcer, but failure is not caused by the changes in this PR.

I will take a look.

MaterialLib/MPL/Properties/CapillaryPressureSaturation/SaturationVanGenuchten.h
endJunction: The implementation is slightly different. If I didn't make a mistake in writing this down, it is
Seff(p) = ((p^n + 1)^(-m), where p is pc/pb, and n is 1/(1-m). The difference is in the exponent (here n), and you have 1/m.
At the end there is clamping of S between Sr and Smax. In the beginning too, if p_c <= 0, Smax is returned.

In the documentation you introduce n, which is not explained.

endJunction: For future PR's I recommend to separate different topics in the commits, like documentation is one commit, renaming is another. Makes reviewers' life easier.

wenqing: Corrected. See the screen shot:
f

endJunction: 1.
> At the end there is clamping of S between Sr and Smax. In the beginning too, if p_c <= 0, Smax is returned.

2.
> In the documentation you introduce n, which is not explained.

..., and actually not used.

3.
Additionally, m can not be 1, and it is tested to be in (0,1) open interval.

wenqing: Changed as:
f

wenqing: Also corrected: S(pc) = S_max if pc<0

endJunction: Nice. This looks much closer to the actual implementation.

endJunction:
This was too fast...

wenqing:
> This was too fast...
> I'll push my changes to master.

oops!

MaterialLib/MPL/Properties/CapillaryPressureSaturation/SaturationVanGenuchten.h
endJunction: Could you explain what \alpha is?

wenqing: \alpha is an alternative  parameter of p_b of the model. For some cases, \alpha instead of p_b is given, p_b can be calculated by that formula.

endJunction: You have a reference? In the original vanGenuchten1980 paper I don't see such an expression.
Actually I also don't see the entry pressure there too.

wenqing: See Olaf'book or
https://mooseframework.inl.gov/modules/porous_flow/capillary_pressure.html

PR #2860, opened 2020-03-16
Rename Groundwater flow to SteadyStateDiffusion

TomFischer:
The current implementation of groundwater flow process is only an elliptic process. Therefore I suggest to rename the process to EllipticProcess. Furthermore, the liquid flow process can be used for simulation of groundwater issues.

- [x] changelog


Documentation/ProjectFile/prj/processes/process/ELLIPTIC_PROCESS/c_ELLIPTIC_PROCESS.md
```diff
@@ -0,0 +1 @@
+Process described by an elliptic partial differential equation.
```
endJunction: Maybe it is better to mention the scalar variable...
```suggestion
Process described by a scalar, elliptic partial differential equation.
```

Documentation/ProjectFile/prj/processes/process/ELLIPTIC_PROCESS/process_variables/t_process_variable.md
```diff
@@ -0,0 +1,2 @@
+The process variable for the elliptic process could be for instance
```
endJunction:
```suggestion
The scalar process variable for the elliptic process, for example
```

wenqing:
This name refers to PDE type, while other processes' names refer to physical problems. Besides the variable is still named as pressure. How about name it as SteadyStateSaturatedFlow? With that, we can rename LiquidFlow as SaturatedFlow. The LiquidFlow can use pressure and head as primary varabile, therefore it can be used to simulate liquid flow and groundwater flow.

TomFischer:
> This name refers to PDE type, while other processes' names refer to physical problems. Besides the variable is still named as pressure. How about name it as SteadyStateSaturatedFlow? With that, we can rename LiquidFlow as SaturatedFlow. The LiquidFlow can use pressure and head as primary varabile, therefore it can be used to simulate liquid flow and groundwater flow.

 1. Yes, you are right - the name comes from the mathematical concept. I would say the 'physical' concept is a diffusion process.
 2. I suggest to rename the primary variable from 'pressure' to 'u' that is often used in the literature. What do you think?
 3. We could also rename LiquidFlow to ParabolicProcess (mathematical notation) or AdvectionDiffusionProcess (physical notation) or SaturatedFlow (engineering notation?). But this would be an other PR.

montoyav:
In my opinion it is easier (for users and people not too much in the development) when the name makes reference to the physical or "engineering notation".

endJunction:
Because the formulation allows one to solve many types of physical behaviours in it is also a good idea to call it in general way. In particular "GroundwaterFlow", "SteadyStateSaturatedFlow", "SteadyStateDiffusionOfX" with X being temperature, concentration etc. are specific to an application at hand.

In a particular application the variable name can be generic, like 'x' or 'u', or can be related to the application, like 'temperature', 'concentration'.

Let's postpone the LiquidFlow discussion for the moment.

endJunction:
@montoyav The physical behaviour would be "steady state diffusion" (of pressure, temperature, etc.).

wenqing:
Then I think SeadyStateDiffusion with u is better.

montoyav:
OK, so then for me "SteadyStateDiffusion" would be very clear as @wenqing is suggesting

Tests/Data/Elliptic/circle_radius_1/circle_1e1_axi.prj
```diff
@@ -5,7 +5,7 @@
     <processes>
         <process>
             <name>GW23</name>
```
wenqing: How about change it for a meaningful name?

endJunction: @wenqing This is now 4 years to late for this review ;)
But yes, it could be changed too.

endJunction:
Let us sit it here for one more day, maybe there will be more suggestions.

Tests/Data/Elliptic/circle_radius_1/circle_1e1_axi.prj
TomFischer: When we have agreed on a name I can change the project files, too.

TomFischer:
Summary at 2020-03-18 12:00 : 'SteadyStateDiffusion' is the favorite of @montoyav , @wenqing , @endJunction and me.

TomFischer:
I renamed the process again. Since it is allowed that the diffusion coefficient depends on the primary variable it is necessary to fix the primary variable - at least at the moment. Maybe this will change in future. But for now, I want to keep pressure as primary variable.

scripts/cmake/ProcessesSetup.cmake
```diff
@@ -2,7 +2,7 @@
 # processes to be build.
 set(ProcessesList
     ComponentTransport
-    GroundwaterFlow
+    SteadyStateDiffusion
```
endJunction: Can you put it in more alphabetic order?

TomFischer: I wonder why 'sed' didn't reorder it automatically :smiley:  - I will do it for 'sed'.

bilke:
@TomFischer Tom, can you please have a look at the web links which have to be adapted as well:

```
 content/docs/benchmarks/elliptic/elliptic-neumann.pandoc
 --------------------------------------------------------
https://github.com/ufz/ogs-data/blob/master/Elliptic/square_1x1_GroundWaterFlow/square_1x1.gml
https://github.com/ufz/ogs-data/blob/master/Elliptic/square_1x1_GroundWaterFlow/square_1e2.prj

 content/docs/benchmarks/elliptic/poisson_equation.pandoc
 --------------------------------------------------------
https://github.com/ufz/ogs/tree/master/Tests/Data/Elliptic/square_1x1_GroundWaterFlow_Python/square_1e3_poisson_sin_x_sin_y.prj
https://github.com/ufz/ogs/tree/master/Tests/Data/Elliptic/square_1x1_GroundWaterFlow_Python/sin_x_sin_y_source_term.py

 content/docs/benchmarks/elliptic/elliptic-dirichlet-volumetric-source-term.pandoc
 ---------------------------------------------------------------------------------
https://github.com/ufz/ogs/blob/master/Tests/Data/Elliptic/square_1x1_GroundWaterFlow/square_1e2_volumetricsourceterm.prj

 content/docs/benchmarks/elliptic/elliptic-robin.pandoc
 ------------------------------------------------------
https://github.com/ufz/ogs/tree/master/Tests/Data/Elliptic/line_1_GroundWaterFlow/line_1e1_robin_right_picard.prj
https://github.com/ufz/ogs/tree/master/Tests/Data/Elliptic/line_1_GroundWaterFlow/line_1e1_robin_left_picard.prj
```

PR #2863, opened 2020-03-18
[T/HT] Cleanup prj files.

TomFischer:


TomFischer:
One benchmark not working in CI, but this benchmark passes on envinf1, my desktop at UFZ and my desktop at home.

PR #2865, opened 2020-03-19
Small improvements for parallel computing

TomFischer:

### DIFF
--- a/Applications/CLI/CMakeLists.txt
+++ b/Applications/CLI/CMakeLists.txt
@@ -66,7 +66,7 @@ endif()
 if(OGS_USE_PETSC)
     target_link_libraries(ogs PRIVATE ${PETSC_LIBRARIES})
 endif()
-if(OGS_USE_MPI)
+if(OGS_USE_MPI AND TARGET MPI::MPI_CXX)
     target_link_libraries(ogs PRIVATE MPI::MPI_CXX)
 endif()
 
--- a/MathLib/LinAlg/PETSc/PETScVector.h
+++ b/MathLib/LinAlg/PETSc/PETScVector.h
@@ -168,14 +168,14 @@ public:
     PetscScalar get(const PetscInt idx) const;
 
     //! Exposes the underlying PETSc vector.
-    PETSc_Vec getRawVector() { return _v; }
+    PETSc_Vec& getRawVector() { return _v; }
     /*! Exposes the underlying PETSc vector.
      *
      * \warning
      * This method is dangerous insofar as you can do arbitrary things also
      * with a const PETSc vector!
      */
-    PETSc_Vec getRawVector() const { return _v; }
+    PETSc_Vec const&  getRawVector() const { return _v; }
     /*!
        Copy local entries including ghost ones to an array
        \param u Preallocated vector for the values of local entries.


MathLib/CMakeLists.txt
```diff
@@ -57,7 +57,7 @@ endif()
 
 if(OGS_USE_PETSC)
     target_include_directories(MathLib PUBLIC ${PETSC_INCLUDE_DIRS})
-    target_link_libraries(MathLib PUBLIC ${PETSC_LIBRARIES})
+    target_link_libraries(MathLib PUBLIC ${PETSC_LIBRARIES} HYPRE)
```
bilke: Where does `HYPRE` come from? If it is a library it should be searched for with `find_package()` and then there would be a variable, e.g. `${HYPRE_LIBRARIES}`, to link to. But you would also need a `FindHypre.cmake` (e.g. from [here](https://github.com/LLNL/sundials/blob/master/config/FindHypre.cmake)). If it is a dependency of PETSc then the library should be included in the `${PETSC_LIBRARIES}`-variable.

wenqing: It does not need to link HYPRE. When PETSc is compiled, one just need to enable HYPRE option, i.e. PETSc can include HYPRE.

TomFischer: As far as I understood: HYPRE is a a package that can be optional included in PETSc. HYPRE offers some interesting preconditoners. If HYPRE is enabled in PETSc and we want to use a preconditioner offered by PETSc/HYPRE we have to link also against HYPRE. At least that was recommended by Juelich Supercomputer support people.

TomFischer: I see the problem in CI. Our PETSc seems not to be build with the optional package HYPRE.

endJunction: @bilke Maybe an OGS_USE_HYPRE option could be used if there is no information coming from petsc cmake?

wenqing: When configure  PETSc for compilation the option of `--download-hypre` can include HYPRE, e.g  `./configure --download-hypre ...`   [see the installation instruction of PETSc](https://www.mcs.anl.gov/petsc/documentation/installation.html) . With that configuration , HYPRE is compiled together with PETSc and it is included into the PETSc path, and one does not need to set additional link to HYPRE manually.
 I am not sure how PETSc is configured in Juelich  Supercomputer. When I used that machine, I compiled PETSc and installed it into my home directory with `./configure --download-hypre ... --prefix=/home/...` , and I could use hypre without problem. Maybe the new machine in Juelich  has new policy for software installation.

wenqing: If the new Juelich Supercomputer needs that, I prefer to set the link as a special option for Juelich only. Generally, if PETSc is configured with `--download-hypre`, having 'export PETSC_DIR' in shell environment is enough.

bilke: Our petsc module on envinf was not built with the hypre option. I can add that (but would wait for #2827 because I change modules there anyway). I guess Wenqing is right that there is no need to explicitly link to hypre. @TomFischer Does OGS correctly build without explicitly linking to hypre on Juwels when using the petsc module with hypre enabled?

TomFischer: The PETSc module on JUWELS is build with HYPRE support. I loaded the PETSc module together with all the other modules needed to compile OGS. Then OGS was correctly build on JUWELS, but I had problems using HYPRE preconditioner. These problems only occurred at run time - OGS crashed with a PETSc error. The JSC support advised me to link with HYPRE and the problems disappeared.

endJunction: Maybe some builds of petsc are static, others are dynamic?
This would explain the difference and the need for explicit linkage in one case but not in another.

wenqing: Might be. @TomFischer  Could you please check whether the PETSc libs are dynamic or static? Normally, the compiled libs of PETSc are dynamic.

wenqing: @endJunction  When I test this, I got a problem to run ogs (PETSc or serial one). Can you run ogs compiled from the latest master without any problem.  I have left an message about the error on [town-square](https://mattermost.opengeosys.org/ogs-devs/channels/town-square), please take a look.

I will remove this comment once you read it.

wenqing: The figure shows the libs of PETSc installed on my laptop with the configuration of `--download-hypre`, where the libs of hypre in the same directory of PETSc libs. Besides hypre libs are included in ${PETSC_LIBRARIES} by findPETSc. I can use hypre without any additional setting except that for PETsc.
petcs_lib

It might be that hypre was not installed as the external package of PETSc on  Juwels  super computer but as an individual package. Then you may need findHypre for your change.

TomFischer: When I loaded the PETSc module on JUWELS the following files are in the corresponding folder: 
```
libchaco.a         libhdf5_fortran.la    libhdf5.la         libparmetis.a libspai.a                    libsundials_nvecserial.a   libzmumps.a
libcmumps.a        libhdf5_hl.a          libhdf5.settings   libparms.a    libsundials_cvode.a          libsundials_nvecserial.la  petsc
libdmumps.a        libhdf5hl_fortran.a   libHYPRE.a         libpetsc.a    libsundials_cvode.la         libsuperlu.a               pkgconfig
libhdf5.a          libhdf5hl_fortran.la  libmetis.a         libpord.a     libsundials_nvecparallel.a   libsuperlu_dist.a
libhdf5_fortran.a  libhdf5_hl.la         libmumps_common.a  libsmumps.a   libsundials_nvecparallel.la  libtriangle.a
```
When my understanding is correct the libs are statically linked at compile time!?

TomFischer: Additionally, I found the configure options PETSc was build with. HYPRE is build / installed by the PETSc package.

bilke: Maybe it works when Petsc is dynamically linked..
Anyway, the [FindPETSc.cmake](https://github.com/ufz/ogs/blob/master/scripts/cmake/jedbrown/FindPETSc.cmake) we use does not look for HYPRE or other additional libraries so we have to take care of linking these. To proceed: can we assume that we always want Petsc build HYPRE support? If yes then I can hard-code something ..

endJunction: Tom has looked up, ogs does not link petsc dynamically. He is still looking if the petsc/hypre symbols are indeed statically linked into ogs...

TomFischer: I tested again without explicitly linking with HYPRE and now it runs. I can not reproduce the crashes. I'll adapt the PR. Sorry for the additional work time!

wenqing: @bilke  Building with HYPRE support is better. Also with SupperLU.  Here is my configuration
```
./configure PETSC_ARCH=linux-fast  COPTFLAGS='-O3 -march=native -mtune=native' CXXOPTFLAGS='-O3 -march=native -mtune=native' --with-cc=mpicc --with-cxx=mpicxx --with-fc=mpif90 --download-fblaslapack --download-metis --download-parmetis --download-superlu_dist --download-scalapack --download-mumps  --download-hypre --with-c2html=0 --with-debugging=0 --with-cxx-dialect=C++11 --prefix=/opt/petsc
```

bilke: @wenqing Thanks, I will use your configuration for the next eve module.

PR #2868, opened 2020-03-20
[PVTU] Corrected the values of vtkGhostType

### LINKED ISSUE #2866: Incorrect writing of vtkGhostType-array in PVTU files
This is regarding PR #2846 where the "vtkGhostType"-array was introduced. 

Currently, *all* ghost cells are set to "1" while non-ghost cells are set to "0". Since VTK/ParaView ignore cells with vtkGhostType > 0, loading PVTUs will currently result in elements apparently missing at the boundaries of mesh parts:
ghosts

Instead, VTK expects the array to work a little different. Here's a minimal example:
pvtu

The first part A of the PVTU is orange, the second part B is blue. The pink cells are the overlapping region existing in both parts.

What VTK, ParaView (and especially the vtkRemoveGhosts-filter) expect, is that

* *all* cells for A have a vtkGhostType-value of "0"
* the blue cells of B have a vtkGhostType-value of "0"
* the pink cells of B have a vtkGhostType-value of "1"

I have a working minimal example (the one shown in the screenshot) if anyone is interested. I believe I previously sent this to @wenqing and @endJunction but I can send it again if you like.

Can we - for the moment - remove the changes added in #2846 ? I would argue that the current implementation is actually worse than not having the vtkGhostType at all. Currently you need to somehow trick ParaView into not ignoriering ghost cells, otherwise results are displayed wrong.  
A quick-and-dirty solution would be to simply change the name of the new array. Anything except "vtkGhostType" will be ignored by ParaView and can be handled like any other array.

wenqing:
This PR fixes #2866
Screenshot from 2020-03-20 10-55-35

### DIFF
--- a/MeshLib/IO/MPI_IO/NodePartitionedMeshReader.cpp
+++ b/MeshLib/IO/MPI_IO/NodePartitionedMeshReader.cpp
@@ -111,9 +111,9 @@ MeshLib::NodePartitionedMesh* NodePartitionedMeshReader::read(
         vtkGhostType->reserve(_mesh_info.regular_elements +
                               _mesh_info.ghost_elements);
         std::fill_n(std::back_inserter(*vtkGhostType),
-                    _mesh_info.regular_elements, 0);
+                    _mesh_info.regular_elements, 1);
         std::fill_n(std::back_inserter(*vtkGhostType),
-                    _mesh_info.ghost_elements, 1);
+                    _mesh_info.ghost_elements, 0);
     }
 
     INFO("[time] Reading the mesh took %f s.", timer.elapsed());


rinkk:
Unfortunately, this doesn't solve the problem.

I've just added a comment to the issue because I noticed another thing:
In line 101, the source says

```
mesh->getProperties().createNewPropertyVector<std::int8_t>(
            "vtkGhostType", MeshLib::MeshItemType::Cell, 1);
```

The data type here musst be ```unsigned char``` such that VTK writes this as a ```UInt8```-array (instead of ```Int8``` as it is done now). This will make ParaView automatically interpret the array.

All elements in the array with a value>0 will automatically not be rendered by ParaView (and actually removed by applying the ```vtkRemoveGhosts```-filter).

So ```std::fill_n(std::back_inserter(*vtkGhostType), _mesh_info.regular_elements, 0);``` was correct. 
It's just that *some* of the ghost nodes need to be set to 0 and some need to be set to 1, as shown in the example posted in #2866.

wenqing:
Changing number type to unit8_t does not help, and I am looking for other reason.

wenqing:
`unit8_t ` gives UInt8 in vtu.

rinkk:
Changing the data type to Uint8 does not solve this particular problem, but it's essential for correctly displaying the datasets in ParaView and for vtkRemoveGhosts to work.

wenqing:
I seem found the problem. Since there is no detailed description of vtkGhostType available on Internet, I digged its features by trying and found:
1) cells with vtkGhostType = 0 are marked for rendering.
2) cells with vtkGhostType = 1 (ghost cell) must be unique among partitions. For example, a cell is shared by different partitions, meaning the cell is a ghost cell. Then the cell can only be marked as  vtkGhostType = 0 in one partition for rendering, and  it must be marked as vtkGhostType = 1 in the other partitions. With that condition, the rendering in the overlapped domains is skipped by using the filters of RemoveGhostInformation and CleanCellstoGrid

To realize 2), MPI communications have to be conducted to
 * collect  vtkGhostType and global IDs among partitions,
 * make   vtkGhostType = 1 unique in the ghost cells,
 * seed the processed vtkGhostType data to partitions. 

It is a little bit work to do. There might be other better solutions. Therefore  I want to do it in future and remove the previous changes associated with vtkGhostType.

In the following figure, you can see that  only the outer surface of the domain of cells with vtkGhostType=0 is visualized by pvtu (left figure. cells inside the surface are hided for rending ). After applying  the filters of RemoveGhostInformation and CleanCellstoGrid, the cells with vtkGhostType=0 are rendered (right figure).  If ghost cells are set  vtkGhostType=0 uniquely, the whole domain can be rendered.
Screenshot from 2020-03-20 14-32-02
.

rinkk:
Yes. Applying the pipeline you've mentioned to a PVTU to turn in into a a regular (non-overlapping) VTU is basically what the command line utility does that I am currently writing.  

However, for the tool (or the included VTK filters) to work, the vtkGhostType array needs to be set correctly. vtkRemoveGhosts doesn't actually do anything to analyse ghost nodes/elements but just acts based on the information in vtkGhostType.

PR #2869, opened 2020-03-20
[pvtu] Removed vtkGhostType

wenqing:
Removed vtkGhostType due to issue #2866. The overlapping in pvtu will be avoided by using computations together with MPI communications, or by other decent method (see my comment in  # #2868).

### DIFF
--- a/MeshLib/IO/MPI_IO/NodePartitionedMeshReader.cpp
+++ b/MeshLib/IO/MPI_IO/NodePartitionedMeshReader.cpp
@@ -18,7 +18,6 @@
 #include <logog/include/logog.hpp>
 
 #ifdef USE_PETSC
-#include <algorithm>
 #include <mpi.h>
 #endif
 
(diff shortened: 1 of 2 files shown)


MeshLib/IO/MPI_IO/NodePartitionedMeshReader.cpp
```diff
@@ -97,25 +97,6 @@ MeshLib::NodePartitionedMesh* NodePartitionedMeshReader::read(
...
-    {
-        WARN(
-            "Could not create PropertyVector for the cell values of "
-            "vtkGhostType in Mesh.");
-    }
-    else
-    {
-        vtkGhostType->reserve(_mesh_info.regular_elements +
-                              _mesh_info.ghost_elements);
-        std::fill_n(std::back_inserter(*vtkGhostType),
```
endJunction: `<algorithm>` included is not need, too.
The simplest way to revert a commit is to use `git revert`.

wenqing: Reverted by using git revert.

Issue #2861, opened 2020-03-17
ogs {clang|gcc}/debug hangs in logog's mutex


endJunction:
### Description

Compiling ogs +in debug mode with gcc (same with clang in debug mode)+ ~with python~ and running ~python~ ctests like `Tests/Data/Elliptic/square_1x1_GroundWaterFlow/square_1e0.prj` does not advance. The last message is "info: This is OpenGeoSys-6 version 6.3.0-46-g3df4ac5ba."

In the debugger we have (after sufficiently long waiting time):
```
#0  0x00007fffeb17256f in __lll_clocklock_wait () from /usr/lib/libpthread.so.0
#1  0x00007fffeb16aaf9 in __pthread_mutex_lock_full () from /usr/lib/libpthread.so.0
#2  0x00007fffeb2e7d4e in logog::Mutex::MutexLock (this=0x555555722c30) at /home/naumov/w/ogs/s/ThirdParty/logog/src/mutex.cpp:33
#3  0x00005555555ab571 in main (argc=2, argv=0x7fffffffdbf8) at /home/naumov/w/ogs/s/Applications/CLI/ogs.cpp:154
```

## Steps to Reproduce the Problem

On archlinux w/o conan
  1. cmake ogs_sources/ -DOGS_USE_CONAN=Off -G Ninja ~-DOGS_USE_PYTHON=ON~
  2. ninja ogs
  3. bin/ogs Tests/Data/Elliptic/square_1x1_GroundWaterFlow_Python/square_1e3_laplace_eq.prj

**Expected behavior:** [What you expected to happen]
The test must proceed with assembly etc.

**Actual behavior:** [What actually happened]
The full output after few minutes is:
```
info: This is OpenGeoSys-6 version 6.3.0-46-g3df4ac5ba.
```
## Specifications
gcc-9.3.0
python-3.8.2
clang-9.0.1

TomFischer:
I can reproduce the problem in gcc debug build, last 'OGS code line' is BaseLib/ConfigTreeUtil.cpp:70. The filepath variable looks reasonable.

In release mode, but enabled debug output (-l debug), the problem doesn't occur here.

bilke:
I tested this with the container generated in #2765 and it works:

```
[...]

PR #2877, opened 2020-03-26
[CMake] Fix segmentation fault in ThirdParty/IPhreeqc

renchao-lu:
As titled. 

In order to fix the issue mentioned in #2861,  the syntax `add_subdirectory(ThirdParty)` is reverted back to the place where it was.


bilke:
@renchao-lu Thanks for fixing this!

PR #2879, opened 2020-03-26
Fix dead links and use urlchecker

bilke:

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -800,9 +800,7 @@ pipeline {
           }
           steps {
             dir('web') {
-              sh 'yarn'
-              sh 'hugo'
-              sh 'broken-links-checker --path ./public --baseUrl https://www.opengeosys.org'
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -800,9 +800,7 @@ pipeline {
           }
           steps {
             dir('web') {
-              sh 'yarn'
-              sh 'hugo'
-              sh 'broken-links-checker --path ./public --baseUrl https://www.opengeosys.org'
+              sh 'urlchecker check --file-types .pandoc,.md --white-listed-files releases/\* --white-listed-urls https://jenkins.opengeosys.org/job/ufz/job/ogs-container-maker/job/master/build,http://yourproxy.example.com,https://apt.kitware.com/ubuntu/,https://github.com/YOUR-USERNAME/ogs,https://jenkins.opengeosys.org/github-webhook/,http://localhost:1313,https://github.com/ufz/ogs/pull/$1,http://www.opengeosys.org/images/xsd/OpenGeoSysXXX.xsd content'
```
TomFischer: As far as I can see all but one entry on the list is related to OGS. Why is 'http://yourproxy.example.com' listed here?

bilke: This url is given here: https://www.opengeosys.org/docs/devguide/getting-started/prerequisites/#step-install-git as an example on how to use git behind a web proxy.

PR #2880, opened 2020-03-26
Van Genuchten capillary pressure saturation

endJunction:
Mainly fixing the documentation. See individual commits for details.
Change input parameter `<entry_pressure>` to `<p_b>` and `<exponent>` to `<m>`.

### DIFF
--- a/MaterialLib/MPL/Properties/CapillaryPressureSaturation/SaturationVanGenuchten.h
+++ b/MaterialLib/MPL/Properties/CapillaryPressureSaturation/SaturationVanGenuchten.h
@@ -24,10 +24,13 @@ class Component;
  * \f[S_\text{eff}=\frac{S-S_r}{S_{\text{max}}-S_r}.\f]
  * Above, \f$S_r\f$ and \f$S_{\text{max}}\f$ are the residual and the maximum
  * saturations.
- * The exponent \f$m \in (0,1)\f$ and the parameter \f$p_b\f$ (it is equal to
- * \f$\rho g/\alpha\f$ in original publication) are given by the user.
+ * The exponent \f$m \in (0,1)\f$ and the pressure scaling parameter \f$p_b\f$
+ * (it is equal to \f$\rho g/\alpha\f$ in original publication) are given by the
+ * user.
+ * The scaling parameter \f$p_b\f$ is given in same units as pressure.
  *
- * Sometimes a formulation with \f$n = 1 / (1 - m) > 1\f$ is used.
+ * In the original work another exponent \f$n\f$ is used, but usually set to
+ * \f$n = 1 / (1 - m)\f$, and also in this implementation.
  *
  * The saturation is computed from the capillary pressure as follows:
  * \f[S(p_c)=
@@ -46,7 +49,7 @@ public:
     SaturationVanGenuchten(double const residual_liquid_saturation,
                            double const residual_gas_saturation,
                            double const exponent,
-                           double const entry_pressure);
+                           double const p_b);
 
     void setScale(
         std::variant<Medium*, Phase*, Component*> scale_pointer) override
--- a/MaterialLib/MPL/Properties/CapillaryPressureSaturation/SaturationVanGenuchten.h
+++ b/MaterialLib/MPL/Properties/CapillaryPressureSaturation/SaturationVanGenuchten.h
@@ -16,40 +16,29 @@ class Medium;
 class Phase;
 class Component;
 /**
- *   \brief The van Genuchten capillary pressure model:
+ * \brief The van Genuchten capillary pressure model.
  *
- *   \f[p_c(S)=p_b (S_e^{-1/m}-1)^{1-m}\f]
- *   with
- *   \f[S_e=\frac{S-S_r}{S_{\mbox{max}}-S_r}\f]
- *   where
- *    \f{eqnarray*}{
- *       &p_b&            \mbox{ entry pressure,}\\
- *       &S_r&            \mbox{ residual saturation,}\\
- *       &S_{\mbox{max}}& \mbox{ maximum saturation,}\\
- *       &m \in [0,1) &        \mbox{ exponent.}\\
- *    \f}
+ * The van Genuchten capillary pressure model (\cite Genuchten1980) is:
+ * \f[p_c(S)=p_b (S_\text{eff}^{-1/m}-1)^{1-m}\f]
+ * with effective saturation defined as
+ * \f[S_\text{eff}=\frac{S-S_r}{S_{\text{max}}-S_r}.\f]
+ * Above, \f$S_r\f$ and \f$S_{\text{max}}\f$ are the residual and the maximum
+ * saturations.
+ * The exponent \f$m \in (0,1)\f$ and the parameter \f$p_b\f$ (it is equal to
+ * \f$\rho g/\alpha\f$ in original publication) are given by the user.
  *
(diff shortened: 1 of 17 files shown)


wenqing:
I prefer to not change  `<entry_pressure>`  to  `<alpha>`, which is similar to 1/p_d. 

In "A closed-form equation for predicting the hydraulic conductivity of unsaturated soils", the primary variable is pressure head h, alpha is a factor in
`1/(1+(alpha h)^n)`. Which means  alpha is a pressure related parameter with an unit of [1/m]. If  `<entry_pressure>` is changed  to  `<alpha>`, `1/(1+(alpha h)^n)` has to be changed as `1/(1+(alpha    p/(rho g))^n)` because of p = rho g h.  

In the application, the entry pressure can be measured, the van Genuchten model can be therefore extended by changing `1/(1+(alpha    p/(rho g))^n)` to 1/(1+  p/p_d)^n)` with the consideration of p_d = rho * g/alpha. I think the current input of entry_pressure is physically meaningful, and it also feasible to input the alpha with its original meaning by a conversion of p_d=  rho * g/alpha beforehand.

MaterialLib/MPL/Properties/CapillaryPressureSaturation/SaturationVanGenuchten.h
```diff
@@ -16,48 +16,37 @@ class Medium;
...
- *       &m \in [0,1) &        \mbox{ exponent.}\\
- *    \f}
+ * The van Genuchten capillary pressure model (\cite Genuchten1980) is:
+ * \f[p_c(S)=\frac{1}{\alpha} (S_\text{eff}^{-1/m}-1)^{1-m}\f]
+ * with effective saturation defined as
+ * \f[S_\text{eff}=\frac{S-S_r}{S_{\text{max}}-S_r}.\f]
+ * Above, \f$S_r\f$ and \f$S_{\text{max}}\f$ are the residual and the maximum
+ * saturations.
+ * The exponent \f$m \in (0,1)\f$ and the parameter \f$\alpha\f$ are given by
+ * the user.
```
wenqing: Practically, the entry pressure \f$p_d\f$ is introduced to replace \f$ \frac{\rho_i g}{\alpha}\f$.  In the current implementation, the entry pressure \f$p_d\f$ is used.

MaterialLib/MPL/Properties/CapillaryPressureSaturation/SaturationVanGenuchten.h
```diff
@@ -16,48 +16,37 @@ class Medium;
...
- *       &S_r&            \mbox{ residual saturation,}\\
- *       &S_{\mbox{max}}& \mbox{ maximum saturation,}\\
- *       &m \in [0,1) &        \mbox{ exponent.}\\
- *    \f}
+ * The van Genuchten capillary pressure model (\cite Genuchten1980) is:
+ * \f[p_c(S)=\frac{1}{\alpha} (S_\text{eff}^{-1/m}-1)^{1-m}\f]
+ * with effective saturation defined as
+ * \f[S_\text{eff}=\frac{S-S_r}{S_{\text{max}}-S_r}.\f]
+ * Above, \f$S_r\f$ and \f$S_{\text{max}}\f$ are the residual and the maximum
+ * saturations.
```
wenqing:
```suggestion
 * saturations, and \f$\rho_i\f$ is the liquid density.
```

wenqing:
In my comment of yesterday in my PR #2858,  I added a link to the description of van Genuchten model in MOOSE. Since I found MOOSE uses alpha as a sort of 1/p_d, I removed the link.

Scinopode:
There is no such thing as entry-pressure in the van Genuchten model, since Brooks-Corey defined it as the pressure at which the largest pores start to drain. in vG-model, the pores drain immediately (starting from `s_L,eff < 1`), so per definition the entry pressure would be zero.
VanGenuchten never denominated `alpha` as inverse of entry pressure, nor did he relate both parameters. From my understanding, many people use `1/alpha=entry_pressure` to somehow compare BC and vG models, since both curves simply look similar at certain saturations if `alpha=1/p_b`. I also think vG made it clear that `alpha` is merely a fitting parameter for his curve without physical meaning.

From my perspective of poorly understanding this issue, I would suggest to call the parameter something like `<alpha_p>` to make clear that we expect a value of dimension `[(pressure)^(-1)]` rather than `[1/m]` or something like `<p_???>` of dimension `[pressure]`. But I would not call it `<entry_pressure>`, since this parameter exists in BC-Model only.

wenqing:
@Scinopode  For some engineering applications in rock mechanics,  the van Genuchten capillary pressure  model is modified with p_b=rho *g/alpha, which is called `apparent gas entry pressure`. For example in the technique note of the FE experiment, or [in a report by LBNL](http://www.ip-pamina.eu/downloads/pamina3.2.1.pdf):
pc
The same term appeared in the documentations of the DECOVALEX projects (some times `entry pressure` in short.)   

For soil mechanics,  the water density can be assumed as constant, input by either alpha or p_b is alpha is equivalent. For rock mechanics,  the water varies along with pressure and and temperature, and rho *g /alpha is normally treated as a constant parameter.

I am not sure which form of the parameter is mostly used, p_b or 1/p_b. The projects that I have conducted, the value of  that van Genuchten parameter is given by the term of `entry pressure` or `apparent air entry pressure`. There are also some others use alpha = 1/p_b, the term, I guess, was borrowed from the parameter of the original van Genuchten model with alpha= rho *g /p_b.


IMHO, I prefer not to change the p_b as alpha=1/p_b as proposed in this PR but give an adequate term for the parameter in the project file.

endJunction:
@wenqing And what would be your suggestion on how to name the quantity?

wenqing:
> @wenqing And what would be your suggestion on how to name the quantity?
How about `apparent_air_entry_pressure` ?

wenqing:
or just `p_b`

endJunction:
> or just `p_b`

As in the current status of the PR?

Tests/MaterialLib/TestMPLSaturationVanGenuchten.cpp
```diff
@@ -21,11 +21,11 @@ TEST(MaterialPropertyLib, SaturationVanGenuchten)
     double const residual_liquid_saturation = 0.1;
     double const residual_gas_saturation = 0.05;
     double const exponent = 0.79;
-    double const entry_pressure = 5000;
+    double const alpha = 1./5000;
```
wenqing, line 24: Can use p_b directly.

wenqing:
> > or just `p_b`
> 
> As in the current status of the PR?
> Then it can be merged.

Sorry I just replied the Norbert's comment but did not see your current change.

Tests/MaterialLib/TestMPLSaturationVanGenuchten.cpp
endJunction: OK. I'll check the implementation and use p_b directly everywhere.

endJunction:
@wenqing Just to check before another review and merge happens: Is the current implementation in this PR OK, or do we need further discussion?

wenqing:
@endJunction :fast_forward:

Thomas-TK:
since we may use HYDRUS as a reference for some model results here ist a link how they approach it: http://www.pc-progress.com/en/OnlineHelp/HYDRUS3/Hydrus.html?WaterFlowParameters.html. The wording people used depends on their diciplins. HYDRUS background is soil hydraulics

PR #2881, opened 2020-03-27
[CMake] ogs_add_library()-function

bilke:
Added CMake-function `ogs_add_library()` as a drop-in replacement for the `add_library()`-function.

The function takes care of compiler switches (currently warning level only), enabling PCH and installing shared libraries. This makes it easier for developers creating new libraries (just call the new function with your source file list and you are done).


bilke:
As usual:

```
ogs-2D_Adaptive_dt_ThermalConvection_constviscosityStaggeredScheme-time-vtkdiff (Failed)
```

PR #2883, opened 2020-03-28
[CL] Equilibrium-controlled reactions within PhreeqcKernel Interface

renchao-lu:
This pull request implements geochemical calculation for equilibrium-controlled reactions within the faster OGS#Phreeqc Interface.


ChemistryLib/CreateChemicalSolverInterface.cpp
```diff
@@ -186,9 +187,15 @@ createChemicalSolverInterface<ChemicalSolver::PhreeqcKernel>(
         //! \ogs_file_param{prj__chemical_system__rates}
         config.getConfigSubtreeOptional("rates"));
 
+    // pure phase assemblage
```
endJunction: Easier name would be collection. (If you do mean it and not "assembly".)

ChemistryLib/PhreeqcKernel.cpp
```diff
@@ -48,8 +50,24 @@ PhreeqcKernel::PhreeqcKernel(std::size_t const num_chemical_systems,
     }
     use.Set_solution_in(true);
 
+    // For equilibrium-controlled reactions
+    if (pure_phase_assem)
```
endJunction: Maybe `equilibrium_reactions` would be good name.
Anyway don't shorten the assemblage.

ChemistryLib/PhreeqcKernel.cpp
```diff
@@ -221,15 +271,49 @@ void PhreeqcKernel::execute(std::vector<GlobalVector*>& process_solutions)
...
+    }
+
+    // Pure phase assemblage
+    if (!Rxn_pp_assemblage_map.empty())
+    {
+        Rxn_new_pp_assemblage.clear();
+        for (int j = 0; j < count_unknowns; j++)
         {
-            Rxn_kinetics_map[chemical_system_id].Get_steps().clear();
+            if (x == NULL || x[j]->type != PP)
```
endJunction:
```suggestion
            if (x == nullptr || x[j]->type != PP)
```

ChemistryLib/PhreeqcKernelData/CreateInitialAqueousSolution.cpp
```diff
@@ -62,6 +62,17 @@ InitialAqueousSolution createInitialAqueousSolution(
             "for the transport process of hydrogen.");
     }
 
+    auto const means_of_adjusting_charge_in_str =
+        //! \ogs_file_param{prj__chemical_system__solution__means_of_adjusting_charge}
+        config.getConfigParameter<std::string>("means_of_adjusting_charge", "");
```
endJunction: I find "means of something" rather ugly.
Just an idea:
```xml
<charge_adjusting_mechanism>pH</charge_adjusting_mechanism>
```

ChemistryLib/PhreeqcKernelData/CreatePurePhaseAssemblage.cpp
```diff
@@ -0,0 +1,66 @@
+/**
+ * \file
+ * \copyright
+ * Copyright (c) 2012-2019, OpenGeoSys Community (http://www.opengeosys.org)
```
endJunction:
```suggestion
 * Copyright (c) 2012-2020, OpenGeoSys Community (http://www.opengeosys.org)
```

ChemistryLib/PhreeqcKernelData/PurePhaseAssemblage.cpp
```diff
@@ -0,0 +1,38 @@
...
+namespace PhreeqcKernelData
+{
+PurePhase::PurePhase(std::string&& name,
+                     double const initial_amount,
+                     double const saturation_index)
+{
+    this->name = name;
+    moles = initial_amount;
+    si = saturation_index;
+    si_org = saturation_index;
```
endJunction: should go into initializer list...

ChemistryLib/PhreeqcKernelData/CreateInitialAqueousSolution.cpp
renchao-lu: How about `<charge_balance>pH</charge_balance>`?

ChemistryLib/PhreeqcKernelData/PurePhaseAssemblage.cpp
renchao-lu: These parameters belong to the parent class `Phreeqc::cxxPPassemblageComp` and have been initialized in the parent class constructor. I reluctantly wrote in this way.

ChemistryLib/CreateChemicalSolverInterface.cpp
montoyav: The name comes from phreeqc, but we should make the nomenclature compatible with OGS.
What about "phase components"

renchao-lu: This object consists of not only `pure_phases` (`phase_components`) but also other members. That's why I feel object naming is too difficult...

montoyav: If you specify me which other members you are making reference, may I can help, total mass?

renchao-lu: What we call `pure_phase_assemblage` at this moment consists of the members listed below:
`new_def` -> a kind of flag;
`pp_assemblage_comps` -> `pure_phases` (`phase_components`);
`eltList`, -> list of elements in `pure_phases`;
`assemblage_totals` -> total moles of elements;
`vopts` -> ...

`pp_assemblage_comps` is a set of `pure_phase` which holds phase attributes such as:
`name` -> phase name;
`si` -> saturation index
`moles` and so on

renchao-lu: How about naming this assemblage as `EquilibirumPhases`. We call the member `pp_assemblage_comps` `phase_components`. @montoyav @endJunction

montoyav: OK, I understand, all the functionalities that the keyword EQUILIBRIUM_PHASES has in PHREEQC... let me think....

montoyav: I was having a look to all the lines, to make it consistent:
1) EquilibriumPhases as @renchao-lu suggested or
2) EquilibriumReactants

renchao-lu: So `EquilibriumReactants` for the entire object and `PhaseComponents` for the member `pp_assemblage_comps`.

ChemistryLib/PhreeqcKernelData/CreateInitialAqueousSolution.cpp
montoyav: "charge_balance" much more better

ChemistryLib/PhreeqcKernel.cpp
renchao-lu: Renamed.

ChemistryLib/PhreeqcKernelData/CreateInitialAqueousSolution.cpp
renchao-lu: Improved.

ChemistryLib/PhreeqcKernelData/PurePhaseAssemblage.cpp
endJunction: I see. Had a look on the particular class, it indeed has no usable ctors. :roll_eyes: 
Still, could you rewrite the "name" line as
```c++
    name = name_;
```
?

ChemistryLib/CreateChemicalSolverInterface.cpp
```diff
@@ -186,9 +187,15 @@ createChemicalSolverInterface<ChemicalSolver::PhreeqcKernel>(
         //! \ogs_file_param{prj__chemical_system__rates}
         config.getConfigSubtreeOptional("rates"));
 
+    // equilibrium reactants
+    auto equilibrium_reactants = PhreeqcKernelData::createEquilibriumReactants(
+        //! \ogs_file_param{prj__chemical_system__equilibrium_phases}
+        config.getConfigSubtreeOptional("equilibrium_phases"), mesh);
```
endJunction: Just a question. equilibrium phases are same as equilibrium reactants?

ChemistryLib/PhreeqcIOData/AqueousSolution.cpp
```diff
@@ -23,19 +23,19 @@ std::ostream& operator<<(std::ostream& os,
 
     os << "pressure " << aqueous_solution.pressure << "\n";
 
-    switch (aqueous_solution.means_of_adjusting_charge)
+    switch (aqueous_solution.charge_balance)
     {
-        case MeansOfAdjustingCharge::pH:
+        case ChargeBalance::pH:
```
endJunction: :+1:

ChemistryLib/PhreeqcKernelData/EquilibriumReactants.h
```diff
@@ -0,0 +1,60 @@
+/**
+ * \file
+ * \copyright
+ * Copyright (c) 2012-2019, OpenGeoSys Community (http://www.opengeosys.org)
```
endJunction: .

ChemistryLib/PhreeqcKernelData/EquilibriumReactants.cpp
```diff
@@ -0,0 +1,37 @@
...
+    si = saturation_index;
+    si_org = saturation_index;
+}
+
+EquilibriumReactants::EquilibriumReactants(
+    std::vector<PhaseComponent>& phase_components)
+{
+    for (auto& phase_component : phase_components)
+    {
+        auto name = phase_component.getName();
```
endJunction:
```suggestion
        auto& name = phase_component.getName();
```
Why?

ChemistryLib/PhreeqcKernelData/PurePhaseAssemblage.cpp
renchao-lu: changed.

ChemistryLib/CreateChemicalSolverInterface.cpp
renchao-lu: Absolutely.

ChemistryLib/PhreeqcKernelData/EquilibriumReactants.h
renchao-lu: Good catch! fixed.

ChemistryLib/PhreeqcKernelData/EquilibriumReactants.cpp
renchao-lu: Prevent copy, if my understanding is correct.

ChemistryLib/CreateChemicalSolverInterface.cpp
endJunction: Then I would suggest to use only one name, if possible. For me the equality is not obvious.

I understand, that this is one more rename through whole code. Maybe in a discussion a choice could be made before any actual renames...

ChemistryLib/PhreeqcKernelData/EquilibriumReactants.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+namespace ChemistryLib
+{
+namespace PhreeqcKernelData
+{
+PhaseComponent::PhaseComponent(std::string&& name_,
+                               double const initial_amount,
+                               double const saturation_index)
+{
+    name = name_;
```
endJunction: Minor: A move is need here. Not that it makes any difference in runtime, but, I think, if we use r-value references etc. it should be consistent 'til end.

ChemistryLib/CreateChemicalSolverInterface.cpp
renchao-lu: How about update the tag in the .prj file?

montoyav: For me both options are fine, (reactants or phases) but I agree with @endJunction that homogeneous nomenclature facilitate things, maybe to make it clear that we are doing "reactive" things I would suggest "equilibrium_reactants"

renchao-lu: I start to update the tag as suggested.

renchao-lu: What we define in the .prj file is `phase components`. So the data structure in the .prj file would look like either of the below
```
<equilibrium_reactants>
    <equilibrium_reactant>
        ...
    </equilibrium_reactant>
</equilibrium_reactants>
```

```
<equilibrium_reactants>
    <phase_component>
        ...
    <phase_component>
</equilibrium_reactants>
```
which data structure will we use?

endJunction: I like the first one.

renchao-lu: Note that all the dataset between the tag `<equilibrium_reactant>` and `</equilibrium_reactant>` will go into the built-in object `std::vector<PhaseComponent> phase_components`.

endJunction: I see, you have to stop somewhere. One step at a time, but keep the big picture in your head.
If we want this consistent on all layers, then there are few renames to do. Can be independent.
And also every thrive for more consistency will almost always break current semi-consistency; this is unavoidable.

For this PR I'd maybe stop at the `phase_component` as in your second option above.

renchao-lu: Let's take this compromise plan at the present. When we have better workaround to the naming issue, we polish it again.

ChemistryLib/PhreeqcKernelData/EquilibriumReactants.cpp
renchao-lu: Fixed.

ChemistryLib/CreateChemicalSolverInterface.cpp
renchao-lu: Done including Docu, tag replacement in prj files, renaming class member variables and functions in `ChemistryLib::PhreeqcIO` and `ChemistryLib::PhreeqcKernel`.

ChemistryLib/PhreeqcKernel.cpp
```diff
@@ -22,15 +22,17 @@ namespace ChemistryLib
...
-                             std::unique_ptr<Kinetics>
-                                 kinetic_reactants,
-                             std::vector<ReactionRate>&& reaction_rates)
+PhreeqcKernel::PhreeqcKernel(
+    std::size_t const num_chemical_systems,
+    std::vector<std::pair<int, std::string>> const&
+        process_id_to_component_name_map,
+    std::string const database,
+    AqueousSolution aqueous_solution,
+    std::unique_ptr<EquilibriumReactants>&& equilibrium_reactants,
```
endJunction, line 31: Not sure about r-value reference here. It is accessed in the function, but it is not moved. so it will be destroyed after the function finishes...

ChemistryLib/PhreeqcKernel.cpp
```diff
@@ -166,6 +206,11 @@ void PhreeqcKernel::setAqueousSolutions(
                 // Set pH value by hydrogen concentration.
                 double const pH = -std::log10(concentration);
                 aqueous_solution.Set_ph(pH);
+
+                if (components.find("H(1)") != components.end())
```
endJunction, line 210:
```suggestion
                if (auto it = components.find("H(1)"); it != components.end())
```
Then the second search is not needed.

ChemistryLib/PhreeqcKernelData/CreateInitialAqueousSolution.cpp
```diff
@@ -62,6 +62,17 @@ InitialAqueousSolution createInitialAqueousSolution(
             "for the transport process of hydrogen.");
     }
 
+    auto const charge_balance =
+        //! \ogs_file_param{prj__chemical_system__solution__charge_balance}
+        config.getConfigParameter<std::string>("charge_balance", "");
+
+    if (charge_balance == "pH")
```
endJunction, line 69: I recommend to check for other cases too. If empty, then OK, but if not empty it could be mistyped or from future ogs or somethings else, then OGS_FATAL.

ChemistryLib/PhreeqcKernelData/EquilibriumReactants.h
```diff
@@ -0,0 +1,60 @@
...
+public:
+    PhaseComponent(std::string&& name_, double const initial_amount,
+                   double const saturation_index);
+
+    cxxPPassemblageComp const* castToBaseClass() const
+    {
+        return static_cast<cxxPPassemblageComp const*>(this);
+    }
+
+    std::string const getName() const { return Get_name(); }
```
endJunction, line 34: The phreeqc's function is returning a copy or reference?

ChemistryLib/PhreeqcKernelData/CreateInitialAqueousSolution.cpp
renchao-lu: Creation function of charge balance is moved into the folder `Common`, shared with the both of Phreeqc Interface.

ChemistryLib/PhreeqcKernel.cpp
renchao-lu: I don't understand where the second search takes place.

ChemistryLib/PhreeqcKernel.cpp
renchao-lu: Isn't for passing the ownership of the object?

ChemistryLib/PhreeqcKernelData/EquilibriumReactants.h
renchao-lu: Good catch. Return a reference. Fixed.

ChemistryLib/PhreeqcKernel.cpp
endJunction: Yes, it is. `equilibrium_reactants` are now owned by the ctor. They are somehow used in the construction, which is fine.
But when the constructor ends, the unique_ptr will be destroyed. Is this intended behaviour?

ChemistryLib/PhreeqcKernel.cpp
endJunction: in the if-condition, there is `compontes["H(1)"]` this invokes the second search.

montoyav: In fact, I do not understand why this is needed in the code. If I understood correctly this lines will only give the pH value according to the H(1) concentration, but the pH value is a mandatory input parameter. (The pH value in phreeqc is given in log (activity) instead of log (concentration).

ChemistryLib/PhreeqcKernel.cpp
```diff
@@ -48,8 +50,24 @@ PhreeqcKernel::PhreeqcKernel(std::size_t const num_chemical_systems,
...
+    {
+        tidyEquilibriumReactants(*equilibrium_reactants);
+
+        for (std::size_t chemical_system_id = 0;
+             chemical_system_id < num_chemical_systems;
+             ++chemical_system_id)
+        {
+            equilibrium_reactants->setChemicalSystemID(chemical_system_id);
+            Rxn_pp_assemblage_map.emplace(
+                chemical_system_id, *equilibrium_reactants->castToBaseClass());
```
endJunction, line 64:
```suggestion
                chemical_system_id, *equilibrium_reactants.release()->castToBaseClass());
```

ChemistryLib/Common/CreateChargeBalance.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include "BaseLib/Error.h"
+#include "ChargeBalance.h"
+
+namespace ChemistryLib
+{
+ChargeBalance createChargeBalance(BaseLib::ConfigTree const& config)
+{
+    auto const charge_balance =
+        //! \ogs_file_param{prj__chemical_system__solution__charge_balance}
+        config.getConfigParameter<std::string>("charge_balance", "");
```
endJunction, line 22:
```suggestion
        config.getConfigParameterOptional<std::string>("charge_balance");
```

ChemistryLib/PhreeqcKernel.cpp
renchao-lu: With this line, element `H` is considered as a component for the purpose of charge balance.

ChemistryLib/Common/CreateChargeBalance.cpp
renchao-lu: Changed.

ChemistryLib/PhreeqcKernel.cpp
renchao-lu: Fixed.

ChemistryLib/PhreeqcKernel.cpp
renchao-lu: Fixed.

ChemistryLib/PhreeqcKernel.cpp
renchao-lu: Thanks for the instruction.

renchao-lu:
Cleaned up the commits made a bit.

ChemistryLib/PhreeqcKernel.cpp
```diff
@@ -166,6 +212,12 @@ void PhreeqcKernel::setAqueousSolutions(
                 // Set pH value by hydrogen concentration.
                 double const pH = -std::log10(concentration);
                 aqueous_solution.Set_ph(pH);
+
+                if (auto hydrogen = components.find("H(1)");
+                    hydrogen != components.end())
+                {
+                    hydrogen->second.Set_input_conc(pH);
+                }
```
endJunction, line 220: Sorry for you. I forgot we have an old compiler setup here; This feature is only c++17. Old style is:
```suggestion
                {
                    auto hydrogen = components.find("H(1)");
                    if (hydrogen != components.end())
                    {
                        hydrogen->second.Set_input_conc(pH);
                    }
                }
```

renchao-lu:
Rebased.

PR #2884, opened 2020-03-29
[Cmake] Remove ChemistryLib and iphreeqc make for HeatTransportBHE process

ChaofanChen:
As titled.

### DIFF
--- a/ProcessLib/CMakeLists.txt
+++ b/ProcessLib/CMakeLists.txt
@@ -30,9 +30,7 @@ target_link_libraries(ProcessLib
                       PRIVATE ParameterLib GitInfoLib)
 
 if (OGS_BUILD_PROCESS_ComponentTransport
-    OR OGS_BUILD_PROCESS_RichardsComponentTransport
-    OR OGS_BUILD_PROCESS_RichardsComponentTransport
-    OR OGS_BUILD_PROCESS_HeatTransportBHE)
+    OR OGS_BUILD_PROCESS_RichardsComponentTransport)
     target_link_libraries(ProcessLib PUBLIC ChemistryLib)
 endif()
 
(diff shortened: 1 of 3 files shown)


endJunction:
What's the reason?

ChaofanChen:
> What's the reason? Could you add an explanation to the commit message.
> What has been changed is visible by the changes themselves.

Sorry for that, what I am understanding is that before what I changed in this PR, if I want to build the HeatTransportBHE process, the ChemistryLib and iphreeqc will be compiled together and linked, but this BHE process is not related to the chemical thing. So I remove it from corresponding cmake lists. If my understanding is wrong, please let me know. Also, the explanation is added in the commit message.

ProcessLib/CMakeLists.txt
```diff
@@ -31,8 +31,7 @@ target_link_libraries(ProcessLib
 
 if (OGS_BUILD_PROCESS_ComponentTransport
     OR OGS_BUILD_PROCESS_RichardsComponentTransport
```
TomFischer: Why not remove one of the double lines `OR OGS_BUILD_PROCESS_RichardsComponentTransport`?

ThirdParty/CMakeLists.txt
```diff
@@ -11,8 +11,7 @@ option(OGS_USE_OPTIONAL_SUBMODULES "Option for enabling optional submodules" OFF
 # iphreeqc
 if (OGS_BUILD_PROCESS_ComponentTransport
     OR OGS_BUILD_PROCESS_RichardsComponentTransport
```
TomFischer: Why not
```suggestion
```

ProcessLib/CMakeLists.txt
ChaofanChen: Removed.

ThirdParty/CMakeLists.txt
ChaofanChen: Removed.

endJunction:
@ChaofanChen Thanks for the explanation.

PR #2886, opened 2020-03-30
GMSH2OGS: check mesh integration

### LINKED ISSUE #2841: GMSH2OGS surplus boundary elements (that result in det J=0)/element type statistics
One of the most frequent problems people have converting GMSH2 VTU is related to additional elements, I think. One way to address that issue we be to print out a statistics of the element types together with a warning if not all elements have the same dimension.
-e can already remove lines, but there can be also quads and triangles in 3D meshes.
The easiest way to address the problem are warnings. However, it would be even cooler if the OGS tools produce already valid OGS input files, i.e. forbidding mix dimensions, and correcting the node numbering/ordering.

joergbuchwald:
Perform additional mesh checks (e.g. for element type) while converting.
This should facilitate troubleshooting as GMSH meshes nomally don't work out of the box.

Fixes #2841

### DIFF
--- a/Applications/Utils/FileConverter/GMSH2OGS.cpp
+++ b/Applications/Utils/FileConverter/GMSH2OGS.cpp
@@ -117,15 +117,17 @@ int main (int argc, char* argv[])
         }
     }
     // *** print meshinformation
+
     INFO("Please check your mesh carefully!");
     INFO(
         "Degenerated or redundant mesh elements can cause OGS to stop or "
         "misbehave.");
     INFO("Use the -e option to delete redundant line elements.");
+
     // Geometric information
-    const GeoLib::AABB aabb(MeshLib::MeshInformation::getBoundingBox(*mesh));
-    auto minPt(aabb.getMinPoint());
-    auto maxPt(aabb.getMaxPoint());
+    const GeoLib::AABB aabb = MeshLib::MeshInformation::getBoundingBox(*mesh);
+    auto const minPt(aabb.getMinPoint());
+    auto const maxPt(aabb.getMaxPoint());
     INFO("Node coordinates:");
     INFO("\tx [%g, %g] (extent %g)", minPt[0], maxPt[0], maxPt[0] - minPt[0]);
     INFO("\ty [%g, %g] (extent %g)", minPt[1], maxPt[1], maxPt[1] - minPt[1]);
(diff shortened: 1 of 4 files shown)


Applications/Utils/FileConverter/GMSH2OGS.cpp
```diff
@@ -111,6 +116,99 @@ int main (int argc, char* argv[])
...
+        INFO("\t%d hexahedra", nr_ele_types[etype]);
+    }
+    if (nr_ele_types[++etype] > 0)
+    {
+        INFO("\t%d pyramids", nr_ele_types[etype]);
+    }
+    if (nr_ele_types[++etype] > 0)
+    {
+        INFO("\t%d prisms", nr_ele_types[etype]);
+    }
```
endJunction: something like a table would be easier to understand.
```c++
std::array element_names {"lines", ...};
for (int i = 0; i < nr_ele_types.size(); ++i)
{
    INFO("\t%d %s", nr_ele_types[i], element_names[i]);
}
```

Applications/Utils/FileConverter/GMSH2OGS.cpp
```diff
@@ -111,6 +116,99 @@ int main (int argc, char* argv[])
...
+    INFO("Node coordinates:");
+    INFO("\tx [%g, %g] (extent %g)", minPt[0], maxPt[0], maxPt[0] - minPt[0]);
+    INFO("\ty [%g, %g] (extent %g)", minPt[1], maxPt[1], maxPt[1] - minPt[1]);
+    INFO("\tz [%g, %g] (extent %g)", minPt[2], maxPt[2], maxPt[2] - minPt[2]);
+
+    INFO("Edge length: [%g, %g]", mesh->getMinEdgeLength(),
+         mesh->getMaxEdgeLength());
+
+    // Element information
+    const std::array<unsigned, 7> nr_ele_types(
```
endJunction: auto const&

Applications/Utils/FileConverter/GMSH2OGS.cpp
endJunction: @TomFischer This parts were copied from checkMesh.cpp. Where could we move this output routines?

Applications/Utils/FileConverter/GMSH2OGS.cpp
```diff
@@ -111,6 +116,99 @@ int main (int argc, char* argv[])
...
+    // *** print meshinformation
+    INFO("Please check your mesh carefully!");
+    INFO(
+        "Degenerated or redundant mesh elements can cause OGS to stop or "
+        "misbehave.");
+    INFO("Use the -e option to delete redundant line elements.");
+    // Geometric information
+    const GeoLib::AABB aabb(MeshLib::MeshInformation::getBoundingBox(*mesh));
+    auto minPt(aabb.getMinPoint());
+    auto maxPt(aabb.getMaxPoint());
```
TomFischer:
```suggestion
    auto const maxPt(aabb.getMaxPoint());
```

Applications/Utils/FileConverter/GMSH2OGS.cpp
```diff
@@ -111,6 +116,99 @@ int main (int argc, char* argv[])
...
     }
+    // *** print meshinformation
+    INFO("Please check your mesh carefully!");
+    INFO(
+        "Degenerated or redundant mesh elements can cause OGS to stop or "
+        "misbehave.");
+    INFO("Use the -e option to delete redundant line elements.");
+    // Geometric information
+    const GeoLib::AABB aabb(MeshLib::MeshInformation::getBoundingBox(*mesh));
+    auto minPt(aabb.getMinPoint());
```
TomFischer:
```suggestion
    auto const minPt(aabb.getMinPoint());
```

Applications/Utils/FileConverter/GMSH2OGS.cpp
TomFischer: The array returned by `MeshLib::MeshInformation::getNumberOfElementTypes` is only used for the output. Since this output is directly related to the `MeshInformation` I would put it in the class `MeshInformation`. Furthermore, the lines of code used for the output can reduced by a for-loop, as already suggested:
```c++
for (int element_type = 0; element_type < nr_ele_types.size();
         element_type++)
    {
        INFO("\t%d %s", nr_ele_types[element_type],
             MeshLib::MeshElemType2String(
                 static_cast<MeshLib::MeshElemType>(element_type)).c_str());
    }
```

Applications/Utils/FileConverter/GMSH2OGS.cpp
```diff
@@ -111,6 +116,99 @@ int main (int argc, char* argv[])
...
+        {
+            INFO("\t%s: [%g, %g]", vec_name.c_str(), vec_bounds->first,
+                 vec_bounds->second);
+        }
+        else
+        {
+            INFO("\t%s: Could not get value bounds for property vector.",
+                 vec_name.c_str());
+        }
+    }
```
TomFischer: Maybe lines 171-194 would be better in a function/method in class `MeshInformation`. I would denote it as `writePropertyVectorInformation()`. This function can be used here and in checkMesh.cpp.

Applications/Utils/FileConverter/GMSH2OGS.cpp
```diff
@@ -111,6 +116,99 @@ int main (int argc, char* argv[])
...
+
+    if (valid_arg.isSet())
+    {
+        // MeshValidation outputs error messages
+        // Remark: MeshValidation can modify the original mesh
+        MeshLib::MeshValidation validation(*mesh);
```
TomFischer: Lines 200-210 could be moved into a function like `writeMeshValidationResults()`.

joergbuchwald:
Thank you for your comments. I incorporated those things and added now also checkMesh to the pull request.

MeshLib/MeshInformation.cpp
```diff
@@ -63,4 +65,62 @@ std::array<unsigned, 7> MeshInformation::getNumberOfElementTypes(
     return n_element_types;
 }
 
+void MeshInformation::writeAllNumbersOfElementTypes(const MeshLib::Mesh& mesh)
+{
+    auto const& nr_ele_types =
+        MeshLib::MeshInformation::getNumberOfElementTypes(mesh);
+    const std::array<std::string, 7> element_names{
```
TomFischer, line 72: This mapping is already implemented in MeshLib/MeshEnums.h and the corresponding mapping can by achieved as follows:

```c++
for (int element_type = 0; element_type < nr_ele_types.size();
         element_type++)
    {
        INFO("\t%d %s", nr_ele_types[element_type],
             MeshLib::MeshElemType2String(
                 static_cast<MeshLib::MeshElemType>(element_type)).c_str());
    }
```

joergbuchwald: Well, I tried it, and the output was just 'none' so I rewrote this mapping.

joergbuchwald: I think the remaining question is, whether I should add a new mapping to MeshEnums or I should apply the VTK mapping to the array (meaning that we are wasting memory) or whether I should keep it as it is.

TomFischer: I don't understand why the element type to element name mapping is inconsistent. In what cases 'none' is returned? Which cases are handled wrong? Could you please provide a test example?

TomFischer: Just discussed with @joergbuchwald. The OGS-enum is inconsistent with the array provided by the method at hand. It is @joergbuchwald's decision whether to change the array or not.

joergbuchwald: I think I stick with it for the moment. 
When I will have some time, I will change it to the VTK mapping as defined in MeshEnums.h but in a separate PR.

TomFischer:
It is the usual benchmark that fails in the docker test. It isn't related to this PR. If @endJunction is fine with the PR it can be merged.

MeshLib/MeshInformation.h
```diff
@@ -70,6 +70,16 @@ class MeshInformation
...
         const MeshLib::Mesh& mesh);
+
+    /// writes all numbers of element types
+    static void writeAllNumbersOfElementTypes(const MeshLib::Mesh& mesh);
+
+    /// writes out property vector information
+    static void writePropertyVectorInformation(const MeshLib::Mesh& mesh);
+
+    /// writes out mesh validation results
+    // Remark: MeshValidation can modify the original mesh
```
endJunction, line 81:
```suggestion
    /// Remark: MeshValidation can modify the original mesh
```
Otherwise the comment above this one goes to file-documentation and the function is not documented.

PR #2888, opened 2020-03-31
Test file name check

endJunction:
TEST ONLY. Don't merge.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,32 +11,11 @@ repos:
     -   id: check-yaml
 -   repo: local
     hooks:
-    -   id: git-diff-check
-        name: Run git diff --check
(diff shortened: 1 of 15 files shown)


endJunction:
@bilke I think the command line becomes too long at some point.

I tested some permutations of the commits and the first three commits in any order (few variations were tried out, not all) are OK, adding one more on top leads to failure.
If my suspicion is correct, adding more files (maybe with long names) will cause the problem too. Let's try!

endJunction:
Nope. That wasn't it. added almost 7k chars worth of new file names. In the original PR there are only around 2k.

endJunction:
I've no further ideas so far. Feel free to reopen, as needed.

wenqing:
@endJunction Thanks a lot for your efforts!

bilke:
@endJunction Thanks Dima for testing!

PR #2875, opened 2020-03-24
spdlog

endJunction:
Replacing logog logger with modern and powerfull spdlog logger (backed up by the fmt formatting library!).
The most obvious advantage is the type safety (coming from the fmt library). (Indeed I've found two or three places, where the garbage was output.)

1. [ ] Feature description was added to the [changelog](https://github.com/ufz/ogs/wiki/Release-notes-6.3.1)
2. [x] Tests covering your feature were added? (`ogs` :grin: )
3. [ ] Any new feature or behavior change was documented?

- [ ] Linking, not header only.

PR #2891, opened 2020-04-01
Remove compile time log level

endJunction:
the log level can be adjusted in preprocessor. The default values are "debug" for all builds.
If someone actively uses/relies on this feature, now is good point to tell it ;)
This feature is removed in this pull request.

Reducing complexity a little.
Part of #2875

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -150,19 +150,6 @@ endif()
 
 option(OGS_USE_MKL "Use Intel MKL" OFF)
 
-# Logging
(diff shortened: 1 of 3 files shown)


TomFischer:
Is it still possible to enable debug output in the release build (-l debug)? I use this feature often.

endJunction:
@TomFischer Yes, `-l debug` is a runtime feature. Here only compilation is affected. Default is "debug" anyway.

PR #2895, opened 2020-04-02
[SD] Principal stress output

nagelt:
Adds principal stress and principal stress direction output to the SmallDeformation process.
The implementation is in principle the same as the one used in the HydroMechanics process. The hollow sphere test is likewise the same.


nagelt:
@FZill Maybe you can have a look, since you did the HM version.

ProcessLib/SmallDeformation/SmallDeformationFEM.h
```diff
@@ -493,6 +493,50 @@ class SmallDeformationLocalAssembler
...
+        x_position.setElementID(elem_id);
+        unsigned const n_integration_points =
+            _integration_method.getNumberOfPoints();
+
+        constexpr int symmetric_tensor_size =
+            MathLib::KelvinVector::KelvinVectorDimensions<
+                DisplacementDim>::value;
+
+        auto sigma_sum = MathLib::KelvinVector::tensorToKelvin<DisplacementDim>(
+            Eigen::Matrix<double, 3, 3>::Zero());
```
TomFischer, line 510: Is there a relation between the DisplacementDim the fixed size 3 of the matrix?

nagelt: not in this case. We're always dealing with 3 normal stresses, from which we calculate the principal values.

TomFischer: Thanks for answer.

FZill:
Looks good to me. Regarding the output to be optional - If we do that, we should do it in HM too.

nagelt:
> Looks good to me. Regarding the output to be optional - If we do that, we should do it in HM too.

I think so, too. Maybe this could be a follow-up PR that makes this feature in both processes optional.

wenqing:
> > Looks good to me. Regarding the output to be optional - If we do that, we should do it in HM too.
> 
> I think so, too. Maybe this could be a follow-up PR that makes this feature in both processes optional.

That is good.

PR #2899, opened 2020-04-06
Elemtypes map

joergbuchwald:
Substitute getNumberOfElementTypes() array by std::map to avoid multiple different element type mappings.

I'm not able to check the output of the DataExplorer right now as the binary makes some throws some OpenGL/QT errors on my machine. But at least it compiles.

### DIFF
--- a/MeshLib/MeshInformation.cpp
+++ b/MeshLib/MeshInformation.cpp
@@ -25,42 +25,15 @@ GeoLib::AABB MeshInformation::getBoundingBox(const MeshLib::Mesh& mesh)
     return GeoLib::AABB(nodes.begin(), nodes.end());
 }
 
-std::array<unsigned, 7> MeshInformation::getNumberOfElementTypes(
+std::map<MeshElemType, unsigned> MeshInformation::getNumberOfElementTypes(
     const MeshLib::Mesh& mesh)
 {
-    std::array<unsigned, 7> n_element_types = {{0, 0, 0, 0, 0, 0, 0}};
+    std::map<MeshElemType, unsigned> n_element_types;
     const std::vector<MeshLib::Element*>& elements(mesh.getElements());
     for (auto element : elements)
     {
         MeshElemType t = element->getGeomType();
-        if (t == MeshElemType::LINE)
-        {
-            n_element_types[0]++;
-        }
-        if (t == MeshElemType::TRIANGLE)
-        {
-            n_element_types[1]++;
-        }
-        if (t == MeshElemType::QUAD)
-        {
-            n_element_types[2]++;
-        }
-        if (t == MeshElemType::TETRAHEDRON)
-        {
-            n_element_types[3]++;
-        }
-        if (t == MeshElemType::HEXAHEDRON)
-        {
-            n_element_types[4]++;
-        }
-        if (t == MeshElemType::PYRAMID)
-        {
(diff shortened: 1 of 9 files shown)


Tests/MeshLib/TestAddLayerToMesh.cpp
```diff
@@ -70,9 +70,9 @@ TEST(MeshLib, AddTopLayerToLineMesh)
     ASSERT_EQ(2*mesh->getNumberOfNodes(), result->getNumberOfNodes());
     ASSERT_EQ(2*mesh->getNumberOfElements(), result->getNumberOfElements());
 
-    std::array<unsigned, 7> const n_elems (MeshLib::MeshInformation::getNumberOfElementTypes(*result));
-    ASSERT_EQ(5, n_elems[0]); // tests if 5 lines are present
-    ASSERT_EQ(5, n_elems[2]); // tests if 5 quads are present
+    auto n_elems = MeshLib::MeshInformation::getNumberOfElementTypes(*result);
```
endJunction, line 73: keep const

MeshLib/MeshInformation.cpp
```diff
@@ -25,58 +24,28 @@ GeoLib::AABB MeshInformation::getBoundingBox(const MeshLib::Mesh& mesh)
...
 void MeshInformation::writeAllNumbersOfElementTypes(const MeshLib::Mesh& mesh)
 {
-    auto const& nr_ele_types =
-        MeshLib::MeshInformation::getNumberOfElementTypes(mesh);
-    const std::array<std::string, 7> element_names{
-        "lines",     "triangles", "quads", "tetrahedra",
-        "hexahedra", "pyramids",  "prisms"};
-    for (unsigned int element_type = 0; element_type < nr_ele_types.size();
-         element_type++)
+    auto nr_ele_types = MeshLib::MeshInformation::getNumberOfElementTypes(mesh);
```
endJunction, line 43: const&

MeshLib/MeshInformation.cpp
```diff
@@ -25,58 +24,28 @@ GeoLib::AABB MeshInformation::getBoundingBox(const MeshLib::Mesh& mesh)
...
+
+    INFO("Number of elements in the mesh:")
+    for (std::map<MeshElemType, unsigned>::iterator it=nr_ele_types.begin(); it!=nr_ele_types.end(); ++it)
```
endJunction, line 46: formatting

endJunction: Better yet to use the for-range-loop.

MeshLib/MeshInformation.cpp
```diff
@@ -25,58 +24,28 @@ GeoLib::AABB MeshInformation::getBoundingBox(const MeshLib::Mesh& mesh)
...
     {
-        INFO("\t%d %s ", nr_ele_types[element_type],
-             element_names[element_type].c_str());
+        INFO("\t%ss: %d ", MeshLib::MeshElemType2String(static_cast<MeshLib::MeshElemType>(it->first)).c_str(), it->second);
```
endJunction, line 48: Is the double s correct? `%ss`

joergbuchwald: That's correct. Because the casting generates singular typenames, but I want to use plural here.

PR #2901, opened 2020-04-14
Add bulk element mappings for triangles.

TomFischer:
         Yes, I created a pdf. The question is where to put the sketches?

### DIFF
--- a/MeshLib/Elements/MapBulkElementPoint.cpp
+++ b/MeshLib/Elements/MapBulkElementPoint.cpp
@@ -14,6 +14,25 @@
 
 namespace MeshLib
 {
+MathLib::Point3d getBulkElementPoint(MeshLib::Tri const& /*tri*/,
+                                     std::size_t const face_id,
+                                     MathLib::WeightedPoint1D const& wp)
+{
+    switch (face_id)
+    {
+        case 0:
+            return MathLib::Point3d{std::array<double, 3>{{wp[0], 0.0, 0.0}}};
+        case 1:
+            return MathLib::Point3d{
+                std::array<double, 3>{{1 - wp[0], wp[0], 0.0}}};
+        case 2:
+            return MathLib::Point3d{
+                std::array<double, 3>{{0.0, 1 - wp[0], 0.0}}};
+        default:
+            OGS_FATAL("Invalid face id '{:d}' for the tri.", face_id);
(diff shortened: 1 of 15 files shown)


MeshLib/Elements/MapBulkElementPoint.h
```diff
@@ -17,6 +17,18 @@
 
 namespace MeshLib
 {
+/// Maps the given lower dimensional boundary point \c wp of a line, i.e. the 1d
+/// gauss point given in local coordinates of a line, to higher dimensional
+/// point of the tri face (defined by the tri element and the face id) also in
+/// local coordinates of the tri face.
+/// \param tri the tri element
+/// \param face_id the id of the tri face the point will be mapped on
+/// \param wp the gauss point of the lower dimensional element
```
endJunction, line 26: Minor; Would call it integration point at least.
Better would be to call it something like local coordinate of the line element. (and then to pass only the coordinates w/o the integration weight.) But this is a long story...

MeshLib/Elements/MapBulkElementPoint.cpp
```diff
@@ -14,6 +14,25 @@
...
+        case 0:
+            return MathLib::Point3d{std::array<double, 3>{{wp[0], 0.0, 0.0}}};
+        case 1:
+            return MathLib::Point3d{
+                std::array<double, 3>{{1 - wp[0], wp[0], 0.0}}};
+        case 2:
+            return MathLib::Point3d{
+                std::array<double, 3>{{0.0, 1 - wp[0], 0.0}}};
+        default:
+            OGS_FATAL("Invalid face id '%u' for the tri.", face_id);
```
wenqing, line 32:
```suggestion
            OGS_FATAL("Invalid face id '{:d}' for the tri.", face_id);
```

TomFischer: Yes, you are right. Will fix it tomorrow.

TomFischer:
I don't see the reason for the doxygen warning:
```
argument 'quad' of command @param is not found in the argument list of MeshLib::getBulkElementPoint(MeshLib::Tri const &tri, std::size_t const face_id, MathLib::WeightedPoint1D const ℘)
```
Seems to be a false positive?

PR #2903, opened 2020-04-15
Add prism element type to structure mesh generator tool.

TomFischer:
This allows to generate tests for the prism element.

  The feature is tested already in `Tests/MeshLib/TestMeshGenerator.cpp`
  Command line help is updated.

### DIFF
--- a/Applications/Utils/SimpleMeshCreation/generateStructuredMesh.cpp
+++ b/Applications/Utils/SimpleMeshCreation/generateStructuredMesh.cpp
@@ -77,10 +77,13 @@ int main (int argc, char* argv[])
     allowed_ele_types.emplace_back("tri");
     allowed_ele_types.emplace_back("quad");
     allowed_ele_types.emplace_back("hex");
+    allowed_ele_types.emplace_back("prism");
     allowed_ele_types.emplace_back("tet");
     TCLAP::ValuesConstraint<std::string> allowedVals(allowed_ele_types);
-    TCLAP::ValueArg<std::string> eleTypeArg("e", "element-type",
-                                          "element type to be created: line | tri | quad | hex | tet", true, "line", &allowedVals);
+    TCLAP::ValueArg<std::string> eleTypeArg(
+        "e", "element-type",
+        "element type to be created: line | tri | quad | hex | prism | tet",
+        true, "line", &allowedVals);
     cmd.add(eleTypeArg);
     TCLAP::ValueArg<std::string> mesh_out("o", "mesh-output-file",
                                           "the name of the file the mesh will be written to", true,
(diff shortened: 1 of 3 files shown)


MeshLib/MeshEnums.cpp
```diff
@@ -91,35 +93,35 @@ std::string MeshElemType2StringShort(const MeshElemType t)
...
     {
         return MeshElemType::TRIANGLE;
     }
-    if (s == "tet" || s == "Tetrahedron")
+    if (boost::iequals(s, "tet") || boost::iequals(s, "Tetrahedron"))
     {
         return MeshElemType::TETRAHEDRON;
     }
-    if (s == "pris" || s == "Prism")
+    if (boost::iequals(s, "pris") || boost::iequals(s, "prism"))
```
endJunction, line 120:
```suggestion
    if (boost::iequals(s, "pris") || boost::iequals(s, "Prism"))
```
or all lower-case...

Applications/Utils/SimpleMeshCreation/generateStructuredMesh.cpp
```diff
@@ -221,6 +224,11 @@ int main (int argc, char* argv[])
     case MeshLib::MeshElemType::HEXAHEDRON:
         mesh.reset(MeshLib::MeshGenerator::generateRegularHexMesh(*vec_div[0], *vec_div[1], *vec_div[2]));
         break;
+    case MeshLib::MeshElemType::PRISM:
+        mesh.reset(MeshLib::MeshGenerator::generateRegularPrismMesh(
+            n_subdivision[0], n_subdivision[1], n_subdivision[2], vec_dx[0],
+            vec_dx[1], vec_dx[2]));
```
endJunction, line 230: Why is `vec_dx` passed here and `vec_div` in other calls?

TomFischer: There doesn't exist a corresponding interface in `MeshLib/MeshGenerators/MeshGenerator.h`.

endJunction: https://github.com/ufz/ogs/blob/master/MeshLib/MeshGenerators/MeshGenerator.h#L389-L396

TomFischer: Okay, you are right. There is an interface that takes the length and the number of subdivisions.

The other generator functions take `BaseLib::ISubdivision` as a parameter. An overload for the generateRegularPrismMesh that takes `BaseLib::ISubdivision` doesn't exist. So I have chosen the above function without a special reason. I'll change it to the suggested function to be a bit more consistent to the other generator functions.

bilke:
@TomFischer  ThirdParty/cmake-modules was changed (accidentally I guess...).

TomFischer:
> @TomFischer ThirdParty/cmake-modules was changed (accidentally I guess...).

Yes, this was an accident. I'll fix it.

PR #2905, opened 2020-04-16
Clang-tidy cleanups.

TomFischer:
Cleanups in BaseLib, Applications/FileIO, and MeshLib.
Fixed warnings are
- google-readability-braces-around-statements
- cppcoreguidelines-pro-type-member-init

### DIFF
--- a/Applications/FileIO/CsvInterface.cpp
+++ b/Applications/FileIO/CsvInterface.cpp
@@ -88,7 +88,7 @@ int CsvInterface::readPoints(std::string const& fname, char delim,
             continue;
         }
         it = fields.begin();
-        std::array<double, 3> point;
+        std::array<double, 3> point{};
         try {
             point[0] = std::stod(*it);
             point[1] = std::stod(*(++it));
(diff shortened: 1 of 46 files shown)


MathLib/Vector3.h
```diff
@@ -52,7 +52,7 @@ class TemplateVector3 : public MathLib::TemplatePoint<T>
     /**
      * Construct Vector3 from TemplatePoint.
      */
-    TemplateVector3(TemplatePoint<T,3> const& p) :
+    explicit TemplateVector3(TemplatePoint<T,3> const& p) :
```
endJunction, line 55: If I remember correctly marking this ctor `explicit` required more changes in the code.

TomFischer: No, I don't.

TomFischer: Maybe this is the reason why most of the tests are failing. Will revert the change.

endJunction: Drop the change. It's easier to switch to Eigen::Vector3d.

PR #2909, opened 2020-04-20
Create regular mesh consisting of pyramid elements.

TomFischer:

### DIFF
--- a/Applications/Utils/SimpleMeshCreation/generateStructuredMesh.cpp
+++ b/Applications/Utils/SimpleMeshCreation/generateStructuredMesh.cpp
@@ -80,10 +80,12 @@ int main (int argc, char* argv[])
     allowed_ele_types.emplace_back("hex");
     allowed_ele_types.emplace_back("prism");
     allowed_ele_types.emplace_back("tet");
+    allowed_ele_types.emplace_back("pyramid");
     TCLAP::ValuesConstraint<std::string> allowedVals(allowed_ele_types);
     TCLAP::ValueArg<std::string> eleTypeArg(
         "e", "element-type",
-        "element type to be created: line | tri | quad | hex | prism | tet",
+        "element type to be created: line | tri | quad | hex | prism | tet | "
+        "pyramid",
         true, "line", &allowedVals);
     cmd.add(eleTypeArg);
     TCLAP::ValueArg<std::string> mesh_out("o", "mesh-output-file",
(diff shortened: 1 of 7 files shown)


Applications/Utils/SimpleMeshCreation/generateStructuredMesh.cpp
```diff
@@ -80,10 +80,12 @@ int main (int argc, char* argv[])
     allowed_ele_types.emplace_back("hex");
     allowed_ele_types.emplace_back("prism");
     allowed_ele_types.emplace_back("tet");
+    allowed_ele_types.emplace_back("Pyramid");
     TCLAP::ValuesConstraint<std::string> allowedVals(allowed_ele_types);
     TCLAP::ValueArg<std::string> eleTypeArg(
         "e", "element-type",
-        "element type to be created: line | tri | quad | hex | prism | tet",
+        "element type to be created: line | tri | quad | hex | prism | tet | "
+        "Pyramid",
```
endJunction, line 88:
```suggestion
        "pyramid",
```

MeshLib/MeshGenerators/MeshGenerator.cpp
```diff
@@ -113,6 +114,35 @@ std::vector<MeshLib::Node*> MeshGenerator::generateRegularNodes(
...
+    std::vector<double> const& x_coords,
+    std::vector<double> const& y_coords,
+    std::vector<double> const& z_coords,
+    const MathLib::Point3d& origin)
+{
+    std::vector<Node*> nodes;
+    nodes.reserve((x_coords.size() - 1) * (y_coords.size() - 1) *
+                  (z_coords.size() - 1));
+
+    for (std::size_t i = 0; i < z_coords.size() - 1; i++)
```
endJunction, line 129: minor: storing the sizes once is usually faster than asking the vector for its size on every iteration.

TomFischer: Maybe the compiler can optimize it because the vector is const.

Applications/Utils/SimpleMeshCreation/generateStructuredMesh.cpp
TomFischer: Fixed.

MeshLib/MeshGenerators/MeshGenerator.cpp
endJunction: No. It is const in this function, but might not be const elsewhere.

TomFischer: Yes, you are right.

PR #2910, opened 2020-04-20
add GMRES restart option

joergbuchwald:
add restart option for the GMRES solver (Eigen-unsupported). 
It sets the number of iterations before a restart is conducted. It can improve the convergence behavior. However, increasing the value comes with a tradeoff in time and memory. First tests suggest that the default value (30) is already quite optimal in most cases.

### DIFF
--- a/MathLib/LinAlg/Eigen/EigenLinearSolver.cpp
+++ b/MathLib/LinAlg/Eigen/EigenLinearSolver.cpp
@@ -124,16 +124,20 @@ private:
     }
 };
 
-template <> 
-void EigenIterativeLinearSolver<Eigen::GMRES<EigenMatrix::RawMatrixType, Eigen::IdentityPreconditioner>>::setRestart(int const restart)
+template <>
+void EigenIterativeLinearSolver<
+    Eigen::GMRES<EigenMatrix::RawMatrixType,
+                 Eigen::IdentityPreconditioner>>::setRestart(int const restart)
 {
     _solver.set_restart(restart);
     INFO("-> set restart value: {:d}", _solver.get_restart());
 }
 
-template <> 
-void EigenIterativeLinearSolver<Eigen::GMRES<EigenMatrix::RawMatrixType, Eigen::DiagonalPreconditioner<double>>>::setRestart(int const restart)
-{   
+template <>
+void EigenIterativeLinearSolver<Eigen::GMRES<
+    EigenMatrix::RawMatrixType,
+    Eigen::DiagonalPreconditioner<double>>>::setRestart(int const restart)
+{
     _solver.set_restart(restart);
     INFO("-> set restart value: {:d}", _solver.get_restart());
 }
@@ -297,7 +301,7 @@ void EigenLinearSolver::setOption(BaseLib::ConfigTree const& option)
 #else
         OGS_FATAL(
             "The code is not compiled with the Eigen unsupported modules. "
-            "GMRES option restart is not available.");
+            "GMRES/GMRES option restart is not available.");
 #endif
     }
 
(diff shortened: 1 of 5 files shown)


MathLib/LinAlg/Eigen/EigenOption.h
```diff
@@ -47,6 +47,7 @@ struct EigenOption final
 #ifdef USE_EIGEN_UNSUPPORTED
     /// Scaling the coefficient matrix and the RHS bector
     bool scaling;
+    int restart;
```
endJunction, line 50: Only used by GMRES, isn't it? If so, add a comment.

MathLib/LinAlg/Eigen/EigenLinearSolver.cpp
```diff
@@ -265,6 +293,18 @@ void EigenLinearSolver::setOption(BaseLib::ConfigTree const& option)
             "scaling is not available.");
 #endif
     }
+    if (auto restart =
+            //! \ogs_file_param{prj__linear_solvers__linear_solver__eigen__restart}
```
endJunction, line 297: tag documentation is required. Preferably refer to the documentation of the `EigenOption::restart` member.

MathLib/LinAlg/Eigen/EigenLinearSolver.cpp
```diff
@@ -119,8 +120,35 @@ class EigenIterativeLinearSolver final : public EigenLinearSolverBase
...
+    _solver.set_restart(restart);
+    INFO("-> set restart value: {:d}", _solver.get_restart());
+}
+
+template <>
+void EigenIterativeLinearSolver<Eigen::GMRES<EigenMatrix::RawMatrixType, Eigen::IncompleteLUT<double>>>::setRestart(int const restart)
+{
+    _solver.set_restart(restart);
+    INFO("-> set restart value: {:d}", _solver.get_restart());
+}
```
TomFischer, line 150: Why is the setting of the `restart` value only possible for `IdentityPreconditioner`, `DiagonalPreconditioner`, and `IncompleteLUT`? The restart should be independent from the preconditioner at all.

endJunction: Not sure if the partial specialization will work here.
@joergbuchwald Please try it.

joergbuchwald: That's true. The problem I had, was with the T_SOLVER type. It is not instantiated via a general preconditioner type but via the specific preconditioners. Therefore, I had to make the code specialization for each preconditioner.

endJunction: Something like:
```c++
template <typename Preconditioner>
￼void EigenIterativeLinearSolver<Eigen::GMRES<EigenMatrix::RawMatrixType, Preconditioner>>::setRestart(int const restart)
```

MathLib/LinAlg/Eigen/EigenLinearSolver.cpp
```diff
@@ -119,8 +120,36 @@ class EigenIterativeLinearSolver final : public EigenLinearSolverBase
...
+void EigenIterativeLinearSolver<Eigen::GMRES<
+    EigenMatrix::RawMatrixType,
+    Eigen::DiagonalPreconditioner<double>>>::setRestart(int const restart)
+{
+    _solver.set_restart(restart);
+    INFO("-> set restart value: {:d}", _solver.get_restart());
+}
+
+template <>
+void EigenIterativeLinearSolver<Eigen::GMRES<EigenMatrix::RawMatrixType, Eigen::IncompleteLUT<double>>>::setRestart(int const restart)
```
endJunction, line 147: @joergbuchwald clang-format. Use pre-commit checks.

endJunction:
@joergbuchwald Small testcase using the option, something from SteadyStateDiffusion, for example, would be good.

PR #2902, opened 2020-04-15
Class of the van Genuchten relative permeability of the non-wetting phase 

wenqing:
As titled, the PR presents the van Genuchten Maulem model for the relative permeability of wetting phase.

Besides,  this PR allows use  `relative_permeability` as  phase property (see issue  #2904).

1. [ ] Feature description was added to the [changelog](https://github.com/ufz/ogs/wiki/Release-notes-6.3.1)? not yet
2. [x] Tests covering your feature were added? Yes, there is an unit test.
k_rel_vanGenuchtenMaulem

3. [x] Any new feature or behavior change was documented? Yes, it was.

PR #2919, opened 2020-04-24
Capillary pressure - van Genuchten model.

endJunction:
An implementation of capillary pressure calculation from liquid saturation. This is not an inverse of the
SaturationVanGenuchten model, because the resulting capillary pressure is cut off at some maximum value.

### DIFF
--- a/Documentation/ProjectFile/properties/property/CapillaryPressureVanGenuchten/t_residual_gas_saturation.md
+++ b/Documentation/ProjectFile/properties/property/CapillaryPressureVanGenuchten/t_residual_gas_saturation.md
@@ -1,0 +1,3 @@
+The smallest degree of saturation of the gas (or non-wetting) phase in the medium.
+The maximum saturation of the liquid (wetting) phase is
+\f$S_{L,\text{res}} = 1-S_{G,\text{res}}\f$.
--- a/MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
+++ b/MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
@@ -1,0 +1,105 @@
+/**
+ * \file
+ * \copyright
+ * Copyright (c) 2012-2020, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ * Created on March 20, 2020, 9:59 AM
+ */
+
+#include "CapillaryPressureVanGenuchten.h"
+
+#include <algorithm>
+#include <cmath>
+
+#include "MaterialLib/MPL/Medium.h"
+
+namespace MaterialPropertyLib
+{
+CapillaryPressureVanGenuchten::CapillaryPressureVanGenuchten(
+    double const residual_liquid_saturation,
+    double const maximum_liquid_saturation, double const exponent,
+    double const p_b, double const max_capillary_pressure)
+    : _residual_saturation(residual_liquid_saturation),
+      _maximuml_saturation(maximum_liquid_saturation),
+      _m(exponent),
+      _p_b(p_b),
+      _pc_max(max_capillary_pressure)
+{
+    if (!(_m > 0 && _m < 1))
+    {
+        OGS_FATAL(
+            "The exponent value m = %g of van Genuchten saturation model, is "
+            "out of its range of (0, 1)",
+            _m);
+    }
+}
+
+PropertyDataType CapillaryPressureVanGenuchten::value(
+    VariableArray const& variable_array,
+    ParameterLib::SpatialPosition const& /*pos*/, double const /*t*/,
+    double const /*dt*/) const
+{
+    const double saturation = std::get<double>(
+        variable_array[static_cast<int>(Variable::liquid_saturation)]);
+
+    const double S =
+        std::clamp(saturation, _residual_saturation, _maximuml_saturation);
+    const double Se = (S - _residual_saturation) /
+                      (_maximuml_saturation - _residual_saturation);
+    const double pc =
+        (Se < 1.0) ? _p_b * std::pow(std::pow(Se, (-1.0 / _m)) - 1.0, 1.0 - _m)
+                   : 0.0;
+    return std::clamp(pc, 0.0, _pc_max);
+}
+
+PropertyDataType CapillaryPressureVanGenuchten::dValue(
+    VariableArray const& variable_array, Variable const /*primary_variable*/,
+    ParameterLib::SpatialPosition const& /*pos*/, double const /*t*/,
+    double const /*dt*/) const
+{
+    const double saturation = std::get<double>(
+        variable_array[static_cast<int>(Variable::liquid_saturation)]);
+
+    const double S =
+        std::clamp(saturation, _residual_saturation, _maximuml_saturation);
+    const double Se = (S - _residual_saturation) /
+                      (_maximuml_saturation - _residual_saturation);
+
+    if (!(Se < 1.0))
+    {
+        return 0.0;
+    }
+    const double val1 = std::pow(Se, -1.0 / _m);
+    const double val2 = std::pow(val1 - 1.0, -_m);
+    return _p_b * (_m - 1.0) * val1 * val2 / (_m * (S - _residual_saturation));
+}
+
+PropertyDataType CapillaryPressureVanGenuchten::d2Value(
+    VariableArray const& variable_array, Variable const /*primary_variable1*/,
+    Variable const /*primary_variable2*/,
+    ParameterLib::SpatialPosition const& /*pos*/, double const /*t*/,
+    double const /*dt*/) const
+{
+    const double saturation = std::get<double>(
+        variable_array[static_cast<int>(Variable::liquid_saturation)]);
+
+    const double S =
+        std::clamp(saturation, _residual_saturation, _maximuml_saturation);
+    const double Se = (S - _residual_saturation) /
+                      (_maximuml_saturation - _residual_saturation);
+    if (!(Se < 1.0))
+    {
+        return 0.0;
+    }
+
+    const double val1 = std::pow(Se, 1.0 / _m);
+    return -_p_b /
+           (_m * _m * (S - _residual_saturation) * (S - _residual_saturation)) *
+           std::pow(1 - val1, -_m - 1) * std::pow(val1, _m - 1) *
+           ((1 - _m * _m) * val1 + _m - 1);
+}
+
+}  // namespace MaterialPropertyLib
--- a/MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
+++ b/MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
@@ -30,12 +30,51 @@ CapillaryPressureVanGenuchten::CapillaryPressureVanGenuchten(
       _p_b(p_b),
       _p_cap_max(maximum_capillary_pressure)
 {
+    if (_S_L_res < 0 || _S_L_res > 1)
+    {
+        OGS_FATAL(
+            "Van Genuchten capillary pressure model: "
+            "The residual liquid saturation value S_L_res = {:g} is out of "
+            "admissible range [0, 1].",
+            _S_L_res);
+    }
+    if (_S_L_max < 0 || _S_L_max > 1)
+    {
+        OGS_FATAL(
+            "Van Genuchten capillary pressure model: "
+            "The maximum liquid saturation value S_L_max = {:g} is out of "
+            "admissible range [0, 1].",
+            _S_L_max);
+    }
+    if (_S_L_res >= _S_L_max)
+    {
+        OGS_FATAL(
+            "Van Genuchten capillary pressure model: "
+            "The maximum liquid saturation S_L_max = {:g} must not be less or "
+            "equal to the residual liquid saturion S_L_res = { : g}.",
+            _S_L_max, _S_L_res);
+    }
     if (!(_m > 0 && _m < 1))
     {
         OGS_FATAL(
-            "The exponent value m = {:g} of van Genuchten capillary pressure "
-            "model, is out of its range of (0, 1)",
+            "Van Genuchten capillary pressure model: "
+            "The exponent value m = {:g} is out of of admissible range (0, 1).",
             _m);
+    }
+    if (_p_b <= 0)
+    {
+        OGS_FATAL(
+            "Van Genuchten capillary pressure model: "
+            "The pressure scaling value p_b = {:g} must be positive.",
+            _p_b);
+    }
+    if (_p_cap_max < 0)
+    {
+        OGS_FATAL(
+            "Van Genuchten capillary pressure model: "
+            "The maximum capillary pressure value p_cap_max = {:g} must be "
+            "non-negative.",
+            _p_cap_max);
     }
 }
 
--- a/MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
+++ b/MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
@@ -103,46 +103,4 @@ PropertyDataType CapillaryPressureVanGenuchten::dValue(
     double const val2 = std::pow(val1 - 1.0, -_m);
     return _p_b * (_m - 1.0) * val1 * val2 / (_m * (S_L - _S_L_res));
 }
-
-PropertyDataType CapillaryPressureVanGenuchten::d2Value(
-    VariableArray const& variable_array, Variable const primary_variable1,
-    Variable const primary_variable2,
-    ParameterLib::SpatialPosition const& /*pos*/, double const /*t*/,
-    double const /*dt*/) const
-{
-    (void)primary_variable1;
-    (void)primary_variable2;
-    assert((primary_variable1 == Variable::liquid_saturation) &&
-           (primary_variable2 == Variable::liquid_saturation) &&
-           "CapillaryPressureVanGenuchten::d2Value is implemented for "
-           "derivatives with respect to liquid saturation only.");
-
-    double const S_L = std::get<double>(
-        variable_array[static_cast<int>(Variable::liquid_saturation)]);
-
-    if (S_L <= _S_L_res)
-    {
-        return 0;
-    }
-
-    if (S_L >= _S_L_max)
-    {
-        return 0;
-    }
-    double const delta_S_L_res = S_L - _S_L_res;
-    double const S_eff = delta_S_L_res / (_S_L_max - _S_L_res);
-
-    assert(0 < S_eff && S_eff < 1.0);
-
-    double const val1 = std::pow(S_eff, -1.0 / _m);
-    double const p_cap = _p_b * std::pow(val1 - 1.0, 1.0 - _m);
-    if (p_cap >= _p_cap_max)
-    {
-        return 0;
-    }
-
-    return -_p_b / (_m * _m * delta_S_L_res * delta_S_L_res) * val1 * val1 *
-           std::pow(val1 - 1, -_m - 1) * ((1 - _m * _m) / val1 + _m - 1);
-}
-
 }  // namespace MaterialPropertyLib
--- a/MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
+++ b/MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
@@ -20,19 +20,21 @@ namespace MaterialPropertyLib
 {
 CapillaryPressureVanGenuchten::CapillaryPressureVanGenuchten(
     double const residual_liquid_saturation,
-    double const maximum_liquid_saturation, double const exponent,
-    double const p_b, double const max_capillary_pressure)
-    : _residual_saturation(residual_liquid_saturation),
-      _maximuml_saturation(maximum_liquid_saturation),
+    double const residual_gas_saturation,
+    double const exponent,
+    double const p_b,
+    double const maximum_capillary_pressure)
+    : _S_L_res(residual_liquid_saturation),
+      _S_L_max(1. - residual_gas_saturation),
       _m(exponent),
       _p_b(p_b),
-      _pc_max(max_capillary_pressure)
+      _p_cap_max(maximum_capillary_pressure)
 {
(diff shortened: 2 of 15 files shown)


MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
```diff
@@ -20,19 +18,21 @@ namespace MaterialPropertyLib
...
     double const residual_liquid_saturation,
-    double const maximum_liquid_saturation, double const exponent,
-    double const p_b, double const max_capillary_pressure)
-    : _residual_saturation(residual_liquid_saturation),
-      _maximuml_saturation(maximum_liquid_saturation),
+    double const residual_gas_saturation,
+    double const exponent,
+    double const p_b,
+    double const maximum_capillary_pressure)
+    : _S_L_res(residual_liquid_saturation),
```
wenqing, line 25: An underscore followed immediately by an uppercase letter should be avoid. This is why I did not use this kind of naming as that in `SaturationVanGenuchten`
See:
https://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier

MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
```diff
@@ -42,24 +42,24 @@ PropertyDataType CapillaryPressureVanGenuchten::value(
...
 
-    const double S =
-        std::clamp(saturation, _residual_saturation, _maximuml_saturation);
-    const double Se = (S - _residual_saturation) /
-                      (_maximuml_saturation - _residual_saturation);
-    const double pc =
-        (Se < 1.0) ? _p_b * std::pow(std::pow(Se, (-1.0 / _m)) - 1.0, 1.0 - _m)
-                   : 0.0;
-    return std::clamp(pc, 0.0, _pc_max);
+    //const double S = std::clamp(saturation, _S_L_res, _S_L_max);
```
wenqing: Why remove this?

MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
```diff
@@ -75,6 +75,8 @@ PropertyDataType CapillaryPressureVanGenuchten::dValue(
     const double val1 = std::pow(Se, -1.0 / _m);
     const double val2 = std::pow(val1 - 1.0, -_m);
     return _p_b * (_m - 1.0) * val1 * val2 / (_m * (S - _residual_saturation));
+    */
+    return 0;
 }
 
 PropertyDataType CapillaryPressureVanGenuchten::d2Value(
```
wenqing, line 105: Now I think this member, d2Value, can be removed. It was added purely for the test purpose and it will not be used anywhere in  computation.

MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.h
```diff
@@ -19,56 +17,73 @@ class Medium;
 class Phase;
 class Component;
 
-/// \copydoc MaterialPropertyLib::SaturationVanGenuchten
+/**
+ * \brief The van Genuchten capillary pressure model.
```
wenqing, line 21: Why not use \copydoc MaterialPropertyLib::SaturationVanGenuchten or use it  together with some additional words?

Documentation/ProjectFile/properties/property/CapillaryPressureVanGenuchten/t_residual_gas_saturation.md
```diff
@@ -0,0 +1 @@
+The smallest degree of saturation of the gas (or non-wetting) phase in the medium.
```
wenqing, line 1: Need to discuss together with @Scinopode and @boyanmeng about the parameter. The class handles the function of p_c(S^l), one needs to restriction the variable of S^l in (S^l_r, S^l_max). Input S^g_r then convert it to S^l_max is weird, especially for single phase flow.

MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
endJunction: Why is it needed?

Tests/MaterialLib/TestMPLCapillaryPressureVanGenuchten.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+        double const Dp_cap = (p_cap_plus - p_cap_minus) / 2 / eps;
+        double const Dp_cap2 =
+            (p_cap_plus - 2 * p_cap + p_cap_minus) / (eps * eps);
+
+        //
+        // First order derivative tests
+        //
+
+        // Testing relative error up to the S_L_max with different tolerances as
```
wenqing, line 109: For a smooth continue function, e.g., dp_c(S) of this model, testing it in any segment of the variable range is enough.   Therefore, the test with the smallest tolerance is enough.

MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.h
endJunction: This is not a part of new property. If something needs to be changed across different code parts, it should be done in a different PR.

MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
endJunction: > An underscore followed immediately by an uppercase letter should be avoided. This is why I did not use this kind of naming as that in `SaturationVanGenuchten`
This implementation uses same convention as the `SaturationVanGenuchten` [implementation](https://github.com/ufz/ogs/blob/master/MaterialLib/MPL/Properties/CapillaryPressureSaturation/SaturationVanGenuchten.h#L85-L88) for consistency.

> https://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier
The reference is interesting indeed. The rules should be checked automatically across the whole code base. Any ideas on how to do it?

Documentation/ProjectFile/properties/property/CapillaryPressureVanGenuchten/t_residual_gas_saturation.md
endJunction: This property is required only by multi-phase problems and is especially written to be used in the TwoPhaseFlowWithPrho process, isn't it?

MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
endJunction: And again, if this affects multiple places, which it does, open a new issue/pull-request.

endJunction: As far as I understand this the rule has effect on the global scope only: [n3242 / C++11 draft](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2011/n3242.pdf) 17.6.4.3.2 [global names].

Documentation/ProjectFile/properties/property/CapillaryPressureVanGenuchten/t_residual_gas_saturation.md
wenqing: The p(S) is not only used for two-phase flow. The Richards' equation can also use the saturation as primary variable.

Tests/MaterialLib/TestMPLCapillaryPressureVanGenuchten.cpp
endJunction: How do you guarantee, that the function is not called outside of its domain of definition?
How should a Newton method behave if it guesses a value outside of the domain of definition and gets a NaN or -inf as result?

endJunction: The p_c(S) is not smooth, just continuous.

MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.h
wenqing: If you want to keep the full documentation here. I suggest to use \copydoc  in MaterialPropertyLib::SaturationVanGenuchten to avoid duplication (could be in another PR.).

endJunction: That's absolutely fine to have a single page describing the van Genuchten capillary pressure - saturation. In another PR.

MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
wenqing: You can keep `_S...` naming, and we can resolve this issue later (whether change or not). I just remembered  that there is case of `_A`  name caused some problem. Maybe the new C++ standard removed some identification reservation.

BTW, cpplint may be used to check the naming issue. @bilke  may know this.

MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
wenqing: Sorry. It is not needed.

Tests/MaterialLib/TestMPLCapillaryPressureVanGenuchten.cpp
wenqing: I replace `smooth` with ` differentiable` in my comment. With the variable restriction of (S_r, S_m) and value restriction (0, p_c_max), test within (S_r+epsilon, S_m-epsilon) with a relative small tolerance is enough to validate the correction of the implementation. Anyway, the heavy test is also good.

Documentation/ProjectFile/properties/property/CapillaryPressureVanGenuchten/t_residual_gas_saturation.md
wenqing: By the way, this naming some how is related to the relative permeability, which should be a property of phase. For the relative permeability, these two restriction variables were suggested to remove phase info from their names (#2902) as
```
residual_saturation
maximum_saturation
```
after the relative permeability property is moved to phase property.

MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
wenqing: I still see the following statement on page 20 of [C++17 final draft](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/n4659.pdf). It is the same as before.
```
In addition, some identifiers are reserved for use by C++implementations
 and shall not be used otherwise; no diagnostic is required.
—(3.1)Each identifier that contains a double underscore__or begins
 with an underscore followed by an uppercase letter is reserved to
 the implementation
 for any use.
```

Tests/MaterialLib/TestMPLCapillaryPressureVanGenuchten.cpp
endJunction: The smoothness issue aside, the point of the test is to verify the implementation behaviour outside of the "middle"/well-behaved domain, especially around (in both directions) the residual saturation, the pressure cutoff, and the maximum saturation.

MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
endJunction: I see. Google's style guide also recommends to put the underscores after the variable names.
Please open an issue. Intermediate solution suggestion is to keep the leading for consistent names. New, independent code parts should have trailing underscores for private class members. 
@wenqing Please find a way to automatically check this, preferably with clang-tidy or cppcheck, such that we can eliminate this issue step by step.

wenqing:
only one comment:  d2Value may be removed.

MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
```diff
@@ -0,0 +1,106 @@
...
+{
+    (void)primary_variable;
+    assert((primary_variable == Variable::liquid_saturation) &&
+           "CapillaryPressureVanGenuchten::dValue is implemented for "
+           "derivatives with respect to liquid saturation only.");
+
+    double const S_L = std::get<double>(
+        variable_array[static_cast<int>(Variable::liquid_saturation)]);
+
+    if (S_L <= _S_L_res)
```
wenqing, line 82:
```suggestion
    if (!(S_L > _S_L_res))
```
in order to avoid the identical comparison of two real numbers.

MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
```diff
@@ -0,0 +1,106 @@
...
+    {
+        return 0;
+    }
+
+    if (S_L >= _S_L_max)
```
wenqing, line 87:
```suggestion
    if (!(S_L < _S_L_max))
```

MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
```diff
@@ -0,0 +1,106 @@
...
+    {
+        return 0;
+    }
+
+    double const S_eff = (S_L - _S_L_res) / (_S_L_max - _S_L_res);
+
+    assert(0 < S_eff && S_eff < 1.0);
```
wenqing, line 94: Unnecessary due to the if-conditions for S_l \in  [S_r, S_max].
The above three comments are for the same lines in dValue() as well.

MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+    double const val1 = std::pow(S_eff, -1.0 / _m);
+    double const p_cap = _p_b * std::pow(val1 - 1.0, 1.0 - _m);
+    if (p_cap >= _p_cap_max)
```
wenqing, line 98:
```suggestion
    if (!(p_cap < _p_cap_max))
```

MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
endJunction: Interesting. Do you have an example where
```c++
assert((a <= b) == (!(a > b)));
```
fails? I [tried](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(fontScale:14,j:1,lang:c%2B%2B,selection:(endColumn:3,endLineNumber:26,positionColumn:3,positionLineNumber:26,selectionStartColumn:3,selectionStartLineNumber:26,startColumn:3,startLineNumber:26),source:'%23include+%3Climits%3E%0A%23include+%3Ciostream%3E%0A%0Adouble+g(double+a,+double+b)%0A%7B%0A++++if+(a+%3E%3D+b)%0A++++return+1%3B%0A++++else%0A++++return+0%3B%0A%7D%0A%0Adouble+f(double+a,+double+b)%0A%7B%0A++++if+(!!(a+%3C+b))%0A++++return+1%3B%0A++++else%0A++++return+0%3B%0A%7D%0A/*%0Aint+main()%0A%7B%0A++++double+a,+b%3B%0A++++std::cout+%3C%3C+0.*g(a,+b)+%3C%3C+%22+%22+%3C%3C+0.+*+f(a,+b)+%3C%3C+%22%5Cn%22%3B%0A++++return+0%3B%0A%7D%0A*/'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:30.773018217626785,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:vcpp_v19_24_x64,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'0',intel:'0',libraryCode:'1',trim:'1'),fontScale:14,j:1,lang:c%2B%2B,libs:!(),options:/Ox,selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'x64+msvc+v19.24+(Editor+%231,+Compiler+%231)+C%2B%2B',t:'0')),k:35.89364844903989,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compiler:1,editor:1,fontScale:14,wrap:'1'),l:'5',n:'0',o:'%231+with+x64+msvc+v19.24',t:'0')),k:33.33333333333333,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4) to find difference with different compilers and optimization levels and there are some.
For gcc-9.3.0 and msvc 19.24 and optimization >= O1 the comparisons are using the same instruction `comisd`: https://www.felixcloutier.com/x86/comisd
clang-10.0 and optimization >= 01 yields "different" instructions `cmpltsd` `cmpnlesd`, which, to my understanding, are just so called pseudo-op for `cmpsd`
```
CMPLTSD xmm1, xmm2	CMPSD xmm1, xmm2, 1
```
where the last 1 indicates less-then predicate. https://www.felixcloutier.com/x86/cmpsd

`comisd` sets multiple flags depending on the comparison. `cmpsd` set's destination register to 0xFF or 0x00 depending on the outcome.

As I do understand how the floating point numbers are compared the above assertion can never fail. Please show a case, where it fails.

MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
endJunction: This assertion catches possible programming errors for the code which follows it, not to check if the above if conditions were programmed correctly.

wenqing: I see.

MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
wenqing: Nice explanation. Go forward.

endJunction:
@Scinopode Seems we are done here. Any comments from your side?

MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
TomFischer: - IMHO the version `x <= y` is much more readable than `!(x > y)`. Therefore I would keep the code as it is. 
- As far as I remember Scott Meyers recommended to implement the operator `<` using the negation of the operator `<=`. So the above expression should be identical.

endJunction: @wenqing Please show a case where this relation operators do fail for doubles.

wenqing: For example, `S_L=NaN`. You can also use `std::islessequal` and `std::isgreaterequal`. For float number comparison, IMHO, operator of `==`, `<=`, `>=` should be avoided.

wenqing: If you want to keep `if (S_L <= _S_L_res)`,  please go forward. `S_L=NaN` seldom occurs, even   in the equation with saturation as the primary variable. Anyway, the program can be halt by linear/non-linear solvers under such case.

endJunction: I see how that example is showing inconsistency. The reason for NaN comparisons to be different is that **any** comparison with NaN is `false` but the "not-equal". Then negating the result is not really making it correct, though the expression becomes `true`. The `std::isgreaterequal(a,b)` is also not equal to `!(a < b)` for NaN's, beacause `(a < b)` for a NaN is always `false`.
If this would restore some logic, yes, I'd agree that the expression with negation is superior.
C'mon, NaN is representation of an *undefined value*, how does the negated version save the logic?
To catch runtime errors, use `ogs --enable-fpe x.prj`.

From `NaN < b` being `false` you cannot conclude that `b > NaN` is `true`, the latter is also `false`.

Scinopode:
> @Scinopode Seems we are done here. Any comments from your side?

It's fine for me.

endJunction:
@Scinopode Thanks! Will fix the mentioned points.

MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
```diff
@@ -0,0 +1,106 @@
...
+      _p_cap_max(maximum_capillary_pressure)
+{
+    if (!(_m > 0 && _m < 1))
+    {
+        OGS_FATAL(
+            "The exponent value m = {:g} of van Genuchten capillary pressure "
+            "model, is out of its range of (0, 1)",
+            _m);
+    }
+}
```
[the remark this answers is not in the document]
endJunction: Added tests also for S_L_res and S_L_max, as well as for p_b.

Testing p_cap_max for being positive, though. But I don't get the relation between p_b and p_cap_max; they seem to be independent of each other.... help!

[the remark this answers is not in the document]
endJunction: Clarified. Just checking p_cap_max for being positive, independent of p_b.

TomFischer:
Can the PR be merged?

PR #2920, opened 2020-04-24
MPL; use trailing underscores for private members

endJunction:
As pointed out by @wenqing this fixes possible issue when using underscores in front of an identifier.

Review after #2919

### DIFF
--- a/Documentation/ProjectFile/properties/property/BishopsPowerLaw/t_exponent.md
+++ b/Documentation/ProjectFile/properties/property/BishopsPowerLaw/t_exponent.md
@@ -1,1 +1,1 @@
-\copydoc MaterialPropertyLib::BishopsPowerLaw::_m
+\copydoc MaterialPropertyLib::BishopsPowerLaw::m_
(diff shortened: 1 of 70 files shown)


wenqing:
All looks fine except:
1) non-public member naming, which is under discussion.
2) whether d2Value should be kept or not.

wenqing:
I am fine with this trailing underscore naming. If we decide to use it, the names in other classes have to  be changed as well (may by a script?).

endJunction:
> All looks fine except:
> 
> 1. non-public member naming, which is under discussion.
> 2. whether d2Value should be kept or not.

@wenqing  You wanted to comment in #2919 ?

wenqing:
> @wenqing You wanted to comment in #2919 ?

Just made a comment in  #2919.

endJunction:
@wenqing Now this is merged, how do you want to proceed with the other libraries?

wenqing:
> @wenqing Now this is merged, how do you want to proceed with the other libraries?

There are so many such private members including their documentations have to be replaced. I suggest to find someone being good at python script to write a short script to do such replacements. If no body is willing to do it, I will manually  change those in the classes in my on going PRs. The replacements in other classes can be done by who will touch the classes.

The trailing underscore looks ugly. I remembered that we had some discussion about the issue over my implementation of PETSc matrix, where `_A` was initially used.

endJunction:
Make it in chunks, like library wise, utils, etc. Partial changes are difficult to enforce in reviews because people always come with an argument, "but the others do it this way too...".

PR #2921, opened 2020-04-26
RichardsMechanics; Update mass lumping procedure.

endJunction:
2. [x] Tests covering your feature were added? updated.


nagelt:
looks good

nagelt:
> Did not understand the formula. PR looks technical okay. fast_forward

It replaces S' with S/Delta p N p'_nodes, in order to turn a nodal vector into a "mass" matrix (still don't like this term). This way, we can apply mass lumping. In contrast to a \partial S/\partial p formulation we maintain the good mass conservation properties of the S' formulation.

ProcessLib/RichardsMechanics/RichardsMechanicsFEM-impl.h
```diff
@@ -802,23 +808,24 @@ void RichardsMechanicsLocalAssembler<ShapeFunctionDisplacement,
 
         storage_p_a_p.noalias() +=
             N_p.transpose() * rho_LR * specific_storage_a_p * N_p * w;
-        storage_p_a_S.noalias() += N_p.transpose() * rho_LR *
-                                   specific_storage_a_S * (S_L - S_L_prev) /
-                                   dt * w;
+        if (p_cap_dot_ip != 0)  // prevent division by zero.
```
wenqing, line 811: If `p_cap_dot_ip` is very small, this condition would causes a problem.

nagelt: > If `p_cap_dot_ip` is very small, this condition would cause a problem.

Yeah. In the saturated range p_cap_dot_ip = -p_L_dot_ip, and the potentially small pressure rate goes along with S_L = S_L_prev. So this case should be fine, I guess.
In the unsaturated case, the ratio of both depends on the shape of the SWCC, but in many cases it should work.

@endJunction if none of the benchmarks nor our additional tests run into any problems of this kind, then I'd be fine with this for now and only address it if an actual problem occurs.

TomFischer:
> 
> It replaces S' with S/Delta p N p'_nodes, in order to turn a nodal vector into the product of a "mass" matrix (still don't like this term) and a nodal vector. This way, we can apply mass lumping. In contrast to a \partial S/\partial p formulation we maintain the good mass conservation properties of the S' formulation.

Thanks for the explanation.

ProcessLib/RichardsMechanics/RichardsMechanicsFEM-impl.h
endJunction: So far I didn't come across a case where it would produce anything strange.

PR #2931, opened 2020-05-04
[DE] Raster translation fix

rinkk:
Small changes regarding positioning raster data in geographic context.

Previously, position data had been sometimes pulled from image algorithm object and sometimes been given explicitely via parameters. Now it's *always* within the image algorithm, making the interface slimmer and reducing the number of ambiguous cases.

(I cannot currently test this for GeoTIFFs. It *should* work but I'm not 100% sure.)

### DIFF
--- a/Applications/DataExplorer/VtkVis/VtkRaster.cpp
+++ b/Applications/DataExplorer/VtkVis/VtkRaster.cpp
@@ -120,7 +120,6 @@ vtkImageAlgorithm* VtkRaster::loadImageFromTIFF(const std::string& fileName)
 
         int version[3];
         int count (0);
-        double x0, double y0, cellsize;
         GTIFDirectoryInfo(geoTiff, version, &count);
 
         if (count == 0)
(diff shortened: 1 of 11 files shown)


Applications/DataExplorer/VtkVis/VtkRaster.cpp
```diff
@@ -126,6 +120,7 @@ vtkImageAlgorithm* VtkRaster::loadImageFromTIFF(const std::string& fileName,
 
         int version[3];
         int count (0);
+        double x0, double y0, cellsize;
```
TomFischer, line 123: Should be moved into the `if (geoTiff)` branch.
 ```suggestion
        double x0 = 0.0; 
        double y0 = 0.0;
        double cellsize = 0.0;
```

TomFischer:
Just out of curiosity: Why can't you test the changes for GeoTIFFs?

Applications/DataExplorer/VtkVis/VtkRaster.cpp
rinkk: ✔️

rinkk:
GeoTiff is currently disabled (at least on Windows). IIRC because of some Conan issue that happened the last week we were int he office. But you'd have to ask Lars for details. 

(For the record: You can still read regular Tiffs because I use the VTK-method for that if libgeotiff isn't found. However, the VTK method ignores the GeoTiff-specific tags containing information regarding georeferencing.)

PR #2932, opened 2020-05-05
Correction of the calcualtion of dPc/dS when S<S_r in the van Genuchten model.

wenqing:
In PR. #2919, `dPc/dS` when `S<S_r` is set to zero. It is not correct.

The correction in this PR is that: `dPc/dS` is computed with `S_r + 1.e-9` or `S` for `pc_max` if `S<S_r`. For this correction, a new member of  `_S_for_p_cap_max` is introduced to `CapillaryPressureVanGenuchten`.

### DIFF
--- a/MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
+++ b/MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
@@ -15,9 +15,20 @@
 #include <cmath>
 
 #include "MaterialLib/MPL/Medium.h"
+#include "GetSaturationVanGenuchten.h"
 
 namespace MaterialPropertyLib
 {
+double getCapillaryPressure(double const S_L, double const p_b,
+                            double const S_L_res, double const S_L_max,
+                            double const m)
+{
+    double const S_eff = (S_L - S_L_res) / (S_L_max - S_L_res);
+    assert(0 <= S_eff && S_eff <= 1);
+
+    return p_b * std::pow(std::pow(S_eff, -1.0 / m) - 1.0, 1.0 - m);
+}
+
 CapillaryPressureVanGenuchten::CapillaryPressureVanGenuchten(
     double const residual_liquid_saturation,
     double const residual_gas_saturation,
@@ -28,16 +39,13 @@ CapillaryPressureVanGenuchten::CapillaryPressureVanGenuchten(
       S_L_max_(1. - residual_gas_saturation),
       m_(exponent),
       p_b_(p_b),
-      p_cap_max_(maximum_capillary_pressure)
+      p_cap_max_(std::min(maximum_capillary_pressure,
+                          getCapillaryPressure(S_L_res_ + 1.0e-9, p_b_,
+                                               S_L_res_, S_L_max_, m_))),
+      S_for_p_cap_max_(std::max(
+          S_L_res_ + 1.0e-9,
+          getSaturationVanGenuchten(p_cap_max_, p_b_, S_L_res_, S_L_max_, m_)))
 {
-    if (S_L_res_ < 0 || S_L_res_ > 1)
-    {
-        OGS_FATAL(
-            "Van Genuchten capillary pressure model: "
-            "The residual liquid saturation value S_L_res = {:g} is out of "
-            "admissible range [0, 1].",
-            S_L_res_);
-    }
     if (S_L_max_ < 0 || S_L_max_ > 1)
     {
         OGS_FATAL(
@@ -86,7 +94,7 @@ PropertyDataType CapillaryPressureVanGenuchten::value(
     double const S_L = std::get<double>(
         variable_array[static_cast<int>(Variable::liquid_saturation)]);
 
-    if (S_L <= S_L_res_)
+    if (S_L <= S_for_p_cap_max_)
     {
         return p_cap_max_;
     }
@@ -96,12 +104,8 @@ PropertyDataType CapillaryPressureVanGenuchten::value(
         return 0;
     }
 
-    double const S_eff = (S_L - S_L_res_) / (S_L_max_ - S_L_res_);
-    assert(0 <= S_eff && S_eff <= 1);
-
     double const p_cap =
-        p_b_ * std::pow(std::pow(S_eff, -1.0 / m_) - 1.0, 1.0 - m_);
-    assert(p_cap > 0);
+        getCapillaryPressure(S_L, p_b_, S_L_res_, S_L_max_, m_);
     return std::min(p_cap, p_cap_max_);
 }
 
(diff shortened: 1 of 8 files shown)


MaterialLib/MPL/Properties/CapillaryPressureSaturation/CapillaryPressureVanGenuchten.cpp
```diff
@@ -96,12 +104,8 @@ PropertyDataType CapillaryPressureVanGenuchten::value(
         return 0;
     }
 
-    double const S_eff = (S_L - _S_L_res) / (_S_L_max - _S_L_res);
-    assert(0 <= S_eff && S_eff <= 1);
-
     double const p_cap =
-        _p_b * std::pow(std::pow(S_eff, -1.0 / _m) - 1.0, 1.0 - _m);
-    assert(p_cap > 0);
```
wenqing, line 104: Removed this assertion because it is unnecessary. The formula always gives a positive value with the restrictions of all parameters in ctor.

Tests/MaterialLib/TestMPLCapillaryPressureVanGenuchten.cpp
```diff
@@ -61,14 +72,14 @@ TEST(MaterialPropertyLib, CapillaryPressureVanGenuchten)
 
         if (S_L < residual_liquid_saturation)
         {
-            EXPECT_GE(S_L_roundtrip, S_L)
-                << "for liquid saturation " << S_L << " and capillary pressure "
-                << p_cap;
+            EXPECT_GE(S_L_roundtrip, S_L) << "for liquid saturation " << S_L
+                                          << " and capillary pressure "
+                                          << p_cap;
```
wenqing, line 77: Just format

endJunction:
The `p_cap(S_L)` returns 0 for `S_L < S_res`.  Why should the derivative not be zero when the function is constant?

wenqing:
> The `p_cap(S_L)` returns a constant `p_cap_max` for `S_L < S_res`. Why should the derivative not be zero when the function is constant?

Both Pc(S) and dPc(S)/dS have singularity at S=Sr. That is why we cannot set dPc(S)/dS=0 at  S=Sr.  We can have cutoffs for Pc(S) and dPc(S)/dS by setting a maximum Pc for numerical stability. 
I plotted a figure of the model for my explanation.
cp

endJunction:
I'm not sure I follow your argumentation. In the code we have a condition for `S_L<=S_res`. For that interval `(-inf, S_res]` the return value is a constant. Therefore the derivative in that interval is zero.
The function `p_cap(S_L)` (as it is implemented now) is not differentiable at `S_L = S_res`. At that point one can argue what is numericall better to return; a zero or +inf would be both possible.

wenqing:
> I'm not sure I follow your argumentation. In the code we have a condition for `S_L<=S_res`. For that interval `(-inf, S_res]` the return value is a constant. Therefore the derivative in that interval is zero.
> The function `p_cap(S_L)` (as it is implemented now) is not differentiable at `S_L = S_res`. At that point one can argue what is numericall better to return; a zero or -inf would be both possible.

There is a restriction of S \in [Sr, Smax]. I hope what I drew in this following figure can help to understand my changes.
IMG_20200506_121244
One has to avoid to have dPc(S)/dS=0 at S=Sr, and then to have it a jump to a vary large negative value.

wenqing:
The curve is not flatten in (Sr, Sr+epsilon). Otherwise, there is very large negative jump or dPc(S)/dS from zero. With p_max, only the segment of the curve in [Sr+epsilon, Smax] is used.

boyanmeng:
I found this feature very useful, since now the derivative of capillary pressure w.r.t. saturation can actually be constrained by giving a respective `pc_max`. The implementation further imporoves the numerical stability of van Genuchten model towards or below Sr.

endJunction:
Better to move the capillary pressure cutoff out of the MPL and into the specific process implementation.
I strongly disagree with a function's derivative not being correctly (mathematically) implemented.

wenqing:
> Better to move the capillary pressure cutoff out of the MPL and into the specific process implementation.
> I strongly disagree with a function's derivative not being correctly (mathematically) implemented.

The model held by this class can be treated as the raw van Genuchten model with an extension such that
```
if (S<S_pc_max)), set S=S_pc_max. 
```
Therefore if S<S_pc_max, one gets Pc=pc_max, dPc/dS=dPc/dS|(S_pc_max). Obviously, there is no error at all in the function's derivative (mathematically) under such consideration. Besides, the  pc_max is a model parameter. It is not adequate  to move the material property parameter to specific process classes.  

I will add some more description in the documentation about this class, and possibly set pc_max as an optional parameter. In the functions of value and dValue, only the following condition will be used.
```
if (S<S_pc_max)
{
   S=S_pc_max
}
```

endJunction:
I didn't make myself clear on this, but given a function
```c++
double f(double x)
{
    if (x < x_0)
       return f_0;
    else
       return x*x;
}
```
the derivative is
```c++
double dfdx(double x)
{
    if (x < x_0)
        return 0;
    else
        return 2*x;
}
```
Probably we are talking about different functions, who knows. But this is the current implementation in the master branch. Changing derivative to anything else is incorrect.

wenqing:
> I didn't make myself clear on this, but given a function
> 
> ```c++
> double f(double x)
> {
>     if (x < x_0)
>        return f_0;
>     else
>        return x*x;
> }
> ```
> 
> the derivative is
> 
> ```c++
> double dfdx(double x)
> {
>     if (x < x_0)
>         return 0;
>     else
>         return 2*x;
> }
> ```
> 
> Probably we are talking about different functions, who knows. But this is the current implementation in the master branch. Changing derivative to anything else is incorrect.

The case is
```
   if (x<x_l)
   {
       x=x_l
   }
   return f(x);
```
 with x_l the lower bound of the variable of a function.  Please see the improved  documentation of the class.

endJunction:
> The case is
> 
> ```
>    if (x<x_l)
>    {
>        x=x_l
>    }
>    return f(x);
> ```

The derivative of this function is
```c++
    if (x < x_l)
        return 0;
    else
        return dfdx(x);
```

wenqing:
> > The case is
> > ```
> >    if (x<x_l)
> >    {
> >        x=x_l
> >    }
> >    return f(x);
> > ```
> 
> The derivative of this function is
> 
> ```c++
>     if (x < x_l)
>         return 0;
>     else
>         return dfdx(x);
> ```

I hope that the following mathematical definition would convince you. 

f

endJunction:
The definition of spaces does not change the derivative.
If you would compare a numerical derivative to your provided one in the range [-0.1,1.1] you would see some differences. The way you changed the test where there is comparison to the numerical derivative is not acceptable.

I have mentioned my view on this subject sufficiently many times and quit the discussion.

wenqing:
> A derivative of a function must be correct on the whole range of the function's input.

The function pc(S) and its derivative are only valid in S \in [Sr, Smax]. The constant extension pc(S) to [0,Sr] is not applicable. The current PR uses a simple extension by the range mapping.  A comment can be added to the if-condition concerning to the mapping.

boyanmeng:
As Wenqing has pointed out, the original van Geunuchten function (including its derivative) is only defined within (Sr, Smax]. However, in order to handle numerical cases where Sw<=Sr, a mapping is proposed so that if the abovementioned condition holds, Sw is set to S_pc_max. Such mapping is rather a numerical compromise and should not be treated completely equivalent to an extension of the van Genuchten function to (-∞, Sr].

wenqing:
I agree that my mathematical explanation by mapping fails, i.e d(barS(S))/dS=0 for all S in[0, Sr]. But dpc/dS cannot be 0 at S=Sr. To follow such purity definition dValue, while guarantee the correction of model, the only way could be
1. restriction S in [Sm, Smax] in value() and dValue() as that is defined in the model such as

     if (S<Sm|| S>Smax)

     { OGS_FATAL(.....)}

     or assert(S>=Sr && S<=Smax)

2. restriction pc and dpc/dS at S=Sm in local assemble via process property by passing std::clamp(S, Sm, Smax) to value and dValue, or the mapping is done in local assembler.

PR #2935, opened 2020-05-06
Pass xdot to compute secondary

endJunction:
Pass xdot and dt to Process::computeSecondaryVariables().
In this PR the xdot and dt is mostly commented and is needed in RichardsMechanics process (see following PR #2936) to compute rate dependent secondary variables.

Main change is the storage of `x_prev` value in the time loop.
Review commit-wise.

1. [ ] Feature description was added to the [changelog](https://github.com/ufz/ogs/wiki/Release-notes-6.3.1)
2. [x] Tests covering your feature were added? Covered by ctests
3. [x] Any new feature or behavior change was documented? No user changes.

PR #2936, opened 2020-05-06
RichardsMechanics; restart.

endJunction:
Follow up of #2935 


ProcessLib/TimeLoop.cpp
```diff
@@ -648,8 +648,8 @@ void postTimestepForAllProcesses(
         }
         auto& x = *process_solutions[process_id];
         auto& x_dot = *x_dots[process_id];
-        pcs.postTimestep(process_solutions, t, dt, process_id);
         pcs.computeSecondaryVariable(t, dt, x, x_dot, process_id);
+        pcs.postTimestep(process_solutions, t, dt, process_id);
```
renchao-lu, line 652: Despite subtle, this change is very beneficial for the understanding of physical processes. Can this change be understood as the member function `computeSecondaryVariable` serves as named solely for computing secondary variables and `postTimestep` serve for processing local state variables. Is my understanding correct?

endJunction: Yes. I tried to summarize this in the commit message---have a look, can be improved if you suggest something.
The usage of postTimestep is more general, not only pushing states, but the computation of secondary indeed must not be mixed with other things because of the extra initialization call before first time step.

renchao-lu: IMHO, it is more clear to message like `computation of secondary variables` instead of `computation of some of those`. For me, the stress/strain field is derived from the displacement field and thought of not as a local variable.

endJunction: :+1: Will update with next push.

ProcessLib/Output/ProcessOutput.cpp
```diff
@@ -10,14 +10,18 @@
 
 #include "ProcessOutput.h"
 
+#ifndef _WIN32
+#ifndef __APPLE__
+#include <cfenv>
+#endif  // __APPLE__
+#endif  // _WIN32
```
TomFischer, line 17: What happens on WIN32 and APPLE platforms when floating point exceptions occur?

endJunction: TL;DR; Same as before. Change only for linux.

Old behaviour everywhere but on linux: On windows nothing happens at all. On mac the exceptions are not changed for this particular snippet, i.e. if FPE are set **and** if FPE happens in vtk somewhere, it will stop the programme, as before. See code comments below, where it is used for more explanations.

Issue #2226, opened 2018-09-28
[Bug] Transience in Richards(Component)-Flow 


waltherm:
Maybe it is just me, but it seems that the transience in RichardsFlow and RichardsComponentTransport are wrong. 

I used the first setup from this file as a reference (https://www.pc-progress.com/Downloads/Public_Lib_H1D/HYDRUS-1D_Tutorial_V1.00_2018.pdf), it is a simple infiltration setup: IC unsaturated (p=-9810 Pa), one BC at the top with p=0 Pa, relevant van-Genuchten parameters are given in figure 1.10 (p. 11).

With RichardsFlow, nothing happens at all: results are equal to ICs.
With RichardsComponentFlow, the saturation front is too fast (see figure below and compare this to 1.18, p. 18). NB: I used SI units, so x-axis is given in seconds (not days), y-axis in meter (not cm); the figure only shows the points at x={0.8, 0.6, 0.4, 0.2}m (equivalent to N1, N2, N3, N4 of fig. 1.18).

The setup is enclosed.
[example1_infiltration.tar.gz](https://github.com/ufz/ogs/files/2428733/example1_infiltration.tar.gz)

![out3](https://user-images.githubusercontent.com/6029954/46217330-fbd30000-c341-11e8-865f-87170f740244.png)

waltherm:
@endJunction @TomFischer @Thomas-TK @OlafKolditz @HBShaoUFZ 

Dear all,

I have updated this benchmark with the latest OGS6 branch and created the documentation in this branch: https://github.com/waltherm/ogs/tree/1d-infiltration 

I would like to ask you:
* for comments on the documentation
* ideas/comments about how close we want the results to be and what should be done to decrease differences
* support to set up the same project for richards-mechanic and richards-component

Furthermore, I suggest that another example will be set up. I am willing to do this, but would like to ask you for ideas which setup should be tested.

Thanks for your help,
Marc

waltherm:
@endJunction @TomFischer @Thomas-TK @OlafKolditz @HBShaoUFZ

I updated the description based on @endJunction's comments. 
New commit: https://github.com/waltherm/ogs/commit/5721765f9b68578bfc4b178bfba5a695f24d260b

ogsbot:
OpenGeoSys development has been moved to [GitLab](https://gitlab.opengeosys.org/ogs/ogs).

[See this issue on GitLab](https://gitlab.opengeosys.org/ogs/ogs/-/issues/2226).

[...]

PR #2938, opened 2020-05-07
Benchmark Richards flow (1d infiltration)

waltherm:

Benchmark provided based on https://github.com/ufz/ogs/issues/2226.


web/content/docs/benchmarks/richards-flow/infiltration_1D.pandoc
```diff
@@ -0,0 +1,62 @@
...
+
+
+## Problem description
+
+The problem describes a transient infiltration process. The area of the simulation domain that is evaluated is $1 m$ deep and initially unsaturated where the initial condition is defined as hydraulic pressure $p_{IC} = -1 m = -9810 Pa$. The upper boundary condition is a Dirichlet type with $p_{top} = 0.01 m = 98.1 Pa $. The original lower boundary condition in HYDRUS-1D is set as a "free drainage" type (see [2]); in both OGS versions, we extend the model domain to become $2 m$ and mimic the same behavior (the lower boundary condition for OGS is a no-flow Neumann type). The domain is discretized in HYDRUS-1D with line segments of $0.01 m$, in OGS with line segments of $0.005 m$.
+
+The parametrization is summarized in the following table. HYDRUS-1D and OGS5 use the classical Mualem-van-Genuchten formulation with parameters $\alpha$ and $n$, while OGS6 uses the formulation given in ###.
```
TomFischer, line 26: ` ... given in ###.` Is this correct or did you forget the reference?

waltherm: Ah, good spotting! No, I did not forget this initially (yet, when pushing), but it was intended to link to the description in doxygen. Unfortunately, this description is still missing.

bilke:
@waltherm I fixed the image in the web doc (and renamed the pandoc file).

ProcessLib/RichardsFlow/Tests.cmake
```diff
@@ -116,3 +116,33 @@ AddTest(
     RichardsFlow_2d_richardsflow_pcs_0_ts_99_t_1900.000000.vtu RichardsFlow_2d_richardsflow_pcs_0_ts_99_t_1900_000000_0.vtu pressure pressure 5e-8 1e-10
     RichardsFlow_2d_richardsflow_pcs_0_ts_99_t_1900.000000.vtu RichardsFlow_2d_richardsflow_pcs_0_ts_99_t_1900_000000_0.vtu saturation saturation 1e-10 1e-11
 )
+
+# 1d infiltration with RichardsFlow
+AddTest(
+        NAME 1d-infiltration
+        PATH Parabolic/Richards
```
TomFischer, line 123:
```suggestion
        PATH Parabolic/Richards/infiltration_1D
```

waltherm: Thanks, @TomFischer

waltherm:
Hi @TomFischer @endJunction ,
so currently, the VTKDiff for this new example is failing. I ran the initial setup on my pc and compared this with the results Tom got from envinf1 (these were fine). Tom already pointed out that the issue might be related to the variable time stepping. What is your advice here how we can make this test a little more reliable?
Thanks,
Marc

Tests/Data/Parabolic/Richards/infiltration_1D/1d-infiltration.prj
```diff
@@ -0,0 +1,178 @@
...
+        </medium>
+    </media>
+    <time_loop>
+        <processes>
+            <process ref="Hus">
+                <nonlinear_solver>NLS</nonlinear_solver>
+                <convergence_criterion>
+                    <type>DeltaX</type>
+                    <norm_type>NORM2</norm_type>
+                    <reltol>1e-8</reltol>
```
endJunction, line 99: Can you tighten the tolerance? I'd set the minimum which is still converging. This will reduce the spread of solutions due to different compiler/system setups.

waltherm: Can try, what would you consider a reasonable limit for this value?

endJunction: It is problem-dependent. I'd run it with 1e-16 first and observe the convergence behaviour. Then choose bigger value based on the output, s.t. the global iterations converge in reasonable number of steps (which, you guess it already, is problem-dependent ;) .

waltherm: For the sake of completeness: why starting with 1e-16?

TomFischer: I think with a tol. of 1e-16 the equations are solved as accurately as possible on different CPU architectures / different operating systems / different compilers. I tried various thresholds and obtained different timesteps in which the timestamp (due to adaptive timestepping) is different from the reference file timestamp.

reltol   | timestep with different timestamp
---------|---------------------------------------------
1e-16 | 105
1e-14 | 525
1e-12 | 350

endJunction: 1e-16 is close to machine epsilon for doubles.

TomFischer: I once removed the timestamp from the file name, but also the actual simulation results differ quite a lot in these time steps.

TomFischer: I am not sure how we can achieve that benchmarks with adaptive time-stepping on different architectures always converge in the same way. Maybe it is better to create the output files on envinf1 or in a docker container and only test the benchmarks in the same environment.

waltherm: > I am not sure how we can achieve that benchmarks with adaptive time-stepping on different architectures always converge in the same way. Maybe it is better to create the output files on envinf1 or in a docker container and only test the benchmarks in the same environment.

Or we have to set a defined time stepping size. Somewhat cumbersome...

wenqing: You may try `fixed_output_times ` for the adaptive time-stepping.  If that does not work due to file name difference,  you can dump the time step sizes by the  adaptive time-stepping on your machines, and then use the fixed time step with the dumped   time step sizes.

waltherm: Ok, thanks @wenqing for the hint. I will try this (using constant steps from one adaptive run) in the afternoon if there are no objections.

waltherm: I took a look at the time stepping and produced the following graph for different configurations of maximum time step size (60 and 120 s) and the non-linear iteration tolerance (1e-12 and 1e-9).
Bild1

I have the following concerns:
- You can see that all configurations show seemingly random drops in the time step size after the max has been reached and the simulation seemed to have stabilized (note that there is no change in boundary conditions, grid resolution nor parameter homogeneity). I cannot explain this, but it worries me somehow.
- Trying to find a version without these drops, I firstly decreased the tolerance (initially started with 1e-14), but couldn't find one without drops. I could decrease this further, but I guess @endJunction wouldn't be happy with this (and this is fine).
- Even more, for smaller maximum time step sizes, you can see that there are more frequent drops than for larger ones. Don't know if this is a good sign...

Now my question: how should we continue? I guess it is not a meaningful idea to decrease the tolerance further, nor does it seem rational to explicitely define all random drops (as they may, again, be different on other machines).

endJunction: @waltherm If you are happy with that tolerance, if you are confident that (most) errors will be caught when the code changes, you can choose any tolerance. Strict tolerances give less room for "error creep-in".

waltherm:
I have updated the prj based on the comments: change to constant time step sizes, decrease non-linear tolerance.

Tests/Data/Parabolic/Richards/infiltration_1D/1d-infiltration.prj
waltherm: @endJunction As I am already unconfident with the previous analysis (smaller time steps -> larger number of fluctuations), I don't know if I should be certain with this setup...

ProcessLib/RichardsFlow/Tests.cmake
```diff
@@ -125,24 +125,44 @@ AddTest(
...
-    out_ex1_pcs_0_ts_500_t_47303.496783_expected.vtu out_ex1_ts_500.vtu pressure pressure 1e-4 1e-5
-    out_ex1_pcs_0_ts_550_t_52303.496783_expected.vtu out_ex1_ts_550.vtu pressure pressure 1e-4 1e-5
-    out_ex1_pcs_0_ts_600_t_57303.496783_expected.vtu out_ex1_ts_600.vtu pressure pressure 1e-4 1e-5
-    out_ex1_pcs_0_ts_650_t_62293.496783_expected.vtu out_ex1_ts_650.vtu pressure pressure 1e-4 1e-5
-    out_ex1_pcs_0_ts_700_t_67293.496783_expected.vtu out_ex1_ts_700.vtu pressure pressure 1e-4 1e-5
-    out_ex1_pcs_0_ts_750_t_72293.496783_expected.vtu out_ex1_ts_750.vtu pressure pressure 1e-4 1e-5
-    out_ex1_pcs_0_ts_800_t_77181.778033_expected.vtu out_ex1_ts_800.vtu pressure pressure 1e-4 1e-5
-    out_ex1_pcs_0_ts_850_t_82181.778033_expected.vtu out_ex1_ts_850.vtu pressure pressure 1e-4 1e-5
-    out_ex1_pcs_0_ts_894_t_86400.000000_expected.vtu out_ex1_ts_894.vtu pressure pressure 1e-4 1e-5
+    out_ex1_ts_0_expected.vtu out_ex1_ts_0.vtu pressure pressure 1e-8 1e-8
```
bilke, line 128: @waltherm I think the filename for the first argument is wrong, it should be `out_ex1_ts_0.vtu`. Full line would be:

```
out_ex1_ts_0.vtu out_ex1_ts_0.vtu pressure pressure 1e-8 1e-8
```

waltherm: You are right... I thought I had changed the reference file names...

bilke:
I checked the output:

## On Windows

-  `out_ex1_ts_340.vtu` is the last file written

error output:

```
360:   info: [time] Linear solver took 0.0003547 s.
360:
360:   info: Convergence criterion: |dx|=3.3939e-02, |x|=1.6024e+05,
360:   |dx|/|x|=2.1180e-07
360:
360:   info: [time] Iteration #30 took 0.0064527 s.
360:
360:   info: [time] Solving process #0 took 0.192618 s in time step #340
360:
360:   error: The nonlinear solver failed in time step #340 at t = 37464.5 s for
360:   process #0.
360:
360:   info: [time] Output of timestep 340 took 0.0044219 s.
360:
360:   critical: E:/jenkins/workspace/ufz_ogs_PR-2938/ProcessLib/TimeLoop.cpp:680
360:   ProcessLib::TimeLoop::solveUncoupledEquationSystems()
360:
360:   warning: There are still some matrices and vectors in use.  This might be
360:   an indicator of a possible waste of memory.
360:
360:   error: Time stepper cannot reduce the time step size further.
360:
360:   info: OGS terminated on 2020-05-27 09:40:25+0200.
360:
360:   error: OGS terminated with error.
```

## frontend2 (eve)

-  `out_ex1_ts_340.vtu` is the last file written

error output:

```
362:   info: [time] Linear solver took 0.00022539 s.
362:
362:   info: Convergence criterion: |dx|=3.4041e-02, |x|=1.6024e+05,
362:   |dx|/|x|=2.1244e-07
362:
362:   info: [time] Iteration #30 took 0.00427144 s.
362:
362:   info: [time] Solving process #0 took 0.124821 s in time step #340
362:
362:   error: The nonlinear solver failed in time step #340 at t = 37464.5 s for
362:   process #0.
362:
362:   info: [time] Output of timestep 340 took 0.00218314 s.
362:
362:   critical:
362:   /gpfs1/data/ogs/jenkins-frontend2/workspace/ufz_ogs_PR-2938/ProcessLib/TimeLoop.cpp:680
362:   solveUncoupledEquationSystems()
362:
362:   warning: There are still some matrices and vectors in use.  This might be
362:   an indicator of a possible waste of memory.
362:
362:   error: Time stepper cannot reduce the time step size further.
362:
362:   info: OGS terminated on 2020-05-27 10:19:55+0200.
362:
362:   error: OGS terminated with error.
```

waltherm:
Thanks, @bilke. I also had a talk with @TomFischer and we decided to try using the exact same mesh as HYDRUS does (= resolution 1cm, ie. half the size of the current mesh), but I think that this still will not suffice as we are most likely formulating the Richards equation differently and definitely use a different approach to solve it (which therefore may require a different mesh than the original one). 

It seems pure luck whether the setup converges at a specific system with a fixed time step and I am trying around with this original mesh and a constant time step, but couldn't find a converging time step size yet.

bilke:
What about testing with variable time stepping on just one platform (Linux)?

waltherm:
@bilke: Wouldn't this actually be against the intention of a benchmark? If we cannot test that the setup runs on different systems, we will not see whether it is affected by any changes... we could, however, setup several benchmarks for different systems and I had to verify all results once. Would this be possible?

endJunction:
Rebased. Changes in last new commit only.

Setting tolerances tight for the moment, will see how it runs on other platforms. Probably will need to go down to the order non-linear solver accuracy.

TomFischer:
Not so nice, but a bit better: Creating for every CI platform its own test that is only enabled on this platform. What do you think?

waltherm:
> Not so nice, but a bit better: Creating for every CI platform its own test that is only enabled on this platform. What do you think?

Yes, this is what I meant with my last comment. But @endJunction was active with some adaptive time stepping... maybe wait for this to have an outcome?

endJunction:
So far:
@waltherm  RM does not suffer from same degradation and RF. Though implementation and the results are different. So there is something wrong with the implementation.

@wenqing Running the `Parabolic/Richards/infiltration_1D/1d-infiltration-RM.prj` test from this PR does not output all requested times. Actually not all of the fixed time steps are taken. Can you solve this issue?

Tests/Data/Parabolic/Richards/infiltration_1D/1d-infiltration-RM.prj
```diff
@@ -0,0 +1,315 @@
...
+            </process>
+        </processes>
+        <output>
+            <type>VTK</type>
+            <prefix>1d-infiltration-RM</prefix>
+            <suffix>_t_{:gtime}</suffix>
+            <timesteps>
+                <pair>
+                    <repeat>1</repeat>
+                    <each_steps>10000</each_steps>
```
endJunction, line 154: @waltherm Change this to
```xml
<repeat>100000</repeat>
<each_steps>1</each_steps>
```
for full output.

waltherm: Yes, I set this before to have output during a fixed time stepping at specific times.

waltherm:
> @waltherm RM does not suffer from same degradation and RF. Though implementation and the results are different. So there is something wrong with the implementation. Can you run the RM version of the test and do your comparisons?

Thanks for testing. I will try to run the RM version and check the output.

When you write "not [the] same degradation", what deviation can you see in RM?

wenqing:
> 
> @wenqing Running the `Parabolic/Richards/infiltration_1D/1d-infiltration-RM.prj` test from this PR does not output all requested times. Actually not all of the fixed time steps are taken. Can you solve this issue?

I am going to take a look.

endJunction:
> When you write "not [the] same degradation", what deviation can you see in RM?

1e-11

waltherm:
I have run the RM setup and found that the results are close to those of OGS5:
compare_ogs5-6-hydrus_RM
(I can't figure out how to have output after t>44400s, though.)

That being said, we still need to decide about which our reference case should be. From my point of view, the results are fine for RM, being close to OGS5 and somewhat similar to HYDRUS. I guess, a different setup should be used to see whether differences are similar or problem-specific (hopefully not depending on the implementation).

endJunction:
The RM implementation does not handle 1D cases because of mechanics.

Regarding the output for t>44400s, feel free to push updated project here, I'll have a look.

waltherm:
> The RM implementation does not handle 1D cases because of mechanics.

Ok.

> Regarding the output for t>44400s, feel free to push updated project here, I'll have a look.

True, sorry I forgot to push the prj.

endJunction:
> (I can't figure out how to have output after t>44400s, though.)

Confirm. There is no output after that time. This is related to the time stepper providing wrong increments to reach the fixed time. Same bug as mentioned above => @wenqing 


For now change lines ca. 153+ to
```xml
                    <repeat>10000</repeat>
                    <each_steps>1</each_steps>
```
Then you can plot using all taken time steps resulting in much finer time resolution in the graphs you posted above.

wenqing:
> @wenqing Running the `Parabolic/Richards/infiltration_1D/1d-infiltration-RM.prj` test from this PR does not output all requested times. Actually not all of the fixed time steps are taken. Can you solve this issue?

After calling Output::shallDoOutput, the top entry of _fixed_output_times is popped out. The reason is then caused by somewhere unnecessary calling Output::shallDoOutput, e.g. in a rejected step.
```
    const double specific_time = _fixed_output_times.back();
    const double zero_threshold = std::numeric_limits<double>::min();
    if (std::fabs(specific_time - t) < zero_threshold)
    {
        _fixed_output_times.pop_back();
        make_output = true;
    }
```
 I am looking for the exact place for that unnecessary calling, or I think about to not change the vector of _fixed_output_times.

endJunction:
> I am looking for the exact place for that unnecessary calling, or I think about to not change the vector of _fixed_output_times.

Tried this. This is responsible for the rejected fixed output time step not being output, but the problem is that the fixed time steps are not reached. This is problem in the time stepper, not output.
For the output I shall push the changes for review.

PR #2945, opened 2020-05-14
regex project file parameter 

bilke:
Adds a `regex` project file parameter to the `vtkdiff`-section.

The main idea is to put the test tolerances/criteria together with the project files. Currently it is possible by specifying individual files to be compared, which is a little lengthy. With globbing this shall become much easier.

TODO:

- [ ] Handle parallel case (Other PR)
- [x] Migrate other benchmarks

---


Applications/ApplicationsLib/TestDefinition.cpp
```diff
@@ -215,16 +233,26 @@ TestDefinition::TestDefinition(BaseLib::ConfigTree const& config_tree,
...
-        INFO("Will run '{:s}'", command_line);
-        _command_lines.emplace_back(std::move(command_line));
+        for (auto &filename : filenames)
+        {
+            std::string const& output_filename =
+                BaseLib::joinPaths(output_directory, filename);
+            _output_files.push_back(output_filename);
+            // TODO (naumov) expand filename relative to ref path for globbing.
+            std::string const& reference_filename =
+                BaseLib::joinPaths(reference_path, filename);
```
endJunction, line 243: This looks OK to me. What do you mean by "expand filename relative to reference path"?

Applications/ApplicationsLib/TestDefinition.cpp
```diff
@@ -163,27 +166,43 @@ TestDefinition::TestDefinition(BaseLib::ConfigTree const& config_tree,
...
 
+        std::vector<std::string> filenames;
+        if (auto const regex_string =
+            //! \ogs_file_param{prj__test_definition__vtkdiff__regex}
+            vtkdiff_config.getConfigParameterOptional<std::string>("regex"))
+        {
+            // TODO: insert rank into regex for mpi case
+            DBUG("vtkdiff regex is '{}'.", *regex_string);
+            auto const regex = std::regex(*regex_string);
+            for (auto& p: fs::directory_iterator(fs::path(reference_path)))
```
endJunction, line 177: Yes, that's good. The file names are taken from the reference directory.

Applications/ApplicationsLib/TestDefinition.cpp
bilke: You added this line: https://github.com/ufz/ogs/blob/master/Applications/ApplicationsLib/TestDefinition.cpp#L184

I guess this line can be removed now?

endJunction: Oh, that was me :) Who knows what it should mean.
Can be dropped.

endJunction:
I'll provide a script to convert `AddTest` to `OgsTest` for `GLOB`bed cases...

endJunction:
So long I'm busy with the converter maybe there are some comments about the extension. Would be nice!

TomFischer:
> maybe there are some comments about the extension. Would be nice!

IMHO it is a useful extension. Will do a review when the status is changed from Draft to PR.

endJunction:
Replaced all "GLOB"bed RichardsMechanics tests with prj-file `<test_definition>` and cmake's `OgsTest`. If the reviewer agree on the changes, I'll finish the conversion for **all** processes.

Generation of the `<test_definition>` section is automatic with a script in the cmake file: see "[cmake] Parsing of GLOBbed vtkdiffs to xml." commit. The `AddTest` to `OgsTest` is manual because of additional conditions and comments.

Applications/ApplicationsLib/TestDefinition.cpp
```diff
@@ -163,27 +166,43 @@ TestDefinition::TestDefinition(BaseLib::ConfigTree const& config_tree,
...
-        std::string const& filename =
-            MeshLib::IO::getVtuFileNameForPetscOutputWithoutExtension(
-                //! \ogs_file_param{prj__test_definition__vtkdiff__file}
-                vtkdiff_config.getConfigParameter<std::string>("file")) +
-            "_" + std::to_string(rank) + ".vtu";
+            int rank;
+            MPI_Comm_rank(PETSC_COMM_WORLD, &rank);
+            int mpi_size;
+            MPI_Comm_size(PETSC_COMM_WORLD, &mpi_size);
+            std::string const& filename =
```
TomFischer, line 194: I don't understand why filename should be a reference. The referenced object on the right hand side is constructed.

Applications/ApplicationsLib/TestDefinition.cpp
TomFischer, line 177: Is the following possible?
```suggestion
            for (auto const & p: fs::directory_iterator(fs::path(reference_path)))
```

Applications/ApplicationsLib/TestDefinition.cpp
endJunction: Why not. The lifetime of the returned temporary (if it is a temporary) is extended to the lifetime of the reference. The copy happens in line 204...
The `getVtuFileName...()` could return a reference to something local.

Applications/ApplicationsLib/TestDefinition.cpp
```diff
@@ -215,16 +233,25 @@ TestDefinition::TestDefinition(BaseLib::ConfigTree const& config_tree,
...
-        // Construct command line.
-        //
-        std::string command_line =
-            vtkdiff + " -a " + safeString(field_name) + " -b " +
-            safeString(field_name) + " " + safeString(reference_filename) +
-            " " + safeString(output_filename) + " " +
-            absolute_tolerance_parameter + " " + relative_tolerance_parameter;
-        INFO("Will run '{:s}'", command_line);
-        _command_lines.emplace_back(std::move(command_line));
+        for (auto &filename : filenames)
```
endJunction, line 236:
```suggestion
        for (auto const &filename : filenames)
```

endJunction:
All `GLOB`bed ctests but LIE/cohezive zone converted. The particular test uses different name for the two fields, which is not yet supported by the `<test_definition>`s.

PR #2949, opened 2020-05-18
Add PVD output for all output domains.

TomFischer:

**Update** Extracted PR #2952 and PR #2953 from this PR. Please review first PR #2952, then PR #2953. I'll rebase this PR to ease the review.


MeshLib/IO/VtkIO/PVDFile.h
```diff
@@ -33,6 +33,8 @@ class PVDFile
     //! Add a VTU file to this PVD file.
     void addVTUFile(std::string const& vtu_fname, double timestep);
 
+    std::string const& getName() const;
+
 private:
     std::string const _pvd_filename;
```
endJunction, line 39: I'd just move it `const` as it is to public section and don't add the extra `getName()`.
```suggestion
public:
    std::string const pvd_filename;
```

ProcessLib/Output/Output.cpp
```diff
@@ -110,32 +122,50 @@ Output::Output(std::string output_directory, std::string output_file_prefix,
...
+        constructPVDName(_output_directory, _output_file_prefix, process_id,
+                         mesh_name_for_output);
     auto range = _process_to_pvd_file.equal_range(&process);
     int counter = 0;
     MeshLib::IO::PVDFile* pvd_file = nullptr;
     for (auto spd_it = range.first; spd_it != range.second; ++spd_it)
     {
-        if (counter == process_id)
+        DBUG(
+            "findPVDFile: spd_it->second.getName(): '{:s}', process_id: {:d}, "
```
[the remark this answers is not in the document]
TomFischer: I'll remove this debug message. It has little value for a user.

MeshLib/IO/VtkIO/PVDFile.h
TomFischer: Yes, I know less code contains fewer errors. But sorry, I'm old school object oriented programmer. I'd like to keep it that way.

TomFischer: Changed to @endJunction suggested version. It's not only less code it is even less assembly code and thus performs better. Since the class is very simple we can do without encapsulation.

endJunction:
Any more comments?

PR #2952, opened 2020-05-20
Cleanup extracted from PR #2949

TomFischer:


bilke:
> We choose to go to the Moon in this decade and do the other things, not because they are easy, but because they are hard.

PR #2953, opened 2020-05-20
Renames extracted from PR #2949

TomFischer:
It is based on the extracted Cleanup PR #2952. As soon as PR #2952 is merged i'll rebase.


ProcessLib/Output/Output.cpp
```diff
@@ -90,7 +90,7 @@ Output::Output(std::string output_directory, std::string output_file_prefix,
                bool const output_nonlinear_iteration_results,
                std::vector<PairRepeatEachSteps> repeats_each_steps,
                std::vector<double>&& fixed_output_times,
-               ProcessOutput&& process_output,
+               OutputDataSpecification&& output_data_specification,
```
endJunction, line 93: I guess the move-ctor of `Output` can vanish too...

PR #2955, opened 2020-05-20
MFront get bulk modulus; RM solid compressibility.

endJunction:
The changes are twofold:
 - Add `getBulkModulus` function to MFront models. The function existed before but was not applicable to MFront models and is extended to handle that case.
 - RichardsMechanics uses solid compressibility `beta_SR` from now on. The previous implementation was ambigious: three of two parameters were specified in the project files: porous medium bulk modulus `K_SR`, solid bulk modulus `K_S`(computed from Young's modulus `E` and Poisson's ratio `nu`), and the Biot coefficient `alpha`. With introduction of compressibility computed from the Biot coefficient and the solid bulk modulus the ambiguity is resolved.

2. [x] Tests covering your feature were added? Updated


MaterialLib/MPL/VariableType.h
```diff
@@ -42,19 +42,20 @@ using Tensor = std::array<double, 9>;
 /// is missing, simply add it somewhere at the list, but above the last entry.
 enum class Variable : int
 {
-    concentration,
-    phase_pressure,
-    effective_pore_pressure_rate,
     capillary_pressure,
```
TomFischer, line 45: I like the new consistent sorting in the enum. It would be nice and even more consistent if you would sort also the corresponding translation function `convertStringToVariable` in the cpp file. ;-)

endJunction: Kept it for a cleanup session. Especially there half of the enum is not handled...

endJunction:
Based on Tom's comment elsewhere, the `assert(C!=nullptr)` was changed to `if (C==nullptr) OGS_FATAL(...);` The latter runs (surprisingly) faster then the assert version.
image

MaterialLib/MPL/VariableType.h
TomFischer: That is fine for me.

endJunction:
Any further comments?

PR #2956, opened 2020-05-20
[PL] Consistent naming of integr. points number.

endJunction:


endJunction:
Removed second double usage of `n_integration_points` in RM and THM. Updated the commit message.

PR #2960, opened 2020-05-25
[MPL] Rename MPL::XProperty to MPL::X for consistency.

TomFischer:
Renames:
-  `LinearProperty` to `Linear`
- `ExponentialProperty` to 'Exponential`
- `CurveProperty` to `Curve`
- `ParameterProperty` to `Parameter`
 in order to be consistent with all the other MPL property names.


MaterialLib/MPL/Properties/CreateConstantProperty.cpp
```diff
@@ -11,11 +11,12 @@
...
+#include "ConstantProperty.h"
 
 namespace MaterialPropertyLib
 {
-std::unique_ptr<Constant> createConstant(BaseLib::ConfigTree const& config)
+std::unique_ptr<ConstantProperty> createConstantProperty(
+    BaseLib::ConfigTree const& config)
 {
     //! \ogs_file_param{properties__property__type}
     config.checkConfigParameter("type", "Constant");
```
endJunction, line 22: The project file tag should be changed too, then. But let's wait for @joergbuchwald comment about the change.

endJunction: Same for the `LinearProperty`...

joergbuchwald: To be consistent, we should change the projectfile tag too, I agree. 
This quite confusing right now in the doxygen as they are called LinearProperty and ExponentialProperty there. But ogs accepts only 'Linear' and 'Exponential'. Perhaps it should accept both variants to ensure backward compatibility? Or is it still not a thing in ogs6?

TomFischer: In the project file we have the tag type within the MPL description. This makes it clear in my opinion that the type of the (MPL) property is meant. In the source code, a variable constant of the type property does not always have a quickly obvious reference to the MPL. I personally would keep the tags in the project file. If @joergbuchwald  as a user or other users would prefer a change, I can change that too.

PR #2970, opened 2020-05-28
[geolib] copy geometrical object names when duplicating geometries

rinkk:
Previously, names of objects were ignored. Now they're copied along with the geometric information.

### DIFF
--- a/GeoLib/DuplicateGeometry.cpp
+++ b/GeoLib/DuplicateGeometry.cpp
@@ -44,9 +44,7 @@ void DuplicateGeometry::duplicate(std::string const& input_name)
     new_pnts->reserve(pnts->size());
     std::transform(pnts->cbegin(), pnts->cend(), std::back_inserter(*new_pnts),
         [](GeoLib::Point* point) { return new GeoLib::Point(*point); });
-    auto pnt_name_id_map =
-        std::make_unique<std::map<std::string, std::size_t>>();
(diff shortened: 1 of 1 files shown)


GeoLib/DuplicateGeometry.cpp
```diff
@@ -43,20 +44,38 @@ void DuplicateGeometry::duplicate(std::string const& input_name)
     new_pnts->reserve(pnts->size());
     std::transform(pnts->cbegin(), pnts->cend(), std::back_inserter(*new_pnts),
         [](GeoLib::Point* point) { return new GeoLib::Point(*point); });
-    _geo_objects.addPointVec(std::move(new_pnts), _output_name);
+    auto pnt_name_id_map =
+        std::make_unique<std::map<std::string, std::size_t>>();
+    pnt_name_id_map->insert(
+        _geo_objects.getPointVecObj(input_name)->getNameIDMapBegin(),
+        _geo_objects.getPointVecObj(input_name)->getNameIDMapEnd());
```
endJunction, line 51: minor, a little shorte. `map` has a constructor accepting a range:
```suggestion
    auto pnt_name_id_map =
        std::make_unique<std::map<std::string, std::size_t>>(
        _geo_objects.getPointVecObj(input_name)->getNameIDMapBegin(),
        _geo_objects.getPointVecObj(input_name)->getNameIDMapEnd());
```

PR #2971, opened 2020-05-28
Some changes in Tests

wenqing:
* Set createTestProperty as a common function for MPL test.
* Added namespace to two common functions.
* Fixed the error that new file in Tests gives a conflict with VTK headers.

### DIFF
--- a/Tests/MaterialLib/TestMPL.h
+++ b/Tests/MaterialLib/TestMPL.h
@@ -13,9 +13,31 @@
 
 #pragma once
 
-#include "Tests/TestTools.h"
+#include <functional>
+#include <memory>
 
 #include "MaterialLib/MPL/Medium.h"
+
 namespace MPL = MaterialPropertyLib;
 
+namespace BaseLib
+{
+class ConfigTree;
+}
+
+namespace MaterialPropertyLib
+{
+class Property;
+}
+
 std::unique_ptr<MPL::Medium> createTestMaterial(std::string const& xml);
+
+namespace Tests
+{
+std::unique_ptr<MaterialPropertyLib::Property> createTestProperty(
+    const char xml[],
+    std::function<std::unique_ptr<MaterialPropertyLib::Property>(
+        BaseLib::ConfigTree const& config)>
+        createProperty);
+
+}  // namespace Tests
--- a/Tests/MaterialLib/TestMPL.h
+++ b/Tests/MaterialLib/TestMPL.h
@@ -30,10 +30,10 @@ namespace MaterialPropertyLib
 class Property;
 }
 
-std::unique_ptr<MPL::Medium> createTestMaterial(std::string const& xml);
-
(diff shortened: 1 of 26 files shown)


endJunction:
clang-format should be executed on each commit.

bilke:
Does this fix the unity build issue you described yesterday in Mattermost?

wenqing:
> Does this fix the unity build issue you described yesterday in Mattermost?

Not yet. I am trying to find the reason. If I fix it, I will add the fixing to this PR.

Tests/MaterialLib/TestMPL.h
```diff
@@ -13,9 +13,33 @@
 
 #pragma once
 
-#include "Tests/TestTools.h"
+#include <functional>
+#include <memory>
 
+#include "Tests/TestTools.h"
```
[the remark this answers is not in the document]
wenqing: Removed.

wenqing:
> clang-format should be executed on each commit.

Ran git-clang-format but it seems take no effect.

wenqing:
> Does this fix the unity build issue you described yesterday in Mattermost?

The problem
```
/usr/include/vtk/vtkAtomic.h:360:11: error: reference to ‘detail’ is ambiguous

```
 is caused by the namespace name of `detail`, which is used in VTK as well. The last commit of this PR fixed the issue.

Tests/TestTools.cpp
```diff
@@ -11,14 +11,16 @@
...
+boost::property_tree::ptree readXml(const char xml[])
 {
     boost::property_tree::ptree ptree;
     std::istringstream xml_str(xml);
     boost::property_tree::read_xml(
-        xml_str, ptree,
-        boost::property_tree::xml_parser::no_comments |
-        boost::property_tree::xml_parser::trim_whitespace);
+        xml_str, ptree, boost::property_tree::xml_parser::no_comments |
+                            boost::property_tree::xml_parser::trim_whitespace);
```
endJunction, line 22: this formatting is still rather random. Which clang version do you use?

endJunction:
Which vtk version?

Tests/TestTools.cpp
wenqing: clang-format version 10.0.0

wenqing:
> Which vtk version?

vtk-8.2.0-13

Tests/TestTools.cpp
wenqing: Ran several times, but that ugly format does not change.

wenqing:
Any further comments on this PR?

Tests/TestTools.cpp
endJunction: Should be fixed. Actually there is no change in those lines, remove them from the PR.

wenqing: Recovered to the previous format.

PR #2975, opened 2020-05-29
[netcdf] allow explicitly setting no data value

rinkk:
Allows to manually set the no data value when converting netCDF data.

It's not straightforward to automatically extract this from the file, so this allows to handle the issue in the rare cases when not the default value is used.

### DIFF
--- a/Applications/Utils/FileConverter/NetCdfConverter.cpp
+++ b/Applications/Utils/FileConverter/NetCdfConverter.cpp
@@ -164,35 +164,38 @@ static std::pair<double, double> getBoundaries(NcVar const& var)
 }
 
 static MathLib::Point3d getOrigin(NcFile const& dataset, NcVar const& var,
-                           std::vector<std::size_t> const& dim_idx_map,
-                           bool is_time_dep)
+                                  std::vector<std::size_t> const& dim_idx_map,
+                                  std::size_t const time_offset)
 {
-    std::size_t const temp_offset = (is_time_dep) ? 1 : 0;
     MathLib::Point3d origin(MathLib::ORIGIN);
     std::size_t const n_dims = var.getDimCount();
-    for (std::size_t i = temp_offset; i < n_dims; ++i)
+    for (std::size_t i = time_offset; i < n_dims; ++i)
     {
         NcVar const& dim = getDimVar(dataset, var, dim_idx_map[i]);
         auto const bounds = (dim.isNull()) ? getDimLength(var, dim_idx_map[i])
                                            : getBoundaries(dim);
-        origin[i - temp_offset] =
+        origin[i - time_offset] =
             (bounds.first < bounds.second) ? bounds.first : bounds.second;
     }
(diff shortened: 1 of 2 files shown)


Applications/Utils/FileConverter/NetCdfConverter.cpp
```diff
@@ -620,6 +622,12 @@ int main(int argc, char* argv[])
             "(http://www.opengeosys.org)",
         ' ', GitInfoLib::GitInfo::ogs_version);
 
+    TCLAP::ValueArg<int> arg_nodata(
+        "n", "nodata",
+        "explicitely specifies the no data value used in the dataset (usually it not necessary to set this)",
```
TomFischer, line 627:
```suggestion
        "explicitely specifies the no data value used in the dataset (usually it is not necessary to set this)",
```

Applications/Utils/FileConverter/NetCdfConverter.cpp
```diff
@@ -620,6 +622,12 @@ int main(int argc, char* argv[])
...
+        false, -9999, "integer specifying no data value");
```
endJunction, line 628: better to use the above constant in place of `-9999`, just in case it changes.

Applications/Utils/FileConverter/NetCdfConverter.cpp
```diff
@@ -31,7 +31,8 @@
 
 using namespace netCDF;
 
-const double no_data = -9999;
+const double no_data_output = -9999;
+double no_data_input = -9999;
```
endJunction, line 35:
```suggestion
static constexpr double no_data_output = -9999;
static double no_data_input = -9999;
```
`no_data_input` could be local I think. As well as no_data_output. But at least both must have internal linkage.

Applications/Utils/FileConverter/NetCdfConverter.cpp
rinkk: ✔️

Applications/Utils/FileConverter/NetCdfConverter.cpp
rinkk: ✔️

Applications/Utils/FileConverter/NetCdfConverter.cpp
rinkk: Both are used in different places. I'd prefer to keep them in the beginning of the file.  
I've set both ```static```, though.

rinkk:
Question: one of the NetCDF tests is currently failing because now an issue with flipping rasters pops up because I hadn't considered before that flipping 3D rasters might be necessary.

I've fixed the issue already locally ... should I add the fix to *this* PR or open a new one for that?

endJunction:
Sure, fix it here, otherwise the master will fail too...
I assume you refer to the `NetCdfConverter-NetCDF_3D_Test-vtkdiff` test.

PR #2979, opened 2020-06-03
Fix some of the CppCheck warnings.

TomFischer:


Applications/DataExplorer/VtkVis/VtkVisPipelineView.cpp
```diff
@@ -85,26 +85,25 @@ void VtkVisPipelineView::contextMenuEvent( QContextMenuEvent* event )
...
         QAction* addFilterAction = menu.addAction("Add filter...");
 
-        QAction* addLUTAction(nullptr);
-        QAction* addMeshingAction(nullptr);
         if (objectType == VTK_IMAGE_DATA)
         {
             // this exception is needed as image object are only displayed in the vis-pipeline
             isSourceItem = false;
-            addMeshingAction = menu.addAction("Convert Image to Mesh...");
+            QAction* addMeshingAction =
```
rinkk, line 92: Have you tested this? I suspect this won't work if the UI file doesn't find the QActions to build the menu?!  
(Some for all the other QAction-ptrs.)

Applications/DataExplorer/VtkVis/VtkVisTabWidget.cpp
```diff
@@ -326,16 +326,15 @@ void VtkVisTabWidget::buildProportiesDialog(VtkVisPipelineItem* item)
...
             if (!values.empty())
             {
                 QList<QString> valuesAsString;
                 foreach (QVariant variant, values)
                 valuesAsString.push_back(variant.toString());
 
-                vectorEdit = new VtkAlgorithmPropertyVectorEdit(valuesAsString, key,
-                                                                values.front().type(),
-                                                                algProps);
+                VtkAlgorithmPropertyVectorEdit* vectorEdit =
```
rinkk, line 335: Have you tested this? Does the UI still find the array?

Applications/DataExplorer/VtkVis/VtkVisPipelineView.cpp
TomFischer: I compiled the DataExplorer and the testrunner. The testrunner worked fine. Is there another test for this?

MeshLib/convertMeshToGeo.cpp
```diff
@@ -133,12 +133,12 @@ bool convertMeshToGeo(const MeshLib::Mesh& mesh,
         addElementToSurface(*elements[i], id_map, *(*sfcs)[surfaceId]);
     }
 
-    std::for_each(sfcs->begin(), sfcs->end(), [](GeoLib::Surface* sfc) {
+    std::for_each(sfcs->begin(), sfcs->end(), [](GeoLib::Surface*& sfc) {
         if (sfc->getNumberOfTriangles() == 0)
         {
             delete sfc;
+            sfc = nullptr;
```
rinkk, line 140: This is tricky. *Technically* it shouldn't be possible to add surfaces without triangles. So most likely the test here is not necessary. However, if this *does* happen, then the surface shouldn't be removed here. It'll mess up the Data Explorer if there's suddenly a nullptr in a loaded geometry out of nowhere.

TomFischer: Yes, that is true. But two lines below the vector is cleaned up. The problem here is: The surface containing zero triangles is deleted, but the pointer isn't changed. So the code below doesn't tidy up the vector. IMHO it is possible to remove the entire `for_each` construct and the following lines. What do you think?

Applications/DataExplorer/VtkVis/VtkVisPipelineView.cpp
bilke: I guess this works, as the menu is populated during runtime (`menu.addAction()`).

endJunction:
Nice! :fast_forward:

Applications/DataExplorer/VtkVis/VtkVisTabWidget.cpp
endJunction: Is this (and actually the same question above) still open, or can we merge?

rinkk: @endJunction I haven't had any feedback regarding the PropertyVector.

Tom tested the QAction-thing and it seems to be working fine.

MeshLib/MeshGenerators/MeshGenerator.cpp
```diff
@@ -30,19 +30,30 @@ std::vector<MeshLib::Node*> MeshGenerator::generateRegularNodes(
     const std::vector<const std::vector<double>*> &vec_xyz_coords,
     const MathLib::Point3d& origin)
 {
+    auto const shift_coordinates = [](auto const& in, auto& out,
```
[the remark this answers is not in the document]
endJunction: all elements of `in` are shifted by `shift` distance and the result is stored in `out`.
On it self it does nothing - it's just a variable holding a lambda. It is called below (line 42) to transform coordinates.

Applications/DataExplorer/VtkVis/VtkVisTabWidget.cpp
TomFischer: @rinkk What are the necessary steps to test if the changes are okay?

rinkk: Apply any filter with a multi-value parameter to a dataset to see if the widgets are displayed correctly.

E.g. apply a threshold filter to a mesh and check if two line edits are displayed for the "Range"-parameter.

TomFischer: Tested this, threshold filter works. So my changes don't destroy anything in principle.

PR #2962, opened 2020-05-25
Private data member renaming: use underscore as name suffix

wenqing:
As tilted.

The fixes #2762.

Do we need name suffix or prefix? The trailing underscore makes name really ugly. Any modern editor can highlight the data members.

PR #2980, opened 2020-06-03
Part 1 of PR #2962: Application, BaseLib, GeoLib, MathLib, MeshLib

wenqing:
as titled.


endJunction:
5000 is still too much. Make it 1000.

wenqing:
> 5000 is still too much. Make it 1000.

Then for compilation and test, some of private members have not to be touched for a later change back. This makes the change a little mess. How do you think my way of the checking (see it in PR #2962).

wenqing:
Once these basic stuffs are merged, the other separated PRs can be relatively very small.

PR #2988, opened 2020-06-08
Pass the geometry dimension to functions for creating properties

wenqing:
As titled,  `ProjectData::_mesh_vec[0]->getDimension()` is passed into the functions for creating properties.

There exists a  property creation function, 
 [CreatePermeabilityOrthotropicPowerLaw.cpp](https://github.com/ufz/ogs/blob/master/MaterialLib/MPL/Properties/CreatePermeabilityOrthotropicPowerLaw.cpp#L64-L82), 
which use the size of the input parameter as the geometry dimension. For properties that such parameter size may not be identical to the the geometry dimension, such consideration is invalid. Besides, the input info should be made as unique as possible. Therefore, the changes in this PR were made.

PR #2989, opened 2020-06-08
PermeabilityMohrCoulombFailureIndexModel

wenqing:
As titled.

1. [x] Feature description was added to the [changelog](https://github.com/ufz/ogs/wiki/Release-notes-6.3.1).
2. [x] Tests covering your feature were added? Yes, it does. An unit test is added.
3. [x] Any new feature or behavior change was documented? Yes, it is. The formulas of the model and some references are provided in the source documentation, the input parameters are documented as well.

Reviewing can be started once PR #2971 and PR #2988 are merged.

Next PR: benchmark.

### DIFF
--- a/Documentation/ProjectFile/properties/property/PermeabilityMohrCoulombFailureIndexModel/c_PermeabilityMohrCoulombFailureIndexModel.md
+++ b/Documentation/ProjectFile/properties/property/PermeabilityMohrCoulombFailureIndexModel/c_PermeabilityMohrCoulombFailureIndexModel.md
@@ -1,0 +1,1 @@
+\copydoc MaterialPropertyLib::PermeabilityMohrCoulombFailureIndexModel
--- a/Documentation/ProjectFile/properties/property/PermeabilityMohrCoulombFailureIndexModel/t_initial_permeability.md
+++ b/Documentation/ProjectFile/properties/property/PermeabilityMohrCoulombFailureIndexModel/t_initial_permeability.md
@@ -1,0 +1,2 @@
+\copydoc MaterialPropertyLib::PermeabilityMohrCoulombFailureIndexModel::k0_
+Parameter name to be used as initial permeability.
--- a/MaterialLib/MPL/Properties/CreatePermeabilityMohrCoulombFailureIndexModel.cpp
+++ b/MaterialLib/MPL/Properties/CreatePermeabilityMohrCoulombFailureIndexModel.cpp
@@ -1,0 +1,83 @@
+/**
+ * \file
+ * \copyright
+ * Copyright (c) 2012-2020, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ * Created on June 5, 2020, 9:07 AM
+ */
+
+#include "CreatePermeabilityMohrCoulombFailureIndexModel.h"
+
+#include <string>
+
+#include "BaseLib/ConfigTree.h"
+#include "Parameter.h"
+#include "ParameterLib/CoordinateSystem.h"
+#include "ParameterLib/Parameter.h"
+#include "ParameterLib/Utils.h"
+#include "PermeabilityMohrCoulombFailureIndexModel.h"
+
+namespace MaterialPropertyLib
+{
+std::unique_ptr<Property> createPermeabilityMohrCoulombFailureIndexModel(
+    int const geometry_dimension,
+    BaseLib::ConfigTree const& config,
+    std::vector<std::unique_ptr<ParameterLib::ParameterBase>> const& parameters,
+    ParameterLib::CoordinateSystem const* const local_coordinate_system)
+{
+    if ((geometry_dimension != 2) && (geometry_dimension != 3))
+    {
+        OGS_FATAL(
+            "The PermeabilityMohrCoulombFailureIndexModel is implemented only "
+            "for 2D or 3D problems");
+    }
+
+    //! \ogs_file_param{properties__property__type}
+    config.checkConfigParameter("type",
+                                "PermeabilityMohrCoulombFailureIndexModel");
+
+    // Second access for storage.
+    //! \ogs_file_param{properties__property__name}
+    auto property_name = config.peekConfigParameter<std::string>("name");
+
+    DBUG("Create PermeabilityMohrCoulombFailureIndexModel property {:s}.",
+         property_name);
+
+    std::string const& parameter_name =
+        //! \ogs_file_param{properties__property__PermeabilityMohrCoulombFailureIndexModel__initial_permeability}
+        config.getConfigParameter<std::string>(
+            "initial_permeability");
+    auto const& parameter_k0 = ParameterLib::findParameter<double>(
+        parameter_name, parameters, 0, nullptr);
+
+    auto const kr =
+        //! \ogs_file_param{properties__property__PermeabilityMohrCoulombFailureIndexModel__reference_permeability}
+        config.getConfigParameter<double>("reference_permeability");
+    auto const b =
+        //! \ogs_file_param{properties__property__PermeabilityMohrCoulombFailureIndexModel__fitting_factor}
+        config.getConfigParameter<double>("fitting_factor");
+    auto const c =
+        //! \ogs_file_param{properties__property__PermeabilityMohrCoulombFailureIndexModel__cohesion}
+        config.getConfigParameter<double>("cohesion");
+    auto const phi =
+        //! \ogs_file_param{properties__property__PermeabilityMohrCoulombFailureIndexModel__friction_angle}
+        config.getConfigParameter<double>("friction_angle");
+    auto const max_k =
+        //! \ogs_file_param{properties__property__PermeabilityMohrCoulombFailureIndexModel__maximum_permeability}
+        config.getConfigParameter<double>("maximum_permeability");
+
+    if (geometry_dimension == 2)
+    {
+        return std::make_unique<PermeabilityMohrCoulombFailureIndexModel<2>>(
+            std::move(property_name), parameter_k0, kr, b, c, phi, max_k,
+            local_coordinate_system);
+    }
+
+    return std::make_unique<PermeabilityMohrCoulombFailureIndexModel<3>>(
+        std::move(property_name), parameter_k0, kr, b, c, phi, max_k,
+        local_coordinate_system);
+}
+}  // namespace MaterialPropertyLib
--- a/MaterialLib/MPL/Properties/CreatePermeabilityMohrCoulombFailureIndexModel.cpp
+++ b/MaterialLib/MPL/Properties/CreatePermeabilityMohrCoulombFailureIndexModel.cpp
@@ -68,16 +68,19 @@ std::unique_ptr<Property> createPermeabilityMohrCoulombFailureIndexModel(
     auto const max_k =
         //! \ogs_file_param{properties__property__PermeabilityMohrCoulombFailureIndexModel__maximum_permeability}
         config.getConfigParameter<double>("maximum_permeability");
+    auto const t_sigma_max =
+        //! \ogs_file_param{properties__property__PermeabilityMohrCoulombFailureIndexModel__tensile_strength_parameter}
+        config.getConfigParameter<double>("tensile_strength_parameter");
 
     if (geometry_dimension == 2)
     {
         return std::make_unique<PermeabilityMohrCoulombFailureIndexModel<2>>(
             std::move(property_name), parameter_k0, kr, b, c, phi, max_k,
-            local_coordinate_system);
+            t_sigma_max, local_coordinate_system);
     }
 
     return std::make_unique<PermeabilityMohrCoulombFailureIndexModel<3>>(
         std::move(property_name), parameter_k0, kr, b, c, phi, max_k,
-        local_coordinate_system);
+        t_sigma_max, local_coordinate_system);
 }
 }  // namespace MaterialPropertyLib
--- a/MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
+++ b/MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
@@ -31,7 +31,7 @@ PermeabilityMohrCoulombFailureIndexModel<DisplacementDim>::
     PermeabilityMohrCoulombFailureIndexModel(
         std::string name, ParameterLib::Parameter<double> const& k0,
         double const kr, double const b, double const c, double const phi,
-        double const k_max,
+        double const k_max, double const t_sigma_max,
         ParameterLib::CoordinateSystem const* const local_coordinate_system)
     : k0_(k0),
       kr_(kr),
@@ -39,8 +39,20 @@ PermeabilityMohrCoulombFailureIndexModel<DisplacementDim>::
       c_(c),
       phi_(boost::math::constants::degree<double>() * phi),
       k_max_(k_max),
+      t_sigma_max_(t_sigma_max),
       local_coordinate_system_(local_coordinate_system)
 {
+    const double t_sigma_upper = c_ / std::tan(phi_);
+    if (t_sigma_max_ <= 0.0 || t_sigma_max_ > t_sigma_upper ||
+        std::fabs(t_sigma_max_ - t_sigma_upper) <
+            std::numeric_limits<double>::epsilon())
+    {
+        OGS_FATAL(
+            "Tensile strength parameter of {:e} is out of the range (0, "
+            "c/tan(phi)) = (0, {:e})",
+            t_sigma_max_, t_sigma_upper);
+    }
+
     name_ = std::move(name);
 }
 
@@ -85,15 +97,24 @@ PermeabilityMohrCoulombFailureIndexModel<DisplacementDim>::value(
     }
 
     double const sigma_m = 0.5 * (sigma[2] + sigma[0]);
-    double tau_tt = c_ * std::cos(phi_) - sigma_m * std::sin(phi_);
-    const double apex_cut_offset = 0.001;
-    if (std::fabs(tau_tt) < apex_cut_offset)
-    {
-        tau_tt = apex_cut_offset * c_ * std::cos(phi_);
-    }
 
-    // +- tau_t = tau_tt
-    double const f = 0.5 * std::fabs(sigma[2] - sigma[0]) / tau_tt;
+    double const tau_m = 0.5 * std::fabs(sigma[2] - sigma[0]);
+    double f = 0.0;
+    if (sigma_m > t_sigma_max_)
+    {
+        // tensile failure criterion
+        f = sigma_m / t_sigma_max_;
+
+        double const tau_tt =
+            c_ * std::cos(phi_) - t_sigma_max_ * std::sin(phi_);
+
+        f = std::max(f, tau_m / tau_tt);
+    }
+    else
+    {
+        // von Mohr Coulomb failure criterion
+        f = tau_m / (c_ * std::cos(phi_) - sigma_m * std::sin(phi_));
+    }
 
     if (f >= 1.0)
     {
--- a/MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
+++ b/MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
@@ -112,7 +112,7 @@ PermeabilityMohrCoulombFailureIndexModel<DisplacementDim>::value(
     }
     else
     {
-        // von Mohr Coulomb failure criterion
+        // Mohr Coulomb failure criterion
         f = tau_m / (c_ * std::cos(phi_) - sigma_m * std::sin(phi_));
     }
 
--- a/MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
+++ b/MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
@@ -1,0 +1,143 @@
+/**
+ * \file
+ * \copyright
+ * Copyright (c) 2012-2020, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ * Created on June 4, 2020, 10:13 AM
+ */
+
+#include "PermeabilityMohrCoulombFailureIndexModel.h"
+
+#include <boost/math/constants/constants.hpp>
+#include <cmath>
+#include <limits>
+
+#include "BaseLib/Error.h"
+#include "MaterialLib/MPL/Medium.h"
+#include "MaterialLib/MPL/Utils/FormEigenTensor.h"
+#include "MaterialLib/MPL/Utils/GetSymmetricTensor.h"
+#include "MathLib/KelvinVector.h"
+#include "MathLib/MathTools.h"
+#include "ParameterLib/CoordinateSystem.h"
+#include "ParameterLib/Parameter.h"
+
+namespace MaterialPropertyLib
+{
+template <int DisplacementDim>
+PermeabilityMohrCoulombFailureIndexModel<DisplacementDim>::
+    PermeabilityMohrCoulombFailureIndexModel(
+        std::string name, ParameterLib::Parameter<double> const& k0,
+        double const kr, double const b, double const c, double const phi,
+        double const k_max,
+        ParameterLib::CoordinateSystem const* const local_coordinate_system)
+    : k0_(k0),
+      kr_(kr),
+      b_(b),
+      c_(c),
+      phi_(boost::math::constants::degree<double>() * phi),
+      k_max_(k_max),
+      local_coordinate_system_(local_coordinate_system)
+{
+    name_ = std::move(name);
+}
+
+template <int DisplacementDim>
+void PermeabilityMohrCoulombFailureIndexModel<DisplacementDim>::checkScale()
+    const
+{
+    if (!std::holds_alternative<Medium*>(scale_))
+    {
+        OGS_FATAL(
+            "The property 'PermeabilityMohrCoulombFailureIndexModel' is "
+            "implemented on the 'medium' scale only.");
+    }
+}
+
+template <int DisplacementDim>
+PropertyDataType
+PermeabilityMohrCoulombFailureIndexModel<DisplacementDim>::value(
+    VariableArray const& variable_array,
+    ParameterLib::SpatialPosition const& pos, double const t,
+    double const /*dt*/) const
+{
+    auto const& stress_vector = std::get<SymmetricTensor<DisplacementDim>>(
+        variable_array[static_cast<int>(Variable::stress)]);
+
+    auto const& stress_tensor =
+        formEigenTensor<3>(static_cast<PropertyDataType>(stress_vector));
+
+    Eigen::SelfAdjointEigenSolver<Eigen::Matrix<double, 3, 3>>
+        eigenvalue_solver(stress_tensor);
+
+    // Principle stress
+    auto const sigma = eigenvalue_solver.eigenvalues();
+
+    auto k_data = k0_(t, pos);
+
+    double const max_sigma = std::max(std::fabs(sigma[0]), std::fabs(sigma[2]));
+
+    if (max_sigma < std::numeric_limits<double>::epsilon())
+    {
+        return fromVector(k_data);
+    }
+
+    double const sigma_m = 0.5 * (sigma[2] + sigma[0]);
+    double tau_tt = c_ * std::cos(phi_) - sigma_m * std::sin(phi_);
+    const double apex_cut_offset = 0.001;
+    if (std::fabs(tau_tt) < apex_cut_offset)
+    {
+        tau_tt = apex_cut_offset * c_ * std::cos(phi_);
+    }
+
+    // +- tau_t = tau_tt
+    double const f = 0.5 * std::fabs(sigma[2] - sigma[0]) / tau_tt;
+
+    if (f >= 1.0)
+    {
+        const double exp_value = std::exp(b_ * f);
+        for (auto& k_i : k_data)
+        {
+            k_i = std::min(k_i + kr_ * exp_value, k_max_);
+        }
+    }
+
+    // Local coordinate transformation is only applied for the case that the
+    // initial intrinsic permeability is given with orthotropic assumption.
+    if (local_coordinate_system_ && (k_data.size() == DisplacementDim))
+    {
+        Eigen::Matrix<double, DisplacementDim, DisplacementDim> const e =
+            local_coordinate_system_->transformation<DisplacementDim>(pos);
+        Eigen::Matrix<double, DisplacementDim, DisplacementDim> k =
+            Eigen::Matrix<double, DisplacementDim, DisplacementDim>::Zero();
+
+        for (int i = 0; i < DisplacementDim; ++i)
+        {
+            Eigen::Matrix<double, DisplacementDim, DisplacementDim> const
+                ei_otimes_ei = e.col(i) * e.col(i).transpose();
+
+            k += k_data[i] * ei_otimes_ei;
+        }
+        return k;
+    }
+
+    return fromVector(k_data);
+}
+
+template <int DisplacementDim>
+PropertyDataType
+PermeabilityMohrCoulombFailureIndexModel<DisplacementDim>::dValue(
+    VariableArray const& /*variable_array*/, Variable const /*variable*/,
+    ParameterLib::SpatialPosition const& /*pos*/, double const /*t*/,
+    double const /*dt*/) const
+{
+    OGS_FATAL(
+        "The derivative of the intrinsic permeability k(sigma, ...) with "
+        "respect to stress tensor (sigma) is not implemented because that "
+        "dk/du is normally omitted.");
+}
+template class PermeabilityMohrCoulombFailureIndexModel<2>;
+template class PermeabilityMohrCoulombFailureIndexModel<3>;
+}  // namespace MaterialPropertyLib
--- a/MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.h
+++ b/MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.h
@@ -36,23 +36,45 @@ namespace MaterialPropertyLib
  * \f$k_\text{r}\f$ is a reference permeability,
  * \f$b\f$ is a fitting parameter.
  * \f$k_\text{r}\f$ and \f$b\f$ can be calibrated by experimental data.
+ *
  * The failure index \f$f\f$ is calculated from  the Mohr Coulomb failure
- * criterion comparing an acting shear stress. The Mohr Coulomb failure
+ * criterion comparing an acting shear stress for the shear dominated failure.
+ * The tensile failure is governed by an input parameter of
+ * tensile_strength_parameter .
+ *
+ *  The Mohr Coulomb failure
  * criterion \cite labuz2012mohr takes
  * the form
  *   \f[\tau(\sigma)=c-\sigma \mathrm{tan} \phi\f]
  *   with \f$\tau\f$ the shear stress, \f$c\f$ the cohesion, \f$\sigma\f$ the
  * tensile stress, and \f$\phi\f$ the internal friction angle.
  *
- *  The failure index is calculated by
+ *  The failure index of the Mohr Coulomb model is calculated by
  *   \f[
- *         f=\frac{\tau_m }{\cos(\phi)\tau(\sigma_m)}
+ *         f_{MC}=\frac{\tau_m }{\cos(\phi)\tau(\sigma_m)}
  *   \f]
  *   with
  *   \f$\tau_m=(\sigma_3-\sigma_1)/2\f$
  *   and \f$\sigma_m=(\sigma_1+\sigma_3)/2\f$,
  *  where \f$\sigma_1\f$ and \f$\sigma_3\f$ are the minimum and maximum shear
  * stress, respectively.
+ *
+ * The tensile failure index is calculated by
+ *  \f[
+ *    f_{t} = \sigma_m / \sigma^t_{max}
+ * \f]
+ * with, \f$\sigma^t_{max} < c \tan(\phi) \f$, a parameter of tensile strength for the cutting
+ * of the apex of the Mohr Coulomb model.
+ *
+ * The tensile stress status is determined by a condition of \f$\sigma_m>
+ * \sigma^t_{max}\f$. The failure index is then calculated by
+ * \f[
+ *   f =
+ *  \begin{cases}
+ *    f=f_{MC}, & \sigma_{m} <\sigma^t_{max}\\
+ *    f=max(f_{MC}, f_t), & \sigma_{m} \geq \sigma^t_{max}\\
+ *  \end{cases}
+ * \f]
  *
  *  Note: the conventional mechanics notations are used, which mean that tensile
  * stress is positive.
@@ -65,7 +87,7 @@ public:
     PermeabilityMohrCoulombFailureIndexModel(
         std::string name, ParameterLib::Parameter<double> const& k0,
         double const kr, double const b, double const c, double const phi,
-        double const k_max,
+        double const k_max, double const t_sigma_max,
         ParameterLib::CoordinateSystem const* const local_coordinate_system);
 
     void checkScale() const override;
@@ -92,6 +114,9 @@ private:
 
     /// Maximum permeability.
     double const k_max_;
+
+    /// Tensile strength parameter.
+    double const t_sigma_max_;
     ParameterLib::CoordinateSystem const* const local_coordinate_system_;
 };
 
--- a/MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.h
+++ b/MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.h
@@ -1,0 +1,98 @@
+/**
+ * \file
+ * \copyright
+ * Copyright (c) 2012-2020, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ * Created on June 4, 2020, 10:13 AM
+ */
+
+#pragma once
+
+#include "MaterialLib/MPL/Property.h"
+#include "MaterialLib/MPL/VariableType.h"
+
+namespace ParameterLib
+{
+struct CoordinateSystem;
+template <typename T>
+struct Parameter;
+}  // namespace ParameterLib
+
+namespace MaterialPropertyLib
+{
+/**
+ * \brief A failure index dependent permeability model \cite wangetalPerm2020
+ *
+ * \f[ \mathbf{k} =
+ *      \mathbf{k}_0+ H(f-1) k_\text{r} \mathrm{e}^{b f}\mathbf{I}\f]
+ *
+ * where
+ *   \f$\mathbf{k}_0\f$ is the intrinsic permeability
+ *  of the undamaged material,
+ * \f$H\f$ is the Heaviside step function, \f$f\f$ is the failure index,
+ * \f$k_\text{r}\f$ is a reference permeability,
+ * \f$b\f$ is a fitting parameter.
+ * \f$k_\text{r}\f$ and \f$b\f$ can be calibrated by experimental data.
+ * The failure index \f$f\f$ is calculated from  the Mohr Coulomb failure
+ * criterion comparing an acting shear stress. The Mohr Coulomb failure
+ * criterion \cite labuz2012mohr takes
+ * the form
+ *   \f[\tau(\sigma)=c-\sigma \mathrm{tan} \phi\f]
+ *   with \f$\tau\f$ the shear stress, \f$c\f$ the cohesion, \f$\sigma\f$ the
+ * tensile stress, and \f$\phi\f$ the internal friction angle.
+ *
+ *  The failure index is calculated by
+ *   \f[
+ *         f=\frac{\tau_m }{\cos(\phi)\tau(\sigma_m)}
+ *   \f]
+ *   with
+ *   \f$\tau_m=(\sigma_3-\sigma_1)/2\f$
+ *   and \f$\sigma_m=(\sigma_1+\sigma_3)/2\f$,
+ *  where \f$\sigma_1\f$ and \f$\sigma_3\f$ are the minimum and maximum shear
+ * stress, respectively.
+ *
+ *  Note: the conventional mechanics notations are used, which mean that tensile
+ * stress is positive.
+ *
+ */
+template <int DisplacementDim>
+class PermeabilityMohrCoulombFailureIndexModel final : public Property
+{
+public:
+    PermeabilityMohrCoulombFailureIndexModel(
+        std::string name, ParameterLib::Parameter<double> const& k0,
+        double const kr, double const b, double const c, double const phi,
+        double const k_max,
+        ParameterLib::CoordinateSystem const* const local_coordinate_system);
+
+    void checkScale() const override;
+
+    PropertyDataType value(VariableArray const& variable_array,
+                           ParameterLib::SpatialPosition const& pos,
+                           double const t, double const dt) const override;
+    PropertyDataType dValue(VariableArray const& variable_array,
+                            Variable const variable,
+                            ParameterLib::SpatialPosition const& pos,
+                            double const t, double const dt) const override;
+
+private:
+    /// Intrinsic permeability for undamaged material.
+    ParameterLib::Parameter<double> const& k0_;
+    /// Reference permeability.
+    double const kr_;
+    /// Fitting parameter.
+    double const b_;
+    /// Cohesion.
+    double const c_;
+    /// Angle of internal friction.
+    double const phi_;
+
+    /// Maximum permeability.
+    double const k_max_;
+    ParameterLib::CoordinateSystem const* const local_coordinate_system_;
+};
+
+}  // namespace MaterialPropertyLib
--- a/MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.h
+++ b/MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.h
@@ -47,11 +47,11 @@ namespace MaterialPropertyLib
  * the form
  *   \f[\tau(\sigma)=c-\sigma \mathrm{tan} \phi\f]
  *   with \f$\tau\f$ the shear stress, \f$c\f$ the cohesion, \f$\sigma\f$ the
- * tensile stress, and \f$\phi\f$ the internal friction angle.
+ * normal stress, and \f$\phi\f$ the internal friction angle.
  *
  *  The failure index of the Mohr Coulomb model is calculated by
  *   \f[
- *         f_{MC}=\frac{\tau_m }{\cos(\phi)\tau(\sigma_m)}
+ *         f_{MC}=\frac{|\tau_m| }{\cos(\phi)\tau(\sigma_m)}
  *   \f]
  *   with
  *   \f$\tau_m=(\sigma_3-\sigma_1)/2\f$
@@ -63,18 +63,25 @@ namespace MaterialPropertyLib
  *  \f[
  *    f_{t} = \sigma_m / \sigma^t_{max}
  * \f]
- * with, \f$\sigma^t_{max} < c \tan(\phi) \f$, a parameter of tensile strength for the cutting
- * of the apex of the Mohr Coulomb model.
+ * with, \f$0 < \sigma^t_{max} < c \tan(\phi) \f$, a parameter of tensile
+ * strength for the cutting of the apex of the Mohr Coulomb model.
  *
  * The tensile stress status is determined by a condition of \f$\sigma_m>
  * \sigma^t_{max}\f$. The failure index is then calculated by
  * \f[
  *   f =
  *  \begin{cases}
- *    f=f_{MC}, & \sigma_{m} <\sigma^t_{max}\\
- *    f=max(f_{MC}, f_t), & \sigma_{m} \geq \sigma^t_{max}\\
+ *    f_{MC}, & \sigma_{m} \leq \sigma^t_{max}\\
+ *    max(f_{MC}, f_t), & \sigma_{m} > \sigma^t_{max}\\
  *  \end{cases}
  * \f]
+ *
+ *  The computed permeability components are restricted with an upper bound,
+ *  i.e. \f$\mathbf{k}:=k_{ij} < k_{max}\f$.
+ *
+ * If \f$\mathbf{k}_0\f$ is orthogonal, i.e input two or three numbers
+ * for its diagonal entries, a coordinate system rotation of \f$\mathbf{k}\f$
+ * is possible if it is needed.
  *
  *  Note: the conventional mechanics notations are used, which mean that tensile
  * stress is positive.
@@ -101,7 +108,8 @@ public:
                             double const t, double const dt) const override;
 
 private:
-    /// Intrinsic permeability for undamaged material.
+    /// Intrinsic permeability for undamaged material. It can be a scalar or
+    /// tensor for anisotropic material.
     ParameterLib::Parameter<double> const& k0_;
     /// Reference permeability.
     double const kr_;
--- a/Tests/MaterialLib/TestMPLPermeabilityMohrCoulombFailureIndexModel.cpp
+++ b/Tests/MaterialLib/TestMPLPermeabilityMohrCoulombFailureIndexModel.cpp
@@ -9,8 +9,6 @@
  * Created on June 8, 2020, 9:17 AM
  */
 
-#pragma once
-
 #include <gtest/gtest.h>
 
 #include <Eigen/Eigen>
@@ -27,16 +25,16 @@
 
 TEST(MaterialPropertyLib, PermeabilityMohrCoulombFailureIndexModel)
 {
-    ParameterLib::ConstantParameter<double> k0 =
-        ParameterLib::ConstantParameter<double>("k0", 1.e-20);
+    ParameterLib::ConstantParameter<double> const k0("k0", 1.e-20);
     double const kr = 1.0e-19;
     double const b = 3.0;
     double const c = 1.0e+6;
     double const phi = 15.0;
     double const k_max = 1.e-10;
+    double const t_sigma_max = c;
 
     auto const k_model = MPL::PermeabilityMohrCoulombFailureIndexModel<3>(
-        "k_f", k0, kr, b, c, phi, k_max, nullptr);
+        "k_f", k0, kr, b, c, phi, k_max, t_sigma_max, nullptr);
 
     const int symmetric_tensor_size = 6;
     using SymmetricTensor = Eigen::Matrix<double, symmetric_tensor_size, 1>;
@@ -62,7 +60,7 @@ TEST(MaterialPropertyLib, PermeabilityMohrCoulombFailureIndexModel)
 
     double const k_expected = 1.1398264890628033e-15;
 
-    ASSERT_LE(std::fabs((k_expected - k(0, 0))) / k_expected, 1e-10)
+    ASSERT_LE(std::fabs(k_expected - k(0, 0)) / k_expected, 1e-10)
         << "for expected changed permeability " << k_expected
         << " and for computed changed permeability." << k(0, 0);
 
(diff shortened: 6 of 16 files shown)


MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,148 @@
...
+    Eigen::SelfAdjointEigenSolver<Eigen::Matrix<double, 3, 3>>
+        eigenvalue_solver(stress_tensor);
+
+    // Principle stress
+    auto const sigma = eigenvalue_solver.eigenvalues();
+
+    double const f =
+        std::fabs(0.5 * (sigma[0] - sigma[2]) /
+                  (c_ + 0.5 * (sigma[0] + sigma[2]) * std::tan(phi_)));
+
```
nagelt, line 86: This is not what we have in the paper ... there we have:

        0.5 * std::fabs(sigma[0] - sigma[2]) /
                  (c_ * std::(cos(phi_) + 0.5 * (sigma[0] + sigma[2]) * std::sin(phi_));

which is the actual MohrCoulomb formulation (not maximum shear stress drives failure, but worst ratio of shear stress to normal stress).

The std::fabs around everything also does not seem to make sense in the hypothetical far tension area (beyond the hydrostatic tensile strength) .... this should be treated differently even if it will rarely arise ... 

By the way, if the stress state is at this point, the denominator will become zero and you get a NaN, so this should also be caught ...

nagelt: Thinking about this, I suggest three unit tests:
-- one stress point within the yield surface (you have that)
-- one stress point outside of the yield surface (you have that)
-- one stress point beyond the apex of the yield surface (to test this "special behaviour")

MathLib/MathematicalConstant.h
```diff
@@ -0,0 +1,20 @@
...
+ * Created on June 4, 2020, 5:09 PM
+ */
+
+#pragma once
+
+namespace MathLib
+{
+namespace MathematicalConstant
+{
+constexpr double pi = 3.14159265359;
```
endJunction, line 18: Use boost for constants. No need for 25th version of that number.

MathLib/KelvinVector.h
```diff
@@ -235,6 +235,13 @@ KelvinVectorType<DisplacementDim> symmetricTensorToKelvinVector(
             1));
 }
 
+/// Conversion of a symmetric tensor, which is stored in a column of Eigen
+/// matrix with a row size of 4 or 6, to a full tensor in a 3x3 matrix
+template <int KelvinVectorSize>
+Eigen::Matrix<double, 3, 3> symmetricTensorToFullTensor(
```
endJunction, line 241: This is wrong place for this function which has nothing to do with Kelvin-mapped vectors.

MaterialLib/MPL/Properties/CreatePermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+    std::string const& parameter_name =
+        //! \ogs_file_param{properties__property__Parameter__parameter_name}
+        config.getConfigParameter<std::string>("parameter_name");
+    auto const& parameter_k0 = ParameterLib::findParameter<double>(
+        parameter_name, parameters, 0, nullptr);
+
+    auto const kr =
+        //! \ogs_file_param{properties__property__PermeabilityMohrCoulombFailureIndexModel__kr}
+        config.getConfigParameter<double>("kr");
```
endJunction, line 58: Names of the tags should be understandable w/o looking into documentation. Use long names.

MaterialLib/MPL/Properties/CreatePermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,80 @@
...
+std::unique_ptr<Property> createPermeabilityMohrCoulombFailureIndexModel(
+    int const geometry_dimension,
+    BaseLib::ConfigTree const& config,
+    std::vector<std::unique_ptr<ParameterLib::ParameterBase>> const& parameters,
+    ParameterLib::CoordinateSystem const* const local_coordinate_system)
+{
+    if (geometry_dimension == 1)
+    {
+        OGS_FATAL(
+            "The PermeabilityMohrCoulombFailureIndexModel is only for 2D or 3D "
```
endJunction, line 35:
```suggestion
            "The PermeabilityMohrCoulombFailureIndexModel is implemented only for 2D or 3D "
```

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,148 @@
...
+template <int DisplacementDim>
+void PermeabilityMohrCoulombFailureIndexModel<DisplacementDim>::checkScale()
+    const
+{
+    if (!std::holds_alternative<Phase*>(scale_))
+    {
+        OGS_FATAL(
+            "The property 'PermeabilityMohrCoulombFailureIndexModel' is "
+            "implemented on the 'phase' scales only.");
```
endJunction, line 48:
```suggestion
            "implemented on the 'phase' scale only.");
```

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.h
```diff
@@ -0,0 +1,92 @@
...
+    PropertyDataType value(VariableArray const& variable_array,
+                           ParameterLib::SpatialPosition const& pos,
+                           double const t, double const dt) const override;
+    PropertyDataType dValue(VariableArray const& variable_array,
+                            Variable const variable,
+                            ParameterLib::SpatialPosition const& pos,
+                            double const t, double const dt) const override;
+
+private:
+    /// Intrinsic permeability in undamaged status.
```
endJunction, line 78:
```suggestion
    /// Intrinsic permeability for undamaged material.
```

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+template <int DisplacementDim>
+PropertyDataType
+PermeabilityMohrCoulombFailureIndexModel<DisplacementDim>::dValue(
+    VariableArray const& /*variable_array*/, Variable const /*variable*/,
+    ParameterLib::SpatialPosition const& /*pos*/, double const /*t*/,
+    double const /*dt*/) const
+{
+    // no need to compute the derivative.
+    return 0.0;
```
endJunction, line 144: OGS_FATAL("Derivative not implemented for etc. and so forth...");
```suggestion
    OGS_FATAL("Derivative not implemented for etc. and so forth...");
```

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
wenqing: > This is not what we have in the paper ... there we have:
> 
> ```
>     0.5 * std::fabs(sigma[0] - sigma[2]) /
>               (c_ * std::(cos(phi_) + 0.5 * (sigma[0] + sigma[2]) * std::sin(phi_));
> ```
> 
> which is the actual MohrCoulomb formulation (not maximum shear stress drives failure, but worst ratio of shear stress to normal stress).
> 
> The std::fabs around everything also does not seem to make sense in the hypothetical far tension area (beyond the hydrostatic tensile strength) .... this should be treated differently even if it will rarely arise ...
> 
> By the way, if the stress state is at this point, the denominator will become zero and you get a NaN, so this should also be caught ...

Thanks. The formula will be changed to the  actual MohrCoulomb one. The test for the stress point beyond the apex  will be added.

MathLib/MathematicalConstant.h
wenqing: Thanks.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,148 @@
...
+
+    auto const& k0_variant = fromVector(k_data);
+
+    PropertyDataType k = std::visit(
+        [&](auto const& k0_variant) -> PropertyDataType {
+            using T = std::decay_t<decltype(k0_variant)>;
+            if
+                constexpr(std::is_same_v<T, double>) { return k0_variant; }
+            else
+            {
```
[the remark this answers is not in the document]
wenqing: > braces

Do you mean a pire of braces is missing?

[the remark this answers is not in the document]
endJunction: Sorry, not braces, as I see now. Bad formatting led me to wrong conclusion.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
wenqing: > The std::fabs around everything also does not seem to make sense in the hypothetical far tension area (beyond the hydrostatic tensile strength) .... this should be treated differently even if it will rarely arise ...

Removed std:fabs

wenqing: > -- one stress point beyond the apex of the yield surface (to test this "special behaviour")

Added.

MathLib/KelvinVector.h
wenqing: Moved to MathTools.h

MaterialLib/MPL/Properties/CreatePermeabilityMohrCoulombFailureIndexModel.cpp
wenqing: changed.

MaterialLib/MPL/Properties/CreatePermeabilityMohrCoulombFailureIndexModel.cpp
wenqing: changed.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
wenqing: changed.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.h
wenqing: changed.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
wenqing: Changed with a more detailed info.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,153 @@
...
+        eigenvalue_solver(stress_tensor);
+
+    // Principle stress
+    auto const sigma = eigenvalue_solver.eigenvalues();
+
+    int const max_principle_stress_id = DisplacementDim - 1;
+    double const f = std::fabs(
+        0.5 * (sigma[0] - sigma[max_principle_stress_id]) /
+        (c_ * std::cos(phi_) +
+         0.5 * (sigma[0] + sigma[max_principle_stress_id]) * std::sin(phi_)));
```
wenqing, line 80: @nagelt  New changes:
1. still use std::fabs(f)
2. for 2D plane strain problem, use sigma[0] and sigma[1]. This took me a while to figure out  for 2D benchmark.

Please take a look.

nagelt: ok ... if I understand your changes correctly, there may be a few issues to consider:

The Eigen docu says that eigenvalues are returned in increasing order so that if you use 
sigma[2] - sigma[0] for example, your maximum shear stress should remain positive simply thanks to sorting by Eigen ... if you want to stay safe, I can see why you would want to use std::fabs(f) around the maximum shear stress, but not the entire expression ... because this seems to raise some issues:

If you are on the compressive side of the MohrCoulomb apex (c/tan(phi)), then your denominator is also positive and you don't need a fabs there either ... if you're on the tensile side of the apex, introducing a fabs will decrease the distance to the maximum shear stress again (and hence f), but intuitively I'd expect f to keep increasing (because it's more damaging) ... so you need another distance measure there I think ... in other words: if your denominator turns negative, things get worse ... but taking fabs() and making it positive artificially makes things better ... this is unwanted behaviour I think

By using sigma[0] and sigma[1] you seem to choose the middle principal stress. This can vary depending on loading and material parameters (particularly Poisson's ratio and dilatancy). So if you want to ensure that the slip happens "in-plane", I think you need to pass (sigma_xx, sigma_yy, sigma_xy) to the eigendecomposition, instead of (sigma_xx,sigma_yy,sigma_zz,sigma_xy) and then dropping the highest eigenvalue (sigma[2]) ... But in the solid material models we make no such "trick" but let the stresses and boundary conditions play it out ...

wenqing: Thanks. Yes, you are right that || is not needed, and that I wanted to ensure that the slip happens "in-plane". For a plane strain test with high in-plane pressure and with off-plane stress purely caused by the Poisson ratio, f with  sigma[0] and  sigma[1]  (occasionally in-plane, i.e. eigenvalues of 2x2 matrix by sigma_xx, sigma_yy, sigma_xy ) and f   with  sigma[0] and  sigma[2] are all larger than one. I my previous consideration,  f with  sigma[0] and  sigma[1] was selected for its better magnitude for the model. Yes, the "trick" seems not for general problems. I will figure out a decent way to choose f soon.

wenqing: @nagelt 
> so all that's missing now is the case of c*cos(a) = 0.5(sig_3 + sig_1) sin(a) where the denominator vanishes (apex) to avoid a division by zero ... and how you treat the point beyond

I got a idea for the apex point: assuming a cutting of the apex by 0.0001 (or even smaller) of c  * atan(phi).

MaterialLib/MPL/Properties/CreatePermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,79 @@
...
+    if (geometry_dimension == 2)
+    {
+        return std::make_unique<PermeabilityMohrCoulombFailureIndexModel<2>>(
+            std::move(property_name), parameter_k0, kr, b, c, phi,
+            local_coordinate_system);
+    }
+
+    return std::make_unique<PermeabilityMohrCoulombFailureIndexModel<3>>(
+        std::move(property_name), parameter_k0, kr, b, c, phi,
+        local_coordinate_system);
```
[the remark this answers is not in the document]
wenqing: There is already one OGS_FATAL for `(geometry_dimension == 1)` at the beginning of the function.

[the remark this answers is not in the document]
endJunction: Could be 0, or 4, or some other error. Don't leave space for errors.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
nagelt: @wenqing 
so your denominator would be max( epsilon, (c + tan(phi))*cos(phi) ? 
this would correspond to a cutoff of f at f_max as the denominator goes to 0 (which would now let f go to infinity) ... so effectively this would mean a maximum permeability in the model (which can be practical) ... would do the trick for now, I guess. Did I understand your intention correctly?

another idea could be formulating the criterion as a difference instead of as a ratio  (no problems with division by zero or f going to infty, easier use of an Euclidean distance to the apex) but this would be a different model to the one in the paper, of course ... but conceptually similar ...

wenqing: @nagelt  The idea is when sigma_m=(sigma_0+sigma_0), set sigma_m = epsilon * sigma_m, then the dominator  is epsilon * c. I will also think about your idea of a difference.

Documentation/ProjectFile/properties/property/PermeabilityMohrCoulombFailureIndexModel/c_PermeabilityMohrCoulombFailureIndexModel.md
```diff
@@ -0,0 +1,4 @@
+\copydoc MaterialPropertyLib::PermeabilityMohrCoulombFailureIndexModel
+The intrinsic permeability in undamaged status k0,
```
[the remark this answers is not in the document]
wenqing: Changed to `The intrinsic permeability of undamaged porous media, \f$k_0\f$, is
input as Parameter by a tag of <parameter_name>`.

Documentation/ProjectFile/properties/property/PermeabilityMohrCoulombFailureIndexModel/c_PermeabilityMohrCoulombFailureIndexModel.md
```diff
@@ -0,0 +1,4 @@
...
+ MaterialPropertyLib::PermeabilityMohrCoulombFailureIndexModel::k0_, is required
+to be input as Parameter.
```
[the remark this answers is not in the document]
wenqing: Generally, it is needed.

MaterialLib/MPL/Properties/CreatePermeabilityMohrCoulombFailureIndexModel.cpp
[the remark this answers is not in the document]
wenqing: replaced the 1st if-condition with ` if (geometry_dimension != 2 || geometry_dimension != 3)`.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,154 @@
...
+PermeabilityMohrCoulombFailureIndexModel<DisplacementDim>::
+    PermeabilityMohrCoulombFailureIndexModel(
+        std::string name, ParameterLib::Parameter<double> const& k0,
+        double const kr, double const b, double const c, double const phi,
+        ParameterLib::CoordinateSystem const* const local_coordinate_system)
+    : k0_(k0),
+      kr_(kr),
+      b_(b),
+      c_(c),
+      phi_(boost::math::constants::pi<double>() * phi / 180.0),
```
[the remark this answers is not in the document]
wenqing: use it.

MathLib/MathTools.cpp
```diff
@@ -44,4 +46,46 @@ double getAngle (const double p0[3], const double p1[3], const double p2[3])
     return std::acos (scalarProduct<double,3> (v0,v1) / (std::sqrt(scalarProduct<double,3>(v0,v0)) * sqrt(scalarProduct<double,3>(v1,v1))));
 }
 
+template <>
+Eigen::Matrix<double, 3, 3> symmetricTensorToFullTensor(
```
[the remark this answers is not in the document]
wenqing: moved to a new file in MPL/utils

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,154 @@
...
+template <int DisplacementDim>
+PropertyDataType
+PermeabilityMohrCoulombFailureIndexModel<DisplacementDim>::value(
+    VariableArray const& variable_array,
+    ParameterLib::SpatialPosition const& pos, double const t,
+    double const /*dt*/) const
+{
+    constexpr int symmetric_tensor_size =
+        MathLib::KelvinVector::KelvinVectorDimensions<DisplacementDim>::value;
+    using SymmetricTensor = Eigen::Matrix<double, symmetric_tensor_size, 1>;
```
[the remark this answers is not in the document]
wenqing: The argument is different.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,154 @@
...
+        {
+            const double val = b_ * f;
+            if (val > 50.0)
+            {
+                WARN(
+                    "b*f = '{:e}' in PermeabilityMohrCoulombFailureIndexModel "
+                    "is too large",
+                    val);
+            }
+            const double exponent = (val > 50.0) ? 50.0 : val;
```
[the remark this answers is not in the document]
wenqing: It is gone.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,154 @@
...
+                ei_otimes_ei = e.col(i) * e.col(i).transpose();
+
+            k += k_data[i] * ei_otimes_ei;
+        }
+        return k;
+    }
+
+    auto const& k0_variant = fromVector(k_data);
+
+    PropertyDataType k = std::visit(
```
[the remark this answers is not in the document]
wenqing: return fromVector(k_data) directly.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
wenqing: @nagelt 
 New changes for stress point at the apex
1. introduced a parameter of maximum_permeability.
2. cut-off at the apex with a small offset,
3. if f is too large, i.e. exp(bf) is infinite, returns maximum_permeability (changed the unit test for this as well).
4. if the k = max (k,  maximum_permeability)

This would guarantee that the permeability is in a reasonable magnitude at the extremist stress statue.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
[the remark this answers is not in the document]
endJunction: Please explain. I don't see any difference.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,155 @@
...
+            if (std::isinf(exp_value))
+            {
+                k_i = k_max_;
+            }
+
+            k_i += kr_ * exp_value;
+            if (k_i > k_max_)
+            {
+                k_i = k_max_;
+            }
```
endJunction, line 112: use min/max/clamp

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,155 @@
...
+        }
+    }
+
+    // Local coordinate transformation is only applied for the case that the
+    // initial intrinsic permeability is given with orthotropic assumption.
+    if (local_coordinate_system_ &&
+        ((k_data.size() == 2) || (k_data.size() == 3)))
+    {
+        assert(k_data.size() == DisplacementDim);
```
endJunction, line 121: assert condition is always fulfilled, so moving it to the if condition would yield:
```suggestion
    if (local_coordinate_system_ && k_data.size() == DisplacementDim)
    {
```
which is simpler.

MathLib/MathTools.cpp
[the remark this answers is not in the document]
endJunction: Can you explain what difference there is between your implementation `symmetricTensorToFullTensor` and the one in [MaterialLib/MPL/Utils/FormEigenTensor.cpp](https://github.com/ufz/ogs/blob/master/MaterialLib/MPL/Utils/FormEigenTensor.cpp#L68-L98)

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
wenqing: Took it.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
wenqing: use  k_i = std::min(k_i + kr_ * exp_value, k_max_);

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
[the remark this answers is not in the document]
wenqing: I see that it is for using symmetric_tensor_size not for using getSymmetricTensor.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,154 @@
...
+    {
+        for (auto& k_i : k_data)
+        {
+            const double val = b_ * f;
+            if (val > 50.0)
+            {
+                WARN(
+                    "b*f = '{:e}' in PermeabilityMohrCoulombFailureIndexModel "
+                    "is too large",
+                    val);
```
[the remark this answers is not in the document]
wenqing: Dropped.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,145 @@
...
+    double const max_sigma = std::max(std::fabs(sigma[0]), std::fabs(sigma[2]));
+
+    if (max_sigma < std::numeric_limits<double>::epsilon())
+    {
+        return fromVector(k_data);
+    }
+
+    double const sigma_m = 0.5 * (sigma[2] + sigma[0]);
+    double tau_tt = c_ * std::cos(phi_) - sigma_m * std::sin(phi_);
+    const double apex_cut_offset = 0.001;
```
endJunction, line 87: what is this magic number? It was a machine epsilon last time....

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,145 @@
...
+        tau_tt = apex_cut_offset * c_ * std::cos(phi_);
+    }
+
+    double const f = 0.5 * (sigma[2] - sigma[0]) / tau_tt;
+
+    if (f >= 1.0)
+    {
+        for (auto& k_i : k_data)
+        {
+            const double exp_value = std::exp(b_ * f);
```
endJunction, line 99: This can be moved up, because it does not change with each iteration over k_data.

MaterialLib/MPL/Properties/CreatePermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,82 @@
...
+    // Second access for storage.
+    //! \ogs_file_param{properties__property__name}
+    auto property_name = config.peekConfigParameter<std::string>("name");
+
+    DBUG("Create PermeabilityMohrCoulombFailureIndexModel property {:s}.",
+         property_name);
+
+    std::string const& parameter_name =
+        //! \ogs_file_param{properties__property__Parameter__parameter_name}
+        config.getConfigParameter<std::string>("parameter_name");
```
endJunction, line 51: `parameter_name` is not good description of the k0 parameter.

wenqing: Changed to initial_permeability. Please see  https://github.com/wenqing/ogs/blob/permeablility_stress/MaterialLib/MPL/Properties/CreatePermeabilityMohrCoulombFailureIndexModel.cpp#L49-L54

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
wenqing: Moved.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
wenqing: It did appear before. It was introduced for that case of  tau_tt=0 with an ratio tau_tt/max_sigma. In the latest change, the dominator of max_sigma is dropped, which means a more small tolerance for the restriction.

MathLib/MathTools.cpp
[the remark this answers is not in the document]
wenqing: Come to this point later.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,146 @@
...
+    {
+        const double exp_value = std::exp(b_ * f);
+        for (auto& k_i : k_data)
+        {
+            if (std::isinf(exp_value))
+            {
+                k_i = k_max_;
+            }
+
+            k_i = std::min(k_i + kr_ * exp_value, k_max_);
```
endJunction, line 106: It appears to me now, that this case of +infinity is already covered by std::min, isn't it?
```suggestion
            k_i = std::min(k_i + kr_ * exp_value, k_max_);
```

Documentation/ProjectFile/properties/property/PermeabilityMohrCoulombFailureIndexModel/t_initial_permeability.md
```diff
@@ -0,0 +1,2 @@
+\copydoc MaterialPropertyLib::PermeabilityMohrCoulombFailureIndexModel::k0_
+It must be input as a Parameter variable.
```
endJunction, line 2:
```suggestion
Parameter name to be used as initial permeability.
```

MaterialLib/MPL/Properties/CreatePermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,83 @@
...
+    //! \ogs_file_param{properties__property__name}
+    auto property_name = config.peekConfigParameter<std::string>("name");
+
+    DBUG("Create PermeabilityMohrCoulombFailureIndexModel property {:s}.",
+         property_name);
+
+    std::string const& parameter_name =
+        //! \ogs_file_param{properties__property__PermeabilityMohrCoulombFailureIndexModel__initial_permeability}
+        config.getConfigParameter<std::string>(
+            "initial_permeability_parameter");
```
endJunction, line 52: The ogs_file_param string does not match the parsed tag.
```suggestion
        //! \ogs_file_param{properties__property__PermeabilityMohrCoulombFailureIndexModel__initial_permeability}
        config.getConfigParameter<std::string>(
            "initial_permeability");
```

Documentation/ProjectFile/properties/property/PermeabilityMohrCoulombFailureIndexModel/t_initial_permeability.md
wenqing: Took it.

MaterialLib/MPL/Properties/CreatePermeabilityMohrCoulombFailureIndexModel.cpp
wenqing: Changed.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
wenqing: Yes it is.

MathLib/MathTools.cpp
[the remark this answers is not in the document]
wenqing: Drop symmetricTensorToFullTensor and use FormEigenTensor with a type conversion for the variable for the argument.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.h
```diff
@@ -0,0 +1,98 @@
...
+struct CoordinateSystem;
+template <typename T>
+struct Parameter;
+}  // namespace ParameterLib
+
+namespace MaterialPropertyLib
+{
+/**
+ * \brief A failure index dependent permeability model \cite wangetalPerm2020
+ *
```
endJunction, line 28: Did you update documentation for the recent changes in the code. Looks like apex treatment is missing.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+    double const sigma_m = 0.5 * (sigma[2] + sigma[0]);
+    double tau_tt = c_ * std::cos(phi_) - sigma_m * std::sin(phi_);
+    const double apex_cut_offset = 0.001;
+    if (std::fabs(tau_tt) < apex_cut_offset)
+    {
+        tau_tt = apex_cut_offset * c_ * std::cos(phi_);
+    }
+
+    // +- tau_t = tau_tt
```
endJunction, line 95: What is tau_t?

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,143 @@
...
+        return fromVector(k_data);
+    }
+
+    double const sigma_m = 0.5 * (sigma[2] + sigma[0]);
+    double tau_tt = c_ * std::cos(phi_) - sigma_m * std::sin(phi_);
+    const double apex_cut_offset = 0.001;
+    if (std::fabs(tau_tt) < apex_cut_offset)
+    {
+        tau_tt = apex_cut_offset * c_ * std::cos(phi_);
+    }
```
nagelt, line 93: First apex_cut_offset is compared with fabs(tau_tt) and therefore has units of stress ... later you compute tau_tt by multiplying apex_cut_offset with c, so you get units stress squared for tau_tt. In the case below a large tensile normal stress will produce f = 0 if sigma_2 = sigma_0.

So to fix both, maybe simply introduce a tension cut-off parameter (units of stress, definition in input file), and have one shear mechanism for failure (as you already have), and one tensile-like mechanism:

f_tensile = sigma_m / tensile_strength_parameter (greater than one only if tensile_strength_parameter is exceeded)
f_shear = tau_max / (c * cos_phi - min(sigma_m, tensile_strength_parameter) * sin_phi)

f = max(f_tensile, f_shear)

so the tensile_strength_parameter is an upper limit for sigma_m here, so a vertical line in the tau-sigma-diagram ...

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
nagelt: thanks for all your changes, Wenqing. 1, 3, 4 are useful ... for 2 I left another comment, which should be a small fix but give the parameters some meaning ... then I'm ok with it!

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
wenqing: @nagelt  Thanks. Added the von Mises criterion for sigma_m > 0.5 * c /tan(phi) with a new parameter.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.h
wenqing: It was assumed to be updated once the final criterion is decided. Now, temporally added some words.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
wenqing: Just a temporary variable name. Now, it is gone.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,156 @@
...
+            int const i2 = (i + 1) % 3;
+            val += (sigma[i2] - sigma[i1]) * (sigma[i2] - sigma[i1]);
+        }
+        double J = std::sqrt(0.5 * val);
+        f = J / t_sigma_max_;
+    }
+    else
+    {
+        // von Mohr Coulomb failure criterion
+        // +- tau_t
```
endJunction, line 106: What is tau_t? It is not mentioned in the documentation, which has tau and tau_m.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,156 @@
...
+    {
+        // von Mises failure criterion
+        double val = 0.0;
+        for (int i = 0; i < 3; i++)
+        {
+            int const i1 = i % 3;
+            int const i2 = (i + 1) % 3;
+            val += (sigma[i2] - sigma[i1]) * (sigma[i2] - sigma[i1]);
+        }
+        double J = std::sqrt(0.5 * val);
```
endJunction, line 100:
```suggestion
        double const J = std::sqrt(0.5 * val);
```

wenqing: Changed.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
wenqing: Dropped that line.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,155 @@
...
+        }
+        double const J = std::sqrt(0.5 * val);
+        f = J / t_sigma_max_;
+    }
+    else
+    {
+        // von Mohr Coulomb failure criterion
+        f = 0.5 * std::fabs(sigma[2] - sigma[0]) /
+            (c_ * std::cos(phi_) - sigma_m * std::sin(phi_));
+    }
```
nagelt, line 108: @wenqing please help me understand your changes:

1. the apex value of sigma_m is c/tan phi ... based on what logic did you decide to make the distinction at exactly 0.5 that value in L90? If there's no reason, is it a magic number?
2. f_mc (sigma_m = 0.5 c/tan(phi) = (sigma_2 - sigma_1)/(c*cos(phi)) Is it ensured that f_vonMises = f_MC at this point? They should transition continuously. Isn't t_sigma_max_ is an independent parameter? So how would continuity be guaranteed?
3. With a transition to von Mises without an actual tension cutoff, low shear but high tension will still not lead to failure.

wenqing: > @wenqing please help me understand your changes:
> 
>     1. the apex value of sigma_m is c/tan phi ... based on what logic did you decide to make the distinction at exactly 0.5 that value in L90? If there's no reason, is it a magic number?
> 
>     2. f_mc (sigma_m = 0.5 c/tan(phi) = (sigma_2 - sigma_1)/(c*cos(phi)) Is it ensured that f_vonMises = f_MC at this point? They should transition continuously. Isn't t_sigma_max_ is an independent parameter? So how would continuity be guaranteed?
> 
>     3. With a transition to von Mises without an actual tension cutoff, low shear but high tension will still not lead to failure.

Because the criterion of
```
f_shear = tau_max / (c * cos_phi - min(sigma_m, tensile_strength_parameter) * sin_phi)
```
may still have the apex point if tensile_strength_parameter=c/tan(phi). My previous idea is that if sigma_m is greater than the the half of c/tan(phi) (half distance to the apex from 0, not good with that artificial conditon), use von Mises, otherwise use the Mohr-Coulomb.  That does not fit (2) and (3) of your comments. I will provide an improvement soon taking account of  your suggestions.

nagelt: I see ... I wouldn't worry too much about this aspect: tensile_strength_parameter=c/tan(phi). We perform no parameter consistency checks in any of our models, I believe. So if people want to enter false numbers, they can. At least we provide a well defined (documented) criterion and people can choose meaningful parameters.

nagelt: GIGO principle ;-)

wenqing: @nagelt  Just updated. The changes are:
1. Input name is changed to tensile_strength_parameter.
2. if (sigma_m > tensile_strength_parameter), compute f = max(f_MC, f_vM) as you suggested. Meanwhile, f = f_vM if sigma_m = c/tan(phi).
3. if  (sigma_m <= tensile_strength_parameter), f = f_MC.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,156 @@
...
+            c_ * std::cos(phi_) - t_sigma_max_ * std::sin(phi_);
+
+        if (fabs(std::fabs(tau_tt)) > std::numeric_limits<double>::epsilon())
+        {
+            f = std::max(f, tau_m / tau_tt);
+        }
+    }
+    else
+    {
+        // von Mohr Coulomb failure criterion
```
wenqing, line 107: @nagelt  The changes according to our talk are committed. Please take a look. Besides the class documentation was changed accordingly as well.

nagelt: looks good!

nagelt:
Otherwise no further comments at this stage from my side.

endJunction:
@wenqing Did you update all of the documentation to the current implementation?

Not sure, but the unit-test might need an update too.

wenqing:
> @wenqing Did you update all of the documentation to the current implementation?
> 
Already updated. Now changed one word in the class documentation.

> Not sure, but the unit-test might need an update too.
Already updated before.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,156 @@
...
+    double const max_sigma = std::max(std::fabs(sigma[0]), std::fabs(sigma[2]));
+
+    if (max_sigma < std::numeric_limits<double>::epsilon())
+    {
+        return fromVector(k_data);
+    }
+
+    double const sigma_m = 0.5 * (sigma[2] + sigma[0]);
+
+    double const tau_m = 0.5 * std::fabs(sigma[2] - sigma[0]);
```
endJunction, line 90: In the documentation (both of them) tau_m is simply the (sigma_2-sigma_0)/2 difference without abs. How come?

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,156 @@
...
+        const double exp_value = std::exp(b_ * f);
+        for (auto& k_i : k_data)
+        {
+            k_i = std::min(k_i + kr_ * exp_value, k_max_);
+        }
+    }
+
+    // Local coordinate transformation is only applied for the case that the
+    // initial intrinsic permeability is given with orthotropic assumption.
+    if (local_coordinate_system_ && (k_data.size() == DisplacementDim))
```
endJunction, line 122: Orthotropy is not mentioned in the documentation.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,156 @@
...
+    if (max_sigma < std::numeric_limits<double>::epsilon())
+    {
+        return fromVector(k_data);
+    }
+
+    double const sigma_m = 0.5 * (sigma[2] + sigma[0]);
+
+    double const tau_m = 0.5 * std::fabs(sigma[2] - sigma[0]);
+    double f = 0.0;
+    if (sigma_m > t_sigma_max_)
```
endJunction, line 92: Documentation says >=.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,156 @@
...
+    {
+        // tensile failure criterion
+        f = sigma_m / t_sigma_max_;
+
+        double const tau_tt =
+            c_ * std::cos(phi_) - t_sigma_max_ * std::sin(phi_);
+
+        if (fabs(std::fabs(tau_tt)) > std::numeric_limits<double>::epsilon())
```
endJunction, line 100: double fabs

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
endJunction, line 107: minor: von Mises, but Mohr Coulomb.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,156 @@
...
+        // von Mohr Coulomb failure criterion
+        f = tau_m / (c_ * std::cos(phi_) - sigma_m * std::sin(phi_));
+    }
+
+    if (f >= 1.0)
+    {
+        const double exp_value = std::exp(b_ * f);
+        for (auto& k_i : k_data)
+        {
+            k_i = std::min(k_i + kr_ * exp_value, k_max_);
```
endJunction, line 116: This minimum/maximum is not mentioned in the documentation.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
endJunction: Documentation does not say anything about the case when tau_tt > epsilon.

wenqing: removed.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
wenqing: Changed the doc.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
wenqing: Added a word for the member documentation to mention k can be a tensor.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
wenqing: Changed the doc.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
wenqing: Added some words about the upper bound.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
wenqing: dropped 'von'

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.h
```diff
@@ -0,0 +1,127 @@
...
+ * The failure index \f$f\f$ is calculated from  the Mohr Coulomb failure
+ * criterion comparing an acting shear stress for the shear dominated failure.
+ * The tensile failure is governed by an input parameter of
+ * tensile_strength_parameter .
+ *
+ *  The Mohr Coulomb failure
+ * criterion \cite labuz2012mohr takes
+ * the form
+ *   \f[\tau(\sigma)=c-\sigma \mathrm{tan} \phi\f]
```
endJunction, line 48: Other documentation says sigma_n. Writing only sigma looks like full stress tensor.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.h
```diff
@@ -0,0 +1,127 @@
...
+ * \f[
+ *   f =
+ *  \begin{cases}
+ *    f_{MC}, & \sigma_{m} \leq \sigma^t_{max}\\
+ *    max(f_{MC}, f_t), & \sigma_{m} > \sigma^t_{max}\\
+ *  \end{cases}
+ * \f]
+ *
+ *  The computed permeability components are restricted with an upper bound,
+ *  i.e. \f$\mathbf{k}:=k_{ij} < k_{max}\f$.
```
endJunction, line 80: But only for some specific case, not in all cases. And also before rotation.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.h
wenqing: They are the commonly use stress symbols for the MC model. I will copy the context to other doc. Normally in the mechanics notations, the stress tensor symbols take the bold type.

endJunction: That's fine. Choose one consistent notation for the pdf, doxygen, and the code. Makes life and error hunting easier.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.h
wenqing: It for the computed value of k=k0+H(f−1) k_r exp(bI). We do not need to write that restriction in the source code for H(f-1) =0 (f<1) for k=k0. Therefore it is a general expression of the restriction of the permeability of the model. The rotation (only needed for symmetrical k tensor) is a mapping between the two systems, and the restriction have to be applied in the original system, i.e. the local system.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.h
```diff
@@ -0,0 +1,127 @@
...
+ * criterion comparing an acting shear stress for the shear dominated failure.
+ * The tensile failure is governed by an input parameter of
+ * tensile_strength_parameter .
+ *
+ *  The Mohr Coulomb failure
+ * criterion \cite labuz2012mohr takes
+ * the form
+ *   \f[\tau(\sigma)=c-\sigma \mathrm{tan} \phi\f]
+ *   with \f$\tau\f$ the shear stress, \f$c\f$ the cohesion, \f$\sigma\f$ the
+ * tensile stress, and \f$\phi\f$ the internal friction angle.
```
endJunction, line 50:
```suggestion
 * normal stress, and \f$\phi\f$ the internal friction angle.
```

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
endJunction: Still open.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.h
endJunction: I was too fast to close it. I agree on the first part. But the documentation should mention, that the restriction is done before rotation and not after. The result is different, *e.g.* if k = Id and k_max is 0.5. Restrict and rotate yields different value than rotate and restrict.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,156 @@
...
+        Eigen::Matrix<double, DisplacementDim, DisplacementDim> k =
+            Eigen::Matrix<double, DisplacementDim, DisplacementDim>::Zero();
+
+        for (int i = 0; i < DisplacementDim; ++i)
+        {
+            Eigen::Matrix<double, DisplacementDim, DisplacementDim> const
+                ei_otimes_ei = e.col(i) * e.col(i).transpose();
+
+            k += k_data[i] * ei_otimes_ei;
+        }
```
endJunction, line 135: Rotation should be mentioned in the documentation. That k can be a tensor does not imply rotations.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
wenqing: Replaced it as a parameter checking in ctor.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.h
wenqing: The restriction is only for the model. The rotation is computational stuff. Anyway, added a sentence in the class documentation:
```
 * If \f$\mathbf{k}_0\f$ is orthogonal, i.e input two or three numbers
 * for the diagonal entries of \f$\mathbf{k}_0\f$, a coordinate system rotation
 * is possible if it is needed.
```

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
wenqing: Please see the reply to you last comment. 

If user inputs two or three numbers for the parameter, he knows that a rotation is possible as other permeability or conductivity input (this should be documented in the related rotation classes). If a user inputs one, four or nine numbers for the parameter ( i.e a full tensor), he definitely knows that there will be no rotation needed. This is the consideration about  the rotation in this implementation.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
endJunction: This is nice, that you move the constant check in the constructor, but the question was about documentation. Anyway, the users will find out the surprising situations.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
endJunction: Though I still think the rotations should be mentioned in the documentation it is a minor issue and we will see if the users rise more question, in which case additional documentation will be needed.

endJunction:
Don't forget the changelog ;)

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
wenqing: > Though I still think the rotations should be mentioned in the documentation it is a minor issue and we will see if the users rise more question, in which case additional documentation will be needed.

Is the added sentence in the class documentation about the rotation adequate?

wenqing:
> Don't forget the changelog ;)

Added.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
endJunction: This is question to the users.

In my opinion it is not describing the full picture. Especially the order of limiting to k_max and rotations is not mentioned, what influence the presence of `local_coordinate_system` has on the result.

wenqing:
rebased.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.cpp
wenqing: The restriction of the parameter was documented.

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.h
```diff
@@ -0,0 +1,131 @@
...
+{
+/**
+ * \brief A failure index dependent permeability model \cite wangetalPerm2020
+ *
+ * \f[ \mathbf{k} =
+ *      \mathbf{k}_0+ H(f-1) k_\text{r} \mathrm{e}^{b f}\mathbf{I}\f]
+ *
+ * where
+ *   \f$\mathbf{k}_0\f$ is the intrinsic permeability.
+ *  of the undamaged material,
```
TomFischer, line 34:
```suggestion
 *   \f$\mathbf{k}_0\f$ is the intrinsic permeability of the undamaged material,
```

Tests/MaterialLib/TestMPLPermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+    ParameterLib::SpatialPosition const pos;
+    double const t = std::numeric_limits<double>::quiet_NaN();
+    double const dt = std::numeric_limits<double>::quiet_NaN();
+
+    auto const k = MPL::formEigenTensor<3>(k_model.value(vars, pos, t, dt));
+
+    double const k_expected = 1.1398264890628033e-15;
+
+    ASSERT_LE(std::fabs((k_expected - k(0, 0))) / k_expected, 1e-10)
```
TomFischer, line 64:
```suggestion
    ASSERT_LE(std::fabs(k_expected - k(0, 0)) / k_expected, 1e-10)
```

Tests/MaterialLib/TestMPLPermeabilityMohrCoulombFailureIndexModel.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include "MaterialLib/MPL/Utils/FormEigenTensor.h"
+#include "MathLib/KelvinVector.h"
+#include "ParameterLib/ConstantParameter.h"
+#include "TestMPL.h"
+#include "Tests/TestTools.h"
+
+TEST(MaterialPropertyLib, PermeabilityMohrCoulombFailureIndexModel)
+{
+    ParameterLib::ConstantParameter<double> k0 =
+        ParameterLib::ConstantParameter<double>("k0", 1.e-20);
```
TomFischer, line 29:
```suggestion
    ParameterLib::ConstantParameter<double> k0("k0", 1.e-20);
```

endJunction: and const

TomFischer: Updated.

Tests/MaterialLib/TestMPLPermeabilityMohrCoulombFailureIndexModel.cpp
wenqing: Changed. https://github.com/ufz/ogs/pull/2989/commits/df4733f08f7e3f8fbbfe9f8c114868e27e35bd4a#diff-bb5a3a2f4bc492d8735683d3c0ef44a7R28

Tests/MaterialLib/TestMPLPermeabilityMohrCoulombFailureIndexModel.cpp
wenqing: Changed. https://github.com/ufz/ogs/pull/2989/commits/df4733f08f7e3f8fbbfe9f8c114868e27e35bd4a#diff-bb5a3a2f4bc492d8735683d3c0ef44a7R28

MaterialLib/MPL/Properties/PermeabilityMohrCoulombFailureIndexModel.h
wenqing: removed that period '.'

PR #2991, opened 2020-06-08
Add switch to function addLayerToMesh to copy the MaterialID.

TomFischer:
In some applications it is practical to adopt the material ID of the underlying layer.

### DIFF
--- a/Applications/DataExplorer/DataView/MeshView.cpp
+++ b/Applications/DataExplorer/DataView/MeshView.cpp
@@ -244,9 +244,11 @@ void MeshView::openAddLayerDialog()
         return;
     }
 
-    double const thickness (dlg.getThickness());
-    std::unique_ptr<MeshLib::Mesh> result(MeshLib::addLayerToMesh(
-        *mesh, thickness, dlg.getName(), dlg.isTopLayer()));
+    bool const copy_material_ids = false;
+    double const thickness(dlg.getThickness());
+    std::unique_ptr<MeshLib::Mesh> result(
+        MeshLib::addLayerToMesh(*mesh, thickness, dlg.getName(),
+                                dlg.isTopLayer(), copy_material_ids));
 
     if (result)
     {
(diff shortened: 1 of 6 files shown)


Applications/DataExplorer/DataView/MeshView.cpp
```diff
@@ -244,9 +244,11 @@ void MeshView::openAddLayerDialog()
         return;
     }
 
-    double const thickness (dlg.getThickness());
-    std::unique_ptr<MeshLib::Mesh> result(MeshLib::addLayerToMesh(
-        *mesh, thickness, dlg.getName(), dlg.isTopLayer()));
+    bool const take_over_material_ids = false;
```
endJunction, line 247: Take over the world, that's right. Here pass material ids, or copy material ids would be better, before we take over the world. It's a funny evening today ;)

MeshLib/MeshEditing/AddLayerToMesh.cpp
```diff
@@ -192,12 +194,25 @@ MeshLib::Mesh* addLayerToMesh(MeshLib::Mesh const& mesh, double thickness,
...
+        sfc_mesh->getProperties().existsPropertyVector<int>("MaterialIDs"))
+    {
+        auto const& sfc_material_ids =
+            *sfc_mesh->getProperties().getPropertyVector<int>("MaterialIDs");
+        std::copy(sfc_material_ids.cbegin(), sfc_material_ids.cend(),
+                  std::back_inserter(*new_materials));
+    }
+    else
+    {
+        int new_layer_id(
```
endJunction, line 211:
```suggestion
        int const new_layer_id(
```

Applications/DataExplorer/DataView/MeshView.cpp
TomFischer: Renamed as suggested.

MeshLib/MeshEditing/AddLayerToMesh.cpp
TomFischer: Const added.

endJunction:
If no other reviews, will be merged tonight.

PR #2994, opened 2020-06-09
Use STL algorithms instead of raw for-loops.

endJunction:
Manual changes on selected for-loops as suggested by cppcheck.

### DIFF
--- a/GeoLib/Polygon.cpp
+++ b/GeoLib/Polygon.cpp
@@ -230,20 +230,17 @@ bool Polygon::isPartOfPolylineInPolygon(const Polyline& ply) const
         }
     }
 
-    GeoLib::Point s;
-    for (auto polygon_seg : *this)
-    {
-        if (std::any_of(ply.begin(), ply.end(),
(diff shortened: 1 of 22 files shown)


GeoLib/Polygon.cpp
```diff
@@ -230,20 +230,17 @@ bool Polygon::isPartOfPolylineInPolygon(const Polyline& ply) const
...
+    auto polygon_segment_intersects_line = [&](auto const& polygon_seg) {
+        GeoLib::Point s;
+        return std::any_of(ply.begin(), ply.end(),
+                           [&polygon_seg, &s](auto const& polyline_seg) {
+                               return GeoLib::lineSegmentIntersect(
+                                   polyline_seg, polygon_seg, s);
+                           });
+    };
+
+    return any_of(std::cbegin(*this), std::cend(*this),
```
TomFischer, line 242:
```suggestion
    return std::any_of(std::cbegin(*this), std::cend(*this),
```

PR #2995, opened 2020-06-09
Benchmark for PermeabilityMohrCoulombFailureIndexModel

wenqing:
as titled. Temporally, this PR contains the commits of PR # #2989.

1. [ ] Feature description was added to the [changelog](https://github.com/ufz/ogs/wiki/Release-notes-6.3.1) Not yet.
2. [x] Tests covering your feature were added? Yes.
hm_perm

3. [ ] Any new feature or behavior change was documented? Not yet.


ProcessLib/RichardsMechanics/RichardsMechanicsFEM.h
```diff
@@ -65,10 +65,12 @@ class RichardsMechanicsLocalAssembler
                              ShapeMatricesTypePressure, DisplacementDim,
                              ShapeFunctionDisplacement::NPOINTS>;
 
-    static int const KelvinVectorSize =
+    static constexpr int const KelvinVectorSize =
```
endJunction, line 68: either const or constexpr, but both are not necessary, especially the static const is already a constant expression. If it wouldn't be, one could not use it as a template argument below.

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
```diff
@@ -221,15 +225,25 @@ void HydroMechanicsLocalAssembler<ShapeFunctionDisplacement,
...
+        vars[static_cast<int>(MPL::Variable::phase_pressure)] = p_int_pt;
 
         auto const K_S = solid_material.getBulkModulus(t, x_position);
 
+        auto const alpha = solid.property(MPL::PropertyType::biot_coefficient)
+                               .template value<double>(vars, x_position, t, dt);
+
+        // For stress dependent permeability.
+        // TODO (wenqing): get a model type, and use it to avoid the following
+        // assignment if no stress dependent permeability model is used.
```
endJunction, line 238: Just to get your idea, how would you do it? How would you find out if an arbitrary permeability model is stress dependent?

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
```diff
@@ -374,13 +385,27 @@ HydroMechanicsLocalAssembler<ShapeFunctionDisplacement, ShapeFunctionPressure,
...
+        vars[static_cast<int>(MPL::Variable::stress)].emplace<SymmetricTensor>(
+            MathLib::KelvinVector::kelvinVectorToSymmetricTensor(
+                static_cast<SymmetricTensor>(_ip_data[ip].sigma_eff -
+                                             alpha * identity2 * p_int_pt)));
```
endJunction, line 408: Why the cast?
The type of the sigma_eff*alpha*I*p is KelvinVector. Conversion to SymmetricTensor requires sqrt(2) multiplications.

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
wenqing: Can we add a member to Property to return its type?

endJunction: I don't quite get it. Would you elaborate?
So there will be a member in the Property class, let's say x(). It would return what?
How would you use the result here to find out if an *arbitrary* permeability model is stress dependent?

Tests/Data/HydroMechanics/FailureIndexDependentPermeability/quad_with_half_hole.prj
```diff
@@ -0,0 +1,272 @@
...
+                    <properties>
+                        <property>
+                            <name>viscosity</name>
+                            <type>Constant</type>
+                            <value>1e-3</value>
+                        </property>
+                        <property>
+                            <name>density</name>
+                            <type>Constant</type>
+                            <value>1</value>
```
endJunction, line 105: Density in the documentation is 1000. And also other values are not the same. This should be fixed.

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
wenqing: > Why the cast?

If only ``sigma_eff``, no cast is needed. `identity2` needs a cast.

> The type of the `sigma_eff*alpha*I*p` is KelvinVector. Conversion to SymmetricTensor requires sqrt(2) multiplications.

Please take a look at my new PR, #3010, about  kelvinVectorToSymmetricTensor. I assumed that function handles that multiplier correctly.

Tests/Data/HydroMechanics/FailureIndexDependentPermeability/quad_with_half_hole.prj
wenqing: Corrected. Since the gravity term in the Darcy is omitted, it only makes a tiny change in permeability and stresses, which cannot be distinguished in the contour plot.

ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h
wenqing: For example:
```
virtual bool Property::hasStressVariable() const {return false;}

 bool PermeabilityMohrCoulombFailureIndexModel::hasStressVariable() const override {return true;}
 ```

File: Tests/Data/HydroMechanics/FailureIndexDependentPermeability/quad_with_half_hole.prj (hunk 4 above)
Comment (endJunction) [reply Tests/Data/HydroMechanics/FailureIndexDependentPermeability/quad_with_half_hole.prj]: Did you check all of the other values?

File: ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h (hunk 3 above)
Comment (wenqing) [reply ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h:408]: As discussed in the closed PR #3010,  kelvinVectorToSymmetricTensor converts the Kelvin stress to conventional stress correctly.

File: Tests/Data/HydroMechanics/FailureIndexDependentPermeability/quad_with_half_hole.prj (hunk 4 above)
Comment (wenqing) [reply Tests/Data/HydroMechanics/FailureIndexDependentPermeability/quad_with_half_hole.prj]: Yes.
HM_k

File: ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h (hunk 2 above)
Comment (endJunction) [reply ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h:238]: This does not scale well, because for every new variable like stress etc. the Property interface has to be updated. So, then one would add a new enum value, two string conversions (back and forth), and additionally a test function and an if-condition to every process which is going to use it. Also how do you find out for MFront materials if there is dependency on a particular variable?
I'd drop the comment. Keeping things simple is an advantage. How much time would it save for each assembly?

File: ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h (hunk 3 above)
Comment (endJunction) [reply ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h:408]: The `kelvinVectorToSymmetricTensor` handles the sqrt(2) multiplier correctly. static_cast is wrong way to bend the types. Use `eval`.

File: Tests/Data/HydroMechanics/FailureIndexDependentPermeability/quad_with_half_hole.prj (hunk 4 above)
Comment (endJunction) [reply Tests/Data/HydroMechanics/FailureIndexDependentPermeability/quad_with_half_hole.prj]: what about solid density?

File: ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h (hunk 3 above)
Comment (wenqing) [reply ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h]: use .eval().

File: Tests/Data/HydroMechanics/FailureIndexDependentPermeability/quad_with_half_hole.prj (hunk 4 above)
Comment (wenqing) [reply Tests/Data/HydroMechanics/FailureIndexDependentPermeability/quad_with_half_hole.prj]: Changed to 2.65e+3.

File: ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h (hunk 2 above)
Comment (wenqing) [reply ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h:238]: The major computation is a two vector (with a size of 4 or 6) addition at the integration points. I have not compared the time increases in the local assembler for large problems. I think, it may increase the local assembly time for the maximum magnitude of 1%. 

An enum type might be good. This would be put to another PR if an if-condition for passing stress to Property is needed.

File: web/content/docs/benchmarks/hydro-mechanics/PermeabilityMohrCoulombFailureIndexModel.pandoc [hunk 5]
```diff
@@ -0,0 +1,24 @@
...
+
++++
+
+{{< data-link >}}
+
+For the description of the benchmark and its related permeability model,
+  please read [the PDF](../PermeabilityMohrCoulombFailureIndexModel.pdf).
+
+For a full document of the failure index dependent
+ permeability model, benchmark and its the implementation in ogs6, etc.,
```
wenqing, line 20: @OlafKolditz 
 The link to the document on Overleaf was add. Please see [Preview](https://deploy-preview-2995--ogs.netlify.app/docs/benchmarks/hydro-mechanics/permeabilitymohrcoulombfailureindexmodel/)

wenqing:
@endJunction: @nagelt  asked you a question in  [his comment](https://github.com/ufz/ogs/commit/1fa184fe9981f3247e3b96c8b5d14615dba9b455#commitcomment-40002981) about the comment by @OlafKolditz about use Overleaf in web/doc. These comments was detached due to the update of this PR.

wenqing:
> The documentation and the project file should describe exactly same problems.

Changes in doc:
1. added parameters for the permeability model,
2. added the description of the initial conditions of pore pressure and stress.
3. corrected a typo in row 2 of Table 2, tau_x=-15 MPa instead of \tau_y

Changes in prj:
porosity -> 0.15. The ref. result file and the figures for ths doc. were updated accordingly.

PR #2996, opened 2020-06-09
[MPL] Change exponential prop. to a*exp(m(x-x0)).

endJunction:
Previous version negates the exponent factor m internally, which is not obvious.


endJunction:
Needs another thoughtful review, please.

PR #2997, opened 2020-06-10
[T] RasterToMesh: Const, auto, unused uninit vars.

endJunction:
Triggered by fixing a warning of uninitialized spacing variable.

### DIFF
--- a/Tests/MeshLib/TestRasterToMesh.cpp
+++ b/Tests/MeshLib/TestRasterToMesh.cpp
@@ -49,15 +49,15 @@ protected:
 
 TEST_F(RasterToMeshTest, convertRasterToTriMeshElevation)
 {
-    std::unique_ptr<MeshLib::Mesh> mesh(MeshLib::RasterToMesh::convert(
+    auto const mesh = MeshLib::RasterToMesh::convert(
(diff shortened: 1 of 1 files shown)


Tests/MeshLib/TestRasterToMesh.cpp
```diff
@@ -270,26 +271,25 @@ TEST_F(RasterToMeshTest, convertRasterToQuadMeshNone)
 #ifdef OGS_BUILD_GUI
 TEST_F(RasterToMeshTest, vtkImage)
 {
-    double x0;
-    double y0;
-    double spacing;
-    vtkImageAlgorithm* raster = VtkRaster::loadImage(_file_name);
+    double const spacing = std::numeric_limits<double>::quite_NaN();
```
TomFischer, line 274: Is the correct header included for `quite_NaN`?

endJunction: There are uses of epsilon somewhere else in that file...
But those, who are mighty of proper spelling, don't get errors... :)

endJunction:
```suggestion
    double const spacing = std::numeric_limits<double>::quiet_NaN();
```

PR #2999, opened 2020-06-10
RM; pass solid pressure to density

endJunction:
Followup of #2996 

Required for solid pressure dependent density.

2. [x] Tests covering your feature were added? ctest with swelling was updated to include the new dependency


endJunction:
exponential_solid_density

dashed lines - constant solid density
solid lines - exponential solid density.

All but the dry densities plots are identical.
See benchmark itself for details. Short: mechanical load phase in the beginning, desaturation at the end with swelling happening.

ProcessLib/RichardsMechanics/RichardsMechanicsFEM-impl.h
```diff
@@ -803,6 +800,15 @@ void RichardsMechanicsLocalAssembler<ShapeFunctionDisplacement,
                 displacement_index, displacement_index)
             .noalias() += B.transpose() * C * B * w;
 
+        auto const phi = _ip_data[ip].porosity;
```
TomFischer, line 803: Why not:
```suggestion
        auto const& phi = _ip_data[ip].porosity;
```
?

ProcessLib/RichardsMechanics/RichardsMechanicsFEM-impl.h
```diff
@@ -1535,10 +1541,15 @@ void RichardsMechanicsLocalAssembler<ShapeFunctionDisplacement,
 
         GlobalDimMatrixType const K_over_mu = k_rel * K_intrinsic / mu;
 
+        auto const phi = _ip_data[ip].porosity;
```
TomFischer, line 1544: Same as above, why not
```suggestion
        auto const& phi = _ip_data[ip].porosity;
```

ProcessLib/RichardsMechanics/RichardsMechanicsFEM-impl.h
endJunction: Is a double.

ProcessLib/RichardsMechanics/RichardsMechanicsFEM-impl.h
```diff
@@ -803,6 +800,15 @@ void RichardsMechanicsLocalAssembler<ShapeFunctionDisplacement,
                 displacement_index, displacement_index)
             .noalias() += B.transpose() * C * B * w;
 
+        auto const phi = _ip_data[ip].porosity;
+        double const p_FR = -chi_S_L * p_cap_ip;
+        // p_SR
+        variables[static_cast<int>(MPL::Variable::solid_grain_pressure)] =
+            p_FR - (sigma_eff + sigma_sw).dot(identity2) / (3 * (1 - phi));
```
wenqing, line 807:
```suggestion
            p_FR - std::max(0.0, (sigma_eff + sigma_sw).dot(identity2) / (3 * (1 - phi)));
```

wenqing: or 
```
std::max(0.0, p_FR - (sigma_eff + sigma_sw).dot(identity2) / (3 * (1 - phi)));
```

endJunction: why?

wenqing: I do know what it is exactly the density model. If the density model describes the density change only under pressure, or compression, and no change in extension, all three principle stresses must be negative. If it is not the case, your implementation is roughly good, which means we do not care compressed or extended in directions but with a dominated status.

endJunction: The density model in the ctest is exponential one describing the change in both directions, compression and extension.

Tests/Data/RichardsMechanics/deformation_dependent_porosity_swelling.prj
```diff
@@ -72,8 +72,16 @@
...
+                            <type>Exponential</type>
+                            <reference_value>1</reference_value>
+                            <exponent>
+                                <variable_name>solid_grain_pressure</variable_name>
+                                <reference_condition>0</reference_condition>
+                                <!-- beta_SR = (1 - alpha)/K_S. K_S = E/(3(1-2 nu)) -->
+                                <!-- K_S = 1/(3*(1-2*0.3)) = 5/6. -->
+                                <!-- beta_SR = (1 - 11/12)/K_S = 0.1 -->
+                                <factor>0.1</factor>
+                            </exponent>
```
endJunction, line 84: @wenqing the used density model.

wenqing: I see.

PR #2998, opened 2020-06-10
[CL] Interpolate nodal process solutions to int_pt

renchao-lu:
Pull requests #2998, #3001, and #3003 make changes to the chemical solver interface. Pull request #2998 serves to interpolate nodal process solutions to integration points; pull request #3001 updates the PhreeqcIO interface to loop speciation calculation over all intergration points; pull request #3003 serves to extrapolate post-reaction solutions to nodes. Previously, chemical calculation is carried out by node-wise. With the above pull requests, chemical calculation will be done on the integration points.

1. [x] Feature description was added to the [changelog](https://github.com/ufz/ogs/wiki/Release-notes-6.3.1)
2. [x] Tests covering your feature were added?
3. [ ] Any new feature or behavior change was documented?

PR #3003, opened 2020-06-15
[CL] Extrapolate post reaction solution to nodes

renchao-lu:
This pull requests follows up #2998 and #3001.

### DIFF
--- a/ProcessLib/ComponentTransport/ComponentTransportProcess.cpp
+++ b/ProcessLib/ComponentTransport/ComponentTransportProcess.cpp
@@ -174,6 +174,32 @@ void ComponentTransportProcess::
         _local_assemblers, pv.getActiveElementIDs(), _coupled_solutions);
 }
 
+void ComponentTransportProcess::extrapolateIntegrationPointValuesToNodes(
+    const double t, std::vector<GlobalVector*> const& int_pt_x,
+    std::vector<GlobalVector*>& x)
+{
+    auto& extrapolator = getExtrapolator();
+    auto const extrapolatables =
+        NumLib::makeExtrapolatable(_local_assemblers,
+                                   &ComponentTransportLocalAssemblerInterface::
+                                       getInterpolatedLocalSolution);
+
+    for (unsigned transport_process_id = 0;
+         transport_process_id < int_pt_x.size();
+         ++transport_process_id)
+    {
+        auto const& pv = _process_variables[transport_process_id][0].get();
+        auto const& int_pt_C = int_pt_x[transport_process_id];
+
+        extrapolator.extrapolate(pv.getNumberOfComponents(), extrapolatables, t,
+                                 {int_pt_C},
+                                 {_local_to_global_index_map.get()});
+
+        auto const& nodal_values = extrapolator.getNodalValues();
+        MathLib::LinAlg::copy(nodal_values, *x[transport_process_id + 1]);
+    }
+}
+
 void ComponentTransportProcess::preTimestepConcreteProcess(
     std::vector<GlobalVector*> const& x, const double /*t*/,
     const double /*delta_t*/, int const process_id)
--- a/ProcessLib/Process.h
+++ b/ProcessLib/Process.h
@@ -100,6 +100,14 @@ public:
         int const /*process_id*/)
     {
     }
+
+    virtual void extrapolateIntegrationPointValuesToNodes(
+        const double /*t*/,
+        std::vector<GlobalVector*> const& /*integration_point_values_vectors*/,
+        std::vector<GlobalVector*>& /*nodal_values_vectors*/)
+    {
+    }
+
     void preAssemble(const double t, double const dt,
                      GlobalVector const& x) final;
     void assemble(const double t, double const dt,
--- a/ProcessLib/ComponentTransport/ChemicalProcessData.h
+++ b/ProcessLib/ComponentTransport/ChemicalProcessData.h
@@ -1,0 +1,32 @@
+/**
+ * \file
+ * \copyright
+ * Copyright (c) 2012-2020, OpenGeoSys Community (http://www.opengeosys.org)
+ *            Distributed under a Modified BSD License.
+ *              See accompanying file LICENSE.txt or
+ *              http://www.opengeosys.org/project/license
+ *
+ */
+
+#pragma once
+
+#include <vector>
+
+#include "MathLib/LinAlg/GlobalMatrixVectorTypes.h"
+
+namespace ProcessLib
+{
+namespace ComponentTransport
+{
+struct ChemicalProcessData
+{
+    ChemicalProcessData(
+        std::vector<std::vector<GlobalIndexType>>& chemical_system_index_map_)
+        : chemical_system_index_map(chemical_system_index_map_)
+    {
+    }
+
+    std::vector<std::vector<GlobalIndexType>>& chemical_system_index_map;
+};
+}  // namespace ComponentTransport
+}  // namespace ProcessLib
--- a/ProcessLib/ComponentTransport/ComponentTransportFEM.h
+++ b/ProcessLib/ComponentTransport/ComponentTransportFEM.h
@@ -70,6 +70,12 @@ public:
         std::vector<NumLib::LocalToGlobalIndexMap const*> const& dof_table,
         std::vector<double>& cache) const = 0;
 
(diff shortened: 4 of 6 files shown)


renchao-lu:
Extracted post-reaction extrapolator and added a unit test

NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.cpp
```diff
@@ -103,6 +103,68 @@ void LocalLinearLeastSquaresExtrapolator::extrapolate(
...
+    {  // Create num_components times version of ghost_indices arranged by
+       // location. For example for 3 components and ghost_indices {5,6,10} we
+       // compute {15, 16, 17,  18, 19, 20,  30, 31, 32}.
+        auto const& single_component_ghost_indices =
+            _dof_table_single_component.getGhostIndices();
+        auto const single_component_ghost_indices_size =
+            single_component_ghost_indices.size();
+        ghost_indices.reserve(single_component_ghost_indices_size *
+                              num_components);
+        for (unsigned i = 0; i < single_component_ghost_indices_size; ++i)
```
wenqing, line 125:
```suggestion
        for (std::size_t i = 0; i < single_component_ghost_indices_size; ++i)
```

NumLib/Extrapolation/Extrapolator.h
```diff
@@ -33,6 +33,15 @@ class Extrapolator
         std::vector<GlobalVector*> const& x,
         std::vector<NumLib::LocalToGlobalIndexMap const*> const& dof_table) = 0;
 
+    //! Extrapolates post-reaction process solutions from the given local
+    //! assemblers.
```
endJunction, line 37: It's better if NumLib does not include any process specific nomenclature, like post-reaction

NumLib/Extrapolation/Extrapolator.h
```diff
@@ -33,6 +33,15 @@ class Extrapolator
...
+    virtual void extrapolate(
+        const unsigned num_components,
+        ExtrapolatableElementCollection const& extrapolatables,
+        GlobalVector const& int_pt_x,
+        std::vector<std::vector<GlobalIndexType>>&
+            chemical_system_index_map) = 0;
```
endJunction, line 43: Same remark here: chemical is not part of NumLib.

NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.cpp
```diff
@@ -103,6 +103,68 @@ void LocalLinearLeastSquaresExtrapolator::extrapolate(
...
+    for (std::size_t i = 0; i < size; ++i)
+    {
+        extrapolateElement(i, num_components, extrapolatables, int_pt_x,
+                           chemical_system_index_map, *counts);
+    }
+
+    MathLib::LinAlg::finalizeAssembly(*_nodal_values);
+
+    MathLib::LinAlg::componentwiseDivide(*_nodal_values, *_nodal_values,
+                                         *counts);
```
endJunction, line 165: This part but for one line is equal to the existing implementation. This is a lot of code duplication.

NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.cpp
```diff
@@ -271,6 +333,102 @@ void LocalLinearLeastSquaresExtrapolator::extrapolateElement(
...
+    const unsigned num_components,
+    ExtrapolatableElementCollection const& extrapolatables,
+    GlobalVector const& int_pt_x,
+    std::vector<std::vector<GlobalIndexType>>& chemical_system_index_map,
+    GlobalVector& counts)
+{
+    assert(num_components == 1);
+    _integration_point_values_cache.clear();
+    auto const& indices = chemical_system_index_map[element_index];
+    _integration_point_values_cache = int_pt_x.get(indices);
```
endJunction, line 347: Better to use `getIntegrationPointValues` interface of extrapolatables to get the values, instead of passing a single vector of those values.
In the process you have a reference to the `GlobalVector int_pt_x` and you get the element number through the `getIntegrationPointValues` interface. From these two and the "chemical_system_index_map" (which must be part of the process, otherwise you couldn't pass it here) you get the `_integration_point_values_cache` filled and all of the duplication in the NumLib is not necessary.

renchao-lu: I was looking for a workaround to get rid of duplications, but no good idea comes up with me. Thanks for the thoughtful suggestion! I will redo the extrapolation in the way you suggested.

NumLib/Extrapolation/LocalLinearLeastSquaresExtrapolator.cpp
renchao-lu: Reimplemented.

renchao-lu:
> Better to use `getIntegrationPointValues` interface of extrapolatables to get the values, instead of passing a single vector of those values.
> In the process you have a reference to the `GlobalVector int_pt_x` and you get the element number through the `getIntegrationPointValues` interface. From these two and the "chemical_system_index_map" (which must be part of the process, otherwise you couldn't pass it here) you get the `_integration_point_values_cache` filled and all of the duplication in the NumLib is not necessary.

Reimplemented. I think current implementation is much better than before.

ProcessLib/ComponentTransport/ComponentTransportProcess.cpp
```diff
@@ -174,6 +174,32 @@ void ComponentTransportProcess::
...
+void ComponentTransportProcess::extrapolateIntPtProcessSolutions(
+    const double t, std::vector<GlobalVector*>& x,
+    std::vector<GlobalVector*> const int_pt_x)
+{
+    auto& extrapolator = getExtrapolator();
+    auto const extrapolatables = NumLib::makeExtrapolatable(
+        _local_assemblers,
+        &ComponentTransportLocalAssemblerInterface::getIntPtLocalSolution);
+
+    assert(_process_data.chemical_process_data);
```
endJunction, line 186: The variable is not used in the function. Why checking?

ProcessLib/Process.h
```diff
@@ -100,6 +100,14 @@ class Process
         int const /*process_id*/)
     {
     }
+
+    virtual void extrapolateIntPtProcessSolutions(
+        const double /*t*/,
+        std::vector<GlobalVector*>& /*process_solutions*/,
+        std::vector<GlobalVector*> const /*transport_solutions*/)
```
endJunction, line 107: Not clear from the arguments' names who is who. Together with the function name it is not clear what is being extrapolated and what the result will be. Also same applies here, that in the general process interface there should be no names specific to some process.
A suggestion, though you can come up with a better names:
`extrapolateIntegrationPointValuesToNodes(t, integration_point_values_vectors, node_values_vectors)`

Also place the output variables at the end.

Second argument should be const&.

ProcessLib/ComponentTransport/ComponentTransportProcess.cpp
renchao-lu: removed.

ProcessLib/Process.h
renchao-lu: Fixed.

ProcessLib/ComponentTransport/ComponentTransportProcess.cpp
```diff
@@ -174,6 +174,32 @@ void ComponentTransportProcess::
...
+    {
+        auto const& pv = _process_variables[transport_process_id][0].get();
+        auto const& int_pt_C = int_pt_x[transport_process_id];
+
+        extrapolator.extrapolate(pv.getNumberOfComponents(), extrapolatables, t,
+                                 {int_pt_C},
+                                 {_local_to_global_index_map.get()});
+
+        auto const& nodal_values = extrapolator.getNodalValues();
+        *x[transport_process_id + 1] = nodal_values;
```
wenqing, line 199: Since x is a distributed vector under MPI, here you have to use LinAlg::copy(...), which is declared  in MathLib/LinAlg/LinAlg.h. With that, the compilation error under MPI should go.

renchao-lu: Let me have a try. Thanks, Wenqing!

renchao-lu: it works out.

PR #3010, opened 2020-06-18
[Math/KelvinVector] Added an argument of double factor to kelvinVectorToSymmetricTensor

wenqing:
As titled. The default value for that factor is 1/sqrt(2). I wonder that  the factor in the original implementation is incorrect, which should be sqrt(2) instead of 1/sqrt(2.0). If it is true, this PR  just for an correction of that.  

Function  kelvinVectorToSymmetricTensor is used in
[HydroMechanicsLocalAssembler::getEpsilon](https://github.com/ufz/ogs/blob/master/ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h#L788-L792) for strain output. 
and
[HydroMechanicsLocalAssembler::computeSecondaryVariableConcrete](https://github.com/ufz/ogs/blob/master/ProcessLib/HydroMechanics/HydroMechanicsFEM-impl.h#L840-L856) for the principle stress calculation.

### DIFF
--- a/MathLib/KelvinVector.cpp
+++ b/MathLib/KelvinVector.cpp
@@ -140,20 +140,21 @@ KelvinVectorType<3> tensorToKelvin<3>(Eigen::Matrix<double, 3, 3> const& m)
 
 template <>
 Eigen::Matrix<double, 4, 1> kelvinVectorToSymmetricTensor(
-    Eigen::Matrix<double, 4, 1, Eigen::ColMajor, 4, 1> const& v)
+    Eigen::Matrix<double, 4, 1, Eigen::ColMajor, 4, 1> const& v,
+    double const factor)
 {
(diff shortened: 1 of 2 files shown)


endJunction:
I don't quite get the intentions. The mapping from Kelvin vector to tensors is unique and the implementation is correct. There is no ambiguity where to put the sqrt(2).
See (19) and (20) in https://arxiv.org/pdf/1605.09606.pdf

wenqing:
> I don't quite get the intentions. The mapping from Kelvin vector to tensors is unique and the implementation is correct. There is no ambiguity where to put the sqrt(2).
> See (19) and (20) in https://arxiv.org/pdf/1605.09606.pdf

I see that shear stresss and strains are multiplied with sqrt(2). Thanks. Then the stress passing in   PR #2995 is correct as well.

PR #3011, opened 2020-06-18
Use <cmath> constants of sqrt(2) and 1/sqrt(2) for the Kelvin vector related computation 

wenqing:
As titled.

### DIFF
--- a/MaterialLib/SolidModels/Ehlers.cpp
+++ b/MaterialLib/SolidModels/Ehlers.cpp
@@ -8,10 +8,14 @@
  */
 
 #include "Ehlers.h"
+
 #include <boost/math/special_functions/pow.hpp>
-#include "MathLib/LinAlg/Eigen/EigenMapTools.h"
+// to use math constants of <cmath>, e.g M_SQRT1_2: 1/sqrt(2)
+#define _USE_MATH_DEFINES
+#include <cmath>
 
 #include "LinearElasticIsotropic.h"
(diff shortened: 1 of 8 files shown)


endJunction:
What is the performance benefit?

endJunction:
And the boost library already contains all of this constants and many more, is very well maintained and tested. Why?

endJunction:
Can you show the machine dependent difference, please, which is resolved by adding name aliases?

wenqing:
> What is the performance benefit?

At least, not compute sqrt(2).

wenqing:
> And the boost library already contains all of this constants and many more, is very well maintained and tested. Why?

It uses boost constants but with name alias. If std have constants, we just change that that name aliases. Anyway, the name alias file, MathConstant.h, can be dropped, and this PR is only for using boost::root_two.

wenqing:
> Can you show the machine dependent difference, please, which is resolved by adding name aliases?

See my reply above.

TomFischer:
> > What is the performance benefit?
> 
> At least, not compute sqrt(2).

Is this constant expression really evaluated at run time?

wenqing:
> > > What is the performance benefit?
> > 
> > 
> > At least, not compute sqrt(2).
> 
> Is this constant expression really evaluated at run time?

I am not sure that different compilers, or different compilers on difference OSs give the same sqrt(2).I 

I will change this PR for use <cmath> constant of  M_SQRT2.

